Issue #1039, opened 2022-03-06
Slack Status - Set yourself as active/away


MohamedISoliman:
# Extension – [Slack status](https://github.com/raycast/extensions/tree/56eada530324f6dc839f976d92ecc77d7b6e40a1/extensions/slack-status)

Author: @PitNikola

As a Raycast user, I want to have the ability to change the Slack status with an active/away toggle.

**Currently only the status changes without the active/away green dot**

**Who will benefit from this feature?**
Everyone using slack hopefully

xswirelab:
For the implementer:

**API endpoint:** `https://slack.com/api/users.profile.set`
```bash
#!/bin/bash
# Api call to set the status(like it works now)
curl 'https://slack.com/api/users.profile.set' \
--header "Authorization: Bearer $SLACK_TOKEN" \
--header 'Content-Type: application/json; charset=utf-8' \
--data-raw "{
   \"profile\": {
        \"status_text\": \"404\",
        \"status_emoji\": \":coffee:\",
        \"status_expiration\": 0
    }
}"
```

**API endpoint:** `https://slack.com/api/users.setPresence`
```bash
#!/bin/bash
# Api call to set the presence
curl 'https://slack.com/api/users.setPresence' \
--header "Authorization: Bearer $SLACK_TOKEN" \
--header 'Content-Type: application/json; charset=utf-8' \
--data-raw "{
   \"presence\": \"away\"
}"
```

The `presence` parameter takes two options:
- `away` (string)
- `auto` (string)

The presence API also needs you to update the `OAUTH2 scope`. You will need to add an extra permission:
- `users.write`

Looking forward to the update.

PitNikola:
Good suggestion, thank! I'll keep it in mind when I nave time to update the extension.

MohamedISoliman:
@PitNikola any updates on this

PitNikola:
Not yet. Unfortunately, haven't found time to look into it. The trickiest thing is to gracefully handle missing scopes and prompt users to update tokens without blocking them from using existing functionality. There is [another PR](https://github.com/raycast/extensions/pull/1330) for adding "pause notifications" checkbox which would require another scope update, it might be good to bundle it all together.
Not sure when I'll find time for it though. If anybody wants to contribute, it would be much appreciated.

PR #1330, opened 2022-04-08
[slack-status] Added "pause notification" support

dotlouis:
## Description

Added ability to snooze slack notification when setting status. (notice the "pause notification" label)

Screenshot 2022-04-08 at 13 01 29

A simple check box inspired by Slack's own UI

Screenshot 2022-04-08 at 13 03 12

Define the behavior in your preset or in a custom status

Screenshot 2022-04-08 at 13 01 53

## Behavior

- The snooze duration is based on the duration defined for the statuses
- Snooze duration max out at 24h.
- Clearing the status also clears snooze.

## Technical changes

- I basically tapped into the [dnd Slack API](https://api.slack.com/methods/dnd.setSnooze)
- The required scopes are updated to reflect that

## Checklist

### DIFF
--- a/extensions/slack-status/README.md
+++ b/extensions/slack-status/README.md
@@ -24,7 +24,9 @@ Extension for quickly changing your Slack status. Set up presets for common stat
             "user": [
                 "emoji:read",
                 "users.profile:write",
-                "users.profile:read"
+                "users.profile:read",
+                "dnd:read",
+                "dnd:write"
             ]
         }
     },
--- a/extensions/slack-status/src/defaultStatuses.ts
+++ b/extensions/slack-status/src/defaultStatuses.ts
@@ -5,11 +5,13 @@ export const defaultStatuses: SlackStatusPreset[] = [
     title: "Focus Mode",
     emojiCode: ":technologist:",
     defaultDuration: 120,
+    dnd: true,
   },
   {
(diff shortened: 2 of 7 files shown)


PitNikola:
Hey @dotlouis, thanks for the addition!

First thing I noticed is that it breaks current status for existing users because of the missing scope. This is not great, because all the users will be facing it without understanding what happened. They will be forced to go and update their Slack app which I think will require approval again.

CleanShot 2022-04-13 at 07 35 23

Ideally, we should have a more graceful way to deal with it. E.g. trying fetching dnd status, and if received `missing_scope`, add new section at the top with list item that would say "Auth scope update required" which would open detail page explaining what user should do. And until we confirm that new scope is set, the extension continues to work as before. I know it's a bit of a hassle, but I really want to avoid users being faced with suddenly broken extension.

This is outside of the scope of this PR, but I was wondering if you would be interested in also adding [this feature](https://github.com/raycast/extensions/issues/1039). Looks like it will also require scope changes so it might be best to put it all together.

dotlouis:
Hi @PitNikola,
You're right I completely missed the upgrade scenario. Thank you for the suggestion. I'll fix it this weekend :)
Regarding the active/away feature, I'll take a look but I'm not sure how to integrate it into the status in a clear and simple manner.

PR #1348, opened 2022-04-09
Add Folder Bomb Extension

koinzhang:
## Description

# Folder Bomb

Quickly take files out and move them to directory where parent folder is located.

**Command**:

- **Move File out**: Move File out of the parent folder.
- **Copy File out**: Copy File out of the parent folder.

***Warning***: (When using the  extension, please note the following points)

- ⚠️You can hide the Raycast interface. But don't quit the extension when using it to move/copy files. (Quitting the extension will cause the interruption of moving/copying files to fail)
- ⚠️Do not use shortcut keys to call other commands or extensions, this behavior will lead to quit this extension. (This will result in quitting this extension)
- ⚠️The operation of moving files with this extension cannot be reversed, so please operate with caution.

**Preference**:

- [ ] Open Destination Directory: After successfully moving out the files, open the destination folder.
- [ ] Delete Empty Directory: After successfully moving out the files, delete the original folder if the folder is empty.

**Notes on use**:

- Because of UI limitations, when detecting duplicate files, only the "Cancel" and "Overwrite All" options are provided.

- This folder will be ignored when it has the same name as the parent folder.

⚠️**Caution**: The file cannot be moved to the /User directory.

## Screencast

https://user-images.githubusercontent.com/36128970/162559744-1a24de2f-4e1d-410b-a0d6-3b3e764beab4.mp4 

## Checklist

### DIFF
--- a/extensions/folder-bomb/package.json
+++ b/extensions/folder-bomb/package.json
@@ -2,7 +2,7 @@
   "$schema": "https://www.raycast.com/schemas/extension.json",
   "name": "folder-bomb",
   "title": "Folder Bomb",
-  "description": "Quickly take files out and move them to directory where parent folder is located.",
+  "description": "Takes selected content from Finder and moves it to the directory where parent folder is located.",
   "icon": "extension-icon.png",
   "author": "koinzhang",
   "categories": [
@@ -13,16 +13,18 @@
   "commands": [
     {
       "name": "move-file-out",
-      "title": "Move File out",
+      "title": "Move Selected Item to Parent Folder",
+      "subtitle": "Folder Bomb",
       "icon": "move-file-out.png",
-      "description": "Quickly take files out of a folder and move them to the parent directory.",
+      "description": "Takes selected content and moves it to the parent directory.",
       "mode": "no-view"
     },
     {
       "name": "copy-file-out",
-      "title": "Copy File out",
+      "title": "Copy Selected Item to Parent Folder",
+      "subtitle": "Folder Bomb",
       "icon": "copy-file-out.png",
-      "description": "Quickly copy files to the parent directory.",
+      "description": "Copies selected content to the parent directory.",
       "mode": "no-view"
     }
   ],
--- a/extensions/folder-bomb/README.md
+++ b/extensions/folder-bomb/README.md
@@ -1,11 +1,11 @@
 # Move File out
 
-Quickly take files out and move them to directory where parent folder is located.
+Takes selected content from Finder and moves it to the directory where parent folder is located.
 
 **Command**:
 
-- **Move File out**: Move File out of the parent folder.
-- **Copy File out**: Copy File out of the parent folder.
+- **Move File out**: Takes selected content and moves it to the parent directory.
+- **Copy File out**: Copies selected content to the parent directory.
 
 
 


extensions/folder-bomb/package.json
```diff
@@ -0,0 +1,70 @@
...
+  "author": "koinzhang",
+  "categories": [
+    "Productivity",
+    "System"
+  ],
+  "license": "MIT",
+  "commands": [
+    {
+      "name": "move-file-out",
+      "title": "Move File out",
```
mattisssa, line 16: I'm not sure, but "Move File Out" doesn't really mean something to me that I understand what the command will do 😬 
Let's probably call the command:
```suggestion
      "title": "Move Selected Item to Parent Folder",
      "subtitle": "Folder Bomb",
```

extensions/folder-bomb/package.json
```diff
@@ -0,0 +1,70 @@
...
+      "icon": "move-file-out.png",
+      "description": "Quickly take files out of a folder and move them to the parent directory.",
```
mattisssa, line 18:
```suggestion
      "description": "Takes selected content and moves it to the parent directory.",
```

extensions/folder-bomb/package.json
```diff
@@ -0,0 +1,70 @@
...
+      "mode": "no-view"
+    },
+    {
+      "name": "copy-file-out",
+      "title": "Copy File out",
```
mattisssa, line 23:
```suggestion
      "title": "Copy Selected Item to Parent",
      "subtitle": "Folder Bomb",
```

extensions/folder-bomb/package.json
```diff
@@ -0,0 +1,70 @@
...
+      "icon": "copy-file-out.png",
+      "description": "Quickly copy files to the parent directory.",
```
mattisssa, line 25:
```suggestion
      "description": "Copies selected content to the parent directory.",
```

extensions/folder-bomb/src/move-file-out.ts
```diff
@@ -0,0 +1,7 @@
+import { ActionType, getItemAndSend } from "./send-file-utils";
+import { showToast, Toast } from "@raycast/api";
+
+export default async () => {
+  await showToast(Toast.Style.Animated, `Moving... Don't quit.`);
```
mattisssa, line 5: We shouldn't really force end-users to wait until something is happening in an extension. Read my comment about background tasks.

mattisssa (commented):
Hey! Thanks for the contribution! 🚀 
While doing the review I found a few things that I mention in the comments below.

But what bothers me more is the general approach to solving the problem with files moving. Asking end-users to keep the extension open while there is some activity happening (file movements in your case) is not something that they expect from Raycast. In your extension, if an end-user will close the extension while there are file movements in progress what's going to happen? As I could reproduce – some files get lost. 

In general, to avoid all the warnings to end-users I'd probably recommend keeping the extension in a fork from our repo or in a separate branch for a while. 
We plan to support background activities somewhere later that perfectly suit your extension.

extensions/folder-bomb/src/move-file-out.ts
koinzhang: Hi @mattisssa Thanks for the review, I've made all the changes to the text ambiguity.

> In general, to avoid all the warnings to end-users I'd probably recommend keeping the extension in a fork from our repo or in a separate branch for a while.
As for the part you mentioned, is it that I have to close this PR and resubmit a PR to another branch?

mattisssa:
Thank you for the updates! 👏 

If it's ok for you to wait until we introduce background tasks, 
What developers usually do – they create a forked repo and keep extensions there. But the extension won't be included in the Store of course.

--- 

Another thought I have is how we can actually have this extension is probably allow to commands proceed files/folders that less than _X_ size. To avoid this messy state when some files get lost if an end-user closed the command. And remove these warnings about "Don't close the command" and bla. Changing them just into the "Moving files..." message.

In my case – I closed the command while there was a moving folder with 36 GBs size, later I looked at the moved folder and there was some content missing.

mattisssa:
Yeah, thanks for understanding 🙌
At least you can share the branch later in the community and those who really needs the extension can use it by installing it manually. 

If we figure out what can be better than just waiting for the background tasks API we will reach you out here or in Slack! 😉

Feel free to reopen the PR if you will have another solution for this.

PR #1350, opened 2022-04-09
Iconify: add "search all icons" command

Destiner:
## Description

Adds a new command, "Search All Icons", to the Iconify extension. This command runs the query against all icon sets supported by Iconify.

## Screencast

Screenshot 2022-04-09 at 14 13 21

## Checklist

### DIFF
--- a/extensions/iconify/src/search-icons.tsx
+++ b/extensions/iconify/src/search-icons.tsx
@@ -14,7 +14,6 @@ function Command() {
   async function queryIcons(text: string) {
     setQuery(text);
     setLoading(true);
-    setIcons([]);
     const icons = await service.queryIcons(text);
     setIcons(icons);
     setLoading(false);


extensions/iconify/package.json
```diff
@@ -22,10 +22,17 @@
       "subtitle": "Iconify",
       "description": "View available icons",
       "mode": "view"
+    },
+    {
+      "name": "search-icons",
```
sxn, line 27: would it make sense to just add an `All` item in the `searchBarAccessory` of the other command? 🤔

extensions/iconify/src/search-icons.tsx
```diff
@@ -0,0 +1,50 @@
...
+    setLoading(true);
+    setIcons([]);
+    const icons = await service.queryIcons(text);
+    setIcons(icons);
+    setLoading(false);
+  }
+
+  return (
+    <List throttle isLoading={isLoading} onSearchTextChange={queryIcons}>
+      {icons.map((icon) => {
```
sxn, line 23: Mind adding an [EmptyView](https://developers.raycast.com/api-reference/user-interface/list#list.emptyview) for when there are no items to show? It's a bit nicer than greeting users with a `No Results` page 😄

extensions/iconify/package.json
Destiner: Hmm. Possible, but dunno. They behave differently both from the implementation and UX perspective.

extensions/iconify/src/search-icons.tsx
Destiner: 👍 Good one, will add in a bit.

Destiner: Added in [41687ff](https://github.com/raycast/extensions/pull/1350/commits/41687ff37950d0771ba70c46b85881f6ef7bf537)

sxn:
🎉

PR #1363, opened 2022-04-11
[Bitwarden] Add password history and improve generation

jomifepe:
### History

The password history is an important feature, which bitwarden has on the browser, mobile, and others. It's important to implement this feature with some security, given we're dealing with sensitive data, so feel free to analyze the security of the implementation.

**If bitwarden ever implements password history on their CLI, this implementation should be changed to use that.**

Implementation details:

- History is written on an external file using Node APIs;
- Content is encrypted using the `aes-256` algorithm, and saved alongside the random initialization vector that was used (this is a safe practice, more info [here](https://security.stackexchange.com/questions/12071/how-to-store-iv-and-key-temporarily-but-securely/12076#12076));
- The encryption is done using the `client_secret`, which the user specifies on the extension settings;
- The name of the history file is a hash of the `client_id`, which means that, if you change your account, a new file will be create, but if you go back to the previous one, and didn't erase the files, the password history will still be there;
- You are able to clear your history with an action;
- Some keywords (like today, yesterday, passphrase, etc) can be used to search passwords on the list.

<details>
  <summary>🎥  Demonstration</summary>
  
https://user-images.githubusercontent.com/7235666/162644451-f978166b-314c-4ea9-b668-d3b41c7916c6.mp4
</details>

### Generator bug fixing

The first iteration of the password generation has a bug. When you changed the options, it generated a new password, and if you didn't wait for it to finish, and changed another option, it would ignore it and the result would be incorrect. In this PR, the previous action is aborted when you trigger a new one, using an `AbortController`.

<details>
  <summary>🎥 Demonstration</summary>
  
https://user-images.githubusercontent.com/7235666/162644727-039282bd-0be1-4747-9393-929e56dc7149.mp4
</details>

## Checklist

### DIFF
--- a/extensions/bitwarden/src/generate-password.tsx
+++ b/extensions/bitwarden/src/generate-password.tsx
@@ -245,6 +245,12 @@ function PasswordHistory() {
                   onAction={handleClear}
                   shortcut={{ key: "backspace", modifiers: ["cmd", "shift"] }}
                 />
+                <Action.Paste
+                  title="Paste password to active app"
+                  icon={Icon.Text}
+                  content={password}
+                  shortcut={{ key: "enter", modifiers: ["cmd", "shift"] }}
+                />
               </ActionPanel>
             }
           />


PSalant726:
I agree @pomdtr - the password history implementation is not a necessary feature. Also, if added, it creates the potential for new attack vectors. IMO, this extension does not need to be a 1-1 replacement for all Bitwarden platforms, and wrapping the CLI/API features is plenty to support the most common use cases.

@jomifepe I definitely appreciate your effort to build a working implementation, but given the danger and complexity, I don't believe it should be included in this extension until Bitwarden supports it natively via their CLI and/or API. Also, in the future, I'd recommend opening separate PRs with the bug fix (password generation) and the feature addition (password history). As feedback on your implementation, I would only recommend against reusing the client secret as your AES encryption key, as it's stored within the extension's `localStorage`, and potentially accessible to the filesystem (although it definitely shouldn't be).

jomifepe:
@pomdtr @PSalant726 Thank you for the feedback. I totally understand the concerns, I was also hesitant to develop it, because I don't want to be responsible for a possible breach of security for bitwarden users. Although, I believe it's an important feature to have to avoid losing passwords. I'm also confident in the implementation, as I'm using the same algorithm as bitwarden (`aes-256-cbc`), and I'm not storing any sensitive key (without encryption) in the disk. 

@PSalant726 Answering the `localStorage` recommendation. The `client_secret` is being retrieved directly from the built-in `getPreferenceValues` method. I believe this is the correct way, and raycast seems to take care of the [isolation of the process](https://developers.raycast.com/information/security#runtime-model).

Because I'm no security expert, I completely understand, and I'm gonna remove the password history code from the PR and change it to be a fix on the generator.

pomdtr:
Thanks for your comprehension @jomifepe ❤️ 

I'm confident that your implementation is right, but I mainly want to avoid overcomplexifing the extension code to be able maintain it in my free time. As soon as this feature is implemented in the bitwarden cli, we will revisit this.

jomifepe:
I chose to open a new pull request, [here it is](https://github.com/raycast/extensions/pull/1372). 

Thank you for your time in this one 🙂

> Thanks for your comprehension @jomifepe ❤️
> 
> I'm confident that your implementation is right, but I mainly want to avoid overcomplexifing the extension code to be able maintain it in my free time. As soon as this feature is implemented in the bitwarden cli, we will revisit this.

PR #1389, opened 2022-04-13
fix(Lark): fix sometime couldn't open documents

htoooth:
## Checklist

- [] I checked that files in the `assets` folder are used by the extension itself
- [] I checked that assets used by the `README` are placed outside of the `metadata` folder

### DIFF
--- a/extensions/lark/src/services/space.ts
+++ b/extensions/lark/src/services/space.ts
@@ -223,6 +223,7 @@ export async function searchDocs(params: SearchDocsParams): Promise<SearchDocsRe
         title: trimTagsAndDecodeEntities(objEntity.title),
         preview: trimTagsAndDecodeEntities(objEntity.preview),
         url: computeRedirectedUrl(objEntity),
+        type: computeType(objEntity),
       };
     });
     return data;
@@ -262,6 +263,14 @@ const computeRedirectedUrl = (objEntity: ObjEntity) => {
     .replace(/\/space\/bitable\//, '/base/')
     .replace(/\/space\/mindnote\//, '/mindnotes/')
     .replace(/\/space\//, '/');
+};
+
+const computeType = (objEntity: ObjEntity) => {
+  if (objEntity.wiki_infos) {
+    return NodeType.Wik;
+  }
+
+  return objEntity.type;
 };
 
 const CACHE_DOCS_RECENT_LIST = 'CACHE_DOCS_RECENT_LIST';


extensions/lark/src/services/space.ts
```diff
@@ -240,12 +240,19 @@ export async function searchDocs(params: SearchDocsParams): Promise<SearchDocsRe
...
+      .replace(/\/space\/sheet\//, '/sheets/')
+      .replace(/\/space\/bitable\//, '/base/')
+      .replace(/\/space\/mindnote\//, '/mindnotes/')
+      .replace(/\/space\//, '/');
+  }
+
+  if (objEntity.wiki_infos) {
+    objEntity.type = NodeType.Wik
+    return objEntity.wiki_infos[0].wiki_url
+  }
```
tmkx, line 255: Not every condition returns a value.

htoooth: ok，i fix it. @tmkx

tmkx:
LGTM

extensions/lark/src/components/space-list-item.tsx
```diff
@@ -32,7 +32,7 @@ export const SpaceListItem: React.FC<SpaceListItemProps> = ({ node, owner, actio
       node.type === NodeType.Box
         ? `space-icons/type-${node.type}-${node.subtype}.png`
         : `space-icons/type-${node.type}.png`;
-    accessoryTitle = node.author;
+    accessoryTitle = `${node.author} ${dayjs(node.edit_time * 1000).format('YYYY-MM-DD')}`;
```
tmkx, line 35: Why do this? Could you please provide a screenshot?

htoooth: @tmkx  This shows the latest editing time. You can judge whether this document is the latest or old, so you don't need to open the old one.

image

tmkx: If the author's name is very long (maybe an English name or a minority's), the `accessoryTitle` will be too wide:
image

We can sort the search result by updated time to solve your doubt. I may make it more convenient by adding an overview panel later.

htoooth: @tmkx  Yes, you are right. Maybe It would be better to have a text box here. 

image

tmkx: I think so, could you revert this first? I'll redesign it in few days.

htoooth:
@tmkx  I have reverted this.

tmkx:
LGTM

PR #1391, opened 2022-04-13
Add Quick Access extension

koinzhang:
## Description

- Keep your files and folders on the shelf, ready to be copied, opened or revealed in Finder.
- If the original file is deleted, the file on shelf will be deleted and not be available too.

## Screencast

https://user-images.githubusercontent.com/36128970/163184828-cbf32a94-98ab-48e1-bfbe-6272634ec4d2.mp4

## Checklist

### DIFF
--- a/extensions/quick-access/src/hooks/hooks.ts
+++ b/extensions/quick-access/src/hooks/hooks.ts
@@ -92,8 +92,6 @@ export const copyLatestFile = (autoCopyLatestFile: boolean, pinnedDirectoryConte
       }
       setIsCopy(true);
     }
-    //init applescript
-    await runAppleScript("");
   }, [pinnedDirectoryContent]);
 
   useEffect(() => {


extensions/quick-access/package.json
```diff
@@ -0,0 +1,125 @@
...
+      "name": "pin-directory",
+      "title": "Pin Directory",
+      "icon": "pin-directory.png",
+      "subtitle": "Quick Access",
+      "description": "Pin directories that can be accessed quickly.",
+      "mode": "no-view"
+    },
+    {
+      "name": "access-directory",
+      "title": "Access Directory",
```
thomaspaulmann, line 20: I think this describes the command better.

```suggestion
      "title": "Searched Pinned Directory",
```

extensions/quick-access/package.json
```diff
@@ -0,0 +1,125 @@
...
+          "default": true
+        },
+        {
+          "name": "showDeleteAction",
+          "type": "checkbox",
+          "required": true,
+          "label": "Show Delete File Action",
+          "description": "Show permanent file deletion action, please check carefully.",
+          "default": true
+        },
```
thomaspaulmann, line 42: Let's remove this preference. There shouldn't be any harm having this.

```suggestion
```

extensions/quick-access/package.json
```diff
@@ -0,0 +1,125 @@
...
+        {
+          "name": "fileShowNumber",
+          "type": "dropdown",
+          "required": true,
+          "title": "Number of files shown",
+          "description": "Number of files displayed per directory.",
+          "default": "3",
```
thomaspaulmann, line 49: Is this preference really necessary? If you keep it, I would suggest to use "All" as default option and remove the `required` option.

extensions/quick-access/package.json
```diff
@@ -0,0 +1,125 @@
...
+            {
+              "title": "All",
+              "value": "All"
+            }
+          ]
+        },
+        {
+          "name": "sortBy",
+          "type": "dropdown",
+          "required": true,
```
thomaspaulmann, line 80: Let's not make this required and pick the best default.
```suggestion
```

extensions/quick-access/package.json
```diff
@@ -0,0 +1,125 @@
...
+      "title": "Access Directory",
+      "icon": "access-directory.png",
+      "subtitle": "Quick Access",
+      "description": "Quick copy, open, reveal the latest files in pinned Finder directories.",
+      "mode": "view",
+      "preferences": [
+        {
+          "name": "autoCopyLatestFile",
+          "type": "checkbox",
+          "required": true,
```
thomaspaulmann, line 29: Let's not make this required. Also, it's unclear what this option does.
```suggestion
```

thomaspaulmann (changes requested):
Thanks for the contribution. I left a few inline comments to make the extension easier to understand. I had a hard time to understand what it does. I think changing some naming makes it more approachable.

Few more things:
- [ ] Could you add some screenshots? You can find more info [here](https://developers.raycast.com/basics/prepare-an-extension-for-store#screenshots). 
- [ ] Can you swap your light/dark icons? Right now you have dark background in the dark theme. It should be light background on the dark theme to have higher contrast.

Thanks a lot!

extensions/quick-access/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+
+**Command**:
+
+- Pin Directory: Pin directories that can be accessed quickly.
+- Access Directory: Quick copy, open, reveal the latest files in pinned Finder directories.
+  - Copy to Clipboard
+  - Open in Default App/Open in Finder
+  - Reveal in Finder
+  - Pin Directory
```
thomaspaulmann, line 19: This action didn't work for me :-(

extensions/quick-access/package.json
```diff
@@ -0,0 +1,125 @@
...
+  "name": "quick-access",
+  "title": "Quick Access",
+  "description": "Quick copy, open, reveal the latest files in pinned Finder directories.",
+  "icon": "extension-icon.png",
+  "author": "koinzhang",
+  "license": "MIT",
+  "commands": [
+    {
+      "name": "pin-directory",
+      "title": "Pin Directory",
```
thomaspaulmann, line 12: Is `Pin` the right name?

extensions/quick-access/package.json
koinzhang: Here I think it is necessary.
First let me explain the inspiration for this extension. One of the partners in the community wanted to create an extension: Copy Download Latest File, which is to copy the latest file in the download directory.

I was inspired by the idea that when we download, take a screenshot, or save a file to some common folder, like the download folder or the screenshot folder, and we want to copy or open the file immediately, we can quickly get the latest file with the help of this extension (so the files in the pinned directory are sorted by modification time).
If I only want to get the latest few files, this preference is necessary, especially in a small view window like raycast, if  show all the files, then the files in a directory may fill up the whole window, and it becomes very troublesome to find the files.

Many download files or screenshots from a long time ago are actually invalid, so it is very necessary to be able to display only the latest few files.

The functionality of this extension is similar to Raycast's newly updated Search Screnshot, except that this extension will display all types of files.

Quick Access
Search Screenshot

extensions/quick-access/package.json
koinzhang: > Could you add some screenshots? You can find more info [here](https://developers.raycast.com/basics/prepare-an-extension-for-store#screenshots).
 Can you swap your light/dark icons? Right now you have dark background in the dark theme. It should be light background on the dark theme to have higher contrast.
 
I forgot the metadata, I will add it.
As for the icon issue, I will try to modify it, and if it doesn't perform well after swapping icon themes, I might just keep the light-colored icons

> I think this describes the command better.

Good idea, I will modify it。

extensions/quick-access/package.json
koinzhang: The name of this extension was inspired by the Quick Access feature on windows.
So before I set the name to Quick Access and Pin Directory.
Do you have a good suggestion?

02iOkvupvcqZ5Ybx270ZmGP-28 fit_lim size_1536x

extensions/quick-access/README.md
koinzhang: Which action?
If it is "Copy File to Clipboard ", the implementation of this action uses AppleScript and accesses the file, you need to give Raycast the appropriate permissions in the system settings

extensions/quick-access/package.json
thomaspaulmann: Wouldn't sorting the files by recently modified/added solve this problem? Than the latest file is always at the top. It's just confusing that I don't see all my downloads by default. I didn't expect that and thought the extension is broken.

extensions/quick-access/README.md
thomaspaulmann: The `Pin Directory`. It shows me this toast. To be honest, I also don't know what should happen. I expected that the selcted directory would be pinned and then accessible in the dropdown but it wasn't.
CleanShot 2022-04-22 at 12 14 48

extensions/quick-access/package.json
thomaspaulmann: Ah I wasn't aware of this feature. Then let's go with this name and trust Microsoft's naming ;-)

extensions/quick-access/package.json
koinzhang: Indeed it may confuse the user.
So here I will change the default to all, but the preferences will remain so that there can be more options.

thomaspaulmann: Sounds good. I wouldn't make it required. Then the user can start using the extension and can optimize in the preferences later.

Generally, required preferences make mostly sense for things that need to be set up to make the extension work. This pref doesn't seem to be the case. So it's better to keep it non-required.

extensions/quick-access/README.md
koinzhang: If you want to use the Pin Directory action, you need to select a folder or, alternatively, open a Finder window first.
If you do not select any folder or open a Finder window, but use the Pin Directory action, it will detect and indicate that No directories are pinned.

koinzhang: Another situation is that if you select a folder that has already been pinned, when you pin it again, since the duplicate folder will not be added, it will also prompt that No directories are pinned.

extensions/quick-access/package.json
koinzhang: In fact, I prefer to set the require to True for some preferences.
Because most of the time, users don't know that your extension has a preference option and rarely go to the preferences page to check it.

koinzhang:
Hi, @thomaspaulmann. Thanks for your review. Sorry to make you wait so long. I have submitted a commit to update extension.
Specific changes :

- Add metadata
- delete dark icon
- Update "Access Directoey" to "Search Pinned Directory"
- Update Preference options
- Update a bug that Toast will frequently show when you add or remove directory
- Update List Empty View

Please continue to review.

extensions/quick-access/package.json
```diff
@@ -0,0 +1,117 @@
...
+      "icon": "pin-directory.png",
+      "subtitle": "Quick Access",
+      "description": "Pin directories that can be accessed quickly. To use this command, you need to select folders or open a Finder window first.",
+      "mode": "no-view"
+    },
+    {
+      "name": "search-pinned-directory",
+      "title": "Search Pinned Directory",
```
thomaspaulmann, line 20:
```suggestion
      "title": "Search Pinned Directories",
```

extensions/quick-access/package.json
```diff
@@ -0,0 +1,117 @@
...
+      "icon": "access-directory.png",
+      "subtitle": "Quick Access",
+      "description": "Quick copy, open, reveal the latest files in pinned Finder directories.",
+      "mode": "view",
+      "preferences": [
+        {
+          "name": "autoCopyLatestFile",
+          "type": "checkbox",
+          "required": true,
```
thomaspaulmann, line 29: Let's not mark this as required and pick the default that you think is best. I think not copying it automatically. Mostly because I see this screen when I open the extension for the first time and I don't know what to do here.
```suggestion
          "required": false,
```
CleanShot 2022-04-22 at 4 04 15

extensions/quick-access/src/search-pinned-directory.tsx
```diff
@@ -0,0 +1,240 @@
...
+                  key={directory.directory.id}
+                  title={directory.directory.name}
+                  subtitle={parse(directory.directory.path).dir}
+                >
+                  {directory.files.map((value, index) => (
+                    <List.Item
+                      key={value.id}
+                      icon={isImage(parse(value.path).ext) ? { source: value.path } : { fileIcon: value.path }}
+                      title={value.name}
+                      accessories={index === 0 ? [{ icon: "🌟" }] : []}
```
thomaspaulmann, line 68: This icon feels a lil out of place. If you decide to keep it, can you also add a `tooltip` to describe it?

thomaspaulmann (changes requested):
Left a few minor bits. Somehow the command feels slow for what it does. I couldn't pinpoint it  but might be worth to look into it in a second iteration. Anything you think could cause it?

extensions/quick-access/package.json
koinzhang: Yes, the file name and command name have been changed to "Search Pinned Directories".

extensions/quick-access/package.json
koinzhang: Okay, I'll set it to false by default and stop popping up.

extensions/quick-access/src/search-pinned-directory.tsx
koinzhang: Maybe I can find a better way to describe it, and if I do, I'll change it; if not, I'll change 🌟 to "Latest File".

koinzhang:
Hi @thomaspaulmann . Thanks for your review.
Latest Changes:
- Update command name
- Updated preferences
- Removed "🌟"

As for the startup speed of the extension, I haven't found the reason yet, guessing that it might be the time consuming to read the folder content?

koinzhang:
Hi @thomaspaulmann 
I have found the cause of the slow startup and solved it, please keep reviewing.

thomaspaulmann (approved):
Great that you find the problem and fixed it. It's ready to be published. Thanks for all the effort to make this extension seamless.

PitNikola:
@koinzhang, thanks for building this extension. 
Quick feedback: let's find a better shortcut for deleting pinned directory than `⌘ + Delete`. This shortcut is reserved for deleting text to the beginning in the search field. I was quite confused when I wanted to delete the text and then suddenly everything was gone.

koinzhang:
> @koinzhang, thanks for building this extension.
> Quick feedback: let's find a better shortcut for deleting pinned directory than `⌘ + Delete`. This shortcut is reserved for deleting text to the beginning in the search field. I was quite confused when I wanted to delete the text and then suddenly everything was gone.

I didn't notice before that cmd+Backspace removes the text from the search bar, so I'm very sorry for causing this problem.
I will submit a PR to update the extension.

PR #1392, opened 2022-04-13
Update Lorem Ipsum Extension: Allow the action of the command to be configured

harryaswan:
## Description

Creates a configuration option so that users can select what the action of the command does, either copy to clipboard (as it does just now) or paste to the current active app (new functionality).

## Screencast

Screenshot 2022-04-13 at 15 25 03

## Checklist

### DIFF
--- a/extensions/lorem-ipsum/src/paragraph.tsx
+++ b/extensions/lorem-ipsum/src/paragraph.tsx
@@ -2,7 +2,7 @@ import { getPreferenceValues } from "@raycast/api";
 import { generateParagraph, preformAction } from "./utils";
 
 export default async function ParagraphCommand() {
-  const { action="clipboard" } = getPreferenceValues();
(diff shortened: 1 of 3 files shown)


AntonNiklasson:
This is great, thanks for the change. I'll have a look in a day or two.

harryaswan:
> Perhaps it makes more sense to have the paste be on cmd+enter, similar to Raycast's snippets? It seems more natural to me, what do you think?

I did think about that however, unless I am missing something, I don't think you can add additional options to the main search view, only within the extension where you can add an ActionPanel - my initial idea was to do something similar to the way the snippets work where you can set the default action (screenshot) with the alternative becoming the `cmd`+`enter` option.

Screenshot 2022-04-14 at 10 50 06

pernielsentikaer (approved):

PR #1393, opened 2022-04-13
feat: support custom journal location & support org mode!

futantan:
## Description

- Users can config the custom journal location in logseq using `:journals-directory "your-directory"`, this PR covers this config
- Support org mode

## Checklist

### DIFF
--- a/extensions/logseq/src/index.tsx
+++ b/extensions/logseq/src/index.tsx
@@ -16,7 +16,7 @@ export default function Command() {
 
   async function handleSubmit(values: CommandForm) {
     if (!values.content) {
-      return showToast({ style: Toast.Style.Failure, title: "🐝 Type something to get started" })
+      return showToast({ style: Toast.Style.Failure, title: "🐝 Type something to get started" });
     }
     validateUserConfigGraphPath()
       .catch((e) => {


sxn:
nice! 🥳 

As the extension is starting to support more and more of `logseq` config options, do you think it would make sense to add a `README.md` (and a `CHANGELOG.md` 😸) to make sure these capabilities are clear to new users? 

On the actual functionality implemented here, I've run into an issue: if you specify a custom journal folder using `journals-directory` but the folder doesn't exist, the extension kind of "hangs"; it's still responsive, but it never seems to finish. Could you have a look at catching that and either reporting it to the user (maybe offer a [toast action](https://developers.raycast.com/api-reference/feedback/toast#toast) to have the user confirm that the directory should be created) or just catching the case and creating the directory?

futantan:
@sxn Thanks for your feedback!
Issue fixed.
Changelog added.
The readme looks good to me since what we want is to let our users know nothing about how it works, but it's just working!

sxn:
Thanks for the changes! 😄 I've pushed one small change to the changelog: I've reversed the order so that the latest update is at the top of the file, according to [keep a changelog](https://keepachangelog.com/en/1.0.0/), hope you don't mind. 😺 

One more thing I noticed just now - it's possible to submit the form even without typing anything, resulting in an empty bullet point being added to the journal file. Is that intended? 🤔

futantan:
@sxn Thanks for your feedback!
Updated.

sxn:
🎉

PR #1405, opened 2022-04-14
[Extension] - Remember the Date

pernielsentikaer:
## Description

Adds "Remember the Date" - great for remembering how many years of marriage or upcoming events

## Screencast

A195Vf9bZv9Cqj8U

## Checklist

### DIFF
--- a/extensions/remember-the-date/package.json
+++ b/extensions/remember-the-date/package.json
@@ -16,7 +16,7 @@
     },
     {
       "name": "list",
-      "title": "List Days",
+      "title": "Days List",
       "subtitle": "Remember the Date",
       "description": "List of all dates to remember",
       "mode": "view"


mattisssa (changes requested):
Wow, congrats on your another extension in the Store! 🚀 

I haven't looked at code yet, but here is some bits I noticed during testing your extension:

Could be nice to have here an action with the "Create Date" form 😉 

CleanShot 2022-04-14 at 18 23 39@2x

I think until we haven't introduced the validation, it could be nice to check if the Date field is filled and prevent submitting if it's empty. Since without date it won't be visible in the List.
CleanShot 2022-04-14 at 18 26 14@2x

Seems like this is something off with the icon here in case I don't provide an Icon in the Create Date command. The title should be aligned on the left side. And in general, I don't think this is a good UI having inconsistent list Items – somewhere there are icons somewhere there are don't, the title-subtitle looks a bit jumpy because of this.
Probably some placeholder Icon would be nice? 

CleanShot 2022-04-14 at 18 27 41@2x

pernielsentikaer:
Hi @mattisssa 👋 

Thanks for giving it a quick spin, could you test it again with you have time 😄

mattisssa:
> Hi @mattisssa 👋
> 
> Thanks for giving it a quick spin, could you test it again with you have time 😄

Hey! Checked again and now it works 👍 
But how about adding some placeholder for the list item on my last screenshot? 

And question about the general idea of extension, what is the point having this extension if I can use Reminders, that's also synced with my mobile phone in example? Or same for the Calendar app.

pernielsentikaer:
It's mostly for the past dates for me, I need to remember alot of important days.

I don't use notes and that would not give me a quick overview of how long time ago it was.

There is no need for placeholder icon since I removed the option for no icon :)

extensions/remember-the-date/package.json
```diff
@@ -0,0 +1,59 @@
...
+    {
+      "name": "index",
+      "title": "Create Date",
+      "subtitle": "Remember the Date",
+      "description": "Create a new until element",
+      "mode": "view"
+    },
+    {
+      "name": "list",
+      "title": "List Days",
```
mattisssa, line 19:
```suggestion
      "title": "Days List",
```

extensions/remember-the-date/package.json
```diff
@@ -0,0 +1,59 @@
...
+      "subtitle": "Remember the Date",
+      "description": "Create a new until element",
```
mattisssa, line 21:
```suggestion
      "description": "List of all dates to remember",
```

extensions/remember-the-date/package.json
```diff
@@ -0,0 +1,59 @@
...
+  "description": "Creates and shows the most important date which should be remembered.",
+  "icon": "countdown.png",
+  "author": "pernielsentikaer",
+  "license": "MIT",
+  "commands": [
+    {
+      "name": "index",
+      "title": "Create Date",
+      "subtitle": "Remember the Date",
+      "description": "Create a new until element",
```
mattisssa, line 14:
```suggestion
      "description": "Create new date to remember",
```

mattisssa (approved):

PR #1410, opened 2022-04-15
added aws profile preference

dima-m711:
## Description

added aws profile preference

## Screencast

image

## Checklist

### DIFF
--- a/extensions/amazon-aws/src/cloudformation.tsx
+++ b/extensions/amazon-aws/src/cloudformation.tsx
@@ -5,9 +5,10 @@ import { Preferences } from "./types";
 import { StackSummary } from "aws-sdk/clients/cloudformation";
 import setupAws from "./util/setupAws";
 
+setupAws();
+
 export default function ListStacks() {
   const preferences: Preferences = getPreferenceValues();
-  setupAws();
   const cloudformation = new AWS.CloudFormation({ apiVersion: "2016-11-15" });
 
   const [state, setState] = useState<{
--- a/extensions/amazon-aws/src/codepipelines.tsx
+++ b/extensions/amazon-aws/src/codepipelines.tsx
@@ -4,6 +4,8 @@ import * as AWS from "aws-sdk";
 import { Preferences } from "./types";
 import setupAws from "./util/setupAws";
 
+setupAws();
+
 const getExecutionState = (client: AWS.CodePipeline, pipelineName: string) =>
   new Promise<AWS.CodePipeline.PipelineExecutionSummary | null>((resolve, reject) => {
     return client.listPipelineExecutions(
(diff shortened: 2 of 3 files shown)


extensions/amazon-aws/src/util/setupAws.ts
```diff
@@ -0,0 +1,10 @@
+import AWS from "aws-sdk";
+import { getPreferenceValues } from "@raycast/api";
+import { Preferences } from "../types";
+
+export default function setupAws() {
+  const preferences: Preferences = getPreferenceValues();
+  const credentials = new AWS.SharedIniFileCredentials({profile: preferences.aws_profile});
+  AWS.config.credentials = credentials;
+  AWS.config.update({  region: preferences.region });
+}
```
gebeto, line 10: Let's use hook here, to not configuring AWS on each render, also getPreferenceValues is a generic function, so it will be better to put Preferences type as generic parameter

```suggestion
export default function setupAws() {
  const preferences = getPreferenceValues<Preferences>();
  
  React.useEffect(() => {
    const credentials = new AWS.SharedIniFileCredentials({profile: preferences.aws_profile});
    AWS.config.credentials = credentials;
    AWS.config.update({  region: preferences.region });
  }, [preferences];
}
```

gebeto (approved):

gebeto:
Does anybody know, is `getPreferenceValues` a hook or not? where it will be better to call this function?

When it is updating, and does the app reloading when preferences is changing?

possibly it will be better to run it outside of the React lifecycle(outside of the components)? or directly inside the useEffect with [] dependencies(if root component remounting on preferences update)

extensions/amazon-aws/src/util/setupAws.ts
dima-m711: 1. added preferences.
2. converted this to hook complicate things, because the useEffect is running async, and we need to create the Instance of SQS or EC2 with updated config.
so I think the simple approach we can have a global init at the beginning of the command, right now it's just setupAws, WDYT?

gebeto: Yeah make sense, but I don't really know is it okay to run it outside of the react component, but I hope it is Okay

Just test it and if it works fine - great, let's do it😄

For now it looks great for me

gebeto (approved):

mattisssa:
const { token }: Preferences = getPreferenceValues();

> Does anybody know, is `getPreferenceValues` a hook or not? where it will be better to call this function?
> 
> When it is updating, and does the app reloading when preferences is changing?
> 
> possibly it will be better to run it outside of the React lifecycle(outside of the components)? or directly inside the useEffect with [] dependencies(if root component remounting on preferences update)

It's better to call this outside of the components, we have many examples of its usage [in our repo](https://github.com/raycast/extensions/blob/328e27757a5661ec034166d6ce49692f5847665f/extensions/trello/src/utils/fetchTodos.tsx#L11) 😉  
`const { token }: Preferences = getPreferenceValues();`

gebeto:
Thanks, @mattisssa 🙌

PR #1426, opened 2022-04-17
Improve swift playground extension

qeude:
## Description

Made few changes to the "Create Swift Playground" command:
- Add an UIKit template to the templates list..
- Add the capability to set a custom default location.
- Divide the existing action into two actions, one to force create, and the other one to open or create.

## Checklist


pernielsentikaer:
Hi @qeude 👋 

Thanks for taking your time to contribute to this extension 🔥 

It looks like you forgot to format with Prettier, could you do that?

@SvenTiigi do you have time to take a fast peek?

qeude:
@pernielsentikaer Sorry for that, forgot to run it 🤦🏼 It should be better now 🙌🏼

SvenTiigi:
@qeude 

> Add an UIKit template to the templates list..

The addition of a UIKit template is a great addition 👍 

> Add the capability to set a custom default location.

The implementation looks good to me 👍 
Maybe the [`placeholder`](
https://github.com/raycast/extensions/pull/1426/files#diff-7e2c1ed5cf3c8d4d56c87b428384732930a1ac83f6d8de6e2f13515ca6c2905fR84) value of the preference could be adjusted to `~/Desktop` instead of `/tmp`

> Divide the existing action into two actions, one to force create, and the other one to open or create.

Adding dedicated actions to `create` and `create and open` is a nice idea. 

I think it would be even better to match the implementation to the [Playground CLI](https://github.com/JohnSundell/Playground) by JohnSundell where the Playground is created with the following file name format `Playground-{DATE}.playground` where the date consist of the day, month and the year.

This way only one playground will be created for the same day as most of the iOS developers are re-using the same Playground throughout the day as well as to easily archive them. Additionally, a secondary action could be added to force create a new Playground which could be named with an incremental counter after the date e.g. `Playground-{DATE}-2.playground`.

extensions/xcode/package.json
```diff
@@ -72,7 +75,18 @@
...
       "subtitle": "Xcode",
       "description": "Create a new Swift Playground and open it directly in Xcode",
-      "mode": "view"
+      "mode": "view",
+      "preferences": [
+        {
+          "name": "playgroundDefaultLocation",
+          "type": "textfield",
+          "required": true,
+          "placeholder": "/tmp",
```
SvenTiigi, line 84: Maybe `~/Desktop` would be more suitable instead of `/tmp`

extensions/xcode/package.json
```diff
@@ -72,7 +75,18 @@
       "title": "Create Swift Playground",
       "subtitle": "Xcode",
       "description": "Create a new Swift Playground and open it directly in Xcode",
-      "mode": "view"
+      "mode": "view",
+      "preferences": [
+        {
+          "name": "playgroundDefaultLocation",
+          "type": "textfield",
+          "required": true,
```
SvenTiigi, line 83: I guess `false` would be more suitable as the default value `~Desktop` should be okay for most of the users.

qeude: In that case I choose to put `~/Desktop` as `default` on l85 which seems to fit more with the idea of turning this into a Raycast preference.

qeude:
@SvenTiigi I've made few changes related to your feedbacks, hopefully it will be good 🚀  I also invited the two form actions since as you said its maybe more relevant for an iOS dev to have one playground per day by default.

pernielsentikaer:
Can you do another check @SvenTiigi, if you have the time?

extensions/xcode/package.json
SvenTiigi: I was thinking about disabling the `required` property to omit the Raycast preference configuration screen when opening the `Create Swift Playground` command for the first time

Screenshot

As `~/Desktop` is a good default value for most of the users and therefore not truly required. 
Those users who want to specify a different a path can easily adjust the preference via the Raycast Extensions Tab.

extensions/xcode/package.json
```diff
@@ -72,7 +75,18 @@
...
+          "placeholder": "~/Desktop",
+          "default": "~/Desktop",
+          "title": "Default location",
+          "description": "Set the default location in which you want to save your Playgrounds. Ensure to provide an existing path or you will get an error on create."
```
SvenTiigi, line 87: I guess the last sentence can be omitted `Ensure to provide an existing path or you will get an error on create.` as errors are gracefully handled by the `Create Swift Playground` command via a failure `Toast`.

extensions/xcode/src/services/xcode-swift-playground.service.ts
```diff
@@ -45,16 +46,18 @@ export class XcodeSwiftPlaygroundService {
...
-  async createSwiftPlayground(parameters: XcodeSwiftPlaygroundCreationParameters): Promise<XcodeSwiftPlayground> {
+  async createSwiftPlayground(
+    parameters: XcodeSwiftPlaygroundCreationParameters,
+    forceCreate: boolean
+  ): Promise<XcodeSwiftPlayground> {
     // Initialize Playground Path
-    const playgroundPath = joinPathComponents(
-      // Replace tilde (~) with home directory
+    const playgroundPath = await getValidPath(
       parameters.location.replace(/^~/, os.homedir()),
```
SvenTiigi, line 55: Prefer the usage of `untildify`
```typescript
untildify(parameters.location)
```

extensions/xcode/src/shared/check-path-or-increment.ts
```diff
@@ -0,0 +1,23 @@
...
+import * as os from "os";
+import { existsAsync } from "../shared/fs-async";
+
+/**
+ * return a valid file path
+ * @param location: location in which we want to save the playground file.
+ * @param filename: filename of the created or opened playground file.
+ * @param forceCreate: define if we want to force create or not.
+ */
+export async function getValidPath(location: string, filename: string, forceCreate: boolean): Promise<string> {
```
SvenTiigi, line 11: I am not really sure if this function would be better placed in the `XcodeSwiftPlaygroundService` class 🤔

The implementation of the `getValidPath` function is tightly coupled to creating a Swift Playground based on the `.playground` file extension and therefore not really meant to be shared across the codebase of the Xcode Raycast Extension itself.

extensions/xcode/src/shared/check-path-or-increment.ts
```diff
@@ -0,0 +1,23 @@
...
+  let path = "";
+  let iteration = null;
+  const targetLocation = location.replace(/^~/, os.homedir());
+  do {
+    const dateString = new Date().toISOString().slice(0, 10).replace(/-/g, "");
```
SvenTiigi, line 16: Currently the `dateString` is represented in the following format `yyyyMMdd` I guess adding a hyphen between the components would improve the readability like `yyyy-MM-dd`

Example: `MyPlayground-2022-04-22.playground`

Additionally, I guess it would be even better to use the [`toLocaleDateString`](https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleDateString) API of the Date object.

extensions/xcode/package.json
qeude: Thanks for the explanation on that point, I agree with you 😄

qeude:
@SvenTiigi Thanks for these feedbacks 🙌🏼  Should be good now

SvenTiigi:
@qeude looks good to me 🙂
Great work 👍

pernielsentikaer (approved):
Hi @qeude 👋 

Thanks for the contribution 💯

PR #1456, opened 2022-04-21
[Bitwarden] Add support for self hosted instances

### LINKED ISSUE #1447: [Bitwarden] Support self-hosted Bitwarden Vault
# Extension – [Bitwarden](https://github.com/raycast/extensions/tree/main/extensions/bitwarden)

Author: @pomdtr 

I am using a self-hosted Bitwarden (Vaultwarden actually) but seemed like the extension does not support connecting to the self-hosted vaults.
I successfully logged from the bitwarden-cli but cannot find a way to change the other settings in the extension.

**Who will benefit from this feature?**

A lot of users using self-hosted vaultwarden would benefit from this, there would be no need to install the App anymore!

**Any Other info.**

No.

daniel-stoneuk:
## Description

Fixes #1447.

* Added server url and certs preferences
* Set Bitwarden CLI server when these preferences change

**TODO** 

- [x] needs to be tested on a self hosted instance
- [x] check that upgrade works fine

## Checklist

### DIFF
--- a/extensions/bitwarden/src/search.tsx
+++ b/extensions/bitwarden/src/search.tsx
@@ -307,7 +307,7 @@ function VaultActions(props: { syncItems: () => void; lockVault: () => void; log
         shortcut={{ modifiers: ["cmd", "shift"], key: "l" }}
         onAction={props.lockVault}
       />
-      <Action title="Logout Vault" icon={Icon.XmarkCircle} onAction={props.logoutVault} />
+      <Action title="Logout" icon={Icon.XmarkCircle} onAction={props.logoutVault} />
     </Fragment>
   );
 }


daniel-stoneuk:
@aozaki-kuro please could you give this a go on your self hosted instance!

daniel-stoneuk:
@xorob0 I noticed you +1ed the issue, do you have a self-hosted server you could test on as well?

I'm about to add a "Switching server" toast to make it clear when the CLI is switching between servers.

extensions/bitwarden/src/api.ts
```diff
@@ -69,7 +111,12 @@ export class Bitwarden {
...
   }
 
-  private async exec(args: string[], abortController?: AbortController): Promise<ExecaChildProcess> {
+  private async exec(
+    args: string[],
+    abortController?: AbortController,
+    waitForInit = true
+  ): Promise<ExecaChildProcess> {
+    if (waitForInit) await this.initPromise;
     return execa(this.cliPath, args, { env: this.env, input: "", signal: abortController?.signal });
```
jomifepe, line 120: Positional parameters quickly get inconvenient, so it's more clear accept an `options` object here. 
This way we also avoid passing `undefined` on the `abortController` 🙂

```suggestion
    options?: { abortController?: AbortController, waitForInit?: boolean }
  ): Promise<ExecaChildProcess> {
    const { abortController, waitForInit = true } = options ?? {};
    if (waitForInit) await this.initPromise;
    return execa(this.cliPath, args, { env: this.env, input: "", signal: abortController?.signal });
```

jomifepe (commented):
Code looks great 🚀
Can you provide a demonstration of this feature? 🙂

extensions/bitwarden/src/api.ts
daniel-stoneuk: This is much better. Thanks!

Coming from Python most of the time I tried to call the functions with `exec(["logout", waitForInit = true)` 🤦

daniel-stoneuk:
> Code looks great overall 🚀

Thanks! I'm not super happy with the use of `Toast` within the API class, but to pull it out to the view logic would add much more needless complexity.

pomdtr:
Thanks @daniel-stoneuk for adding the feature ! I will do a proper tomorrow.

daniel-stoneuk:
No worries @pomdtr, hope you like it 😁 

Quick demo video (the extension needs to be reloaded for preferences to take effect anyway)

https://user-images.githubusercontent.com/15860949/164558875-8ba17c86-77bd-460d-9af8-3257a10900b4.mp4

pomdtr:
Looks good! However I'm not sure we want to surface the server information to regular users.

And advanced users hosting their own server should have the reflex to both read the README and look at the prefs.

daniel-stoneuk:
Good point. I tried surfacing it only when it wasn't the default but having it appear and push the rest down after the status load was a bit jarring.

Just had another think though, and I'll show the server field if:

a) the preference is set
b) the preference is unset but for some reason status returns the default

Thanks for the feedback!

aozaki-kuro:
Sorry for the late response, I was at work and spent a long time to figure out how to import an extension...

image

... And it seemed to be working totally fine on my side🙏

pernielsentikaer:
Let's await feedback from @pomdtr

extensions/bitwarden/package.json
```diff
@@ -73,6 +73,15 @@
       "description": "The client_secret value seen in the My Account section of a Bitwarden web vault.",
       "placeholder": "client_secret"
     },
+    {
+      "name": "fetchFavicons",
+      "type": "checkbox",
+      "required": true,
```
pomdtr, line 79: Good idea !

extensions/bitwarden/src/components.tsx
```diff
@@ -25,32 +39,68 @@ export function TroubleshootingGuide(): JSX.Element {
...
-      } else {
-        setVaultStatus(`Locked (${vaultState.userEmail})`);
-      }
+      setVaultState(vaultState);
     });
   }, []);
 
+  let userMessage = "...";
+  let serverMessage = "...";
+  if (vaultState) {
```
pomdtr, line 55: Can we move this block to a separate function ?

extensions/bitwarden/src/api.ts
```diff
@@ -1,27 +1,71 @@
...
 export class Bitwarden {
   private env: Record<string, string>;
   cliPath: string;
-  constructor(clientId: string, clientSecret: string, cliPath: string) {
-    if (!cliPath) {
-      cliPath = process.arch == "arm64" ? "/opt/homebrew/bin/bw" : "/usr/local/bin/bw";
-    }
-    if (!existsSync(cliPath)) {
-      throw new Error(`Bitwarden CLI not found at ${cliPath}`);
+  initPromise: Promise<void>;
```
pomdtr, line 12: Is this initPromise logic a frequent pattern in javascript ? It took me a bit of time to figure out what was happening.

I would prefer to move the logic to the exec function. I'll propose another solution today.

pomdtr (approved):

extensions/bitwarden/src/components.tsx
daniel-stoneuk: Good idea, I'll move this into another component.

extensions/bitwarden/package.json
daniel-stoneuk: Thanks!

extensions/bitwarden/src/api.ts
daniel-stoneuk: I'm not sure if it is a pattern tbh, thanks for taking the time to review the code.

This was the only way that I could think of that would delay any calls before the server is set. 

For example, the user is logged out and the server is changed. The user enters their password before the status call is complete. Other methods I thought of either a) did the config call twice or b) did it once and let the second go through before it was finished. Some kind of checking on an interval would also work the same as this. 

There's probably a better way, so I look forward to seeing wht you think 👍

pomdtr: Thanks for the explanation, I did not thought of the concurrency issue.
Can we move all of logic related to the server to the setServerUrl function ?

Ideally I would like to only have this line in the constructor: 

```
this.initPromise = this.setupServer()
```

with all of the check at the top of the setupServer method, and some early return if nothing has to be done

pomdtr (changes requested):
(forgot to set the pull request status to `Request changes`)

xorob0:
Can confirm it works with my self hosted server

extensions/bitwarden/src/api.ts
daniel-stoneuk: Resolved in e8d428247f168a96a346cf52e5cea704b299ce5e

pomdtr (approved):

pomdtr:
@daniel-stoneuk Can you add a mention of the feature in the README ?

pernielsentikaer:
Awesome, it looks like we're almost ready for a merge when README (and maybe CHANGELOG) is updated

extensions/bitwarden/src/components.tsx
daniel-stoneuk: I've refactored this into a hook.

pomdtr: Thanks !

daniel-stoneuk:
Thank you so much @aozaki-kuro and @xorob0 for testing this on your self hosted instances!

pomdtr:
Can you use this PR to fix change `Logout Vault` to `Logout` in the Action Panel ?
 
image

daniel-stoneuk:
Yep, will do that now!

pernielsentikaer:
Is this ready for merge now?

pomdtr:
Go ahead. Thanks for your contribution @daniel-stoneuk !

pernielsentikaer:
Awesome work! Thanks @daniel-stoneuk 🎊

PR #1467, opened 2022-04-22
Add Self-Hosted JIRA Extension

emanguy:
## Description

This extension is a fork of the existing JIRA Search extension by sven which has been modified to work with Self-Hosted JIRA installations. I also removed some commands which worked with JIRA cloud that do not work with Self-Hosted installations. The only thing that's really changed between the two is that all the JIRA API URLs have been updated to use [REST API V2](https://developer.atlassian.com/cloud/jira/platform/rest/v2/intro/) which is available on self-hosted as opposed to [REST API V3](https://developer.atlassian.com/cloud/jira/platform/rest/v3/intro/) which is cloud-only. Additionally, the extension has been updated to use Personal Access Tokens rather than the API tokens used on JIRA Cloud.

## Screencast

The screencast in the assets folder should still apply, as none of the UI has changed.

## Checklist


fe9lix (approved):
Hey @emanguy – thanks for your contribution! I don't have an easy way to test this, but since it's based on the Jira Search extension, that's okay.

PR #582, opened 2021-12-31
Added Kaalam Extension

heertheeswaran:
## Description
Added a new extension called Kaalam, which means Time. 

This shows the current time in available date formats.

## Type of change

- New extension

## Screencast

The extension has only one command which is Show Time

image

## Checklist

- [x] I read the [extension guidelines](https://developers.raycast.com/basics/prepare-an-extension-for-store)
- [x] I read the [documentation about publishing](https://developers.raycast.com/basics/publish-an-extension)

PR #1496, opened 2022-04-25
[New Extension] Now

basefas:
## Description

get now time string and can be format by config.

## Screencast

image

image

result
image

## Checklist

### DIFF
--- a/extensions/kaalam/CHANGELOG.md
+++ b/extensions/kaalam/CHANGELOG.md
@@ -3,3 +3,4 @@
  ## [Update] - 2022-06-13
 
  - Added command to paste formatted date into frontmost active application
+ - Updated Raycast API


koinzhang:
Hi @basefas.
It seems that there is already a similar extension [Kaalam](https://www.raycast.com/heerthees/kaalam) in the store.

basefas:
> 

hello～ I tried Kaalam. The core focus is diiferent between both extensions. Now focus get now time string format and insert it in one step, [now] + [return] is all. User does not change the format frequently, and it can be changed easily in preferences. Use Kaalam you need [show time], [up] or [down] to select string you want, [return], and then you get it only in your clipboard, you must paste it again to insert it where you want. And it can only use the format it provided. We use raycast to make our workflow simplify, isn't it? By the way, Kaalam maybe hardly to be find by user who want to use this function.

pernielsentikaer (commented):
Hi @basefas 👋 

Thanks for your contribution. It could make sense to contribute to the other extension; you're allowed to add another command to that one. I think it's easy enough to find since you can search for things like `time` and still find it.

Could you consider that?

basefas:
> Hi @basefas 👋
> 
> Thanks for your contribution. It could make sense to contribute to the other extension; you're allowed to add another command to that one. I think it's easy enough to find since you can search for things like `time` and still find it.
> 
> Could you consider that?

Hi, I still think they are two different extensions, but if you insist I'll try to merge this feature into kaalam

koinzhang:
@pernielsentikaer @basefas 
Maybe renaming the Kaalam extension would be a good way to go?

What does kaalam mean? I can't seem to find any relation between kaalam and time. If the extension name does not visually reflect the extension, shouldn't the extension be renamed or some other improvement made?

I looked up Kaalam, which seems to be an [Indian novel about time travel](https://en.wikipedia.org/wiki/Kaalam_(novel)), so maybe this is the only point where kaalam is related to time.

pernielsentikaer:
Hi @basefas 👋 

A not long time ago we had a suggestion session about new ideas for "Snippets" and one of the most "voted" was the possibility to add date modification replacements.

I think the best approach would be to merge it into #582, and we could ask @heertheeswaran about the naming

What do you think?

heertheeswaran:
Hi @basefas @koinzhang @pernielsentikaer 

Kaalam means Time in Tamil. 

I think we can rename the extension, so that everybody can understand it easily. Can you guys suggest some name?

basefas:
Hi @pernielsentikaer @koinzhang @heertheeswaran 

Sorry for not replying in time, I have been busy with work last week. I have merged the now function into kaalam, I don't know if my workflow is correct.

About name, I think「Time Tools 」maybe a good choice, we can put more time-related features into this extension and  it is easier for users to find it.

heertheeswaran:
Hi @basefas 

Can you raise a PR for that I will check, if the workflow is correct.

basefas:
> Hi @basefas
> 
> Can you raise a PR for that I will check, if the workflow is correct?

I have edit it under this pr, you can see what changed in [ Files changed ] tab.

basefas:
any updates？

pernielsentikaer:
Hi @basefas 👋 

Sorry for the long response time, somehow I got the notification removed 😞 

I think we can merge this, most of this is however now possible in default Raycast, soon we add support for custom format:

CleanShot 2022-06-13 at 08 14 16@2x

PR #1499, opened 2022-04-25
[Emoji] Disable recently used while searching, fix search glitches

### LINKED ISSUE #1368: [Extension Feature Request] Disable "recently used" for Emoji search
# Extension – [Emoji](https://github.com/raycast/extensions/tree/main/extensions/emoji)

Author: @FezVrasta 

image
Would be great to have an option to disable the "recently used"-section when searching. :) 
To me it happens quite often that the result I'm looking for is not in that section and then I have to hit the down arrow quite a few times to move past the recently-used section and to get to the right result.

daniel-stoneuk:
## Description

A small change to disable the recently used category when the search text is not empty. Fixes #1368

* I've also changed the categories to use the persistent state hook.
* Removed the `useCallback` since this is not necessary here.

## Checklist

### DIFF
--- a/extensions/emoji/CHANGELOG.md
+++ b/extensions/emoji/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Emoji Changelog
 
+## [Fix] - 2022-04-26
+
+- Disables recently used emojis while searching
+
 ## [Fuzzy Search and Categories Dropdown] - 2022-03-21
 
 - New extension icon;
@@ -8,20 +12,20 @@
 
 ## [Add preferences] - 2022-02-11
 
-Added in functionality to take user back to root search after copying an emoji.
+- Added in functionality to take user back to root search after copying an emoji.
 
 ## [Add preferences] - 2021-11-17
 
-Add preferences to Emoji, and print nicer names
+- Add preferences to Emoji, and print nicer names
 
 ## [Fixed typo] - 2021-11-16
 
-Fix Emoji extension PasteAction title typo
+- Fix Emoji extension PasteAction title typo
 
 ## [Fix] - 2021-10-15
 
-Fix typo.
+- Fix typo.
 
 ## [Add Extension] - 2021-10-15
 
-Add Emoji extension
+- Add Emoji extension


extensions/emoji/src/emoji.tsx
```diff
@@ -95,14 +97,7 @@ export default function Main(): ReactElement {
       }
     >
       {!isLoading
-        ? filterList(
-            [
-              { category: "Recently Used", emojis: recentlyUsed },
-              { category: category || "Emojis", emojis: list },
-            ],
```
FezVrasta, line 102: Rather than doing all these changes wouldn't it be simpler to add this section conditionally?

```
            [
              !searchText && { category: "Recently Used", emojis: recentlyUsed },
              { category: category || "Emojis", emojis: list },
            ].filter(Boolean)

Comment (daniel-stoneuk) [reply extensions/emoji/src/emoji.tsx:102]: To do it like this we'd have to cast it like so

```typescript
[
  !searchText && { category: "Recently Used", emojis: recentlyUsed },
  { category: category || "Emojis", emojis: list },
].filter(Boolean) as Category[]
```

or do

```typescript
[
  ...(!searchText ? [{ category: "Recently Used", emojis: recentlyUsed }] : []),
  { category: category || "Emojis", emojis: list },
].
```

I was afraid of putting too much logic in this method arguments, and thought it might be better off filter method. I'm of course open to either of these approaches though, which do you prefer?

FezVrasta: I'd like not to overload the filter function with non-filter related logic. I think your first example would make sense, it would also reduce a lot the line changes of this PR.

FezVrasta: Alternatively we could have an additional function that is responsible of adding the recently used category.

```
addRecentlyUsed([{ category: category || "Emojis", emojis: list }], !searchText).map(... etc
```

FezVrasta (approved):

pernielsentikaer (approved):
Hi @daniel-stoneuk 👋 

Thanks for an update, it works as expected 🔥

bookercodes:
Amazing work @FezVrasta and @daniel-stoneuk, I am so chuffed I get to benefit from this update!

PR #1518, opened 2022-04-27
Adding support for 1Password 8

dteare:
## Description

I worked with @thomaspaulmann to add support for 1Password 8 to the existing 1Password extension. 

## Screencast

CleanShot 2022-04-27 at 11 24 14@2x
screenshot-01
screenshot-02
screenshot-03

## Checklist

```
❯ npm run build

> build
> ray build -e dist

info  - entry points [src/list.tsx]
info  - checked TypeScript
ready - built extension successfully

❯ npm run lint 

> lint
> ray lint

ready - run ESLint
ready - run Prettier
ready - no ESLint or Prettier issues found
```

### DIFF
--- a/extensions/1password7/CHANGELOG.md
+++ b/extensions/1password7/CHANGELOG.md
@@ -1,6 +1,6 @@
 # 1Password Changelog
 
-## [Support for 1Password 8] - 2022-03-27
+## [Support for 1Password 8] - 2022-05-03
 
 - Added support for 1Password 8
 - Added a preference to specify which app version the extension interacts with. By default, it automatically detects the installed one.


khasbilegt:
Hello @dteare @thomaspaulmann,

First of all, thank you all for the contribution. I have been using 1Password 8 for quite some time so I haven't been able to do much about the issues and requests of this extension. However, I was planning to rewrite it using the new 1Password CLI. Correct me if I'm wrong but looks like this changes are possible because 1Password 8 has some backward compatibility on creating metadata files like 1Password 7 does. So in order to integrate with the new CLI would it better to create a new extension? or is it better to do it here but keep the logic separately?

dteare:
Hello @khasbilegt! You're very welcome. And thank you for creating this extension! 🤗 I had reached out to you on the Raycast Slack channel to jam but I'm guessing you don't hang out there any more. 

As for the 1Password CLI, that's exactly right. With the new CLI we have complete flexibility and can use it directly within this Raycast extension. We can even do a lot more like recreate the full item view and even editing if we wanted to, all directly within Raycast and without relying on the metadata files whatsoever. That will be great and is the plan for the future. 

This PR is a stepping stone towards that future by leveraging the [opbookmarks](https://github.com/dteare/opbookmarks) helper which creates the same set of metadata files that 1Password 7 creates. Admittedly it is more complicated and we need to simplify it so this is absolutely just the first step. 

> So in order to integrate with the new CLI would it better to create a new extension? 

Yes! It would be. But that's not my decision to make. 🙂 My original plan was to use [a separate 1Password 8 extension](https://github.com/raycast/extensions/pull/1090) whose implementation would deviate completely to use the CLI and extend the capabilities that we have today. I was unable to convince @thomaspaulmann that this was the best way forward so some gymnastics will be needed to preserve backwards compatibility once we evolve beyond the simple metadata file approach.

PR #600, opened 2022-01-04
Adding BTTV Emotes Extension

Aayush9029:
## Description

Lets you browse [BTTV Emote library](https://betterttv.com/emotes/top) and copy gif, image url directly to clipboard!

## Type of change

- New extension

## Screencast
AShot-2022 01 04-08 12 30

## Checklist

- [x] I read the [extension guidelines](https://developers.raycast.com/basics/prepare-an-extension-for-store)
- [x] I read the [documentation about publishing](https://developers.raycast.com/basics/publish-an-extension)

PR #1529, opened 2022-04-29
New extension: FrankerFaceZ

davidchalifoux:
## Description

This is a new extension which allows you to search the [FrankerFaceZ](https://www.frankerfacez.com/) emote library.
## Screencast
Screen Shot 2022-04-28 at 11 00 13 PM

## Checklist

### DIFF
--- a/extensions/frankerfacez/src/index.tsx
+++ b/extensions/frankerfacez/src/index.tsx
@@ -32,6 +32,7 @@ function SearchListItem({ searchResult }: { searchResult: SearchResult }) {
         <ActionPanel>
           <ActionPanel.Section>
             <Action.CopyToClipboard content={searchResult.urls[1]} />
+            <Action.Paste content={searchResult.urls[1]} />
             <Action.OpenInBrowser url={"https://www.frankerfacez.com/emoticon/" + searchResult.id} />
           </ActionPanel.Section>
           <ActionPanel.Section>


pernielsentikaer (changes requested):
Hi @davidchalifoux 👋 

I have some feedback for you:

- Would it be possible to use our detail view, like it's done in [this extension](https://www.raycast.com/josephschmitt/gif-search) so it's possible to see the images?

It looks like you actually tried to [implement it](https://github.com/raycast/extensions/pull/1529/files#diff-d2c0750574c76bba175dbf083e98ce2c527a7ca694c09036463df82692b057e0R31-R33), let me know if you need help.

Aayush9029:
@davidchalifoux Amazing Work! I've made similar https://github.com/raycast/extensions/pull/600 but for BTTV!
We could try to merge both of these extensions in one and even add 7tv 😄 

What do you think?
>  Completely understandable if you'd rather have separate extensions for each providers! 💪🏼

davidchalifoux:
> @davidchalifoux Amazing Work! I've made similar #600 but for BTTV! We could try to merge both of these extensions in one and even add 7tv 😄
> 
> What do you think?
> 
> > Completely understandable if you'd rather have separate extensions for each providers! 💪🏼

Thanks @Aayush9029 ! Your extension is actually what inspired me to make one for FrankerFaceZ!

I thought about your idea of having one extension for searching all of those services, and I would do that only if they were separate commands within the extension (ie "Search BTTV" and "Search FrankerFaceZ"). I took a look at the "GIF Search" extension which can search a few different GIF services, but I don't like needing to use the dropdown to change providers.

Here's how I would do it:

- Create a new extension called "Search Emotes"
- Create commands for each service and name them something like "Search BTTV" and "Search FrankerFaceZ"
- Use the detail view per @pernielsentikaer's suggestion

Let me know what ya think. @pernielsentikaer I'd love to hear your thoughts too if you have any opinion.

Aayush9029:
@davidchalifoux Don't mind me, I love to throw out some ideas💡

I've been using emotes (extension) a lot lately, replacing GIFs during conversations 😆

Definitely, Looking forward to your extension!

pernielsentikaer:
Hi @davidchalifoux 👋 

It would be awesome to have it as one, the only problem with that is that we might end up with a lot of extensions stalling with an older version of BTTV since we can't uninstall extensions from the users.

I think we're better with having them separated, even though it would be cool!

davidchalifoux:
Added detail view per @pernielsentikaer.
Also cleaned everything up a bit.

Screen Shot 2022-05-03 at 12 19 48 AM

pernielsentikaer:
Hi @davidchalifoux 👋 

Awesome update! Could it be an idea to add a [Paste](https://developers.raycast.com/api-reference/clipboard#paste) action too?

davidchalifoux:
I added a paste action, but it doesn't seem to be working for me. It copies the content to the clipboard but never pastes it. Any ideas @pernielsentikaer?

pernielsentikaer:
It should work since the action is right; I think we should merge it now 🎉 

Does it works from other extensions like the `Clipboard History`?

PR #1541, opened 2022-04-30
Added filter preferences to Google Drive

nametake:
## Description

- Added flag to display directories
  - To help copy local files
- Added exclude paths preference
- Added Reindex action to EmptyView
  - When the Google Drive root path changed, it was necessary to reindex
 
## Checklist

### DIFF
--- a/extensions/google-drive/package.json
+++ b/extensions/google-drive/package.json
@@ -9,6 +9,7 @@
     "Developer Tools",
     "Productivity"
   ],
+  "contributors": ["nametake"],
   "license": "MIT",
   "commands": [
     {


extensions/google-drive/src/index.tsx
```diff
@@ -158,7 +158,29 @@ export default function Command() {
...
+                  />
+                  <Action
+                    title="Clear File Previews Cache"
+                    icon={Icon.Trash}
+                    onAction={clearFilePreviewsCache}
+                    shortcut={{ modifiers: ["ctrl", "shift"], key: "x" }}
+                  />
+                </>
+              )}
+            </ActionPanel>
```
vishaltelangre, line 181: Would it make sense to extract these two actions into a reusable component since they're duplicated?

extensions/google-drive/src/utils.ts
```diff
@@ -42,6 +42,14 @@ export const getDriveRootPath = (): string => {
 
   return resolve(preferences.googleDriveRootPath.trim().replace("~", homedir()));
 };
+export const getExcludePaths = (): string[] => {
```
vishaltelangre, line 45:
```suggestion
export const getExcludePaths = (): Array<string> => {
```

I prefer this style. :)

extensions/google-drive/src/utils.ts
```diff
@@ -58,11 +66,14 @@ export const getDirectories = (path: PathLike): Array<PathLike> =>
...
-export const saveFilesInDirectory = (path: PathLike, db: Database) =>
+export const saveFilesInDirectory = (path: PathLike, db: Database) => {
+  const preferences = getPreferenceValues<Preferences>();
+  const excludePaths = getExcludePaths();
   readdirSync(path).forEach((file) => {
     const filePath = join(path.toLocaleString(), file);
 
-    if (!isFile(filePath)) return;
+    if (!preferences.shouldShowDirectories && !isFile(filePath)) return;
+    if (excludePaths.includes(filePath)) return;
```
vishaltelangre, line 76: Should `getDirectories()` filter `excludePaths` instead?

vishaltelangre (commented):
@nametake Added some minor comments. Otherwise these changes are looking great! 👍 

Also, please update CHANGELOG.md.

extensions/google-drive/src/index.tsx
nametake: https://github.com/raycast/extensions/pull/1541/commits/8923b5fa5545c205db628770d623549a62c2434b

Split as an action components.

extensions/google-drive/src/utils.ts
nametake: No. `saveFilesInDirectory` inserts all files and directories in the passed path, so without this filter, exclude path would show in the results.

extensions/google-drive/src/index.tsx
```diff
@@ -140,25 +164,27 @@ export default function Command() {
...
       ) : (
-        <List.EmptyView title={isFetching ? "Fetching files, please wait..." : "No files found"} />
+        <List.EmptyView
+          title={isFetching ? "Fetching files, please wait..." : "No files found"}
+          actions={
+            <ActionPanel>
+              {!isFetching && (
+                <>
+                  <ReindexFilesCacheAction reindexFiles={reindexFiles} />
+                  <ClearFilePreviewsCacheAction clearFilePreviewsCache={clearFilePreviewsCache} />
```
vishaltelangre, line 182: Don't think we need to pass `clearFilePreviewsCache` as a prop here since it can be called directly from the extracted out action component (as it is already imported as an external dependency).

vishaltelangre (commented):
@nametake Just one more minor comment before we can request this to get merged. Thanks a lot for addressing my previous suggestions!

extensions/google-drive/src/index.tsx
nametake: Thank you!

Fixed it https://github.com/raycast/extensions/pull/1541/commits/2d8e8a360b79302f43b5ef2e29c2538615cf9bf5.

vishaltelangre (approved):
Thank you, @nametake! All changes look good to me. 🎉 

@pernielsentikaer Can you please have a final look? Thanks.

vishaltelangre:
@nametake It would be great if you could add yourself as a `contributors` to this extension. See https://developers.raycast.com/information/manifest#extension-properties.

PR #1567, opened 2022-05-02
massCode: Add detail of the snippet to the view

mkitane:
## Description

Added a detail view to the massCode extension

## Screencast
Capture d’écran 2022-05-03 à 00 32 35

## Checklist

### DIFF
--- a/extensions/masscode-assistant/CHANGELOG.md
+++ b/extensions/masscode-assistant/CHANGELOG.md
@@ -1,5 +1,11 @@
 # massCode Changelog
 
+## [1.0.2] - 2022-05-09
+
+### Features
+
+- Add detail view to display the content of the snippet within raycast
+
 ## [1.0.1] - 2022-05-05
 
 ### Bug Fixes
--- a/extensions/masscode-assistant/src/components/snippet-content.tsx
+++ b/extensions/masscode-assistant/src/components/snippet-content.tsx
@@ -12,7 +12,7 @@ const SnippetContent = ({ snippet, selectedFragment }: { snippet: Snippet; selec
     if (language == "markdown") {
       code = text;
     } else {
-      code =  `~~~${language}
+      code = `~~~${language}
 ${text}
 ~~~
       `;


unnamedd:
Hello @mkitane,

Based on the `<List.Item.Detail />`'s documentation you should avoid to make use of the accessories in the `<List.Item />` and bring those informations to the markdown content.

> When shown, it is recommended not to show any accessories on the List.Item and instead bring those additional information in the List.Item.Detail view.
https://developers.raycast.com/api-reference/user-interface/list#list.item.detail

In the way you presented in the screenshot, you have no space to show the subtitle nor the "Kubernetes" word.

mkitane:
> Hello @mkitane,
> 
> Based on the `<List.Item.Detail />`'s documentation you should avoid to make use of the accessories in the `<List.Item />` and bring those informations to the markdown content.
> 
> > When shown, it is recommended not to show any accessories on the List.Item and instead bring those additional information in the List.Item.Detail view.
> > https://developers.raycast.com/api-reference/user-interface/list#list.item.detail
> 
> In the way you presented in the screenshot, you have no space to show the subtitle nor the "Kubernetes" word.

Good point, I moved that piece of info to the Detail view.

antonreshetov:
@mkitane It looks cool. We also need to update the changelog.

antonreshetov:
@mkitane I don't know if it's an implementation bug, but I faced a random bug. Sometimes the code is cut off in the preview. 
preview

pernielsentikaer (changes requested):
Hi @mkitane 👋 

Thanks for the update 🔥 

It beginning to look quite nice! It seems like breaks disappear when using plaintext, is that something that you can do something about?

CleanShot 2022-05-03 at 05 23 03@2x

mkitane:
> It beginning to look quite nice! It seems like breaks disappear when using plaintext, is that something that you can do something about?

Unfortunately, Raycast only allows markdown for `List.Item.Detail`, which means that the plaintext is treated as markdown (and hence why you see the line break disappear). 
A workaround would be:
-  to wrap it around a code block, just like it is done for the other languages.
- Use a txt to markdown converter, but I couldn't find a good one.

An ideal solution would be to have Raycast support a text field additionally to the markdown one.

How do you think we should approach it?

pernielsentikaer:
Hi @mkitane 👋

I think I would just wrap it inside a block actually 😄 And after this is merged we can look into if we can find a better solution, but this is a nice enhancement that could be awesome to have out public 💯 

mkitane:
> @mkitane I don't know if it's an implementation bug, but I faced a random bug. Sometimes the code is cut off in the preview.

@antonreshetov I updated the changelog.
Also, I can't reproduce this issue, the view is never cutoff for me.

pernielsentikaer (approved):
Awesome @mkitane 🔥

antonreshetov:
> I updated the changelog.

Adding a new feature increases the minor version, not the patch. 
https://semver.org/

antonreshetov:
@pernielsentikaer Can this be fixed?

PR #1582, opened 2022-05-04
Renamed extension

pernielsentikaer:
## Description

Renamed the extension


L3K0V:
@pernielsentikaer does this mean that it's working with 1Password 7 and 1Password 8?

L3K0V:
Nevermind @pernielsentikaer I found it, I will just reference it here: https://github.com/raycast/extensions/pull/1518

pernielsentikaer:
Yes it does, @L3K0V - it should autodetect the version that you have, if not, then you can set it in extension preference 😄

L3K0V:
Yes! Unfortunately I have used 1Password 7 before, and it looks that I have some left overs and the auto detection load old entries even that 1Password 8 is not configured using opbookmarks :( Later today I will try to get it working properly

PR #1612, opened 2022-05-06
Add Converter extension

koinzhang:
## Description

Arbitrary input, direct output. Includes Base converter, Case converter, Code converter, etc.

**🌟🌟Tips:**

- You can turn on automatic detection of selected text or clipboard in the settings, which will automatically fill in the detected text when you enter the command.

**Preferences:**

- Converter:

  - Auto Detect: Convert between different cases: camel, pascal, snake, kebab, upper, lower, title.
  - Priority detection: Automatically detects selected text or clipboard when opening commands.

- Base Converter:

  - Advance View: In the advanced view, you can freely choose the bases of mutual conversion.
  - Advance View Location: Select the location of the advanced view: top or bottom.

## Screencast

Metadata-1
Metadata-2
Metadata-3
Metadata-4

## Checklist


pernielsentikaer:
Hi @koinzhang 👋 

Almost all of this is already in both possible in[Toolbox](https://www.raycast.com/Kang/toolbox) and [Change Case](https://www.raycast.com/es183923/change-case) - is it on purpose that you're creating an extension with this?

I'm not sure your title case is correct since it shouldn't uppercase the first character in every word, [see this one](https://www.raycast.com/es183923/change-case)

koinzhang:
Hi @pernielsentikaer 
I originally made only Base Converter and Code Converter.
Base Converter: is used to convert binary, which seems to be no relevant extension in the store nowadays
Code Converter: is used to convert the character encoding, such as from native to unicode, ASCII, UTF-8, these are the key elements.
Later, because I wanted to create a Opacity conversion extension, I added Color Converter directly (in two parts, Color Converter and Opacity Converter). Among the features of Opacity Converter that are not available in the store now.

As for Case Converter is additional content, indeed and other extensions have duplicate places, I will delete Case Converter

koinzhang:
Hi @pernielsentikaer 
I have removed the Case Converter, please keep reviewing.
Base Converter: This command is not duplicated at all
Code Converter: `Base64` and `Encode URL` are two functions available in other extensions, other items are not duplicated
Color Converter: `Color` function is also available in other extensions, `Opacity` function is not duplicated。

I can also remove the above duplicate functions if needed.

pernielsentikaer (approved):
Let's get it merged 💪

PR #1613, opened 2022-05-06
[Google Drive] Faster indexing with progress

vishaltelangre:
## Description

- Added comparatively faster indexing mechanism by fetching the nested paths (along with file stats) with the help of [`fast-glob`](https://github.com/mrmlnc/fast-glob) package.
- Now it will show the indexing progress in percentage + processed/total format.

## Screencast

https://user-images.githubusercontent.com/876195/167167255-757ad8f5-5726-48ae-b212-92539e85a94d.mp4

## Checklist

### DIFF
--- a/extensions/google-drive/src/constants.ts
+++ b/extensions/google-drive/src/constants.ts
@@ -7,11 +7,11 @@ export const IGNORED_GLOBS = [
   "**/.git",
   "**/tmp",
   "**/.DS_Store",
-  ".Trashes",
-  ".VolumeIcon.icns",
-  ".TemporaryItems",
-  ".shortcut-targets-by-id",
-  ".file-revisions-by-id",
+  "**/.Trashes",
+  "**/.VolumeIcon.icns",
+  "**/.TemporaryItems",
+  "**/.shortcut-targets-by-id",
+  "**/.file-revisions-by-id",
 ];
 export const FILE_SIZE_UNITS = ["B", "KB", "MB", "GB", "TB"];
 export const NON_PREVIEWABLE_EXTENSIONS = [


extensions/google-drive/package.json
```diff
@@ -58,15 +60,20 @@
...
   },
   "devDependencies": {
+    "@types/download": "^8.0.1",
     "@types/node": "~16.10.0",
     "@types/react": "^17.0.28",
     "@types/sql.js": "^1.4.3",
+    "@types/tar": "^6.1.1",
     "@typescript-eslint/eslint-plugin": "^5.0.0",
     "@typescript-eslint/parser": "^5.0.0",
+    "electron": "^18.2.0",
```
vishaltelangre, line 76: The `download` package doesn't work otherwise.

vishaltelangre:
@pernielsentikaer Can you please test + review? Thanks.

extensions/google-drive/package.json
daniel-stoneuk: How big is the `electron` package? I imagine this is a rather large download to include in the extension.

extensions/google-drive/src/constants.ts
```diff
@@ -28,3 +29,9 @@ export const MAX_TMP_FILE_PREVIEWS_LIMIT = 500; // Average size of a file previe
 export const DB_FILE_PATH = join(homedir(), ".raycast-google-drive-sqlite.db");
 export const TMP_FILE_PREVIEWS_PATH = join(tmpdir(), "raycast-google-drive-file-previews");
 export const FILES_LAST_INDEXED_AT_KEY = "filesLastIndexedAt";
+export const FD_PATH = path.join(environment.supportPath, "fd");
+export const FD_ARCHIVE_URL =
+  "https://github.com/sharkdp/fd/releases/download/v8.3.2/fd-v8.3.2-x86_64-apple-darwin.tar.gz";
```
daniel-stoneuk, line 34: Does this binary run on M1 natively or will it have to use Rosetta?

extensions/google-drive/package.json
```diff
@@ -58,15 +60,20 @@
   ],
   "dependencies": {
     "@raycast/api": "^1.33.2",
+    "download": "^8.0.0",
```
daniel-stoneuk, line 63: Is this package necessary? Could we download the file and then unzip with the system utililites - removes the need for `electron`?

daniel-stoneuk (commented):
This looks like a cool update, especially with the progress bar. That'll make it much easier to estimate how long the indexing is going to take. Hopefully more users will be inclined to wait once they see some progress.

I do wonder if using the external binary `fd` is a great idea for this? 

I have come across https://github.com/mrmlnc/fast-glob and the derivative https://github.com/sindresorhus/globby. These look like packages that would be great for indexing all the files if they work, and use built-in Node features.

`fast-glob` claims it's fast*, though I wonder if this means it can compare to the parallelisation used by `fd`. At some point the Google Drive application will be a bottleneck, so it might be worth doing some comparisons.

Would you be open to experimenting with these two packages? I'd love to help out as well, but have limited time over the next two weeks. After that, though, I'll have lots of time!

*from a quick read, it looks like it does fewer system calls by fetching directory contents and details with a single call.

vishaltelangre:
@daniel-stoneuk Thanks for your thoughts. I am open to ideas. The only reason for me to switch to fd is to utilize all the CPU cores in parallel to perform the indexing. Although there may be a way to perform multi-threaded processing in nodejs, it appears that the Raycast extension can operate in a single thread of its own. We _also_ want to collect the stats (especially file's modified time so we can sort the results by their recency).

The proposed approach does that and seemed to be comparatively fast to me.

So if there's a comparatively even faster approach (using native nodejs primitives) I will be more than happy.

extensions/google-drive/package.json
vishaltelangre: I don't like it either. I will actually kill the `download` package later.

extensions/google-drive/src/constants.ts
vishaltelangre: My mac runs on a Intel chip but I'm hopeful that it can work on an M1 as well.

daniel-stoneuk:
As we discussed on Slack, I have run a few tests with the different options:

All of these get the dates of file modification and creation

* Extension currently on Store: 2m14s
* This PR: 1m21s
* This PR with toast and database insertion removed: 30s
* `globby` Script: 1m4s
* `fd` Script: 37s

<details>
 <summary>`globby` Script</summary>
```javascript

import { globby } from 'globby';
const paths = await globby(['/Users/username/Google Drive/My Drive', "!.Trashes", "!.TemporaryItems", "!node_modules", "!.git", "!tmp"], { stats: true });
console.log(paths)
```
</details>

<details>
<summary>`fd` script</summary>
```javascript
import { spawn } from "child_process";

const test = new Promise((resolve) => {
    const child = spawn('fd -L -H --type file --exclude node_modules --exclude bower_components --exclude .git --exclude tmp . "/Users/username/Google Drive/My Drive" -x stat -L -f "%SB | %Sm | %z | %N"', { shell: true });
    let filesIndexed = 0;

    child.stdout.on("data", (data) => {
        data
            .toString()
            .split("\n")
            .filter((entry) => entry.trim().length > 0)
            .map((entry) => entry.split(" | ").map((entry) => entry.trim()))
            .forEach((entry) => {
                if (entry.length != 4) return;
                const [createdAt, updatedAt, sizeInBytes, path] = entry;

                // if (path === "" || isDotUnderscore(path)) return;

                filesIndexed += 1;

                // toast.message = `Progress: ${Math.round((filesIndexed / totalFiles) * 100)}% (${filesIndexed}/${totalFiles})`;
                // insertFile(db, {
                //     name: basename(path),
                //     path,
                //     displayPath: displayPath(path),
                //     fileSizeFormatted: formatBytes(sizeInBytes.length > 0 ? parseInt(sizeInBytes, 10) : 0),
                //     createdAt: createdAt.length > 0 ? new Date(createdAt) : new Date(),
                //     updatedAt: updatedAt.length > 0 ? new Date(updatedAt) : new Date(),
                //     favorite: false,
                // });
            });
    });

    child.on("close", (code) => {
        console.log(`code: ${code}`)

        resolve();
    });
});

test.then((result) => {
    console.log("done");
})
```
</details>

daniel-stoneuk:
From these results it looks like Raycast adds a bit of overhead, and it's possible that Google Drive FS is bottlenecking the listing process.

The following screenshot is from running this PR. I imagine the Raycast usage is so high because of the Toast, but I think the Google Drive usage indicates that it working as hard as it can to provide files.

Screenshot 2022-05-08 at 16 44 51

However, it does appear that `fd` is faster in testing although I am not convinced it is using multiple processes to do this. Hence the speedups might just be from running in a faster language (Rust). The following screenshot shows that the `fd` process in this PR only spins up one thread.

Screenshot 2022-05-08 at 17 29 20

Hopefully this analysis has been useful!

vishaltelangre:
> globby Script: 1m4s

Is this calculated with progress toast update + SQLite insertion logic or without it? Would be good to get the numbers mimicking the same circumstances for both the approaches.

daniel-stoneuk:
> Is this calculated with progress toast update + SQLite insertion logic or without it? Would be good to get the numbers mimicking the same circumstances for both the approaches.

This is by running the attached script at the end of the comment. 

I just changed the implementation of this PR to use globby, and it runs in 39 seconds!

vishaltelangre:
Nice! I have tested it and it appears that `globby` would be a good choice!

Based on my testing (and based on how many files I have in my Google Drive) the difference wasn't very noticeable but having NOT to deal with a third-party binary is definitely an advantage!

@daniel-stoneuk I am going to add you as a collaborator to my fork so you can make the approach changes to this PR for `globby`. Thanks for the suggestions and all the help!

Please also note that (as discussed on Slack), I needed to explicitly specify `{ dot: true }` so that it could index the nested dot directories as well. So please accommodate that change.

extensions/google-drive/src/db.ts
```diff
@@ -169,27 +171,79 @@ export const insertFile = (
...
+      updatedAt: (stats.mtime ? stats.mtime : new Date()).toISOString(),
+      favorite: false,
+    });
+  }
 };
 
 type IndexFilesOptions = { force?: boolean };
 export const indexFiles = async (
-  path: PathLike,
+  path: string,
```
vishaltelangre, line 233: Can we keep this `PathLike` and then wherever required use `path.toLocaleString()`?

extensions/google-drive/src/db.ts
```diff
@@ -169,27 +171,79 @@ export const insertFile = (
...
-  getDirectories(path).map((dir) => walkRecursivelyAndSaveFiles(dir, db));
+const listFilesAndInsertIntoDb = async (path: string, db: Database, toast: Toast): Promise<void> => {
+  const updateToastMessage = throttledUpdateToastMessage(toast);
+
+  let totalFiles = 0;
+
+  if (!pathExists(path)) {
+    // TODO: message that path does not exist
+    return;
+  }
```
vishaltelangre, line 189: Don't think we need it here. We already check for the `drivePath` to exist and show a message in `index.tsx`.

daniel-stoneuk: Cool, I'll remove it from here.

extensions/google-drive/src/db.ts
```diff
@@ -169,27 +171,79 @@ export const insertFile = (
...
+
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
vishaltelangre, line 191: 😓 I hate this rule.

extensions/google-drive/src/db.ts
daniel-stoneuk: I noticed here that the only time this was being used was with the `string` provided by the preference, so I went through the codebase and tried to make it clearer where only string paths are used Vs the PathLike from Node's `fs`.

extensions/google-drive/src/db.ts
```diff
@@ -169,27 +171,79 @@ export const insertFile = (
...
+    suppressErrors: true,
+  })) {
+    totalFiles += 1;
+    updateToastMessage(`Counting files: ${totalFiles}`);
+  }
+
+  let filesIndexed = 0;
+
+  for await (const file of fg.stream([join(path, "**")], {
+    ignore: ["**/.Trashes", "**/.TemporaryItems", "**/node_modules", "**/.git", "**/tmp"],
```
vishaltelangre, line 204: Same here.

extensions/google-drive/src/db.ts
```diff
@@ -169,27 +171,79 @@ export const insertFile = (
...
+    dot: true,
+    suppressErrors: true,
+    stats: true,
+  })) {
+    const { name, path, stats } = file as unknown as Entry;
```
vishaltelangre, line 209: The `file` has `string | Buffer` type here. Can we explicitly switch-case over these ADTs to avoid any runtime errors?

extensions/google-drive/src/index.tsx
```diff
@@ -48,7 +48,9 @@ const Command = () => {
   const [filesIndexGeneratedAt, setFilesIndexGeneratedAt] = useState<Date | null>(null);
   const db = useDb();
 
-  useEffect(initialSetup, []);
+  useEffect(() => {
+    initialSetup();
+  }, []);
```
vishaltelangre, line 53: Can revert this back to how it was before.

extensions/google-drive/src/types.ts
```diff
@@ -3,8 +3,8 @@ export type FileInfo = {
   path: string;
   displayPath: string;
   fileSizeFormatted: string;
-  createdAt: Date;
-  updatedAt: Date;
+  createdAt: string;
+  updatedAt: string;
```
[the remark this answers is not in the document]
daniel-stoneuk: When being put into the database it's passed as a Date, but as its taken out it's passed as a string. The typings say `Date` even though it's a string - hence the `new Date(...)` in the markdown

extensions/google-drive/src/utils.ts
```diff
@@ -148,7 +115,7 @@ const clearLeastAccessedFilePreviewsCache = (previewFiles: Array<string>) => {
   });
 };
 
-export const initialSetup = () => {
+export const initialSetup = async () => {
```
vishaltelangre, line 118: Think there's no need to make this an async function anymore.

extensions/google-drive/src/utils.ts
```diff
@@ -186,10 +153,23 @@ ${file.fileSizeFormatted}
...
-${file.createdAt.toLocaleString()}
+${new Date(file.createdAt).toLocaleString()}
 
 ---
 
 **Updated**\n
-${file.updatedAt.toLocaleString()}`;
+${new Date(file.updatedAt).toLocaleString()}`;
+};
+// TODO: Check that the new Date() is not needed
```
vishaltelangre, line 163: ✂️ Please fix and kill this TODO.

extensions/google-drive/src/utils.ts
```diff
@@ -186,10 +153,23 @@ ${file.fileSizeFormatted}
...
+
+export const throttledUpdateToastMessage = (toast: Toast) => {
+  const interval = 100; // TODO: move to constants
```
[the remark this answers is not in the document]
daniel-stoneuk: Resolved in 0764c98

extensions/google-drive/src/utils.ts
daniel-stoneuk: Resolved in ec741b6

extensions/google-drive/src/index.tsx
daniel-stoneuk: Resolved in ec741b6

vishaltelangre: I don't see `clearLeastAccessedFilePreviewsCache()` as an async function either which is run from `initialSetup()` so no point in having async...await stuff here.

extensions/google-drive/src/utils.ts
```diff
@@ -186,10 +153,21 @@ ${file.fileSizeFormatted}
 ---
 
 **Created**\n
-${file.createdAt.toLocaleString()}
+${new Date(file.createdAt).toLocaleString()}
```
vishaltelangre, line 156: @daniel-stoneuk In ref to this - https://github.com/raycast/extensions/pull/1613/files#r868024738, the earlier version `file.createdAt.toLocaleString()` was printing the datetime in the user's locale, AFAIK.

extensions/google-drive/src/utils.ts
vishaltelangre: ^ That commit seems unrelated?

extensions/google-drive/src/db.ts
```diff
@@ -169,27 +171,79 @@ export const insertFile = (
...
+  let totalFiles = 0;
+
+  if (!pathExists(path)) {
+    // TODO: message that path does not exist
+    return;
+  }
+
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  for await (const _ of fg.stream([join(path, "**")], {
+    ignore: ["**/.Trashes", "**/.TemporaryItems", "**/node_modules", "**/.git", "**/tmp"],
```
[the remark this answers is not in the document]
daniel-stoneuk: Resolved in bd8c8b4

extensions/google-drive/src/db.ts
daniel-stoneuk: Resolved in bd8c8b4

extensions/google-drive/src/utils.ts
```diff
@@ -161,6 +164,33 @@ ${new Date(file.createdAt).toLocaleString()}
 ${new Date(file.updatedAt).toLocaleString()}`;
 };
 
+export const fileStream = ({ stats = false }: { stats?: boolean } = {}) => {
```
vishaltelangre, line 167:
```suggestion
type FileStreamOptions = { stats?: boolean };
export const fileStream = ({ stats }: FileStreamOptions = {}): Transform => {
```

for better readability

extensions/google-drive/src/utils.ts
```diff
@@ -161,6 +164,33 @@ ${new Date(file.createdAt).toLocaleString()}
...
+  const driveRootPath = getDriveRootPath();
+  const preferences = getPreferenceValues<Preferences>();
+
+  const filter = new Transform({
```
vishaltelangre, line 171:
```suggestion
  const transform = new Transform({
```

extensions/google-drive/src/utils.ts
```diff
@@ -161,6 +164,33 @@ ${new Date(file.createdAt).toLocaleString()}
...
+    objectMode: true,
+    transform(chunk: Entry, encoding, callback) {
+      // Don't include directories if the preference is false
+      callback(null, chunk.dirent.isDirectory() && !preferences.shouldShowDirectories ? undefined : chunk);
+    },
+  });
+
+  const excludePaths = getExcludePaths()
+    .concat(IGNORED_DIRECTORIES)
+    .map((p) => path.join("**", p));
```
vishaltelangre, line 181: We don't want globbing syntax for the `excludePaths` preference (since they'd be absolute paths.

It is fine for the `IGNORED_DIRECTORIES`.

extensions/google-drive/src/utils.ts
vishaltelangre, line 167: Can we name this explicitly to something like `drivePathsStream()` since it is not a generic stream/transform?

daniel-stoneuk: Resolved in [397e485](https://github.com/raycast/extensions/pull/1613/commits/397e48545d88d8c9a479a66a88f434e896478825)

extensions/google-drive/src/utils.ts
daniel-stoneuk: Removed in favour of the `node-glob` function.

extensions/google-drive/src/utils.ts
```diff
@@ -48,10 +43,11 @@ export const getExcludePaths = (): Array<string> => {
   return preferences.excludePaths
     .split(",")
     .map((p) => p.trim())
-    .map((p) => p.replace("~", homedir()))
+    .filter((p) => p.length > 0)
+    .map((p) => p.replace("~", homedir()).replace(`${getDriveRootPath()}/`, ""))
```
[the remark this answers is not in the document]
daniel-stoneuk: Resolved in [e4b1bc2](https://github.com/raycast/extensions/pull/1613/commits/e4b1bc2ac7f096e6d36177754697c1be5d1cd64c)

extensions/google-drive/src/types.ts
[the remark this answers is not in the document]
daniel-stoneuk: I think before this PR they were being passed as `string`s by `statement.getAsObject()`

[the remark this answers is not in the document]
vishaltelangre: I see. This makes sense to me in that case. Also, the dates in the side panel look readable so we are good here!

vishaltelangre (commented):
Many thanks, @daniel-stoneuk for working hard to get this done! 🙏 

This is looking great to me! 

@pernielsentikaer Can you please review + test? Thanks.

extensions/google-drive/src/index.tsx
```diff
@@ -225,16 +211,18 @@ type ListItemProps = {
...
   return (
     <List.Item
       id={`${idPrefix}${file.displayPath}`}
       key={file.displayPath}
       icon={{ fileIcon: file.path }}
       title={file.name}
-      accessoryIcon={file.favorite ? "⭐" : undefined}
-      detail={<List.Item.Detail markdown={fileDetailsMarkup} />}
+      accessories={
+        file.favorite ? [{ icon: { source: Icon.Star, tintColor: Color.Yellow }, tooltip: "Favorite" }] : undefined
```
daniel-stoneuk, line 223: I changed the emoji to a Raycast Icon here, let me know if this is okay.

pernielsentikaer (changes requested):
It seems like pinned files don't have any metadata, if I find the file in the regular list it has the data:

CleanShot 2022-05-09 at 22 10 26@2x

CleanShot 2022-05-09 at 22 11 34@2x

pernielsentikaer:
Much better @daniel-stoneuk 🔥 Let's await final feedback from @vishaltelangre then and merge after 😄

daniel-stoneuk:
Were you able to test the re-indexing? The `.VolumeIcon.icns` shouldn't be showing after a re-index, so that could be another bug. There's a way to force it in the action menu. Cheers

pernielsentikaer:
Could be a bug then, I also have some folders 😄 

CleanShot 2022-05-09 at 22 31 53@2x

daniel-stoneuk:
Fantastic, thanks for checking this! I have too many files in my Google Drive to index everything for testing. It takes ages, even with these speedups!

pernielsentikaer (approved):
Hi @daniel-stoneuk 👋 

That's much better, now dotfiles are correctly hidden! 🔥 

I couldn't find more, let's see with @vishaltelangre before merge 😄

vishaltelangre:
@pernielsentikaer LGTM! 🚢

pernielsentikaer (approved):
Let's get it merged then 🔥 

Good work both, awesome to see 💪

PR #1629, opened 2022-05-08
Add Twitter v2 extension

tonka3000:
## Description

⚠️ Work in Progress ⚠️

Add Twitter V2 extension.

* Login with Twitter (no API key needs to be provided by the user anymore)
* `Recent Tweets` command
* `My Recent Tweets` command
* `Send Tweet` command
* `Search User` command

  There is no API in v2 to make it in Raycast directly. The current impl. use the input from the user and open twitter.com in the browser with the given search query.

### Why?
Twitter don't offer all features like v1 API does (v2 is not finished). The v1 Twitter extension for Raycast could therefore not be upgraded to v2 because this would only work with adding error views or remove half of the feature set (`Search User`). This extension provide the feature which works with v2. If v2 get the missing features we can easily upgrade it. If a user miss some feature he/she can use the v1 extension.

UPDATE:
- Twitter updated v2 to get the timeline, which is already added to this PR
- Many features are now working, but following are missing
  - Like and Unlike are 2 actions, because there is no way in v2 if the current user liked a given tweet or not
  - Retweet works as before, but unretweet needs to be done on the retweet itself. In v1 it was done on the original one
- I'm think about back-porting it to the original twitter extension, but based on the preferences from the user if would use the old impl. or the new impl., so both v1 and v2 live next to each other

### Todo
- [x] Add most features what v1 extension has
- [ ] Add fixed clientid
- [ ] Add Unretweet action
- [x] Add refresh tweets option

     The problem is that the v2 API use expansions, which sometimes don't work for every tweet and throw erros. Tweets need to remember which expansions they have used. 

## Checklist

### DIFF
--- a/extensions/twitterv2/src/components/send.tsx
+++ b/extensions/twitterv2/src/components/send.tsx
@@ -212,7 +212,7 @@ export function TweetSendThreadForm(): ReactElement {
       }
     >
       {tweets.map((t, index) => (
-        <TweetFragment index={index} content={t} onTextChange={updateTweet} />
+        <TweetFragment key={index} index={index} content={t} onTextChange={updateTweet} />
       ))}
     </Form>
   );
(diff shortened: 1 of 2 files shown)


tonka3000:
Closed in favor of #2087

PR #1664, opened 2022-05-11
[Toothpick] add address and battery information

albertoxamin:
## Description

Add bluethooth device address and battery information for connected devices

## Screencast

image

## Checklist

### DIFF
--- a/extensions/toothpick/CHANGELOG.md
+++ b/extensions/toothpick/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Toothpick Changelog
 
-## [Add battery level indicator and device address] - 2022-05-16
+## [Add battery level indicator and device address] - 2022-05-19
 
 - Add battery level indicator for each connected Bluetooth device
 - Add Bluetooth device address in menu


albertoxamin:
It could be useful to distinguish devices with the same name, 
However I recognize that it would be better if that information would be shown only when a certain key is pressed like in the macos menu
image

albertoxamin:
An alternative 
image

briankosw:
In terms of solving the scenario you mentioned, where the user has devices with the same name and therefore uses the addresses to distinguish between them, I feel like the UX is a bit clunky. In this scenario, the user will have to first go to one device and open up menu and then go to the other device and open up the menu again to compare addresses.

Having each device's address in the menu is fine for its own sake (and so I think it's fine to have the device address in the menu), but I don't think it specifically improves on the scenario of distinguishing between devices with the same name.

extensions/toothpick/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Toothpick Changelog
 
+## [Add Battery Level] - 2022-05-11
+
+Add more info to bluetooth connections
+
```
briankosw, line 6:
```suggestion
## [Add battery level indicator and device address] - 2022-05-11

- Add battery level indicator for each connected Bluetooth device
- Add Bluetooth device address in menu
```

pernielsentikaer:
Is this ready to be merged and is everyone happy with the PR?

pernielsentikaer (changes requested):
@albertoxamin 👋 

I often end up with `undefined` in the list. Is this something we can look into improving?

CleanShot 2022-05-17 at 07 06 21@2x

pernielsentikaer:
Hi @albertoxamin 👋 

`i` is correct, in my case, it's 0,1,2.

CleanShot 2022-05-18 at 13 53 50@2x

When I run this with `npm run dev` and keep hitting ⌘+r - both device status and the battery is off even though the data in the console.log is right. You should be able to reproduce it, it's happening quite often here.

Let me kniw if you need help debugging it.

albertoxamin:
@pernielsentikaer this should prevent the undefined to ever showing, let me know if this works for you.

pernielsentikaer (approved):
That's better, I still get disconnected icon from time to time, but it's really rare

Let's get it merged 💯 

Thanks for your contribution! 🔥

PR #1690, opened 2022-05-13
Stock Lookup

jones-sam:
## Description

This extension lets you search and view information about a stock via the alphavantage stock API. To run the extension you must get a free API key from [alphavantage]( https://www.alphavantage.co/support/#api-key). (not affiliated with me) Prices are from the previous trading day.

## Screencast

https://user-images.githubusercontent.com/45644425/168375115-980c2252-99f3-4414-9e7a-703b63cdc241.mov

## Checklist

### DIFF
--- a/extensions/stock-lookup/src/StockInfo.tsx
+++ b/extensions/stock-lookup/src/StockInfo.tsx
@@ -104,8 +104,6 @@ As of the latest trading day on ${stockInfo?.lastTradingDay}:
       markdown={md}
       metadata={
         <Detail.Metadata>
-          <Detail.Metadata.Label title="Links" icon={Icon.Link} />
-
           <Detail.Metadata.Link target={googleFinanceUrl} text="Google Finance" title="Google" />
           <Detail.Metadata.Link target={yahooFinanceUrl} text="Yahoo Finance" title="Yahoo" />
           <Detail.Metadata.Link target={tradingViewUrl} text="Trading View" title="Trading View" />


pernielsentikaer (changes requested):
Hi @jones-sam 👋 

I have now tested your extension and I have some feedback for you to look into 💪 

- Is the API key necessary? I just used the API key `hello` and I can't see anything not working?

CleanShot 2022-05-16 at 14 01 49@2x

CleanShot 2022-05-16 at 14 21 26@2x

- I would remove that view and use `preferences` in `package.json` instead, that makes more sense (and then write in the README how to get it)

jones-sam:
Hey @pernielsentikaer,

Thanks for the feedback! I made the changes you requested.

Interestingly, it seems like the API key that is passed to the Alphavantage API on the free tier doesn't seem to have any validation on their end, so that is why when you used `hello` it worked with no issues. I changed it so now it uses a random uuid as the API key, unless you have an API key set in the preferences for the extension.

pernielsentikaer:
Hi @jones-sam 👋 

It seems like we should introduce the API key again as `required` since there is a heavy rate limit going on - could you look into this?

I also changed the code a bit, hope it's okay ❤️

jones-sam:
Hey @pernielsentikaer,

I've experimented with the API keys and it seems like if you put a free API key in the preferences or if the extension uses a random string every request it ends up with the same rate limitation. On their end they must be limiting the requests from the IP of the requester or something similar. Unfortunately I don't think there is a way to stop the rate limiting without purchasing a premium API key.

Maybe their is another API service with a better free tier, but most of these APIs have something similar

pernielsentikaer (approved):
Awesome @jones-sam 🔥 

I think in the next version we can maybe add a lite cache somehow 😄

jones-sam:
Awesome, thank you @pernielsentikaer! 🎉

PR #1691, opened 2022-05-13
feat(todoist): Add task from project + task actions for detail

thomaslombart:
## Description

- Added the ability to add a new task from a project and a project's empty state
- Shared the task actions between the task detail and the task list item

### Add new task from project

https://user-images.githubusercontent.com/16003285/168386178-3f4e66c1-4aee-4e42-8e9d-fd14162a7360.mp4

### Actions for task detail

CleanShot 2022-05-13 at 22 40 23@2x

## Checklist

### DIFF
--- a/extensions/todoist/CHANGELOG.md
+++ b/extensions/todoist/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 - Added the ability to add a new task from a project
 - Shared the task actions between the task detail and the task list item
+- Add flag icon for priorities in the detail view as well as in the action panel
+
 ## [Edit your projects] - 2022-04-30
 
 - Added a new action on projects to edit them


thomaslombart:
It seems there's a bug in the PR. If I update a task from the task detail view, it's not updated afterward but the list item is:

https://user-images.githubusercontent.com/16003285/168387631-105af508-a594-47ae-ac8a-9830623eaa73.mp4

It may be a bug on Raycast's end. It seems that props passed in a `target` prop through an `Action.Push` are not updated correctly. 

What do you think? Am I doing something wrong?

pernielsentikaer:
Hi @thomaslombart 👋 

Thanks for this update, we'll take a look at the possible bug and get back! I'll wait with the merge until we have some information about it

mathieudutour:
So it's not a bug per se, but I can see how it can be misconceived. Basically when the `Action.Push` is triggered, it takes the component that is the target at that point and push it to the navigation stack. 
It means that the target isn't a child of the List and so it doesn't get re-rendered when the list is updated.
I guess using the `useNavigation` and `push` method directly makes it more obvious.

So maybe, to solve your specific issue, you could create the Detail from just an id (that's not mutated) so that when the data is updated, the detail is updated as well?

thomaslombart:
Thanks for the answer. It makes total sense! 
I'll look into that then.

thomaslombart:
A small comment to let you know that I haven't had the time to make the modifications. I still have that in mind, though. Hopefully, I'll get it done this week 🙂

thomaslombart:
@pernielsentikaer It's ready for review again 🙂 

The only downside in fixing the above is that there's a slight loading for each task detail. I intend to fix that by adding persistent cache next week 🚀

pernielsentikaer (changes requested):
Hi @thomaslombart 👋 

Thanks for the update. It's amazing how you can keep improving like that 💪 

Around every two or third time I change the list in `Search Project`, it's loading the wrong list - is that a timing issue we can look into?

https://user-images.githubusercontent.com/38033107/170420352-5d2611f0-feef-42d1-a3a0-169f82144444.mp4

thomaslombart:
@pernielsentikaer Under the hood, [SWR](https://swr.vercel.app/) is used to fetch the different pieces of data including the task list which shares the same key (`Tasks`) across the different commands. So when you change projects, SWR will use the key to display what it had in cache (the previous task list or nothing for the first load), then revalidate and show you the right list. This is why you have this "flickering" list.

This is indeed a bug but it has been there since the beginning. I'm happy to fix it with the persistent cache update as it'll be closely tied to how SWR is used in the extension. Should I do it in this PR? 🙂

pernielsentikaer (approved):
Let's just merge it since this is not a new problem and it will be handled in the next PR 😄

PR #1709, opened 2022-05-16
[Spotify Player] Add the ability to open actions in the Spotify app

### LINKED ISSUE #1684: [Spotify Player] Allow show track/album/playlist actions to open directly in app if present
# Extension – [Spotify Player](https://github.com/raycast/extensions/tree/099fc2f9885620fe64c2c0ed3d97838b94bf4f68/extensions/spotify-player/)
Author: @mattisssa

Right now, each command has a "Show (Track, Album, Playlist)" action which uses `OpenInBrowser` to allow doing just that. I would like the option to open each item directly in the Spotify app.

**Who will benefit with this feature?**
Anyone who has the Spotify desktop app installed, and prefers the app instead of a web browser.

**Any Other info.**
Personally I think it would make sense to for this action to open in the Spotify app by default, only falling back to the browser if Spotify is not installed. But if this is implemented as a separate action that would also work (ideally with its own keyboard shortcut).

I believe this would be possible using [Spotify's URL schemes](https://developer.spotify.com/documentation/general/guides/content-linking-guide/).

andreaselia:
## Description

This adds the ability for the existing actions within the extension to open via the Spotify app instead of always using Spotify Web for everything. This also fixes #1684.

## Checklist

### DIFF
--- a/extensions/spotify-player/CHANGELOG.md
+++ b/extensions/spotify-player/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Spotify Player Changelog
 
+## [Feature] - 2022-05-23
+
+- Add the ability to open via the Spotify app instead of always using Spotify Web for everything
+
 ## [Feature] - 2022-04-26
 
 - Added playlist and album search


djpowers (approved):
This looks great, thanks for the work to add this! 🎉

Only thing I can think of is it might be worth adding a quick line to the changelog about the update.

andreaselia:
@djpowers all sorted thanks to @pernielsentikaer 😁

PR #1715, opened 2022-05-17
fix(amazon-aws): manage the case where there is no public ip in EC2

duboiss:
## Description
Remove public ip information & action when there is none.

**BEFORE**
CleanShot 2022-05-17 at 20 14 21@2x

**AFTER**
CleanShot 2022-05-17 at 20 13 28@2x

### DIFF
--- a/extensions/amazon-aws/CHANGELOG.md
+++ b/extensions/amazon-aws/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Amazon AWS Changelog
 
-## [Bug fix for EC2 Instances] - 2022-05-17
+## [Bug fix for EC2 Instances] - 2022-05-23
 
 - Manage the case where there is no public ip in EC2 ([#1715](https://github.com/raycast/extensions/pull/1715))
 


gebeto (approved):

extensions/amazon-aws/src/ec2.tsx
```diff
@@ -69,13 +69,29 @@ function InstanceListItem(props: { instance: AWS.EC2.Instance }) {
...
+  ipActions.push(copyPrivateIpAction);
+
+  if (instance.PublicIpAddress) {
+    subtitle = instance.InstanceType + " (🔒 " + instance.PrivateIpAddress + " / 🌍 " + instance.PublicIpAddress + ")";
+    ipActions.push(
+      <CopyToClipboardAction key="publicIpAction" title="Copy Public IP" content={instance.PublicIpAddress} />
+    );
+  } else {
+    subtitle = instance.InstanceType + " (🔒 " + instance.PrivateIpAddress + ")";
+  }
```
gebeto, line 87: Let's use useMemo here

```tsx
const subtitle = React.useMemo(() => {
  if (instance.PublicIpAddress) {
    return `instance.InstanceType (🔒 ${instance.PrivateIpAddress} / 🌍 ${instance.PublicIpAddress})`;
  }
  return `${instance.InstanceType} (🔒 ${instance.PrivateIpAddress})`;
}, [instance]);

const ipActions = React.useMemo(() => {
  const result = [];
  const copyPrivateIpAction = (
    <CopyToClipboardAction
      key="privateIpAction"
      title="Copy Private IP"
      content={instance.PrivateIpAddress || ""}
    />
  );
  result.push(copyPrivateIpAction);
  if (instance.PublicIpAddress) {
    result.push(
      <CopyToClipboardAction
        key="publicIpAction"
        title="Copy Public IP"
        content={instance.PublicIpAddress}
      />
    );
  }
  return result;
}, [instance]);
```

extensions/amazon-aws/src/ec2.tsx
```diff
@@ -91,8 +107,7 @@ function InstanceListItem(props: { instance: AWS.EC2.Instance }) {
               instance.InstanceId
             }
           />
-          <CopyToClipboardAction title="Copy Public IP" content={instance.PublicIpAddress || ""} />
-          <CopyToClipboardAction title="Copy Private IP" content={instance.PrivateIpAddress || ""} />
+          {ipActions}
```
gebeto, line 110: here you can use mechanism like this, instead of using array:

```ts
<CopyToClipboardAction key="privateIpAction" title="Copy Private IP" content={instance.PrivateIpAddress || ""} />
{instance.PublicIpAddress && <CopyToClipboardAction key="publicIpAction" title="Copy Public IP" content={instance.PublicIpAddress} />}
```

duboiss: Thanks a lot for your tip

gebeto (approved):
Looks awesome!

pernielsentikaer (approved):
🎉

PR #1720, opened 2022-05-18
feat: Yet another encode-decode tool with only few steps

DarthPestilane:
## Description

Encode or decode input text in base64/url and copy to clipboard with **only few steps**

## Screencast

image
image

## Checklist


koinzhang:
Hi @DarthPestilane 
It seems that several extensions with the same functionality already exist in the store.
[URL Tools](https://www.raycast.com/joshua46/url-tools)
[Base64](https://www.raycast.com/DanielSinclair/base64)
[Converter](https://www.raycast.com/koinzhang/converter)
[ToolBox](https://www.raycast.com/Kang/toolbox)

DarthPestilane:
Hi @koinzhang 
I have searched the store for the function I want, and I found some good extensions. But they are all quite complex, I have to hit the keyboard many times to get the result.

I used to use Alfred workflow, which is fast and with few steps to get the result. Raycast is a good but new tool, I'd love to use it and make convenient extensions compare with Aflred.

And is there any further plan to support the third-party store to install extensions? Since developers may publish and share their self-used extensions

Thanks.

pernielsentikaer:
Hi @DarthPestilane 👋 

As already mentioned earlier in this thread, extensions are already doing this; instead of adding one more, it would be better to open a feature request or create a PR for an existing extension with a different command.

I hope that the above makes sense and that you'll get my point. Otherwise, we end up with too many tools doing the same work.

Feel free to contact me here or at [Slack](https://www.raycast.com/community) or Twitter if you have any questions.

pernielsentikaer:
I'm closing this due to inactivity; feel free to open it again or comment in the thread when you're ready to continue working on it 🙂

You can also catch me here or in [Slack](https://www.raycast.com/community) if you want to discuss 😄

PR #1729, opened 2022-05-18
feat(doc): improve contribution documentation page

duboiss:
## Description
- Replace "clone" by "fork" because we can't push to the repository
- Specify `npx ray lint` in order to be able to run eslint/prettier
By the way, I have the impression that prettier run with this command does not behave normally. I find it disturbing that depending on the extension it is specified in the package.json or not.


mathieudutour (approved):

duboiss:
@mathieudutour It was precisely because the CI was failing that I searched for many minutes to launch prettier (aws extension). I had to install prettier globally and find the fix command. So the DX was very bad :(

PR #1741, opened 2022-05-19
Todoist improvements

gebeto:
### Add colors to create project select
Metadata

### Add priority icons
Metadata-2

## Checklist

### DIFF
--- a/extensions/todoist/package-lock.json
+++ b/extensions/todoist/package-lock.json
@@ -27,9 +27,9 @@
       }
     },
     "node_modules/@babel/runtime": {
-      "version": "7.17.0",
-      "resolved": "https://registry.npmjs.org/@babel/runtime/-/runtime-7.17.0.tgz",
-      "integrity": "sha512-etcO/ohMNaNA2UBdaXBBSX/3aEzFMRrVfaPv8Ptc0k+cWpWW0QFiGZ2XnVqQZI1Cf734LbPGmqBKWESfW4x/dQ==",
+      "version": "7.17.9",
+      "resolved": "https://registry.npmjs.org/@babel/runtime/-/runtime-7.17.9.tgz",
+      "integrity": "sha512-lSiBBvodq29uShpWGNbgFdKYNiFDo5/HIYsaCEY9ff4sb10x9jizo2+pRrSyF4jKZCXqgzuqBOQKbUm90gQwJg==",
       "dependencies": {
         "regenerator-runtime": "^0.13.4"
       },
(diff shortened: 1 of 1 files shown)


thomaslombart (approved):
Looks good to me except for the `package-lock.json` file that has been modified even though there are no changes in `package.json`.

gebeto:
@thomaslombart there was some minor versions updates, it is because package.json was not updated
Screenshot 2022-05-19 at 15 44 12

After update it looks so:
Screenshot 2022-05-19 at 15 49 33

gebeto:
There was no changes in package.json

thomaslombart:
Then `package-lock.json` shouldn't be part of the PR. As stated in [npm Docs](https://docs.npmjs.com/cli/v8/configuring-npm/package-lock-json):

> package-lock.json is automatically generated for any operations where npm modifies either the node_modules tree, or package.json. It describes the exact tree that was generated, such that subsequent installs are able to generate identical trees, regardless of intermediate dependency updates.

Could you remove the `package-lock.json` file from the changed files of the PR? 🙂

gebeto:
Yep, but do we really need this? I think it will be better to leave it here, because it will guarantee that npm installed the same packages versions on your, my or any other machines

And all the packages updates will be explicit

Dependencies in the package.json declared with ^, for example eslint: ^8.4.1, it means that this project require eslint version 8, but it can include bigger minor or patch versions, but not major. But package-lock.json save the concrete version of the package that should be installed, to have the guaranty that it is the same everywhere

You can read more about that why package.json was not updated here: https://docs.npmjs.com/about-semantic-versioning#using-semantic-versioning-to-specify-update-types-your-package-can-accept

thomaslombart:
Ok, I think there's a misunderstanding here.

- `package-lock.json` as you and I said is here to guarantee that npm installs the exact same package versions on different machines. It's been committed since the first release of the extension.
- `package-lock.json` is updated (as stated in npm docs) where npm modifies either the `node_modules` tree, or `package.json`. However, this PR doesn't contain any updates to the `package.json` file. Thus, the `package-lock.json` file should stay as-is.
- The few cases where `package-lock.json` file can get modified without `package.json` changes are for example when you delete the `package-lock.json` file and run `npm install` afterward. This would cause npm to pick up the latest dependencies depending on the versioning (e.g `"eslint": "^8.4.1",` would cause npm to install the latest minor version which is 8.15.0 as of now). But doing so implies that others won't have the same package versions unless they do an `npm install` after pulling the changes.

In every case, I don't think updating it will break on others machines. But the thing to get here is that if the `package.json` is not modified, the `package-lock.json` file shouldn't be modified as well 🙂

gebeto:
> * The few cases where `package-lock.json` file can get modified without `package.json` changes are for example when you delete the `package-lock.json` file and run `npm install` afterward

Also you can use `npm upgrade` to update outdated packages, I've ran this to update all the packages. 


> * But doing so implies that others won't have the same package versions unless they do an `npm install` after pulling the changes

Developers should keep track on this, and it is their problem that they are not updating dependencies after pulling new changes

It is okay to update lock file without updating `package.json`, also when Raycast CI building the package, it is using package-lock, so there is guaranties that CI build it successfully because it has the same packages versions that developer has on his machine, and build is not depends on the time when package is building(because there can be some time between creating PR and merging it and some packages can be updated, so Raycast CI build will be different than your old local build)

thomaslombart:
Ok, I won't debate furthermore here. Let's leave it up to the maintainers.

pernielsentikaer:
The lock file can be updated without `package.json - when you run `npm update` to update all dependencies to the newest versions.

I would say it’s ok to make an update like this, especially if there were some vulnerabilities in the older version, for example - so there is nothing wrong in updating purely `package-lock.json` without touching the `package.json itself.

I checked the current store version with `npm audit` it had one high severity:

```
follow-redirects  <=1.14.7
Severity: high
Exposure of Sensitive Information to an Unauthorized Actor in follow-redirects - https://github.com/advisories/GHSA-pw2r-vq6v-hr8c
Exposure of sensitive information in follow-redirects - https://github.com/advisories/GHSA-74fj-2j2h-c42q
fix available via `npm audit fix`
node_modules/follow-redirects

One high severity vulnerability
```

and it will be fixed with `npm update` - so in that case - I would say it’s worth including. Otherwise, it makes no sense 😄 

Let's get it merged 🔥

Issue #1479, opened 2022-04-23
[Spotify Player] Connect to a Device


timcastillogill:
# Extension – [Spotify Player](https://github.com/raycast/extensions/tree/317b7f140dfc4a304317e5bde1408221e0777e30/extensions/spotify-player/)
Author: @mattisssa

**Describe the feature and the current behavior/state.**

I would love to be able to choose a connected device - that is the online speakers available on your wifi network to connect to via the `connect to a device` menu. The button to do this via Spotify is located right next to the Spotify Volume rocker.

**Who will benefit with this feature?**

Anyone who uses a connected/smart speaker to listen to music via Spotify.

**Any Other info.**

It looks as if this API endpoint in in Beta but details can be found [HERE](https://developer.spotify.com/documentation/web-api/guides/using-connect-web-api/)

mattisssa:
Hey @andreaselia and @timcastillogill! I just have reached the issue and realised that it wouldn't be possible to implement once my #1766 PR merged because the developer app on the Spotify side doesn't have a proper scope for the feature 🙈 
And their review team is pretty weird about updating these developer apps; it shows to me that I cannot request new scopes for a while.

andreaselia:
Awh, that's unfortunate. I wonder if contacting them, they'd make an exception to the delay between updating the scopes of the app? 😅 

Alternatively, I guess anyone can create one of these developer apps for approval and PR an update to the client ID/scope, alongside this feature, right? Or would you want to avoid anyone having that sort of control over the Spotify Player extensions functionality? 🤔

mattisssa:
>I wonder if contacting them, they'd make an exception to the delay between updating the scopes of the app

I have tried to contact them and ask about it, but they just keep ignoring me, and there was no response from the Spotify team 🤦 

>Alternatively, I guess anyone can create one of these developer apps for approval and PR an update to the client ID/scope, alongside this feature, right

Yes, but it ruins all the advantages of installing the extension and don't worry about anything. Before the update, you had to go to the Spotify developer page, create developer app, update secrets in the Raycast Spotify extension, it's pretty hard.

>Or would you want to avoid anyone having that sort of control over the Spotify Player extensions functionality

If I correctly understand the question, is it about the Spotify's Devices control? Or having another Spotiy extension in the store?

andreaselia:
Ah I see, that sucks that they aren't being responsive 😬 

[...]

PR #1766, opened 2022-05-22
Spotify Player: Introduce OAuth and several useful commands

mattisssa:
Finally! 🎉  The Spotify review team has reviewed my app on their side.

## Description

In the new version, you don't need to create a developer app, generate tokens and bla.., but what is more astonishing is that it will support a few more valuable things like: Lik Current Song, Search Artists -> Search Albums -> Play Albums entirely, and many more.

## Checklist

### DIFF
--- a/extensions/spotify-player/CHANGELOG.md
+++ b/extensions/spotify-player/CHANGELOG.md
@@ -1,8 +1,15 @@
 # Spotify Player Changelog
 
+## [Feature] - 2022-05-24
+
+- Added OAuth Support
+- Introduce new command "Like Curent Song"
+- Introduce new command "Search Artists"
+- Redesign for existing commands, add feature to preview albums
+
 ## [Feature] - 2022-05-23
 
-- Add the ability to open via the Spotify app instead of always using Spotify Web for everything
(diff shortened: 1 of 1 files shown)


andreaselia:
This is great! I was going to take a look at resolving #1479, but with the work you've done here, it will be even easier to implement.

pernielsentikaer (changes requested):
Hi @mattisssa 👋 

What an update, really looking forward to using it! 

While testing I found that the `Like Current Song` is throwing an error instead of gracefully showing it, it happened in two scenarios:

- Uninstalled Spotify and used the command
- Installed Spotify and didn't have a song selected yet (This should not be happening often)

The error I received was this:
```
Error: type mismatch at params.title : Expected to decode String but found a dictionary instead.
```

The rest of the command popped a OSX window asking me to select Spotify. Just after installing Spotify again it all (Except that I didn't have a song played yet) just worked again.

It's a tough one to find all edge cases so there might be things and combinations that we might see later on.

mattisssa:
Hey @pernielsentikaer, thanks for the good catches! I've updated the PR, please take a look at the extension again when you have time 👏

PR #1768, opened 2022-05-22
[Extension] Add Yandex Music extension

AlexStrNik:


AlexStrNik:
@koinzhang Hi! Can you look please?

koinzhang:
Hi @AlexStrNik. I am very sorry, I am not a Raycast staff member.
@pernielsentikaer is in charge of extension review and hope he can review your extension soon.

mattisssa (changes requested):
Hey! Thanks for the contribution, but unfortunately, I couldn't run any commands from the extension. I got the issue:
CleanShot 2022-05-23 at 20 45 46@2x

I'm not sure what could be wrong here, I tried to configure Safari or Chrome in the extension preferences and both browsers crashed with the issue from the screenshot above.

Let me know if you need some additional info.

AlexStrNik:
Okaaay. The problem is that Raycast app has no permissions to control either Safari or Chrome. But it should ask permission to do it for first time of usage. No idea why it isn't working like this with `runAppleScript`

AlexStrNik:
@mattisssa  can you look once again please? My assumption about permissions was wrong. The problem was in two things: Safari has a different syntax for executing JS (wtf Apple?) and in some cases option that allows executing of javascript in browser may turned off, now in that case extension will show a HUD with message

extensions/yandex-music/src/utils.ts
```diff
@@ -0,0 +1,41 @@
...
+                end repeat
+            end tell
+        `);
+  } catch (e) {
+    const message = (e as OsaError).stderr;
+
+    if (message.includes("Allow JavaScript from Apple Events")) {
+      showHUD(
+        `You must enable the 'Allow JavaScript from Apple Events' option in ${browser.name}'s Develop menu to use extension`
+      );
```
mattisssa, line 38: I think it would be better to use a simple toast for long messages like this, once I got the issue I couldn't see the text because HUD disappeared straight away 🙈.

```suggestion
      showToast(
        Toast.Style.Failure,
        `You must enable the 'Allow JavaScript from Apple Events' option in ${browser.name}'s Develop menu to use extension`
      );
```

And to make this work more proper I think it could be nice to return in the function some result _(can be a simple boolean)_, and depending on that result call the `await closeMainWindow();` from outside _(now you're doing this before trigger the applescript)_

extensions/yandex-music/src/utils.ts
```diff
@@ -0,0 +1,41 @@
+import { runAppleScript } from "run-applescript";
+import { Application, getPreferenceValues, showHUD } from "@raycast/api";
+
+// May work in other browser too, but it's confirmed to work only in these
+type SupportedBrowsers = "Safari" | "Chrome" | "Yandex";
```
mattisssa, line 5: I haven't seen any checks if a selected browser is different from this. Shall we show some error toast for the case user configures an unsupported browser?

extensions/yandex-music/src/utils.ts
```diff
@@ -0,0 +1,41 @@
...
+}
+
+export async function runJSInYandexMusicTab(code: string) {
+  const browser = getPreferenceValues<{ browser: Application }>().browser;
+
+  try {
+    await runAppleScript(`
+            tell application "${browser.name}"
+                repeat with w in (every window)		
+                    repeat with t in (every tab whose URL contains "music.yandex.ru") of w
```
mattisssa, line 26: I'm not sure if it's possible here, but I think it would be nice if the `music.yandex.ru` is opened in a selected browser. Otherwise, I'd show another toast with a description like "The Yandex Music page hasn't been found"

mattisssa (changes requested):
Hey, thanks for fixing the issues 👍 

It would be really nice to add a few more toasts to keep a better UX for end-users. Everything works fine to me in my browsers, but I commented on how we could make the extension better.

extensions/yandex-music/src/utils.ts
AlexStrNik: My fault. I meant that extension can work with almost every browser but i checked just these. Because actually any Chromium browser should be supported and Firefox based browsers too. The idea was just to list browsers for which it's known that extension work, not to restrict usage of extension with this list. I will correct the comment

PR #484, opened 2021-12-06
Add Google Maps Search extension

ratoru:
## Description

Add `google-maps-search` extension that allows you to quickly pull up Google Maps directions in your browser. After setting home address and preferred mode of travel in preferences, this extension provides the following 3 commands:

- `Travel Home` - open Google Maps with directions from your current location to your home using the preferred method of transportation.
- `Travel To` - open Google Maps with directions from your current location, your home, or a custom address to a specified address.
- `Find Place` - Open Google Maps with the given search parameter.

This extension should be most useful for people who commute via public transport and want to check when the next train to their home leaves. (--> This can be done with one "no-view" command.)

## Type of change

- New extension

## Screencast

https://user-images.githubusercontent.com/44789473/144770767-df4aba6b-e78e-4e98-ad7f-d4bc77f12d83.mp4

## Checklist

- [x] I read the [extension guidelines](https://developers.raycast.com/basics/prepare-an-extension-for-store)
- [x] I read the [documentation about publishing](https://developers.raycast.com/basics/publish-an-extension)
- [x] I also made sure that the `png` is `512x512`

PR #1774, opened 2022-05-22
Add Apple Maps Search extension

djpowers:
## Description

This adds a new "Apple Maps Search" extension, which is a fork of the ["Google Maps Search" extension](https://github.com/raycast/extensions/tree/main/extensions/google-maps-search) by @ratoru added in #484. This offers the same functionality and commands, but using Apple Maps as the data source instead:

- `Travel Home` - open Apple Maps with directions from your current location to your home using the preferred method of transportation.
- `Travel To` - open Apple Maps with directions from your current location, your home, or a custom address to a specified address.
- `Find Place` - Open Apple Maps with the given search parameter.

I had considered changing some copy and to better match the Apple Maps terminology (e.g., "Search Maps" command instead of Find Place, using "Transport Type" instead of travel mode), but I thought it would be a better user experience to match the same command interface and UI copy as the original extension where possible. Open to suggestions if others have thoughts here.

## Screencast

https://user-images.githubusercontent.com/4978418/169706545-48800e6c-2b6e-455b-9c8c-8b863a2a1fd0.mp4

## Checklist

### DIFF
--- a/extensions/apple-maps-search/src/utils.ts
+++ b/extensions/apple-maps-search/src/utils.ts
@@ -1,7 +1,7 @@
 /**
  * Corresponds to Apple Map's four possible modes of travel.
  */
- export enum TravelMode {
+export enum TravelMode {
   Driving = "d",
   Walking = "w",
   Transit = "r",


ratoru:
Hi Dave! 
This looks great! If you feel like the command names should be changed in order to improve user experience, I might also be able to change them for the Google Maps Search extension.

pernielsentikaer (commented):
Hi @djpowers 👋 

Excellent with a port 🔥 

It works as expected, and I do have one question:

- Is there no way to get `Travel Home` to not open a browser before opening maps? The two other commands don't do that 😄 

I'm not that into maps, so I couldn't verify myself 😄

djpowers:
Thank you both!

@ratoru, I think the Google Maps commands are great as is. I only had the thought to change the Apple Maps commands since Apples refers to those actions slightly differently, but I think the current commands are clear as to the functionality so I'm fine matching the original.

@pernielsentikaer, I noticed this as well. It's using the same exact `http://maps.apple.com` base URL as the `Travel To` command, so my guess is there's something to do with that command being invoked with `exec` as opposed to using `OpenInBrowserAction` from the Raycast API.

I had initially planned to use one of the [Apple Maps URL Schemes](https://itnext.io/apple-maps-url-schemes-e1d3ac7340af), but thought using the URL from the docs might be more robust, especially if someone did not have Apple Maps installed.

Using the URL scheme opens directly in the app (possibly even a little quicker), so that seems like a good solution for now. I'm not sure if either option handles things differently if Maps is not installed or not available, but a future update could possibly check for that and fallback as needed.

ratoru:
Sounds good @djpowers!

I believe that there was no `OpenInBrowserAction` when I wrote Google Maps Search. So, don't rely on what I did as the golden standard. Seems like I should replace the `exec` call with the official API call.

pernielsentikaer (approved):
Awesome, great work @djpowers 🔥 

I updated the code to newest Raycast API 💪

PR #1781, opened 2022-05-23
New extension: QuickFile

danielroe:
## Description

[QuickFile](https://www.quickfile.co.uk/) is highly configurable cloud accounting software that is free for small to medium sized accounts with excellent support and integration for the UK tax (including VAT, Making Tax Digital and Companies House).

## Screencast

CleanShot 2022-05-23 at 13 57 58

## Checklist

### DIFF
--- a/extensions/quickfile/CHANGELOG.md
+++ b/extensions/quickfile/CHANGELOG.md
@@ -1,6 +1,6 @@
 # quickfile Changelog
 
-## [Initial Version] - {PR_MERGE_DATE}
+## [Initial Version] - 2022-06-01
 
 - Add `View Account Balances`, including warning when Open Banking consent needs to be refreshed.
 - Add `Account Details` view to see up to 20 individual transactions for each account.


extensions/quickfile/src/balances.tsx
```diff
@@ -0,0 +1,94 @@
...
+                  }
+                  title={account.Name}
+                  accessories={[{ text: balance }]}
+                  actions={
+                    <ActionPanel>
+                      {!!details && (
+                        <Action.Push
+                          title="Account Details"
+                          target={<AccountDetail transactions={details} url={url} />}
+                        />
```
pernielsentikaer, line 82: Misses an icon since `Action.OpenInBrowser` has, if one item has an icon then all should have 

```suggestion
                        <Action.Push
                          title="Account Details"
                          icon="Icon.List"
                          target={<AccountDetail transactions={details} url={url} />}
                        />
```

extensions/quickfile/src/balances.tsx
```diff
@@ -0,0 +1,94 @@
...
+  const { data: transactions = [], loading: loadingTransactions } = useAsyncData(bankCodes, (codes) =>
+    Promise.all(codes.map((code) => bankGetAccountTransactions(code, { number: 20 })))
+  );
+
+  const accountTypes = unique(accounts.map((a) => a.BankType)).sort(
+    (a, b) => (bankTypePriority[b] || 0) - (bankTypePriority[a] || 0)
+  );
+
+  return (
+    <List isLoading={loadingDomain || loadingAccounts || loadingTransactions}>
```
pernielsentikaer, line 55: Could you set the `isLoading` to `false` when the error occurs, this way we see the `No Results` screen instead.

CleanShot 2022-05-24 at 05 09 56@2x

extensions/quickfile/src/balances.tsx
```diff
@@ -0,0 +1,94 @@
...
+      {accountTypes.map((type) => (
```
pernielsentikaer, line 56: You could add a [List.EmptyView](https://developers.raycast.com/api-reference/user-interface/list#list.emptyview) here to make the UX even better, something like this:

```suggestion
      <List.EmptyView icon={{ source: "icon.png" }} title="No Results" />
      {accountTypes.map((type) => (
```

You could make a copy of the current logo and resize it to 64x64 pixels and it will look something like this:

CleanShot 2022-05-24 at 05 37 50@2x

If you also make some conditional text, if an error accours then you could change the text to something else 😄

extensions/quickfile/src/views/AccountDetail.tsx
```diff
@@ -0,0 +1,33 @@
...
+  url?: string;
+}
+
+export default (props: AccountDetailProps) => {
+  return (
+    <List isLoading={!props.url}>
+      {props.transactions.Transactions.Transaction.map((transaction) => (
+        <List.Item
+          title={transaction.Reference}
+          icon={
```
pernielsentikaer, line 16: All items here should have an icon to avoid the text jumping around.

pernielsentikaer (changes requested):
Hi @danielroe 👋 

Thanks for your contribution, it looks good 💪 

I have some feedback and suggestions for you to address:

Your metadaimages should only contain either light or dark theme, and should not have toltips sticking out 😄 I would show this:

- quickfile-screenshot-1.png
- quickfile-screenshot-1.png with ActionPanel open
- quickfile-screenshot-2.png without tooltip

extensions/quickfile/src/utils/hooks.ts
```diff
@@ -0,0 +1,46 @@
...
+) => {
+  const [data, setData] = useState<T>();
+  const [loading, setLoading] = useState(true);
+  const [error, setError] = useState<any>();
+
+  const key: string = Array.isArray(_key) ? _key.join(",") : _key;
+  useEffect(() => {
+    if (!key) return;
+
+    Promise.race([
```
sxn, line 16: would it make sense to use [Promise.any](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/any) here instead? In case the first promise (`LocalStorage.getItem`) rejects due to e.g. `JSON.parse` throwing an exception, we'd probably want to wait for the API call to resolve, only _then_ resolving the parent promise.

danielroe: Agreed 👍 Nice catch.

extensions/quickfile/src/utils/hooks.ts
```diff
@@ -0,0 +1,46 @@
...
+      LocalStorage.getItem(key).then((storedValue: unknown) => {
+        if (!storedValue || typeof storedValue !== "string") return;
+        return JSON.parse(storedValue) as T;
+      }),
+      fetcher(_key)
```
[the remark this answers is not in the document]
danielroe: We should only make one request per unique key ... Or are you thinking if the user invokes 'View Account Balances' several times in succession?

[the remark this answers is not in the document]
sxn: > Or are you thinking if the user invokes 'View Account Balances' several times in succession?
that's one example, yeah 😸

But since this is a general util, it could be used in other places where multiple requests are plausible. Commonly this happens if `List.Item.Detail` dispatches an API call on mount; quickly moving selection up/down would end up sending a lot of API calls, despite the fact that once selection has changed, the result of the API call can be discarded.

But anyway, feel free to ignore my suggestion; we can always come back to it in the future if it somehow becomes necessary. 😉

pernielsentikaer (approved):
Hi @danielroe 👋 

Thanks for your contribution, let's get it merged 🔥

PR #1784, opened 2022-05-23
New Extension: Nature Sounds 🔥

jsheffers:
## Description
This extension brings the sounds of nature to Raycast.  You can mix and match six different sounds from nature, or pick your favorite. 

## Screencast
nature-sounds-01

## Checklist

### DIFF
--- a/extensions/nature-sounds/src/Actions.tsx
+++ b/extensions/nature-sounds/src/Actions.tsx
@@ -20,35 +20,39 @@ export default function Actions(props: { item: Audio; playing: Item[]; setPlayin
           />
         )}
 
-        {props.playing && !props.playing.some((item: Item) => item.name === props.item.filename) && (
-          <Action
-            title="Play"
-            onAction={() => {
-              const played = exec(`afplay ${__dirname}/assets/${props.item.filename}.mp3 && $$`);
-              const tempState: Item[] = [...props.playing, { name: props.item.filename, pid: played.pid! }];
-              props.setPlaying(tempState);
-              showToast({
-                style: Toast.Style.Success,
-                title: `Now playing — ${props.item.icon} ${props.item.name}`,
-              });
-            }}
-          />
-        )}
+        {props.playing &&
+          props.item.filename !== "stop" &&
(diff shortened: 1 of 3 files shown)


pernielsentikaer (changes requested):
Hi @jsheffers  👋 

Thanks for your contribution!

The assets are a bit huge, is it possible to trim/compress the files a bit more?

jsheffers:
I was able to compress a few (change is pushed), but I think these are as small as I can get them now.

extensions/nature-sounds/src/Actions.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+export default function Actions(props: { item: Audio; playing: Item[]; setPlaying: PlayingItem["setPlaying"] }) {
+  return (
+    <ActionPanel title={props.item.name}>
+      <ActionPanel.Section>
+        {props.item.filename === "stop" && (
+          <Action
+            title="Stop All Sounds"
+            onAction={() => {
+              exec(`killall afplay`);
```
pernielsentikaer, line 13: It could be better to loop the array of playing items and killing them individually since we don't know if the user plays anything else with `afplay`

pernielsentikaer (changes requested):
Hi @jsheffers 👋 

I think the assets are a bit better now, later on, we can discuss a way to download them on the fly instead, but for now, I think we're good to go 😄 

I changed a small thing in the code since I was able to play "Stop All Sounds" 💪 

I left a single comment. Is that something that you can look into as the last thing?

pernielsentikaer:
Hi @jsheffers 👋 

Are you still around? I tested this a few times and since we don't have the possibility to run this in the background it just stops the sounds after playing them one time, is there an application we can hook into instead? Otherwise it might be a bit silly when it stops after playing it once

extensions/nature-sounds/src/Actions.tsx
pernielsentikaer:
```suggestion
              props.playing.forEach((item) => {
                exec(`kill -9 ${item.pid + 1}`);
              });
```
And this could work here

PR #1836, opened 2022-05-28
jetbrains: improve keywords to match dashes and low-dashes when searching with space

adrienbaron:
## Description

Improve the keywords generated for each jetbrains projects in the list so that they match when searching with space.
For example a project called `cx-api-python` will now show up when typing `cx api` in the search.

## Screencast

**Before:**

https://user-images.githubusercontent.com/4410247/170826915-7aed3b25-ee3e-4d5d-ac8b-05568a18650a.mov

**After:**

https://user-images.githubusercontent.com/4410247/170826920-23932949-60d1-4817-b6d2-d623987576a8.mov

## Checklist

### DIFF
--- a/extensions/jetbrains/CHANGELOG.md
+++ b/extensions/jetbrains/CHANGELOG.md
@@ -1,0 +1,5 @@
+# Jetbrains Changelog
+
+ ## [Update] - 2022-06-11
+
+Improve keywords to match dashes and low-dashes when searching


gdsmith:
@adrienbaron not sure what the Raycast team thinks here, but the filtering here is the same as how it works elsewhere in Raycast so I'm not sure it's something we should make custom here. You're probably already aware, but you can do the search without spaces (e.g. `cxapi`) and get your results.

adrienbaron:
@gdsmith good shout! I didn't know the version without space worked 🤦‍♂️.

I use the Github search extension too and it does work with spaces (`cx api`), and doesn't without (`cxapi`). I think this is due to it using Github search API under the hood and the disconnect between the two is throwing me off 😅.

I think in my mental model `-` and `_` are here to represent spaces in strings that don't support them, so they should be considered word separators IMO.
This means searching with ` ` (which is easier to type) should match them.

However this might be something to change in Raycast default search matching algorithm rather than manually here? Wonder what the Raycast team thinks about that 🤔.

gdsmith:
@pernielsentikaer if you are happy, I'm happy

pernielsentikaer (approved):
Hi @adrienbaron 👋 

Works as expected, let's get it merged 🚢 

Would be nice to have some screenshots added soon, anyone want to do that?

gdsmith:
@pernielsentikaer screenshots added in #1686

PR #1847, opened 2022-05-30
iMessage search extension

timkendall:
## Description

Adds a very basic (but useful!) extension to search iMessages. It is implemented using the approach taken by the [safari extension](https://github.com/raycast/extensions/tree/main/extensions/safari) in that it uses the WebAssembly version of [sql.js](https://github.com/sql-js/sql.js/) to load the SQLite databases of both the iMessage and Contacts apps into memory. 

We could alternatively use native bindings to SQLite (such as [sqlite3](https://www.npmjs.com/package/sqlite3) or [prisma](https://github.com/prisma/prisma)) to support standard paging from the fs however this is likely overkill for this simple use-case (tested against my iMessage db that has ~200k messages in it and clocks in at around **160mb** and my Contact's db with 339 contacts and is **~6mb**). The sql.js approach has the benefit of not requiring unique Node native module binaries per ARM/x86 as well.

**Some improvements for future contributors:**
- Improve contact lookup by searching against normalized phone numbers (Contact's `ZABCDPHONENUMBER.ZFULLNUMBER` is not normalized it seems)
- Improve contact lookup by searching against email addresses (iMessage's `handle.id` can be a phone number, email address, or `null` it seems)
- Render Contact image avatars (it seems like the data exists in Contact's `ZABCDRECORD.ZTHUMBNAILIMAGEDATA` however I was not immediately able to determine the binary format)
- Avoid loading all data into memory (perhaps via. a WebAssembly-targeted version of Prisma client) 
- Use a first-class iMessage and Contacts API instead of reading the apps internal data :)

## Screenshot

imessage-search-0

## Checklist

### DIFF
--- a/extensions/imessage-search/CHANGELOG.md
+++ b/extensions/imessage-search/CHANGELOG.md
@@ -1,3 +1,3 @@
 # iMessage Search Changelog
 
-## [Initial Version] - 2022-05-29
+## [Initial Version] - {PR_MERGE_DATE}


pernielsentikaer (changes requested):
Hi @timkendall 👋 

Just a question for now:
- Is there no other way to achieve this so we could avoid asking for full disk access?

pernielsentikaer:
Could it be an idea to build #1954 into this and make it a single extension instead? 

Is there a way to show conversations by default instead of no results?

timkendall:
Hey @pernielsentikaer thanks for reviewing! 

>Is there no other way to achieve this so we could avoid asking for full disk access?

Non that I could find unfortunately. If Apple ever offered a public Swift or REST API for reading personal iMessages we could potentially use that.

>Could it be an idea to build https://github.com/raycast/extensions/pull/1954 into this and make it a single extension instead?

I believe this implementation would support https://github.com/raycast/extensions/pull/1954 's use-case already.

>Is there a way to show conversations by default instead of no results?

Sure I can make this change.

pernielsentikaer:
Hi @BalliAsghar 👋 

Sorry for the late response, your PR #1954 was marked as resolved on my personal list.

Could it make sense to make one big extension with both this and yours?

pernielsentikaer:
I'm closing this due to inactivity; feel free to open it again or comment in the thread when you're ready to continue working on it 🙂

You can also catch me in [Slack](https://www.raycast.com/community) if you want to discuss 😄

Issue #1855, opened 2022-05-30
[Google Drive] ... 


Anshuljkt:
# Extension – [Google Drive](https://github.com/raycast/extensions/tree/1655e5205d460085b88656682dd9ce9f099c928c/extensions/google-drive/)
Author: @vishaltelangre

## Description
Google Drive folder search broken, always crashes when attempting to access search results.

## Error:
```
Error: ENOENT: no such file or directory, stat  '....REDACTED..../Insurance May 2022-1.jpg'

statSync:node:fs:1536:3
Nl:index.js:6:10098
processTicksAndRejections:task_queues:96:5
```

## Steps To Reproduce

1. Search for a file, attempt to look through results...
2. ends up broken

### Raycast version
Version: 1.35.2

vishaltelangre:
@Anshuljkt Thanks a lot for the report. But unfortunately, I am unable to reproduce this error. If you are running this extension from the [source](https://github.com/raycast/extensions/tree/1655e5205d460085b88656682dd9ce9f099c928c/extensions/google-drive/), can you please post the exact backtrace from the terminal since it is not clear where exactly it is crashing from the error above? BTW, if you're on Slack, ping me so we can debug + fix it further. Thanks!

PR #1857, opened 2022-05-30
[Google Drive] Fix: Prevent retrieving stats of a non-existing file path

vishaltelangre:
## Description

Fixes https://github.com/raycast/extensions/issues/1855.

### DIFF
--- a/extensions/google-drive/CHANGELOG.md
+++ b/extensions/google-drive/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Google Drive Changelog
 
-## [Fix retrieving stats of a non-existing files] - {PR_MERGE_DATE}
+## [Fix retrieving stats of a non-existing files] - 2022-05-31
 
 - Fixed [an issue](https://github.com/raycast/extensions/issues/1855) due to which the extension was crashing while retrieving the stats of a file that doesn't exists anymore.
 - When the cache is reindexed (automatically after it gets invalidated at the configured interval or forcefully via the Actions menu), now it would overwrite all existing indexed files (by retaining the favorites) so as to avoid having non-existing file paths in the database.


extensions/google-drive/src/db.ts
```diff
@@ -243,17 +243,14 @@ export const indexFiles = async (db: Database, options: IndexFilesOptions = { fo
...
-    if (options.force) {
-      // Backup the favorite file paths before force indexing
-      favoriteFilePaths = queryFavoriteFiles(db, 1000).map((file) => file.path);
+    // Backup the favorite file paths before indexing
+    const favoriteFilePaths = queryFavoriteFiles(db, 1000).map((file) => file.path);
 
-      // Delete all the old indexed files
-      db.exec("DELETE from files");
+    // Delete all the old indexed files
+    db.exec("DELETE from files");
```
vishaltelangre, line 251: So that there's no difference between forced indexing and automatic indexing (on cache expiry). This will ensure that the no longer existing paths will be nuked off from the database (due to which the mentioned bug seemed to have occurred).

extensions/google-drive/src/utils.ts
```diff
@@ -132,20 +132,22 @@ export const initialSetup = () => {
 };
 
 export const filePreview = async (file: FileInfo | null, controller: AbortController): Promise<string> => {
-  if (!file) {
-    return "";
+  const previewImage = (path: string) => `<img src="file://${path}" width="192" height="192" />`;
+
+  if (!file || !pathExists(file.path)) {
```
vishaltelangre, line 137: Checks if the path exists before doing anything further.

vishaltelangre:
@pernielsentikaer Please review.

PR #1869, opened 2022-06-02
Modules Page

yug2005:
Added modules page, error message through toasts, more icons, and fixed bug with useEffect

## Description

When you click on the courses, there is a new modules page that shows. This page includes items of the following type : 
* Page
* File
* ExternalUrl
* Assignment
* Quiz

If a file ends with an appropriate prefix, the icon for that file is a code icon. 

Other changes include: 
* New icons on home window.
* Subheaders are removed. 
* Error is shown using showToast
* useEffect is now fixed to only run once. 

## Screencast

homepage
modulepage

## Checklist

### DIFF
--- a/extensions/canvascast/CHANGELOG.md
+++ b/extensions/canvascast/CHANGELOG.md
@@ -5,4 +5,3 @@
 - Added modules page
 - Better error handling
 - Added icons
-- Updated API


extensions/canvascast/src/index.tsx
```diff
@@ -100,47 +148,52 @@ export default function main() {
...
         {items?.map((item, index) => (
           <List.Item
             key={index}
             title={item.name}
-            subtitle={item.code}
-            icon={{ source: Icon.Pin, tintColor: Color[Colors[index % Colors.length]] }}
+            icon={{ source: Icons["Course"], tintColor: Color[Colors[index % Colors.length]] }}
             actions={
-              <ActionPanel title="Title">
-                <OpenInBrowserAction url={"https://" + preferences.domain + "/courses/" + item.id} />
```
garyhtou, line 131: I love this new update. Thanks for making the extension better 😄!

I do have some small feedback: It was really convenient to be able to easily visit the home page of courses. It seems like this functionality was replaced with the modules (which is totally fine!). Would it be possible to keep the link to course home pages?

YodaLightsabr:
@garyhtou @yug2005 Since you seem to be active users of this extension, do you think it would be a good idea to add more features and API coverage, or just act as a shortcut to access the website? I'm considering adding pages and assignments support to the extension itself, but I'm not sure if people would want it / use it.

garyhtou:
Personally, I think the extension should serve two purposes:

- **Short cuts**. Make it easy to access commonly used pages in each course.
- **Add additional read-only functionality on top of what Canvas already provides**. These could be new views that aggregate data in a friendlier way. For example, a list of upcoming assignments/quizzes from all courses, ordered by due date. afaik, Canvas doesn't provide a good aggregated view of upcoming due dates - besides the calendar, which in my opinion is pretty messy. I don't think the extension should modify/write any data via the API. Read-only feels much safer, especially since this extension is already against their TOS.

YodaLightsabr:
Got it, that makes sense.

PR #1906, opened 2022-06-06
Fast Dash docset filtering from MultiDocsetSearch

dannydb:
## Description

When the first word of the search string matches one of the docset keywords, we can append a `:` to the keyword and immediately filter the results down to just that docset, similar to the experience with Dash’s Alfred workflow. This solves the speed issue moving from Alfred in my use.

## Screencast

CleanShot 2022-06-06 at 14 57 46

## Checklist


RSO (approved):
This makes a lot of sense to me! Thanks for the great addition 😄

PR #1910, opened 2022-06-07
Added new extension, Country Lookup

pernielsentikaer:
## Description

Quickly lookup any country and get informational data

## Checklist

### DIFF
--- a/extensions/country-lookup/src/countries.tsx
+++ b/extensions/country-lookup/src/countries.tsx
@@ -1,4 +1,4 @@
-import { Action, ActionPanel, environment, getPreferenceValues, List, showToast, Toast } from "@raycast/api";
+import { Action, ActionPanel, environment, getPreferenceValues, Icon, List, showToast, Toast } from "@raycast/api";
 import { getData } from "./api";
 import React, { useState } from "react";
 import { Preferences } from "./interface";
@@ -109,7 +109,7 @@ export default function Countries() {
             }
             actions={
               <ActionPanel>
-                <Action title="Toggle Coat of Arms" onAction={() => setShowMarkdown(!showMarkdown)} />
+                <Action icon={Icon.Eye} title="Toggle Coat of Arms" onAction={() => setShowMarkdown(!showMarkdown)} />
                 {entry.maps.googleMaps && (
                   <Action.OpenInBrowser title="Open in Google Maps" url={entry.maps.googleMaps} />
                 )}


extensions/country-lookup/package.json
```diff
@@ -0,0 +1,49 @@
...
+  "name": "country-lookup",
+  "title": "Country Lookup",
+  "description": "Quickly lookup countries and get informational data",
+  "icon": "command-icon.png",
+  "author": "pernielsentikaer",
+  "license": "MIT",
+  "commands": [
+    {
+      "name": "index",
+      "title": "Country Lookup",
```
mattisssa, line 12: Not sure about subtitle here, but I think it would be better:
```suggestion
      "title": "Search Country",
```

extensions/country-lookup/package.json
```diff
@@ -0,0 +1,49 @@
...
+      "description": "Quickly lookup countries",
+      "mode": "view"
+    }
+  ],
+  "preferences": [
+    {
+      "title": "Coat of Arms",
+      "name": "showCoatOfArms",
+      "type": "checkbox",
+      "required": true,
```
mattisssa, line 22:
```suggestion
      "required": false,
```

extensions/country-lookup/package.json
```diff
@@ -0,0 +1,49 @@
...
+      "default": true,
+      "label": "Yes, Show Coat of Arms by default",
```
mattisssa, line 24:
```suggestion
      "label": "Display the Coat of Arms by default",
```

extensions/country-lookup/src/cache.ts
```diff
@@ -0,0 +1,48 @@
+import { environment, trash } from "@raycast/api";
+import { readFileSync } from "fs";
+import { writeFile } from "fs/promises";
+import { resolve } from "path";
+import { Cache, Key } from "swr";
+
+const CACHE_KEY = "swr-cache";
```
mattisssa, line 7: Shouldn't the cache key be unique for the command here?

mattisssa (changes requested):
Can we change this preference to non-required? And just enable it by default.

CleanShot 2022-06-09 at 20 56 31@2x

Would be nice to have an icon here to match the guidelines 😉 
CleanShot 2022-06-09 at 20 58 16@2x

extensions/country-lookup/src/cache.ts
pernielsentikaer: There is no need to change anything here since it's stored inside the extension's own directory

mattisssa (approved):

PR #1919, opened 2022-06-07
Add VSCode extension to docs

tonka3000:
## Description

Add [VSCode extension](https://marketplace.visualstudio.com/items?itemName=tonka3000.raycast) to

- Getting Started
- Debug an extension

@thomaspaulmann @mathieudutour

## Questions
- Add Images?
- Correct pages or move to different locations?


mathieudutour:
> Add Images?

I feel like you could add some in the readme of the vscode extension but I don't think they are needed in the docs

> Correct pages or move to different locations?

Your additions make sense to me. Maybe if we have more tools like this in the future, we can create a separate page for them - but let's go with that for now

mathieudutour (approved):

PR #1931, opened 2022-06-08
Update Placeholder extension

koinzhang:
## Description

- Add new image layout options: List or Grid.

## Screencast

Metadata-1

## Checklist


mattisssa (changes requested):
Hey, thanks for the update! I find one small thing, can we add an icon here to match the Store Guidelines? And we good to go.
CleanShot 2022-06-09 at 22 12 31@2x


I also found crazy performance issues in your extension and though that you could probably use a smaller size for images in `Grid.Item` and for export use the original sizes. Would be nice.

What about performance issues – we're investigating the problem and will fix it soon.

extensions/placeholder/src/components/action-on-placeholder-image.tsx
```diff
@@ -0,0 +1,83 @@
...
+  picsumImage: PicsumImage;
+  preferences: Preferences;
+  page: number;
+  setPage: Dispatch<SetStateAction<number>>;
+}) {
+  const { picsumImage, preferences, page, setPage } = props;
+  return (
+    <ActionPanel>
+      <Action.Push
+        icon={{
```
koinzhang, line 22: Hi @mattisssa 
> Hey, thanks for the update! I find one small thing, can we add an icon here to match the Store Guidelines? And we good to go.

There is an icon here, but I don't know if it is not displayed on your device, I am displaying it normally here

> I also found crazy performance issues in your extension and though that you could probably use a smaller size for images in Grid.Item and for export use the original sizes. Would be nice.

Thank you for your suggestion, I will make changes, it will become more beautiful and better performance

koinzhang:
Hi @mattisssa. Thanks for your review and suggests!
I submitted the latest commit, please keep it under review

extensions/placeholder/src/components/action-on-placeholder-image.tsx
mattisssa: >There is an icon here, but I don't know if it is not displayed on your device, I am displaying it normally here

Can this be the theme issues or something? I'll take a look later

koinzhang: I just opened the command and noticed that this action also does not have an icon, but it was quickly displayed. Maybe it takes time to load the icons in local assets?

koinzhang:
Hi @mattisssa Did you encounter any problems with the review of this extension?

mattisssa:
@koinzhang ops, sorry, forgot to get back to the review here 🙈 
We have fixed this issue natively and improved the performance, will be available in the next update 😉

PR #1955, opened 2022-06-10
Gif Search: Use slugified title as filename

PitNikola:
# Description
This is a minor improvement that uses slugified title as filename instead of the slug provided by the service (which is usually some global unique name containing bunch of unreadable information).

This makes the gifs sent via Slack look slightly nicer.

# Screencast
CleanShot 2022-06-10 at 15 22 55

# Checklist
- [x]  I ran npm run build and [tested this distribution build in Raycast](https://developers.raycast.com/basics/prepare-an-extension-for-store#metadata-and-configuration)
- [x]  I checked that files in the assets folder are used by the extension itself
- [x]  I checked that assets used by the README are placed outside of the metadata folder
—

Created via [Raycast](https://www.raycast.com?ref=signatureGithub)

### DIFF
--- a/extensions/gif-search/src/models/finergifs.ts
+++ b/extensions/gif-search/src/models/finergifs.ts
@@ -1,6 +1,6 @@
 import fetch from "node-fetch";
 
-import type { APIOpt, IGif, IGifAPI } from "./gif";
+import { APIOpt, IGif, IGifAPI, slugify } from "./gif";
 
 export interface FinerGifsClubResults {
   results: FinerGif[];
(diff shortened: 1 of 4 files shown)


extensions/gif-search/src/components/GifDetailsActions.tsx
```diff
@@ -37,13 +37,21 @@ export function getActions(item: IGif, showViewDetails: boolean, service?: Servi
 
   const removeFav = () => dispatch({ type: "remove", save: true, favIds: actionIds, service });
 
+  const slugify = (str: string) =>
```
josephschmitt, line 40: I think this logic belongs in the service model's map function. It can then be applied to the slug properly of it which all gif services already use.

josephschmitt (approved):
🙌🙌

PitNikola:
Forgot to add myself as contributor 😱 

disappointed-kevin-sorbo-gif

Issue #1033, opened 2022-03-05
[Search npm Packages] Include popularity, stars etc. in search results


MKaciuba:
# Extension – [Search npm Packages](https://github.com/raycast/extensions/tree/de047f2e9ad02d2c79a526f2291f64f74aca05b6/extensions/search-npm/)
Author: @mrmartineau

First of all, I absolutely love the extension!

I think one useful addition would be to include info such as popularity, GitHub star count, last publish date etc. I usually base my decision about using a particular package on this info.

mrmartineau:
@MKaciuba hi mate, thanks for the feedback. I would also like to add more info to the output, but there isn't much space for it in the Raycast UI. 3rd party extensions have a `title`, `description` (second line), image and the right-hand bit where we put the version, to put information. That doesn't leave much space for the new things that you've suggested. Have a look at the Raycast extensions API docs and see if you can figure out where that information could go.

Actually, we could put that info in the action panel, would that be acceptable?

MKaciuba:
What do you think about giving a configurable option to choose one of those data points to appear next to the version simply by concatenation, like this: `1.2.3 | ⭐456`? Then you could have more detailed info in the action panel as you propose.

mrmartineau:
They could also work. If I get some time, I will add this stuff.

mrmartineau:
@MKaciuba FWIW a typical response from the npms.io suggestions API (that the extension uses) looks like this:

```json
{
  "package": {
    "name": "react",
    "scope": "unscoped",
    "version": "18.0.0",
    "description": "React is a JavaScript library for building user interfaces.",
    "keywords": [
      "react"
    ],
    "date": "2022-03-29T16:00:04.223Z",
    "links": {
      "npm": "https://www.npmjs.com/package/react",
      "homepage": "https://reactjs.org/",
      "repository": "https://github.com/facebook/react",
      "bugs": "https://github.com/facebook/react/issues"
    },
    "publisher": {
      "username": "acdlite",
      "email": "npm@andrewclark.io"
    },
    "maintainers": [
      {
        "username": "gaearon",
        "email": "dan.abramov@gmail.com"
      },
      {
        "username": "acdlite",
        "email": "npm@andrewclark.io"
      },
      {
        "username": "fb",
        "email": "opensource+npm@fb.com"
      },
      {
        "username": "trueadm",
        "email": "dg@domgan.com"
      },
      {
        "username": "sophiebits",
        "email": "npm@sophiebits.com"
      },
[...]

PR #1967, opened 2022-06-12
Create bundlephobia extension

thuutri2710:
## Description

A simple extension to search bundle size and related information of npm packages via Bundlephobia API.

## Screencast

image

## Checklist


mathieudutour:
What do you think about adding this information to the existing npm extension instead of creating a new one? Sounds pretty useful to have when browsing npm?

thuutri2710:
image
As you see, the npm extension, it doesnt have enough space to add more information about bundle size. We can show it in the detail of every item, but the npm extension doesnt have detail currently. Moreover, in some specific cases, we only need to find the bundle size of the package instantly, my extension helps to do that conveniently. @mathieudutour

mathieudutour:
I'm sure @mrmartineau would be open to have a Detail view (some experimentation done here: https://github.com/raycast/extensions/issues/1033#issuecomment-1111457960).

> Moreover, in some specific cases, we only need to find the bundle size of the package instantly, my extension helps to do that conveniently

I don't really understand how it'd be more convenient than if it's shown in the npm extension?

mathieudutour:
yeah there could be a toggle to switch between the list + detail and just the list, and we could store this state in the local storage to make sure your choice is remembered

mrmartineau:
Hi folks, I gotta say that I think a specific bundlephobia extension would be better than adding this functionality to my npm extension. Even though there is an action in my extension for navigating to the bundlephobia site, I use my "cost" Quicklink frequently to navigate to bundlephobia more quickly.

### npm extension action
image

### Quicklink
image

thuutri2710:
@mrmartineau agreed with my suggestion. @mathieudutour how about your thought? Again, I want to remind that my extension is to help to see the bundle size of the npm package as soon as possible.

extensions/bundlephobia/src/index.tsx
```diff
@@ -0,0 +1,98 @@
...
+  const [results, setResults] = useState<NpmsFetchResponse>([]);
+  const [detailResult, setDetailResult] = useState<PackageResultModel>(null);
+  const [loading, setLoading] = useState<boolean>(false);
+  const [loadingDetail, setLoadingDetail] = useState<boolean>(false);
+  const [searchTerm, setSearchTerm] = useState<string>("");
+  const [links, setLinks] = useState<Links | null>(null);
+
+  const onSearchTextChange = async (text: string) => {
+    setLoading(true);
+    const response = await fetchSuggestionsBunldephobia(text);
```
mathieudutour, line 18: For both `onSearchTextChange` and `onSelectionChange` you have to abort the previous call when making a new one otherwise you might end up with race conditions and show the result that doesn't correspond to the query/selection. Have a look at https://github.com/raycast/extensions/blob/main/examples/api-examples/src/typeahead-search.tsx if you need an example

mathieudutour (changes requested):
Left a comment about race conditions. Could you also remove the list-icon.png asset?

extensions/bundlephobia/src/index.tsx
thuutri2710: I resolved the race condition, please take a took @mathieudutour 🙏

extensions/bundlephobia/src/fetchUtils.ts
```diff
@@ -0,0 +1,42 @@
...
+    const response = await fetch(`https://api.npms.io/v2/search/suggestions?q=${packageName}`, {
+      method: "get",
+      signal,
+    });
+    const json = await response.json();
+
+    return json as NpmsFetchResponse;
+    //eslint-disable-next-line @typescript-eslint/no-explicit-any
+  } catch (e: any) {
+    showToast(Toast.Style.Failure, "Could not fetch matched packages");
```
mathieudutour, line 38: I don't think you should show a toast here since you one in `useSearcha already

mathieudutour (approved):

PR #1987, opened 2022-06-13
Clamp: Add Calculate Linear Clamp Extension

teziovsky:
## Description

This extension can generate a responsive linear clamp function based on px or rem values.
It's created based on the [CSS Tricks article](https://css-tricks.com/linearly-scale-font-size-with-css-clamp-based-on-the-viewport/).

## Screencast

calculate-linear-clamp-01

## Checklist


pernielsentikaer (changes requested):
Hi @teziovsky  👋 

I would be nice with some validations in the two commands to ensure that. the input is not empty and only numbers

If I write "hello" and "you" in one of them I get this output: `clamp(NaNrem, NaNrem + NaNvw, NaNrem)`

teziovsky:
Hi @pernielsentikaer 👋🏼

It's my pleasure!

I recently added validation to all fields. I hope now this extension will be safer in use.

You can review this extension now 🔥

teziovsky:
I think that it's a good idea, but for now, I don't see any use cases to extend. Maybe we could add another field for example select, to select bezier curves (not only linear).

pernielsentikaer:
Then just keep it as it for now :)

pernielsentikaer:
We could change the name/title of the extension to something like `CSS Calculations` and the folder to `css-calculations` instead?

teziovsky:
Yes, of course, done 🚀

PR #1999, opened 2022-06-14
Fix character encoding in `TextArea`

ickas:
## Description

The TextArea value is read through the `text` state, which is always triggered when the item's value changes. This conflicts with some character encodings (e.g., Portuguese words, Spanish, etc.) because it logs whenever a character or graphic sign (e.g., accent) is clicked.

## Screencast

Before ❌

CleanShot 2022-06-14 at 20 19 54@2x

After ✅

CleanShot 2022-06-14 at 20 20 16@2x

## Checklist

### DIFF
--- a/extensions/google-translate/CHANGELOG.md
+++ b/extensions/google-translate/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Google Translate Changelog
 
-## [Fix] - 2022-06-14
+## [Fix] - 2022-06-15
 
 Fix character encoding in `TextArea` on translate form
 


gebeto (approved):
Looks good

I'm not sure is it correct to not pass `value` when `onChange` used, but if it is working, it is good to me

mattisssa:
Hey @henriquemacedo! 
I just returned back the `value` configuration into the extension to check why did even happen and cannot repro the case you're showing in the description. Can you recommend me the case? How do you enter the `ç` and other characters? Because If enter them with _option + a_ or other letters everything works fine.

PR #2054, opened 2022-06-21
[ensk-is] Add fuse.js sorting to ensk-is for improved result ranking

jokull:
## Description

Add fuse.js sorting to ensk-is for improved result ranking


pernielsentikaer (approved):
Looks good to me, let's get it in 🚀

PR #2092, opened 2022-06-26
Add Regex REPL extension

hongkiulam:
## Description

Simple extension for testing regular expressions:
- The user can type in a regular expression and some text which they want to test.
- The extension will give instant feedback by underlining the matching section
- The extension will provide validation for the inputted regex

## Screencast

image
image
image
image

## Checklist


pernielsentikaer:
@hongkiulam 🔥

What about just having another command into the "current" extension instead?

hongkiulam:
@pernielsentikaer fine with me :) however are you considering just dropping it in?

I feel work is needed to separate out the various concerns to ensure each command we have is clearly doing only one thing, instead of two commands which do a similar thing but in a different way

pernielsentikaer:
What are the differences? It seems like it's doing the same but different for me.

pernielsentikaer:
Hi @hongkiulam 👋 

Do you still want to finish this?

hongkiulam:
Hey @pernielsentikaer, sorry for the lack of activity here.
Happy to finish it but still not sure how to fit it into the current extension... as you said they are same but different.

The existing extension allows the user to enter a test string and then proceed to test a pattern against it in the following step.
This PR allows the user to enter the pattern, modifiers and test string in a single form and receive a live result.

Packing them into the same extension seems odd to me and my preference would be to have separate extensions, but I'm sure that's not desirable on your end.

Lmk your thoughts, I'm also fine with closing this PR as I'm not too bothered whether this gets published or not

pernielsentikaer:
I would like to have it in the store, we can go with a standalone extension. We just need to find another name(also for the folder) - can you think of anything?

hongkiulam:
Apologies, completely forgot about this - stalebot reminded me xD Renamed to regex-repl

pernielsentikaer (approved):

PR #2130, opened 2022-06-29
Portuguese Primeira Liga: Add cache and other improvements

danielsequeira:
## Description

Added cache mechanism and removed detail view when opening a newspaper cover.

## Screencast

screenshot-04
screenshot-05

## Checklist

### DIFF
--- a/extensions/portuguese-primeira-liga/CHANGELOG.md
+++ b/extensions/portuguese-primeira-liga/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Portuguese Primeira Liga Changelog
 
+## [Improvements] - 2022-07-01
+
+- Added cache mechanism
+- Removed detail view for newspaper covers
+
 ## [New Additions] - 2022-06-18
 
 - Removed inset from Grid


mattisssa:
I quickly looked at your extension and found that your Standings command with List is a bit laggy while I'm searching something. I found that you're [using throttler](https://github.com/raycast/extensions/blob/a1bb28b7699502df44c049d109f58f92f6aa8b6a/extensions/portuguese-primeira-liga/src/standings.tsx#L10) inside the List. 

What is purpose of using it since you don't have networking searches?

danielsequeira:
> I quickly looked at your extension and found that your Standings command with List is a bit laggy while I'm searching something. I found that you're [using throttler](https://github.com/raycast/extensions/blob/a1bb28b7699502df44c049d109f58f92f6aa8b6a/extensions/portuguese-primeira-liga/src/standings.tsx#L10) inside the List.
> 
> What is purpose of using it since you don't have networking searches?

Good catch, no purpose whatsoever. I just have been playing around with filtering and other properties of the List component since I started developing the extension and totally missed removing the `throttle`. I will do it now. Thank you

PR #2133, opened 2022-06-29
Harvest

eluce2:
## Description

Update for v1.37.1
Add support for automatically focusing the next field after selecting a project or task

## Screencast

https://share.cleanshot.com/yZ0BkA

## Checklist

### DIFF
--- a/extensions/harvest/CHANGELOG.md
+++ b/extensions/harvest/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Harvest Changelog
 
+## [Updated branding & icons] - 2022-07-30
+
+- Updated brading and icons
+- Updated Raycast API
+
 ## [Edit Time Entries] 2022-04-08
 
 - Added: You can now edit previous time entries from the action menu or `cmd + E`.


pernielsentikaer:
Hi @eluce2 👋 

I'm not so sure this is the way to go, it would be the only thing in Raycast with this behavior, this would cause a really weird UX

eluce2:
@pernielsentikaer This is the behavior I wanted when I first built the extension, but it wasn't possible until a recent API update. I better matches the Harvest Desktop app, which is what most people would be coming from when they try this extension. It has always felt very unnatural to me to remain focused on the field after selecting something from a drop-down menu.

If those arguments don't convince you, could I keep this feature behind a preference that is disabled by default?

eluce2:
@pernielsentikaer I have reverted this change, now this PR is to simply update to the latest version of the API.

While it saddens me to not have this behavior, in my experience using it over the past few weeks it hasn't always worked in the way I would expect it to, so I'm ok with reverting

pernielsentikaer:
Hi @eluce2 👋 

Sorry for the long response time. I'm on vacation and only checking from time to time 😄 

Yes, I could imagine there are some drawbacks since it's not made for that; let's keep it as is for now.

Approved! 🔥

PR #2136, opened 2022-06-29
Apple Music: Added artwork for tracks, albums, and playlists in grid view.

yug2005:
## Description

Added optional grid view for tracks, albums, and playlists. There are options for grid item size as well. 

Showing the artwork using AppleScript. However due to performance reasons, I need to : 

* Resize some images (this increases loading time) 
* Limit the number of shown results. 
* Removed the `loadAll` feature for albums.

Let me know if you have any ideas to improve performance. Perhaps another method to resize the images faster. 

However, that being said, I still think this is awesome!

## Screencast

music 2022-06-29 at 16 19 27

music 2022-06-29 at 16 19 53

music 2022-06-29 at 16 22 38

## Checklist

### DIFF
--- a/extensions/music/package.json
+++ b/extensions/music/package.json
@@ -232,6 +232,13 @@
         "track",
         "rating"
       ]
+    },
+    {
+      "name": "repeat", 
+      "title": "Toggle Repeat",
+      "subtitle": "Apple Music",
+      "description": "Toggle repeat mode.",
+      "mode": "no-view"
     }
   ],
   "dependencies": {
--- a/extensions/music/src/repeat.tsx
+++ b/extensions/music/src/repeat.tsx
@@ -1,0 +1,24 @@
+import { Toast, showToast } from "@raycast/api";
+import { runAppleScript } from "run-applescript";
+
+export default async () => {
+  const script = `
+    tell application "Music"
+      set currentState to get song repeat
+      if currentState is equal to one or currentState is equal to none then
+        set song repeat to all
+        return "all"
+      else 
+        set song repeat to one
+        return "one"
+      end if
+    end tell
+    `;
+  try {
+    let state = await runAppleScript(script);
+    state = state[0].toUpperCase() + state.slice(1);
+    showToast(Toast.Style.Success, `Set Song Repeat to ${state}`);
+  } catch {
+    showToast(Toast.Style.Failure, `Error Setting Song Repeat`);
+  }
+};


Rawnly:
Thank you! This looks dope! I’ll try it in the afternoon 💪

extensions/music/src/util/scripts/playlists.ts
```diff
@@ -56,3 +59,38 @@ export const getPlaylists = (kind: PlaylistKind): TE.TaskEither<Error, string> =
...
+  const promises = ids.map(async (id) => {
+    const res = await runAppleScript(`
+      tell application "Music"
+        set playlistArtwork to first artwork of first playlist of (every playlist whose id is ${id})
+        set playlistImage to data of playlistArtwork
+      end tell
+      return playlistImage
+    `);
+    let binary = null;
+    try {
```
Rawnly, line 77: Everything seems ok to me. The only thing I would do differently is by wrapping the following try/catch into a separate generic function (since the code seems to be almost identical in 3 files.)

extensions/music/src/play-library-album.tsx
```diff
@@ -1,85 +1,101 @@
...
             O.matchW(() => [] as ReadonlyArray<Album>, parseResult<Album>())
-          )
+          );
+        }
       ),
       T.map(setAlbums)
     )();
   };
 
+  useEffect(() => {
```
Rawnly, line 52: as in the rest of the extension I would replace `Promises` with `TaskEither`. But I understand that the library might be a wall for new contributors.

Rawnly:
I left a few comments. Feel free to ignore them. These are not hard requirements I can fix those things in a new PR next time!

extensions/music/src/util/scripts/playlists.ts
yug2005: Sounds good, I'll add this soon.

extensions/music/src/play-library-album.tsx
yug2005: Yeah I didn't really understand how to use `TaskEither`. Feel free to make the conversion if it is simple.

Rawnly (approved):

yug2005:
@Rawnly is it possible to play a song on repeat? Also, I am planning to add a push action for playlists, to see the songs of the playlist.

Rawnly:
I don't know honestly, there's a website which references a lot of scripts (right now I don't remember the name)

Rawnly:
http://dougscripts.com/itunes/index.php

yug2005:
Not ready to merge yet. I will add a few features soon.

yug2005:
May I ask what is the reason/advantage of using TaskEither?

yug2005:
Added some awesome new features!

* You can see songs in a playlist. 
  * However the results still have to be limited.
  * The list view is best for seeing more results because the artworks are really small.
  * You can see up to 100 results. 
* List detail view with a lot of information. 
* Preferences for new features in extension configurations

https://user-images.githubusercontent.com/77864399/176850353-93c5b990-b4e1-4edb-b834-01b44c6f1aeb.mov

Rawnly:
> Added some awesome new features!
Love it!

Rawnly:
> May I ask what is the reason/advantage of using TaskEither?

Sure! [`TaskEither`][task-either] is the composition of 2 types, [`Task`][task] and [`Either`][either]. 

The [`Task`][task] type  "represents an asynchronous computation that yields a value of type A and never fails". But what about the errors? Here it comes our  [`TaskEither`][task-either] which is an alias for `type TaskEither<E, A> = Task<Either<E, A>>`.

The [`Either`][either] is a union type that can holds 2 values, an error (`left`) and the return value of the function (`right`). You might see some similarities with the `Result` type in languages like [`rust`](https://doc.rust-lang.org/rust-by-example/error/result.html) and [`swift`](https://developer.apple.com/documentation/swift/result).

Everything is done to be type-safe. Since the error in the `try/catch` [cannot be typed][ts1196]. 
You can see an example usage in the `src/utils/apple-script.ts` file.

This explanation might not be super complete but it should clear the difference between a `Promise` and a `TaskEither`.
Once you've done your edits, if you want, I can contribute your PR by converting the code (if you're having difficulties with it).

I don't want this to be a problem for new contributors so, is fine for me if you want to learn something new or if you find it interesting. But if not, don't worry 😉

[task]: https://gcanti.github.io/fp-ts/modules/Task.ts.html
[either]: https://gcanti.github.io/fp-ts/modules/Either.ts.html
[task-either]: https://gcanti.github.io/fp-ts/modules/TaskEither.ts.html
[ts1196]: https://typescript.tv/best-practices/error-ts1196-catch-clause-variable-type-annotation/

yug2005:
Hello I haven't been able to reproduce the error for start playlist. Could you see if that playlist does not have any artwork? Maybe running the script in script editor to see what the issue is. 

For the previous implementation of the list view, the `loadAll` feature was only there for the albums right? I removed that feature because when you started to search for something while `loadAll` was running, it would show you the results for the search query but then flicker back to show all songs once `loadAll` was complete. List view is default for the playlist tracks view but I do think grid view looks really nice for the other views. Are you thinking about bringing back the loadAll feature for tracks and albums? 

For Task Either, if you would like to implement it, please do. I am busy with some other things now.

Rawnly:
@yug2005 

> Hello I haven't been able to reproduce the error for start playlist. Could you see if that playlist does not have any artwork? Maybe running the script in script editor to see what the issue is.

Yes it was caused by a Playlist without artwork. Can you please handle this? (Also for some tracks i'm having the same problems)

yug2005:
@Rawnly fixed the issue.

extensions/music/src/start-playlist.tsx
```diff
@@ -18,6 +18,7 @@ import {
   getMaxNumberOfResults,
   getGridItemSize,
 } from "./util/listorgrid";
+import { privateEncrypt } from "crypto";
```
Rawnly, line 21: unused import :)

Rawnly (approved):

Rawnly:
I'm having trouble when trying to "show tracks" inside "start playlist". 
Error isn't logged but I believe is with artwork. I'm trying to implement TaskEither so maybe i can fix that too. I'll let you know asap.

music-1

yug2005:
Make sure you are not using the id of the track as that sometimes varies (especially with tracks of playlists). That's why I made the switch to using database ID as that is always consistent.

Rawnly:
@yug2005 I think too much stuff is going on here. Maybe we should split this PR.. Unfortunately I don't have enough time to implement TaskEither, but i'd like to reintroduce the loadAll maybe with pagination and infinite scrolling. You can [do a little trick](https://raycastcommunity.slack.com/archives/C028DTXTW3Y/p1656685668718759?thread_ts=1656684686.778479&cid=C028DTXTW3Y) via `onSelectedId` to make it work. Specific code can be found in the [Github Gist extension source code](https://github.com/raycast/extensions/tree/3b703e4283e2a8a7fad07781540374fbbb74e0b5/extensions/github-gist/)

yug2005:
@Rawnly I tried using `onSelectedId` but it was difficult to implement and did not work well. I was trying to implement it for the grid view. Fetching the artwork takes a long time and once you have a lot of artworks, unless you resize to something very small, you will run out of memory very quickly.

I guess that it would be doable for the list view if you resized to about 32x32 or smaller but even then it is possible to run out of memory. As for the grid view, its impossible if you want the artwork to look decent. 

I think unless TaskEither has performance benefits, it is probably okay to skip for this PR as you could implement in a future one.

pernielsentikaer:
What are the status here now? 🙂

Rawnly:
> What are the status here now? 🙂

@pernielsentikaer 

As i said before, the idea is cool but doesn't convince me 100%. Starting from the memory problem (due to images). I'm sorry but I'd wait for a better solution, maybe some other way to fetch images. The current implementation is slower and I don't think that images are worth the lack of speed :/

pernielsentikaer:
Thanks for the update 🔥 I'm converting this to a draft for now 🙂

PR #2198, opened 2022-07-07
fix(random-email): fix preference typings

akellbl4:
## Description

A bug was introduced because of incorrect typing. A preference option can be `undefined`.

Probably it's worth trying to fix it on `@raycast/api` level by wrapping `Values` generic with `Partial` or infer type from the preferences config.

```typescript
export function getPreferenceValues<Values extends PreferenceValues = PreferenceValues>(): Partial<Values>;
```

## Checklist

### DIFF
--- a/extensions/random-email/src/utils.ts
+++ b/extensions/random-email/src/utils.ts
@@ -2,7 +2,7 @@ import { getPreferenceValues } from "@raycast/api";
 
 interface Preferences {
   domain: string;
-  prefix: string;
+  prefix: string | undefined;
 }
 
 export const generateEmail = () => {


pernielsentikaer:
Hi @rkjaer 👋 

Are you getting any errors with this extension or just fixing a user error?

> .. or infer type from the preferences config

Yep, that's something we could explore at some point 🙂

akellbl4:
@pernielsentikaer, I added an option to configure prefix. Then I saw fix for that.
I was concatenating prefix to email always, but if prefix wasn't  added then the variable is `undefined`

akellbl4:
If you need some help on typings I could try to do inference.

PR #2214, opened 2022-07-08
Apple Notes: Super fast for iCloud Notes

mathieudutour:
## Description

Load the sql database to fetch the iCloud notes - that's a lot faster than using applescript.

## Checklist

### DIFF
--- a/extensions/apple-notes/src/errors.tsx
+++ b/extensions/apple-notes/src/errors.tsx
@@ -1,4 +1,6 @@
-import { ActionPanel, Detail, Icon, Action } from "@raycast/api";
+import { ActionPanel, Detail, Icon, Action, environment } from "@raycast/api";
+import path from "path";
+import fileUrl from "file-url";
 
 export class PermissionError extends Error {
   constructor(message: string) {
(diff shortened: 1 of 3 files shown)


ron-myers:
This is excellent @mathieudutour - huge upgrade.  Well done.

I pulled the branch locally and works as I would expect.

What I am wondering is if we should make searching locally via AppleScript an option that is enabled vs the default behavior.  This could be added at a later time - I can take care of it after this is merged.

I have a local branch optimizing the local search via AppleScript but it won't touch the performance of using the iCloud db.

extensions/apple-notes/src/errors.tsx
```diff
@@ -0,0 +1,39 @@
...
+}
+
+export const isPermissionError = (error: unknown) => {
+  return error instanceof Error && error.name === "PermissionError";
+};
+
+// @TODO: This screen should be handled by Raycast itself (https://github.com/raycast/extensions/issues/101)
+const permissionErrorMarkdown = `## Raycast needs full disk access in order to display your Apple Notes.
+
+![Full Disk Access Preferences Pane](https://i.imgur.com/3SAUwrx.png)
```
ron-myers, line 17: The `full-disk-access` image is in the repo.  Not sure if you intended to use it here.

mathieudutour: Ah you're right!

PR #2237, opened 2022-07-12
Update publish command in documentation

RobertBolender:
## Description

Documentation change only. There is no `publish` script in the default `package.json` created by the `Create Extension` command, so the `npm run publish` command errors. It was intended to be `npx ray publish`, which works.


danielsequeira:
@RobertBolender thanks for pointing it out. We just went ahead and added the publish script to our create extension command. With that said there's no need for this documentation change.

PR #2266, opened 2022-07-14
[GitHub Repository Search] Add dropdown for filtering by user/org

jberglinds:
## Description
Adds a dropdown to the search field for filtering the search by user/orgs, similarly to how the native "Search Repositories" extension works. Also, includes forks when searching for repositories like the native extension.

* "All Repositories" -> The same as before, searching all of Github (Default selection)
* "My Repositories" -> Searches for personal repos and repos of organizations that you are a member of
* "{Github username}" -> Searches only for personal repos
* "{Organization username}" -> Searches only for repos of the chosen org

## Screencast
github-repository-search 2022-07-14 at 16 05 04
github-repository-search 2022-07-14 at 16 05 56
github-repository-search 2022-07-14 at 16 03 31

## Checklist

### DIFF
--- a/extensions/github-repository-search/src/index.tsx
+++ b/extensions/github-repository-search/src/index.tsx
@@ -9,6 +9,7 @@ import {
   useNavigation,
   Action,
   Toast,
+  getPreferenceValues,
 } from "@raycast/api";
 import { useState } from "react";
 import { Repository, UserDataResponse } from "./types";
@@ -20,7 +21,12 @@ import { useRepositoryReleases } from "./useRepositoryReleases";
 import { OpenInWebIDEAction } from "./website";
 import { useUserData } from "./useGithubUser";
 
+const preferences: {
+  includeForks: boolean;
+} = getPreferenceValues();
+
 export default function Command() {
+  console.log(preferences);
   const [searchText, setSearchText] = useState<string>();
   const [searchFilter, setSearchFilter] = useState<string>("");
   const [debouncedSearchText] = useDebounce(searchText, 200);
@@ -28,8 +34,8 @@ export default function Command() {
     data: repositories,
     error: searchRepositoriesError,
     isLoading: isLoadingRepositories,
-  } = useRepositories(debouncedSearchText, searchFilter);
-  const { data: userData, error: fetchUserDataError } = useUserData();
+  } = useRepositories(`${searchFilter} ${debouncedSearchText} fork:${preferences.includeForks}`);
+  const { data: viewer, error: fetchUserDataError } = useUserData();
 
   const {
     repositories: visitedRepositories,
@@ -59,20 +65,20 @@ export default function Command() {
     <List
       isLoading={isLoading}
       onSearchTextChange={setSearchText}
-      searchBarAccessory={<FilterDropdown userData={userData} onFilterChange={setSearchFilter} />}
+      searchBarAccessory={<FilterDropdown viewer={viewer} onFilterChange={setSearchFilter} />}
     >
       <List.Section
         title="Visited Repositories"
         subtitle={visitedRepositories ? String(visitedRepositories.length) : undefined}
       >
         {visitedRepositories
-          ?.filter(
-            (r) =>
-              r.nameWithOwner.includes(searchText ?? "") &&
-              r.nameWithOwner.match(
-                // Converting query filter string to regexp:
-                `${searchFilter?.replaceAll(/org:|user:/g, "").replaceAll(" ", "|")}/.*`
-              )
+          ?.filter((r) => r.nameWithOwner.includes(searchText ?? ""))
+          .filter((r) =>
+            // Filter on dropdown selection value
+            r.nameWithOwner.match(
+              // Converting query filter string to regexp:
+              `${searchFilter?.replaceAll(/org:|user:/g, "").replaceAll(" ", "|")}/.*`
+            )
           )
           .map((repository) => (
             <RepositoryListItem key={repository.id} repository={repository} onVisit={visitRepository} />
@@ -91,31 +97,31 @@ export default function Command() {
 }
 
 function FilterDropdown({
-  userData,
+  viewer,
   onFilterChange,
 }: {
-  userData: UserDataResponse["viewer"] | undefined;
+  viewer?: UserDataResponse["viewer"];
   onFilterChange: (filter: string) => void;
 }) {
-  return (
+  return viewer ? (
     <List.Dropdown tooltip="Filter Repositories" storeValue={true} onChange={onFilterChange}>
       <List.Dropdown.Item title={"All Repositories"} value={""} />
-      {userData && (
+      {viewer && (
         <List.Dropdown.Item
           title={"My Repositories"}
-          value={`user:${userData.login} ${userData.organizations.nodes.map((org) => `org:${org.login}`).join(" ")}`}
+          value={`user:${viewer.login} ${viewer.organizations.nodes.map((org) => `org:${org.login}`).join(" ")}`}
         />
       )}
-      {userData && (
+      {viewer && (
         <List.Dropdown.Section>
-          <List.Dropdown.Item title={userData.login} value={`user:${userData.login}`} />
-          {userData.organizations.nodes.map((org) => (
+          <List.Dropdown.Item title={viewer.login} value={`user:${viewer.login}`} />
+          {viewer.organizations.nodes.map((org) => (
             <List.Dropdown.Item key={org.login} title={org.login} value={`org:${org.login}`} />
           ))}
         </List.Dropdown.Section>
       )}
     </List.Dropdown>
-  );
+  ) : null;
 }
 
 function RepositoryListItem(props: { repository: Repository; onVisit: (repository: Repository) => void }) {
--- a/extensions/github-repository-search/src/useRepositories.ts
+++ b/extensions/github-repository-search/src/useRepositories.ts
@@ -34,7 +34,7 @@ query SearchRepositories($searchQuery: String!) {
   }
 }`;
 
-export function useRepositories(searchText: string | undefined, searchFilter: string | undefined) {
+export function useRepositories(searchQuery: string | undefined) {
   const [state, setState] = useState<{
     data?: SearchRepositoriesResponse["search"];
     error?: Error;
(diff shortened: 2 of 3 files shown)


extensions/github-repository-search/src/index.tsx
```diff
@@ -95,7 +153,7 @@ function Actions(props: { repository: Repository; onVisit: (repository: Reposito
           icon={{ source: "pull-request.png", tintColor: Color.PrimaryText }}
           title="Open Pull Requests"
           url={`${props.repository.url}/pulls`}
-          shortcut={{ modifiers: ["cmd"], key: "p" }}
+          shortcut={{ modifiers: ["cmd", "shift"], key: "p" }}
```
jberglinds, line 156: cmd+p is the default hotkey to trigger the dropdown so this had to change

extensions/github-repository-search/src/index.tsx
```diff
@@ -121,7 +179,7 @@ function Actions(props: { repository: Repository; onVisit: (repository: Reposito
             icon={{ source: "project.png", tintColor: Color.PrimaryText }}
             title="Open Projects"
             url={`${props.repository.url}/projects`}
-            shortcut={{ modifiers: ["cmd", "shift"], key: "p" }}
+            shortcut={{ modifiers: ["cmd", "shift", "opt"], key: "p" }}
```
jberglinds, line 182: this is quite a lot, open to suggestions

extensions/github-repository-search/src/index.tsx
```diff
@@ -11,55 +11,113 @@ import {
...
         subtitle={visitedRepositories ? String(visitedRepositories.length) : undefined}
       >
         {visitedRepositories
-          ?.filter((r) => r.nameWithOwner.includes(searchText ?? ""))
+          ?.filter(
+            (r) =>
+              r.nameWithOwner.includes(searchText ?? "") &&
+              r.nameWithOwner.match(
+                // Converting query filter string to regexp:
+                `${searchFilter?.replaceAll(/org:|user:/g, "").replaceAll(" ", "|")}\/.*`
```
jberglinds, line 74: not proud of this but the alternative is quite some extra state to manage

DonDebonair:
This looks great! How can we help test this?

jberglinds:
As this week is "launch week" at raycast I'm guessing that Thomas is really busy. 

If you want to test it out before it gets merged you can clone the changes and run `npm install && npm build` and it should install as a separate extension, @DonDebonair

DonDebonair:
I tested it (need to run `npm run dev` to get it to work btw) and it works great! The added option to search only your own repos or even within a specific org, together with caching visited repos, is exactly what I need!

Would be nice if this could be merged and published soon :)

thomaspaulmann:
Sorry that this drags so long. Been busy with Launch Week. Gonna pick it up next week.

jberglinds:
Hey @thomaspaulmann, any chance you could take a look at this soon-ish?

extensions/github-repository-search/src/index.tsx
thomaspaulmann: Makes sense, thanks!

extensions/github-repository-search/tsconfig.json
```diff
@@ -3,7 +3,7 @@
   "display": "Node 16",
   "include": ["src/**/*"],
   "compilerOptions": {
-    "lib": ["es2020"],
+    "lib": ["es2021"],
```
thomaspaulmann, line 6: Why did you change this?

extensions/github-repository-search/src/useRepositories.ts
```diff
@@ -53,7 +53,9 @@ export function useRepositories(searchText: string | undefined) {
       setState((oldState) => ({ ...oldState, isLoading: true }));
 
       try {
-        const { search } = await octokit.graphql<SearchRepositoriesResponse>(SEARCH_REPOSITORIES_QUERY, { searchText });
+        const { search } = await octokit.graphql<SearchRepositoriesResponse>(SEARCH_REPOSITORIES_QUERY, {
+          searchQuery: `${searchFilter} fork:true ${searchText}`,
```
thomaspaulmann, line 57: Should it be a preference if you want to include forks in the search? It's a bit confusing that you have `searchText` and `searchFilter`. Wouldn't it be easier to inject the entire `searchQuery` from outside the hook?

extensions/github-repository-search/src/index.tsx
thomaspaulmann: I guess it's fine.

extensions/github-repository-search/src/index.tsx
```diff
@@ -11,55 +11,113 @@ import {
...
       </List.Section>
     </List>
   );
 }
 
+function FilterDropdown({
+  userData,
+  onFilterChange,
+}: {
+  userData: UserDataResponse["viewer"] | undefined;
```
thomaspaulmann, line 97: Isn't that the `viewer`?
```suggestion
  viewer?: UserDataResponse["viewer"];
```

extensions/github-repository-search/src/index.tsx
```diff
@@ -11,55 +11,113 @@ import {
...
+        <List.Dropdown.Section>
+          <List.Dropdown.Item title={userData.login} value={`user:${userData.login}`} />
+          {userData.organizations.nodes.map((org) => (
+            <List.Dropdown.Item key={org.login} title={org.login} value={`org:${org.login}`} />
+          ))}
+        </List.Dropdown.Section>
+      )}
+    </List.Dropdown>
+  );
+}
```
thomaspaulmann, line 119: How about not returning a dropdown if we don't have any user data? A dropdown with just a single item doesn't make much sense.
```suggestion
  return userData ? (
    <List.Dropdown tooltip="Filter Repositories" storeValue={true} onChange={onFilterChange}>
      <List.Dropdown.Item title="All Repositories" value="" />
        <List.Dropdown.Item
          title="My Repositories"
          value={`user:${userData.login} ${userData.organizations.nodes.map((org) => `org:${org.login}`).join(" ")}`}
        />
        <List.Dropdown.Section>
          <List.Dropdown.Item title={userData.login} value={`user:${userData.login}`} />
          {userData.organizations.nodes.map((org) => (
            <List.Dropdown.Item key={org.login} title={org.login} value={`org:${org.login}`} />
          ))}
        </List.Dropdown.Section>
    </List.Dropdown> : null
  );
}
```

extensions/github-repository-search/src/index.tsx
```diff
@@ -11,55 +11,113 @@ import {
...
       >
         {visitedRepositories
-          ?.filter((r) => r.nameWithOwner.includes(searchText ?? ""))
+          ?.filter(
+            (r) =>
+              r.nameWithOwner.includes(searchText ?? "") &&
+              r.nameWithOwner.match(
+                // Converting query filter string to regexp:
+                `${searchFilter?.replaceAll(/org:|user:/g, "").replaceAll(" ", "|")}/.*`
+              )
```
thomaspaulmann, line 75: Shouldn't that be a disjunction?
```suggestion
              r.nameWithOwner.includes(searchText ?? "") ||
              r.nameWithOwner.match(
                // Converting query filter string to regexp:
                `${searchFilter?.replaceAll(/org:|user:/g, "").replaceAll(" ", "|")}/.*`
              )
```

thomaspaulmann (changes requested):
I left a few comments in the PR. My main concern is the new required scopes. People that don't have a PAT with the user and org scope will constantly get failure toasts. Given that the extension is used by a lot of people, we should find a way to work around this.

github-repository-search 2022-07-29 at 10 37 46

A few ideas:
- Somehow check the scope and do the right thing. If user/org scope is available, add your changes, if not, then stick with the previous behavior.
- Put your changes behind a new preference that people have to opt-in. I don't like preferences for such things.

We have this problem with other extensions and need some elegant way to solve it. I'm open for suggestions here.

thomaspaulmann:
@jberglinds left a few comments. Sorry that it took me longer than expected.

extensions/github-repository-search/tsconfig.json
jberglinds:
```
src/index.tsx:74:34 - error TS2550: Property 'replaceAll' does not exist on type 'string'. Do you need to change your target library? Try changing the 'lib' compiler option to 'es2021' or later.
```

extensions/github-repository-search/src/index.tsx
jberglinds: Why?
The second filter just targets the owner part of the repository id. Even if you get a hit with the first filter on something I think it should still filter the results by whatever you have selected in the dropdown

extensions/github-repository-search/src/index.tsx
jberglinds: Sure it is.
I renamed it because I didn't think 'viewer' was a great name anymore when you detach it from the context of the request, but I guess you still have it in the type information

extensions/github-repository-search/src/useRepositories.ts
jberglinds: Sure, can do it

jberglinds:
I believed I've addressed most of your comments. 
Regarding the error messages I'm not sure what to do about it. I think the error message it pretty clear about whats wrong.

thomaspaulmann:
It isn't about that the error message isn't clear. It's about that existing users see this message from one day to another. This causes confusion.

I have an idea to be smarter:
- Show a one-time view to explain the new features and mention that you need a token with the expected scopes
- Build the API requests in a more robust way
  - Essentially, if the user doesn't have user scope, don't search private repos. And if the user's scope doesn't have access to orgs, don't search them.
  - The dropdown would then be optional.

Benefit is that existing users can keep using the extension as is and don't need to change the scope of the token. Happy to help implementing this. WDYT?

jberglinds:
I've had absolutely no time lately to look at this. Please help out if you can find the time 🙏

thomaspaulmann:
I started to work on the changes in [this branch](https://github.com/raycast/extensions/tree/pr/2266). For now I upgraded to the latest APIs. Next, I'll convert the history to use our utils and then I'll look into the graceful handling of the tokens.

thomaspaulmann:
@jberglinds I found the time to upgrade the extension. I created a new PR for it, it was easier after all the additional changes. I'm closing this PR for #2901. Please give it a try and let me know if there is anything else that you would need.

PR #2270, opened 2022-07-14
Create an extension to search dbt documentation

b-per:
## Description
This is a new extension to support searching the docs for dbt on https://docs.getdbt.com/
This is similar to the [Vue.js documentation extension](https://www.raycast.com/DavidPascaud/vuejs-documentation) and leverages most of the code from [Search with Algolia](https://www.raycast.com/valentindotxyz/search-with-algolia).

I tried to use Search with Algolia to query the dbt docs but it didn't work as its current implementation doesn't allow nested fields in the configuration.

In relation to capitilazition rules for commands and extensions, dbt should always be lower case.

## Screencast
dbt-documentation-1

## Checklist


danielsequeira:
Hi @b-per 👋 

Thank you so much for your contribution.

I noticed that you are missing a `README.md` file though, could you please add one?

b-per:
Done 😄 

According to the docs I thought the README was only required for extensions that have parameters or require previous setup (which is not the case for this one).

danielsequeira:
Sorry @b-per you are totally right, I missed the fact that was not the case with this extension 🤦

extensions/dbt-documentation/CHANGELOG.md
```diff
@@ -0,0 +1,3 @@
+# dbt-documentation Changelog
+
+## [Initial Version] - 2022-07-14
```
danielsequeira, line 3:
```suggestion
## [Initial Version] - 2022-07-15
```

extensions/dbt-documentation/package.json
```diff
@@ -0,0 +1,39 @@
+{
```
danielsequeira, line 1: nitpicking here, but the command together with the extension name look a bit verbose. I would suggest maybe call the extension title "dbt" and then the command title "Search Documentation" with the subtitle "dbt".
Raycast 2022-07-15 at 12 40 41

extensions/dbt-documentation/package.json
```diff
@@ -0,0 +1,39 @@
...
+  "title": "dbt Documentation Search",
+  "description": "Search the dbt documentation website https://docs.getdbt.com/",
+  "icon": "dbt-icon.png",
+  "author": "b-per",
+  "license": "MIT",
+  "categories": ["Documentation"],
+  "commands": [
+    {
+      "name": "index",
+      "title": "Search dbt Documentation",
```
danielsequeira, line 13: nitpicking here, but the command together with the extension name look a bit verbose. I would suggest maybe call the extension title "dbt" and then the command title "Search Documentation" with the subtitle "dbt".

extensions/dbt-documentation/src/index.tsx
```diff
@@ -0,0 +1,62 @@
+import { ActionPanel, getPreferenceValues, List, Action, showToast, Toast } from "@raycast/api";
+import { useEffect, useMemo, useState } from "react";
+import algoliasearch from "algoliasearch/lite";
+
+const APPID = "57ZWAOQC7F";
+const APIKEY = "04a5092253f5120fdff2c77b3847d0e1";
```
danielsequeira, line 6: why didn't you consider adding this as a preference? I guess this is not an account or user related API key?

extensions/dbt-documentation/CHANGELOG.md
b-per: This is updated now

extensions/dbt-documentation/package.json
b-per: Completely agree, the suggestion on the command title/subtitle is great and I just changed it.

In regards to the extension name, there is [already one](https://www.raycast.com/zsombor-flds/dbtcloud) related to dbt but focused on the commercial product dbt Cloud, so I thought it'd still make sense to mention that this is about the docs search for dbt.

extensions/dbt-documentation/src/index.tsx
b-per: These are the API creds to connect to Algolia for it to return results for the dbt documentation. This is not something that the users need to know/fill-in themselves.

This is the same approach used by the vuejs-documentation extension:
https://github.com/raycast/extensions/blob/0a576d1ce4a411840d8ff1765c0ab3d67d999708/extensions/vuejs-documentation/src/index.tsx#L8-L10

extensions/dbt-documentation/package.json
danielsequeira: Ah ok, I didn't check. In that case I guess "dbt Documentation" as the name is fine.

danielsequeira (approved):

danielsequeira:
Thanks @b-per, before I merge I just noticed that with the name changes your screenshot is outdated. Could you please upload a new one?

b-per:
Sure! Done.

danielsequeira (approved):

PR #2278, opened 2022-07-15
Supernotes: New Browser Shorcuts, Improved Creation Form

pomdtr:
## Description

- Add shortcuts for home, daily, thoughts, inbox, and junk views
- Allow open cards in the browser from the search and create commands
- Default to showing an HUD on new card creation, add shortcut to open the preview
- add a confirmation alert on card deletion

## Screencast

image

## Checklist

### DIFF
--- a/extensions/supernotes/package.json
+++ b/extensions/supernotes/package.json
@@ -29,35 +29,35 @@
     },
     {
       "name": "home",
-      "title": "Open Home",
+      "title": "Go Home",
       "subtitle": "Supernotes",
       "description": "Open Home View.",
       "mode": "no-view"
     },
     {
       "name": "daily",
-      "title": "Open Daily",
+      "title": "Go to Daily",
       "subtitle": "Supernotes",
       "description": "Open Daily Collection.",
       "mode": "no-view"
     },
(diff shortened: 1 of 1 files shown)


acnebs:
Hi @pomdtr, awesome work. Couple initial thoughts: 

- Might make more sense to have a "go to"-style command rather than having a separate command for each view. For example, the command could be called "Open View" which takes you to a list of all the primary views. On top of this, we will be adding a URL scheme (e.g. `supernotes://`) to the Supernotes desktop application in the next version, so it probably makes sense to wait for that so that users can open views directly in the desktop app rather than just on the webapp.
- You have inadvertently removed the recents cards from the search command by adding an empty list view when there is no search.

Finally, we actually have a separate repo for this extension's development [here](https://github.com/supernotes/raycast-extension). We think it would be better if PR's were initially made against that repo and then when changes are ready to go into the Raycast store we can make a PR from that repo to here. That way we can keep discussion about what features we think are best (for Supernotes specifically) and other distracting things out of the Raycast repo.

Thanks again for working on this and look forward to going through your PR over there!

PR #2286, opened 2022-07-16
Gopass: Add Show Details Action

5mutawa:
## Description
Add “show details” action to display the secret contents

Displays the secrets content in a new view as pairs of key and value. Each content item has the action options of "Copy to Clipboard" or "Paste to Active App”.

## Screencast

image
image

## Checklist

### DIFF
--- a/extensions/gopass/CHANGELOG.md
+++ b/extensions/gopass/CHANGELOG.md
@@ -1,6 +1,6 @@
 # pass Changelog
 
-## [New Additions] - 2022-07-16
+## [New Additions] - 2022-07-19
 
 - Add action to display the content of a secret.
 - Add action to copy or paste values from the selected secret content.


bake:
Hi @5mutawa, thank you for the addition!

I actually had a similar details view built previously but removed it before publishing. The problem when displaying information stored in gopass within Raycast is that the GPG key has to be already loaded. Someone might have set GPG Suite to not remember the key at all. When they try to open the details view, Pinetry Mac will open and close Raycast.

We could probably tell the users to set GPG Suite to remember the key for at least one minute, check if it is loaded and ask for the password within Raycast. Do you have an idea on how to implement something like this?

Another solution would be to ask @danielsequeira or @pernielsentikaer if it is possible to keep Raycast open when on a specific view.

5mutawa:
@bake Yes good point, I have GPG set to not to remember the passphrase myself. When you click on Pinetry Mac it will close Raycast, but after you enter your key and reopen raycast it will be on the details view.

The current implementation in this PR works just fine for me. It would be nice to keep Raycast open while Pinetry Mac is in focus! Not sure of a way to achieve that just yet.

I doubt there is a secure way to ask for the GPG passphrase within Raycast. I will look into it and let you know.

extensions/gopass/src/details.tsx
```diff
@@ -0,0 +1,58 @@
...
+      })
+      .finally(() => setLoading(false));
+  }, []);
+
+  return (
+    <List isLoading={loading}>
+      {details.map((item, index) => {
+        const [key, value] = item.split(": ");
+        const copyAction = () => (index === 0 ? copyPassword(entry) : copy(key, value));
+        const pasteAction = () => (index === 0 ? pastePassword(entry) : paste(key, value));
```
bake, line 40: Do you think this would be more clear if you'd create the first `List.Item` manually outside the map? We could get rid of these the `concat` in line 26 and these two actions. We also wouldn't have to compare to an index.

extensions/gopass/src/details.tsx
```diff
@@ -0,0 +1,58 @@
+import { Action, ActionPanel, Clipboard, closeMainWindow, Icon, List, showHUD, showToast, Toast } from "@raycast/api";
+import { useEffect, useState } from "react";
+import gopass from "./gopass";
+import { humanize } from "./utils";
+import { copyPassword, pastePassword } from "./index";
+
+async function copy(key: string, value: string): Promise<void> {
+  Clipboard.copy(value);
```
bake, line 8: I don't know if it is important but you're using `await` on `Clipboard.paste`.

extensions/gopass/src/index.tsx
```diff
@@ -1,8 +1,9 @@
 import { Action, ActionPanel, Clipboard, closeMainWindow, Icon, List, showHUD, showToast, Toast } from "@raycast/api";
 import { useEffect, useState } from "react";
 import gopass from "./gopass";
+import Details from "./details";
 
-async function copyPassword(entry: string): Promise<void> {
+export async function copyPassword(entry: string): Promise<void> {
```
bake, line 6: Since you already created a `utils.ts`, I think it would be a good fit for `copyPassword` and `pastePassword`. That would also clean up `index.ts` a little bit. What do you think?

bake (changes requested):

danielsequeira:
@bake @5mutawa unfortunately to keep raycast window open while other apps are in focus

extensions/gopass/src/details.tsx
5mutawa: Yes, you are right. I just forgot about it.

extensions/gopass/src/details.tsx
```diff
@@ -0,0 +1,65 @@
...
+      {!loading && (
+        <List.Item
+          title="Password"
+          subtitle="*****************"
+          actions={<Actions copy={() => copyPassword(entry)} paste={() => pastePassword(entry)}></Actions>}
+        />
+      )}
+
+      {details.map((item, index) => {
+        const [key, value] = item.split(": ");
```
bake, line 52: I have several secrets containing `: `. What do you think of doing something like this?

```js
const [key, ...values] = item.split(": ");
const value = values.join(": ");
```

bake (approved):

bake:
@5mutawa Thank you so much!

PR #2305, opened 2022-07-19
Live IRC chat for Twitch.tv

Aayush9029:
## Description

View live chat for any channel on [twitch.tv](https://www.twitch.tv)

- I've stressed tested with 100k+ viewer streams and it has worked without any issues 👍🏼 

## Screencast
CleanShot 2022-07-19 at 16 00 43

## Checklist

### DIFF
--- a/extensions/twitch-chat/src/index.tsx
+++ b/extensions/twitch-chat/src/index.tsx
@@ -35,7 +35,7 @@ export default function Command() {
     <List navigationTitle={`${getPreferenceValues().streamName}'s Chat`} searchBarPlaceholder="Filter users">
       <List.Item
         icon={Icon.Message}
-        title="Open chat in browser"
+        title="Open Chat in Browser"
         subtitle={`twitch.tv/${getPreferenceValues().streamName}`}
         actions={
           <ActionPanel>


danielsequeira:
@Aayush9029 is the idea of this extension just to be able to see the chat messages? I tried it out and because of the constant stream of messages, it is not possible to keep the action panel open.

Aayush9029:
Yes the idea is just to glance at twitch chat for a specific channel!

danielsequeira:
@Aayush9029 I would suggest to at least add an action to open the chat in the browser and it I think it would look better to add a twitch.tv logo to this extension

Aayush9029:
@danielsequeira lmk if I need to change anything else :) 

> because of the constant stream of messages, it is not possible to keep the action panel open.

I ended up adding a list item on top which fixes this

danielsequeira:
> @danielsequeira lmk if I need to change anything else :)
> 
> > because of the constant stream of messages, it is not possible to keep the action panel open.
> 
> I ended up adding a list item on top which fixes this

@Aayush9029 couldn't you just add the action to the action panel, without the list item?

Aayush9029:
> couldn't you just add the action to the action panel, without the list item?

It seems to refresh whenever a new list item is introduced, unless I'm implementing it in a weird way (which I probably am). Adding the action to the parent didn't work either. Is there an example you could point me to?

danielsequeira:
> > couldn't you just add the action to the action panel, without the list item?
> 
> It seems to refresh whenever a new list item is introduced, unless I'm implementing it in a weird way (which I probably am). Adding the action to the parent didn't work either. Is there an example you could point me to?

I guess that is the same issue that I previously mentioned of not being able to keep the action panel open, after pressing cmd+K.

Could you share the code you are using for this?

danielsequeira:
@Aayush9029 I think your solution with adding the list item at the top is indeed the best. I tried to add the action to every list item of the chat, but because they are constantly being added, it will result in the action flickering, which is worse. The main thing is that the List component was not intended to be used like this, as a stream.

extensions/twitch-chat/src/index.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+  return (
+    <List navigationTitle={`${getPreferenceValues().streamName}'s Chat`} searchBarPlaceholder="Filter users">
+      <List.Item
+        icon={Icon.Message}
+        title="Open chat in browser"
+        subtitle={`twitch.tv/${getPreferenceValues().streamName}`}
+        actions={
+          <ActionPanel>
+            <Action title="Open chat in browser" onAction={openPopupChat} icon={Icon.Message} />
```
danielsequeira, line 42:
```suggestion
            <Action title="Open Chat in Browser" onAction={openPopupChat} icon={Icon.Message} />
```

extensions/twitch-chat/src/index.tsx
```diff
@@ -0,0 +1,66 @@
...
+    popToRoot({ clearSearchBar: true });
+    await open(`https://www.twitch.tv/popout/${getPreferenceValues().streamName}/chat`);
+    client.disconnect();
+  };
+
+  return (
+    <List navigationTitle={`${getPreferenceValues().streamName}'s Chat`} searchBarPlaceholder="Filter users">
+      <List.Item
+        icon={Icon.Message}
+        title="Open chat in browser"
```
danielsequeira, line 38:
```suggestion
        title="Open Chat in Browser"
```

danielsequeira:
Anyway if you go ahead and apply my last two nitpicking comments, I can go ahead and merge it 🙂

Aayush9029:
> List component was not intended to be used like this, as a stream

Absolutely! It's fun to push Raycast and try new and weird things! Thanks a lot for being super helpful!! 😄

danielsequeira (approved):

PR #2334, opened 2022-07-21
[Todo List] - Add menu bar support

ben-greenwood:
## Description
Adds menubar support for Todo List

## Screencast
https://share.cleanshot.com/NyOlJy

## Checklist


extensions/todo-list/src/menu_bar_todo_item.tsx
```diff
@@ -0,0 +1,18 @@
...
+  idx,
+  sectionKey,
+}: {
+  item: TodoItem;
+  idx: number;
+  sectionKey: keyof TodoSections;
+}) {
+  const { toggleTodo } = useTodo({ item, idx, sectionKey });
+
+  return <MenuBarExtra.Item title={`${item.completed ? "✓" : "○"}  ${item.title}`} onAction={() => toggleTodo()} />;
```
sxn, line 17: Right now, Raycast isn't guaranteed to wait for a render once an action handler has been executed. Because of this, the list is only re-rendered once you invoke the command from the root search or when you click the item again in the menu bar. We're currently looking into making sure we always perform a render after an action handler, but in the mean time you should be able to work around it by adding an `await setTimeout(1)` in your action handler… 😅

```
import { setTimeout } from "timers/promises";

…

return (
  <MenuBarExtra.Item
    title={`${item.completed ? "✓" : "○"}  ${item.title}`}
    onAction={async () => {
      toggleTodo();
      await setTimeout(1);
    }}
  />
);
``` 

Without await:
https://user-images.githubusercontent.com/1155589/183705350-4fd2e85f-fed5-45a3-ad02-ef807d555916.mov

With await:
https://user-images.githubusercontent.com/1155589/183705901-3f73546a-7638-4620-b095-30ec0df1cc1a.mov

extensions/todo-list/src/menu_bar.tsx
```diff
@@ -0,0 +1,31 @@
...
+import { MenuBarExtra } from "@raycast/api";
+import MenuBarTodoItem from "./menu_bar_todo_item";
+import { SECTIONS_DATA } from "./config";
+import { useAtom } from "jotai";
+
+export default function MenuBar() {
+  const [todoSections] = useAtom(todoAtom);
+
+  return (
+    <MenuBarExtra icon="../assets/command-icon-menubar.png" tooltip="Your Pull Requests">
```
sxn, line 12: if you have no todos, `MenuBarExtra` will have no children, so it won't have a menu, so clicking it will appear to do nothing, which could be a bit confusing. To avoid that awkwardness, I'd suggest adding a placeholder Item when there are no todos.

sxn:
hi @ben-greenwood, thanks for the changes! Overall looks good - I left a couple of comments that should help improve UX a bit. 😸 
Oh and there seems to be a conflict w/ `package.json`, mind fixing it?

sxn:
🎉

PR #2342, opened 2022-07-22
Add DeepL extension

pindab0ter:
## Description

This extension allows you to translate text using the DeepL API. A [similar extension already exists](https://www.raycast.com/mooxl/deepcast), but I found the user experience lacking. The differences between their and my approach are fundamental enough that I felt a separate extension was warranted.

## Screencast

https://user-images.githubusercontent.com/5128166/180440783-528fdc73-8c4c-49b5-b981-e93bfcca2144.mp4

## Checklist

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -75,7 +75,7 @@
 /extensions/cloudflare                               @Destiner @niklaswa
 /extensions/coda-bookmarks-search                    @allienx
 /extensions/code-stash                               @ben-greenwood
-/extensions/coffee                                   @mooxl @bibixx
+/extensions/coffee                                   @mooxl @bibixx @GastroGeek
 /extensions/coinbase-pro                             @farisaziz12
 /extensions/coingecko                                @Destiner @mike182uk
 /extensions/coinmarketcap-crypto-crawler             @moojing
--- a/.github/raycast2github.json
+++ b/.github/raycast2github.json
@@ -360,5 +360,6 @@
   "acnebs": "acnebs",
   "b-per": "b-per",
   "devmoath": "devmoath",
-  "stuart": "stuart"
+  "stuart": "stuart",
+  "GastroGeek": "GastroGeek"
 }


pindab0ter:
The build action fails with 

```error  Invalid author "pindab0ter". Error Body: {"error": {"code": "502", "message": "An error occurred with this application."}} (RequestID: )```

The author name is checked against this regex: `^[a-zA-Z0-9-*~][a-zA-Z0-9-*._~]*$` ([source](https://www.raycast.com/schemas/extension.json)), but that should be fine.
It is also the same as my Raycast account name: [https://www.raycast.com/pindab0ter](https://www.raycast.com/pindab0ter), which is also a requirement.

Any idea what the issue could be?

kidonng:
Hi @pindab0ter, I love this, much more intuitive!

Just a suggestion, instead of having two DeepL extension that do things differently, maybe this flow can be integrated into the existing Deepcast extension, possibly under a preference switch? It could become the default if you ask the author.

I'm asking this because I think it's also neat that Deepcast can translate selected text directly, and have a command for a more "familiar" interface for switching between languages:

image

I would like to see both, without cluttering the menu with similarly named commands.

pindab0ter:
> Hi @pindab0ter, I love this, much more intuitive!
> 
> Just a suggestion, instead of having two DeepL extensions that do things differently, maybe this flow can be integrated into the existing Deepcast extension, possibly under a preference switch? It could become the default if you ask the author.
> 
> I'm asking this because I think it's also neat that Deepcast can translate selected text directly, and have a command for a more "familiar" interface for switching between languages:
> 
> image
> 
> I would like to see both, without cluttering the menu with similarly named commands.

> Hi @pindab0ter, I love this, much more intuitive!

Thanks! Glad to hear you like it!

> it's also neat that Deepcast can translate selected text directly

Would this be solved by reading what text is currently in your pasteboard?

What DeepCast currently does is read what is in your pasteboard, translate that and replace that text in your pasteboard. I really dislike that approach because it's opaque and there's no way to check when the translation is. DeepL is _really_ good, but not perfect.

My alternative would be to enter the selected text into the search bar by default, but that comes with the risk of sensitive information ending up on your screen by accident, so if I were to do that I would make that a preference like you said.

My question to you is: how is this then better than just pasting your text into the search bar?
(This is not a rhetoric question, I would like to hear your take.)

> and have a command for a more "familiar" interface for switching between languages:

What does this interface offer in terms of functionality, other than it being more familiar?

Finally, there's two reasons that I chose to build my own extension over seeking collaboration in this case:

1. I feel that DeepCast's interface is opaque (replacing what's on your clipboard) and cumbersome (the screenshot you sent), but I can also see that people value those. Their and my approach are mutually exclusive.
2. I strongly prefer extensions named for what they do. I find "DeepCast" to be a poor name since it is also the name of an [AI platform](https://www.deepcast.ai/) and a [music website](https://www.melbournedeepcast.net/), an [AI assisted 'crowdcast' technique](https://ieeexplore.ieee.org/document/9047133) (whatever that is), a [Kansas based recruitment agency](https://www.facebook.com/DeepCast-146403245402232/), and probably a few more things. But none of them have anything to do with DeepL. That's why I chose to name my extension simply "DeepL", because that's what I want to show up in my UI.

pindab0ter:
I did a big rewrite into using `useFetch` instead of `node-fetch`, but I'm holding off on that until #2392 is fixed.

Also, I regret not using a feature branch for this PR. Is there any way I can update this PR to point to the feature branch instead? Or is that just tough luck now?

kidonng:
> My question to you is: how is this then better than just pasting your text into the search bar and using the copy shortcut?

I have thought about that (pasting), sorry for not making it clear in my last comment. I think they are in feature parity.

> What does this interface offer in terms of functionality, other than it being more familiar?

You can switch the "from" and "to" languages, or change any of them inside the command (instead of needing to go back to root). But I don't seem them as essential.

> 2\. I strongly prefer extensions named for what they do.

This is a good point! Searching for "deepl", Deepcast is actually shown at the second place (and I still don't know why it is named after so):

image

PR #2345, opened 2022-07-22
[gitlab] Add option to show MR details.

metakirby5:
## Description

- Add a new setting to show details in merge request list via a metadata view.
- Switch `My Reviews` to use `MRListItem` to display merge requests.
- Upgrade Raycast API and fix resulting compilation errors.

## Screencast

Setting on:

image
image

Setting off:

image

## Checklist

### DIFF
--- a/extensions/gitlab/CHANGELOG.md
+++ b/extensions/gitlab/CHANGELOG.md
@@ -1,6 +1,6 @@
 # GitLab Changelog
 
-## [MR Details] - 2022-07-27
+## [MR Details] - 2022-07-28
 
 - Add a new setting to show details in merge request list via a metadata view.
 - Switch `My Reviews` to use `MRListItem` to display merge requests.


L3K0V:
Hey @metakirby5!

I have one very small feature request!\
Could you add in this MR an action to go to epic board and issue board for a group?

https://github.com/raycast/extensions/blob/5611a310e978c70fc491495baf02da27fe4999aa/extensions/gitlab/src/components/groups.tsx#L79

Just need to add here two new entries opening a web url:

- `-/boards/` for issue board with a shortcut `cmd+b`
- `-/epic_boards/` for epic board with a shortcut `cmd+shift+e`

I'm missing this feature for some time and I believe this will be merged quicker than me starting a new PR.

Please, please 🙏

metakirby5:
@L3K0V I believe smaller pull requests are usually better. The new actions need their own feedback cycle, so I'm not comfortable including them in this unrelated change.

L3K0V:
I will try to open new one then. Thanks.

extensions/gitlab/src/cache.ts
```diff
@@ -136,7 +136,7 @@ export function useCache<T>(
   const [isLoading, setIsLoading] = useState<boolean>(true);
   const [timestamp, setTimestamp] = useState<Date>(new Date());
   const shouldForceRefetch = useRef(false);
-  const depsAll = [timestamp];
+  const depsAll: any[] = [timestamp];
```
tonka3000, line 139: Why you changed it to `any[]`?

tonka3000 (commented):
Like the change, but some thing would be good to optimize.

The description field is now very small and it is really hard to read the text in mid-size description (10 lines and above. I personally avoid to switch to the browser just for reading the text. I added a `Hide Details` (Opt + D) action to my twitter extension, which hide the detail section on demand. With that action you get at least the full vertical space to read.

I'm not quite sure if it would be a good option that the feature still would respect the `Primary Action` preference. This would open the "old" detail view when `Open in Raycast` is active which has more space to read and give a better overview.

extensions/gitlab/src/cache.ts
metakirby5: Without the change, there is actually a compilation error in the current `main` branch:

image

This is because the items that are pushed are contianed in the type `type DependencyList = ReadonlyArray<unknown>;`. Upon further inspection, it seems better to type this as `unknown[]`, so I've changed it 👀

metakirby5:
>The description field is now very small and it is really hard to read the text in mid-size description (10 lines and above. I personally avoid to switch to the browser just for reading the text. I added a Hide Details (Opt + D) action to my twitter extension, which hide the detail section on demand. With that action you get at least the full vertical space to read.
>
>I'm not quite sure if it would be a good option that the feature still would respect the Primary Action preference. This would open the "old" detail view when Open in Raycast is active which has more space to read and give a better overview.

Good suggestions! I added collapse/expand detail as the primary action, to keep parity with the original implementation. Please take another look.

image
image

tonka3000:
I would add the new hide/show details to the second section, because now it is the default action. I would also add the shortcut `Opt` + `D` to it.

metakirby5:
>I would add the new hide/show details to the second section, because now it is the default action. I would also add the shortcut Opt + D to it.

I tried it out, but IMO it doesn't make sense from a UX perspective. The content shifts left which is jarring, and it shows information that was already present in the previous view. Having two menu items, one to "expand" details and one to "show" details adds some further confusion. Furthermore, the second section has been reserved for actions relating to actual GitLab functionality rather than Raycast UI, and I think breaking that rule with this convention would be counter-intuitive.

I think it makes more sense to treat the details pane view as an alternative UI, and the `Show Details` action as analogous to `Expand/Collapse Details`. The aim is to provide an alternative interface which prioritizes quick overviews over individual details, and have it be completely opt-in. What do you think?

danielsequeira:
@tonka3000 ping here 🙂

tonka3000:
@danielsequeira thanks for the ping. I will check it in the evening.

tonka3000 (approved):

danielsequeira (approved):

PR #2351, opened 2022-07-23
Add Sourcetree extension

i906:
## Description

View, search and open repositories in Sourcetree.

## Screencast

sourcetree-2

## Checklist

### DIFF
--- a/extensions/sourcetree/src/BinNotAvailable.tsx
+++ b/extensions/sourcetree/src/BinNotAvailable.tsx
@@ -1,0 +1,25 @@
+import { Action, ActionPanel, Detail, openCommandPreferences } from "@raycast/api";
+
+interface BinNotAvailableProps {
+  bin: string;
+}
+
+export function BinNotAvailable({ bin }: BinNotAvailableProps): JSX.Element {
+  const message = [
+    "# Unable to find Sourcetree command line tools",
+    `The path for the command line tools are currently set to:`,
+    `\`${bin}\``,
+    `If you have installed Sourcetree via Homebrew, you can set the path to \`/opt/homebrew/bin/stree\` in the extension settings.`,
+  ];
+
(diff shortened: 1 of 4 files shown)


mattisssa (changes requested):
I got infinite loader here and some errors in console:

CleanShot 2022-08-01 at 16 22 23@2x

I have Sourcetree installed, what could go wrong? I guess it's the path of `bin`. 
First: would be nice to handle this and show failure toast if something goes wrong
Second: how can we improve that? I haven't change anything since I installed the Sourcetree, so it seems like the default folder is wrong in preferences 🤔

i906:
> I have Sourcetree installed, what could go wrong? I guess it's the path of `bin`. First: would be nice to handle this and show failure toast if something goes wrong

I'll update the toast to the failure state if any errors occur.

> Second: how can we improve that? I haven't change anything since I installed the Sourcetree, so it seems like the default folder is wrong in preferences 🤔

If you don't mind, can you share the full path of the `bin`, if any? I'll change the default bin path in the preference to that.

I installed Sourcetree using homebrew, so the bin file might have also been installed by homebrew.

I'll add a dialog similar to [Jetbrains extension](https://github.com/raycast/extensions/blob/c614a2dfc24aad9174e55bacc6053238781d77a7/extensions/jetbrains/src/recent.tsx#L148-L156) when the path is not available.

mattisssa:
>I installed Sourcetree using homebrew, so the bin file might have also been installed by homebrew.

I installed It from their website a long time ago 😄 

>If you don't mind, can you share the full path of the bin, if any? I'll change the default bin path in the preference to that.

`Error: Command failed: /opt/homebrew/bin/stree /Users/myname/proj/Raycast`

mattisssa (approved):

PR #2352, opened 2022-07-23
Replace `js-base64` usage with native `Buffer`

kidonng:
## Description

This is more of a "lint" change than actual functional improvements, but it does save a little bytes for everyone 😉

Node.js has a built-in [`Buffer`](https://nodejs.org/api/buffer.html) since forever, so pulling an extra `js-base64` dependency is usually unnecessary (some extensions even install it without actually using!). The only valid usage I found is in the `ray-so` extension, which uses the `encodeURI` function from `js-base64` which has no equivalent in `Buffer`.

## Checklist


mattisssa:
Hey @kidonng, could you please resolve the conflicts so I could do the review?

kidonng:
@mattisssa Hi, could you please take another look?

PR #2354, opened 2022-07-23
add arguments for searchstring to dash

AaronDavidSchneider:
## Description

1. Add argument to dash search
2. Update API to 1.38.0

In principle one could also add a second argument for a docset. I would be happy to implement this as well, but I think I would need a littlebit of help for that (I did not fully understand the docset search).

Before merging, we should update the extension metadata (screencasts, versionnumber and changelog?). I didn't do so yet.

## Screencast

grafik

## Checklist

### DIFF
--- a/extensions/dash/package.json
+++ b/extensions/dash/package.json
@@ -19,7 +19,7 @@
       "arguments": [
         {
           "name": "searchstring",
-          "placeholder": "Searchstring",
+          "placeholder": "Query",
           "type": "text",
           "required": false
         }


otakustay:
Dash may requires more than 3 arguments, e.g. I have 27 docsets installed.

Also Dash extension may have to interop with Dash to find which docsets are installed to dynamically generate the `commands` section in `package.json`, I'm not sure whether RayCast API currently supports this

AaronDavidSchneider:
> Dash may requires more than 3 arguments, e.g. I have 27 docsets installed.
> 
> Also Dash extension may have to interop with Dash to find which docsets are installed to dynamically generate the `commands` section in `package.json`, I'm not sure whether RayCast API currently supports this

Thanks for your comments. I don't think thats needed/beneficial. The user can always just use the dash extension in the normal way, ie without the arguments. The arguments should instead just be a shortcut to search either in all docsets or (if codeowners decide so) in one specific docset, in which case we would need just two arguments: (docset + searchstring)

extensions/dash/package.json
```diff
@@ -15,11 +15,19 @@
       "title": "Find in Dash",
       "subtitle": "Dash",
       "description": "Filter a Dash docset and search for the provided string in it.",
-      "mode": "view"
+      "mode": "view",
+      "arguments": [
+        {
+          "name": "searchstring",
+          "placeholder": "Searchstring",
```
djpowers, line 22:
```suggestion
          "placeholder": "Search query",
```
Breaking this up into two words would serve as a more helpful prompt. I think "query" would be more consistent with the Dash convention (as used in the "Web Search" preference pane), plus you can search in Dash for non-string terms like numbers of symbols so this would more accurately convey that.

djpowers (changes requested):
Nice work getting this implemented quickly! Perfect use case for arguments.

djpowers (approved):

RSO (approved):
Nice use of arguments! Thanks for the PR 😄

PR #2357, opened 2022-07-23
can-i-use: add release date and more options

kidonng:
## Description

This pull request adds several options to the Can I Use extension to make it more informative and useful.

## Screencast

- Adds option to show release date (default on):

  image

- Adds option to show partial support details (default off):

  image

- Adds option for brief mode:

  image

- Adds option for default query when no project is configured.

## Checklist

### DIFF
--- a/extensions/can-i-use/CHANGELOG.md
+++ b/extensions/can-i-use/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Can I Use Changelog
 
+## [Improvements] - 2022-07-30
+
+- Add option to Show release date
+- Add option to Show partial support details
+- Add option for Brief Mode
+- Add option for Default Query
+
 ## [Update dependencies] - 2022-04-06
 
 - Update Raycast and Can I Use packages


thomaslombart (commented):
Thanks for the contribution. These are great improvements! 💪 

I have two feedback regarding the UX:
- Shouldn't the "Show partial support" preference be enabled by default? It doesn't seem consistent with the "Show release date" preference. The month could also be shortened in brief mode (e.g Oct 26, 2009 instead of October 26, 2009)
- In the brief mode, I believe the version number should be prefixed by a `v` to make it clear we're talking about the version (e.g `v3` instead of `3`).

kidonng:
Glad you like it! Regarding your feedbacks:

> * Shouldn't the "Show partial support" preference be enabled by default?

I didn't enable it by default because with the current code, only the latest version of a browser is considered, so the "partial support" version number is not very accurate, and I want to encourage concerning users to visit Can I use to see details.

For example, regarding the [AAC audio feature](https://caniuse.com/aac), Firefox has "partial support" since forever:

image

But currently the extension only shows status for the latest (development!) version:

image

I plan to change this behavior and then we can show partial support by default.

> * In the brief mode, I believe the version number should be prefixed by a `v` to make it clear we're talking about the version (e.g `v3` instead of `3`).

I don't think prefixing version numbers with "v" is a norm now, especially for browsers (Safari v15, anyone?). If we want to be serious, Chrome/Chromium actually prefix version numbers with "M":

image

My point is that since brief mode is not enabled by default, the number can't be taken as other meaning.

> The month could also be shortened (e.g Oct 26, 2009 instead of October 26, 2009)

Good idea! I've implemented this.

thomaslombart:
> But currently the extension only shows status for the latest (development!) 

Hum, indeed. This looks like a bug to fix 🙂 

> I don't think prefixing version numbers with "v" is a norm now, especially for browsers (Safari v15, anyone?). If we want to be serious, Chrome/Chromium actually prefix version numbers with "M":

Fair enough, that makes sense! In a way, users have to opt-in to brief mode so they'll likely be familiar with the extension's UI.

thomaslombart (approved):

pernielsentikaer:
Is this ready to be merged @kidonng since this is still marked as a draft? Maybe you can add a CHANGELOG entry to the extension too.

Thanks @thomaslombart for the review 🔥

kidonng:
@pernielsentikaer Thanks for the nudge, I will implement the remaining changes in the future, so let's not block this.

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #2372, opened 2022-07-24
[Spotify-Player] Add menubar title length limit

tonka3000:
## Description

Add an optional menubar length limit. In cases where the display is not that big it is helpful to limit the length of the menubar item. The max size is fully optional. I also added a tooltip for the whole title.

`<= 0` will also create no limitation. But maybe setting it to `0` would make the title in general fully optional for people who wanna just see it on hover.

### Long Title
CleanShot 2022-07-24 at 16 54 14@2x

### Shorter Title
CleanShot 2022-07-24 at 18 42 38@2x

CleanShot 2022-07-24 at 18 48 47@2x

## Checklist

### DIFF
--- a/extensions/spotify-player/CHANGELOG.md
+++ b/extensions/spotify-player/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Spotify Player Changelog
 
+## [Feature] - 2022-07-24
+
+- Add max. length preference for menubar item
+
 ## [Feature] - 2022-07-21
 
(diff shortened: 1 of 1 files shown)


mattisssa:
Hah, I had exactly the same suggestion from @sxn for the case but decided not include it into the released version. I'll take a look a bit later for the PR, thanks 👍

mattisssa (approved):

Issue #2329, opened 2022-07-21
[Coffee] No longer caffeinating


XT7x8QfPvupZab8RwbVU:
# Extension – [Coffee](https://github.com/raycast/extensions/tree/a492802a9ab60a9f01d3f696720f441864f4bca2/extensions/coffee/)
Author: @mooxl

## Description
Running coffee/caffeinate in raycast no longer works since upgrade to 1.38.0. It says caffeinated but then running caffeination status it says decaffeinated.

## Steps To Reproduce

1. Upgrade to 1.38.0
2. Run caffeinate using raycast. 

## The current behavior
Caffeination status shows decaffeinated.

## The expected behavior
Caffeination status should decaffeinated.

### Raycast version
Version: 1.38.0

PR #2388, opened 2022-07-26
check that args is a string before using - else raycast may pass empt…

GastroGeek:
…y arguments object

## Description

```
git commit -m "check that args is a string before using - else raycast may pass empty arguments object"
```

## Checklist

### DIFF
--- a/extensions/coffee/package.json
+++ b/extensions/coffee/package.json
@@ -6,7 +6,8 @@
   "icon": "logo.png",
   "author": "mooxl",
   "contributors": [
-    "bibixx"
+    "bibixx",
+    "GastroGeek"
   ],
   "license": "MIT",
   "commands": [


mooxl:
LGTM and fixes #2329

GastroGeek:
> 

forgive my potential ignorance but do you mean in the Caffeine() function? something like

```
const Caffeinate = async (args: "") => 
```

? only that won't help I don't think since the problem is when you execute just `Caffeinate` Raycast sends `{ arguments: {} }` which overrides that default... so same problem?

extensions/coffee/src/caffeinate.ts
```diff
@@ -2,11 +2,13 @@ import { showHUD } from "@raycast/api";
...
 
 const Caffeinate = async (args?: string) => {
+  const validStringCaffeinateArgs = typeof args === "string";
+
   try {
     await runAppleScript(`do shell script "pgrep caffeinate"`);
     await showHUD("Your Mac is already caffeinated");
   } catch (_) {
-    runAppleScript(`do shell script "caffeinate -di${args ? ` ${args}` : ""}"`);
+    runAppleScript(`do shell script "caffeinate -di${validStringCaffeinateArgs ? ` ${args}` : ""}"`);
```
mooxl, line 11: just a nit: `${typeof args === "string" ? ...}` would save one line:)

mooxl (approved):

extensions/coffee/src/caffeinate.ts
GastroGeek: should be on PR now 😅 that's the hardest _simple_ fix, I've ever done heh.

mooxl: if you want, you could also add yourself to the contributors in the package.json with that one liner :) @GastroGeek

GastroGeek: why not, eh? 😄 done! thanks!

mattisssa (approved):
Let's get it merge until we introduce the fix in the next release. 🚀 
Sorry for inconvenience

PR #2396, opened 2022-07-27
Add DeepL Extension

pindab0ter:
## Description

This extension allows you to translate text using the DeepL API. A [similar extension already exists](https://www.raycast.com/mooxl/deepcast), but I found the user experience lacking. The differences between their and my approach are fundamental enough that I felt a separate extension was warranted.

For more information and some discussion, please see my previous PR #2342.

## Screencast

https://user-images.githubusercontent.com/5128166/180440783-528fdc73-8c4c-49b5-b981-e93bfcca2144.mp4

## Checklist

### DIFF
--- a/extensions/deepl/src/Command.tsx
+++ b/extensions/deepl/src/Command.tsx
@@ -6,20 +6,25 @@ import SourceLanguageDropdown from "./components/SourceLanguageDropdown";
 
 export default function Command(targetLanguage: Language): () => JSX.Element {
   return () => {
-    const [text, setText] = useState("");
+    const [searchText, setSearchText] = useState("");
     const [sourceLanguage, setSourceLanguage] = useState<Language | undefined>(undefined);
-    const { isLoading: isLoadingTranslation, data: translation } = useTranslation(text, sourceLanguage, targetLanguage);
+    const { isLoading: isLoadingTranslation, data: translation } = useTranslation(
+      searchText,
+      sourceLanguage,
+      targetLanguage
+    );
     const { isLoading: isLoadingUsage, data: usage, revalidate: revalidateUsage } = useUsage();
     const isLoading = isLoadingUsage || isLoadingTranslation;
-    const hasInput = text.length > 0;
-    const hasTranslation = hasInput && translation !== undefined;
+    const hasInput = searchText.length > 0;
+    const hasTranslation = translation !== undefined;
 
     useEffect(() => revalidateUsage(), [translation]);
 
     return (
       <List
         isLoading={isLoading && hasInput}
-        onSearchTextChange={setText}
+        searchText={searchText}
+        onSearchTextChange={setSearchText}
         searchBarPlaceholder={`Translate to ${targetLanguage.name} using DeepL…`}
         searchBarAccessory={
           <SourceLanguageDropdown sourceLanguages={sourceLanguages} onSourceLanguageChange={setSourceLanguage} />


danielsequeira:
Hi @pindab0ter,

Indeed we really try to not have duplicated extensions in our store and this one is very similar to Deepcast, as you mentioned.

What would you say are "fundamental differences" between this extension and the existing one? And why can't you make an improvement to the existing one instead of creating a new one?

pindab0ter:
The fundamental difference is:

The DeepCast extension takes what you have currently selected, and replaces the contents of your clipboard on command invocation. In my personal opinion, this is very poor UX. There is no way to check either your input or your output.

My extension shows a search bar that allows you to edit source translations on the go.

This, in my opinion, is an absolute must. Since DeepL's translations are good, but not so good as to never require adjustments. My extension allows for that.

The reason I chose not to offer this as an improvement to the DeepCast extension is that current users of that extension will not be able to use that workflow any more if my approach were to be used. They are mutually exclusive.

I also think that the name 'DeepCast' is very confusing, as I elaborated on in the mentioned PR.

danielsequeira:
Hi @pindab0ter,

I have checked with the rest of the team and we really want to avoid having multiple extensions for the same application.

We do agree with you regarding some of the shortcomings of the current Deepcast extension, mainly regarding the UX and on commands usage and we also think that having one command to switch language is a lot more intuitive.

Having said that Deepcast is a heavily used extension in our store, with more than 5400 installations. And we would prefer if we could somehow get your changes as an improvement to the current extension. To do that it will probably be a good idea to contact @mooxl and collaborate with him. We can also support you in achieving this. From our perspective just because one extension behaves differently than another it is not enough reason for having multiple ones.

pindab0ter:
Hey @danielsequeira, thank you for taking the time to formulate such a measured response. I understand all your reasons.

I have a few questions:

1. Since the features my extension offers is basically a complete rewrite, how does this work? What do I even put in my PR?

2. What is your policy on backwards compatibility? If my implementation were to be accepted, this would break the workflow of the people that are used to using Deepcast as it is now. It is of course possible to have both workflows exist next to each other as separate commands, but that would require at least renaming the commands. What is your stance on this?

3. Is it possible to change an extensions' name? The name "Deepcast" bothers me, since it doesn't accurately describe what it does or what it's for, and doesn't follow the convention of using the actual name of the application you're interfacing with.

mooxl:
Hey @pindab0ter,

thanks for your input regarding the user experience at Deepcast. And thanks to @danielsequeira for mentioning me, as I've only now seen that there are approaches to improving Deepcast.

I have so far used the persistent settings of the classic `Translate` command to copy my text and then translate it. But I understand your reasoning and like your suggestion so we can rewrite the direct translation to meet your needs. 

As we don't want to disrupt the workflow of existing users, I would suggest that we add manual text input to direct translation as an option in the welcoming screen where you also insert your key.
If that's okay with you, feel free to port your code to Deepcast and I'll post the reviews as soon as possible. If you want, I can also resolve my issues afterwards:)

I don't know if it is possible to rename the extension. Since this extension was one of the first in the store and I have a penchant for word games, it was fitting at the time, but in retrospect it would be nicer from my side if we renamed Deepcast to DeepL.

danielsequeira:
Changing the title of the extension should be fine and not have any side effect

pindab0ter:
> Changing the title of the extension should be fine and not have any side effect

That’s awesome!

mooxl:
Hey @pindab0ter , 

I've been working on porting your code today, as discussed in Slack. 
After working on it for a while, I noticed that we get a redundancy of `Translate` and `Translate Into` due to the changes. Now I have reinterpreted the existing concept of Deepcast and your suggestion and want to know what you think of it.

I have already told you before that I don't want to have an automatic translation, especially in free plan, because otherwise you reach the letter limit too quickly in my opinion. 

That's why I came to the conclusion that it's only necessary to upgrade the existing `Translate` command a little - there really is no need  a `Translate Into` command any more.

Namely:

- I would like to continue to use the existing interface with `Form`, omitting last the textarea with the translated text.
- After pressing `Cmd + Enter` to translate, you get to the overview you suggested, where you can also see your current consumption.
- The target language becomes persistent and the source defaults to autodetect.
- You can mark up to three different languages as favourites in the settings and use `Control + C` to switch the target language with them.
- There would be an option in the preferences to autoinsert from and autocopy to the clipboard.

pindab0ter:
> I have already told you before that I don't want to have an automatic translation, especially in free plan, because otherwise you reach the letter limit too quickly in my opinion.

I don't remember discussing this. This is actually one of the main reasons I started developing my own version. My use case is this: I often have a word, phrase or, a part of a sentence that I need to translate. These often require rewording to get good results. This is the reason that I want as little friction as possible.

Of course, this doesn't eat into your free tier API usage as much as other use cases might.

> That's why I came to the conclusion that it's only necessary to upgrade the existing `Translate` command a little - there really is no need a `Translate Into` command any more.

Would this not prevent the binding of aliases? That way, you will not be able to use an alias to translate to a specific language, which is not an unusual use case.

> * I would like to continue to use the existing interface with `Form`, omitting last the textarea with the translated text.
> * After pressing `Cmd + Enter` to translate, you get to the overview you suggested, where you can also see your current consumption.
> * The target language becomes persistent and the source defaults to autodetect.
> * You can mark up to three different languages as favourites in the settings and use `Control + C` to switch the target language with them.
> * There would be an option in the preferences to autoinsert from and autocopy to the clipboard.

The main reason I created my own version of a DeepL extension is because I think that the existing “Translate into [Language]” commands are obtuse from a UX perspective. It is unclear what they do as they mix UX contexts. You have to first select text, then launch Raycast, use the required command, and then it uses the text you selected in an app that is now no longer in the foreground or even visible any more. Most importantly; they offer no way to view or adjust the translation result.

I do think, however, that there is a solution that accommodates most—if not all—use cases!

* Use the form for all ‘Translate into…’ commands
* It automatically inserts selected text or text in the pasteboard if enabled in the preferences as you suggested
* The source language defaults to ‘Autodetect’
* The target language is still a dropdown, but defaults to whatever the chosen language for the command is
  (I personally don't see a scenario where you would want to change the _target_ language, but maybe people use this?)
* The usage of the following actions:
  * <kbd>⌘⏎</kbd> would translate the input; no automatic translation to prevent unnecessary API usage
  * <kbd>⌘⇧⏎</kbd> would copy the translation text and close the window, which could also be used without first translating using <kbd>⌘⏎</kbd>, in case you really don't care about evaluating the translation result first
* Showing the usage below the translation result in a [`Form.Description`](https://developers.raycast.com/api-reference/user-interface/form#form.description)
* Scrap the “Translate” command since it is no longer needed

The advantages of my proposal are:
* A better and more discoverable UX
* The ability to view the translation result and adjust if necessary
* Less friction because of less required inputs when wanting to translate to a specific language (by using the `Form`)
* The retaining of existing functionality at the cost of one single extra keystroke
* The ability to use aliases to translate to a specific target language, reducing friction in comparison to the ‘Translate’ command

What do you think?

danielsequeira:
@mooxl @pindab0ter do you still plan to work together on this?

pindab0ter:
I'm open for collaborating.

PR #2403, opened 2022-07-27
new extension: network-usage

oktaysenkan:
## Description

Display network bandwith from Raycast

## Screencast

https://user-images.githubusercontent.com/42527467/181379634-e7ac9299-1c9f-41dc-87f6-07cc2bef0586.mp4

## Checklist

### DIFF
--- a/extensions/network-usage/package-lock.json
+++ b/extensions/network-usage/package-lock.json
@@ -9,10 +9,12 @@
       "dependencies": {
         "@raycast/api": "^1.38.2",
         "millify": "^5.0.0",
-        "moment": "^2.29.4"
+        "moment": "^2.29.4",
+        "node-schedule": "^2.1.0"
       },
       "devDependencies": {
         "@types/node": "~16.10.0",
+        "@types/node-schedule": "^2.1.0",
         "@types/react": "^17.0.28",
         "@typescript-eslint/eslint-plugin": "^5.0.0",
         "@typescript-eslint/parser": "^5.0.0",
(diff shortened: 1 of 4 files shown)


oktaysenkan:
`MenuBarExtra` component need `width` or `minWidth` property to prevent changing width. I tried JavaScript's `padEnd` and `padStart` functions but they didn't work.

sxn:
hi @oktaysenkan - thanks for building the extension!
Regarding `width` or `minWidth`: there's unfortunately no way to do this right now; I'll add it to the feature requests list :)
In the mean time, you could get away with using [NumberFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat) and using something like `Intl.NumberFormat("en-IN", { maximumSignificantDigits: 4, minimumSignificantDigits: 4 }).format(speed)`. You'll still see _some_ jumping, as the text is not monospaced, but it should somewhat alleviate it. Alternatively, you could not make the menu bar item dynamic, but only the menu items… 🤔 

Now, regarding the extension - I've got a few other remarks:
- when it's run, the extension checks whether `vnstat` is available and, if it's not, runs `brew install vnstat`. This is not ideal because, first, it assumes the user has `brew` installed - crashing if that's not the case - and second, because it uses the user's package manager to install software without the user's permission. Generally, extensions' `view` commands communicate missing dependencies using something like [this](https://github.com/raycast/extensions/blob/main/extensions/bunch/src/components/bunch-not-install-view.tsx). Since in this case the command is a `menu-bar` command, we don't really have a "standard" way of doing this yet. However, you _could_ try adding a different version of the `MenuBarExtra`, explaining what the problem is, explain _why_ `vnstat` is needed and how it can be installed
- would you mind updating the README so that it explains what the extension does, what `vnstat` is and why it's necessary?

Thanks! Ping me if something else comes up, is unclear etc! 😺

PR #2423, opened 2022-07-29
Edit extension: Add menubar command that holds pending actionable PRs

kudrykv:
## Description

This is a menu bar extension. It shows PRs you were asked to review and PRs that have had motion after your last involvement.

## Screencast

CleanShot 2022-07-29 at 21 29 28

Once clicked, it opens a browser with a PR. Then, the extension removes the PR from the active list when possible.

## Checklist

### DIFF
--- a/extensions/github-review-requests/package.json
+++ b/extensions/github-review-requests/package.json
@@ -22,7 +22,7 @@
       "title": "Pending Pull Requests",
       "description": "List actionable pending pull requests",
       "mode": "menu-bar",
-      "interval": "1m"
+      "interval": "5m"
     }
   ],
   "preferences": [


kudrykv:
Is there anything I can do to help review this PR?

extensions/github-review-requests/src/actionablePullRequests.tsx
```diff
@@ -0,0 +1,50 @@
...
+const actionablePullRequests = () => {
+  const { isLoading, login, updatedPulls, recentlyVisitedPulls, visitPull } = usePulls();
+
+  const title = getTitle(updatedPulls);
+  const icon = getIcon(updatedPulls);
+
+  const showSeparator = updatedPulls.length > 0 && recentlyVisitedPulls.length > 0;
+
+  return (
+    <MenuBarExtra isLoading={isLoading} icon={icon} title={title}>
```
sxn, line 15: if both `updatedPulls` and `recentlyVisitedPulls` are empty, then I get an `All good` item in the menu bar that (seemingly) doesn't react to being clicked. This feels a bit…confusing?
What I would suggest is either:
- always add a `No updated pulls` or similar item in this case, making sure that the menu bar extra has a menu
- or make the extension return `null` instead of a `MenuBarExtra`. This will result in the nothing being placed in the menu bar in this case, which…sounds appropriate? 🤔

extensions/github-review-requests/package.json
```diff
@@ -13,6 +16,13 @@
       "subtitle": "GitHub",
       "description": "Search review requests on GitHub and listing them.",
       "mode": "view"
+    },
+    {
+      "name": "actionablePullRequests",
+      "title": "Pending Pull Requests",
+      "description": "List actionable pending pull requests",
+      "mode": "menu-bar",
+      "interval": "1m"
```
sxn, line 25: refreshing every minute feels a bit overenthusiastic 😅  what about using 5 / 10 / 15 minutes instead? 🤔

extensions/github-review-requests/src/actionablePullRequests.tsx
kudrykv: The first option sound better in this case. This would give an impression that something happens, and once visited a PR from the menu bar, there will always be a `Recently Visited` submenu item. Thank you for the suggestion!

extensions/github-review-requests/package.json
kudrykv: It would be great if this value vas configurable from settings \*sigh\*

One GraphQL call with the given query burns 3 points. 60 calls in an hour result in 180 points. Github's resource limitation is [5000 points per hour](https://docs.github.com/en/graphql/overview/resource-limitations#rate-limit). Let's round it to 200: then, the extension would consume 4% of the available quota.

I would want to get updates as soon as they are available, although I understand the reasons you might give.

I'll be bargaining here a little bit: do 3 minutes still sound too much? Otherwise, I would go with 5 minutes.

sxn:
🎉

jhogervorst:
I’m trying out this command, but it remains saying "All good", "No updated pulls at this time", while I would expect to see some PRs. I set up the token with the full `repo` scopes. When I use the "Search Review Requests" command, I do see PRs listed, so it seems that the token is working correctly.

Is there anything else I can try? I already restarted Raycast.

kudrykv:
@jhogervorst there are indeed some issues. First, please also grant `read:org` from `admin:org` privilege to your token. I don't remember, but the absence might break the GraphQL query.
The second issue, is that there is no way to force the refresh manually. It happens in the background once in 5 minutes. You'll need to wait up to 5 minutes to see whether there are changes

kudrykv:
@jhogervorst hi! Did it help?

PR #2432, opened 2022-07-30
feat: add today summary menu bar command

Fatpandac:
## Description

Add today's summary menu bar command.

## Screencast

image

image

## Checklist

### DIFF
--- a/extensions/wakatime/src/utils/api.ts
+++ b/extensions/wakatime/src/utils/api.ts
@@ -6,10 +6,13 @@ async function routeHandler<T>(endpoint: string): Promise<(T & { ok: true }) | {
   const URL = "https://wakatime.com/api/v1";
 
   try {
-    const response = await fetch(`${URL}${endpoint}`, setAuthHeader());
+    const res = await fetch(`${URL}${endpoint}`, setAuthHeader());
+    const result = (await res.json()) as T | { error: string };
+
+    if ("error" in result) throw new Error(result.error);
     return {
       ok: true,
-      ...((await response.json()) as T),
+      ...result,
     };
   } catch (error) {
     return {


extensions/wakatime/package.json
```diff
@@ -31,6 +31,14 @@
       "title": "Show Private Leaderboards",
       "description": "Show your private leaderboards",
       "mode": "view"
+    },
+    {
+      "name": "today-summary",
+      "subtitle": "WakaTime",
+      "title": "Show Today Summary",
+      "description": "Show your WakaTime ToDay Summary in Menu Bar",
```
iammola, line 39:
```suggestion
      "description": "Show Today's Summary in the Menu Bar",
```
How about something like this as the description?

extensions/wakatime/src/today-summary.tsx
```diff
@@ -0,0 +1,21 @@
...
+  const summary = useSummary();
+
+  return (
+    <MenuBarExtra
+      icon="../assets/icon.png"
+      title={`${summary.data
+        ?.map(
+          ([key, range]) =>
+            key === "Today" &&
+            (range.cummulative_total.seconds === 0 ? "Are we coding today?" : `Today: ${range.cummulative_total.text}`)
```
iammola, line 15: I think this should be an optional preference. Something that asks if they want to reminded the person hasn't recorded any time today.

extensions/wakatime/src/today-summary.tsx
```diff
@@ -0,0 +1,21 @@
...
+import { MenuBarExtra } from "@raycast/api";
+
+export default function Command() {
+  const summary = useSummary();
+
+  return (
+    <MenuBarExtra
+      icon="../assets/icon.png"
+      title={`${summary.data
+        ?.map(
```
iammola, line 12: I'll prefer you evaluate this outside the prop. just a design preference of mine.

extensions/wakatime/src/today-summary.tsx
```diff
@@ -0,0 +1,21 @@
...
+          ([key, range]) =>
+            key === "Today" &&
+            (range.cummulative_total.seconds === 0 ? "Are we coding today?" : `Today: ${range.cummulative_total.text}`)
+        )
+        .filter((item) => item)}`}
+      isLoading={summary.isLoading}
+    ></MenuBarExtra>
```
iammola, line 19:
```suggestion
    />
```
Since there aren't any children.

iammola (changes requested):
Hey @Fatpandac. This will be a really good feature to the extension. I've got some suggestions, but it all depends on @pernielsentikaer and what he thinks.

Fatpandac:
@iammola All is done. Thank for your review.

extensions/wakatime/src/today-summary.tsx
Fatpandac: image

You can make it be empty if you do not want to have any asks.

extensions/wakatime/package.json
```diff
@@ -31,6 +31,22 @@
...
+      "name": "today-summary",
+      "subtitle": "WakaTime",
+      "title": "Show Today Summary",
+      "description": "Show Today's Summary in the Menu Bar",
+      "mode": "menu-bar",
+      "interval": "1m",
+      "preferences": [{
+        "name": "showAsksSentence",
+        "title": "Show Asks Sentence",
+        "description": "Show asks sentence for you when your coding time is zero",
```
iammola, line 45: I don't think the message should be configurable but I'm not against it.

But there should be a `checkbox` preference on whether to show the menu bar item at all.

extensions/wakatime/src/today-summary.tsx
```diff
@@ -0,0 +1,23 @@
...
+        ([key, range]) =>
+          key === "Today" &&
+          (range.cummulative_total.seconds === 0 && asksSentence
+            ? asksSentence
+            : `Today: ${range.cummulative_total.text}`)
+      )
+      .filter((item) => item)}`;
+  }
+
+  return <MenuBarExtra icon="../assets/icon.png" title={getTitle()} isLoading={summary.isLoading} />;
```
iammola, line 22: With all my suggestions, I hope for it to look something like this. Where `null` is returned when the preference is disabled and the time for `Today` is 0.

```ts
const title = getTitle();

// Strictly false meaning today's time is 0
return title === false && [preferenceToShowZeroDays] !== true ? null : (
  <MenuBarExtra
    icon="../assets/icon.png"
    // isLoading expected to be true when title is undefined
    // and false when title is false or the time as a string
    isLoading={summary.isLoading}
    title={title || "Are we coding today?"}
  />
);

```

iammola (changes requested):

extensions/wakatime/src/today-summary.tsx
```diff
@@ -0,0 +1,23 @@
+import { useSummary } from "./hooks";
+
+import { MenuBarExtra, getPreferenceValues } from "@raycast/api";
+
+export default function Command() {
+  const summary = useSummary();
```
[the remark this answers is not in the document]
Fatpandac: OK

extensions/wakatime/src/today-summary.tsx
```diff
@@ -0,0 +1,23 @@
...
+  const asksSentence = preference.showAsksSentence;
+
+  function getTitle(): string {
+    return `${summary.data
+      ?.map(
+        ([key, range]) =>
+          key === "Today" &&
+          (range.cummulative_total.seconds === 0 && asksSentence
+            ? asksSentence
+            : `Today: ${range.cummulative_total.text}`)
```
iammola, line 17: You should use the `getDuration` helper in the utils. For consistency as it's used in other places in the extension.

https://github.com/raycast/extensions/blob/3a019140df76f4ff0720117a9daebbd4b1021ea6/extensions/wakatime/src/utils/index.ts#L3

iammola (changes requested):
One more thing...

extensions/wakatime/src/today-summary.tsx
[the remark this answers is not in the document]
Fatpandac: I have no API to enable or disable this command, but the user can switch-enabled it in Raycast preference.
image

extensions/wakatime/src/today-summary.tsx
```diff
@@ -0,0 +1,23 @@
...
+      )
+      .filter((item) => item)}`;
```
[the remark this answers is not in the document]
Fatpandac: This is used to filter out the null, maybe I need to make a new hook for the command.

extensions/wakatime/src/today-summary.tsx
Fatpandac: OK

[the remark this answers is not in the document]
iammola: Yeah I meant to show the `Are we coding today?` message. Whether anything should be shown when they haven't recorded any activity on that day.

Should probably default to `true`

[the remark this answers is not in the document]
Fatpandac: You meant is add a checkbox to config show or not show for asks, not add a text field to config show what sentence?

mattisssa (changes requested):
Hey! I got this issue once I tried to run the command **Show Summary**: 

CleanShot 2022-08-02 at 16 22 12@2x 

Probably this happened because I have a new empty account.

Also got into this for the new menu-bar command:

CleanShot 2022-08-02 at 16 23 48@2x

I think the reason is the same as for the first issue.

Fatpandac:
@mattisssa I created a new empty account, but I can not reproduce this bug.
image

Fatpandac:
@iammola Could you take a look?

extensions/wakatime/src/today-summary.tsx
```diff
@@ -0,0 +1,29 @@
...
+  function getTitle(): string | undefined {
+    const today = todaySummary.data;
+
+    if (today === undefined) return;
+
+    const cumulative_total = today.cummulative_total;
+    const todayText = getDuration(cumulative_total.seconds);
+
+    return cumulative_total.seconds === 0 && showAsksSentence ? undefined : `Today: ${todayText}`;
+  }
```
iammola, line 20: Hey @Fatpandac @mattisssa
I've been travelling the past few days, so I saw the messages but couldn't do anything at the time.

I've checked it now and have found the reasons for the @mattisssa's errors.

I couldn't reproduce it, but I encountered a similar error as I'm working on #2434. WakaTime's documentation doesn't say this, but some fields can be `null`.

You can change your function to this.
```suggestion
  function getTitle(): string | undefined {
    const { seconds } = todaySummary?.data?.cummulative_total ?? {};
    if (seconds == undefined) return;
    
    return seconds === 0 && showAsksSentence ? undefined : `Today: ${getDuration(seconds)}`;
  }
```

I'll fix the same thing along with the `NaN%` in @Fatpandac's image, other issues and some new features in the PR above.

iammola (changes requested):
I have a suggestion!!

iammola:
@Fatpandac even after your last changes, the menu bar will still show content even when the user doesn't want zero days

Fatpandac:
@iammola The menu bar will show asks when the user doesn't want to show zero days.
<div>
image
image
</div>

extensions/wakatime/src/today-summary.tsx
Fatpandac: Thx.

iammola:
I like the "`Are you coding today?`" replacing the "`Today: 0 sec`" message. So I don't think that should be configurable.

My intention for the 'asksSentence' preference was so on zero coding days like Sundays, the command won't take space it doesn't need to.

But you can ignore that and always show the "`Are you coding today?`" message. Remove the preference and we'll let @mattisssa review it again.

Fatpandac:
@iammola Do you want to show what on zero coding days?

extensions/wakatime/src/today-summary.tsx
```diff
@@ -0,0 +1,25 @@
+import { useTodaySummary } from "./hooks";
+import { getDuration } from "./utils";
+
+import { MenuBarExtra, getPreferenceValues } from "@raycast/api";
+
+export default function Command() {
+  const todaySummary = useTodaySummary();
+  const preference = getPreferenceValues();
+  const showAsksSentence = preference.showAsksSentence;
```
iammola, line 9: You have to remove uses of this as well

iammola (changes requested):
You can use the '`Are you coding today?`' message. But let's see what @mattisssa prefers.

extensions/wakatime/src/today-summary.tsx
Fatpandac: Done.

iammola:
@mattisssa @pernielsentikaer

mattisssa:
Hey! Sorry, it took long to get back to the review, looking at the changes now 🙌

mattisssa:
I still have the issue for **Today Summary** command:

>Error: Cannot find module 'react/jsx-runtime'
> Require stack:
> - /Users/.../.config/raycast/extensions/wakatime/today-summary.js
/Applications/Raycast.app/Contents/Resources/RaycastCommands_RaycastCommands.bundle/Contents/Resources/api/node>_modules/@raycast/api/index.js
Require stack:
>- /Users/.../.config/raycast/extensions/wakatime/today-summary.js
Function.Module._resolveFilename:loader:933:15
Function.Module._load:loader:778:27
Module.require:loader:1005:19
require:helpers:102:18
Object.<anonymous>:today-summary.tsx:1:1
Module._compile:loader:1101:14
Object.Module._extensions..js:loader:1153:10
Module.load:loader:981:32
Function.Module._load:loader:822:12
Module.require:loader:1005:19


And exactly the same issue I have for the **Show Summary** command.

Could you probably try to reproduce it with my token? 
It's `6303f8b5-42e8-4ff5-97fa-bd19b92b5634`

cc @Fatpandac @iammola

mattisssa:
And with the logs I have this in Raycast:

CleanShot 2022-08-08 at 13 26 00@2x

https://user-images.githubusercontent.com/15333662/183397756-cf04e8ca-3af4-495d-9d0d-706623539ec6.mp4

Fatpandac:
Maybe @iammola fixed it in #2434.

iammola:
So the issue is caused by the user's timezone not being set.

The WakaTime API then returns an object with an `error` property. I didn't know that was possible, so I never handled it.

I've made a commit for that now in https://github.com/iammola/raycast-extensions/commit/cad98c5de6ec2490742f4c933c5f077470d32c26, and it should throw an error if the property is in its response.

This also means that the suggestion I had in https://github.com/raycast/extensions/pull/2432#discussion_r939091987 was wrong. The `cummulative_total` property wasn't `null`, it didn't exist because the object only had the `error` property thus causing the `undefined` error when accessing the seconds.

I don't know why you're getting a missing `react/jsx-runtime` module for the `Show Today Summary` command though @mattisssa, maybe you're missing dependencies?

Regardless, please set the timezone to continue testing or delete this one and create a new user so the API key you shared won't be usable anymore.

mattisssa:
>I've made a commit for that now in https://github.com/iammola/raycast-extensions/commit/91000f8434df58d6922bc3200e719ef6208714fd, and it should throw an error if the property is in its response.

Does it make sense to have the commit as part of the PR? Since having the timezone not configured anyway should be proceeded properly 😬 


>I don't know why you're getting a missing react/jsx-runtime module for the Show Today Summary command though @mattisssa, maybe you're missing dependencies?

I ran `npm install` again to ensure I'm up to date. So I think it's something different. 

Here is another screencast of what I'm doing to repro:
https://user-images.githubusercontent.com/15333662/183614734-1ff71583-f331-48be-bf5c-2ae84c3b768d.mp4

iammola:
> Does it make sense to have the commit as part of the PR? Since having the timezone not configured anyway should be proceeded properly 😬

Yeah you're right. @Fatpandac, please push a commit checking if there's an error property in the response of the fetch in the `routeHandler` API helper.

it could help catch other errors or exceptions to the WakaTime API as well.

Fatpandac:
Checking what property?

iammola:
> Checking what property?

`error` is the key. Check the commit I've linked above.

mattisssa (approved):
Great! Thank you guys for tackling all the reported issues 🙌

PR #2434, opened 2022-07-30
✨ feat(wakatime): new features and fixes

iammola:
## Description
- Added Pagination to Leaderboards.
- Added Details to User's Activity Change.
- Replace user's email with timezone in accessory.
- Allow use of extension without `apiKey` in some commands.
- Fixed Sorting of Recent Projects.
- Fixed `NaN%` in activity change.

## Checklist

### DIFF
--- a/extensions/wakatime/src/utils/api.ts
+++ b/extensions/wakatime/src/utils/api.ts
@@ -8,7 +8,7 @@ import { getPreferenceValues } from "@raycast/api";
  * @param {string} endpoint - string - The endpoint to hit.
  * @returns A function that returns a promise that resolves to a RouteResponse<T>
  */
-async function routeHandler<T>(endpoint: string): Promise<Types.RouteResponse<T>> {
+async function routeHandler<T extends object>(endpoint: string): Promise<Types.RouteResponse<T>> {
   const baseURL = "https://wakatime.com/api/v1";
 
   try {


iammola:
I believe this is ready for review.

But I want #2432 to be added to the extension, so I can merge conflicts and update it to match the structural changes for other hooks and commands in this PR.

extensions/wakatime/package.json
```diff
@@ -16,47 +16,66 @@
...
+        }
+      ]
     }
   ],
   "preferences": [
     {
       "name": "apiKey",
-      "required": true,
+      "required": false,
       "title": "WakaTime API Key",
```
iammola, line 58: Since I made this optional, I had to repeat it in the `summary` and `private-leaderboards` commands that require it.

I don't know if there's a better way of handling this, because I expected the `apiKey` preference to be shared across all commands. But adding it to one command doesn't apply it to the others/entire extension.

extensions/wakatime/package.json
```diff
@@ -16,47 +16,66 @@
...
+          "title": "WakaTime API Key"
+        }
+      ]
     },
     {
       "name": "leaderboard",
       "subtitle": "WakaTime",
       "title": "Show Public Leaderboard",
-      "description": "Show your rank in the public global leaderboard",
+      "description": "An overview of the public leaderboard. Check your rank against others all around the world.",
```
iammola, line 34: Thoughts on this?

extensions/wakatime/src/hooks/use-user.ts
```diff
@@ -1,36 +1,27 @@
...
-      }
-
-      setIsLoading(false);
-    }
-
-    getData();
-  }, []);
-
-  return { ...(data ?? {}), isLoading };
+  const [cachedUser, setCachedUser] = useCachedState<WakaTime.User>("user");
```
iammola, line 8: Should I be caching the user?

extensions/wakatime/src/hooks/use-today-summary.ts
```diff
@@ -1,35 +1,23 @@
...
+    handler: useCallback(async () => {
+      const data = await getSummary("Today", new Date());
+      if (!data.ok) throw new Error(data.error);
+
+      return data;
+    }, []),
+    toasts: {
+      error: (err) => ({
+        title: "Error Loading Today's Summary",
+        message: err.message,
```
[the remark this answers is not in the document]
sxn: yeah this looks like something that should be fixed.
You can check whether the command was launched by the user or by a background refresh by checking [environment.launchType](https://developers.raycast.com/api-reference/environment#launchtype).

extensions/wakatime/src/hooks/use-user.ts
sxn: should be fine, yeah

extensions/wakatime/package.json
sxn: Unfortunately there's no current way of defining a per-extension required preference and make it not required for a particular command, so I guess this is the only way to go about it currently. 😅

extensions/wakatime/package.json
sxn: `Check your rank against other all around the world.` should be enough 😸

extensions/wakatime/src/leaderboard.tsx
```diff
@@ -1,17 +1,61 @@
...
+              {data.page > 1 && (
+                <Action
+                  title="Previous Page"
+                  onAction={() => setPage(String(data.page - 1))}
+                  shortcut={{ key: "arrowLeft", modifiers: ["shift"] }}
+                />
+              )}
+              {data.page < data.total_pages && (
+                <Action
+                  title="Next Page"
```
sxn, line 51: mind adding an `icon` prop to these two actions? either all `Action`s in the [ActionPanel](https://developers.raycast.com/basics/prepare-an-extension-for-store#action-panel) should have an icon, or none of them should 😸

extensions/wakatime/package.json
iammola: This works! My only problem is the user still has to provide the API key for that command as well. It's still better than the way I had it before, because now it's only one other command that asks for it.

Just to be sure, is there a way to set the preference programmatically? I checked but didn't find anything.

sxn: > the user still has to provide the API key for that command as well

do you mean the fact that the field still appears in preferences for that particular command? or something else?

The current version of the branch works as follows for me:

https://user-images.githubusercontent.com/1155589/198271493-d3f58d8c-3c3e-422e-9ffc-d0cf4ce71f3c.mov

iammola: That's not my direct problem with it, but yeah I'll like for the same API key to be used across the extension and updating in the public leaderboard won't update the extension-wide one.

sxn:
🎉

PR #2435, opened 2022-07-30
Gopass: Minor Fixes

5mutawa:
## Description
- Fix an issue with clearing clipboard after password copy.
- Fix an issue with retrieving passwords.
- Add the functionality to open links in the browser.

## Screencast
N/A

## Checklist

### DIFF
--- a/extensions/gopass/package.json
+++ b/extensions/gopass/package.json
@@ -14,7 +14,7 @@
     {
       "name": "index",
       "title": "Search Passwords",
-      "subtitle": "gopass",
+      "subtitle": "Gopass",
       "description": "Search and copy passwords from gopass password manager",
       "mode": "view"
     }


extensions/gopass/src/details.tsx
```diff
@@ -58,7 +59,13 @@ export default function ({ entry }: { entry: string }): JSX.Element {
               key={index}
               title={humanize(key)}
               subtitle={value}
-              actions={<Actions copy={() => copy(key, value)} paste={() => paste(key, value)}></Actions>}
+              actions={
+                <Actions
+                  copy={() => copy(key, value)}
+                  paste={() => paste(key, value)}
+                  url={isValidUrl(value) ? value : ""}
```
mattisssa, line 66: Shouldn't this be `undefined` to avoid adding the `{url && <Action.OpenInBrowser url={url} />}` from the function above?
```suggestion
                  url={isValidUrl(value) ? value : undefined}
```

mattisssa (changes requested):
Hey! Thanks for the contribution, check my comment about the URL action you have changed. Not sure if it's a valid point but could be important. 

Also you have a warning about the command title "gopass", please follow the description:
>Extension's title has to be Title Cased. Expected "Gopass"

And to me `Gopass` is more a `subtitle` so the title should be a bit different. Our [guidelines](https://developers.raycast.com/basics/prepare-an-extension-for-store#extensions-and-commands-naming) can help 😉

extensions/gopass/src/details.tsx
5mutawa: Thank you @mattisssa, I have addressed the title case. 

The condition to show `OpenInBrowser` wasn't the best. I made a small refactor to make the code easier to understand.

mattisssa (approved):

PR #2473, opened 2022-08-03
GitLab: adds option to filter "My Merge Requests"

wierdbytes:
## Description

Before this PR command "My Merge Requests" displayed MRs with any state (opened, closed, merged, ...). I found this not convinient for my workflow. This changes will provide ability to choose option in preferences to specify which state of MRs should be displayed.

## Screencast

изображение

## Checklist

### DIFF
--- a/extensions/gitlab/package.json
+++ b/extensions/gitlab/package.json
@@ -261,7 +261,7 @@
     "immint023",
     "duboiss",
     "metakirby5",
-    "mentro"
+    "wierdbytes"
   ],
   "dependencies": {
     "@apollo/client": "^3.4.8",


Loskir:

Have you tried using [list dropdown](https://developers.raycast.com/api-reference/user-interface/list#dropdown) instead?

danielsequeira:
> Thank you for your contribution!
> 
> Have you tried using [list dropdown](https://developers.raycast.com/api-reference/user-interface/list#dropdown) instead?

I was actually going to propose the same thing. I think that in terms of UX makes a lot more sense to have this as a list dropdown inside the "My Merge Requests" command instead of having it as a preferences setting.

Loskir:
I see, interesting...
Could someone from Raycast give an advice?

tonka3000:
@wierdbytes @Loskir Use [ActionPanel.Submenu](https://developers.raycast.com/api-reference/user-interface/action-panel#actionpanel.submenu). With this you can achieve the same. Use the cache features from raycast api if you wanna remember the action.

metakirby5:
Another suggestion is to have a checkbox preference to toggle which control gets to be shown as the list filter. Then, users can see the one that fits their preference.

Issue #2138, opened 2022-06-29
[Brew] Upgrade commands failing to execute


1activegeek:
# Extension – [Brew](https://github.com/raycast/extensions/tree/c857437cb85229cc3e11177172ea408b8e03afa1/extensions/brew/)
Author: @nhojb 

## Description
Upgrades are failing for any installed app I try to update with Home-brew.
## Steps To Reproduce

1. Use the "brew" command in Raycast. 
2. Select the Show Outdated option. 
3. Choose to upgrade the outdated app
4. Error shown and action not taken

```
sudo: a terminal is required to read the password; either use the -S option to read from standard input or configure an askpass helper
sudo: a password is required
Error: blockblock: Failure while executing; `/usr/bin/sudo -E -- /opt/homebrew/Caskroom/blockblock/2.0.5/BlockBlock\ Installer.app/Contents/MacOS/BlockBlock\ Installer -uninstall` exited with 1. Here's the output:
sudo: a terminal is required to read the password; either use the -S option to read from standard input or configure an askpass helper
sudo: a password is required
```

## The current behavior
The actions that require Sudo, are failing.

## The expected behavior
The actions that require Sudo, would work successfully.

### Raycast version

1.36.1

1activegeek:
Checking in here - anything I can do @nhojb? Any other information needed. Or am I missing something obvious for enabling this?

nhojb:
Thanks for reporting the issue. In most cases brew is designed not to require `sudo`. Do you know why it is required in this case? Looks like it's running an uninstall operation for some app?

It looks like we can force `sudo` to prompt the user for their password setting the `SUDO_ASKPASS` environment variable in the extension. I'll have a look into this.

1activegeek:
Updating with context from the Slack chat in case others are searching.

It seems that not ALL requests are requiring SUDO, but when they do, it's failing as there is no prompt. Specifically the issue I hit this with was with 1Password, I've since updated it already out of band so it was no longer a culprit. Here I have an example though of attempting to update the out of data LuLu package. Output from a standard Homebrew upgrade of the package at the command line for context:

```
user@machine % brew upgrade lulu
==> Upgrading 1 outdated package:
lulu 2.4.1 -> 2.4.2
==> Upgrading lulu
==> Downloading https://github.com/objective-see/LuLu/releases/download/v2.4.2/LuLu_2.4.2.dmg
[...]

PR #2479, opened 2022-08-04
Brew support sudo

nhojb:
## Description

It's quite common for cask install/uninstall/upgrade to require sudo. Normally the password would be entered via the terminal, but this isn't possible in the extension.

Instead we set the `SUDO_ASKPASS` environment variable (see `man sudo`). This must be set to an executable which prompts the user for their login password.

Unfortunately macOS does not seem to include a suitable executable. One common work-around (based on Google research) is to prompt the user via an AppleScript dialog. This is the solution taken by the extension.

Notes:
- The user is only prompted for a password if the installation etc requires sudo.
- Brew itself cannot be run with admin rights. SUDO_ASKPASS is currently the only solution to authorising other installation/uninstallation tasks.
- SUDO_ASKPASS requires a valid executable. So we can't prompt the user for their password via a (yet to be implemented) raycast API. There's no way we can forward this password to sudo from the extension.
- Raycast could, in future, provide such an executable, which could display (via XPC) the password prompt within the app.
- PR is based on #2478 - otherwise it would introduce conflicts.

Caveats:
- The Raycast window is closed when the AppleScript password dialog is displayed. This isn't a great UX 😞 

Fixes [2138](https://github.com/raycast/extensions/issues/2138)

## Screencast

https://user-images.githubusercontent.com/312540/182859106-54c7348b-d9b2-4824-8488-6c2a5225321b.mov

Screenshot 2022-08-04 at 14 15 32

## Checklist

### DIFF
--- a/extensions/brew/CHANGELOG.md
+++ b/extensions/brew/CHANGELOG.md
@@ -1,12 +1,9 @@
 # Brew Changelog
 
-## [Sudo] - 2022-08-10
-
-- The extension will now prompt for user's login password if brew requires authorization when installed or uninstalling casks.
-
-## [Custom Path] - 2022-08-10
+## [Improvements] - 2022-08-10
 
 - Add a "Custom Brew Path" preference for users with brew installed in a non-standard path
+- The extension will now prompt for user's login password if brew requires authorization when installed or uninstalling casks.
 
 ## [Improvements] - 2022-06-06
 


pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #2497, opened 2022-08-05
Include files of shared drives, not only my drive

marcusgrando:
## Description

Added a dropdown to search not only on My Drive but also on Shared Drives.

Reference: https://developers.google.com/drive/api/v3/reference/files/list#includeItemsFromAllDrives

### DIFF
--- a/extensions/google-workspace/src/search-google-drive-files.tsx
+++ b/extensions/google-workspace/src/search-google-drive-files.tsx
@@ -44,7 +44,7 @@ function SearchGoogleDriveFiles() {
           />
           <List.Dropdown.Item title="In full text in My Drive" value={`${QueryTypes.fullText}-${ScopeTypes.user}`} />
           <List.Dropdown.Item
-            title="in full text in All Drives"
+            title="In full text in All Drives"
             value={`${QueryTypes.fullText}-${ScopeTypes.allDrives}`}
           />
         </List.Dropdown>


mathieudutour:
@thomaslombart might be worth adding a dropdown to switch between the two

marcusgrando:
Hi @mathieudutour. I've tried to implement as preference instead of dropdown because it's make more sense to me, also fits for me daily usage.

mathieudutour:
I'm not sure I understand what you mean. How does it fits more your usage? The dropdown can remember the last value you set, so the usage wouldn't change much - but it'd make it a lot more visible

marcusgrando:
Hi @mathieudutour, sorry for the poor explanation. For example, today I use it for work, so my daily basis use is searching on shared drives, not on my drive. Because of that makes more sense to me to turn on and keep it on as a preference, not changing it every time on dropdown even with shortcuts.

marcusgrando:
@mathieudutour now I think I understand. You say that dropdown remembers my last usage. So, I don't know which method is the best fit.

mathieudutour:
> You say that dropdown remembers my last usage

Yes exactly, we have a `storeValue` that's made just for that! We tend to use dropdowns for things like filtering or sorting a list and I feel like this is sort of a filter (the options could be "All", "Shared With Me", "My Files" or something)

thomaslombart:
@mathieudutour Are you talking about a dropdown in the "Search Google Drive Files" command? If yes, there's already a search bar dropdown to select how you want to filter the files (file name or full text)

mathieudutour:
oh I missed that sorry. I kind of feel like the "file name vs full text" should be the preference rather than the "shared with me vs my files" 😄

thomaslombart:
To be honest, I'm using both searches depending on my current use case. When I'm quickly searching for a file, I'm using the file name. When I don't know the file name but vaguely remember the content, I'm using the full-text search.

As @marcusgrando is saying, the full-text search includes the search at file name. So maybe, we can just get rid of the file name search to include the My Drive/Shared Drives dropdown?

marcusgrando:
@mathieudutour any thoughts about the last comments?

thomaslombart:
I've checked out Google Drive's API and they specify that the `supportsAllDrives` query parameter should be added as well. @marcusgrando Did it work out for you locally without this parameter? I got a forbidden error on my end.

CleanShot 2022-08-17 at 10 52 30@2x

zhouk:
Getting "Failed to retrieve files" when searching all drives.

zhouk:
@thomaslombart and @marcusgrando I tried it locally and it looks like these two lines need to go together to search all drives.

`    params.append("supportsAllDrives", "true");`
`    params.append("includeItemsFromAllDrives", "true");`

marcusgrando:
> @thomaslombart and @marcusgrando I tried it locally and it looks like these two lines need to go together to search all drives.
> 
> ` params.append("supportsAllDrives", "true");` ` params.append("includeItemsFromAllDrives", "true");`

Thanks. Updated patch. I'm not able to test it before.

Best regards

marcusgrando:
@mathieudutour @thomaslombart Just to know, there's anything yet to do here? Let me know 😊

thomaslombart:
@marcusgrando FYI, I don't have any rights over this PR 🙂 
If getting the files from shared drives also returns personal files, I'd recommend not adding more options in the dropdown and simply adding the parameters by default.

mathieudutour (approved):

PR #2518, opened 2022-08-09
[Spotify-Player] Introduce "Now Playing" (view) command

dancannon:
## Description

This PR migrates the now playing command out of #2456 without using the overly complicated sub commands. The primary difference is the use of the Detail component over a typeahead/list details, this means the title and artwork are more visible. 

Because this introduces a "Now Playing" command I would like to propose renaming the menu bar to "Now Playing Menu Bar" to differentiate the two (I think this command should take the shorter name as its more likely to be something the user types given that the menu bar command isnt something that can be interacted with). Feedback especially welcome here around what would be the "Raycast way of naming" for this kind of situation.

Thankfully as this change is much smaller in scope the PR is quite a bit simpler. I've also some things I broke while resolving merge conflicts before, in particular the track notifications not appearing and the menu bar not updating when changing tracks.

## Screencast

I've updated the assets to include some new screenshots but this is basically a simplified version of #2456

spotify-05

## Checklist

### DIFF
--- a/extensions/spotify-player/src/nowPlayingMenuBar.tsx
+++ b/extensions/spotify-player/src/nowPlayingMenuBar.tsx
@@ -10,6 +10,7 @@ import {
   Icon,
   Color,
   getPreferenceValues,
+  updateCommandMetadata,
 } from "@raycast/api";
 import { useEffect, useState } from "react";
 import { likeCurrentlyPlayingTrack, startPlaySimilar } from "./spotify/client";
@@ -44,7 +45,14 @@ export default function NowPlayingMenuBar() {
         setCurrentSpotifyState(state);
         setCurrentlyPlayingTrack(track);
         result = [state, track];
+
+        let newSubtitle: string | undefined;
+        if (track && track.id) {
+          newSubtitle = `${track.artist} – ${track.name}`;
+        }
+        await updateCommandMetadata({ subtitle: newSubtitle });
       } catch (err) {
+        await updateCommandMetadata({ subtitle: undefined });
         if (environment.launchType != LaunchType.Background) {
           showToast(Toast.Style.Failure, String(err));
         }


extensions/spotify-player/CONTRIBUTING.md
```diff
@@ -0,0 +1,13 @@
...
+
+Please follow the [Contribute to an Extension](https://developers.raycast.com/basics/contribute-to-an-extension) guide.
+
+We're still figuring things out and updating our guidelines. If at any point something is unclear, please tell us in [our community](https://raycast.com/community).
+
+---
+
+### Wallpapers
+
+Screenshots are taken with the "[Moonrise](https://www.raycast.com/uploads/wallpapers/moonrise.png)" background from https://www.raycast.com/wallpapers
```
dancannon, line 13: I couldnt find the one previously used but thought it would make sense to use something from the public set? Just to make future screenshots safer/easier

mattisssa: yeah, don't worry about it 👍 
hm, interesting why the one I used is not a part of the wallpapers set 🤔

mattisssa (changes requested):
Hey! Everything looks good, I just got into this when I have spotify app closed and opened **Now Playing** command:

CleanShot 2022-08-13 at 10 28 10@2x

Could you please have a look?

mattisssa:
By the way,  please rebase my latest changes from the `main` repo, I updated a `client.ts` a bit

extensions/spotify-player/package.json
```diff
@@ -38,7 +38,7 @@
     },
     {
       "name": "nowPlayingMenuBar",
-      "title": "Now Playing Menu Bar",
+      "title": "Menu Bar Player",
```
dancannon, line 41: Good thinking

extensions/spotify-player/src/nowPlaying.tsx
```diff
@@ -191,21 +189,23 @@ export default function NowPlayingMenuBar() {
...
-              <Action
-                icon={Icon.Heart}
-                title="Like"
-                onAction={async () => {
-                  try {
-                    const response = await likeCurrentlyPlayingTrack();
-                    if (response?.result) {
-                      const title = `${response.result.artist} – ${response.result.name}`;
-                      showHUD(`💚 ${title}`);
+              {authorized && (
```
dancannon, line 192: 👍

extensions/spotify-player/src/nowPlayingMenuBar.tsx
```diff
@@ -26,41 +25,37 @@ export default function NowPlayingMenuBar() {
...
-          const [state, track] = await Promise.all([getState(), getTrack()]);
-
-          setCurrentSpotifyState(state);
-          setCurrentlyPlayingTrack(track);
-
-          let newSubtitle: string | undefined;
-          if (track.id) {
-            newSubtitle = `${track.artist} – ${track.name}`;
-          }
-          await updateCommandMetadata({ subtitle: newSubtitle });
```
dancannon, line 52: Is this correct? If I understand the API correctly then the command metadata will only be updated when the now playing view command is executed (not in the background). That was my reasoning for keeping this code path in both the now playing view command AND the menu bar. 

In fact now that I think about it maybe the command metadata should only be updated in the menu bar command and not the view command since that will be come out of date unless the user interacts with it.

I wonder if a raycast API for updating other the command metadata for other commands in the extension would be useful here?

dancannon:
I have pushed some new commits to tidy up the handling when Spotify is not running (and rebased to include your changes)

extensions/spotify-player/src/nowPlayingMenuBar.tsx
mattisssa: >In fact now that I think about it maybe the command metadata should only be updated in the menu bar command and not the view command since that will be come out of date unless the user interacts with it.

yeah, makes sense. 
I removed the subtitle updating from the menu bar command since I didn't want to have it for both commands. 
I'm thinking now that probably it makes sense to have the subtitle updating in the menu bar command only and remove it from **"Now Playing"** 🤔

mattisssa (approved):

extensions/spotify-player/src/nowPlayingMenuBar.tsx
dancannon: > I'm thinking now that probably it makes sense to have the subtitle updating in the menu bar command only and remove it from "Now Playing" 🤔

I think this would make sense, I removed it from "Now Playing" for now.

mattisssa: I brought it back

PR #2531, opened 2022-08-11
Bibmanager search improvements

AaronDavidSchneider:
## Description

The main change of this PR is the custom fuzzy search, similar to the one used in the `zotero` raycast extension. Additionally, this PR updates screenshots and adds shortcuts for actions.

## Screencast

bibmanager-4
bibmanager-3
bibmanager-2
bibmanager-1

## Checklist

### DIFF
--- a/extensions/bibmanager/src/index.tsx
+++ b/extensions/bibmanager/src/index.tsx
@@ -1,4 +1,4 @@
-import {useEffect, useMemo, useState} from "react";
+import { useEffect, useMemo, useState } from "react";
 import { List, environment, getPreferenceValues, showToast, Toast } from "@raycast/api";
 import { ListBibmItem } from "./ListBibmItem";
 import { Item, Preferences, State } from "./types";


AaronDavidSchneider:
Not really an issue, but a small glitch:
When triggering the command, there is a very short (less than 0.1s maybe) transition from loading to the final list, where I get an emptyview for some reason.

Could the reviewer of this PR help me with that?

extensions/bibmanager/src/index.tsx
```diff
@@ -46,10 +40,22 @@ export default function Command() {
     }
     loadItems();
   }, []);
+
+  useEffect(() => {
+    if (state.items.length > 0) {
+      filterList(filterFct(state.items, state.searchText));
```
mathieudutour, line 46: Instead of using another state, you could probably use `useMemo` to avoid another render

AaronDavidSchneider: Thanks @mathieudutour! Tried to do so. Do you mean like this?
At least it seems to avoid the issue that I described!

mathieudutour: Not really. `useMemo` returns the result of the operation inside. So instead of using a state for `filteredList`, you can just use the result of `useMemo`. Does that make sense?

AaronDavidSchneider: Ah... Now I get it. How about this one?

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #2550, opened 2022-08-12
[CUID] Add extension to generate CUIDs

dgrcode:
## Description

This PR adds a new extension to generate CUIDs. A bit more info about CUIDs [here](https://github.com/ericelliott/cuid)

One possible bug is at the final state. Raycast closes, but when I open raycast again it is "inside" the cuid generator extension, which doesn't have any interface, so it's sitting on an empty raycast window. Gif below:
Screen Recording 2022-08-12 at 18 29 32

Is this related with being running a development version of the extension?

## Screencast

https://user-images.githubusercontent.com/17493200/184402147-8beffc04-87f3-446f-a9e6-1b95b48d59fe.mov

## Checklist

### DIFF
--- a/extensions/cuid-generator/package.json
+++ b/extensions/cuid-generator/package.json
@@ -13,7 +13,7 @@
     {
       "name": "generate",
       "title": "Copy CUID",
-      "description": "Generates a new cuid",
+      "description": "Generates a new CUID",
       "mode": "no-view"
     }
   ],


extensions/cuid-generator/package.json
```diff
@@ -0,0 +1,39 @@
...
+  "categories": [
+    "Developer Tools"
+  ],
+  "license": "MIT",
+  "commands": [
+    {
+      "name": "generate",
+      "title": "Generate cuid",
+      "description": "Generates a new cuid",
+      "mode": "view"
```
sxn, line 17: > One possible bug is at the final state. Raycast closes, but when I open raycast again it is "inside" the cuid generator extension, which doesn't have any interface, so it's sitting on an empty raycast window. Gif below:

The reason for this behaviour is that the command is a `view` command, when it doesn't actually render any view. 
```suggestion
      "mode": "no-view"
```

extensions/cuid-generator/package.json
```diff
@@ -0,0 +1,39 @@
...
+  "icon": "command-icon.png",
+  "author": "dgrcode",
+  "categories": [
+    "Developer Tools"
+  ],
+  "license": "MIT",
+  "commands": [
+    {
+      "name": "generate",
+      "title": "Generate cuid",
```
sxn, line 15: Since `CUID Generator` is already the extension's title and will thus show up as the subtitle, there's no need to repeat it in the command's title. Instead, you could change the command to something like `Copy CUID`.

Instead of just copying to the clipboard, you could provide a preference to allow directly [pasting](https://developers.raycast.com/api-reference/clipboard#clipboard.paste) to the frontmost application. You can find an example of it [here](https://github.com/raycast/extensions/blob/main/extensions/cpf-cnpj-generator/package.json#L29-L36) 😸 


```suggestion
      "title": "Copy CUID",
```

dgrcode:
Awesome! Thanks for the review :)

Regarding the possible actions (pasting to active window instead of copy, etc), I'll keep the copy to clipboard to make it similar to the UUID extension. But it's good to know there are more options there 😄

sxn:
the build's failing because of the lock file. Mind having a look? 😄 we should be fine to merge once CI passes.

dgrcode:
It seems I had quite an old version of npm. Upgraded and reinstalled. The previous error should be gone now 🤞

And all check just passed 🎉

dgrcode:
@sxn thanks for your support here. And also, great nit addition on [8411aa1](https://github.com/raycast/extensions/pull/2550/commits/8411aa10066590542abbe71faeda86a33959e72f). I was actually a bit concerned this morning when I saw this:
image
Now it's more consistent and I can be happy again haha (to be honest I think it should be "Generate UUID" since it only generates one, but that's a different extension and I can definitely live with it 😄).

sxn:
Happy to help! 😸

> to be honest I think it should be "Generate UUID" since it only generates one

A [recent PR](https://github.com/raycast/extensions/pull/2592) changed that behaviour so now you can generate anywhere from 1 to 10000 UUIDs. But feel free to open a PR updating the name if you think it should be singular. Worst case the author disagrees and we close the PR, best case it gets merged and your itch gets scratched. 😉

dgrcode:
> A https://github.com/raycast/extensions/pull/2592 updated the extension so now you can generate anywhere from 1 to 10000 UUIDs

I missed that update. I take that comment away then :)

PR #2552, opened 2022-08-12
Icons8 Update

yug2005:
## Description

* Added option to change background color and add padding.
* Improved loading times, preview image quality, field for custom color. 
* Fixed bugs with configure action within view icon, copy image url, and moving pinned icons. 
* Added new metadata images and updated README.md.

## Screencast

06 Configure Icon

## Checklist

### DIFF
--- a/extensions/icons8/src/components/empty-view.tsx
+++ b/extensions/icons8/src/components/empty-view.tsx
@@ -6,7 +6,7 @@ export const EmptyView = (): JSX.Element => {
 
 export const InvalidAPIKey = (): JSX.Element => {
   return (
-    <Grid>
+    <Grid searchBarPlaceholder="Search Icons">
       <Grid.EmptyView
         title="Invalid API Key"
         description="Get Valid Api Key From Icons8 Developers."


pernielsentikaer (changes requested):
Hi @yug2005 👋 

This looks like a pretty awesome update, I just tested the scenario when you have wrong a wrong API key, do you think we could use isLoading to avoid this flicker?

https://user-images.githubusercontent.com/38033107/185051877-90535730-d27e-49d8-b74d-92f3ae088af5.mp4

yug2005:
Hey @pernielsentikaer, I know this bug looks bad but I think that it is better with this bug because when there is a correct API key, the loading feels instant. This is because the way I validate the API key is by loading in the icon styles (for the grid dropdown). The default style is always "All Styles" so I don't need to finish loading the styles before showing the home view. Therefore, I set `isLoading` to `false` before getting the styles and validating the API key, and this makes the loading feel instant. 

It also leads to the following bug where if you press "cmd + p" instantly, the other styles will not show but I am sure no one would notice this : 

https://user-images.githubusercontent.com/77864399/185054931-a9c9955f-0550-48de-9d4c-bac4a4d1b813.mov

Is this okay? Personally, I really think that the feel of an instant load time of the home view is worth it for these small bugs. Here is what the fixed state would look like on every use : 

https://user-images.githubusercontent.com/77864399/185056461-6244643c-2593-47eb-baca-be81a70e17aa.mov

pernielsentikaer (approved):

PR #2562, opened 2022-08-13
Brew update

Aayush9029:
## Description

*Large casks and formulaes take a while to install, which leads Raycast window being un-usable for the time being, this updates allows one to copy un/install command and run them directly in their terminal.*

- Add a copy install command action
- Add run install command in terminal action

## Screencast

N / A

## Checklist


nhojb (changes requested):
Hi @Aayush9029, thanks for your contribution 🎉 

I have a couple of suggested improvements:

## 1. Actions
I wonder if we can improve the naming of the "Run in Terminal" action and also the grouping of actions? I think it's better if the action is explicit about what it is running in the terminal - as your copy action already is. For example "Run Install in Terminal".

### Uninstalled Actions:
Here I think we could reorganise & rename the command palette actions like this:

```
Show Details
Install
----
Copy Formula Name
Copy Install Command
Run Install in Terminal
----
Open in Browser
Copy URL
```

### Installed Actions:
Here I think it makes more sense to group the new actions with the "Uninstall" action itself like this:

```
Show Details
Show in Finder
Pin
----
Open in Browser
Copy URL
----
Uninstall
Copy Uninstall Command
Run Uninstall in Terminal
```

## 2. Brew Path
If would be better if the action content used the full path to `brew` (which the user can customise), rather than relying on the current environment. For example:

```
content={`${brewExecutable()} install ${formula.name}`}
```

You'll need to export `brewExecutable()` (from brew.ts).

Thanks!

Aayush9029:
@nhojb I've added the changes you've mentioned including organizing commands and using brewPath.

Lmk if there's anything that needs to be changed

nhojb (approved):
Looks good. Thanks for the changes and your contribution 😄 

One minor thing: I'm curious why in the commit history you deleted the `extension/brew` directory, then uploaded new files and did a bunch of renames? Would it not have been easier just to update the source files in-place?

Not that it matters, since we squash & rebase 😅

Aayush9029:
@nhojb i didn't want to clone the entire raycast extensions repo just to work on couple of files.

Just a personal preference thing :)

PR #2573, opened 2022-08-14
Github extension

qeude:
## Description

I created a Github extension that makes use of the extension api so it can uses new features such as a menu bar extension and so on. The idea is to avoid having multiple Github extension that almost do the same thing to have a single one that gather every Github features. To do so I kind of started by replicating the "My Pull Requests" feature from the built-in extension so we can start iterate on it.
I also added a menu command that list pending requested reviews.

## Screencast
Screenshot 2022-08-14 at 23 02 08
Screenshot_2022-08-14_at_23_02_23

## Checklist

### DIFF
--- a/extensions/github/package.json
+++ b/extensions/github/package.json
@@ -30,8 +30,8 @@
       "name": "token",
       "type": "password",
       "required": true,
-      "title": "API Token",
-      "description": "Your Github API token.",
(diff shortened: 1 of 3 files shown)


michaelwschultz:
This looks so good @qeude pumped to check it out when it's live.

PR #2588, opened 2022-08-16
New extension: ColorSlurp

Jordan-Ellis:
## Description

An extension that allows you to control [ColorSlurp's](https://colorslurp.com/) features from Raycast.

## Screencast

screenshot-1

## Checklist

### DIFF
--- a/extensions/colorslurp/package.json
+++ b/extensions/colorslurp/package.json
@@ -19,7 +19,6 @@
       "mode": "no-view",
       "keywords": [
         "dropper",
-        "pick color",
         "magnifier"
       ]
     },
@@ -28,9 +27,6 @@
       "title": "Edit Color",
       "description": "Shows the specified color in ColorSlurp.",
       "mode": "no-view",
-      "keywords": [
-        "picker"
-      ],
       "arguments": [
         {
           "name": "color",


extensions/colorslurp/package.json
```diff
@@ -0,0 +1,94 @@
...
+  "author": "Jordan-Ellis",
+  "categories": [
+    "Applications",
+    "Design Tools",
+    "Developer Tools"
+  ],
+  "license": "MIT",
+  "commands": [
+    {
+      "name": "show-magnifier",
```
sxn, line 16: shouldn't this be `show-picker`? 🤔

sxn:
hi @Jordan-Ellis - nice work w/ the extension!

One issue I've run into: when you don't have the application running, none of the commands work as intended; instead, running any command just results in the application being opened. 🤔

extensions/colorslurp/package.json
Jordan-Ellis: The Magnifier in ColorSlurp is what allows you to pick colors from your screen, so “Pick Color” is supposed to be a shortcut to showing the magnifier.  I could rename it to “Show Magnifier“, to be more inline with what it does if you’d like. 

I realized that the name of the other command pick-color didn’t align with the title of “Edit Color“, so I just pushed a change that fixes the naming.

Jordan-Ellis:
> One issue I've run into: when you don't have the application running, none of the commands work as intended; instead, running any command just results in the application being opened. 🤔

Hi @sxn! Looks like you found a bug with ColorSlurp, so I’ll have to release a ColorSlurp update. I already got it fixed in our dev version of ColorSlurp, so it’s not a problem with the extension. Thanks for pointing that out!

sxn:
oh - I hadn't realised you're the creator of ColorSlurp! 😸 glad to hear the issue's been fixed. Let's merge the PR then - apart from that issue, the extension works great. 🎉

Jordan-Ellis:
Awesome, thanks! 🎉

extensions/colorslurp/src/utilities.ts
```diff
@@ -0,0 +1,24 @@
...
+    return bundleId === "com.IdeaPunch.ColorSlurp";
+  });
+}
+
+export async function openColorSlurpUrl(url: string) {
+  if (!(await isColorSlurpInstalled())) {
+    await showToast({
+      style: Toast.Style.Failure,
+      title: "ColorSlurp isn't installed",
+      message: "Install it from: https://colorslurp.com",
```
sxn, line 16: one thing I realised just after merging - [Toasts](https://developers.raycast.com/api-reference/feedback/toast#toast) allow you to specify a `primaryAction` and `secondaryAction` that the user can invoke. So something like this

```suggestion
      message: "You can download it from the website or the app store",
      primaryAction: { title: "Open Homepage", onAction: () => open("https://colorslurp.com") },
      secondaryAction: {
        title: "Open App Store",
        onAction: () => open("https://apps.apple.com/us/app/colorslurp/id1287239339#?platform=mac"),
      },
```

would show up like this:

image

Not a big deal, but would make it ever so slightly simpler for users to access your website. 😸

Jordan-Ellis: Oh that's super cool! Should I create a new pull request?

sxn: go for it! 😉

PR #2591, opened 2022-08-16
[Hacker News] 429 API error fix

GastroGeek:
## Description

Set a random UA string in API calls to prevent 429 error responses.

## Checklist

### DIFF
--- a/extensions/hacker-news/src/StoryListItem.tsx
+++ b/extensions/hacker-news/src/StoryListItem.tsx
@@ -3,7 +3,7 @@ import { ActionPanel, List, Action, Icon, Image } from "@raycast/api";
 import Parser from "rss-parser";
 
 export function StoryListItem(props: { item: Parser.Item; index: number }) {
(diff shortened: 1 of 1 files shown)


sxn:
hey @GastroGeek - thanks for the PR! It feels a bit bad to use a random UA to get past the rate limiting. I'm thinking we should create a custom user agent instead? Something like:
```
const userAgent = `Hacker News Extension, Raycast/${environment.raycastVersion} (${os.type()} ${os.release()})`;
```

Additionally, according to [this issue](https://github.com/hnrss/hnrss/issues/71), results are cached server-side for 5 minutes, so we could do the same: only perform an API call if the last result is older than 5 mins, otherwise render from the cache. That way we'd be mindful of the service, while also making it clear where the requests are coming from.

WDYT?

sxn:
hope you don't mind, I took the liberty and implemented the aforementioned suggestions :D the UX should now be a bit better since we cache responses now 😸

PR #2606, opened 2022-08-17
New Extension: Empty Screenshot Folder

chanceaclark:
## Description

Adds a new extension that cleans up the screenshot folder. My screenshot folder ends up getting polluted over time with old screenshots and screen recordings, so I thought I'd bring a way to clean it up more easily. By default it'll delete any screenshots/screen recordings older than 30 days, but you can configure it between 1, 7, 30, and 90 days.

## Screencast

https://user-images.githubusercontent.com/10539418/185220734-ef4f1e8c-ea2b-4535-8d79-980e79a0d3b6.mov

## Checklist

### DIFF
--- a/extensions/empty-screenshots/CHANGELOG.md
+++ b/extensions/empty-screenshots/CHANGELOG.md
@@ -1,5 +1,5 @@
 # Empty Screenshot Folder Changelog
 
-## [Initial Version] - 2022-08-17
+## [Initial Version] - 2022-09-20
 
 Added inital code


extensions/empty-screenshots/src/index.ts
```diff
@@ -0,0 +1,49 @@
...
+  const { stdout, stderr } = await asyncExec(
+    "defaults read com.apple.screencapture | plutil  -convert json -r -o - -- -"
+  );
+
+  if (stderr) {
+    showToast(Toast.Style.Failure, "Something went wrong", stderr);
+  }
+
+  const numberOfDays = parseInt(daysToKeep, 10);
+  const { location } = JSON.parse(stdout);
```
sxn, line 24: in my case, `stdout` is:
```
{
  "last-analytics-stamp" : "682863474.458861",
  "last-selection" : {
    "Height" : "315",
    "Width" : "364",
    "X" : "453",
    "Y" : "0"
  },
  "last-selection-display" : "0",
  "style" : "selection",
  "video" : "1"
}
```
so `location` ends up being undefined, crashing the extension 😅 

image

What about making this a preference instead?

pernielsentikaer: Just out of curiosity, how do you even set it? I tried to look in macOS Ventura for it and I'm not sure where to find it (Also getting the same error as @sxn)

chanceaclark: Not sure about Ventura, since I'm still on Monterey, but [here's an article](https://www.macworld.com/article/673251/how-to-change-where-screenshots-are-saved-on-a-mac.html#how-to-change-screenshot-location-on-a-mac) on how to do it on other OS versions.

pernielsentikaer: We're not all in Ventura actually, only a few 🙂 I think the best approach here would be to add a preference value where you can set the directory path (maybe with a tilde(~) support)

chanceaclark: Done (and done)! 😄 Let me know if there are any other concerns 🙏

extensions/empty-screenshots/src/index.ts
```diff
@@ -0,0 +1,42 @@
...
+
+export default async function main() {
+  const { daysToKeep, screenshotFolder } = getPreferenceValues<Preferences>();
+  const numberOfDays = parseInt(daysToKeep, 10);
+  const folder = resolve(screenshotFolder.replace("~", homedir));
+
+  const files = await readdir(folder);
+
+  const screenshots = files
+    .filter((file) => file.endsWith(".png") || file.endsWith(".mov"))
```
pernielsentikaer, line 19: Could it be an idea to make it a preference which filetypes should be removed? You could have a comma-separated list of extensions (with .mov/.png as default selected)

extensions/empty-screenshots/src/index.ts
```diff
@@ -0,0 +1,42 @@
...
+    .map((file) => join(folder, file));
+
+  const today = new Date();
+  const priorDate = new Date(new Date().setDate(today.getDate() - numberOfDays));
+
+  await showToast({
+    style: Toast.Style.Animated,
+    title: `🗑 Cleaning ${numberOfDays} worth screenshots`,
```
pernielsentikaer, line 27: I'm not sure the sentence make sense?

pernielsentikaer (changes requested):
Hi 👋 

Thanks for doing the corrections 🔥 

I left a few suggestions we could consider, what do you think 🙂

extensions/empty-screenshots/src/index.ts
chanceaclark: Ah, this used to be the number of screenshots it was going to try and delete, but changed it at some point! Good catch! Updated this to `🗑 Cleaning ${numberOfDays} days worth of screenshots`

extensions/empty-screenshots/src/index.ts
pernielsentikaer: Sure!

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

chanceaclark:
Thanks @pernielsentikaer!

Issue #2427, opened 2022-07-29
[Linear] Copy ID and Title in one hotkey


dmastylo:
# Extension – [Linear](https://github.com/raycast/extensions/tree/0aea27a6d2b3bf0f75d8ba2a972004e462c5400f/extensions/linear/)
Author: @thomaslombart

**Describe the feature and the current behavior/state.**
When I track what I've worked on during the day in my other applications, I like to have the ID and the Title (so I have context on what it means) of the linear issue I worked on. I have to do several keystrokes to do this as there are just separate copy actions for ID and title but not together.

EXAMPLE ACTION: Copy ID and Title
OUTPUT:  STE-749 Fix bug on position input component

**Who will benefit with this feature?**
Anyone who wants more context than the ID or just title alone

**Any Other info.**

thomaslombart:
Thanks for the feedback. Right now, the extension has the same copy actions as Linear. 

This copy action seems specific to your workflow and I'm afraid adding it would bloat the actions. I'm thinking instead of something more customizable: a preference on which you can define a template of what to copy (e.g `{ISSUE_ID} {ISSUE_TITLE}`).

Then we could add an option (called "Custom copy" or something similar) so that other users could define their preferred format, regardless of their workflow. What do you think?

dmastylo:
Love that idea!

PR #2627, opened 2022-08-19
feat(linear): add issue custom copy action

thomaslombart:
## Description

Add a custom copy action to the issue actions so that the user can copy to the clipboard their desired format. The list of available keys is:

- `{ISSUE_TITLE}`: the issue's title
- `{ISSUE_ID}`: the issue's identifier
- `{ISSUE_URL}`: the issue's URL
- `{ISSUE_BRANCH_NAME}`: the issue's branch name

Closes https://github.com/raycast/extensions/issues/2427

## Screencast

https://user-images.githubusercontent.com/16003285/185646390-09ebe7a7-cdf6-4854-a7d6-a49045186f2a.mp4

## Checklist

### DIFF
--- a/extensions/linear/package.json
+++ b/extensions/linear/package.json
@@ -123,6 +123,15 @@
       ]
     }
   ],
+  "preferences": [
+    {
+      "name": "issueCustomCopyAction",
+      "type": "textfield",
+      "title": "Issue Custom Copy Action",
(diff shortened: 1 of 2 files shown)


extensions/linear/package.json
```diff
@@ -123,6 +123,15 @@
       ]
     }
   ],
+  "preferences": [
```
thomaspaulmann, line 126: Would it be better to add the preference only to the relevant commands instead of a global one for the extension?

thomaslombart: Since it's an issue-related action, it means the preference is used whenever a command shows an issue list/detail which is the case for Active Cycle, Assigned Issues, Created Issues, Search Issues, Search Projects (project issues), Notifications (when you open the related issue).

In my opinion, users would rather want to customize how they copy and send the issue to teammates on a global level than personalize it for 6 different commands. What do you think?

thomaspaulmann: Yeah, makes sense.

PR #2678, opened 2022-08-25
Apple Mail Extension

yug2005:
## Description

Extension for apple mail. 

Commands : 

- See Mail Accounts: See all accounts and then see all their mailboxes
- See Important Mail: See important mail for all accounts. 
- See Recent Mail: See recent unread messages for all accounts. 
- New Message: Create a new outgoing message. 
- Share With Mail: Select files or folders in Finder to create a new message with those files attached.
- Check For New Mail: Checks for new mail in all accounts. This is a no-view command so right now, it is also used for the background refresh. 

The extension uses the cache and background refresh. 

I need to implement the reply, forward, and redirect actions for messages but other than that, I am almost done with the extension. I just wanted to get some feedback. What other features would you like to see? 

Also, for the background refresh, the process is slightly heavy. It is done daily and it takes around 40s in total to complete. I use applescript to determine all the possible recipients for a message.  This is used for the to, cc, and bcc recipients field in the outgoing message form as these fields are implemented using a tag picker. @pernielsentikaer Do you think this is acceptable for background refresh?

## Screencast

mail 2022-08-25 at 11 20 29

mail 2022-08-25 at 11 18 05

mail 2022-08-25 at 11 18 51

mail 2022-08-25 at 11 19 22

## Checklist

### DIFF
--- a/extensions/mail/src/utils/finder.ts
+++ b/extensions/mail/src/utils/finder.ts
@@ -5,7 +5,7 @@ import { join } from "path";
 import mime from "mime-types";
 
 const isHidden = (item: string) => {
-  return item === "Icon\r" || /(^|[/])[.][^\/\.]/g.test(item);
+  return item === "Icon\r" || /(^|\/)\.[^/.]/g.test(item);
 };
 
 export const getDirectoryItems = async (dir: string): Promise<{ files: string[]; subDirectories: string[] }> => {


pernielsentikaer:
It's a tough one to give a piece of good advice about; once every day for caching data could be fine, I'm more worried about the time it will take, it's around 40 seconds on your machine but could take longer for older machines with more contacts and since it uses AppleScript it sounds like there could be some edge cases where some very good error handling would be necessary.

At least it should also be mentioned in the command description and in the README so the users are aware of the usage of background refreshes and that it works if the user disables it.

What happens if I want to write to a new contact that is not on the list? Using a `TagPicker` at the moment doesn't allow new entries.

PR #2817, opened 2022-09-06
[Spotify Controls] Added a way to change playback device

CryogenicPlanet:
## Description

Add a new command to spotify controls which allows you to change which device spotify is playing on. 

As far as I can tell, there is no way to do this with apple script and needs the spotify api to achieve this.

I have applied for a quota extension from [spotify](https://developer.spotify.com/community/news/2021/05/27/improving-the-developer-and-user-experience-for-third-party-apps/), but in the meantime happy to manually add an email so you can test the extension

## Screencast

https://www.loom.com/share/8c38d912935d4408974f57066cd7811e

image

## Checklist

### DIFF
--- a/extensions/spotify-controls/src/changeDevice.tsx
+++ b/extensions/spotify-controls/src/changeDevice.tsx
@@ -41,7 +41,7 @@ export default function ChangeDevice() {
                   : device.type.toLowerCase() === "tablet"
                   ? Icon.Window
                   : Icon.SpeakerOn,
-              tintColor: device.is_active ? Color.Green : undefined
+              tintColor: device.is_active ? Color.Green : undefined,
             }}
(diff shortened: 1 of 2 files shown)


thomaspaulmann:
@CryogenicPlanet I would suggest to add these commands to the [Spotify Player extension](https://www.raycast.com/mattisssa/spotify-player) given that it already has support for OAuth. Do you need any additional scope for it? 

In the long run, we might even merge both extensions. It seems like both extensions get closer to each other. Also for users, it would be easier to just install one extension and are good to go.

thomaspaulmann:
@CryogenicPlanet I closed the PR for now. Feel free to re-open it when you have integrated it with the other extension. Or create a new one if that's easier.

PR #2834, opened 2022-09-08
Add extension - Orion

Plonq:
## Description

Orion is a new browser (https://browser.kagi.com). This is an extension to interact with it - view bookmarks, reading list, and history.

I borrowed a lot from the [Safari](https://www.raycast.com/loris/safari) extension, except instead of interacting with tabs I added bookmark support, since the official Browser Bookmarks extension doesn't include Orion.

A note - Full Disk Access permission isn't required to access Orion data (unlike Safari), so I have not included the same permission checks and error screens.

## Screencast

https://user-images.githubusercontent.com/7709415/189087744-7bde23ad-104c-4d0b-9cc2-f1d23d49b6bb.mp4

## Checklist

### DIFF
--- a/extensions/orion/package-lock.json
+++ b/extensions/orion/package-lock.json
@@ -8,7 +8,7 @@
       "license": "MIT",
       "dependencies": {
         "@raycast/api": "^1.39.3",
-        "@raycast/utils": "^1.4.1",
+        "@raycast/utils": "^1.4.7",
         "bplist-parser": "^0.3.2"
       },
       "devDependencies": {
@@ -223,9 +223,9 @@
       }
     },
     "node_modules/@raycast/utils": {
-      "version": "1.4.1",
-      "resolved": "https://registry.npmjs.org/@raycast/utils/-/utils-1.4.1.tgz",
-      "integrity": "sha512-f3HBJytXmlkhEd8tD++YP3csv/0ryq9CH4cdSU33xf/OFfMT1LyWS7rcpU8ehcGX0/1Hs7j6rbyDTb9AyxlTLA==",
+      "version": "1.4.7",
+      "resolved": "https://registry.npmjs.org/@raycast/utils/-/utils-1.4.7.tgz",
+      "integrity": "sha512-v0Ba4UnqlUaQfdJBOS9KKhhFUzw3wwgZ6EBVSDg/dJYWOobXWeA6TbjgBBiY9Up5W0ei+IfegbwRICXxehLiww==",
       "dependencies": {
         "content-type": "^1.0.4",
         "dequal": "^2.0.3",
@@ -233,9 +233,6 @@
         "object-hash": "^3.0.0",
         "signal-exit": "^3.0.7",
         "undici": "^5.10.0"
-      },
-      "optionalDependencies": {
-        "sql.js": "^1.7.0"
       },
       "peerDependencies": {
         "@raycast/api": ">=1.39.2"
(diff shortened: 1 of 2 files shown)


pernielsentikaer (changes requested):
Hi 👋 

I have now tested this extension and it works most of the time 😃

- I fixed state flicker in bookmarks and made sure that the file exists so it dies gracefully.

- I can't get the `History` command to work, it just does this, do you know anything about that? 

CleanShot 2022-09-13 at 11 50 25@2x

We also recently added a [useSql](https://developers.raycast.com/utilities/react-hooks/usesql) hook that might be something for you to look into, could that be a better way?

Plonq:
> Hi 👋
> 
> Thanks for your contribution 💪
> 
> I have now tested this extension and it works most of the time 😃

Thanks for checking this out!

> * I fixed state flicker in bookmarks and made sure that the file exists so it dies gracefully.

Oh cheers for that, something I overlooked.

> * I can't get the `History` command to work, it just does this, do you know anything about that?
> 
> CleanShot 2022-09-13 at 11 50 25@2x

Hmm, that error looks to be like it can't find the `better_sqlite3.node` file, which is in assets. Is there any reason an asset file would fail to be found?

> We also recently added a [useSql](https://developers.raycast.com/utilities/react-hooks/usesql) hook that might be something for you to look into, could that be a better way?

I'll definitely check this out.

pernielsentikaer:
Yes, that hook should work, I'm not sure you can the current way to work just like that 🙂

Plonq:
> Yes, that hook should work, I'm not sure you can the current way to work just like that 🙂

I've updated the PR and replaced the custom sql hook with the official one. However, there seems to be an issue with `sql.js` - it shows an outdated version of the database. Sometimes it updates, but I can't figure out what triggers it to get the latest version. I had the same problem when I used `sql.js` manually, which is why I switched to `better-sqlite3`. The Safari extension has the same issue, which also uses `sql.js`. I don't know whether it's a bug in the library or not, but I guess since this extension now uses the official `useSQL` hook, it's probably your (Raycast's) problem now 😛 .

Here's a screenshot showing the issue. Notice zombo.com is not showing in Raycast, even though it's an item in the history.

image

mathieudutour:
Oh that's a super interesting bug! 😄 I'll have a look.

So FYI, the reason why we aren't too keen on native bindings is because it's super hard to support different architecture (eg. the `.node` might work for you if you have an M1 mac, but might not for someone who has an Intel).

mathieudutour:
After a quick investigation in the Safari extension, looks like the history shown in the extension is updated as soon as you quit Safari. That makes me think that there is somehow a lock or that the data is flushed then. I don't really understand it tho because if I open the database file directly, I do see the latest history, even if before Safari "flushed" it.

Plonq:
> Oh that's a super interesting bug! 😄 I'll have a look.
> 
> So FYI, the reason why we aren't too keen on native bindings is because it's super hard to support different architecture (eg. the `.node` might work for you if you have an M1 mac, but might not for someone who has an Intel).

That's fair enough!

>After a quick investigation in the Safari extension, looks like the history shown in the extension is updated as soon as you quit Safari. That makes me think that there is somehow a lock or that the data is flushed then. I don't really understand it tho because if I open the database file directly, I do see the latest history, even if before Safari "flushed" it.

I thought the same thing, but then I couldn't get it to 'flush' consistently by quitting the app. (I was testing with Orion though.) 🤷🏻

mathieudutour:
So... I'm by no mean an SQL expert but I think the issue is that SQL.js (the library the hook is using) can't acquire file locks or memory map files and has to load the entire library in moment. This is obviously not ideal.

I found a way to deal with it in `@raycast/utils@1.4.2`: use the `sqlite3` shipped with macos by default 😄 It's really fast and super lightweight since we don't have to ship anything ourselves.

Plonq:
@mathieudutour Anything else I need to do for this PR? Or are you going to fix the sqlite issue upstream first?

mathieudutour:
Ah sorry, forgot to update here but it should be fixed in the very latest version of the utils

Plonq:
> Ah sorry, forgot to update here but it should be fixed in the very latest version of the utils

Updated to latest utils version, and quick test indicates that fixed the issue

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #2920, opened 2022-09-17
Fix import in docs example

danpalmer:
Minor fix for code example in documentation.


pernielsentikaer (approved):
Thanks for the correction 🔥

PR #2925, opened 2022-09-17
Update some extension icon with a better look and style

RaffeYang:
…size.

## Checklist


thomaspaulmann:
Oh thanks for doing this. Was there a specific workflow that you used to achieve a consistent size?

Jax0rz:
> Oh thanks for doing this. Was there a specific workflow that you used to achieve a consistent size?

I think this [script](https://github.com/elrumo/macOS_Big_Sur_icons_replacements) is very useful.
You could force developers to use fixed size and rounded corner style icons, just like iOS mandates, so that raycast interactions would be more coordinated and uniform,here is a example showcase :
CleanShot 2022-09-19 at 14 47 13@2x

thomaspaulmann (changes requested):
@Jax0rz looks like some images are not the correct size. Can you please fix and re-request a review?

thomaspaulmann (changes requested):
Thanks for fixing it quickly. Looks like you added a new extensions called `chinese-numbers` as a git submodule. Could you please remove the extension? And if you want to publish it, open a new pull request without being it a submodule. Then, we should be ready to merge this change.

Jax0rz (commented):
Update some extension icon

Jax0rz (commented):
Update some extension icon

Jax0rz (commented):
Update some extension icon with a better look and style

Jax0rz (commented):
Add more icon

peduarte:
good effort!

PR #2965, opened 2022-09-20
Updated Any Website Search extension

rben01:
Extension: `extensions/any-website-search`

## Description

The initial release had bug that ignored the selected search suggestion; this PR fixes it. Also added a setting for search suggestions provider and improved how DDG bang-related intelligence works.

### DIFF
--- a/extensions/any-website-search/README.md
+++ b/extensions/any-website-search/README.md
@@ -41,7 +41,7 @@
 _Default: DuckDuckGo_
 
 While entering a query, suggestions can be shown for your query, the same as you'd see
-in a search engine in your browser. This setting controls search engine to get
+in a search engine in your browser. This setting controls the search engine to get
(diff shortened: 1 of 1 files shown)


pernielsentikaer (changes requested):
Hi @rben01 👋 

Can you also add so when I change to another engine in the dropdown, then the search results are refreshed?

rben01:
> Can you also add so when I change to another engine in the dropdown, then the search results are refreshed?

The suggestion always come from the user's configured search suggestions provider: Google or DuckDuckGo (or none). The selected site in the dropdown is only the "destination" and is not used to get suggestions. So there is nothing to refresh when changing the search engine.

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #2967, opened 2022-09-21
Bitwarden

pernielsentikaer:
## Description

Added items folder names if any

## Screencast

CleanShot 2022-09-21 at 07 08 42@2x

### DIFF
--- a/extensions/bitwarden/package.json
+++ b/extensions/bitwarden/package.json
@@ -8,7 +8,8 @@
   "contributors": [
     "daniel-stoneuk",
     "jomifepe",
-    "andreaselia"
+    "andreaselia",
+    "pernielsentikaer"
   ],
   "license": "MIT",
   "keywords": [
@@ -109,7 +110,7 @@
     }
   ],
   "dependencies": {
-    "@raycast/api": "^1.38.1",
+    "@raycast/api": "^1.39.3",
     "execa": "^6.1.0"
   },
   "devDependencies": {


pernielsentikaer:
The next step could be to use a dropdown to filter items + maybe add the folder to the keyword attribute.

andreaselia:
Looks good to me, nice work @pernielsentikaer 👍

extensions/bitwarden/src/search.tsx
```diff
@@ -54,16 +54,18 @@ export default function Search() {
...
     items: [],
+    folders: [],
     isLocked: false,
     isLoading: true,
   });
 
   async function loadItems(sessionToken: string) {
     try {
+      const folders = await bitwardenApi.listFolders(sessionToken);
       const items = await bitwardenApi.listItems(sessionToken);
```
pomdtr, line 67: We should make those two requests in parallel here.
The startup time of the bitwarden extension is already pretty slow, so the last thing we want is too increase it.

pomdtr (changes requested):

pomdtr:
Thanks for the contribution @pernielsentikaer !

pomdtr:
Looks good to me ! Btw we should use the new usePromise api to simplify the code :)

pomdtr:
Also what do you think of adding a dropdown to filter by folder ?

pernielsentikaer:
Sounds like a good idea with usePromise, but I don't think I have enough time to do it at the moment, maybe a project for another?

[Dropdown](https://github.com/raycast/extensions/pull/2967#issuecomment-1253216533) is next step and will be in another PR when I have time, I don't think I can do it now so another might take that too

PR #2980, opened 2022-09-21
revamp gcp extension

pomdtr:
## Description

Simplify the code of the extension, and use a single view for the projects and products lists.

## Screencast

gcp-search-1

## Checklist


pernielsentikaer:
Thanks for the update @pomdtr 🔥 Seems like the metadata images are the wrong size, can you take them on your retina screen?

Can you check this update @klaa97

klaa97:
@pomdtr would it be possible to show in a video how you switch between projects?

My fear of a single step procedure to access the product page is risking to make it harder to switch between projects - especially common in case you use multiple projects usually

pomdtr:
@klaa97 here is a demo. It requires to hit `cmd+p` then choose your project. Since it remember the last selection, on average it saves me clicks

https://user-images.githubusercontent.com/17577332/193237473-0a4331fc-5dab-4d42-8300-45f8a1ec73ed.mov

klaa97:
I am doubtful on this :( but I also have not very clear how most users use it (user interviews needed 😂).. personally I switch pretty often between 10-11 projects and this would be add keystrokes rather than removing them. It could be cool if there could be: 
- a way to go in this page directly 
- A way to first query the list of available tasks
so basically two different commands? 🤔 your is sort of like “I’m feeling lucky” pattern that you care about the last project.

Wdyt? Not opposed to this change to start with

pomdtr:
Maybe a good solution would be to reuse the previous design, but order the project by recently visited ? It would surface often used project on top

klaa97:
That's true absolutely! It is in my backlog but in these months I am quite full 😞  I know that some extensions are doing it already, like the github one, but could probably need some refactors.

Regarding the design itself, it seems to me it'd be more similar to the (for example) `CircleCI` one where you do the same:
- Select a project
- After selecting project, you see last entries

PR #2986, opened 2022-09-22
feat(iTranslate): translate by capture

404J:
## Description

- Add new command `Translate by OCR`

## Screencast

itranslate-2

## Checklist

### DIFF
--- a/extensions/itranslate/package.json
+++ b/extensions/itranslate/package.json
@@ -1255,8 +1255,8 @@
       "type": "checkbox",
       "required": false,
       "label": "Close window when capture",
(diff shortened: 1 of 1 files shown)


pernielsentikaer:
Thanks for your update 🔥 However, if having this setting it doesn't work as it is right now since the extension is unloaded before it's done.

I'm not sure we can work around this at the moment 😕

404J:
@pernielsentikaer Which setting are you talking about?

pernielsentikaer:
@404J strange! I'm sure I included the image I prepared, here it is 😊

image

404J:
@pernielsentikaer This command has a configuration item that defaults value is false and gives some hints when the user wants to change its value. This will prevent the extension from being unloaded before it's done.
CleanShot 2022-09-28 at 16 03 16@2x

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #2987, opened 2022-09-22
New Extension: Switch DNS

glenhooo:
## Description

This extension help you switch DNS quickly.

## Checklist

### DIFF
--- a/extensions/switch-dns/README.md
+++ b/extensions/switch-dns/README.md
@@ -8,3 +8,7 @@ Switch DNS quickly
 
 The settings provide the option to run switch DNS command with sudo (as root).
 If you don't already know what this means or how to do it, you probably shouldn't do it.
+
+## Admin Password
+
+If "Use Sudo" checked, pre-enter admin password for commands.
--- a/extensions/switch-dns/package.json
+++ b/extensions/switch-dns/package.json
@@ -24,6 +24,15 @@
       "description": "See https://github.com/GallenHu/extensions/blob/main/extensions/switch-dns/README.md",
       "default": false,
       "label": "Run Switch DNS commands as sudo"
+    },
+    {
+      "name": "password",
+      "type": "password",
+      "required": false,
+      "title": "Admin Password",
+      "description": "Required if \"Use Sudo\" checked",
+      "default": false,
+      "label": "Admin Password"
     }
   ],
   "dependencies": {
--- a/extensions/switch-dns/src/index.tsx
+++ b/extensions/switch-dns/src/index.tsx
@@ -3,7 +3,7 @@ import { useEffect, useState } from "react";
 import Edit from "./components/edit";
 import DEFAULT_DNS from "./config";
 import { DNSItem } from "./types/types";
-import { switchDNS, getCurrentDNS, useSudo } from "./utils/utils";
+import { switchDNS, getCurrentDNS } from "./utils/utils";
 import StorageUtils from "./utils/storage-utils";
 
 export default function Command() {
@@ -11,24 +11,28 @@ export default function Command() {
   const [loading, setLoading] = useState(false);
   const [customDNS, setCustomDNS] = useState<DNSItem[]>([]);
 
-  const apply = async (item: DNSItem) => {
-    const { error, data } = await switchDNS(item.dns);
+  const apply = (item: DNSItem) => {
+    showToast({
+      title: "Applying",
+      message: "DNS Switching...",
+      style: Toast.Style.Animated,
+    });
 
-    if (error) {
-      if (useSudo) {
-        showHUD(`Failed: ${error}`);
-      } else {
+    setTimeout(async () => {
+      const { error, data } = await switchDNS(item.dns);
+
+      if (error) {
         showToast({
           title: "Failed",
-          message: `Enable "Use Sudo" in the Extension Preferences and try again.`,
+          message: error,
           style: Toast.Style.Failure,
         });
+      } else {
+        await showHUD("DNS changed to " + data);
+        setCurrentDNS(item.dns);
+        popToRoot({ clearSearchBar: false });
       }
-    } else {
-      await showHUD("DNS changed to " + data);
-      setCurrentDNS(item.dns);
-      popToRoot({ clearSearchBar: false });
-    }
+    }, 0);
   };
 
   const loadCustom = async () => {
@@ -95,8 +99,12 @@ export default function Command() {
     });
   }, []);
 
-  return (
-    <List isLoading={loading}>
+  return loading ? (
+    <List isLoading>
+      <List.Item icon={Icon.Repeat} title="Configure Loading..."></List.Item>
+    </List>
+  ) : (
+    <List>
       {customDNS.map((item) => (
         <ListItem key={item.title} isCustom data={item} />
       ))}
--- a/extensions/switch-dns/src/utils/utils.ts
+++ b/extensions/switch-dns/src/utils/utils.ts
@@ -1,56 +1,40 @@
-import util from "util";
-import { exec } from "child_process";
+import { execSync, ExecSyncOptions } from "child_process";
 import { getCachedEnv } from "./shell-utils";
 import { DEFAULT_DNS } from "../config";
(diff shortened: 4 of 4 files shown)


extensions/switch-dns/src/config.ts
```diff
@@ -0,0 +1,9 @@
+export const DNS_LIST = [
+  { title: "Alibaba", dns: "223.5.5.5,223.6.6.6", accessory: "alidns.com" },
+  { title: "Google", dns: "8.8.8.8,8.8.4.4", accessory: "Google" },
+  { title: "Cloudflare", dns: "1.1.1.1,1.0.0.1", accessory: "Cloudflare DNS" },
+  { title: "114", dns: "114.114.114.114,114.114.115.115", accessory: "114 Public DNS" },
+  { title: "Empty DNS", dns: "Empty", accessory: "Empty DNS" },
```
danpalmer, line 6: I'd probably call this "Revert to network default" – there are always DNS servers set, for most networks, but this is often provided by default by the network. It'll normally be the ISP.

A common use case I have for changing DNS is when I connect to a public WiFi network that doesn't support custom DNS, so I want to revert to the network default DNS.

extensions/switch-dns/src/config.ts
```diff
@@ -0,0 +1,9 @@
+export const DNS_LIST = [
+  { title: "Alibaba", dns: "223.5.5.5,223.6.6.6", accessory: "alidns.com" },
+  { title: "Google", dns: "8.8.8.8,8.8.4.4", accessory: "Google" },
+  { title: "Cloudflare", dns: "1.1.1.1,1.0.0.1", accessory: "Cloudflare DNS" },
+  { title: "114", dns: "114.114.114.114,114.114.115.115", accessory: "114 Public DNS" },
```
danpalmer, line 5: I think this set of choices could perhaps be improved.

I'm not familiar with the Chinese market, and my Chinese is very rusty, but Alibaba and 114 are almost impossible to find any details for on the English speaking public internet. Are these services accessible outside of China? Are they safe to use outside of China?

Separately, the list is missing OpenDNS, which has historically been a well known and trustworthy provider, they were the first custom DNS provider I used probably 15 years ago. I'd have expected a list of options to include them.

However... these issues lead me to think about whether this is the best way to configure this extension. Do we want to add DNS servers for the Indian market? Do we want to add DNS servers for the Russian, Iranian, or North Korean markets that might be sanctioned and having difficulty using others? This all gets difficult and political.

Because of this, I'd propose an alternative: Have a setting for users to provide the custom DNS they want to use, and then select from those options. I think this avoids the issue of which DNS servers are worthy of inclusion, and also prevents users from accidentally setting DNS servers that they may not want to use with one mis-click.

What do you think?

danpalmer (commented):
I'm not a maintainer, but saw this extension, I'm keen for the functionality and hope that my comments help refine it.

extensions/switch-dns/src/config.ts
GallenHu: Thanks for your suggestion. It's my mistake not considering international universality. Originally, custom configuration is my plan. I will submit a new commit with those changes:  
- Provide `Google`, `Cloudflare`, `Default DNS` as default options
- Support user custom configuration

I think that "one-click switching " is useful for most users (installed this extention). Maybe I don't want extra confirmation to slow down my operation. How do you think?

danpalmer: Sounds great!

pernielsentikaer:
Hi @GallenHu 👋 

I'm on Ventura and it doesn't seem to work as expected here, when opening the commands I see this error (but the extension looks as it should(?))

CleanShot 2022-09-27 at 21 58 54@2x

```
21:59:11.541 Error: Command failed: networksetup -getdnsservers "Wi-Fi" | grep -v 'any'
    at checkExecSyncError (node:child_process:826:11)
    at execSync (node:child_process:900:15)
    at null.exec (/Users/pernielsentikaer/Documents/Raycast/tmp/switch-dns/src/utils/utils.ts:8:20)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at null.getCurrentDNS (/Users/pernielsentikaer/Documents/Raycast/tmp/switch-dns/src/utils/utils.ts:35:15) {
  status: 1,
  signal: null,
  output: [ null, <Buffer >, <Buffer > ],
  pid: 5019,
  stdout: <Buffer >,
  stderr: <Buffer >
}
```

When trying to set a DNS it show me the HUD saying that it's done, but nothing happens
```
21:59:31.331 networksetup -setdnsservers "Wi-Fi" 208.67.222.222 208.67.220.220
21:59:31.331 Error: Command failed: networksetup -setdnsservers "Wi-Fi" 208.67.222.222 208.67.220.220
    at checkExecSyncError (node:child_process:826:11)
    at execSync (node:child_process:900:15)
    at null.exec (/Users/pernielsentikaer/Documents/Raycast/tmp/switch-dns/src/utils/utils.ts:8:20)
    at null.switchDNS (/Users/pernielsentikaer/Documents/Raycast/tmp/switch-dns/src/utils/utils.ts:49:3)
    at null.apply (/Users/pernielsentikaer/Documents/Raycast/tmp/switch-dns/src/index.tsx:17:17) {
  status: 14,
  signal: null,
  output: [
    null,
    <Buffer 2a 2a 20 45 72 72 6f 72 3a 20 43 6f 6d 6d 61 6e 64 20 72 65 71 75 69 72 65 73 20 61 64 6d 69 6e 20 70 72 69 76 69 6c 65 67 65 73 2e 0a>,
    <Buffer >
  ],
  pid: 5085,
  stdout: <Buffer 2a 2a 20 45 72 72 6f 72 3a 20 43 6f 6d 6d 61 6e 64 20 72 65 71 75 69 72 65 73 20 61 64 6d 69 6e 20 70 72 69 76 69 6c 65 67 65 73 2e 0a>,
  stderr: <Buffer >
}
```

Do you have an idea what we can try?

GallenHu:
Hi, @pernielsentikaer 
How about trying to run this script on your terminal ?
```sh
networksetup -setdnsservers "Wi-Fi" 208.67.222.222 208.67.220.220
```
It works on my machine.

GallenHu:
This error (`Command failed: networksetup -getdnsservers "Wi-Fi" | grep -v 'any'`) is only displayed when DNS is empty, I'm thinking of catching and hiding it to avoid misunderstanding.

pernielsentikaer:
I'm getting this error

```
networksetup -setdnsservers "Wi-Fi" 208.67.222.222 208.67.220.220
** Error: Command requires admin privileges.
```

Might be related to Ventura, we had the same issue in #2857

GallenHu:
> Might be related to Ventura, we had the same issue in #2857

According to this issue, this extension also add an option to enable run with `sudo`. Plz try again. @pernielsentikaer

pernielsentikaer (changes requested):
Thanks for the update 🔥 

It works as expected, well done! I notice two things (not sure we can do anything about it)

1) When the password box is visible Raycast closes, when having this setting I don't see the HUD when it's changing
2) It could be cool if data first returned when both custom and the default list is received, it feels a bit funny as it is right now

Can you look into above, let me know if you need any help 😊

GallenHu:
I don't know why `sudo-prompt` (and `child_process.exec` with `sudo` command) do not trigger callback. And sudo-prompt mentioned that [never use sudo to launch any graphical applications](https://github.com/jorangreef/sudo-prompt#non-graphical-terminal-commands-only) , Maybe [`passwordless use of the sudo command`](https://github.com/raycast/extensions/tree/9d73618c41fda5d66c26016dc65e102a012414a1/extensions/port-manager/#use-sudo) or Preset-Sudo-Pasword in preferences a workable solution. Any suggestion here?

GallenHu:
Hi, @pernielsentikaer, I changed extension preferences,  please try again.

pernielsentikaer:
It doesn't feel right to have the `sudo` password directly in preferences, this can cause a lot of problems if something changes in the code. We recently added [LaunchCommand](https://developers.raycast.com/api-reference/utilities#launchcommand) which might be able to keep it alive - could we try that out (and do you have time)

PR #3028, opened 2022-09-26
[New Feature: Lorem Ipsum] update api/deps, allow support for multiple sentences/paragraphs

GastroGeek:
## Description

* General maintenance (also added a CHANGELOG)
* Add support for generating multiple sentences/paragraphs

## Checklist


pernielsentikaer:
Do you have time to check this @AntonNiklasson

extensions/lorem-ipsum/package.json
```diff
@@ -6,20 +6,53 @@
   "icon": "paragraph-icon.png",
   "author": "AntonNiklasson",
   "license": "MIT",
+  "contributors": [
+    "GastroGeek"
+  ],
+  "categories": [
+    "Data",
+    "Design Tools",
```
AntonNiklasson, line 14: Does this really fit in `Design Tools`, `Web`, and `Developer Tools`? Thoughts @pernielsentikaer ?

extensions/lorem-ipsum/src/utils.tsx
```diff
@@ -1,6 +1,9 @@
-import { closeMainWindow, copyTextToClipboard, pasteText, showHUD } from "@raycast/api";
+import { closeMainWindow, Clipboard, showHUD } from "@raycast/api";
 import { LoremIpsum } from "lorem-ipsum";
 
+// don't want to cause a heap error, so cap it 😱
+const LOREM_IPSUM_MAX_NUMBER = 10000;
```
AntonNiklasson, line 5: If this is a risk, perhaps 10k is still a bit high?

extensions/lorem-ipsum/src/utils.tsx
```diff
@@ -15,27 +18,63 @@ const generator = new LoremIpsum({
 // generator.generateWords(1);
 // generator.generateParagraphs(7);
 
-export const generateParagraph = () => {
-  return generator.generateSentences(10);
+export const generateParagraphs = (count: number) => {
+  return Array.from(Array(count))
+    .map(() => generator.generateSentences(6))
```
AntonNiklasson, line 23: This hardcoded 6 makes the parargraphs less dynamic imo. You think we could have this be a random number between say 4 and 10?

AntonNiklasson (commented):
Reading the code as a first pass, I'll run the changes locally when I some more time 😊

extensions/lorem-ipsum/package.json
GastroGeek: Just my 2c... **yes**, since as a developer and/or designer I often need to generate lots of lorem text as placeholder text... 🤷 both of which often relate to 'web'...

extensions/lorem-ipsum/src/utils.tsx
GastroGeek: Simple enough to reduce - whatever you feel approproate/realistic without being restrictive.

extensions/lorem-ipsum/src/utils.tsx
GastroGeek: It was set as default/_hard-coded_ to 10 before via `return generator.generateSentences(10);` 😕 😛 

Again, simple enough fix - go ahead and tweak to randomize if desired.

extensions/lorem-ipsum/package.json
pernielsentikaer: I think we can keep it as a Design Tool too 😊

extensions/lorem-ipsum/src/utils.tsx
pernielsentikaer: Who will do the tweaks, and yes, it was hardcoded before as mentioned.

Are we going for something like `Math.floor(Math.random() * 7) + 4;`?

extensions/lorem-ipsum/src/utils.tsx
pernielsentikaer: Might also depend on the command, what about 2k then? Most users will do 1 anyways 😊

extensions/lorem-ipsum/src/utils.tsx
GastroGeek: Seems reasonable from my PoV... does not need to be _truly_ random 🤓 

Or perhaps re-use the options for LoremIpsum lib?

```js
const loremIpsumOptions = {
  sentencesPerParagraph: {
    max: 8,
    min: 4,
  },
  wordsPerSentence: {
    max: 16,
    min: 4,
  },
};

const generator = new LoremIpsum(loremIpsumOptions);
```

...

```js
export const generateParagraphs = (count: number) => {
  return Array.from(Array(count))
    .map(() =>
      generator.generateSentences(
        Math.floor(
          Math.random() *
            (loremIpsumOptions.sentencesPerParagraph.max - loremIpsumOptions.sentencesPerParagraph.min + 1)
        ) + loremIpsumOptions.sentencesPerParagraph.min
      )
    )
    .join("\r\n\r\n"); // newline + seperator line
};
```

AntonNiklasson: If it was hardcoded before, I'm fine with just keeping it as is! It's been a while since I wrote that, I don't remember all the details. Looks great!

extensions/lorem-ipsum/package.json
AntonNiklasson: Great, makes a lot of sense! ✨

extensions/lorem-ipsum/src/utils.tsx
pernielsentikaer: If we have that already added in the code it might be fine just reusing that @GastroGeek, can you do that?

GastroGeek: Well, that is my change to make that so its accessible but sure, will push up some tweaks in relation to this convo

GastroGeek: Done! @pernielsentikaer and updated CHANGELOG too...

AntonNiklasson:
I'm not sure I like the idea of a manual CHANGELOG like this. Are other extensions doing this? If it's a common pattern, perhaps a centralized solution like [conventional changelog](https://github.com/conventional-changelog/standard-version) for this repo would be more scalable approach, @pernielsentikaer?

pernielsentikaer:
Hi @AntonNiklasson 👋 

That's the way every [extension is doing it](https://developers.raycast.com/basics/prepare-an-extension-for-store#version-history), the changelog will be shown in the app and in the store.

You can see an [example here](https://www.raycast.com/thomaslombart/todoist/changelog)

AntonNiklasson:
> That's the way every extension is doing it
Alright, cool! I haven't kept up to speed here ☺️

AntonNiklasson (approved):
Great job, I'm looking forward to using this 🌟

pernielsentikaer (approved):

PR #3079, opened 2022-09-30
New Extension: YubiKey Code

yoyoyeti:
## Description

A new extension to help get OTPs from your YubiKey.

## Screencast

https://user-images.githubusercontent.com/9059772/193303916-919b78a1-5c42-4437-a944-68e4f63cbd02.mp4

## Checklist

### DIFF
--- a/extensions/yubikey-code/CHANGELOG.md
+++ b/extensions/yubikey-code/CHANGELOG.md
@@ -1,3 +1,3 @@
 # YubiKey Code Changelog
 
-## [Initial Version] - 2022-10-04
+## [Initial Version] - 2022-11-18


pernielsentikaer (changes requested):
Hi 👋 

Seems to do the job as expected, but you could maybe look into finding the brew path automatically like the [Brew extension](https://www.raycast.com/nhojb/brew) is doing it

What do you think about that?

yoyoyeti:
> Seems to do the job as expected, but you could maybe look into finding the brew path automatically like the [Brew extension](https://www.raycast.com/nhojb/brew) is doing it
> 
> What do you think about that?

That seems like a good improvement. I'll try to get that worked in when I have a chance.

yoyoyeti:
@pernielsentikaer Alright, so now that I've had a chance to sit down and look at this for a minute, I have a potentially silly question. I believe the code you were referring to is this bit [here](https://github.com/raycast/extensions/blob/dd6d1181a4d488063098d5561759026bafd36f6d/extensions/brew/src/brew.ts#L103), but when I try to run that locally, it fails with "command not found" and always defaults back to the `cpus()[0].model.includes("Apple") ? "/opt/homebrew" : "/usr/local"` line.

I'm able to run `brew --prefix` through the terminal with no issues, so is there something about my local dev environment which could cause this, or is it possible that this code isn't working in the Brew extension either? If that is the case, I can just add the `cpus()[0].model.includes("Apple")` check as a sensible default with the option to set a custom path through the preferences. Let me know what you think.

pernielsentikaer:
Hi 👋 

I gave it another spin, I have some feedback for you 😊

* It would be useful if it told you that `ykman` doesn't exist instead, since right now it can be hard to understand what's going on - maybe even combine with a nice List.Emptyview 😊

image

Do you have some ideas for that?

yoyoyeti:
@pernielsentikaer  thanks for the helpful feedback. I made some updates to try to improve the things you mentioned. Let me know what you think 😊

danpalmer:
@yoyoyeti 

> @pernielsentikaer Alright, so now that I've had a chance to sit down and look at this for a minute, I have a potentially silly question. I believe the code you were referring to is this bit [here](https://github.com/raycast/extensions/blob/dd6d1181a4d488063098d5561759026bafd36f6d/extensions/brew/src/brew.ts#L103), but when I try to run that locally, it fails with "command not found" and always defaults back to the cpus()[0].model.includes("Apple") ? "/opt/homebrew" : "/usr/local" line.

I think that code is incorrect, or at least incomplete. I think you need to pass `{ shell: true }` to `execSync`. This indicates that the command should be run via the user's shell, rather than directly. When not running in a shell, as is the default, `brew` can't be found because there is no `PATH` to be searched, or if there is it doesn't have `/usr/local/bin` on it.

It's probably better to pass a full path to `/usr/local/bin/brew` if it's always installed in the same place as using a shell is quite a bit heavier weight, but up to you to judge.

danpalmer:
Apologies, slightly wrong in my previous comment, `execSync` doesn't use the _user's_ shell, but rather `/bin/sh` by default. This might be enough to see binaries at `/usr/local/lib` though so I'd give it a go.

yoyoyeti:
@danpalmer Thanks for getting back to me on that. I'm getting an error that looks more like `/bin/sh: bin/brew: No such file or directory` with `{ shell: true }`, but still no luck. 

Even if I could make this work, I found that adding the extra command to find the brew prefix slowed down the workflow enough to make it a little frustrating to use. I feel like the current setup is faster and will work out of the box for most people. There's also the option to override the path for anybody who needs it. I'm willing to keep trying if you have any more suggestions, but do you think the extension would be fine in its current state?

yoyoyeti:
@pernielsentikaer is there anything else this needs before it can be merged?

pernielsentikaer (approved):

PR #3086, opened 2022-10-01
add aws-ecr extension

abelkuruvilla:
## Description

Added a new extension AWS ECR to update ECR Images

## Screencast
ecr-tag-change-2

Screen Shot 2022-10-01 at 8 22 58 PM

## Checklist

### DIFF
--- a/extensions/aws-ecr/package-lock.json
+++ b/extensions/aws-ecr/package-lock.json
@@ -1,7 +1,5393 @@
 {
   "name": "ecr-tag-change",
-  "requires": true,
   "lockfileVersion": 2,
+  "requires": true,
+  "packages": {
+    "": {
+      "name": "ecr-tag-change",
+      "license": "MIT",
+      "dependencies": {
+        "@raycast/api": "^1.40.1",
+        "@raycast/utils": "^1.4.8",
+        "aws": "0.0.3-2",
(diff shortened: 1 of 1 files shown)


pernielsentikaer:
Hi @abelkuruvilla 👋 

Thanks for the contribution 💪 

Would it make sense to add it to [this extension](https://www.raycast.com/Falcon/aws) instead so we keep only having one extension?

abelkuruvilla:
@pernielsentikaer This extension works with aws cli, whereas the other one works with aws SDK. I wanted to use SDK, but it does not support the `query` param that cli has for describe-images. Do you still think its best to combine both ?

pernielsentikaer:
Then it might be a good idea to split it, otherwise, it would be a weird UX.

Maybe one of @victor-falcon @Hodglim @JonathanWbn @gebeto @momme-rtf @duboiss @hexpl0it knows how to do what you're doing here in the SDK?

JonathanWbn:
I also agree that this behaviour can be integrated with the existing `amazon-aws` extension. This could be a "List ECR Repositories" command with an action to add a tag.

Even if the SDK doesn't allow for the desired behaviour, I don't see a problem with mixing SDK + CLI operations in the same extension.

PR #3108, opened 2022-10-03
[New Extension]: Airsy

piotrzaborow:
## Description
Airsy is an air pollution checker in Polish cities.
It gathers data from GIOS API: https://powietrze.gios.gov.pl/pjp/content/api.
It shows data for all stations in Poland and for selected city (needs to be written in preferences).

## Screencast

https://user-images.githubusercontent.com/5574918/193596502-7ee8250c-c4a4-413c-b4e1-715b5916b9da.mp4

## Checklist

### DIFF
--- a/extensions/airsy/src/components/IndexDetails.tsx
+++ b/extensions/airsy/src/components/IndexDetails.tsx
@@ -16,7 +16,7 @@ const IndexDetails: FC<{ title: string; indexValue: string | undefined; value?: 
       <List.Item.Detail.Metadata.TagList title="Index">
         <List.Item.Detail.Metadata.TagList.Item text={text} color={color} />
       </List.Item.Detail.Metadata.TagList>
-      {value && (
+      {typeof value === "number" && (
         <List.Item.Detail.Metadata.TagList title="Value">
           <List.Item.Detail.Metadata.TagList.Item text={value ? parseFloat(value + "").toFixed(5) : "Unknown"} />
         </List.Item.Detail.Metadata.TagList>


extensions/airsy/CHANGELOG.md
```diff
@@ -0,0 +1,9 @@
+# Airsy Changelog
+
+## [0.1.0] - 2022-10-03
+### Added GIOS API as a service
+### Added city selection in preferences
+### Preparation for future translations
+
+## [Initial Version] - 2022-09-29
+
```
pernielsentikaer, line 9:
```suggestion
# Airsy Changelog

## [0.1.0] - 2022-10-03
- Added GIOS API as a service
- Added city selection in preferences
- Preparation for future translations

## [Initial Version] - 2022-09-29

```

extensions/airsy/src/index.tsx
```diff
@@ -0,0 +1,18 @@
+import { List } from "@raycast/api";
+import useStations from "./services/GIOS";
+import StationDetails from "./components/StationDetails";
+import { IStationResult } from "./types";
+
+export default function Command() {
+  const { stations, isLoading } = useStations();
+  return (
+    <List navigationTitle={`Airsy - AllStations`} isLoading={isLoading} isShowingDetail>
```
pernielsentikaer, line 9:
```suggestion
    <List isLoading={isLoading} isShowingDetail>
```
Don't change navigationTitle in the root command

pernielsentikaer (changes requested):
Hi 👋 

I tested your extension and have some feedback for you 😊

- You could use [useFetch](https://developers.raycast.com/utilities/react-hooks/usefetch) from @raycast/utils - with this you can get the response cached next time you open the list and refreshing in the background (`initialData: true`) 

- I never get any data from the two commands, are they really slow to respond?

- Seems like the list from `All Stations` crashes if you scroll a bit down
```
StationAddress:StationDetails.tsx:11:39

---
 8:   station: IStationResult;
 9: }
10: 
11: const StationAddress: FC<IProps> = ({ station }) => {
12:   const city = station.city.name;
13: 
14:   const ad = station.city.addressName.split("/")[0];
---
```

piotrzaborow:
I will apply the suggestions.

Just one question about `useFetch`. 
As my flow works that way:
 -> fetch stations from 1st endpoint
 -> map through stations to get sensors data
 -> map through sensors to get values for air pollution

I want to wait for all data to be get and then display it. 
Any suggestion how to do it with `useFetch`?

With Promises, it is simple as `Promise.all`, with `useFetch` hook, I will constantly need to check if data has arrived and if none of the responses are errors. 

I can cache the data, but I would rather stick with Promises, they are more robust. Fetch API is also coming in Node.js 18, so there will no need to have any external deps like `node-fetch`.

mathieudutour:
You're right that `useFetch` might not be the most appropriate here since you need multiple calls - but you could use `useCachedPromise` to cache the results

mathieudutour:
it's cached forever, and refreshed when you open the command. It's there to make sure you don't have to wait for an api call before showing some data

piotrzaborow (commented):
I've used useCachedPromise in my own hook `useStations`. That should do the trick, and it is very React-way 😄

extensions/airsy/src/components/IndexDetails.tsx
```diff
@@ -0,0 +1,29 @@
...
+  const text = translateAndColor(indexValue).text;
+  const color = translateAndColor(indexValue).color;
+
+  return (
+    <>
+      <List.Item.Detail.Metadata.Label title={title} />
+      <List.Item.Detail.Metadata.TagList title="Index">
+        <List.Item.Detail.Metadata.TagList.Item text={text} color={color} />
+      </List.Item.Detail.Metadata.TagList>
+      {value && (
```
pernielsentikaer, line 19:
```suggestion
      {typeof value === "number" && (
```
Should fix the error: A `Metadata` only accepts `Metadata.Label`, `Metadata.Link`, `Metadata.Separator` and `Metadata.TagList` children. Ignoring "string".

pernielsentikaer (changes requested):
Hi 👋 

Could you fix this last error and then I think we're ready to proceed 😊

piotrzaborow:
Done :)

pernielsentikaer (approved):

PR #3109, opened 2022-10-03
[Notion] Support Status properties

### LINKED ISSUE #2412: [Notion] Could we have the support for Notion's newest status property?
# Extension – [Notion](https://github.com/raycast/extensions/tree/8ea3c6f9563bf73663218d2ead8c32e683ef7b34/extensions/notion/)
Author: @ice-Yaya

### Describe the feature and the current behavior/state.
Loving 2 current functions to support Kanban view and customising visible properties for when searching for a database.
But the new status property is not an option in setting up the Kanban and choosing what to see.
Hope this could be supported.

### Who will benefit with this feature?
As Notion launched this new status property feature, I believe **more and more user will transform their initial single-selection into status**. This will more relevant to the Kanban view in this Raycast extension.

### Any Other info.
- [Notion help center on status property](https://www.notion.so/help/guides/status-property-gives-clarity-on-tasks)
- [Notion API info on status property](https://developers.notion.com/reference/property-value-object#status-property-values)

jasonappah:
## Description

This PR currently adds support for creating and editing pages which use the new Status property type. Marking as draft since there are a few other tasks I need to tackle.

To-do:
- [ ] display Status properties on pages in the list view
- [ ] add the Status property as an option for driving kanban boards
- [ ] Screencast

Closes #2412.

## Screencast

TO-DO

## Checklist

### DIFF
--- a/extensions/notion/src/components/PageListItem.tsx
+++ b/extensions/notion/src/components/PageListItem.tsx
@@ -65,7 +65,7 @@ export function PageListItem(props: {
   }
 
   const quickEditProperties = databaseProperties?.filter((property) =>
-    ["checkbox", "select", "multi_select", "people"].includes(property.type)
+    ["checkbox", "select", "multi_select", "people", "status"].includes(property.type)
   );
 
   const visiblePropertiesIds: string[] = [];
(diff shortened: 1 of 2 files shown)


jasonappah:
This is likely going to be on hold for a bit. I've discovered what appears to be an issue with Notion's API where pages with status properties return null for those properties if they haven't been manually set or updated, despite being given a default value. Updating those status properties, then returning them to their original state gets the API to return the expected response.

PR #3146, opened 2022-10-06
[Docs] Fix wrong command name in changelog

tonka3000:
## Description

Fix wrong command name in changelog


mathieudutour (approved):
Good catch!

PR #3148, opened 2022-10-06
Downloads Manager: Added command "Reveal Latest Download in Finder"

infused-kim:
## Description

Added the command `Reveal Latest Download in Finder` to the `Downloads Manager` extension.

## Checklist

### DIFF
--- a/extensions/downloads-manager/CHANGELOG.md
+++ b/extensions/downloads-manager/CHANGELOG.md
@@ -1,3 +1,7 @@
 # Downloads Manager Changelog
 
+## [Show in Finder] - 2022-10-08
+
+Added the command `Show Latest Download`, which reveals the latest download in Finder.
+
 ## [Initial Version] - 2022-08-22


extensions/downloads-manager/src/reveal-latest-download.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+  if (downloads.length < 1) {
+    await showHUD("No downloads found");
+    return;
+  }
+
+  const latestDownload = downloads[0];
+
+  try {
+    revealFile(latestDownload.path);
```
thomaspaulmann, line 15: You should be able to use our built-in [showInFinder](https://developers.raycast.com/api-reference/utilities#showinfinder) function.

```suggestion
    await showInFinder(latestDownload.path);
```

thomaspaulmann (changes requested):
Nice addition, thanks! I left a few suggestions. I was torn apart naming the command to `Show Latest Download` to match better our `Show in Finder` action but wasn't sure. Any thoughts? Alternatively, you could add a keyword "show" to the command to make sure that others would find it easily in the root search.

infused-kim:
@thomaspaulmann thank you for the suggestions. I agree with everything and pushed a commit implementing them.

thomaspaulmann:
Looks like you have an issue with code formatting. Could you run prettier locally again and fix the issue? Then we're ready to merge.

infused-kim:
It should be good to go now. I have also updated the changelog.

thomaspaulmann:
Thanks, let's ship it! Also, would be great if you can add yourself as `contributor` in the `package.json`. This way you see the extension on your Raycast profile ;-)

PR #3157, opened 2022-10-07
Add extension to search Font Awesome icons

dutzi:
## Description

Added a command - Search Font Awesome Icons.

Similar to the Emoji picker, it displays a grid of Font Awesome icons. You can search for icons by name, and select one to insert it into the current document.

Each grid item (icon) has to actions: 

1. Copy as SVG
2. Open In Browser (opens the icon's page on https://fontawesome.com/)

This extension adds 6000 files to the repo (the PNGs and SVGs of the icons). I'm not sure that's the way to go, if that's not possible I'd appreciate guidance as to what's the best way to serve all the icons. 

## Screencast

https://share.cleanshot.com/fyIgKK

## Checklist

### DIFF
--- a/extensions/fontawesome/CHANGELOG.md
+++ b/extensions/fontawesome/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Font Awesome Changelog
 
-## [Initial Version] - 2022-10-08
+## [Initial Version] - 2022-10-16
 
 - Removed static icons files
 - Added iconography fetching from Font Awesome API


DanielFerguson:
This is awesome, and would be a really useful tool. Since FontAwesome icons are already hosted on a CDN, I would definitely think about pulling a list from somewhere and then just rendering the images dynamically. That way you won't need to keep it up to date manually.

DanielFerguson:
I'd take a look at the [FontAwesome GraphQL] (https://fontawesome.com/docs/apis/graphql/get-started) API to start with, and go from there.

dutzi:
Thanks! I did consider the GraphQL API, and it would be useful to get the SVGs for the Copy as SVG command, but in order to display the images in the grid I need them as PNG, which isn't provided by Font Awesome.

I ran a script locally to convert all the SVGs to PNGs, which creates ~3000 files, and that's only for the dark theme, so double that if we want them to render nicely in a light theme.

Also, I have a payed account in Font Awesome, which gives you access to ~20,000 icons, I was hoping to extend the extension, allowing paying customers to sign in and get access to all of those as well.

I could either:

1. host them myself, but that could be costly, or
2. created a 8,215x8,215 sprite file (`150 * sqrt(3000)`) and slice it up locally, or
3. somehow find a way to convert SVGs to PNGs locally, without stirring up puppeteer or something like that?
 
wdyt?

dutzi:
I realized I could just tar the files... I'll do that :)

dutzi:
Hm, the TAR file weighs ~15 MB (only for the regular pack in dark mode). 

Would that be approved?

DanielFerguson:
I'd have no idea, but that's pretty big. As well, you don't want to be responsible for updating it for every FA icon change or addition. Plus, if you compress the files, clients would probably need to download a whole new update.

I wouldn't host the files yourself either - that doesn't get rid of the issue of manual updates, unless you wrote some sort of cron job; however that's still something external to maintain.

I'm going to try something...

DanielFerguson:
@dutzi Could you please add my to your fork repository? I've got it working with the CDN and API.

dutzi:
Cool! Added.

DanielFerguson:
@dutzi Have a look and let me know your thoughts. Remember not to commit the `yarn.lock` file - the Raycast CI pipeline doesn't like it.

dutzi:
That's great!

dutzi:
Added support for dark mode!

font-awesome 2022-10-08 at 18 22 08

A bit of an overkill, but the way it works is I set up a simple HTTP server that proxies requests to Font Awesome and, if requested by the client (the extension), sets the SVGs fill color to white.

DanielFerguson:
Beautiful, love it. Keen to get this in the store. Good job! 🙏

extensions/fontawesome/src/index.tsx
```diff
@@ -0,0 +1,144 @@
...
+    thin: 'Thin',
+  }[iconStyle];
+}
+
+export default function Command() {
+  const port = useServer();
+  const { iconStyle } = getPreferenceValues<Preferences>();
+
+  const getSvgUrl = (icon: Icon, iconStyle: IconStyle, isDark?: boolean): string => {
+    return `http://localhost:${port}/${getLibraryType(icon, iconStyle)}/${icon.id}.svg?dark=${
```
pernielsentikaer, line 55: How does this work?

DanielFerguson: @dutzi Did you forget to point it to the deployment server? @pernielsentikaer This is supposed to be an intermediary server that handles requests for SVGs. Its pure function is to alter the SVG response to change the color to white, if requested.

dutzi: 1. Raycast grid items accept SVG urls and render whatever is served by that URL
2. For an SVG to render in "white" we need to set the SVG's `fill` to "white"
3. So we need a way to return a modified SVG, given its URL
4. The `useServer` hook sets up a local server that serves Font Awesome icons
5. It starts by listening to the first free port available
6. It adds `fill="white"` to the served SVG, if requested

pernielsentikaer (changes requested):
Thanks for the explanation 💯 

When using the `Open in Browser` action I land here:

* https://fontawesome.com/,,,icons/alipay?s=solid&f=classic

Seems like there is a bug in the path? 😊

DanielFerguson:
@pernielsentikaer Are you sure you're running the latest version of this plugin? I just tried on a number of icons and it's working as intended.

dutzi:
(Same here)

pernielsentikaer:
I'm not sure why I didn't get that, just checked again, all good 😊

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

holman:
This is insanely awesome- thanks for who worked on it!

(If there's more desire to dig into it some more, it'd be lovely to have a quick setting to press enter and copy the FA slug rather than the SVG itself, but that's just a wishful thinking thing. :) )

dutzi:
Thanks! Added :)

PR #3215, opened 2022-10-13
feat(linear): Add customization for creating issues

### LINKED ISSUE #2426: [Linear] I would like more copy actions, including copy link, on the `CMD-T` actions after I create an issue
# Extension – [Linear](https://github.com/raycast/extensions/tree/0aea27a6d2b3bf0f75d8ba2a972004e462c5400f/extensions/linear/)
Author: @thomaslombart

I frequently will want to copy a link after creating an issue in Linear to send to someone.

**Who will benefit with this feature?**
Anyone who shares links with co-workers on issues they've created through RayCast.

**Any Other info.**

thomaslombart:
## Description

- Add a preference to select the toast copy action after creating the issue
- Add a preference to automatically focus the title field
- Fix a bug where the sections were not ordered when searching
- Refactor the project and issue edition to use `useForm`

Closes #2426 
Closes #3144

## Screencast

CleanShot 2022-10-13 at 13 43 04@2x

## Checklist

### DIFF
--- a/extensions/linear/package.json
+++ b/extensions/linear/package.json
@@ -32,13 +32,22 @@
           "default": true
         },
         {
-          "name": "autofocusTitle",
-          "type": "checkbox",
-          "title": "Autofocus",
-          "label": "Automatically focus title field",
-          "description": "When enabled, the title field will be automatically focused once the form is rendered.",
+          "name": "autofocusField",
+          "title": "Autofocus Field",
+          "type": "dropdown",
           "required": false,
-          "default": true
+          "description": "Select the field to focus automatically once the form is rendered",
+          "default": "team",
+          "data": [
+            {
+              "title": "Team",
+              "value": "teamId"
+            },
+            {
+              "title": "Title",
+              "value": "title"
+            }
+          ]
         },
         {
           "name": "copyToastAction",
--- a/extensions/linear/src/components/CreateIssueForm.tsx
+++ b/extensions/linear/src/components/CreateIssueForm.tsx
@@ -55,7 +55,7 @@ export type CreateIssueValues = {
 
 type Preferences = {
   signature: boolean;
-  autofocusTitle: boolean;
+  autofocusField: "teamId" | "title";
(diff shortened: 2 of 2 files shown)


pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

extensions/linear/package.json
```diff
@@ -30,6 +30,37 @@
           "description": "When enabled, \"Created via Raycast\" signature will be added at the bottom of the description.",
           "required": false,
           "default": true
+        },
+        {
+          "name": "autofocusTitle",
+          "type": "checkbox",
+          "title": "Autofocus",
+          "label": "Automatically focus title field",
```
mathieudutour, line 38: could it be a dropdown to pick the field to focus instead?

thomaslombart: It could, but it feels a bit overkill to implement it since there are no user needs for that right now. What do you think?

mathieudutour: Even if there are only 2 options for now, I'd feel that it would be both clearer and more extensible (keep in mind that if we want more options later, a checkbox won't be enough and we will lose the preference

thomaslombart: Are you talking about a dropdown for all fields or only for the team and title fields?

If that's for all fields, I feel like picking the field to focus from a dropdown doesn't seem right as I don't see why a user would want to autofocus a field past the title. I get the need for the fields before the title but not after.

The implementation looks also a bit harder to implement since the `autofocus` prop would have to be conditionally applied to every field. Also, some fields are conditionally hidden if there are no data (e.g cycles), so a user could very well select the "Cycles" field in the preference even if their team has no cycles. That would mean that these edge cases would need to be handled as well.

mathieudutour: A dropdown to choose just between the team and the title. It would make it clearer what you are changing it from (which isn't clear with the checkbox)

mathieudutour (approved):

PR #3251, opened 2022-10-16
add random-identity-generator

nghlt:
## Description
This extension generates a random username, name, address using API from https://rapidapi.com/mrsonj/api/random-username-generate. It also can generate a random password as well as an @duck.com email.

## Screencast

https://user-images.githubusercontent.com/19650585/196023953-c87f23a8-08e3-4261-8a5a-a345c073a0f9.mov

## Checklist

### DIFF
--- a/extensions/random-identity/README.md
+++ b/extensions/random-identity/README.md
@@ -4,7 +4,18 @@ Generate a new random identity
 
 ## Duckduckgo email generator and email converter
 
-Use your own duckduckgo email and token.
+Use your own duckduckgo email and token. Get your token using this instruction below:
+- Install Duckduckgo extension for [Google Chrome](https://chrome.google.com/webstore/detail/duckduckgo-privacy-essent/bkdgflcldnnnapblkhphbgpggdiikppg), [Microsoft Edge Chromium](https://microsoftedge.microsoft.com/addons/detail/duckduckgo-privacy-essent/caoacbimdbbljakfhgikoodekdnlcgpk?city=Lynchburg) or other browsers that supported that extension.
+- Install a HTTP debug tool like Proxyman, Charles Proxy. Turn on SSL Proxying on `quack.duckduckgo.com`. Make sure to install and trust Root CA from HTTP debug tool to perform the man-in-the-middle (MITM) attack.
+- Access `https://duckduckgo.com/email/` to sign in your @duck.com email or sign up for a new account.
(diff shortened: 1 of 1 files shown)


pernielsentikaer (changes requested):
Hi @trungnghiatn 👋 

I have a few questions about this extension:

* How do I get a `duckToken`, it could be nice with some more explanation in the `README`
* Most of the other commands are already something that other extensions have, like:
 * [Random](https://www.raycast.com/loris/random), give you everything for random data
 * There are at least three extensions [generating passwords](https://www.raycast.com/search?q=password)
 
 Maybe the extension should be all about the duckduckgo integration instead?

trungnghiatn:
Hi @pernielsentikaer 
I updated an instruction to get duck_token in `README`.

> * Most of the other commands are already something that other extensions have, like:
> * [Random](https://www.raycast.com/loris/random), give you everything for random data
> * There are at least three extensions [generating passwords](https://www.raycast.com/search?q=password)

I know there are similar extensions but each has its own strength. My purpose is making multiple commands in an extension. Each command is made as simple as possible. I think they are the differences compared to other extensions.

pernielsentikaer:
I still don't think it's a good idea to mix multiple services into one extension; as previously mentioned, both password and random data are already covered - maybe it's a good idea only to improve a duck extension?

trungnghiatn:
Hmm. Instead of going around and finding a few extensions for generate random identities, I prefer providing them all in one. I think I will stop here unless you change your mind.

pernielsentikaer:
Thanks for getting back here 🔥 

Within [Random Data Generator](https://www.raycast.com/loris/random) you can already do:
* Random Password
* Get Address
* Get Username
* Get Name

You could probably continue on the extension about getting data from `RapidAPI`, but then I would leave the `duck` part out and only focus on `RapidAPI` part 😊

Let me know your thoughts

trungnghiatn:
I completely disagree.
1. `RapidAPI` like a store for APIs. Each API has the different response. Making an extension to get data from `RapidAPI` is so specific for individual APIs. So, the loop of making another extension still happens.
2. Given you browse on the internet to looking for an information. You access a forum that requires registration to keep reading. If you sign up, you have to fill several fields:
- username
- name
- email
- password
Other registration sometimes ask for your address. Because it's just a random visit, you don't have to use your real identity. That why I want my extension can includes ability to generate all these information. I don't think install many extension for a specific purpose is a good idea.

PR #3279, opened 2022-10-19
Adds SSH command to GitHub-Codespaces extension

mkwng:
## [Adds SSH command] - 2022-10-19

- Adds SSH option as one of the preferred clients
- Reducing polling period for menubar application


maxbeizer:
Yes please!!!

💖 a codespace SSH user

PR #3333, opened 2022-10-27
Color Casket: Make Picker command a no-view command

fe9lix:
## Description

@slavarazum 
Fixes the Color Casket Picker command to be a no-view command. 
Currently the command is set up with an async main function, which the API only supports for no-view commands.
(This will generate a warning in future API releases...)

## Checklist

### DIFF
--- a/extensions/color-casket/src/components/ColorPickers.tsx
+++ b/extensions/color-casket/src/components/ColorPickers.tsx
@@ -72,7 +72,9 @@ export default function ColorPickers() {
           <Action
             title={`Pick in ${defaultFormat}`}
             icon={Icon.Brush}
-            onAction={() => launchCommand({ name: "picker", type: LaunchType.UserInitiated })}
+            onAction={() =>
+              launchCommand({ name: "picker", type: LaunchType.UserInitiated, context: { open: "fromRender" } })
+            }
           />
           <ActionPanel.Section key="pickers">
             {orderedPickFormats.map((type, index) => (
--- a/extensions/color-casket/src/picker.tsx
+++ b/extensions/color-casket/src/picker.tsx
@@ -1,4 +1,13 @@
-import { Clipboard, closeMainWindow, getPreferenceValues, popToRoot, showHUD } from "@raycast/api";
+import {
+  Clipboard,
+  closeMainWindow,
+  environment,
+  getPreferenceValues,
+  launchCommand,
+  LaunchType,
+  popToRoot,
+  showHUD,
+} from "@raycast/api";
 
 import { ColorType } from "./colors/Color";
 import { prepend } from "./hooks/colorSaver";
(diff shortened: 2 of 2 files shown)


slavarazum:
Great improvement @fe9lix! Previously `no-view` mode did not work with async operations. Looks like now everything is fine.

We still have a problem with the picker call from `index` command. It has 10s timeout after which the color is not saved and returning to the Raycast window does not work.

I have tried to change [this line](https://github.com/raycast/extensions/blob/main/extensions/color-casket/src/components/ColorPickers.tsx#L73) to:

```jsx
onAction={async () => await openColorPicker(defaultFormat)}
```
timeout still exists.

Do we have any ability to create async actions without timeout like with `no-view` mode?

pernielsentikaer:
Hi @slavarazum 👋 

Can you check it again, I just want to make sure that I didn't break anything 😊

slavarazum:
Hi @pernielsentikaer 🙌 
Checked latest change and the behavior has changed from what was originally planned.

Separate **Pick Color** command just copies a result to clipboard.
When you run **Pick a Color** action from **Render Color** command it copies a result to clipboard and returns to the command view to show it in the history list.

As I can see the new `launchCommand` function runs separate **Pick Color** command without ability of further control from the **Render Color** command.
It would be nice to have the ability to run async operation from list item action with ability to return to previous state with updates.

sxn:
hi @slavarazum – I've reverted the previous change, and updated it s.t. the action handler is async.
It seems to work for me as expected w/ this change: with `pop to root` set to immediate, I can pick a colour and it gets added to the clipboard even if I take ~20s to decide what colour I want 🤔 


https://user-images.githubusercontent.com/1155589/198678945-f97dfb7b-696b-4685-8054-ab73f7233ef5.mov

fe9lix:
> It would be nice to have the ability to run async operation from list item action with ability to return to previous state with updates.

@slavarazum If I understand it correctly, you could ahieve that with the new command launching API by passing a context object between the two commands.

slavarazum:
@fe9lix I want to achieve opening color picker from 2 different commands with 2 different behavior. Running one command from another is not suitable to provide different results.

You can try to check the behavior of the Color Casket actual version by running a separate **Pick Color** command and with the **Pick a Color** action from **Render Color** command.

## Expected behavior

https://user-images.githubusercontent.com/5820718/198695767-7faba0d7-3ae3-4473-a9b6-5ffd8ab699f9.mp4

## 10s Timeout bug

https://user-images.githubusercontent.com/5820718/198695924-86525a3e-026a-4339-8cc3-0fe62145977c.mp4

pernielsentikaer:
Hi @slavarazum 👋 

Could you give it another spin 😊

slavarazum:
@pernielsentikaer Made some additions in #3360 
New issue with "root window flickering" appeared. Details in description.

pernielsentikaer:
So let's merge this one, right @slavarazum (and continue in the other one)

slavarazum:
@pernielsentikaer This one only implement picking in default format with "root window flickering" issue and doesn't cover [picking in other formats](https://github.com/raycast/extensions/blob/e9a20ecbdc59a2d7b2cae6c2b704a7f59e44b1f9/extensions/color-casket/src/components/ColorPickers.tsx#L86).

In https://github.com/raycast/extensions/pull/3360 behavior with other formats is already implemented. So, let's merge #3360 to ensure same behavior for picking colors in default and other formats.

As a next step we should wait some additions for `launchCommand` function to fix "root window flickering" for `no-view` commands.

PR #3371, opened 2022-11-01
Add Solana Explorer extension

AnishDe12020:
## Description

New Extension: Solana Explorer 

## Screencast

https://user-images.githubusercontent.com/63192115/199194044-1f684fa0-e21b-4465-8c31-59a05e196f0c.mov

## Checklist

### DIFF
--- a/extensions/solana-explorer-raycast/package.json
+++ b/extensions/solana-explorer-raycast/package.json
@@ -4,7 +4,7 @@
   "title": "Solana Explorer",
   "description": "View Solana Account and Transaction details right from Raycast",
   "icon": "logo.png",
-  "author": "anishde12020",
+  "author": "AnishDe12020",
   "categories": [
     "Developer Tools",
     "Productivity"
@@ -14,9 +14,6 @@
     "explorer",
     "solscan",
     "web3"
-  ],
-  "contributors": [
-    "AnishDe12020"
   ],
   "license": "MIT",
   "commands": [


pernielsentikaer:
Hi @AnishDe12020 👋 

That looks like a solid update to the current extension build by @TrGiLong - could be that this supersedes the other one, do you have any thoughts @TrGiLong

AnishDe12020:
Any update on what is the next step? I am fine with any - it being an update to the existing extension or it being published as a new extension

AnishDe12020:
@pernielsentikaer could you please let me know the next step?

AnishDe12020:
@pernielsentikaer It has been ~6 weeks, could you please let me know what can be done about this pr?

pernielsentikaer:
The feature the other one has, is that relevant? I can't figure out how to do that in yours. Also, all commands don't have a loading indicator when searching; could you look into that?

AnishDe12020:
> The feature the other one has, is that relevant?

Didn't understand this part

> Also, all commands don't have a loading indicator when searching; could you look into that?
Lemme look into this and push a fix in some time

pernielsentikaer:
I'm not really into Solana, but I tried in your extension to be able to search for something like the other one can
Raycast-TGz4ZT2T@2x

AnishDe12020:
Ah I see, seems like it got token search too! I also looked at the code and there is a block and epoch search. I can implement these in my extension in a similar fashion, should I do that?

pernielsentikaer:
It can also be me that's doing in doing it wrong.

https://user-images.githubusercontent.com/38033107/207606651-38108c07-0bcd-4332-8c7c-0e55fdf0fb67.mp4

I think we should aim for a solution where we still only have one version, what do you think about enhancing the current with all yours when it's working as expected?

AnishDe12020:
Oh, so the extension currently only does a search if the character length of the query is 44 or 88 (account address or tx signature) and hence there is no loading state

Regarding contributing to the older extension, it seems like it will be a huge refactor as the structure is quite different. It will be easier to replace it with my code and add the extra features to be honest but lmk what you think.

I am fine with doing any

pernielsentikaer:
Could you also check #3820 - There might be some good stuff in that one - can we all agree on this version being the "one" that will override the current (when the other feature is added)

AnishDe12020:
> Could you also check #3820 - There might be some good stuff in that one - can we all agree on this version being the "one" that will override the current (when the other feature is added)

This one is pretty similar to the current one but it has a better way to handle explorers and a history feature.

I can mix the best features of all 3 (that does mean using code from all 3) and make one extension that replaces the current one. Should I continue in this pr or make a new pr?

pernielsentikaer:
What do you say @aquare11e

pernielsentikaer:
Do you have any last input @aquare11e

pernielsentikaer:
Let us enhance the current with the things you mentioned if you're still up for that?

If we do it correct it's just going to be an update of the current 😊

AnishDe12020:
> Let us enhance the current with the things you mentioned if you're still up for that?
> 
> If we do it correct it's just going to be an update of the current 😊


Sure, I will create a new pr for this

btw, im a lil busy with other stuff but i will work on this, give me the time of a week or 2

pernielsentikaer:
No rush, and if you just use the same folder/package.json from the original, it will just be an update 💝

PR #3391, opened 2022-11-02
bitwarden: Copy passwords with concealed type

eth-p:
## Description

When copying passwords from the BitWarden vault, mark the clipboard contents as concealed/sensitive. As Raycast does not provide a native way to do this yet, this change leverages MacOS 10.10 and newer's ability use call Objective-C classes via AppleScript.

More details on how this should work:
http://nspasteboard.org/

## Checklist

### DIFF
--- a/extensions/bitwarden/src/clipboard.ts
+++ b/extensions/bitwarden/src/clipboard.ts
@@ -1,0 +1,43 @@
+import { runApplescript } from "./utils";
+import { Clipboard } from "@raycast/api";
+
+/**
+ * Copies a password to the system clipboard, setting the appropriate options to conceal it from well-behaved
+ * clipboard history software.
+ *
+ * If this cannot be done (e.g. macOS version < 10.10), it will be copied using the standard clipboard copy mechanism.
+ *
+ * @param password The password to copy.
+ * @return true If copied securely.
+ */
+export async function copyPassword(password: string): Promise<boolean> {
+  const result = runApplescript(
+    `
+use AppleScript version "2.4"
+use framework "Foundation"
+use framework "AppKit"
+use scripting additions
+
+property NSPasteboardTypeString : a reference to current application's NSPasteboardTypeString
+
+on run argv
+  set textToCopy to item 1 of argv
+
+  set cb to current application's NSPasteboard's generalPasteboard() -- get pasteboard
+  cb's clearContents()
+  cb's setString:textToCopy forType:"org.nspasteboard.ConcealedType" -- http://nspasteboard.org/
+  cb's setString:textToCopy forType:"com.agilebits.onepassword" -- 1Password
+  cb's setString:textToCopy forType:NSPasteboardTypeString
+end run
+`,
+    [password]
+  );
+
+  try {
+    await result;
+    return true;
+  } catch (ex) {
+    await Clipboard.copy(password);
+    return false;
+  }
+}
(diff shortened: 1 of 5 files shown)


extensions/bitwarden/src/utils.ts
```diff
@@ -65,3 +66,29 @@ export function extractKeywords(item: Item): string[] {
...
+        if (err != null) {
+          reject(err);
+          return;
+        }
+
+        resolve(stdout);
+      }
+    );
+  });
+}
```
jomifepe, line 94: Small suggestion here, we could make use of the `execa` library we have as a dependency and use the `async await` syntax.

Don't forget to delete the import statement.

```suggestion
export async function runAppleScript(appleScript: string, args: string[]) {
  const { stdout } = await execa("/usr/bin/osascript", ["-e", appleScript, ...args], { shell: false });
  return stdout;
}
```

extensions/bitwarden/src/components.tsx
```diff
@@ -76,3 +77,12 @@ export function UnlockForm(props: { onUnlock: (token: string) => void; bitwarden
     </Form>
   );
 }
+
+export function CopyPasswordToClipboardAction(props: { title: string; content: string }): JSX.Element {
+  async function doCopy() {
+    const copiedSecurely = await copyPassword(props.content);
+    showHUD(copiedSecurely ? "Copied password to clipboard" : "Copied to clipboard");
```
jomifepe, line 84: For better dev experience, we shouldn't rely on a boolean to check if the password was copied securely. Instead we could return a something more explicit, like `{ copiedSecurely: true }` on the `copyPassword` function.

```suggestion
    const { copiedSecurely } = await copyPassword(props.content);
    showHUD(copiedSecurely ? "Copied password to clipboard" : "Copied to clipboard");
```

extensions/bitwarden/src/clipboard.ts
```diff
@@ -0,0 +1,43 @@
...
+ * clipboard history software.
+ *
+ * If this cannot be done (e.g. macOS version < 10.10), it will be copied using the standard clipboard copy mechanism.
+ *
+ * @param password The password to copy.
+ * @return true If copied securely.
+ */
+export async function copyPassword(password: string): Promise<boolean> {
+  const result = runApplescript(
+    `
```
jomifepe, line 15: The script string can be moved to a constant outside the function

jomifepe (approved):
Seems like a great improvement. I believe Raycast's clipboard manager already looks into these properties to hide passwords.

I suggested some small code improvements, but overall everything seems fine, including the AppleScript.

Great job 👍

eth-p:
Implemented your suggestions, thanks!

I'm not too sure about the clipboard manager in Raycast itself, but the one I use (Paste) works flawlessly with this change :)

pernielsentikaer:
@pomdtr can you check this 😊

pomdtr:
it looks good to me, thanks @eth-p for implementing the feature

pernielsentikaer (approved):

pomdtr:
I forgot to say that I agreed with all of @jomifepe suggestions, I'll implement them myself

pernielsentikaer:
Just ping me when you have something ready @pomdtr 😊

eth-p:
Seems I forgot to force-push those changes that @jomifepe suggested :(

I have some other features in the works, but after those are merged I'll do a cleanup PR including the suggestions?

PR #3413, opened 2022-11-04
bitwarden: Master Password Reprompt + Cleanup + Refactoring

eth-p:
## Description

These four commits implement the "Master password re-prompt" option for Raycast's Bitwarden extension.

The first commit is a refactoring of how Bitwarden vault sessions are handled, and aims to decouple the session management from the Search interface while also providing a more React-like usage. This was needed to prevent prop-drilling and make it easier to implement the master password re-prompt feature.  

The second commit cleans up the clipboard password copy code from my last pull request. I forgot to push it 😅

The third and forth commits implement the re-prompt feature and a setting for temporarily disabling re-prompt dialogs for a specified period of time after entering the master password.

*When reviewing, I would recommend looking at it commit-by-commit.*
It was structured so each commit is a single feature :)

## Checklist

### DIFF
--- a/extensions/bitwarden/CHANGELOG.md
+++ b/extensions/bitwarden/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Bitwarden Changelog
 
+## [Enhancements] - 2023-01-17
+
+- Added master password re-prompt feature.
+- Cleans up the clipboard password copy code
+- Added re-prompt feature and a setting for temporarily disabling re-prompt dialogs for a specified period of time after entering the master password.
+
 ## [Copy Enhancement] - 2022-11-02
 
 - Copy passwords in a way that asks clipboard managers to not save it 


extensions/bitwarden/src/session.tsx
```diff
@@ -0,0 +1,307 @@
...
+
+  // Functions.
+  /**
+   * Refresh the state of the session provider.
+   * This updates the component state.
+   *
+   * @param newState The state to update.
+   */
+  async function update(newState: Partial<Session>): Promise<void> {
+    const status = await api.status(newState.token);
```
pomdtr, line 116: The `bw status` command is buggy in the cli, we cannot rely on it. The best approach I found out was to run the command (ex `bw list items`), and to show the form if an error occurs.

See https://github.com/bitwarden/clients/issues/2729 for more infos.

pomdtr (changes requested):
Thanks for for your work, all of these changes are welcome !

I only highlighted a single issue, please take a look at it.

extensions/bitwarden/src/session.tsx
```diff
@@ -0,0 +1,307 @@
...
+      const providedHashed = await hashMasterPasswordForReprompting(password);
+      if (providedHashed !== this.repromptHash) {
+        return false;
+      }
+
+      const now = new Date();
+      await LocalStorage.setItem(REPROMPT_PASSWORD_ENTERED_KEY, new Date().toString());
+      setSession((old) => ({ ...old, passwordEnteredDate: now }));
+      return true;
+    },
```
jomifepe, line 105: I'm a bit confused with this approach. All these functions are being passed as the `useState`'s initial state, which means that they are not reactive to the component's scope, and also, they don't need to be changed programmatically, so no need for them to be here. I'm not sure if there's anything inside the functions that can change at runtime, but nonetheless, these should be moved outside the useState initializer, and passed to the provider below, alongside the rest of the values (e.g. token, isLoading, etc). Same for the `api` variable, which is already passed as prop.

jomifepe (commented):
Great idea on the reprompt feature 🚀  

Although, I pointed out some concerns about the code.

extensions/bitwarden/src/session.tsx
eth-p: Thanks for pointing that out! I'll change it and add a comment to capture the context.

extensions/bitwarden/src/actions.tsx
```diff
@@ -0,0 +1,101 @@
...
+  title?: string;
+  item?: Item;
+}): JSX.Element {
+  async function doCopy() {
+    const { copiedSecurely } = await copyPassword(props.content);
+    await showHUD(copiedSecurely ? "Copied password to clipboard" : "Copied to clipboard");
+  }
+
+  const action = props.reprompt
+    ? reprompt(props.session, doCopy, { item: props.item, what: "copy the password" })
```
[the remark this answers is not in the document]
eth-p: Good catch on that! It was supposed to be more of a wrapper function than a hook, but the `useNavigation` within the `reprompt` function does make it a bit problematic. I'll take your suggestion on this and turn it into a hook instead.

extensions/bitwarden/src/session.tsx
eth-p: I wanted to make it a bit less verbose by just keeping everything within a single object and updating that, rather than having both a `Session` object for public API consumption and a separate `SessionState` object for holding the state data itself.

In hindsight, it was a pretty bad approach that made far less sense for the sake of brevity. I split them back up into a `Session` class and `SessionState` interface in my latest round of changes, and wired it up appropriately so the `Session` exposes only the necessary parts of the `SessionState`.

extensions/bitwarden/src/actions.tsx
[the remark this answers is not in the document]
eth-p: Changed it to be a hook instead, thanks for the suggestion!

extensions/bitwarden/src/session.tsx
eth-p: Context is now captured:

https://github.com/raycast/extensions/pull/3413/files#diff-59101489bd951ec97dd45c500bb43bc2775e174d64b127b034e16df3f41ec595R121-R145

pernielsentikaer:
I lost track of the status. Does anyone need to recheck it? Please ping if anyone needs to do anything 😊

pomdtr:
Please do not merge it yet, I want to take another look at it :)

pernielsentikaer:
Sure, I'll await you 😊

extensions/bitwarden/src/actions.tsx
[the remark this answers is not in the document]
jomifepe: Looks good 👍

extensions/bitwarden/src/session.tsx
jomifepe: Cool, nice work 👍 

I don't really like the class approach, this is React after all, and that could easily be handled by using hooks and some functions, but I'll leave that to be evaluated by @pomdtr.

pomdtr:
I'll revesit this PR tomorrow, sorry for the delay @eth-p

pernielsentikaer:
@pomdtr did you have time to revisit this yet?

pernielsentikaer:
Let's check with @pomdtr one last time since he said he wanted to test it out 😊

pomdtr:
Hey, I finally took the time time to test it (sorry for the delay, I will probably transfer the ownership of the extension soon).

It looks good to me ! However I noticed a bug after messing up the initial setup. Could you please fix the label of the toast ?

image

jomifepe:
> Hey, I finally took the time time to test it (sorry for the delay, I will probably transfer the ownership of the extension soon, @jomifepe would you be interested ?).
> 
> It looks good to me ! However I noticed a bug after messing up the initial setup. Could you please fix the label of the toast ?
> 
> image

Hey, sorry for the delay. Do you mean transfer ownership for GitHub issues and PR's, not on the Raycast store right?

pomdtr:
No problem. If someone steps up for handling issues and prs, he should probably steps up as the author of the extension. I'm fine beeing relegated to a contributor :)

pernielsentikaer:
Should we agree on transferring the extension to @jomifepe and set @pomdtr as a contributor?

pomdtr:
If @jomifepe agrees to takes ownership, go ahead. If not, has we've discussed on slack, @pernielsentikaer could also be the maintainer.

pernielsentikaer:
Voting for @jomifepe too!

jomifepe:
@pernielsentikaer @pomdtr Yeah I can take the ownership 🙂

pernielsentikaer (approved):
Let's make the change 🫣

PR #3421, opened 2022-11-06
[AWS] Improve data fetching

JonathanWbn:
## Description

- use Raycast's `useCachedPromise` for all data fetching
- improve readability by extracting data fetching into separate functions
- parallelize SQS queue status updates to speed up loading times

### DIFF
--- a/extensions/amazon-aws/CHANGELOG.md
+++ b/extensions/amazon-aws/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Amazon AWS Changelog
 
-## [Improve Data Fetching] - 2022-11-07
+## [Improve Data Fetching] - 2022-11-14
 
 Use Raycast's API for all data fetching ([#3421](https://github.com/raycast/extensions/pull/3421))
 


pernielsentikaer:
Hi @JonathanWbn 👋 

Thanks for doing this update 🔥

@Falcon can you check this if you have time 😊

extensions/amazon-aws/src/queues.tsx
```diff
@@ -1,11 +1,11 @@
 import { getPreferenceValues, ActionPanel, List, Detail, Action, confirmAlert, Toast, showToast } from "@raycast/api";
 import { useCachedPromise } from "@raycast/utils";
+import { chunk } from "lodash";
```
gebeto, line 3: let's use direct import instead to prevent loading full lodash module

```suggestion
import chunk from "lodash/chunk";
```

extensions/amazon-aws/src/cloudformation.tsx
```diff
@@ -1,61 +1,33 @@
...
 import setupAws from "./util/setupAws";
+import { useCachedPromise } from "@raycast/utils";
 
 setupAws();
+const cloudformation = new AWS.CloudFormation({ apiVersion: "2016-11-15" });
 
 export default function ListStacks() {
-  const preferences: Preferences = getPreferenceValues();
-  const cloudformation = new AWS.CloudFormation({ apiVersion: "2016-11-15" });
+  const { data: stacks, error, isLoading } = useCachedPromise(fetchStacks);
```
gebeto, line 12: I'm not sure that useCachedPromise is okay here, because if it is storing result between command runs, it means that if `stacks` will be changed in AWS, it will be not refetched because it will get data from catch. So maybe it makes sense to use `usePromise` instead

Screenshot 2022-11-08 at 15 17 11


Did you test that case?

gebeto (changes requested):

extensions/amazon-aws/src/cloudformation.tsx
gebeto: or maybe I just do not understand how it works

if it is refetching on next command run but using cached value as initial value, then it is okay to use this one

extensions/amazon-aws/src/queues.tsx
JonathanWbn: yup, good point, done.

extensions/amazon-aws/src/cloudformation.tsx
JonathanWbn: > if it is refetching on next command run but using cached value as initial value, then it is okay to use useCachedPromise

that was my assumption. I've confirmed that and it **does** work like that. so using `useCachedPromise` is the best choice here IMO.

gebeto: Yep, I agree, cool 🚀

gebeto (approved):
LGTM 👍

JonathanWbn:
@pernielsentikaer Can we get this one merged? I'd love to keep building on it.

pernielsentikaer (approved):

Issue #3465, opened 2022-11-10
onInput callback for Form.TagPicker or Form.Dropdown


thomaslombart:
**Describe the feature and the current behavior/state.**

It'd be nice that `Form.TagPicker` or `Form.Dropdown` includes an `onInput` callback so that we can execute an action each time the user enters text in these fields. The main use case would be for searching in a list that might return a lot of options. 

For example, in the Linear extension, the API returns the 50 first results by default for the queries. This causes a problem for Linear workspaces with more than 50 labels since all labels wouldn't be shown. While it's possible to increase this number, this wouldn't solve the issue since it's not possible to know in advance what's a reasonable number of labels. Instead, it's better to allow users to search for labels directly in the picker or dropdown. That way, they can search for the label they want. (cf [Slack thread](https://raycastcommunity.slack.com/archives/C028DTXTW3Y/p1668003021307629))

**Will this change the current API? How?**

This would be purely additive to the API. No breaking changes to expect.

**Who will benefit from this feature?**

Any users who want to execute some kind of action when the users enters text in `Form.TagPicker` or `Form.Dropdown`

mathieudutour:
In the next release, we are going to add `onSearchTextChange` to `Form.Dropdown`. Going to think about the TagPicker

peduarte:
As Mathieu said, `onSearchTextChange` is now available in `Form.Dropdown` https://developers.raycast.com/api-reference/user-interface/form#props-6

This isn't the case for `Form.TagPicker` though. @mathieudutour Is this under consideration?

vogelino:
I'd love to see that for `Form.TagPicker` as well.

Consider this use case: You want to present the user a list of tags to choose from but want to allow them to also add tags that aren't part of the list yet. In oder to react to that, you would either listen for onSearchTextChangeEvent and add the tag when the tab key is triggered or when a coma character has been inserted.

If there is already a way to do that, please let me know. Cheers!

Here's how Notion does it:
![Screenshot 2024-06-10 at 10 26 05](https://github.com/raycast/extensions/assets/2759340/60b75d0f-dd8a-4080-9561-a5bf0e4bece5)

bkeys818:
@thomaslombart @peduarte @mathieudutour

Any update on adding `onSearchTextChange` and `throttle` to `TagPicker`?

thomaslombart:
This component is tricky, making it more difficult to add these props to it. We're thinking of alternative ways to solve the underlying use cases 🙂

PR #3468, opened 2022-11-10
fix(linear): allow more labels

thomaslombart:
## Description

Linear's API returns 50 results by default. This causes a problem when a workspace has more than 50 labels. This PR allows up to 100 labels.

Note this is a temporary fix. A better fix would be to have some sort of typeahead search on the tag picker (cf #3465)

## Checklist


thomaspaulmann:
Thanks for "fixing" this. We have plans to support typeahead searches in the action panel, e.g. add/remove labels. But good point that it would also be useful for the tag picker.

peduarte:
Good stuff!

peduarte (approved):

PR #3470, opened 2022-11-10
feat: add "Paste Without Newlines" command to "Paste as Plain Text" extension

pantajoe:
## Description

This enhances the extension "Paste as Plain Text" with some preferences:

- "Clean Line Breaks" allows you to paste the clipboard text with now newlines (they are replaced as spaces).
- "Trim" allows you to paste the clipboard text without leading/trailing spaces

## Screencast

image

## Checklist

### DIFF
--- a/extensions/paste-as-plain-text/CHANGELOG.md
+++ b/extensions/paste-as-plain-text/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Paste as Plain text Changelog
 
+## [Enhancements] - 2022-11-16
+
+- Clean Line Breaks allows you to paste the clipboard text with now newlines (they are replaced as spaces).
+- Trim allows you to paste the clipboard text without leading/trailing spaces
+
 ## [Refactor code] - 2022-06-04
 - Refactor code
 


koinzhang:
Hi @pantajoe Did you notice this extension [**Paste as Plain Text**](https://www.raycast.com/koinzhang/paste-as-plain-text) in the store?

pantajoe:
Hi! Yes, I did, but I needed to replace newline characters with spaces as well, which the extension does not do 😅

pernielsentikaer:
I think it would be better to contribute to the other extension instead and add a preferences value, end users might be confused if multiple extensions have almost the same name (and does almost the same)

koinzhang:
Hi @pantajoe Perhaps it would be more appropriate to make this feature a Preferences item? Instead of adding another command, like adding the option to replace line breaks with spaces in the Preferences

pantajoe:
I've thought about it, but then you cannot use both commands simultaneously, right?
Sometimes I want to keep the newlines, and sometimes I do not. 😅

pantajoe:
On second thought, this might be an edge case as most applications have the shortcut CMD+Shift+V to "Paste and Match Style" already. I'll change it, so you can have a checkbox in the preferences panel of the extension. Just a sec! 😄

koinzhang:
But I don't think this is a common feature, and adding a very similar command for such a specific need would confuse the user.
Also, if you simply want to replace line breaks with spaces, you can try using the T[ext Shortcuts](https://www.raycast.com/koinzhang/text-shortcuts) command, I see you use `/[\r\n]/gm` in your code, you can create a _Replace_ action in Text Shortcuts (which supports regular expressions) to replace the line breaks
image

koinzhang:
Just create one of these text shortcuts and you can easily replace all line breaks with spaces😉
image

pantajoe:
Ah, okay. Thanks! Haven't seen this extension before. But I can't define a custom shortcut for one specific action, can I?

koinzhang:
You can customize the text shortcut and then quickly find and use it through the Shortcut Library command.
image
The Shortcut Library command also has some built-in shortcuts that you can use to quickly process text
image

pantajoe:
I see. However, I feel, that this is kind of disrupting as I have to perform this transformation very frequently. Maybe we can add this as a preference option like discussed before?

koinzhang:
You are very welcome to add this feature as a preference, and perhaps provide more preferences, such as removing all spaces and deleting spaces at both ends of the text, but that is all an afterthought

pantajoe:
That's a nice idea. Will do this right away

pantajoe:
Okay, I've added two preference options as discussed and I've tested the combinations 👍🏽

koinzhang:
Hi @pernielsentikaer. I reviewed the code and installed PR for testing and all functions worked fine.

PR #3474, opened 2022-11-10
Soundboard extension

pernielsentikaer:
## Description

Use Raycast as a soundboard and bind sounds to hotkeys

## Screencast

soundboard-1
soundboard-2
soundboard-3

## Checklist

### DIFF
--- a/extensions/soundboard/src/utils.ts
+++ b/extensions/soundboard/src/utils.ts
@@ -1,7 +1,7 @@
 import { exec } from "child_process";
 import { Item } from "./types";
 import { getItem, getItems, saveItems } from "./storage";
-import { showHUD } from "@raycast/api";
+import { launchCommand, LaunchType } from "@raycast/api";
 
 export const playSoundFromIndex = async (index: number) => {
   const sound = await getItem(index);
@@ -10,7 +10,7 @@ export const playSoundFromIndex = async (index: number) => {
     return;
   }
 
-  await showHUD(`❌ Favourite #${index} is not defined`);
+  launchCommand({ name: "index", type: LaunchType.UserInitiated, context: { index: index } });
 };
 
 export const playFile = async (item: Item) => {
--- a/extensions/soundboard/src/index.tsx
+++ b/extensions/soundboard/src/index.tsx
@@ -1,4 +1,15 @@
-import { ActionPanel, Action, List, Icon, confirmAlert, Color, useNavigation } from "@raycast/api";
+import {
+  ActionPanel,
+  Action,
+  List,
+  Icon,
+  confirmAlert,
+  Color,
+  useNavigation,
+  showToast,
+  Toast,
+  environment,
+} from "@raycast/api";
 import { getItems } from "./storage";
 import { Item } from "./types";
 import { useState, useEffect } from "react";
@@ -12,7 +23,13 @@ export default function Command() {
 
   useEffect(() => {
     (async () => {
-      setConnectionsList(await getItems());
+      const items = await getItems();
+      setConnectionsList(items);
+
+      if (environment.launchContext && items.length === 0) {
+        showToast(Toast.Style.Failure, "No sounds found", "Create a sound to get started");
+      }
+
       setLoading(false);
     })();
   }, []);


peduarte:
Love this idea @pernielsentikaer ❤️ 

I tested this locally and have some feedback.

When I first ran the extension, all I saw was a list of "Play Favourite Commands:

CleanShot 2022-11-14 at 17 04 20@2x

I was trying to play them, but getting an error. 

Took me a while to realise there was a "Sounds" Command, which is needed in order to actually add sounds to each of the Favourite commands. 

I found it quite overwhelming to see so many commands 😳

Not sure what to do about this, because we cannot programatically enable commands.

One thing that comes to mind, is consolidate them all into a single "view" command. Such as "View Soundboards". Render each favourite as a `List.Item`. And this way we could render as many as have been defined. 

The downside is that if users want to asign a hotkey, they can't.

---

I also think we should stick with just one keyboard shortcut to add a new sound. In the following page, users are instructed to add a new sound via ⌘+N as well as ⏎

CleanShot 2022-11-14 at 17 07 12@2x

peduarte:
@felixb1010 had an idea which could improve the experience a bit. Which is to instead of displaying a HUD, redirect users to the sound form:
```diff
export const playSoundFromIndex = async (index: number) => {
  const sound = await getItem(index);
  if (sound) {
    await playFile(sound);
    return;
  }

- await showHUD(`❌ Favourite #${index} is not defined`);});
+ launchCommand({ name: "index", type: LaunchType.UserInitiated 
};
```

pernielsentikaer:
Thanks for the review 😊

> One thing that comes to mind, is consolidate them all into a single "view" command. Such as "View Soundboards". Render each favourite as a List.Item. And this way we could render as many as have been defined.

Yes, but as you're saying then it's not possible to bind a hotkey, I agree it would be cool to have a way to have commands disabled from default 😉

> I also think we should stick with just one keyboard shortcut to add a new sound. In the following page, users are instructed to add a new sound via ⌘+N as well as ⏎

As soon as at least one sound is on the list, then ⏎ is mapped to `Play`, so that's why it might be smarter to learn the right shortcut (That people already have in mind, the same for ⌘+E for `Edit`

CleanShot 2022-11-14 at 20 19 50@2x

> Took me a while to realise there was a "Sounds" Command, which is needed in order to actually add sounds to each of the Favourite commands.

> @felixb1010 had an idea which could improve the experience a bit. Which is to instead of displaying a HUD, redirect users to the sound form

It might be a tough one to tackle, since you assign a favorite to a sound, if no sound is added it might confuse since we don't know if the user want's to assign or create a new sound, I'll try to think of a solution for this 😊

peduarte:
> It might be a tough one to tackle, since you assign a favorite to a sound, if no sound is added it might confuse since we don't know if the user want's to assign or create a new sound, I'll try to think of a solution for this 😊

Im not sure I follow, but I trust ya 😄

pernielsentikaer:
Hi @peduarte 

I made a few changes, let me know how it works out for you now 😊

extensions/soundboard/src/utils.ts
```diff
@@ -10,7 +10,7 @@ export const playSoundFromIndex = async (index: number) => {
     return;
   }
 
-  await showHUD(`❌ Favourite #${index} is not defined`);
+  launchCommand({ name: "index", type: LaunchType.UserInitiated, context: { index: index } });
```
peduarte, line 13: What do we use the `index` context for?

pernielsentikaer: Nothing (at the moment) but if I don't pass anything as context it's triggers the toast since this extension always is UserInitiated, but it could be anything actually

peduarte: Oh, I see! This is a bit of hack to know whether the view was opened from a "user" or from a "launchCommand"

/cc @felixb1010 would be nice if there was a more intuitive way to do this.

peduarte (approved):
@pernielsentikaer Very nice Per 🚀

felixb1010:
I am not sure what you guys are talking about?
Which idea? from where?

peduarte:
@felixb1010 Wrong Felix, sorry!

pernielsentikaer:
Actually, @peduarte, I can just use `Launchtype.Background` instead of the context, the plan was that I'm going to incorporate the index in the toast somehow so it's not changing that much

peduarte:
@pernielsentikaer I tried it locally, `Background` doesnt work in this case.

pernielsentikaer:
Ah, no, because it should open the `view` - I had something else in mind sorry 😊

PR #3514, opened 2022-11-13
teziovsky/react-icons new extension

teziovsky:
## Description

React Icons is an extension for [react-icons npm package](https://react-icons.github.io/react-icons/). It shows all SVG icons, which can be searched and pasted or copied by Name, SVG, or even JSX! 🎉

## Screencast

react-icons-01
react-icons-02
react-icons-03
react-icons-04

## Checklist

### DIFF
--- a/extensions/react-icons/CHANGELOG.md
+++ b/extensions/react-icons/CHANGELOG.md
@@ -4,3 +4,6 @@
 
 Initial version code
 
+## [Update dependencies] - 2022-12-05
+
+Update react-icons from ^4.6.0 to ^4.7.1
--- a/extensions/react-icons/package-lock.json
+++ b/extensions/react-icons/package-lock.json
@@ -7,21 +7,21 @@
       "name": "react-icons",
       "license": "MIT",
       "dependencies": {
-        "@raycast/api": "^1.43.2",
-        "react-icons": "^4.6.0"
+        "@raycast/api": "^1.44.0",
+        "react-icons": "^4.7.1"
       },
       "devDependencies": {
-        "@types/node": "18.11.9",
-        "@types/react": "18.0.25",
-        "@types/react-dom": "^18.0.8",
-        "@typescript-eslint/eslint-plugin": "^5.42.1",
-        "@typescript-eslint/parser": "^5.42.1",
-        "eslint": "^8.27.0",
+        "@types/node": "18.11.10",
+        "@types/react": "18.0.26",
+        "@types/react-dom": "^18.0.9",
+        "@typescript-eslint/eslint-plugin": "^5.45.0",
+        "@typescript-eslint/parser": "^5.45.0",
+        "eslint": "^8.29.0",
         "eslint-config-prettier": "^8.5.0",
-        "prettier": "^2.7.1",
+        "prettier": "^2.8.0",
         "react-dom": "^18.2.0",
         "ts-node": "^10.9.1",
-        "typescript": "^4.8.4"
+        "typescript": "^4.9.3"
       }
     },
     "node_modules/@cspotcode/source-map-support": {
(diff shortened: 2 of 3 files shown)


teziovsky:
Hi, I have completed a new extension called `React Icons`, which shows all icons included in this package, but I have a little question. Is it ok to have all icons, `physically on the device`? I had a problem with passing SVG to `Grid.Item` `source` to show icon, so I decided to download all icons from the package, and reference each file in the assets folder in `Grid.Item` 🤔
Is there any way to pass SVG to `Grid.Item` to show this item on a grid?

I tried to encode SVG to something like that, but it doesn't work.
`data:image/svg+xml;charset=US-ASCII...`
I'm asking about this because react-icons have 26k icons inside 😵‍💫

pernielsentikaer (changes requested):
Hi @teziovsky 👋 

Instead of having 20+ commands in the extension, wouldn't it be better to have just one command with a list dropdown to switch the content?

It feels a bit overwhelming when seeing that many commands 😞

teziovsky:
Hi @pernielsentikaer 👋
I thought about that, but I decided that 20+ other commands will be a lot quicker to choose from a list or event with a shortcut instead of one command with a dropdown. But of course, I can transform that to one command 🤔 That will clean the mess up on the list of commands 🤔 Maybe with a cache to store the last selected option in the dropdown. What do you think about it?

pernielsentikaer:
I think that would be the best approach, `⌘+P` opens the Dropdown by default so it shouldn't take that much longer time to navigate actually 😊

pernielsentikaer:
And I'm not sure if there is a overlap somewhere with all the extensions, have you [checked the list](https://www.raycast.com/search?q=icon)

teziovsky:
> I think that would be the best approach, `⌘+P` opens the Dropdown by default so it shouldn't take that much longer time to navigate actually 😊

Ok, so I will refactor this app to just one command with ⌘+P dropdown 👌



> And I'm not sure if there is a overlap somewhere with all the extensions, have you [checked the list](https://www.raycast.com/search?q=icon)

Yes, I checked this list, and hmm. Maybe It overlaps with [css.gg](https://www.raycast.com/astrit/css-gg), [iconify](https://www.raycast.com/destiner/iconify/commands) and [heroicons](https://www.raycast.com/haack/heroicons), but It's typically for `react-icons` library, so you can super easy copy or paste JSX directly to your editor.
When browsing this extension you will have 100% confidence that those icons exist in your installed library during the React web app development 😉

teziovsky:
@pernielsentikaer I tried to refactor this app to just one command, but the problem is that I need to import all icons into one file. This cause the problem with the JavaScript heap out of memory. The dynamic import not solving the problem, so I think that separating those icons into each file will be the best workaround at this moment.

pernielsentikaer:
Hi @teziovsky 👋 

Yes, we're still trying to figure out the best way to deal with this since the compiled bundle size is around 112MB in total, on solution could a to download it when the user using the extension, but I still need to figure out about the 25+ commands

I'll let you know soon

teziovsky:
@pernielsentikaer I refactored those extensions and now it is around 12MB (with images and without node_modules) in total!
I changed the solution to not download all icons, but pass encoded SVG to `Grid.Image source`. Now just the one problem is that I can't import all 26 icons into one extension, because JavaScript then is heap out of memory.
That's why it's separated into 26 extensions for each icon category.

pernielsentikaer:
Hi @teziovsky 👋 

It's still a tricky one to figure out; splitting it up into multiple commands doesn't scale well and might be a bit awkward to use

The extension size is fine, but the bundle size after installation is, in total, over 100MB, loaded into the memory.
image

**Alternative options** (Not sure if that's even doable)
Download and cache the files in the extension, e.g., as a one-time setup and store the files in the extension's support directory; then search dynamically and load only X items into the grid. or, if scrolling and browsing the collection is essential, at least split them up into categories via a dropdown maybe or support "infinite scrolling" by adding a "load more" entry to the bottom of the list (it's our poor man's workaround to support pagination since we don't have that feature yet in the API)

What do you think, also this is a npm package so I'm not sure we can do much?

teziovsky:
Hi, when I will have a little bit more free time, I will think about it and will try to code your suggestions 🤔

pernielsentikaer:
Yes it's a hard one, it will the current way crash the extension eventually :/

teziovsky:
@pernielsentikaer I think that we should close my pull request because @yug2005 did an amazing job in [this pull request](https://github.com/raycast/extensions/pull/4051) in my opinion. I think that his solution works very well, but you should test it of course 🤔
That's what I wanted to achieve, but he was faster 👿😄 Anyway, @yug2005 Thank you 😄

When I will have an idea to extend your extension, of course, I will let you know 😎

yug2005:
> @pernielsentikaer I think that we should close my pull request because @yug2005 did an amazing job in [this pull request](https://github.com/raycast/extensions/pull/4051) in my opinion. 

Thank you 🙏🙏🙏

> When I will have an idea to extend your extension, of course, I will let you know 😎

For sure, let me know if there are any other commands/actions that you would like to see. I just added an action to quickly download the svg file.

PR #3527, opened 2022-11-15
Updated soundboard

pernielsentikaer:
## Description

* Now it updates the subtitle for a favorite command the first time it's used if a sound has been assigned
* You can move items order in the list by hotkeys

## Screencast

CleanShot 2022-11-15 at 07 35 02@2x

### DIFF
--- a/extensions/soundboard/src/index.tsx
+++ b/extensions/soundboard/src/index.tsx
@@ -1,15 +1,4 @@
-import {
-  ActionPanel,
-  Action,
-  List,
-  Icon,
-  confirmAlert,
-  Color,
-  useNavigation,
-  showToast,
-  Toast,
-  environment,
-} from "@raycast/api";
+import { ActionPanel, Action, List, Icon, confirmAlert, Color, showToast, Toast, environment } from "@raycast/api";
 import { getItems, saveItems } from "./storage";
 import { Item } from "./types";
 import { useState, useEffect } from "react";


extensions/soundboard/src/soundform.tsx
```diff
@@ -2,19 +2,26 @@ import { Form, ActionPanel, Action, Icon, useNavigation } from "@raycast/api";
...
   const { pop } = useNavigation();
-  const { handleSubmit, itemProps } = useForm<SignUpFormValues>({
+  const [favoriteInUse, setFavoriteInUse] = useState<string>("");
+  const { handleSubmit, itemProps, values } = useForm<SignUpFormValues>({
     async onSubmit(values) {
-      props.onEdit(props?.item ? { ...values, id: props.item.id } : { ...values, id: nanoid() });
+      props.onEdit(
+        props?.item
+          ? { ...values, id: props.item.id, last_favourite: props.item.favourite }
+          : { ...values, id: nanoid() }
```
peduarte, line 23: If you wanted one less dependency, you could try using React's [useId](https://beta.reactjs.org/apis/react/useId) Hook instead of `nanoid`

extensions/soundboard/src/utils.ts
```diff
@@ -1,28 +1,66 @@
...
+  await updateCommandMetadata({ subtitle: "Soundboard" });
 };
 
 export const playFile = async (item: Item) => {
   exec(`afplay ${item.path} && $$`);
 };
 
+export const updateFavoriteSubtitles = async (item: Item) => {
+  if (item.favourite !== "0") {
+    await launchCommand({ name: `favorite${item.favourite}`, type: LaunchType.Background });
```
peduarte, line 30: I know it's not possible, but it'd be nice if `launchCommand` supported updating the `title` as well. That way, it would help with the initial UX of which commands can/cannot play a sound. 

- "Set Favourite": when none is set
- "Play Favourite": when it's set (it will actually play it in this case)

(ಠ ›ಠ) — @peduarte — C8eyoS2R000045@2x

extensions/soundboard/src/utils.ts
```diff
@@ -1,28 +1,66 @@
...
 
 export const playSoundFromIndex = async (index: number) => {
   const sound = await getItem(index);
   if (sound) {
-    await playFile(sound);
+    if (environment.launchType === LaunchType.UserInitiated) {
+      await playFile(sound);
+    }
+
+    await updateCommandMetadata({ subtitle: `Sound Name: ${sound.title}` });
```
peduarte, line 13: I know it's not possible, but it'd be nice if `updateCommandMetadata` supported updating the `title` as well. That way, it would help with the initial UX of which commands can/cannot play a sound. 

- "Set Favourite": when none is set
- "Play Favourite": when it's set (it will actually play it in this case)

(ಠ ›ಠ) — @peduarte — C8eyoS2R000045@2x

peduarte (approved):
Very nice 🚀

pernielsentikaer:
LGTM!

isnt-it-romantic-thumbs-up-gif

PR #3559, opened 2022-11-17
Add extension: Auto Quit App

koinzhang:
## Description

Automatically exit the application at regular intervals after closing the application window (if the app does not have any windows, run every 10 minutes).

You can set whether to automatically quit the following applications in the command's preferences.

Includes: Notes, Preview, TextEdit, Messages, Terminal, Shortcuts, TV, QuickTime Player, Mail.

## Screencast

auto-quit-app

## Checklist

### DIFF
--- a/extensions/auto-quit-app/CHANGELOG.md
+++ b/extensions/auto-quit-app/CHANGELOG.md
@@ -1,5 +1,5 @@
 # Auto Quit App Changelog
 
-## [Initial Version] - {PR_MERGE_DATE}
+## [Initial Version] - 2022-12-05
 
 - Automatically quit the application after all windows are closed


pernielsentikaer:
Hi @koinzhang 👋 

This seems like something that people install and forgets about it, and suddenly it's doing things that were not expected. 

I gave it a quick spin, and every time it runs, the preview app spawns and closes fast again 

I'm not feeling that keen about it

koinzhang:
> This seems like something that people install and forgets about it, and suddenly it's doing things that were not expected.

There is no better solution to this problem for now, maybe I can turn it into a menu bar command so that there will be an obvious prompt (but obviously this is not a good solution)

> I gave it a quick spin, and every time it runs, the preview app spawns and closes fast again

I have not encountered such problems in my testing, can you describe in more detail how to reproduce it?

koinzhang:
> There is no better solution to this problem for now, maybe I can turn it into a menu bar command so that there will be an obvious prompt (but obviously this is not a good solution)

To this I should be able to add a judgment: determine whether the application is activated or not, so that when the user puts an application at the top, the extension does not actively close it

koinzhang:
Hi @pernielsentikaer. I tried to turn it into a Menu Bar Command, but couldn’t achieve the same effect.
I find that Background Refresh doesn’t seem to trigger correctly in Menu Bar Command, at least it doesn’t trigger the hooks in the Menu Bar Command.

Also, I must remind that when Auto Quit App exits an app for the first time, the system will pop up a warning box warning whether Raycast is allowed to operate the sub-app, which I think will be a good reminder.

Finally, I made an improvement to the extension so that when an app is active (at the frontmost), the extension will not exit it, which helps avoid accidentally exiting the app the user is opening.

If you encounter any other problems, please feel free to point them out and keep them under review.

koinzhang:
Hi @pernielsentikaer. I've changed the Auto Quit App from a no-view command to a Menu Bar command, which now allows users to tap the menu bar icon to actively exit the app, or open the app in a drop-down menu. Also this command will still automatically quit Enabled all apps every 10 minutes. Please keep reviewing

auto-quit-app

pernielsentikaer (commented):
You can see what's happening here everytime it runs

https://user-images.githubusercontent.com/38033107/203073856-0584a414-070b-4177-b671-9d20d9ee130d.mp4

koinzhang:
> You can see what's happening here everytime it runs

This is strange, it doesn't happen in my case, I'll look into it

pernielsentikaer:
Do you think we should add "No apps.." or something like that when no apps are enabled/disabled? 😊

pernielsentikaer (approved):

PR #3568, opened 2022-11-17
[GitHub] Repository search improvements

thomaslombart:
## Description

- Improve throttling when searching for GitHub repositories
- Hide repositories in "Found Repositories" if they are in "Visited Repositories"
- Fix the notification's icon color for the "Unread Notifications" command
- Add required scopes for personal access tokens

### Before
CleanShot 2022-11-17 at 16 48 49@2x
CleanShot 2022-11-17 at 16 48 42@2x

### After

CleanShot 2022-11-17 at 16 47 27@2x
CleanShot 2022-11-17 at 16 48 10@2x

## Checklist

### DIFF
--- a/extensions/github/src/search-repositories.tsx
+++ b/extensions/github/src/search-repositories.tsx
@@ -1,7 +1,6 @@
 import { List, getPreferenceValues } from "@raycast/api";
 import { useCachedPromise } from "@raycast/utils";
 import { useState, useMemo } from "react";
-import { useDebounce } from "use-debounce";
 
 import RepositoryListEmptyView from "./components/RepositoryListEmptyView";
 import RepositoryListItem from "./components/RepositoryListItem";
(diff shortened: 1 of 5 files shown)


extensions/github/src/search-repositories.tsx
```diff
@@ -16,12 +17,14 @@ function SearchRepositories() {
   const preferences = getPreferenceValues<{ includeForks: boolean }>();
 
   const [searchText, setSearchText] = useState("");
+  const [debouncedSearchText] = useDebounce(searchText, 200);
```
mathieudutour, line 20: how does that improve the throttling of the search? What's the issue in the first place?

thomaslombart: That one isn't detailed enough indeed. The problem was that with `throttle` on the `List` component, the visited repositories would be filtered asynchronously even though it can be done immediately. Here's a comparison where the 2nd one is slightly faster.

## Before

https://user-images.githubusercontent.com/16003285/202664563-7367fb06-9f08-4384-8422-7622af86e9f2.mp4

## After

https://user-images.githubusercontent.com/16003285/202664285-853a417b-0f1c-41a9-af44-49fb088d56e4.mp4

mathieudutour: that seems to be because of the `foundRepositories` that you introduced, not the debounce thing

thomaslombart: Fair enough

mathieudutour (approved):

PR #3576, opened 2022-11-18
Obsidian: Fix hardcoded searchText

mathieudutour:
## Description

The `searchText` was a constant value - so searching didn't work

## Checklist

### DIFF
--- a/extensions/obsidian/CHANGELOG.md
+++ b/extensions/obsidian/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Obsidian Changelog
 
+## [Fix] - 2022-11-18
+
+- Fixed a bug where the searchbar text didn't update
+
 ## [Small fix | Version 1.9.1] - 2022-10-28
 
 - Fix command execution with shortcut not working


marcjulianschwarz:
This seems to have partially fixed the problem. However when trying to quickly write text in the search bar, parts of the text will be removed while writing. 
I am testing this on a vault with about 3,000-4,000 notes. 

Might be a problem in parts of my code which now became apparent. 

https://user-images.githubusercontent.com/67844154/202669792-36e1fc04-ed6b-4585-a659-adb86c5fff5b.mov

marcjulianschwarz:
In `NoteListObsidian.tsx` I am using the `useEffect` hook on an empty array which should only run once. In there I load the notes from cache. 
I just fixed the ShowMentioningNotesAction to not load the notes again.

Now only the search is running when typing into the search bar and still text gets removed from the searchText.

marcjulianschwarz:
I rewrote lots of parts that were directly or indirectly connected to the search in the [newest commit here](https://github.com/marcjulianschwarz/obsidian-raycast).

However the issue still persists and I have no idea what could cause it.

mathieudutour:
you could try to add `throttle` to the `List` maybe

marcjulianschwarz:
You can still get it to leave out letters when typing very very fast but for normal fast typing it is now way better than before. Thank you.

marcjulianschwarz:
While testing the current version I noticed that the problem is actually still fairly visible. I mistakenly thought that letters were removed when typing fast but it turns out it happens much more often when typing slowly.

marcjulianschwarz:
I made this simple example command where I experience the same issues as described above.

```tsx
  const values = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10];

  const [text, setText] = useState("");

  return (
    <List
      throttle={true}
      onSearchTextChange={(text) => {
        setText(text);
      }}
      searchText={text}
    >
      {values.map((value) => (
        <List.Item key={value} title={value.toString()}></List.Item>
      ))}
    </List>
  );
```

PR #3577, opened 2022-11-18
[One Thing]: Add preferences to the extension

thomaslombart:
## Description

- Add optional prefix and suffix preferences to prepend or append to the one thing (e.g emojis)
- Add a preference to close Raycast's window defaulting to `true`
- Set the one thing text field to use by default the cached value

## Screencast

CleanShot 2022-11-18 at 11 20 16@2x

https://user-images.githubusercontent.com/16003285/202679536-aa540722-c937-4e7f-9825-d10f2750694b.mp4

## Checklist

- [x] I checked that assets used by the `README` are placed outside of the `metadata` folded

### DIFF
--- a/extensions/one-thing/package.json
+++ b/extensions/one-thing/package.json
@@ -5,7 +5,10 @@
   "description": "Put a single task or goal in your menu bar",
   "icon": "command-icon.png",
   "author": "pernielsentikaer",
-  "contributors": ["mathieudutour", "thomaslombart"],
+  "contributors": [
+    "mathieudutour",
+    "thomaslombart"
+  ],
   "license": "MIT",
   "commands": [
     {


thomaslombart:
There's one annoying bug with the preferences: the menu bar command is re-launched after each modification in the preferences which triggers the `Set One Thing` command (as clicking on the menu bar launches this command)

https://user-images.githubusercontent.com/16003285/202680069-9902bb08-7fa9-4be1-967c-25e430bbc616.mp4

mathieudutour (approved):

pernielsentikaer:
Both ways work for me; I like the first since it keeps it simple - do you have a solution for that?

If everything goes wrong, then solution 2 works too, which is also nice 😁

mathieudutour:
@sxn do you think we could do something like that? Perhaps launch the menu bar in background mode if supported?

sxn:
It could make sense to launch the command with launchType `background` if it's supported, yeah.
Regardless of that, if `MenuBarExtra` had an `onAction`, we could differentiate between these two situations, and the `MenuItem` component wouldn't be necessary… 🤔

thomaslombart:
Yeah, `onAction` on `MenuBarExtra` would work as well. It makes also more sense from a DX point of view in my opinion.

pernielsentikaer:
So what are we going with? 😊

pernielsentikaer:
Is this ready @thomaslombart

thomaslombart:
The bug is still there and still pending for a solution on the Raycast API, so not ready at the moment 😊

PR #3579, opened 2022-11-18
Airtables - Added custom no-view, Corrected naming errors

pernielsentikaer:
## Description

Added custom no-view
Corrected naming errors

## Screencast

image

image

### DIFF
--- a/extensions/airtable/src/BaseSchemaListView.tsx
+++ b/extensions/airtable/src/BaseSchemaListView.tsx
@@ -2,6 +2,7 @@ import { Action, ActionPanel, Detail, Icon, List, showToast, Toast } from "@rayc
 import { useEffect, useState } from "react";
 import { AirtableBaseMetadata, Field, Table, View } from "./types";
 import * as api from "./metadata-api";
+import { Fragment } from "react";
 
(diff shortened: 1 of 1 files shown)


pernielsentikaer:
Could you check this @marks

marks:
Thanks @pernielsentikaer ! Looks good to me and appreciate your help. 

Im mobile and on vacation so I can't test it but recommend merging. 

Do you think the other lists should be updated to show the number of results similarly in the list header?

Thanks again!

pernielsentikaer:
Thanks for the approval, I'll check the rest of the code and add some more corrections 😊

marks:
Awesome. I wouldn't expect any other lists to have 0 results (though not a bad idea to have a fallback), I just assume it'd be desirable to have consistent headers

pernielsentikaer:
It's also showing the [List.EmptyView](https://developers.raycast.com/api-reference/user-interface/list#list.emptyview) when you search doesn't find a match, should be a bit more consistent now 😊

pernielsentikaer:
Let's update with this, if anything pops up I will be available to correct it 😊

seal-of-approval-thumbs-up-gif

PR #3589, opened 2022-11-18
[AWS] Add S3 Command

JonathanWbn:
## Description

Adding an "S3 Buckets" command. Searching and navigating the AWS S3 Console is not a great user experience (search only by prefix, objects are nested, ...). This command provides a better way to find and open buckets and objects.

With this PR I'm also creating more consistency in the folder structure, file names, and copy. These changes pollute the diff view, so this PR might be easiest to review on a commit basis.

## Screencast

https://user-images.githubusercontent.com/19997520/202785952-daa4760d-6a0f-4fe2-b371-2443c93ab8a1.mov

### DIFF
--- a/extensions/amazon-aws/src/s3.tsx
+++ b/extensions/amazon-aws/src/s3.tsx
@@ -1,9 +1,13 @@
-import { ActionPanel, List, Detail, Action, Icon } from "@raycast/api";
+import fs from "fs";
+import { homedir } from "os";
+import { Readable } from "stream";
+import { ActionPanel, List, Detail, Action, Icon, showToast, Toast } from "@raycast/api";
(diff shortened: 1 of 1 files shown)


duboiss:
Awesome. I'm wondering if we can add a download action for bucket objects ?

Hodglim:
Looks great!  I agree, the ability to download from S3 would be amazing if possible.

JonathanWbn:
@duboiss @Hodglim Agree! I added a "Download" action to the object list item, which will download the file to the user's `Downloads` folder.

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #3594, opened 2022-11-18
Fix empty homepage URLs for "NixPkgs Search"

arulagrawal:
## Description

As far as I can tell, this extension currently crashes when the response contains homepage URLs that are empty, for example if we receive `['']` as the `result.package_homepage`. 

As an example, searching for `yarn` results in this behaviour.

This PR simply filters out empty string values from `result.package_homepage`.

## Checklist

### DIFF
--- a/extensions/nixpkgs-search/src/index.tsx
+++ b/extensions/nixpkgs-search/src/index.tsx
@@ -277,7 +277,7 @@ async function performSearch(searchText: string, signal: AbortSignal): Promise<S
       attrName: result.package_attr_name,
       description: result.package_description,
       version: result.package_pversion,
-      homepage: result.package_homepage,
+      homepage: result.package_homepage.filter((url) => url?.length > 0),
       source:
         result.package_position &&
         `https://github.com/NixOS/nixpkgs/blob/nixos-unstable/${result.package_position.replace(/:([0-9]+)$/, "")}`,


pernielsentikaer:
@aiotter can you approve this correction?

aiotter:
I want to see the actual data for `homepage` being zero-length string. It is highly possible that it is not a valid data. This can be a bug on the upstream. We may be better fixing [NixOS/nixos-search](https://github.com/NixOS/nixos-search) or [NixOS/nixpkgs](https://github.com/NixOS/nixpkgs) then.

aiotter:
Oh well, the author provided an example: `yarn` package. I'll take a look into it.

aiotter:
`nixpkgs#yarn2nix`'s `meta.homepage` is `""`. ([source](https://github.com/NixOS/nixpkgs/blob/cf63ade6f74bbc9d2a017290f1b2e33e8fbfa70a/pkgs/development/tools/yarn2nix-moretea/yarn2nix/default.nix#L380))
On the other hand, it seems NixOS/nixos-search does not expect empty string being in `meta.homepage`. ([source](https://github.com/NixOS/nixos-search/blob/f249d4964e38b13f32e25634145e06853db41657/frontend/src/Page/Packages.elm#L1028-L1039))

I'll create an issue on NixOS/nixos-search and ask them who is responsible for the issue.
If NixOS/nixos-search is to be fixed, then this extension should be fixed too.

aiotter:
This issue is going to be fixed on https://github.com/NixOS/nixpkgs/pull/202328

ncfavier:
At a quick glance I think the crash comes from `new URL(url)` which raises an exception when `url` is not a valid URL. Sounds like you want to catch that exception.

arulagrawal:
Not sure what raycast's (and @aiotter's) stance is on whether or not it should be fixed here.

Also not sure about whether it should be fixed with what I did or via exception handling.

pernielsentikaer:
Sounds like a good thing to catch the example, I guess it's not the last time this could happen 😊

aiotter:
I feel it's better to error out instead of silently ignore it when the extension receives invalid data. However I agree that crashing should be avoided. The best solution would be like showing "invalid data" error message at the detail pane of the invalid package.

Nixpkgs is a community-driven repository and Nix has little error detection system about properties of derivations. There can be other format violations than zero-length homepage urls.

aiotter:
I'm now in the middle of relocating my home from Kyoto to Tokyo. I'll not available until all the things set up, which takes 2w or more, so my patch for this will come later. Instead, I think I have some time to review PRs, so I welcome them if you are interested.

pernielsentikaer:
Crashes should be avoided for sure, they should be gracefully handled 😊

PR #3596, opened 2022-11-19
New Extension: Search Zendesk support center

cpenned:
## Description

The extension searches any Zendesk support center for all published, public articles. Located articles show rendered HTML and allow the user to: 
- Open the article in a browser (default)
- Copy the article URL to the system clipboard
- Paste the article URL in an active application

## Screencast

Setup requires a Zendesk support center base. You can optionally provide a locale for the support center. 

zendesk-1

Searching for an article returns all matches, showing a full HTML preview of the article text.

zendesk-2

Open, copy, or paste the URL in your active application.
zendesk-3

## Checklist

### DIFF
--- a/extensions/zendesk/src/search-zendesk-support-center.tsx
+++ b/extensions/zendesk/src/search-zendesk-support-center.tsx
@@ -33,38 +33,44 @@ export default function ZendeskSearch() {
         ))}
       </List.Dropdown.Section>
     </List.Dropdown>
-  )
+  );
 
-  const {data}: LocaleFetchRes = useFetch(`https://${supportCenter}/api/v2/locales/`, {
+  const { data }: LocaleFetchRes = useFetch(`https://${supportCenter}/api/v2/locales/`, {
     keepPreviousData: true,
     onData: (data) => {
-      if(data) {
-        return setLocales(data.locales.map((locale) => ({
-          name: locale.name,
-          locale: locale.locale,
-        })
-        ))
+      if (data) {
+        return setLocales(
+          data.locales.map((locale) => ({
+            name: locale.name,
+            locale: locale.locale,
+          }))
+        );
       }
     },
   });
 
   useEffect(() => {
     setSelectedLocale(locales[0]);
-  }, [locales])
+  }, [locales]);
 
   const {
     isLoading,
     data: articles,
     error,
-  }: ArticleFetchRes = useFetch(`https://${supportCenter}/api/v2/help_center/articles/search.json?query=${query}&locale=${selectedLocale?.locale ? selectedLocale?.locale : ""}`, {
-    keepPreviousData: true,
-    onError: (error) => {
-      if (error.message === "Bad Request") {
-        return;
-      }
-      showToast({ title: "Error", message: error.message, style: Toast.Style.Failure });
-    },
-  });
+  }: ArticleFetchRes = useFetch(
+    `https://${supportCenter}/api/v2/help_center/articles/search.json?query=${query}&locale=${
+      selectedLocale?.locale ? selectedLocale?.locale : ""
+    }`,
+    {
+      keepPreviousData: true,
+      onError: (error) => {
+        if (error.message === "Bad Request") {
+          return;
+        }
+        showToast({ title: "Error", message: error.message, style: Toast.Style.Failure });
+      },
+    }
+  );
 
   const [searchResult, setSearchResult] = useState<FilteredArticle[] | undefined>(undefined);
 
@@ -104,11 +110,11 @@ export default function ZendeskSearch() {
       {searchResult ? (
         searchResult.map((item) => {
           if (item.url === undefined) {
-            return null
-          };
+            return null;
+          }
           return (
             <List.Item
-              key={item.id}
+              key={item.title}
               title={item.title}
               detail={<List.Item.Detail markdown={item.body} />}
               actions={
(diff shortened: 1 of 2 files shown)


zendesk/package.json
```diff
@@ -0,0 +1,60 @@
...
+      "placeholder": "e.g., support.mycompany.com",
+      "description": "The base URL for your Zendesk support center. When viewing your support articles online, the URL will be in the format https://support.mycompany.com/hc/en-us/articles/3600000000000-My-Article-Title. The base URL is everything after the https:// and before the /hc/ part of the URL."
+    },
+    {
+      "name": "locale",
+      "type": "textfield",
+      "required": false,
+      "title": "Support center locale",
+      "placeholder": "e.g., en-us",
+      "description": "Defaults to en-us if no locale is provided."
```
peduarte, line 37: Are all locales supported? Thinking whether we could display this in a [dropdown](https://developers.raycast.com/information/manifest#preference-properties) instead of textfield

peduarte (changes requested):
Could you make sure the extension contains the following files too:
- `.eslintrc.json`
- `.prettierrc`
- `.gitignore`

You can copy them over from other extensions 👍

zendesk/src/search-zendesk-support-center.tsx
```diff
@@ -0,0 +1,100 @@
...
+        searchResult.map((item) => {
+          if (item.url === undefined) return;
+          return (
+            <List.Item
+              title={item.title}
+              detail={<List.Item.Detail markdown={item.body} />}
+              actions={
+                <ActionPanel>
+                  <Action.OpenInBrowser
+                    title="Open in Browser"
```
peduarte, line 77: You can leave out the `title` here, since its the default when you use [`Action.OpenInBrowser`](https://developers.raycast.com/api-reference/user-interface/actions#action.openinbrowser)

zendesk/src/search-zendesk-support-center.tsx
```diff
@@ -0,0 +1,100 @@
...
+                    url={`https://${supportCenter}/hc/${locale}/articles/${item.id}`}
+                  />
+                  <Action.CopyToClipboard content={item.url} shortcut={{ modifiers: ["cmd"], key: "enter" }} />
```
peduarte, line 80: You can leave out `shortcut` here if you want, since ⌘  ⏎ will be used by default as this is the second Action

zendesk/src/search-zendesk-support-center.tsx
```diff
@@ -0,0 +1,100 @@
...
+      throttle={true}
+    >
+      {searchResult ? (
+        searchResult.map((item) => {
+          if (item.url === undefined) return;
+          return (
+            <List.Item
+              title={item.title}
+              detail={<List.Item.Detail markdown={item.body} />}
+              actions={
```
peduarte, line 74: Probably missing a `key` prop here. Once you add `.eslintrcjson` it should warn you 👍

zendesk/src/search-zendesk-support-center.tsx
```diff
@@ -0,0 +1,100 @@
...
+            />
+          );
+        })
+      ) : (
+        <List.EmptyView
+          title={getEmptyViewText()}
+          icon={Icon.AppWindowList}
+          actions={
+            <ActionPanel>
+              <Action title="Search for Article" icon={Icon.Binoculars} />;
```
peduarte, line 93: What does this Action do?

zendesk/package.json
cpenned: Ah, that's a good point. Looking closer into the Zendesk documentation, there's a way to fetch which languages are available for each support center. Perhaps I can make it a searchBarAccessory dropdown? And then whatever they select, in the dropdown persists over uses? Now I just need to figure out how to do that if that's the right direction?

peduarte: Yeah, I think the `Dropdown` Component could work well here 👌

extensions/zendesk/src/search-zendesk-support-center.tsx
```diff
@@ -0,0 +1,137 @@
...
+    >
+      <List.Dropdown.Section title="Available Locales">
+        {locales.map((locale) => (
+          <List.Dropdown.Item key={locale.name} title={locale.name} value={locale.locale} />
+        ))}
+      </List.Dropdown.Section>
+    </List.Dropdown>
+  );
+
+  const { data }: LocaleFetchRes = useFetch(`https://${supportCenter}/api/v2/locales/`, {
```
peduarte, line 38: Looks like `data` isnt used. Do we need it?

peduarte: Could maybe be: 
```tsx
  useFetch(`https://${supportCenter}/api/v2/locales/`, {
    keepPreviousData: true,
    onData: ({ data }: LocaleFetchRes) => {
      if (data) {
        return setLocales(
          data.locales.map((locale) => ({
            name: locale.name,
            locale: locale.locale,
          }))
        );
      }
    },
  });
```

peduarte (commented):
Much better, thanks for the changes.

I dont love that the initial view is empty, is there anything we can do about that?

Eg: Retrieve the top 10 articles? 

image

cpenned:
@peduarte The Zendesk API returns the following when a query isn't included: "The query must contain at least one of the following parameters: query, category, section or label_names."

I could hit a different endpoint to get only article names and urls, but it would need to be an entirely different data view without the HTML preview. Thoughts?

peduarte:
I see, ok. Let's do with that then. Thanks!

peduarte (approved):

Issue #3222, opened 2022-10-14
Support multiple installations/configurations of the same extension


EvansMatthew97:
# Extension – [All extensions](https://github.com/raycast/extensions/tree/main/extensions/*)

Author: @raycast

Support multiple installations/configurations of the same extension

Currently only allows a single installation or configuration per extension. Feature requested would allow the user to create multiple configurations for a single extension. The primary benefit is multi-account support out of the box for every extension.

Possible implementations:
1. Allow users to install the same extension twice, but must assign an alias for repeat installs
2. Allow multiple configurations for the same extension, request that the user choose a configuration each time they type the command (worse user experience than option 1)
3. Multiple Raycast "profiles" - largest change but possibly the largest benefit. The user can have profiles that are either isolated or inherit other profiles' configuration, and can switch between them using a command. This could affect the entire Raycast app config from colour scheme to which extensions are installed based and their configurations on the chosen profile.
Example use case is a "work" and "personal" profile, for BYOD and freelance developers.

This benefits all users who want to use multiple accounts with a single extension. Examples of users requesting multi-account support for different extensions:
- https://github.com/raycast/extensions/issues/305
- https://github.com/raycast/extensions/issues/1517
- https://github.com/raycast/extensions/issues/3057
- https://github.com/raycast/extensions/issues/2741

weitzman:
This is a valid feature request. I dont much benefit in closing it as stale.

Issue #3057, opened 2022-09-29
[Amazon AWS] Multiple profiles


skrollme:
# Extension – [Amazon AWS](https://github.com/raycast/extensions/tree/4a1e82efc770015ce881a5480765e5e949248ac2/extensions/amazon-aws/)
Author:

It would be great if there is a possibility to switch between multiple AWS accounts (credentials) on the fly without going back to the raycast-settings. E.g. via a dropdown in the top/right like in other extensions with multi-account-support.

**Describe the feature and the current behavior/state.**
At the moment I always have to go the settings when I want to switch the AWS accounts

**Who will benefit with this feature?**
Everyone with more than one AWS account ;) 

**Any Other info.**

Issue #3022, opened 2022-09-26
[Amazon AWS] Support for AWS Vault


themanontv:
# Extension – [Amazon AWS](https://github.com/raycast/extensions/tree/4a1e82efc770015ce881a5480765e5e949248ac2/extensions/amazon-aws/)
Author: @victor-falcon

 It would be great if the AWS extension for Raycast supported AWS Vault as this is a common way people store AWS Credentials

**Describe the feature and the current behavior/state.**
The feature would involve using AWS vault to get credentials for an AWS profile as well as the current situation of just using the ~/.aws/credentials file

**Who will benefit with this feature?**
AWS professionals who need to store sensitive credentials and want a secure way to store them but also want the convenience of using Raycast

**Any Other info.**
Ideally there should be a way to select the current profile being used as typically there will be many within a vault.
https://github.com/99designs/aws-vault

themanontv:
I would still like this feature doing as it would be incredibly useful

themanontv:
Accidentally closed the issue with a slip of the finger :)

JonathanWbn:
@themanontv Does #3612 solve your problem?

themanontv:
> @themanontv Does #3612 solve your problem?

It doesn't no, AWS vault is a method of storing AWS profiles. It's certainly a partial workaround but for people that want to securely hold credentials it's not a proper solution.

themanontv:
@JonathanWbn If you try to use the extension with AWS vault then you get a challenge for MFA that the extension doesn't know how
 to handle like in the screenshot below. This seems like it's theoretically possible to add this to this plugin?
![SCR-20221202-ews](https://user-images.githubusercontent.com/39599137/205275966-be948368-04a0-4d97-a964-7d737937e626.jpeg)

I'll give you the link to AWS vault https://github.com/99designs/aws-vault

themanontv:
Potentially this could be solved by presenting the MFA code prompt to the user to respond to

JonathanWbn:
@themanontv Yes, with the latest changes we have certainly moved closer to supporting `aws-vault`. I have this on my list of things I'd like to improve.

themanontv:
I would still like this ticket to be in the backlog

JonathanWbn:
[...]

PR #3612, opened 2022-11-21
[AWS] Support Multiple Profiles

JonathanWbn:
## Description

This removes the hardcoded AWS Profile and Region setting in favor of a simple profile select.

This approach comes with several advantages:
- extension becomes zero-config
- adds profile switching inside the extension via list dropdown _(cached between commands)_
- supports all credential providers ([config and credentials files](https://docs.aws.amazon.com/sdkref/latest/guide/file-format.html), [environment variables](https://docs.aws.amazon.com/sdkref/latest/guide/environment-variables.html), [SSO](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/clients/client-sso/index.html)...)
- simplifies the extension by relying on the built-in behavior of the v3-SDK
- solves
  - #3057
- relates to
  - #3222 
  - #3022

### Demo 1 - Profile Switching

_Notice how the profile is persisted between the "tables" and "stacks" commands._

https://user-images.githubusercontent.com/19997520/203145075-eeeabaf0-9a35-40f0-8d11-472f62a4b705.mov

### Demo 2 - SSO

_In this example, the default profile uses credentials from the `~/.aws/credentials` file, while the `test_sso_profile` uses the `/.aws/config` file to read credentials. Notice that the CLI-login step is only required when the user is logged out. In most workflows, I expect the user to be already logged in._

https://user-images.githubusercontent.com/19997520/204146519-aa4d0321-caf0-4525-8755-4996ce509d6e.mp4

### DIFF
--- a/extensions/amazon-aws/CHANGELOG.md
+++ b/extensions/amazon-aws/CHANGELOG.md
@@ -1,7 +1,6 @@
 # Amazon AWS Changelog
 
-## [Support AWS Profile Switch] - 2022-11-28
-
+## [Support AWS Profile Switch] - 2022-11-30
 Add "Profile" select replacing the "Profile" and "Region" setting ([#3612](https://github.com/raycast/extensions/pull/3612))
 
 ## [S3 Command] - 2022-11-18


duboiss:
I've been waiting for support for this for a long time, I didn't even think we could go this far. Thanks a lot, I will try to take the time to test it.

pernielsentikaer:
Looking forward to the tests 🔥 

Awesome @JonathanWbn 🚀

JonathanWbn:
@duboiss, did you get a chance to test this yet?

victor-falcon:
Great, nice feature. This makes the extension a lot better 🎉

duboiss:
> @duboiss, did you get a chance to test this yet?

It turns out that at work we changed the configuration. The `config` file now contains the profiles but using sso in the following format:
```
[default]
sso_start_url = XXX
sso_region = eu-central-1
sso_account_id = XXX
sso_role_name = PowerUserAccess
region = eu-central-1
output = json

[profile XXX]
sso_start_url = XXX
sso_region = eu-central-1
sso_account_id = XXX
sso_role_name = PowerUserAccess
region = eu-central-1
output = json

[profile XXX]
sso_start_url = XXX
sso_region = eu-central-1
sso_account_id = XXX
sso_role_name = PowerUserAccess
region = us-east-1
output = json

...
```

And that we now have no more `credentials` file. As a result, even before testing your changes, the current state of the extension no longer works for me.
image
I haven't yet taken the time to see if I can continue to use the extension from now on... by changing something.

JonathanWbn:
> It turns out that at work we changed the configuration. The config file now contains the profiles but using sso

@duboiss I have taken this as an opportunity to re-think the whole approach and ended up with something much better: Just setting the `AWS_PROFILE` environment variables and letting the AWS SDK v3 do the rest. This happens to support **all** credentials providers (including SSO). With these changes, I expect the extension to work for you again (see demo 2 in the PR description)

@duboiss @victor-falcon This new approach required significant changes. I have updated the description accordingly. I appreciate any review or testing.

duboiss:
Awesome ! After a reminder to use `aws sso login`, everything works great.
Howerver, I have a weird behavior for S3. As a reminder, S3 is a global service.

```ini
[profile A]
sso_start_url = XXX
sso_region = eu-central-1
sso_account_id = XXX
sso_role_name = PowerUserAccess
region = us-east-1
output = json

[profile B]
source_profile = XXX
region = eu-central-1
output = json
role_arn = XXX
```
On the profiles above, for the second one, "B", I got this :
image
A bucket that I should not have access.
image

I can also show you an other profile where I see rds snapshot bucket...
image

Thanks a lot, you made a great work

duboiss:
@JonathanWbn My bad. By selecting a profile on the console I have these same files. I thought it was at AWS.
Except for a bucket that I can open on the console and not in Raycast (second screenshot above), everything is good.

JonathanWbn:
@duboiss Thanks for the feedback! I have addressed the S3 problem. Do you mind testing again?

pernielsentikaer:
It's really awesome to follow this extension, good work @JonathanWbn 🔥

duboiss:
image
Do you think that we can add a button to run "aws sso login" without the need to open our own terminal ?
Maybe complicated. The output of this command terminates only when we click on "Allow" on the AWS page that opened (or with ctrl+c).
Once clicked, it seems that cmd + r reloads this error directly into the display. But maybe it can also be automated.

For your update @JonathanWbn, now I have this
image

JonathanWbn:
> Do you think that we can add a button to run "aws sso login" without the need to open our own terminal ?

@duboiss I have tried this before, but I gave up on it eventually.

> For your update @JonathanWbn, now I have this

Then unfortunately my workaround is not robust. At this point, I would instead choose good error handling than try to cover a bunch of edge cases. So here's the error screen you should see now:
Screenshot 2022-11-29 at 19 19 10

@duboiss Could you confirm that this error handling is working for you? If so, I would love to get this PR merged.

duboiss:
Same behavior as you, much better. That's ok for me, thanks.

JonathanWbn:
> Same behavior as you, much better. That's ok for me, thanks.

Great!
@pernielsentikaer I would now consider this PR ready to be merged. 🚢

pernielsentikaer (approved):
Thanks for your hard work @JonathanWbn (and all contributors/testers)

arabshapt:
Hi everyone, is it possible to open e.g. secretsmanager in specific profile? In raycast I can switch profiles with cmd+P and then search for secrets within that profile, but when I open it in browser by hitting cmd+K and then navigating to "open Secret" I need to switch the profile manually in the browser... Is It possible to add that information in the query(in URL) to switch the profile automatically? It would also be great to have a shortcut to open that specific secret(or anything else like) without cmd+K plus navigation.
Thank you for this extension!

Issue #3540, opened 2022-11-16
[Set Audio Device] "Favourite output audio device could not be set"


JustDMare:
# Extension – [Set Audio Device](https://github.com/raycast/extensions/tree/9705ba94af72208cff790222ee3ce28e02231825/extensions/audio-device/)
Author: @benvp

## Description
In macOS Ventura the command for selecting your favourite output device always raises the error "Favourite output audio device could not be set".

## Steps To Reproduce

1. Use "Set output device" command.
2. In that command's list, select any output device and copy the device's name through the "Copy device name to clipboard"
3. Go to Raycast settings and paste the device's name in the "Favourite Output Device" input for this extension's settings.
4. Execute the command "Set Output device to favourite"

## The current behavior
Raycast shows the error "Favourite output audio device could not be set"

## The expected behavior
Changing output audio device to the one marked as favourite, in my case, my Airpods.

### Version
Device: Macbook Pro 14" M1 Pro
OS: macOS Ventura 13.0.1
Raycast: 1.43.2

PS: The exact same steps work in my Intel i5 macbook (2020) with macOS Monterrey 12.4 installed, so my guess is that it is caused by macOS Ventura.

Issue #3526, opened 2022-11-15
[Set Audio Device] Favourite output audio device could not be set


tsugur:
# Extension – [Set Audio Device](https://github.com/raycast/extensions/tree/9705ba94af72208cff790222ee3ce28e02231825/extensions/audio-device/)
Author: Raycast

## Description

I have my AirPods as my favourite output device
When I run the **Set Output Device to Favourite** command, I get an error: ”Favourite output audio device could not be set”
My AirPods are called ”Tsugur’s AirPods Pro”
The name is entered properly: no typos, no excessive whitespaces etc.

## Steps To Reproduce

1. Set favourite device to "Tsugur's AirPods Pro"
2. Run  **Set Output Device to Favourite** command

## The current behavior
After running the **Set Output Device to Favourite** command, I see a red dot with a ”Favourite output audio device could not be set” message. The output device is not set

## The expected behavior
The favourite output device is successfully set as expected

### Raycast version

1.43.2

Issue #3479, opened 2022-11-11
[Set Audio Device] ...favourite not working


stanty522:
# Extension – [Set Audio Device](https://github.com/raycast/extensions/tree/9705ba94af72208cff790222ee3ce28e02231825/extensions/audio-device/)
Author: @benvp

## Description
When using the set output device to favourite it does not work. I used my airpods pro in this example.

## Steps To Reproduce

1. Set airpods pro as the favourite output device
2. Use command set output device to favourite

## The current behavior
It fails to connect

## The expected behavior
It connects

### Raycast version
Version: 1.43.2

Issue #3446, opened 2022-11-09
[Set Audio Device] ...


Jinwannng:
# Extension – [Set Audio Device](https://github.com/raycast/extensions/tree/9705ba94af72208cff790222ee3ce28e02231825/extensions/audio-device/)
Author:

## Steps To Reproduce

1.
2.

Jinwannng:
Raycast version 1.42.2

Jinwannng:
Issue: cannot set output device to HomePod(Airplay)

1. before toggle the [include airplay devices], the search results of output devices could be properly showed in the Raycast.
2. After toggle the [include airplay devices], **No Search results** anymore.
3. The [input devices] works well.

Raycast version 1.42.2
extension: latest, I've uninstalled and reinstalled the plugin.

PR #3613, opened 2022-11-21
[Set Audio Device] Fix for macOS Ventura

### LINKED ISSUE #3542: [Set Audio Device] ...
# Extension – [Set Audio Device](https://github.com/raycast/extensions/tree/9705ba94af72208cff790222ee3ce28e02231825/extensions/audio-device/)
Author: @benvp

## Steps To Reproduce

1. Set a favourite device in Raycast Settings -> Extensions -> Set Audio Device
2. Use raycast to select Set output device to Favourite

## The current behavior
Fails with message "Favourite output audio device could not be set"

## The expected behavior
Audio output is set to the favourite device

### Raycast version
Version: 1.43.2

gkoberger:
## Description
Closes #3542 #3415 #3446 #3479 #3526 #3540 #3349

Extension: https://www.raycast.com/benvp/audio-device
Author: @benvp 

This updates Set Audio Device to work with macOS Ventura 13. The issue was that the Settings panel was redesigned in the upgrade, so the Apple Script that powers this when Airplay is checked on (and also powers the Favorite command) no longer works.

**🚨 NOTE:** There is one big problem, which is that it opens the Settings modal rather than doing it behind the scenes. I'm still trying to figure out if it's possible to hide it in Ventura, but maybe @benvp knows how to do that?

## Checklist

### DIFF
--- a/extensions/audio-device/package-lock.json
+++ b/extensions/audio-device/package-lock.json
@@ -4,10 +4,11 @@
   "requires": true,
   "packages": {
     "": {
-      "name": "audio-device",
       "license": "MIT",
       "dependencies": {
         "@raycast/api": "^1.29.1",
+        "airplayer": "^2.0.0",
+        "apple-system-profiler": "^1.1.0",
         "execa": "^6.0.0",
         "macos-audio-devices": "^1.4.0",
         "macos-version": "^5.2.1",
@@ -457,6 +458,42 @@
         "acorn": "^6.0.0 || ^7.0.0 || ^8.0.0"
       }
     },
+    "node_modules/airplay-protocol": {
+      "version": "2.0.2",
+      "resolved": "https://registry.npmjs.org/airplay-protocol/-/airplay-protocol-2.0.2.tgz",
+      "integrity": "sha512-Pm25h2kCF51mA8dg1froC2CSwZknAvyXNRFjlafhcs6yQB5R+heddwSVMqRp7KBQyk5j/nhTq1JxF3pXdwHXVg==",
+      "dependencies": {
+        "bplist-creator": "^0.0.6",
+        "bplist-parser": "^0.1.1",
+        "concat-stream": "^1.5.1",
+        "plist": "^1.2.0",
+        "reverse-http": "^1.2.0"
+      },
+      "engines": {
+        "node": ">= 0.12"
+      }
+    },
+    "node_modules/airplayer": {
+      "version": "2.0.0",
+      "resolved": "https://registry.npmjs.org/airplayer/-/airplayer-2.0.0.tgz",
+      "integrity": "sha512-oQFi0K+TLDpw8vTeDnkX1opLEGPJ03OeT5dKzYqPHf6JjuaYmXoNdc5GQhZDF+IzX8ypWvYZ/NbftabNU02dkA==",
+      "dependencies": {
+        "airplay-protocol": "^2.0.2",
+        "appendable-cli-menu": "^2.0.0",
+        "bonjour": "^3.3.1",
+        "internal-ip": "^1.2.0",
+        "mime": "^1.3.4",
+        "minimist": "^1.2.0",
+        "range-parser": "^1.0.3",
+        "server-destroy": "^1.0.1"
+      },
+      "bin": {
+        "airplayer": "bin.js"
+      },
+      "engines": {
+        "node": ">= 0.12"
+      }
+    },
     "node_modules/ajv": {
       "version": "6.12.6",
       "resolved": "https://registry.npmjs.org/ajv/-/ajv-6.12.6.tgz",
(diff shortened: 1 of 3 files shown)


pernielsentikaer:
HI @gkoberger 👋 

I just tested this in the latest beta os macOS 13 (13.1 Beta 3) and I don't see anything that pops or flashes - do you?

@benvp do you mind checking this

gkoberger:
@pernielsentikaer thanks for checking it out!

Based on my understanding of the original code, most of the time it uses the `audio-devices` binary in `audio-device.ts`, which works like a charm. However, there's two situations when it uses an Applescript: when "Airplay" is checked in the settings, or when using the "Favourite" feature.

My assumption is that because Airplay devices don't show up in the binary being used, the original authors decided an Applescript was the best/only way to do it.

So! You can try to cause it by using "Set Output Device to Favorite" to trigger the Applescript. If that works without a flash, then there's no problem!

(It's very possible there's just something weird on my computer, and for everyone else it works great!)

jdsimcoe:
When I remove the toggle my output devices display normally again.

gkoberger:
@jdsimcoe are you using this PR, or the extension in general? What version of macOS are you on? I'll try to help track down the issues for ya!

FezVrasta (approved):

jdsimcoe:
@gkoberger I'm using the extension in general, not this PR. Hopeful that it actually fixes things for me. I'm on macOS `13.1 Beta (22C5050e)`. Once it gets merged, I'm assuming I'll be able to update via Raycast and try the fix?

gkoberger:
@jdsimcoe Ah yes, this should fix it hopefully! Admittedly I don't know how Raycast extensions update behind the scenes, but I imagine it will.

(There's one caveat I mentioned before, which is that it opens the Settings modal, which is super annoying. I haven't been able to find a way around it, though. This only happens if Airplay is checked.)

gkoberger:
Hey @benvp just wanted to check in and see if you had any thoughts!

gkoberger:
Also fixes #3771

pernielsentikaer:
When I enable `Airplay` and opens Set Output Device is newer stops loading, is that intended?

benvp:
Just tried it out and it's not working for me on Ventura. When running the apple script version it's stuck at the "Appearance" system settings window. 

I'm sorry but I absolutely have no experience using AppleScript and the AppleScript based implementation was done by @FezVrasta.

Right now, I can't help with that.

benvp:
@gkoberger Based on your script, the following works for me, but will need some tweaking. It currently steals the focus of the raycast window and therefore hides raycast for me. When opening it up again, the audio devices appear in the list properly.

```
do shell script "open x-apple.systempreferences:com.apple.Sound-Settings.extension"

set devices to {}

tell application "System Events" to tell application process "System Settings"
	repeat until exists window "Sound"
		delay 0.3
	end repeat
	if value of radio button 1 of tab group 1 of group 2 of scroll area 1 of group 1 of group 2 of splitter group 1 of group 1 of window "Sound" is not 1 then
		click radio button 1 of tab group 1 of group 2 of scroll area 1 of group 1 of group 2 of splitter group 1 of group 1 of window "Sound"
		repeat until value of radio button 1 of tab group 1 of group 2 of scroll area 1 of group 1 of group 2 of splitter group 1 of group 1 of window "Sound" is 1
			delay 0.3
		end repeat
	end if
	
	tell table 1 of scroll area 1 of group 2 of scroll area 1 of group 1 of group 2 of splitter group 1 of group 1 of window "Sound"
		set selected_row to (first UI element whose selected is true)
		set currentOutput to value of static text of group 1 of UI element 1 of selected_row as text
	end tell
	
	set theRows to rows of table 1 of scroll area 1 of group 2 of scroll area 1 of group 1 of group 2 of splitter group 1 of group 1 of window "Sound"
	
	repeat with r in theRows
		try
			set deviceName to value of static text of group 1 of UI element 1 of r as text
			set deviceType to value of static text of group 1 of UI element 2 of r as text
			set end of devices to {deviceName, deviceType}
		end try
	end repeat
end tell

if application "System Settings" is running then
	tell application "System Settings" to quit
end if

return [devices, "currentOutput", currentOutput]
```

gkoberger:
Okay, I have another attempt, this time not using an AppleScript (since it feels so hacky). I use 3 different sources for getting audio output sources (the included binary, [airplayer](https://www.npmjs.com/package/airplayer) and `system_profiler SPBluetoothDataType`).

There's two things I don't love:

- [ ] The output still opens up the System menu... it's annoying but doesn't break the flow like the old version does
- [ ] I had to use a setTimeout to collect Airplay, which I don't like at all

I think I can live with the Sound settings menu still opening, because it only happens once you make a selection. I find the setTimeout to be super flaky (for example, sometimes 3 seconds isn't long enough and sometimes 1 second is long enough), I just don't know what other options we have here. (I imagine we could refactor it so that as the listener reports new devices, the react list could update?)

For a larger refactor, I would suggest:
- [ ] Combine `audio-device` and `utils` into one, because it's really confusing to have two similar functions that operate differently
- [ ] Merge the Bluetooth stuff into `audio-device.ts`, because it's able to pick up Bluetooth devices that are connected but not active (but it can't currently go there because it doesn't have a device.id)
- [ ] Fix it so that based on the source (the binary, Bluetooth or airplay) it knows how to switch the device (the binary can use itself, however bluetooth/airplay have to pop open the settings)

(Also, fixes #3942)

gkoberger:
I think these typescript errors might be where I tap out here :/ I might come back to this later, but it's working for me locally now and I'm not sure I'll be able to get TS definitions done for the two packages I pulled in.

pernielsentikaer:
I gave it another spin, and when I had AirPlay enabled, I just got `No Results`

Raycast-iioWg6yr@2x

When disabling it, I get results.

Raycast-nYEmO9IC@2x

Can I help with some debugging?

PR #3614, opened 2022-11-22
Add World Cup extension language support 

djpowers:
## Description

Added support for additional langauges by detecting locale. Links to FIFA initially always used "ES" language, and this change supports most of the ones listed on fifa.com (in the right-hand side of the footer). Tested with each language included in the PR.

~As a bonus it also uses the translation from the API when showing the group info.~

## Checklist

### DIFF
--- a/extensions/world-cup/CHANGELOG.md
+++ b/extensions/world-cup/CHANGELOG.md
@@ -2,4 +2,6 @@
 
 ## [Add support for additional languages] - 2022-11-21
 
+When using the Action See Match on FIFA.com it now detects if a language is supported by FIFA.com
+
 ## [Initial Version] - 2022-11-21


pernielsentikaer (approved):
Looks good to me, @ismamz do you want to verify?

ismamz:
@djpowers It looks great! We can also add locale for date string: 

image

djpowers:
> @djpowers It looks great! We can also add locale for date string:

Thanks @ismamz! And good suggestion. I pushed a change to pass the locale when using the date-fns translation function. I intentionally left the "finished" and "now" text unchanged as they would need to be manually translated, but I think this should suffice for now.
world-cup 2022-11-22 at 09 57 49

Also I suppressed a TypeScript error I was getting, but I'm not super familiar with TS so not sure if there's a better way of handling it.

pernielsentikaer (changes requested):
At the moment, Raycast doesn't support localization and only supports US English. Therefore, please avoid introducing your custom way to localize your extension.

The link is okay since it's not a part of the UI, could we remove that to not break the rules?

djpowers:
Thanks for the reminder @pernielsentikaer. I updated the PR so that no UI or language changes are made to the extension, and to only use the system locale to determine which language version the outgoing link should use.

ismamz (approved):

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #3637, opened 2022-11-23
[World Cup] Show live time for current match

ismamz:
## Description

Fetch current match live time (refresh every minute).

## Screencast

Metadata

## Checklist

### DIFF
--- a/extensions/world-cup/CHANGELOG.md
+++ b/extensions/world-cup/CHANGELOG.md
@@ -7,6 +7,7 @@
 - Added animated GIF for empty view
 - Fixed time for matches about to start
 - Added goals information
+- Details are hidden by default (use Enter key to toggle)
 
 ## [Enhancements] - 2022-11-23
 


pernielsentikaer:
Hi @ismamz 👋 

Now that we fetch the time, can we also fetch the score?

ismamz:
> Hi @ismamz 👋
> 
> Now that we fetch the time, can we also fetch the score?

Do you mean update the score every minute? Yes, we can.

We can add also the who did the goals and the exact minute, maybe for details section.

Don't you think it would be better to leave the details closed by default?

pernielsentikaer:
> Do you mean to update the score every minute? Yes, we can.

That would make sense, right? Would be weird only to update the time 😊

> We can also add who did the goals and the exact minute, maybe for the details section.

That's also a cool idea, a lot of ideas left like players, etc, we might need to push to a new view with all that data.

> Don't you think it would be better to leave the details closed by default?

Fine with me 😊

ismamz:
@pernielsentikaer i've pushed some changes:

- Show live time and **score** for current match
- Reversed the list for previous matches
- Fixed time for matches about to start
- Added goals information
- Added animated GIF for empty view

Metadata-3

Metadata-4

https://user-images.githubusercontent.com/3452011/203845301-655fdbf6-f925-4211-9f22-7dcf6f30ac9b.mov

peduarte:
🔥

peduarte (approved):
Let's gooo ⚽ ⚽ ⚽ ⚽

PR #3658, opened 2022-11-25
Drupal Toolbox Extension

emircanerkul:
## Description

New Extension: Drupal Toolbox extension

Valuable tools and shortcuts for your Drupal Websites. Manage your Drupal Websites with ease.

image

## Checklist

### DIFF
--- a/extensions/drupal-toolbox/package.json
+++ b/extensions/drupal-toolbox/package.json
@@ -5,6 +5,9 @@
   "description": "Valuable tools and shortcuts for your Drupal Websites",
   "icon": "logo.png",
   "author": "emircanerkul",
+  "contributors": [
+    "pernielsentikaer"
+  ],
   "categories": [
     "Developer Tools",
     "Web",


extensions/drupal-toolbox/src/tools.tsx
```diff
@@ -0,0 +1,116 @@
...
+    id: "open_in_php_storm",
+    title: "Open in PhpStorm",
+    action: (website: DrupalWebsite) =>
+      Tool(website, `phpstorm ${website.root}`, {
+        inprogress: "Preparing to Open in PhpStorm.",
+        completed: "The project opened in PhpStorm.",
+      }),
+    icon: Icon.CodeBlock,
+  },
+];
```
pernielsentikaer, line 68: Could we add [ActionPanel](https://developers.raycast.com/api-reference/user-interface/actions#api-reference) icons here

pernielsentikaer (changes requested):
Hi 👋 

I have now tested your extension and I have a suggestions ready for you

extensions/drupal-toolbox/src/tools.tsx
emircanerkul: Also, thank you for Destructive Style

pernielsentikaer:
Hi @emircanerkul 👋 

I tested it once more, and I don't see a reason to push to another view where only actions are possible, what do you think about doing this instead?

drupal-toolbox 2022-11-30 at 06 02 23

emircanerkul:
> Hi @emircanerkul 👋
> 
> I tested it once more, and I don't see a reason to push to another view where only actions are possible, what do you think about doing this instead?
> 
> drupal-toolbox 2022-11-30 at 06 02 23

To be able to filter, to be able to extend without thinking how many tools have in the list (I think it's not so much usable in that little action menu), and for future possibilities like nested tools.

pernielsentikaer:
Hi @emircanerkul 👋 

That sounds like a good idea to keep tools then, thanks for the clarification 🔥 

I edited a few things, I hope it's okay with you 😊

- Use useForm for adding the sites, this removed a lot of code
- Added `canChooseFiles={false}` to the picker, so only folders can be selected
- Updated API
- Small wording touches

As always, take a look and let me know, everything is up for debate 😊

extensions/drupal-toolbox/src/components/EmptyView.tsx
```diff
@@ -25,8 +25,8 @@ function EmptyView(props: {
   return (
     <List.EmptyView
       icon="drupal-toolbox.png"
-      title="No Drupal websites found"
-      description="You don't have any Drupal websites yet. Why not add some?"
+      title="No Sites Found"
+      description="Press ⌘+N to add a site"
```
emircanerkul, line 29: Drupal, drupal, drupal, druuuupal: https://youtu.be/lZ-s3DRZJKY?t=11

emircanerkul: But okay, not a big deal. I'll stop being drupal partisan :)

emircanerkul:
> Hi @emircanerkul 👋
> 
> That sounds like a good idea to keep tools then, thanks for the clarification 🔥
> 
> I edited a few things, I hope it's okay with you 😊
> 
> * Use useForm for adding the sites, this removed a lot of code
> * Added `canChooseFiles={false}` to the picker, so only folders can be selected
> * Updated API
> * Small wording touches
> 
> As always, take a look and let me know, everything is up for debate 😊

All is okay except for one point that is;

I was already placed canChooseDirectories={true} I'm not sure also canChooseFiles={false} is required. By default, it should not if I mention it as a directory, but not sure if it's a good design. Might be enum could be good for this case? type={Form.FilePicker.Type.File|Form.FilePicker.Type.Folder|Form.FilePicker.Type.Both}

Also good to have this validation section in the example todo app. I just based on this :)

pernielsentikaer:
> I was already placed canChooseDirectories={true} I'm not sure also canChooseFiles={false} is required. By default, it should not if I mention it as a directory, but not sure if it's a good design. Might be enum could be good for this case? type={Form.FilePicker.Type.File|Form.FilePicker.Type.Folder|Form.FilePicker.Type.Both}

If you don't set `canChooseFiles={false}` then you can both pick a file AND a folder since it defaults on

Yes, our new useForm is much better since it handles all the scenarios with validations and remembering data

extensions/drupal-toolbox/src/components/EmptyView.tsx
pernielsentikaer: It's just to shorten the text for the user and this way makes the UI/UX a bit easier. Damn, that's a catchy song 😊

emircanerkul: Happy to your mind :)

emircanerkul:
I think we are all done, right?

pernielsentikaer:
We should be 🚀

pernielsentikaer:
Thanks for your cooperation, it was nice working with you 🔥

PR #3660, opened 2022-11-25
New extension: Curator.bio

Yukaii:
## Description

Add the new extension for https://beta.curator.bio/

## Screencast

Metadata-1

Metadata-2

Metadata-3

## Checklist

### DIFF
--- a/extensions/curator-bio/CHANGELOG.md
+++ b/extensions/curator-bio/CHANGELOG.md
@@ -1,3 +1,3 @@
 # Curator Bio Changelog
 
-## [Initial Version] - 2022-11-25
+## [Initial Version] - 2022-12-05


pernielsentikaer (changes requested):
Hi 👋 

I have now tested your extension and I have some feedback ready for you:

* For all three commands, could you ensure that the loader stops when wrong credentials is entered 😊

Yukaii:
Hi @pernielsentikaer, I've pushed one commit to show message when error occurred. I also have another problem. 

Since curator.bio lacks a public API, I used the Raycast Cache API and `tough-cookie` to preserve login sessions rather than keeping access tokens in preference storage. However the login status of a session cannot be reset by changing the email or password because the cookie is persistent between requests. The possible solutions might be:

1. Completely remove session caching. Each command sends login request every time.
2. Provide a way to clear the cache when changing preferences. But I'm not sure how to achieve this.

Have you got any suggestions?

pernielsentikaer:
1) That would also be a shame since it will make it more responsive and snappy
2) I'm pretty sure we did something similar in [Liveblocks](https://www.raycast.com/andreaselia/liveblocks) - there we saved a hashed value of the password and compared it

andreaselia:
Just for future reference in case someone ends up looking at this issue for something.

The Liveblocks extension is being updated to simply use the secret key, so if you're looking for the previous implementation mentioned in this issue, you'll need to check a historic version of the extension.

PR #3667, opened 2022-11-26
feat: harbor extension

dlvhdr:
# Harbor 🚢

> See which ports are exposed by your devlopment processes and act on them.

Imagine you're a developer that needs to set up multiple devservers to run his app locally. E.g: a server that serves the client side app, and another for the API server etc.
With **Harbor** you can easily see which devservers are running right in your _Menu Bar_!

## Preferences

* `hideByArgs`: command separated string list.
  * Allows you to hide certain processes by their command line args.
  * Useful for long running processes that are not related to your devservers.

## Images

Raycast 2022-11-26 at 16 03 09
SCR-20221126-ma0
SCR-20221126-mbg

Note: uses `lsof` to query for open ports.
Repo: https://github.com/dlvhdr/harbor

Signed-off-by: Dolev Hadar <dolevc2@gmail.com>


pernielsentikaer:
Thanks for your contribution @dlvhdr 🔥 

Would it make sense to build into [Port Manager](https://www.raycast.com/lucaschultz/port-manager) instead, they are quite similar, it feels like a very good enhancement to that one.

dlvhdr:
@lucaschultz wdyt? The menu bar extension I wrote puts more emphasis on locally exposed `node` processes but can be extended to prioritize more.

lucaschultz:
I agree with @pernielsentikaer, while the two extensions each provide a different UI, they offer seemingly the same functionality and their potential user base is probably very similar. Integrating this in Port Manager would give users of the existing extension access two the Menu Bar command, instead of having two _discover_ and install another extension.

This might be an opportunity to polish PortManager up a bit, perhaps adding the possibility too emphasize `node` processes a bit more (I'm thinking `List.Section`) and unifying which information is shown.

PR #3693, opened 2022-11-28
Add new (more flexible) main command to Base64 Extension

sammarks:
## Description

Adds a new main command to the existing Base64 extension. I found the existing extension pretty inflexible when it comes to how to present the encoded / decoded data. Namely, the new index command supports the following:

- Preview the encoded / decoded content directly inside the list view.
- Preview the content you are encoding / decoding.
- Automatically read from clipboard, by default.
- Allow input override via the list search field (this is more seamless to me than using arguments).
- Three actions: copy to clipboard, paste in active app, open URL (if the decoded content is a URL).
- New preference: "default action" - allows users to configure which action gets executed when they press enter on the encode / decode list item.

**I left the existing commands in place for backwards-compatibility.**

**I also elected to improve the existing extension rather than submitting a new one.** I've seen a few other base64-related PRs pop up, only to be closed later because there are enough base64 extensions already in the store.

The main base64 extension has gone quite some time without some love with the new features available in the API, so I though I'd take a crack at it.

## Screencast

https://user-images.githubusercontent.com/424093/204354746-c6c6e7c2-abf6-43bb-b0c4-c4e986c8e2ae.mov

## Checklist


pernielsentikaer:
Hi @sammarks 👋 

Thanks for your contribution 💯 

@DanielSinclair can you check this?

sammarks:
@pernielsentikaer do you have a policy on what happens if the original extension author doesn't respond to requests for a certain period of time?

DanielSinclair:
Hi @sammarks. Looks fine to me! I hope Raycast can reevaluate how distribution/forking works; the namespace squatting and monorepo setup seems untenable as the platform has grown.

pernielsentikaer:
We are going to take some actions, it's a focus point for us to come up with a better solution 😊

PR #3754, opened 2022-12-01
[Linear] Added rightclick support for menubar

pernielsentikaer:
## Description

Added so it's possible to right-click on a menubar item to mark it as read

### DIFF
--- a/extensions/linear/CHANGELOG.md
+++ b/extensions/linear/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Linear Changelog
 
-## [Added right click support to menubar] - 2022-12-01
+## [Added right click support to menubar] - 2022-12-05
 
 - Added right click support to menubar which mark the issue as read.
 


extensions/linear/src/unread-notifications.tsx
```diff
@@ -19,10 +20,7 @@ const preferences = getPreferenceValues<{ alwaysShow: boolean }>();
 function UnreadNotifications() {
   const { isLoadingNotifications, unreadNotifications, urlKey, mutateNotifications } = useNotifications();
 
-  async function openNotification(notification: NotificationResult) {
-    const applications = await getApplications();
-    const linearApp = applications.find((app) => app.bundleId === "com.linear");
-    notification.issue ? await open(notification.issue.url, linearApp) : await openInbox();
+  async function markNotificationAsRead(notification: NotificationResult, showNotification = false) {
```
peduarte, line 23: nit: we dont really need `= false` here, since it's `false` by default already

extensions/linear/src/unread-notifications.tsx
```diff
@@ -35,6 +33,17 @@ function UnreadNotifications() {
       },
       shouldRevalidateAfter: true,
     });
+
+    if (showNotification) {
+      await showHUD("Notification marked as read");
```
peduarte, line 38: yeah, i'm not 100% if we need the toast here to be honest.

peduarte (commented):
Just a couple of comments, code wise just a nit. Im just thinking about the feature itself. I find that right click is not very discoverable/user friendly. But I can get behind how making as read from the menu is handy for power users :)

extensions/linear/src/unread-notifications.tsx
```diff
@@ -84,7 +88,13 @@ function UnreadNotifications() {
               title={baseTitle}
               subtitle={notification.issue?.title ? truncate(notification.issue.title, 20) : ""}
               tooltip={`${notification.issue?.identifier}: ${notification.issue?.title}`}
-              onAction={() => openNotification(notification)}
+              onAction={async (event: MenuBarExtra.ActionEvent) => {
+                if (event.type === "left-click") {
+                  openNotification(notification);
```
thomaspaulmann, line 93: Does it need an await here?

```suggestion
                  await openNotification(notification);
```

pernielsentikaer: I'm not sure, but it didn't work without for me 😕

pernielsentikaer:
Anymore feedback before we merge it?

PR #3765, opened 2022-12-02
[open-in-visual-studio-code] Also work with VSCodium

chohner:
## Description
Enables [VSCodium](https://github.com/VSCodium/vscodium) to open the current Finder folder

## Checklist

### DIFF
--- a/extensions/open-in-visual-studio-code/src/index.ts
+++ b/extensions/open-in-visual-studio-code/src/index.ts
@@ -1,5 +1,9 @@
-import { getApplications, getSelectedFinderItems, open, showToast, Toast } from "@raycast/api";
+import { getApplications, getPreferenceValues, getSelectedFinderItems, open, showToast, Toast } from "@raycast/api";
 import { exec } from "child_process";
+
+interface OpenVSCodePreferences {
+  VSCodeVariant: string;
+}
 
 /**
  * Gets the selected Finder window.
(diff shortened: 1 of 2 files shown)


extensions/open-in-visual-studio-code/src/index.ts
```diff
@@ -32,7 +32,9 @@ const getSelectedFinderWindow = (): Promise<string> => {
 
 export default async () => {
   const applications = await getApplications();
-  const visualStudioCode = applications.find((app) => app.bundleId === "com.microsoft.VSCode");
+  const visualStudioCode = applications.find(
+    (app) => app.bundleId === "com.microsoft.VSCode" || app.bundleId === "com.vscodium"
+  );
```
thomaspaulmann, line 37: This logic can have different results, depending on the order of the applications. Also, what happens to people that have both versions installed? 

I would suggest to do the following:
1. Search for VS Code and VS Codium separately
2. Combine the results with an optional ternary operator. 

Alternatively, you can order the applications by bundle identifier and then use `find` as you do. An improvement could be to add a preference to the extension as a dropdown to select which version of VS Code should be used. By default, you could have an "automatic" way that prefers VS Code over Codium.

thomaspaulmann (changes requested):
Thanks for the addition, this is great. I left a comment to improve the logic for detecting VSCodium.

chohner:
The last commit adds a dropdown menu to the preferences (which defaults to `VSCode`)

Screenshot 2022-12-05 at 18 38 35

Does this match what you expect @thomaspaulmann ?

thomaspaulmann:
Looking good! Thanks for the contribution 👏

PR #3796, opened 2022-12-05
Adding a new command to Folder-Search - Send Finder Item to Folder.

omarshahine:
When using the Folder Search command, it will now optionally take the selected Finder item(s) and allow you to move them to a selected folder

## Description

When using the Folder Search command, it will now optionally take the selected Finder item(s) and allow you to move them to a selected folder via a new Action menu

## Screencast
https://share.cleanshot.com/0anhh5

## Checklist

### DIFF
--- a/extensions/folder-search/CHANGELOG.md
+++ b/extensions/folder-search/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Folder Search Changelog
 
+## [New Features] - 2022-12-15
+
+- **added** Can take selected Finder item(s) and allow you to move them to a selected folder via a new Action menu entry
+
 ## [New Features] - 2022-11-01
 
 - **added** Sorting of results by `Last used` to bring most recently used (opened) to the top of repeated searches.


koinzhang:
Hi @omarshahine. One question, what happens if I move a larger file and exit the extension in the middle of moving the file?
This reminds me of the Common Directory extension, which was planned to also have the ability to move/copy files, but was rejected by Raycast's reviewers because it couldn't handle the above problem

omarshahine:
> Hi @omarshahine. One question, what happens if I move a larger file and exit the extension in the middle of moving the file?
> This reminds me of the [Common Directory](https://github.com/koinzhang/common-directory) extension, which was planned to also have the ability to move/copy files, but was rejected by Raycast's reviewers because it couldn't handle the above problem

Good point. How do other extensions handle this issue?

I have tried moving large files on disk, it was instant. I guess I could restrict moves to not happen across volumes to handle this restriction. I don't support copy.

koinzhang:
> I have tried moving large files on disk, it was instant.

What is the approximate size of the large file you are trying? I mean large files are probably of the tens of gigabytes variety, or even larger, at least that was the file size (35G) when Raycast reviewers ran the test earlier.

> I guess I could restrict moves to not happen across volumes to handle this restriction.

Limiting cross-volume moves doesn't help, as long as it's moving to other folders. 

>  I don't support copy.

Copy does not have the above issue because the original file is not moved, only when Move is done, the original file is moved and only then it may cause file loss.

As for the solution, I probably have two approaches:
1. Maybe consider limiting the size of the files moved? This was a previous idea, but a troublesome and strange solution.
2. Use applescript script to operate Finder to move files, I wonder if the operation will be interrupted when exiting the extension in this way?

The root cause of this problem is that Raycast does not have a mechanism for background tasks, and it does not guarantee that some background persistent tasks continue to execute when the extension is launched.

Maybe @pernielsentikaer  has experience in dealing with this issue?

pernielsentikaer:
Not that I have tried it yet, but what about testing the new option for [CloseMainWindow](https://developers.raycast.com/api-reference/window-and-search-bar#closemainwindow)

Raycast-qB76ELPP@2x

omarshahine:
> > I have tried moving large files on disk, it was instant.
> 
> What is the approximate size of the large file you are trying? I mean large files are probably of the tens of gigabytes variety, or even larger, at least that was the file size (35G) when Raycast reviewers ran the test earlier.
> 
> > I guess I could restrict moves to not happen across volumes to handle this restriction.
> 
> Limiting cross-volume moves doesn't help, as long as it's moving to other folders.
> 
> > I don't support copy.
> 
> Copy does not have the above issue because the original file is not moved, only when Move is done, the original file is moved and only then it may cause file loss.
> 
> As for the solution, I probably have two approaches:
> 
> 1. Maybe consider limiting the size of the files moved? This was a previous idea, but a troublesome and strange solution.
> 2. Use applescript script to operate Finder to move files, I wonder if the operation will be interrupted when exiting the extension in this way?
> 
> The root cause of this problem is that Raycast does not have a mechanism for background tasks, and it does not guarantee that some background persistent tasks continue to execute when the extension is launched.
> 
> Maybe @pernielsentikaer has experience in dealing with this issue?

I moved a 65GB File (it was a VM) and it moved in 2 seconds (APFS). So I don't actually know or have a way to test this issue.

pernielsentikaer:
I just moved a 94GB file from one directory to another; it shouldn't take more than a few seconds - I had `Pop to Root Search` to immediate

pernielsentikaer:
You might be able to use [getSelectedFinderItems](https://developers.raycast.com/api-reference/environment#getselectedfinderitems) instead 😊

omarshahine:
> You might be able to use [getSelectedFinderItems](https://developers.raycast.com/api-reference/environment#getselectedfinderitems) instead 😊

Oh geez I didn't even know about that. I have updated my PR to use this new built in function.

omarshahine:
> Not that I have tried it yet, but what about testing the new option for [CloseMainWindow](https://developers.raycast.com/api-reference/window-and-search-bar#closemainwindow)
> 
> Raycast-qB76ELPP@2x

What would I use this for?

The command already calls 

```
closeMainWindow();
popToRoot({ clearSearchBar: true });
```
When it finishes calling the move function.

Issue #3652, opened 2022-11-24
[Notes] Search Notes doesn't work


iamdannywyatt:
### Extension

https://www.raycast.com/tumtum/apple-notes

### Description

When I try to search for notes, I get this error:
![Screen Shot 2022-11-24 at 1 15 59 PM](https://user-images.githubusercontent.com/96640778/203843504-ec3183eb-36ff-46a5-a4f7-4fd0e5899245.png)

Here's the log:
[search notes log.txt](https://github.com/raycast/extensions/files/10086660/search.notes.log.txt)

ron-myers:
Thank you for reporting this @iamdannywyatt 

Does this happen with all notes or just some? 

From the log it looks like you are selecting a note titled `C` and selecting it for open, does that sound correct?  The error in the log is:

```
osascript -e tell application "Notes" 
show note "C" 
end tell
23:41: execution error: Notes got an error: AppleEvent handler failed. (-10000)
```

Any chance you have multiple notes titled `C`?

iamdannywyatt:
@ron-myers Thank you for the quick reply.
The issue is with any note, regardless of the name.
One thing though: now that you mentioned the "C note", that's not a note, it's a folder inside a folder called Passwords, but it's showing me as if it was a note, so that alone is already a bug.

I tried notes inside the default Notes folder (the one that can't be deleted), notes inside folders, notes inside folders that are inside folders. Everything shows me that error.

Even if I had multiple notes titles C (which is not the case anyway), it would show me both options and the moment I pick one, it should open that one, right? I assume each note has its own "id" or something like that, so I don't see that this could be the issue.
I tried with notes that have unique names and got the same error, so that's not the cause either.

iamdannywyatt:
@ron-myers 
I tried it again after rebooting, just in case something was going wrong, but no luck.
One thing I noticed and I don't know if this is a normal behavior, but Notes was closed completely, and as soon as I typed "Search Notes" and hit Enter, it opened Notes without me even picking one option. So it showed me some notes on the list for like a second, but then it immediately opened Notes.
Also, for a few minutes when I would open the list of notes, it would show me a note that I deleted a long time ago, but it was showing me at the top of the list as if it was inside the Recently Deleted folder. So weird...

ron-myers:
@iamdannywyatt  ok - I have found **an** issue - which may be your issue.  Do any of your folders have any non-alphanumeric characters?

[...]

PR #3806, opened 2022-12-06
[Apple Notes] #3652 Correct issue where searching for notes failed

ron-myers:
## Description

Added error handling for failing to open notes towards #3652

## Checklist

### DIFF
--- a/extensions/apple-notes/CHANGELOG.md
+++ b/extensions/apple-notes/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Apple Notes Changelog
 
-## [Fix] - 2022-12-06
+## [Fix] - 2022-12-11
 
 - Correct issue where searching for notes with quotes in titles failed
 


pernielsentikaer:
Do you mind testing this @tumispro

tumispro:
> Do you mind testing this @tumispro

Yeah I'll schedule some time tonight!

tumispro:
Not sure if this is supposed to work like this, but I tried to open a deleted note and got this:

Screenshot 2022-12-10 at 12 38 42

ron-myers:
Thanks for testing this @tumispro.  

I have a few thoughts on the deleted note opening and related to this extension:

1. this pull request is to correct escaping titles when opening around #3652 - this shouldn't change the behavior of deleted notes.  I created a note with that same name, searched for it, opened it, deleted it, searched for it, and then opened it successfully. My gut says that note is in your search cache but not available in your notes app - is that true?
2. we should add a filter to remove deleted notes from the search
3. we add a preference to keep them in the search for people that prefer them to be available.
4. we should make searching using SQL the default, add a preference for what source people want to use for their notes.

I prefer to merge this to see if we correct #3652, I will work on the above over the holidays.  How does that sound?

tumispro:
Good stuff. Seems to work fine now with folders containing special characters like "Themes & Names." as in the mentioned issue, so I think this can be merged now @pernielsentikaer.

As for your comment about the deleted note: I both deleted it and then emptied trash, and then tried to open it while the list wasn't yet refreshed. But I think this is an edge case that we don't really have to handle, as it is really rare and resolves itself right after. Still, I agree with your second point about being able to filter them out. 

I don't really have much time to work on it currently, but I'm glad to help test stuff that you build :)

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #3821, opened 2022-12-08
hypersonic: v2.0.1

tonimasc:
## Description

- Added an action that lets users share a task after creating it.
- Added an action that lets users to open a task after creating it.
- Changed toast feedback to hud feedback after a successful work sharing action.
- Fixed a bug that caused the extension to crash when the assigned user didn't exist in the workspace.
- Fixed a bug that caused the database fetch to crash when a database lacked a title property.

## Screencast
hypersonic 2022-12-08 at 13 00 30

## Checklist

### DIFF
--- a/extensions/hypersonic/CHANGELOG.md
+++ b/extensions/hypersonic/CHANGELOG.md
@@ -1,5 +1,18 @@
 # Hypersonic Changelog
 
+## [2.0.1] - 2022-12-08
+
+#### New
+
+- Added an action that lets users share a task after creating it.
+- Added an action that lets users to open a task after creating it.
+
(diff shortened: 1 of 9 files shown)


pernielsentikaer (approved):
Hi 👋 

I checked this update, and it works as expected; the only thing I felt a bit weird when adding a task is the delay from when using `Create Task and Copy URL` - it feels like a long time before the window closes; it could maybe be fixed with something like this.

Raycast-xb8Q2MTH@2x

It could be something to play with; catch me in [Slack](https://www.raycast.com/community) as always 🧑‍🚒

PR #3841, opened 2022-12-10
Updating the open-with-app to use the new getSelectedFinderItems command

omarshahine:
## Description

Updating the open-with-app to use the new [getSelectedFinderItems](https://developers.raycast.com/api-reference/environment) command

## Screencast

n/a

## Checklist

### DIFF
--- a/extensions/open-with-app/src/index.tsx
+++ b/extensions/open-with-app/src/index.tsx
@@ -51,7 +51,7 @@ function ApplicationsListItem(props: { application: Application }) {
                 await closeMainWindow({ clearRootSearch: true, popToRootType: PopToRootType.Suspended });
                 if (selectedItems.length != 0) {
                   for (let i = 0; i < selectedItems.length; i++) {
-                    await open(selectedItems[i].path.replace(/"/g, '\\"'), application.bundleId);
+                    await open(selectedItems[i].path, application.bundleId);
                   }
                 } else {
                   await showHUD("⚠️  No Finder selection to open.");


fturcheti:
Thanks for doing this long overdue update, @omarshahine.

pernielsentikaer:
Could you both try it out 😊

fturcheti:
@pernielsentikaer With these changes I'm not able to open multiple files at once. Previously, for example, I could select multiple images, invoke the command, and open all of them with a specific application. After the proposed changes, when I do so, only one image gets opened.

fturcheti:
@pernielsentikaer Actually, even the old implementation is not working correctly. It can open multiple files, but it is not reliable. If I have, for example, 5 files selected, it is opening only 2 or 3.

omarshahine:
> Actually, even the old implementation is not working correctly. It can open multiple files, but it is not reliable. If I have, for example, 5 files selected, it is opening only 2 or 3.
> 
> Although I don't use the extension to open multiple files quite often, I think it used to work properly.
> 
> @pernielsentikaer Do you know how we can make this work for multiple files?

Odd, so this close block does not work deterministically?

`onAction={async () => {
              try {
                const selectedItems = await getSelectedFinderItems();
                if (selectedItems.length != 0) {
                  selectedItems.forEach((item) => {
                    open(item.path.replace(/"/g, '\\"'), application.bundleId);
                  });
                } else {
                  await showHUD("⚠️  No Finder selection to open.");
                }
              } catch (error) {
                await showHUD("⚠️  No Finder selection to open.");
              }
              closeMainWindow();
              popToRoot({ clearSearchBar: true });
            }}`

omarshahine:
@fturcheti I just tested with many files, jpgs etc and it worked just fine

pernielsentikaer:
Hi @fturcheti 👋 

Could you give it another spin, should be fixed now

extensions/open-with-app/src/index.tsx
```diff
@@ -47,10 +48,11 @@ function ApplicationsListItem(props: { application: Application }) {
             onAction={async () => {
               try {
                 const selectedItems = await getSelectedFinderItems();
+                await closeMainWindow({ clearRootSearch: true, popToRootType: PopToRootType.Suspended });
                 if (selectedItems.length != 0) {
-                  selectedItems.forEach((item) => {
-                    open(item.path.replace(/"/g, '\\"'), application.bundleId);
-                  });
+                  for (let i = 0; i < selectedItems.length; i++) {
+                    await open(selectedItems[i].path.replace(/"/g, '\\"'), application.bundleId);
```
fturcheti, line 54: We should remove these last leftovers from the old code. This call could be simplified to:

```js
await open(selectedItems[i].path, application);
```

fturcheti:
Thanks @pernielsentikaer! I was `await`ing good news from you!

pernielsentikaer:
Like this @fturcheti, or?

fturcheti:
Yep. LGTM, @pernielsentikaer!

pernielsentikaer (approved):
Let's go 🔥

PR #3867, opened 2022-12-13
[AWS]: Add Secrets Manager

crisboarna:
## Description
This introduces a new command which covers another common use case for developers using AWS: verifying/copying secrets variables.

- Add "Secrets Manager" to find, view, and copy a secret value

## Screencast

image
image
image
image

## Checklist

### DIFF
--- a/extensions/amazon-aws/src/secrets.tsx
+++ b/extensions/amazon-aws/src/secrets.tsx
@@ -68,14 +68,14 @@ function Secret({
               <List.Item.Detail.Metadata.Label title="Creation Date" text={secret.CreatedDate?.toLocaleString()} />
               <List.Item.Detail.Metadata.Label title="Last Accessed" text={secret.LastAccessedDate?.toLocaleString()} />
               <List.Item.Detail.Metadata.Separator />
-              <List.Item.Detail.Metadata.Label title="Rotation" text={`${secret.RotationEnabled ? "true" : "false"}`} />
+              <List.Item.Detail.Metadata.Label title="Rotation" text={secret.RotationEnabled ? "true" : "false"} />
             </List.Item.Detail.Metadata>
           }
         />
       }
       actions={
         <ActionPanel>
-          <Action.SubmitForm title={showValue ? "Hide Value" : "Show Value"} onSubmit={() => setShowValue()} />
+          <Action title={showValue ? "Hide Value" : "Show Value"} onAction={() => setShowValue()} />
           <Action.CopyToClipboard title="Copy Value" content={secretDetails?.SecretString || ""} />
           <Action.CopyToClipboard title="Copy ARN" content={secret.ARN || ""} />
           <Action.OpenInBrowser


extensions/amazon-aws/src/secrets.tsx
```diff
@@ -0,0 +1,125 @@
...
+              <List.Item.Detail.Metadata.Label title="Last Accessed" text={secret.LastAccessedDate?.toLocaleString()} />
+              <List.Item.Detail.Metadata.Separator />
+              <List.Item.Detail.Metadata.Label title="Rotation" text={`${secret.RotationEnabled ? "true" : "false"}`} />
+            </List.Item.Detail.Metadata>
+          }
+        />
+      }
+      actions={
+        <ActionPanel>
+          <Action.SubmitForm title={showValue ? "Hide Value" : "Show Value"} onSubmit={() => setShowValue()} />
```
JonathanWbn, line 78: I'm aware other commands have used the `SubmitForm` action for such a use case, but a simple [Action](https://developers.raycast.com/api-reference/user-interface/actions#action) is probably a better choice.

```suggestion
          <Action title={showValue ? "Hide Value" : "Show Value"} onSubmit={() => setShowValue()} />
```

JonathanWbn (changes requested):
Awesome new command, great job! 🎉  I really like the approach you chose with the `Detail` panel. 👏 

I have two minor tweaks to suggest.

extensions/amazon-aws/src/secrets.tsx
```diff
@@ -0,0 +1,125 @@
...
+          markdown={secretDetails?.SecretString || secretDetails?.SecretBinary?.toString() || ""}
+          metadata={
+            <List.Item.Detail.Metadata>
+              <List.Item.Detail.Metadata.Label title="Description" text={secret.Description} />
+              <List.Item.Detail.Metadata.Label title="ARN" text={secret.ARN} />
+              <List.Item.Detail.Metadata.Separator />
+              <List.Item.Detail.Metadata.Label title="Creation Date" text={secret.CreatedDate?.toLocaleString()} />
+              <List.Item.Detail.Metadata.Label title="Last Accessed" text={secret.LastAccessedDate?.toLocaleString()} />
+              <List.Item.Detail.Metadata.Separator />
+              <List.Item.Detail.Metadata.Label title="Rotation" text={`${secret.RotationEnabled ? "true" : "false"}`} />
```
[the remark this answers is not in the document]
crisboarna: Fixed

extensions/amazon-aws/src/secrets.tsx
crisboarna: Fixed

crisboarna:
Hi @JonathanWbn,

Could you have a look once more if it is good to go ?

JonathanWbn:
I am wondering if using Raycast's built-in filtering is actually the better choice for the secrets command. I introduced server-side filtering for parameters and log groups because there might be thousands or even tens of thousands of them. I'm not too familiar with the secrets manager, but I would assume there are rarely thousands of secrets per profile. If this is the case, I would suggest removing the search approach for a more straight-forward implementation and better UX. What do you think, @crisboarna?

crisboarna:
Hi @JonathanWbn,

I think that in most scenarios server side filtering is the best way to go as it 

1. Increases speed of commands by reducing payload network traffic
2. Reduces amount of work needed to be done on front-end/raycast
3. Reduces memory footprint overall

This is valid if there are 10 or 1000 secrets. 

I personally have profiles with several hundred secrets for larger projects and Secrets Manager supports up to 500,000 secrets per region so there is the potential to have a lot of secrets and it is best to code for worst case scenarion.

TL;DR: I think server side rendering is best way to go for this as well :)

JonathanWbn (approved):

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #3879, opened 2022-12-14
feat(arc): add command for creating new window

### LINKED ISSUE #3869: [Arc] Create new Arc window
### Extension

https://www.raycast.com/thomas/arc

### Description

It would be nice to create a new Arc window from Raycast. (like in the video below)

Usecase? opening a new window does so in the current space. 

The current way is either
to select an Arc window, press CMD+N and move that to the desired window, or use the mouse. Both are not great.

https://user-images.githubusercontent.com/4562878/207278055-5a1e75c6-2d75-4679-9cb5-b2db189bec66.mp4

_No response_

### Anything else?

_No response_

zach-fuller:
## Description

- Adds command for opening a new Arc window
- Updates existing search history command to support opening in a new window

## Screencast

https://user-images.githubusercontent.com/16906332/207503469-0abbae57-0b20-4aa4-8293-c9805227a6b3.mp4

## Checklist

Closes #3869

### DIFF
--- a/extensions/arc/package.json
+++ b/extensions/arc/package.json
@@ -24,7 +24,7 @@
     },
     {
       "name": "new-window",
-      "title": "New Window",
+      "title": "Open New Window",
       "description": "Creates a new window",
       "mode": "no-view",
       "subtitle": "Arc",
--- a/extensions/arc/src/new-window.tsx
+++ b/extensions/arc/src/new-window.tsx
@@ -4,7 +4,6 @@ import { createNewArcWindow } from "./utils";
 export default async function Command() {
   try {
     await createNewArcWindow();
-    await showHUD("✅ Window created");
   } catch {
     await showHUD("❌ Couldn't create window");
   }


extensions/arc/package.json
```diff
@@ -11,19 +11,30 @@
...
     {
       "name": "search-history",
       "title": "Search History",
       "subtitle": "Arc",
       "description": "Search your browsing history.",
       "mode": "view"
+    },
+    {
+      "name": "new-window",
+      "title": "New Window",
```
thomaspaulmann, line 27: Let's add a verb here to fit the naming. Maybe "Open" or "Create"?
```suggestion
      "title": "Open New Window",
```

extensions/arc/src/new-window.tsx
```diff
@@ -0,0 +1,11 @@
+import { showHUD } from "@raycast/api";
+import { createNewArcWindow } from "./utils";
+
+export default async function Command() {
+  try {
+    await createNewArcWindow();
+    await showHUD("✅ Window created");
+  } catch {
+    await showHUD("❌ Couldn't create window");
+  }
```
thomaspaulmann, line 10: Let's not show the HUD in the success case. The newly created window should be enough of a visual feedback.
```suggestion
  try {
    await createNewArcWindow();
  } catch {
    await showHUD("❌ Couldn't create a new window");
  }
```

extensions/arc/src/utils.ts
```diff
@@ -91,3 +91,12 @@ export class PermissionError extends Error {
 export const isPermissionError = (error: unknown) => {
   return error instanceof Error && error.name === "PermissionError";
 };
+
+export async function createNewArcWindow() {
+  await runAppleScript(`
+    tell application "Arc"
+      make new window
+      activate
+    end tell
```
thomaspaulmann, line 100: Can we somehow set the URL of the new tab here instead of using the `exec` command to open Arc. Later feels a bit like a hack.

thomaspaulmann (changes requested):
Thanks for adding this, it's great! I left a few minor comments.

extensions/arc/src/utils.ts
zach-fuller: I tried a few variations of that approach originally but couldn't get the command to work reliably. However, I have limited AppleScript experience and I'm open to suggestions. I agree the `exec` approach isn't ideal 😕

extensions/arc/src/new-window.tsx
zach-fuller: Good point, updated!

extensions/arc/src/utils.ts
crisboarna: Hey @thomaspaulmann ,

I have been struggling with the same issue regarding `AppleScript` opening a new Arc tab. Using the default `Chromium` `AppleScript` integration for the `tab` dictionary` entry results in the browser crashing when attempting to open the new tab, which is what I imagine @zach-fuller is referring to, akin to attempting to do [this](https://github.com/raycast/extensions/blob/b3f7177eb0875c538cd6ef18737984e3ffb8afe0/extensions/browser-history/src/actions/index.ts#L39). I think it is crashing as they are attempting to do some advanced tab handling with their `Spaces` concept which they have not covered for the `AppleScript` path perhaps.

In any case, I have also reached the same conclusion, albeit not ideal, to open a new `Arc` tab, although my approach still leverages using `AppleScript` for it (which adds a layer of (redirection)[https://github.com/raycast/extensions/pull/3918/commits/30310cf45dc66b48109b230721ee768e0f9ceec4]), in hopes of being able to quickly move it back once this issue is fixed in Arc :).

thomaspaulmann: Well okay, if Arc crashes and this works, we should keep it. Thanks for the collaboration here. I'll reach out to the Arc team to see if they can fix the AppleScript on their side.

thomaspaulmann:
Thanks for the addition!

extensions/arc/src/utils.ts
hurshagrawal: Sorry about this crash folks! Have some changes to fix this crash, and provide a richer AppleScript interface into Arc so you don't have to do the `exec` gnarliness. Aiming to get that into the first release of 2023!

PR #3887, opened 2022-12-15
feat(readwise-reader): Add new extension for Readwise Reader

zach-fuller:
## Description

Adds a new extension for [Readwise Reader](https://readwise.io/read) now that it is in public beta. This initial integration supports opening the main pages in Reader. Additional functionality will be added as Reader's API is expanded.

## Screenshot

Raycast 2022-12-14 at 21 19 05

## Checklist


pernielsentikaer:
Hi @zach-fuller 🔥 

Did you see [this extension](https://www.raycast.com/natterstefan/readwise)? Does it look like it's doing the same as yours?

zach-fuller:
> Hi @zach-fuller 🔥 
> 
> 
> 
> Did you see [this extension](https://www.raycast.com/natterstefan/readwise)? Does it look like it's doing the same as yours?

👋 Thanks for taking a look at this. They're actually different products from the same company. They use different APIs and serve different purposes but the naming is certainly confusing.

pernielsentikaer:
And it doesn't make sense only to have one, you think? Do you have any input here @natterstefan

zach-fuller:
I think it makes sense to have separate extensions since they are different products. One extension can potentially confuse users. For example, there is a concept of viewing your "Library" in Readwise and a similar option of viewing your "Library" in Reader. Any command for opening the Library would be ambiguous without resorting to "Open Library in Reader" or something similar. Additionally, Readwise has announced that they intend to price the products differently. There is certainly overlap at the moment, but that won't be the case long term.

iagomelanias:
Although I agree they must be separate extensions (_for now_), it's totally possible both products may have a bigger overlap in the long term.

**Pricing** - they haven't figured it out how to price it yet:
> [...] Once Reader officially exits beta sometime in 2023, we intend to reprice Readwise/Reader for new subscribers thereafter. Pricing is really hard and complex so we candidly haven't figured out the exact plans yet. [...]

**Products** - they're different products, but they mention the vision to unify them:
> [...] Over time, the vision is to create a unified experience between Readwise and Reader within Reader. [...]

[From their frequently asked questions section on the official website](https://readwise.io/read).

zach-fuller:
Right, technically the plan is for Reader to replace Readwise long term

pernielsentikaer:
What do you think @natterstefan - you're the author of the current

pernielsentikaer:
Pinging a last time @natterstefan

pernielsentikaer:
Did you have time to check @natterstefan - if it makes sense it's fine by me having two if the purpose is not the same / other product

natterstefan:
Hi pernielsentikaer,
hi @all.

Oh, time flies. Task managers only help if you pay attention to the reminder... Thank you all for your patience. So, let me think.

There are, as reported above, two scenarios:

1. both products remain, and Readwise (the company) does not merge them.
2. Readwise will merge both and operate under a common name (e.g., Reader) only.

Now we can either offer both products under one extension (i.e., the extension represents the company that provides two products) or we will separate them. Both cases have pros and cons.

### One extension case:

- Pros
	- Should Readwise (company) integrate Readwise into Reader, we update the extension, and users immediately have access to their highlights, etc., again without noticing that Readwise was merged into Reader.
	- The user can easily move between Reader and Readwise in the same extension without exiting it first (e.g. show me all highlights of a Reader article in Readwise would present the same interface as when the user opens a highlight directly in Readwise) which provides one unified user experience
	- shared codebase which is maybe beneficial
- Cons
	- We probably add toggles in the settings ("use Reader", "use Readwise") if people only pay for one service. At the moment you get both at one price. This could add complexity if they never merge the two products.

### Two extension cases:
- Pros
	- Separation of concerns
	- Flexibility
- Cons
	- Duplicate content if e.g. one wants to interact with highlights listed in Reader (which eventually also is present in Readwise)
	- What happens when Readwise (Company) phases out the Readwise product and integrates it into Reader? Users who only use the Readwise extension would have to install a new extension and we'd sunset the other.
	- Two extensions to maintain
	
Please, keep in mind that this list is incomplete for sure, I only listed the few that came up in my mind right now before having lunch.

> Although I agree they must be separate extensions (for now), it's totally possible both products may have a bigger overlap in the long term.

That's what I'm thinking. Currently, they are two separate products, but we cannot anticipate which one will remain in the long term. Even Readwise does not know (yet, and as far as I know) what to do and how.

> I think it makes sense to have separate extensions since they are different products. One extension can potentially confuse users. For example, there is a concept of viewing your "Library" in Readwise and a similar option of viewing your "Library" in Reader. Any command for opening the Library would be ambiguous without resorting to "Open Library in Reader" or something similar.

I understand the concerns, but even if two extensions exist and both use the same names ("Open Library"), the icon is required to help to differentiate the two products or a suffix ("in Reader" and "in Readwise"). I do not think that having two extensions will ever solve this if the two products use similar language. They even use similar icons. There is potential for confusion, but I do not think that's simply solvable with two extensions in this specific case.

> Additionally, Readwise has announced that they intend to price the products differently. There is certainly overlap at the moment, but that won't be the case long term.

Yes, and that's something we do not know, and it comes down to the question I ask myself whether we want to anticipate this will eventually happen and use only one extension already or not (and have two extensions).

---

tl;dr: In my opinion (but I leave the final decisions up to you @pernielsentikaer), I would provide only one extension anticipating that Readwise (company) will eventually merge the two products into one. This seems to be something they are considering, and that's mentioned in this thread here as well.

Both cases are valid thugh so yeah, difficult decision @pernielsentikaer :)

zach-fuller:
There's a valid case to be made for either option. I prefer separate extensions primarily for the separation of concerns. I think the decision to release Reader as a separate product was intentional. In my opinion, the extensions should reflect that decision. 

I think there is less overlap between the products than it appears at first glance.  While there is a degree of overlap (primarily in regards to highlighting) the primary use case for Readwise is quite different than the use case for Reader. 

The majority of Readwise users don't necessarily use Reader. As a user of Readwise + any other read-it-later app, I would prefer to not have to navigate around Reader functionality while interacting with the Readwise extension. As a user of Reader, I would prefer to have an experience that is focused on Reader while being able to use a dedicated extension to interact with Readwise.

@pernielsentikaer I'm on board with whatever you feel is best here 👍

pernielsentikaer:
I think we need two extensions as it is right now since it's two products, thanks for all the input 👏

It's always a hard decision but I think we need it that way

pernielsentikaer (approved):

PR #3893, opened 2022-12-15
feat(aws): added ecs/lambda resource drill-down for quick debugging

crisboarna:
## Description
For those that have to deal with multiple SAML redirect flows across multiple tab containers to be able to reach the desired resources, the default 'Open in Browser' is not enough, especially when one has 20+ accounts to switch between. For the quick debugging and researching of status of various AWS entities of interest, this PR adds the ability to quickly view details from within the app.

- Add ECS Services listing & details
- Add ECS Service Tasks listing & details
- Add ECS Service Task Containers listing & details
- Add ECS Service Task Container logs viewing
- Add Lambda Log Streams listing
- Add Lambda Log viewing
- Add AWS icons to all AWS entities being listed across all actions as opposed to the generic Raycast inbuilt icons.

## Screencast

Untitled

Also added logs filtering based on predefined search bar accessory dropdown list of options
2022-12-16 at 19 00 00@2x

## Checklist

### DIFF
--- a/extensions/amazon-aws/src/components/ecs/ECSClusterServiceTaskContainers.tsx
+++ b/extensions/amazon-aws/src/components/ecs/ECSClusterServiceTaskContainers.tsx
@@ -67,7 +67,6 @@ function ECSClusterServiceTaskContainers({ taskDefinitionArn }: { taskDefinition
               <ActionPanel>
                 {container.logConfiguration?.logDriver === "awslogs" && (
                   <Action.Push
-                    key={"view"}
                     title={"View Services"}
                     icon={Icon.Eye}
                     target={
--- a/extensions/amazon-aws/src/components/ecs/ECSCluster.tsx
+++ b/extensions/amazon-aws/src/components/ecs/ECSCluster.tsx
@@ -35,7 +35,6 @@ function ECSCluster({ cluster }: { cluster: Cluster }) {
       actions={
         <ActionPanel>
           <Action.Push
-            key={"view"}
             title={"View Services"}
             icon={Icon.Eye}
             target={<ECSClusterServices clusterArn={cluster.clusterArn || ""}></ECSClusterServices>}
--- a/extensions/amazon-aws/src/components/ecs/ECSClusterServiceTasks.tsx
+++ b/extensions/amazon-aws/src/components/ecs/ECSClusterServiceTasks.tsx
@@ -47,7 +47,6 @@ function ECSClusterServiceTasks({ service }: { service: Service }) {
             actions={
               <ActionPanel>
                 <Action.Push
-                  key={"view"}
                   title={"View Containers"}
                   icon={Icon.Eye}
                   target={
--- a/extensions/amazon-aws/src/components/ecs/ECSClusterServices.tsx
+++ b/extensions/amazon-aws/src/components/ecs/ECSClusterServices.tsx
@@ -60,7 +60,6 @@ function ECSClusterServices({ clusterArn }: { clusterArn: string }) {
             actions={
               <ActionPanel>
                 <Action.Push
-                  key={"view"}
                   title={"View Tasks"}
                   icon={Icon.Eye}
                   target={<ECSClusterServiceTasks service={service}></ECSClusterServiceTasks>}


pernielsentikaer:
Hi @crisboarna 🔥 

Thanks for your contribution 💯 

I might need some help with the check from some of the developers:  @victor-falcon @Hodglim @JonathanWbn @gebeto @momme-rtf @duboiss @hexpl0it

Could you check it out 😊

extensions/amazon-aws/src/components/searchbar/CloudwatchLogsTimeDropdown.tsx
```diff
@@ -0,0 +1,26 @@
+import { LogStartTimes } from "../../interfaces";
+import { List } from "@raycast/api";
+import { useEffect, useState } from "react";
+
+function CloudwatchLogsTimeDropdown({ onChange }: { onChange(logStartTime: LogStartTimes): void }) {
+  const [logStartTime, setLogStartTime] = useState<LogStartTimes>(LogStartTimes.OneHour);
+
+  useEffect(() => {
+    onChange(logStartTime);
+  }, [logStartTime]);
```
JonathanWbn, line 10: I suggest making this component fully controlled, since the parent is already keeping the `logStartTime` value.
```suggestion
function CloudwatchLogsTimeDropdown({ value, onChange }: { value: LogStartTimes; onChange(logStartTime: LogStartTimes): void }) {
```

extensions/amazon-aws/src/ecs.tsx
```diff
@@ -1,62 +1,35 @@
...
     <List
-      isLoading={isLoading}
-      searchBarPlaceholder="Filter clusters by name..."
+      isLoading={isLoading || isServicesLoading}
+      searchBarPlaceholder={getFilterPlaceholder("clusters")}
       searchBarAccessory={<AWSProfileDropdown onProfileSelected={revalidate} />}
     >
-      {error ? (
-        <List.EmptyView title={error.name} description={error.message} icon={Icon.Warning} />
+      {clusters ? (
```
JonathanWbn, line 28: This will not work when the returned value is an empty array (which happens in my case).
```suggestion
      {clusters && clusters.length > 0 ? (
```

extensions/amazon-aws/src/ecs.tsx
```diff
@@ -1,62 +1,35 @@
...
-import AWSProfileDropdown, { AWS_URL_BASE } from "./aws-profile-dropdown";
+import { fetchClusters } from "./actions";
+import AWSProfileDropdown from "./components/searchbar/aws-profile-dropdown";
+import { getFilterPlaceholder } from "./util";
+import ECSCluster from "./components/ecs/ECSCluster";
+import { useState } from "react";
 
 export default function ECS() {
-  const { data: clusters, error, isLoading, revalidate } = useCachedPromise(fetchClusters);
+  const [isServicesLoading, setServicesLoading] = useState<boolean>(false);
```
JonathanWbn, line 10: I don't think it's necessary to keep two loading states in this component. I would argue the `isLoading` from our `fetchClusters` is sufficient. The loading of each cluster should be handled by the cluster list item itself (maybe a loading indicator as the `icon`).

Furthermore, the `setServicesLoading` is passed inside a `.map` which means it will be called by each cluster and will lead to those clusters overwriting each other.

JonathanWbn (changes requested):
I love this, especially the new quick-access features! Well done 👏 

I have requested a few minor changes, but I have one more **general suggestion**:
I understand that the `defaultAction` preference increases optionality and has better backward compatibility. **However**, I think the "view details" drill down can reasonably be expected to be the desired behavior for most users. With Raycast's default shortcuts it'll also remain easy to open the selected resource in the browser with `⌘ ↩`.

So **I suggest removing the preference and making "View Details" the primary action**. This will also simplify the codebase and thereby improve maintainability (e.g. functions like `getActions` and `getActionPush` are no longer needed).

extensions/amazon-aws/src/ecs.tsx
```diff
@@ -1,62 +1,35 @@
...
+  const {
+    data: clusters,
+    error,
+    isLoading,
+    revalidate,
+  } = useCachedPromise(fetchClusters, [], { keepPreviousData: true });
+
+  if (error) {
+    return <List.EmptyView title={error.name} description={error.message} icon={Icon.Warning} />;
+  }
```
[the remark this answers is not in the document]
crisboarna: Good catch.

extensions/amazon-aws/src/components/searchbar/CloudwatchLogsTimeDropdown.tsx
crisboarna: Good point, it went through several iterations / cleanups to reach that point so it evaded me.

extensions/amazon-aws/src/ecs.tsx
crisboarna: I agree, good point, replaced each with loading icon if not complete as a replacement.

crisboarna:
> I understand that the `defaultAction` preference increases optionality and has better backward compatibility. **However**, I think the "view details" drill down can reasonably be expected to be the desired behavior for most users. With Raycast's default shortcuts it'll also remain easy to open the selected resource in the browser with `⌘ ↩`.
> 
> So **I suggest removing the preference and making "View Details" the primary action**. This will also simplify the codebase and thereby improve maintainability (e.g. functions like `getActions` and `getActionPush` are no longer needed).

Hi, on reviewing the possible flows for this, I agree it may not have a valid use case to warrant the added code lines for the `getActions` function in most cases, the utility functions contained within are useful as it removes code duplication from all classes into one single point, facilitating refactoring, i.e.: quickly change the title of `Open In Browser` while also reducing code comprehension difficulty.

JonathanWbn:
> the utility functions contained within are useful as it removes code duplication from all classes into one single point, facilitating refactoring, i.e.: quickly changing the title of Open In Browser while also reducing code comprehension difficulty for what is essentially duplicate code across all classes using that action.

Hi @crisboarna,
I can understand the arguments for keeping the `getActionOpenInBrowser` function, but I would still push back against using `getActionPush`:
- the code duplication is minimal (only the `icon` prop would be repeated)
- type-safety is lost with `[key: string]: any;`
- _opinion_ in-place JSX is easier to comprehend for a first-time reader, especially when they are new to React.

extensions/amazon-aws/src/components/ecs/ECSClusterServiceTaskContainers.tsx
```diff
@@ -0,0 +1,109 @@
...
+      logGroupStreamPrefix: container.logConfiguration?.options
+        ? container.logConfiguration?.options["awslogs-stream-prefix"]
+        : "",
+    });
+    const actionViewInBrowser = getActionOpenInBrowser(getTaskContainerUrl(taskDefinitionArn));
+
+    let containerActions = [actionViewInApp, actionViewInBrowser];
+
+    if (container.logConfiguration?.logDriver !== "awslogs") {
+      containerActions = containerActions.filter((action) => action.key !== "logs");
```
JonathanWbn, line 32: I think this can be really difficult for someone to understand what is going on here. Relying on `key` for conditions seems quite fragile. I'm not even sure if this even works as I can't find `"logs"` as a key anywhere. I think in-place conditional rendering will work fine here:
In-place conditions will have better readability IMO:
```jsx
<ActionPanel>
  {container.logConfiguration?.logDriver === "awslogs" &&
    getActionPush({
      title: "View Logs",
      component: CloudwatchLogs,
      logGroupName: container.logConfiguration?.options ? container.logConfiguration?.options["awslogs-group"] : "",
      startTime: logStartTime,
      logGroupStreamPrefix: container.logConfiguration?.options
        ? container.logConfiguration?.options["awslogs-stream-prefix"]
        : "",
    })}
   {getActionOpenInBrowser(getTaskContainerUrl(taskDefinitionArn))}
```

extensions/amazon-aws/src/actions/logs.ts
```diff
@@ -0,0 +1,99 @@
...
+  );
+
+  const combinedEvents = [...(accEvents || []), ...(logStreams || [])];
+
+  if (nextToken) {
+    return fetchAllLogStreams(logGroupName, nextToken, combinedEvents);
+  }
+
+  return combinedEvents;
+}
```
JonathanWbn, line 39: These functions could be combined. In the current setup `fetchLogStreams` is not really doing anything meaningful.

extensions/amazon-aws/src/actions/logs.ts
```diff
@@ -0,0 +1,99 @@
...
+  if (!process.env.AWS_PROFILE) return [];
+  return await fetchAllLogStreams(logGroupName);
+}
+
+async function fetchAllLogStreams(logGroupName: string, token?: string, accEvents?: LogStream[]): Promise<LogStream[]> {
+  const { logStreams, nextToken } = await cloudWatchLogsClient.send(
+    new DescribeLogStreamsCommand({ logGroupName, nextToken: token })
+  );
+
+  const combinedEvents = [...(accEvents || []), ...(logStreams || [])];
```
JonathanWbn, line 32: I'm running into a "heap out of memory" error when listing log streams. That makes sense to me as there can be tens of thousands of log streams for a given log group. I suggest capping this at e.g. 1000:
```js
if (combinedEvents.length > 1000) return combinedEvents;
```

extensions/amazon-aws/src/components/cloudwatch/CloudwatchLogStreams.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+function CloudwatchLogStreams({ logGroupName }: { logGroupName: string }) {
+  const [logStartTime, setLogStartTime] = useState<LogStartTimes>(LogStartTimes.OneHour);
+  const { data: streams, isLoading } = useCachedPromise(fetchLogStreams, [logGroupName], { keepPreviousData: true });
+
+  return (
+    <List
+      isLoading={isLoading}
+      searchBarPlaceholder={getFilterPlaceholder("stream")}
+      searchBarAccessory={<CloudwatchLogsTimeDropdown logStartTime={logStartTime} onChange={setLogStartTime} />}
```
JonathanWbn, line 18: Could you explain me the use case of this time dropdown?
I think a user might expect this to filter the log streams list, but instead, it is only used to change the Open In Browser action, correct?

JonathanWbn (changes requested):

extensions/amazon-aws/src/components/cloudwatch/CloudwatchLogStreams.tsx
crisboarna: It is for the final view logs `Detail` view. As you cannot encompass a `Detail` in a `List` I had to add this on the screen before the logs `Detail` view, otherwise the period to be retrieved cannot be configured.

crisboarna:
Hi @JonathanWbn,
Can you look once more ?

JonathanWbn (approved):
Looks good to me now! Really nice job @crisboarna! 👏

pernielsentikaer (approved):

pernielsentikaer:
Awesome! 🧑‍🚒

PR #3966, opened 2022-12-20
Brave Search extension

CoinFundApp:
## Description

Brave Search with autosuggestion


pernielsentikaer:
What about adding this command to the [Brave](https://www.raycast.com/ron-myers/brave) extension instead?

Spl0itable:
> What about adding this command to the [Brave](https://www.raycast.com/ron-myers/brave) extension instead?

Possible, but I don't use the Brave browser. So, this extension is browser agnostic.

pernielsentikaer:
I was about to merge this and saw that another version was added #4094

Is there something in that one that could benefit this extension, or does it superseed yours?

VishwasShashidhar:
@Spl0itable As @pernielsentikaer mentioned, I've created a similar extension in #4094 with a few more features (like Opening in specific browsers, keeping history etc.), if you think you can add those features, I'll be happy to close my PR.

Spl0itable:
> 

Yes, the one from @VishwasShashidhar has added features like from the Google Search extension. Feel free to close my PR and use theirs. 👍

pernielsentikaer:
Let's do that, but I think we should add you as contributor anyway in the #4094

VishwasShashidhar:
Thank you @Spl0itable & @pernielsentikaer  👍

Spl0itable:
Thank you @VishwasShashidhar 🙌

PR #3969, opened 2022-12-20
[Linear]: Fix icons

thomaslombart:
## Description

- Fix a bug where Linear icons would not show up in the list's accessories for projects
- Fix a bug where Linear icons would not show up if the corresponding icon in the file system doesn't exist
- Rename `assets/project` to `assets/linear-icons`
- Add new predefined icons
 
These two changes should make it so the UI always shows a fallback icon even if the icon is not in the file system or isn't associated with the right file.

## Checklist

### DIFF
--- a/extensions/linear/src/helpers/icons.ts
+++ b/extensions/linear/src/helpers/icons.ts
@@ -1,7 +1,6 @@
-import { Image } from "@raycast/api";
+import { Image, environment } from "@raycast/api";
 import emojis from "node-emoji";
 import fs from "fs";
-import path from "path";
 
 type GetIconParams = {
   icon?: string;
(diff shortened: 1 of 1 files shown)


peduarte (approved):
LGMT 🔥

extensions/linear/src/helpers/icons.ts
```diff
@@ -0,0 +1,31 @@
...
+
+  const emojiRegex = /:(.*):/;
+  if (emojiRegex.test(icon)) {
+    const emoji = emojis.get(icon);
+    // if there's no corresponding emoji, the same emoji code is returned
+    return emoji === icon ? fallbackIcon : emoji;
+  }
+
+  // Linear can add new icons from time to time so some icons may not be in the file system
+  const filePath = path.resolve(__dirname, `assets/linear-icons/${icon.toLowerCase()}.svg`);
```
mathieudutour, line 25: let's use `environment.assetsPath` instead of `__dirname`

mathieudutour (changes requested):

mathieudutour (approved):

PR #3974, opened 2022-12-20
Emoji Search: Vendor `generate-emoji-list` and emoji data.

sxn:
## Description

The extension makes ues of [generate emoji list](https://github.com/leodr/generate-emoji-list), which downloads emoji data from unicode.org and then parses it. Unfortunately, the Unicode Technical Website [is down](https://home.unicode.org/technical-alert-unicode-technical-website-down/), and is expected to be down for at least a few weeks.
As such, I've vendored all of the `generate emoji list` code that the extension relies on, and modified it s.t. instead of feching the data remotely, it is fetched from the `assets` folder.

### DIFF
--- a/extensions/emoji/src/vendor/generate-emoji-list/createEmojiList.ts
+++ b/extensions/emoji/src/vendor/generate-emoji-list/createEmojiList.ts
@@ -1,4 +1,4 @@
-// vendored from https://github.com/leodr/generate-emoji-list
+// vendored from https://github.com/leodr/generate-emoji-list and modified to read emoji data from the `assets/` folder.
 import { environment } from "@raycast/api";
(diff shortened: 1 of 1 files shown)


FezVrasta:
Will this first try to reach unicode.org and if it fails fallback to the local data or it's local-data only?

The advantage of the current approach is that it always provides the most up to date emojis list.

sxn:
no, right now it only fetches the bundled data. We can further update the extension to try both, but for now I thought this should bine fine, since the alternative is the extension not working at all. 😅

thomaspaulmann:
Sounds like a good mitigation for now. We can bring back the fetching of the emojis as soon as the unicode site is back. Does that sound good to you, @FezVrasta?

Issue #3668, opened 2022-11-27
[KeePassXC] Add Yubikey support


silas229:
### Extension

https://github.com/raycast/extensions/tree/9fb1ff46e2ce4185a0c7ff8d769e4fab0677065a/extensions/keepassxc/

### Description

Please add support for Yubikeys:
https://www.mankier.com/1/keepassxc-cli#Options-General_options

Example:
```
keepassxc-cli open -y 1 Documents/keepass.kdbx
```
The slot needs to be configurable (should be mostly 1 or 2).

### Who will benefit from this feature?

Every KeePaasXC user that has their database secured with a Yubikey.

klassm:
Any update? Before the PR goes stale ...

klassm:
Any news? It is again going stale.

silas229:
The PR got closed due to inactivity, is there anyone who can work on it?

klassm:
This issue is again going stale. Any updates?

klassm:
Any news? Or is the extension just dead?

klassm:
Ping

silas229:
Ping

SteffenAuer:
ping

klassm:
Ping

PR #3990, opened 2022-12-21
[KeePassXC]  Add limited Yubikey Support

hmica:
## Description

Try to solve  #3668  by adding limited Yubikey Support

We need to press yubikey on search and action.

To be more usable we should use keepassxc cli with open/close  and reuse opened cli but I will need some helop for this to be done.

## Checklist

### DIFF
--- a/extensions/keepassxc/src/utils/keepassLoader.ts
+++ b/extensions/keepassxc/src/utils/keepassLoader.ts
@@ -48,12 +48,11 @@ const getSearchEntryCommand = async () => ((await getKeepassXCVersion()) >= 2.7 
 const keyFileOption = keyFile != "" && keyFile != null ? ["-k", `${keyFile}`] : [];
 const yubikeyOption = yubikey != "" && yubikey != null ? ["-y", `${yubikey}`] : [];
 // cli options
-const cliOptions = [...keyFileOption,...yubikeyOption, "-q", "-a"];
+const cliOptions = [...keyFileOption, ...yubikeyOption, "-q", "-a"];
 const entryFilter = (entryStr: string) => {
   return entryStr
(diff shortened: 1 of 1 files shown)


extensions/keepassxc/package.json
```diff
@@ -48,6 +48,14 @@
       "description": "KeepassXC.app",
       "placeholder": "Select KeepassXC.app",
       "default": "/Applications/KeePassXC.app"
+    },
+    {
+      "name": "yubikey",
+      "type": "textfield",
+      "required": false,
+      "title": "Yubikey Slot",
+      "description": "Yubikey slot and optional serial used to",
```
zkytech, line 57: It is recommended to give a specific example value in the description.

zkytech: eg: `1` or `1:1`?

zkytech (changes requested):
I don't have a device to test this feature, please make sure that it has been tested and works fine.

extensions/keepassxc/package.json
hmica: > I don't have a device to test this feature, please make sure that it has been tested and works fine.

Hi,

The fact that we need to touch the device each time is a bit cumbersome so I don't thing the PR worth to be merged as it is, the solution were we use the cli to open the database and kept the reference to the opened cli to make research and have an action to close/release cli will be better.

also sometime if the touch has been made to fast or to slow  it create an exception and we need to restart the raycast action. So if  we have the option  to wait until it's successfully opened with good feedback/retry should be considered.

But I thing a good support of keepassXC in Raycast will be very nice.

For now I have sync low critical password to a less secure database so I can I use the plugin as it is.

PR #4000, opened 2022-12-22
Vercast add menubar command

bgiovand:
## Description

Added menubar command to vercast extension and updated packages to latest.

## Screencast

image

## Checklist


extensions/vercast/src/pages/lists/menubar-deployments.tsx
```diff
@@ -0,0 +1,162 @@
...
+const newline = "\n";
+const numShow = 12;
+
+const MenuBarDeployments = ({ projectId }: { projectId?: string }) => {
+  const { user, selectedTeam } = useVercel();
+  const url = getFetchDeploymentsURL(selectedTeam?.id, projectId);
+
+  const { isLoading, data, revalidate } = useFetch<{
+    deployments: Deployment[];
+    // TODO: why can't I `{ headers: FetchHeaders }` here?
```
dgrcode, line 58: I have no idea. Maybe @MaxLeiter or @gleich know?

extensions/vercast/src/pages/lists/menubar-deployments.tsx
```diff
@@ -0,0 +1,162 @@
...
+  const mostRecentDeploymentState = deployments?.[0]?.state || "undefined";
+
+  return (
+    <MenuBarExtra
+      icon={{
+        source: "../assets/vercel.svg",
+        tintColor:
+          mostRecentDeploymentState === "READY" // If the most recent deployment state is READY change the tint to white or dark depending on system theme
+            ? { light: "000000", dark: "FFFFFF" }
+            : stateStyles[mostRecentDeploymentState].color,
```
dgrcode, line 79: Would it be better to add this behavior to the `stateStyles` object?

What I understand is that, for the "READY" case in the `MenuBarExtra` we want it to be either black or white, but in the `MenuBarExtra.Section` we want it to be green. Is that correct?

I was thinking something like the following, but I'm not sure if it's actually better. Just thinking out loud in case you prefer it:

```ts
const stateStyles: StateStyles = {
  READY: {
    icon: Icon.CheckCircle,
    color: Color.Green,
    themedColor: { light: "000000", dark: "FFFFFF" }
  },
  BUILDING: {
    icon: Icon.Bolt,
    color: Color.Yellow,
  },
  // ...
}
```

Then on the component:
```ts
<MenuBarExtra
  icon={{
    source: "../assets/vercel.svg",
    tintColor:
      stateStyles[mostRecentDeploymentState].themedColor ?? stateStyles[mostRecentDeploymentState].color,
  }}
  isLoading={isLoading}
>
...
```

That way, if the `themedColor` property is defined for the `mostRecentDeploymentState`, it uses it, and it defaults to the `color` property if it isn't defined.

I find it a bit cleaner as it removes the exceptional case for "READY" and makes it more generic, but you might prefer the ternary because it's more explicit. I'm happy with both, but just wanted to share this option.

dgrcode (commented):
Looks good to me. Haven't tested locally, just looked at the changes.

Pinging other people that might know the answer to your question:
> why can't I `{ headers: FetchHeaders }` here?

And a suggestion. No need to change if you don't like it :)

extensions/vercast/src/pages/lists/menubar-deployments.tsx
MaxLeiter: Sorry, I definitely wrote that line but can't remember what the actual typing error was.

pernielsentikaer:
I can see you're still working on this @bgiovand - right?

PR #4024, opened 2022-12-23
Arc: Top Sites and Side Bar Tabs

yug2005:
## Description

Hey there, I have been working a little bit on the Arc extension. I wanted to be able to quickly access all items in the side bar of Arc from anywhere through raycast. So I am trying to implement the side bar in raycast. 

I am far from complete but I have ran into some problems so I figured I would make a draft pr and see if anyone else has some suggestions. Check out what I have so far. I have found the file `StorageSidebar.json` and this file contains the spaces, top tabs, pinned tabs, and also open tabs. However, this file does not update instantly. It takes about 5 - 10s for it to update. I am not sure if there is an SQL database that stores this information too, as there is one for user history. 

Another problem is when you open a tab, it will happen through the `Action.OpenInBrowser` and this just makes a new tab with the same url. This is ok for tabs in user history of course but for the side bar, it would be ideal if you did not make a new tab but just went to that tab. Anyone have any ideas on how you could open a specific arc tab through raycast. Of course, tabs are not the only items that can be in the side bar. There can also be easels, notes, and split views that I am currently displaying in raycast but I havent implemented opening them. 

Also, I found that an SQL database contained user top sites so I made a command `top-sites`, basically by copying what was in the `see-history` command. 

## Screencast

When you open the side-bar command, you can see your top tabs and all of your spaces: 
arc 2022-12-23 at 04 28 10
You can even search for any tab or folder from here. You can also go into the spaces: 
arc 2022-12-23 at 04 28 57
There are also nested folders. I have implemented the search such that you search all nested folders for tabs and folders. 
arc 2022-12-23 at 04 29 56

`top-sites` command : 
arc 2022-12-23 at 04 27 13

## Checklist


yug2005:
@thomaspaulmann let me know what you think and if you have any suggestions

thomaspaulmann:
Sorry got a bit delayed. This is super cool. I wasn’t aware of the JSON file. I’m currently looking into AppleScript to read tabs and spaces.  Maybe we can team up?

yug2005:
>Maybe we can team up?

Yep for sure. I am terrible with applescript. How do you plan on using it to read the spaces and tabs? 

Also do you know how to click menu bar items using applescript? I can sometimes get it to work but it takes a pretty long delay. If we could click menu bar items, we could also create Raycast commands for creating new easels and notes in Arc.

yug2005:
Hey I found out how to click menu bar items with applescript from stack overflow and I added a lot of useful commands. Let me know what you think about these commands. Some I am not sure anyone would use, like new folder. 

I think using the StorableSidebar.json file for getting currently open tabs is fine since the delay that it takes to update won't really be noticed. I think you would only use the Raycast command when you have been active on another application for a while. In that time, the json file would have been updated. 

The main thing now for the sidebar command is to somehow open the tabs (as well as easels, notes, and split views) directly rather creating new ones. Do you think there is a way to do this using applescript? I tried some things but they didnt work (again terrible at applescript lol)

thomaspaulmann:
I'm not aware of any APIs to interact with easels, notes, etc. but I agree that this would be nice to have. I'm just wrapping up my AppleScript integration. I do think it's better to rely on the official API instead of proprietary JSON files that might change. How about we merge our changes somehow? I really like the idea of searching a Space and then drill into it. I haven't implemented that yet, maybe you can integrate it on my branch then?

thomaspaulmann:
[Here](https://github.com/raycast/extensions/pull/4240) we go. I've just published my updates that are working on top of Arc's new AppleScript capabilities.

yug2005:
> I do think it's better to rely on the official API instead of proprietary JSON files that might change. 

I think there are a lot of advantages of using the JSON file and I think using it would probably be more stable than using applescript. We also know which tabs belong to which spaces and folders. 

Also there is custom info for icons and colors. I have integrated colors and emoji icons but I am also planning to do these builtin icons somehow.

image

Not sure where these are stored in the assets for Arc. I couldn't find them but I was thinking I could just find similar looking icons.

thomaspaulmann:
Maybe worth reaching out to the Arc team if the JSON is reliable? Then we can use that as an additional source of truth to integrate deeper!

PR #4051, opened 2022-12-25
[New Extension] React Icons

yug2005:
## Description

Search and browse react icons. 

* Copy or paste the icon name
* Copy or paste the icon react componet 
* Copy or paste the icon import statement
* Copy or paste the icon svg

You can also see recent icons and pin icons. 

You cannot view all the icons at once as there is a lot of them but you can search for icons in all categories.

## Screencast

react-icons 2022-12-25 at 03 11 50

react-icons 2022-12-25 at 03 12 08

react-icons 2022-12-25 at 03 19 34

## Checklist

### DIFF
--- a/extensions/react-icons/CHANGELOG.md
+++ b/extensions/react-icons/CHANGELOG.md
@@ -1,3 +1,3 @@
 # React Icons Changelog
 
-## [Initial Version] - 2022-12-24
+## [Initial Version] - 2023-01-05


yug2005:
I see that the first check is failing due to file size being too large. What is the maximum file size allowed for extensions? I can move the SVGs to another repository if that is better?

teziovsky:
I'm trying to create that extension too, but it's complicated, unfortunately. Take a look at [this pull request](https://github.com/raycast/extensions/pull/3514).

At first, I tried to download all icons and it worked properly, but the directory was huge (about 100MB).
Then I decided to split each category of icons, but the comfort of using this extension wasn't very good (It works great without crashing any extension).
Now I'm trying to dynamic import icons with dropdown, but each import is saving in a cache, so when I import a few icon categories, there is info about the JavaScript heap out of memory.
Now I don't have an idea how to solve this problem.

The best that I want to achieve is to have one extension called `React Icons` and select category by dropdown in Grid view.

yug2005:
@teziovsky 
>At first, I tried to download all icons and it worked properly, but the directory was huge (about 100MB).

Yeah, I have that problem too; the directory is around 24 MB. I was thinking I could put the SVG files on another repository on github. I could still determine the URL of the SVG files easily and it should work just as fast without having to cache it. 

>The best that I want to achieve is to have one extension called React Icons and select category by dropdown in Grid view.

This is essentially what I have currently. Initially, I wanted to be able to display all icons so I tried putting all icon names in one JSON file but I could not load this file without getting the JavaScript heap out of memory error. Because there are so many icons, I don't think there is a way you can display all icons at once. I don't think it would be very useful anyway. 

However, I think it is very important to be able to search for icons in all categories. I implemented this with my most recent push. Check it out: 

https://user-images.githubusercontent.com/77864399/210126774-7fdabe20-a976-4385-abc8-ecc409c64611.mov

yug2005:
I tried moving the icons to another repository and it works fine; There is a small load time depending on your internet connection of course. Another drawback is that to use the extension, you would require an internet connection. Can someone from the raycast team confirm whether or not it is okay to keep the SVG files of the icon in the PR?

yug2005:
I have moved the SVG files to another repo so checks are now passing. I think this is ready for review.

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

teziovsky:
🥳🎉

PR #4106, opened 2022-12-31
[bitbucket] Changed the SearchList implementation to work with workspaces with large number of repositories

itsrifat:
👋🏾 

## Description

`SearchList` currently tries to load all the repositories in a workspace into cache before presenting to the user. This works great for workspaces with small number of repositories but it takes forever for workspaces with larger number of repositories. I was trying to use the extension on my workspace that has 10k+ repositories and it never finishes loading the data. 

I changed the implementation to load the fist page of the data as the user types. swr cache will still cache per query. In most cases if the user types 3-5 letters the list shrinks down to less a 100 repositories (even with 10k+ repos!) so imo its not that important to load more than a page.

Also took the opportunity to upgrade to latest version of RayCast.  All other changes except `queries/index.ts` and `searchRepositories.tsx` file are due to RayCast version upgrade.

## Checklist

### DIFF
--- a/extensions/bitbucket/CHANGELOG.md
+++ b/extensions/bitbucket/CHANGELOG.md
@@ -1,0 +1,5 @@
+# Bitbucket Changelog
+
+## [Update] - 2023-01-03
+
+Changed the SearchList implementation to work with workspaces with large number of repositories


Lp-Francois:
Hello! 👋
Thanks for contributing to this extension :)

Could you run `npm audit fix` on your PR? I noticed on NPM install several vulnerabilities (7 moderate, 3 high, 1 critical).

Otherwise LGTM, my workspace only have ~50-ish repositories so I tweaked the page length to 10 in order to experiment the new behavior.

pernielsentikaer:
Is there a reason why this extension has default values for two of the fields? A placeholder should be sufficient.

image

itsrifat:
@pernielsentikaer good catch! Looks like we already had placeholders, I just removed the default values from it.

PR #4147, opened 2023-01-04
Flaticon: Replace global fetch with cross-fetch

fe9lix:
## Description

This replaces usage of global fetch with cross-fetch. Context: Raycast is going to remove support for Node 18's experimental global fetch with an upcoming release to ensure compatibility with the proxy feature.

While testing the extension, another issue popped up: after configuring a generated API key from Flaticon, the extension returns an empty list since the Token expiration check fails. The `expires` property from the API response seems correct with a value of 86400 (one day in seconds) but the code checks for: `expires - Date.now() / 1000 > 30` so it assumes that the returned `expires` value is the current date + the expires time.

@kudrykv Could you check 1) the cross-fetch replacement, 2) the issue with the expiration expression?

## Checklist

### DIFF
--- a/extensions/flaticon/package-lock.json
+++ b/extensions/flaticon/package-lock.json
@@ -9,6 +9,7 @@
       "dependencies": {
         "@raycast/api": "^1.43.2",
         "@raycast/utils": "^1.4.10",
+        "cross-fetch": "^3.1.5",
         "run-applescript": "^6.0.0"
       },
       "devDependencies": {
--- a/extensions/flaticon/package.json
+++ b/extensions/flaticon/package.json
@@ -29,6 +29,7 @@
   "dependencies": {
     "@raycast/api": "^1.43.2",
     "@raycast/utils": "^1.4.10",
+    "cross-fetch": "^3.1.5",
     "run-applescript": "^6.0.0"
   },
   "devDependencies": {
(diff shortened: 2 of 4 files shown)


kudrykv:
> The `expires` property from the API response seems correct with a value of 86400

That is odd. It was working before. I wonder what the chances are of actually me making a mistake. I guess those chances were high.

I treat `expires` as a timestamp, which causes the issue. Though, it is a duration instead. My idea was to look at how much time was left for the token to be alive. If the token is yet to live for 30 seconds or less, I wanted to refresh it.

I suggest converting the duration to a timestamp before saving it to the local storage.

fe9lix:
@kudrykv Could you verify that retrieving a new token via the API returns the duration in seconds (86400) instead of a timestamp?

kudrykv:
@fe9lix yup, I've just did. It indeed returns a duration in seconds.

fe9lix:
@kudrykv Ok, and how come it worked in the past? Did the API change its expires value? Regardless, I'd suggest to address this issue via a separate PR – it just happened to pop up when switching the extension to use cross-fetch.

kudrykv:
@fe9lix to be honest, I have no idea. I realize that it's far more likely that I have made a mistake than the API has changed since. 

Although, it doesn't matter now and we need to fix that. Separate PR is good, too.

fe9lix:
@kudrykv PR is here: https://github.com/raycast/extensions/pull/4152

kudrykv (approved):

PR #4154, opened 2023-01-04
[jira-search] Added a filter in Search Issues to show my issues

itsrifat:
## Description

Added a filter option in the Search Issues page. `All Issues` will do the current behavior and `Assigned to me` option will show issues assigned to the current user.

Had to upgrade to new version of Raycast to take advantage of the new search filter feature. The main changes are in the `Command.tsx` and `issue.tsx` file. Rest of the changes are due to version Raycast version upgrade.

## Screencast

https://user-images.githubusercontent.com/4328772/210659802-75ed2333-357d-4237-af4b-e1f0e8df9416.mov

## Checklist

### DIFF
--- a/extensions/jira-search/CHANGELOG.md
+++ b/extensions/jira-search/CHANGELOG.md
@@ -1,0 +1,5 @@
+# Jira Search Changelog
+
+ ## [Update] - 2023-01-16
+
+ - Added a filter option in the Search Issues page.


pernielsentikaer:
@svenwiegand do you mind checking this 😊

svenwiegand:
Thanks for your contribution @itsrifat! That looks like a useful extension to this extension. I was already thinking about how to utilize the filter accessory for this extension without losing the benefits of typed search queries – this is a great use case.

I've added some comments to your code. They are all kind of personal preferences – So feel free to ignore them if you have another opinion :)

itsrifat:
@svenwiegand I don't see any comments though on this PR 🤔

svenwiegand:
@itsrifat damn, have forgotten to submit the review 🤦 Have done so now. Sorry.

itsrifat:
@svenwiegand ok tried to clean the code up a little 😄. Lemme know if this works. I was also thinking of adding a few more filters like 

* My Open Issues
* Issues in Open sprints (Will allow to quickly see what everyone in `@project` working on )
* My Issues in Open sprints (Will allow to quickly see what issues am I working on in current sprint)

Lemme know if you think these could be useful. I'll add em to the list

extensions/jira-search/src/issue.ts
```diff
@@ -79,13 +81,31 @@ function buildJql(query: string): string {
...
 }
 
-export async function searchIssues(query: string): Promise<ResultItem[]> {
-  const jql = jqlFor(query)
+const isIssueFilter = (filter?: string): filter is IssueFilter => {
+  const filters: readonly string[] = issueFilters
+  return !!(filter && (filters.includes(filter) || filters.includes(filter)))
+}
+
+export async function searchIssues(query: string, filter?: string): Promise<ResultItem[]> {
```
svenwiegand, line 104: Hm, why can‘t we type the `filter` argument here more precisely. You‘ve provided the `SearchCommand` a generic parameter, so that this should be possible or am I missing something. This should completely eliminate the `isIssueFilter()` typeguard.

svenwiegand (commented):
@itsrifat Thanks for the changes and sorry for bugging again, but I think we could safe the typeguard in issue.ts. If I‘m wrong, please let me know – only checked this on an iPad ;)

svenwiegand:
> I was also thinking of adding a few more filters like
> 
> * My Open Issues
> * Issues in Open sprints (Will allow to quickly see what everyone in `@project` working on )
> * My Issues in Open sprints (Will allow to quickly see what issues am I working on in current sprint)
> 
> Lemme know if you think these could be useful. I'll add em to the list

Sounds good. Whatever might be useful to you and others could be added.

extensions/jira-search/src/issue.ts
```diff
@@ -79,13 +81,31 @@ function buildJql(query: string): string {
   return jql + " order by lastViewed desc"
 }
 
-function jqlFor(query: string): string {
-  return isIssueKey(query) ? `key=${query}` : buildJql(query)
+function jqlFor(query: string, filter?: IssueFilter): string {
+  const jql = isIssueKey(query) ? `key=${query}` : buildJql(query)
+
+  if (!filter || filter === "allIssues") return jql
+  let extraJqlForFilter = ""
```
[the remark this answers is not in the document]
itsrifat: ok, added a few more options and moved the thing to its own function

extensions/jira-search/src/types.ts
```diff
@@ -0,0 +1,2 @@
+export const issueFilters = <const>["allIssues", "issuesInOpenSprints", "myIssues", "myIssuesInOpenSprints"]
```
svenwiegand, line 1: As both of the types here are only used inside `issue.ts` I don't see a benefit of having these two in their own file and would move them to `issue.ts` instead.

extensions/jira-search/src/issue.ts
```diff
@@ -79,13 +81,29 @@ function buildJql(query: string): string {
   return jql + " order by lastViewed desc"
 }
 
-function jqlFor(query: string): string {
-  return isIssueKey(query) ? `key=${query}` : buildJql(query)
+function jqlForFilter(filter: IssueFilter) {
+  if (filter === "issuesInOpenSprints") return "sprint in openSprints()"
+  if (filter === "myIssues") return "assignee=currentUser()"
+  if (filter === "myIssuesInOpenSprints") return "assignee=currentUser() AND sprint in openSprints()"
+  return ""
```
svenwiegand, line 88:
```suggestion
  switch (filter) {
    case "allIssues": return ""
    case "issuesInOpenSprints": return "sprint in openSprints()"
    case "myIssues": return "assignee=currentUser()"
    case "myIssuesInOpenSprints": return "assignee=currentUser() AND sprint in openSprints()"
  }
```

In my opinion, this is prettier, and it gives us an exhaustive check of the switch-statement for free: If we add new filters to the type but forget to add them to this switch statement, the typescript compiler will complain about the missing cases. That's why I used `case allIssues` instead of `default`.

svenwiegand (commented):
Two final suggestions.

extensions/jira-search/src/types.ts
```diff
@@ -0,0 +1,2 @@
...
+export type IssueFilter = (typeof issueFilters)[number]
```
svenwiegand, line 2: One more thing: Haven't seen before, that you're not using the array `issueFilters` anywhere, so I don't see a reason to define it at all. Declaring the type would be enough:

```suggestion
export type IssueFilter = "allIssues" | "issuesInOpenSprints" | "myIssues" | "myIssuesInOpenSprints"
```

extensions/jira-search/src/issue.ts
```diff
@@ -79,13 +82,37 @@ function buildJql(query: string): string {
...
+    case "allIssues":
+      return ""
+    case "issuesInOpenSprints":
+      return "sprint in openSprints()"
+    case "myIssues":
+      return "assignee=currentUser()"
+    case "myIssuesInOpenSprints":
+      return "assignee=currentUser() AND sprint in openSprints()"
+    default:
+      throw new Error(`Unknown filter ${filter}`)
```
svenwiegand, line 96: The `default:`-Case isn't necessary here, as the above four cases handle all options allowed by the type `IssueFilter`.

No further comments besides this – PR can be approved.

itsrifat:
@pernielsentikaer you mind merging this PR when you get a chance 🙏🏾

PR #4198, opened 2023-01-09
added json-prettier

lacherogwu:
## Description

Added a package for converting JSON/JS Objects to pretty JSON
The package will paste the pretty JSON automatically from your copied text on your clipboard

## Checklist


lacherogwu:
The reason I was creating this because the existing one works differently I
truly believe mine would give a better DX
For 2 main reasons
1. No window that shows you the output, it’s automatically being pasted
where you need it
2. Handles also JS objects

pernielsentikaer:
You could still add another command to the existing extension, it might be a bit confusing for the end user to pick the "right" one since they have a lot of similarities as in they don't trying to do the same "job"

lacherogwu:
TBH I don't think the existing JSON Format doing good job, I've never seen a case where you need it the formatted value in a separate window, you always need to interact with this.

I think a better suggestion would be to remove/update the old extension because the command they provide is simply unnecessary and don't provide the ideal DX

Aayush9029:
I'm going to side with @pernielsentikaer here, adding it to an existing extension would probably be better for existing thousands of users, as they'll get your amazing update in an instant!!

However this is something you can figure out and decide for yourself.

There's one thing I wished your extension did. Instead of only "directly pasting formatted JSON to an active app", ability to store the formatted JSON to clipboard for later would be nice.

Anyways great work on the extension. The previous extension can definitely be improved upon as it was a bit annoying to work with.

pernielsentikaer:
Hi @lacherogwu 👋 

Sorry for the long response time here 😊

We're not going to allow extensions that directly pass unsanitized input to any eval- or exec-like commands; your `eval` function can be very harmful if you have anything else than a JSON object in the clipboard. This is generally bad practice and can cause serious issues.

> Same as there are multiple vs code extensions(on your store)
None of the vscode extensions does the same; it's 100% different purposes, whereas this one solves the same purpose.

Many extensions have a form and clipboard command, which would be the way to go - I'm sure @Destiner would be up for enhancing it too

lacherogwu:
Hi @pernielsentikaer
This is a very good point.
What is I will sanitize it first and make sure only javascript object can be passed before evaluating
or even find a more creating way instead of `eval()` to read a jsObject from a string and convert it to JSON

Would you approve this PR?

let me know if it is worth my effort.

lacherogwu:
What do you say @pernielsentikaer

Aayush9029:
@pernielsentikaer 
#4654

Aayush9029:
I've created a rough draft update, feel free to change things and please add @lacherogwu as a contributor as well.

lacherogwu:
Hey thanks @Aayush9029
sure no worries, as long as it will be part of the tools I'm good :) not looking for any credit or something like that 😄 

About the code, I saw that you removed the `eval` part, which is okay, but then need to find a better solution to parse js object as a json
since you mention there "Please copy a valid JSON/JS Object, but you are not handling JS Object in any way.

image

pernielsentikaer:
Did you see @lacherogwu's last message @Aayush9029 😊

Aayush9029:
@pernielsentikaer oh, yeah we can remove the JS Object.

On the error line

Aayush9029:
@pernielsentikaer feel free to quick merge https://github.com/raycast/extensions/pull/4745

This should fix the issue raised by @lacherogwu 😄

lacherogwu:
What you did is great, but the main difference here was that you can paste also a JS object, which will be a game-changer for many devs out there...

but you just removed it...

Aayush9029:
@lacherogwu Fair enough :) 

We'll figure out a way to add said feature without using eval, should be simple as importing a library 😭 (surely) 

And we'll get to change the error back to "JSON/JS Object" 🎉

PR #4201, opened 2023-01-09
[todoist] fix focused task when not listed in menubar

kud:
## Description

I've added a small code to check if the task still exists to know if we should display it in the menubar or not (like if it was completed or not) but it seems that this code is buggy. So I've removed it.

If we want to do that, we need to fetch first all the tasks the user got and compare each of them to see if the task still exists or not.

It'll be in another PR one day but let's fix it for right now.

If the user completes the task from Raycast, it'll remove it in the menubar if it's a focused one. However, if the user focuses on a task they complete on the official app, we'll still have the focused task in the menubar. Not so important for the moment. Let's see that another day.

## Checklist

### DIFF
--- a/extensions/todoist/CHANGELOG.md
+++ b/extensions/todoist/CHANGELOG.md
@@ -1,6 +1,10 @@
 # Todoist Changelog
 
-# [Menu bar improvements] - 2023-01-03
+## [Menu bar fix] - 2023-01-10
+
+- Fix focused task when not listed in menubar
+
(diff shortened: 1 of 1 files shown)


pernielsentikaer:
Hi @kud 

Could you add a changelog entry too? 😊

thomaslombart (approved):

kud:
I'll do it the next time yeah @pernielsentikaer 😌

PR #4230, opened 2023-01-11
Add zNotch Extension

zkondor:
## Description

New extension `znotch` to manage macbook's notch visibility directly from `Raycast`. Allows explicitly to hide and show the notch, as well as toggle to an opposite state.

## Screencast

Since `macOS` doesn't provide any option to record laptop's screen's notch, the overall effect looks this way:

usage-demo

## Checklist

### DIFF
--- a/extensions/github/package-lock.json
+++ b/extensions/github/package-lock.json
@@ -7,8 +7,8 @@
       "name": "github",
       "license": "MIT",
       "dependencies": {
-        "@raycast/api": "^1.45.0",
-        "@raycast/utils": "^1.4.11",
+        "@raycast/api": "^1.45.3",
+        "@raycast/utils": "^1.4.15",
         "date-fns": "^2.29.3",
         "graphql-request": "^5.0.0",
         "lodash": "^4.17.21",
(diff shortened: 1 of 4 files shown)


peduarte:
Not sure what's happened with this PR, but shows that every single file has changed. The review page wont even load. 

image

Could you have a look? Thanks!

nhojb:
> Not sure what's happened with this PR, but shows that every single file has changed. The review page wont even load.

It's because @zkondor is attempting to merge directly into `main`. You need to:

- Stash your PR changes somewhere.
- Reset your fork to https://github.com/raycast/extensions
- Create a new branch for your PR
- Apply the changes to the new branch.
- Re-create the PR.

grzegorzkrukowski:
I don't think that's the real problem @nhojb - we should make PR to the `main` branch - it's correct.

The big problem is that on the fork @zkondor removed everything - all extensions and every single file that was in the repository - instead of just adding his extensions.
You can see it by checking his fork content.

grzegorzkrukowski:
I pushed the fix, but in case you need to do it in the future, here is how to do it easily:

```
//clone repository (it will connect to upstream automatically)
gh repo clone zkondor/raycast-extensions
//FINDER: copy extension folder before next steps
//fetch latest changes from our public repository
git fetch upstream
//reset fork to public repository state
git reset --hard upstream/main
// now you have the same state as our public repository, force push changes to fork as we overwrite history
git push origin main --force
// FINDER: re-add the extension
git add .
git commit -m "Add zNotch Extension"
```

grzegorzkrukowski:
I tried to fix the problem, but somehow I was able to force push to it the reset, but I'm not allowed to add the extension back to the repository 🤯

PR #4244, opened 2023-01-13
Fix Laravel Livewire Squishy Icon

austenc:
## Description

The prior version of the icon was missing the white fill in the Jellyfish's (Squishy) eye. This fixes it!

Thank you!


tafhyseni (approved):
Thank you for this change @austenc

PR #4260, opened 2023-01-14
Arc: Add new command to show bookmarks

ankitchouhan1020:
## Description
Arc doesn't support bookmarks for now. Many times, I felt the requirement of saving some tabs that I won't be using frequently. I used to do these via chrome bookmark folders.
In this PR, I added a way to save any tab in the local storage of raycast. Users can show all the bookmarks and remove them using the action panel. I'm utilizing the existing `Search Tab` command to bookmark any tab.

## Screencast
https://youtu.be/_Fgn-b-mQTg

## Checklist


ankitchouhan1020:
The issue with pinned/top tabs is that they are writable, which means if you play around with Tab and move to some other link, the original saved URL is lost. This frustrates me since I don't want to lose the original pinned URL.

I agree though that it can be confusing UX. The best solution will be if Arc opens up a new way of storing read-only links.

I'm not sure how much pin/unpin a tab will be useful. I prefer to move them using drag and drop.

ankitchouhan1020:
Should we play around with naming? Instead of saying bookmarks, it can be just `Saved Tabs` or something..

thomaspaulmann:
> Should we play around with naming? Instead of saying bookmarks, it can be just `Saved Tabs` or something..

That could work. I wonder if this should then be just part of the "Search Tabs" command? So you can save a tab and it will be another section?

I'm planning to add some preferences to sort and filter sections, which can then also cover your usecase. Just a heads up that I'm also adding a new command to search tabs, history, and suggestions in #4294.

ankitchouhan1020:
Yeah... make sense. We can use filters to view the saved tabs instead of a whole command.

Issue #4246, opened 2023-01-13
[Arc] "Search Tabs" and "Search Spaces" commands don't work


aesign:
### Extension

https://www.raycast.com/the-browser-company/arc

### Description

Thank you for the latest update, always wanted to browse arc tabs and spaces with Raycast. Unfortunately, both commands are not working for me as they throw an error.

When I run "Search Tabs":

`Error: Command failed with exit code 1: osascript -e 
    on escape_value(this_text)
      set AppleScript's text item delimiters to the "\""
      set the item_list to every text item of this_text
      set AppleScript's text item delimiters to the "\\\""
      set this_text to the item_list as string
      set AppleScript's text item delimiters to ""
      return this_text
    end replace_chars

    set _output to ""

    tell application "Arc"
      set _window_index to 1
      set _tab_index to 1
      
      repeat with _tab in tabs of first window
        set _title to my escape_value(get title of _tab)
        set _url to get URL of _tab
        set _location to get location of _tab
        
        set _output to (_output & "{ \"title\": \"" & _title & "\", \"url\": \"" & _url & "\", \"windowId\": " & _window_index & ", \"tabId\": " & _tab_index & " , \"location\": \"" & _location & "\" }")
        
        if _tab_index < (count tabs of first window) then
          set _output to (_output & ",\n")
        else
          set _output to (_output & "\n")
        end if
        
        set _tab_index to _tab_index + 1
      end repeat
    end tell
    
    return "[\n" & _output & "\n]"
   
629:637: execution error: Arc got an error: Can’t make location of item 1 of every tab of window 1 into type specifier. (-1700)
    at Lg (/Users/aesign/.config/raycast/extensions/606e26d8-b3c5-4760-8001-dccce1b872c4/search-tabs.js:49:58)
    at /Users/aesign/.config/raycast/extensions/606e26d8-b3c5-4760-8001-dccce1b872c4/search-tabs.js:49:8090
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async we (/Users/aesign/.config/raycast/extensions/606e26d8-b3c5-4760-8001-dccce1b872c4/search-tabs.js:49:41572)
    at async Object.l5 [as current] (/Users/aesign/.config/raycast/extensions/606e26d8-b3c5-4760-8001-dccce1b872c4/search-tabs.js:49:41651)`

When i run "Search Spaces":

`
Error: Command failed with exit code 1: osascript -e 
    set _output to ""

    tell application "Arc"    
[...]

Issue #4258, opened 2023-01-14
[Arc] Unable to search spaces & tabs in new update


abass:
### Extension

https://www.raycast.com/the-browser-company/arc

### Description

<img width="739" alt="CleanShot 2023-01-14 at 00 21 40@2x" src="https://user-images.githubusercontent.com/5505994/212458989-59b2702c-42c3-43ef-936a-8e6e9eb9f1c9.png">

Error: Command failed with exit code 1: osascript -e 
    set _output to ""

    tell application "Arc"    
      set _space_index to 1  
      
      repeat with _space in spaces of front window
        set _title to get title of _space
        
        set _output to (_output & "{ \"title\": \"" & _title & "\", \"id\": " & _space_index & " }")
        
        if _space_index < (count spaces of front window) then
          set _output to (_output & ",\n")
        else
          set _output to (_output & "\n")
        end if
        
        set _space_index to _space_index + 1
      end repeat
    end tell
    
    return "[\n" & _output & "\n]"
   
98:549: execution error: Arc got an error: Can’t make spaces of window 1 into type specifier. (-1700)
    at Z5 (/Users/alex/.config/raycast/extensions/606e26d8-b3c5-4760-8001-dccce1b872c4/search-spaces.js:30:57)
    at /Users/alex/.config/raycast/extensions/606e26d8-b3c5-4760-8001-dccce1b872c4/search-spaces.js:30:8086
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async Re (/Users/alex/.config/raycast/extensions/606e26d8-b3c5-4760-8001-dccce1b872c4/search-spaces.js:30:41401)
    at async Object.Ia [as current] (/Users/alex/.config/raycast/extensions/606e26d8-b3c5-4760-8001-dccce1b872c4/search-spaces.js:38:32)

### Steps To Reproduce

Search spaces, approve permissions, then try to search spaces

### Current Behaviour

Errors out

### Expected Behaviour

Not erroring out

ankitchouhan1020:
Tried using latest Arc version?

thomaspaulmann:
This should be fixed with a version check.

PR #4261, opened 2023-01-14
[Arc] Fix version check issues in Arc Extension

ankitchouhan1020:
Fixes: https://github.com/raycast/extensions/issues/4246
Maybe this too: https://github.com/raycast/extensions/issues/4258

Screenshot with changing required version to 0.85.1
image

### DIFF
--- a/extensions/arc/package.json
+++ b/extensions/arc/package.json
@@ -14,7 +14,8 @@
   ],
   "license": "MIT",
   "contributors": [
-    "zach"
+    "zach",
+    "ankitchouhan1020"
   ],
   "commands": [
     {


thomaspaulmann:
Thanks for the fix! I went and improved the code a bit further with two things:
1. I rely on the cached promise for the version check. This way it's faster than waiting for the AppleScript to finish loading.
2. I added an image to make the Version Check look a bit better.

arc 2023-01-14 at 14 17 12

thomaspaulmann:
And I added you as a contributor 😉

ankitchouhan1020:
Awesome, Thank You! I missed that caching part 🙈

PR #4279, opened 2023-01-16
Messages 2FA Codes: Added polling + general cleanup

thomaspaulmann:
## Description

The main purpose of this PR is to add polling of new 2FA codes. So you can open the command and the code might be a bit delayed. This way it shows up and you don't need to re-open the command.

While I was there, I cleaned up the extension a bit:
- Refactored for easier contribution
- Changed `iMessage` to `Messages` to match official name
- Added keyboard shortcuts to actions
- Cleaned up the UI

## Screencast

imessage-2fa 2023-01-16 at 17 55 51

## Checklist

### DIFF
--- a/extensions/imessage-2fa/src/utils.ts
+++ b/extensions/imessage-2fa/src/utils.ts
@@ -66,4 +66,4 @@ export function extractCode(original: string) {
   }
 
   return code;
-};
+}


thomaspaulmann:
@yuercl given that this extension is about 2FA codes, I was wondering to remove the search type dropdown. I assume just the `code` type should be enough? If we want to keep a search for messages, I would rather build a separate command that is optimized for this. WDYT?

yuercl (approved):
new hook 'useInterval' support

thomaspaulmann:
@yuercl what do you think of removing the filter dropdown as described above?

yuercl:
@thomaspaulmann Perhaps this scenario is more general, most of the time it is to search for all information, and a few times it is to extract the code, or in other words, extracting the code is just a shortcut.  so I agree with your suggestion about "remove filter dropdown menu". This will make the extension more concise.

Issue #4255, opened 2023-01-13
[Arc] Make Open New Tab function like it does in Arc


snelling-a:
### Extension

https://www.raycast.com/the-browser-company/arc

### Description

As it is implemented now, this feature just opens up Google. This is a very different experience to how  <kbd>⌘</kbd>+<kbd>t</kbd> work in Arc. It always felt like that was the intention behind that command not simply opening a new tab. It would be a lot better if we could see the same suggestions, Arc commands, etc. either in Raycast or opening the <kbd>⌘</kbd>+<kbd>t</kbd> menu the topmost page in Arc/Little Arc.

### Who will benefit from this feature?

everyone

### Anything else?

I am happy to help if needed

thomaspaulmann:
@snelling-a that sounds like the universal search command. Given that many folks are interested in it, I'll take a stab at it. It shouldn't be too hard.

snelling-a:
> @snelling-a that sounds like the universal search command. Given that many folks are interested in it, I'll take a stab at it. It shouldn't be too hard.

Nice! looking forward to it

aaronhmiller:
Same. Either be nicer if you could bring up the universal search that Arc uses when launching a New Tab or respect the search engine choice rather than defaulting to google.

thomaspaulmann:
A few ideas here:
- I’ll read the default search engine
- I’ll add an option to define a URL to open
- The other bits should be covered by an universal “Search Arc” command that searches across tabs, history, and suggestions. Then a new tab can be opened via this command. 

WDYT?

snelling-a:
SGTM!

thomaspaulmann:
First step, I added a preference for the URL that opens in the new tab in this PR #4280.

PR #4280, opened 2023-01-16
Arc new tab

thomaspaulmann:
## Description

As mentioned in #4255, I added a preference to define the URL that opens with the "Open New Tab" command. By default it's https://google.com.

## Checklist

### DIFF
--- a/extensions/arc/package.json
+++ b/extensions/arc/package.json
@@ -74,6 +74,7 @@
   "dependencies": {
     "@raycast/api": "^1.45.0",
     "@raycast/utils": "^1.4.0",
+    "is-url": "^1.2.4",
     "lodash": "^4.17.21",
     "run-applescript": "^6.1.0",
     "run-jxa": "^3.0.0",
@@ -81,6 +82,7 @@
     "sql.js": "^1.8.0"
   },
   "devDependencies": {
+    "@types/is-url": "^1.2.30",
     "@types/lodash": "^4.14.191",
     "@types/node": "^18.8.3",
     "@types/react": "^18.0.17",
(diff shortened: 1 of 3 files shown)


extensions/arc/package.json
```diff
@@ -58,19 +58,31 @@
...
+          "required": false,
+          "default": "https://google.com"
+        }
+      ]
     }
   ],
   "dependencies": {
     "@raycast/api": "^1.45.0",
     "@raycast/utils": "^1.4.0",
+    "is-url": "^1.2.4",
```
snelling-a, line 77: do you think it's necessary to use a library for this? it seems like a fairly simple [regex](https://www.oreilly.com/library/view/regular-expressions-cookbook/9781449327453/ch08s01.html) would do the trick.

snelling-a (approved):
🚀 tested and LGTM. 

One comment about validation of the url. Also, related to that, is it possible to do the validation when the user enters the url in the preferences? Might make for a better UX if the feedback is given there instead of when they try to open a new tab.

Screen Shot 2023-01-16 at 9 59 04 PM

extensions/arc/package.json
thomaspaulmann: I checked the package and it does pretty much that, this is why I pulled it in but good research.

thomaspaulmann:
> 🚀 tested and LGTM.
> 
> One comment about validation of the url. Also, related to that, is it possible to do the validation when the user enters the url in the preferences? Might make for a better UX if the feedback is given there instead of when they try to open a new tab.
> 
> Screen Shot 2023-01-16 at 9 59 04 PM

Nice, thanks for testing and verifying it works for you. Unfortunately we don't have validation in the preferences yet. This could be indeed a nice-to-have that would improve the UX. We have plans to revamp the preferences and make it more dynamic and configurable. This should also cover validation. In the meantime I added some validation in the command as you've seen.

PR #4282, opened 2023-01-16
Enhances the Pinboard extension to be able to search with more than just tags (and also to use a cache)

kory-smith:
## Description

This PR adds the ability to search for pins using their title.

## Checklist

### DIFF
--- a/extensions/pinboard/src/api.tsx
+++ b/extensions/pinboard/src/api.tsx
@@ -18,7 +18,7 @@ export function useSearchConstantsBookmarks() {
 
       const data = (await response.json()) as PinboardBookmark[];
       if (data !== undefined) {
-        const constantTagsData = constantTags.split(" ");
+        const constantTagsData = constantTags?.split(" ");
         if (constantTags.length) {
           const items: Bookmark[] = data.map((post) => transformBookmark(post));
           const filtered = items.filter((tag) => {


kory-smith:
Alright, @fe9lix, I'd love it if you'd pull down the code and try it for yourself. I'm hopeful you'll like the experience.

I still have a few changes to make before it's done, but all of the search functions now work with the new useSearchBookmarks hook (which now includes caching). So I'd love your opinion on whether or not it feels right.

pernielsentikaer (commented):
Hi @kory-smith 👋 

Do you have an API token I can borrow for a few minutes to test it, or is that impossible? 

You can always catch me in [Twitter](twitter.com/pertikaer/) and [Slack](https://www.raycast.com/community)

kory-smith:
> Hi @kory-smith 👋
> 
> Do you have an API token I can borrow for a few minutes to test it, or is that impossible?
> 
> You can always catch me in [Twitter](twitter.com/pertikaer/) and [Slack](https://www.raycast.com/community)

Hey @pernielsentikaer! Sure, you can use my API token to test. I'll reach out to you on Twitter DMs to give it to you, and I'll invalidate it once you're done.

pernielsentikaer:
Hi @kory-smith 👋 

I gave the extension some love and rewrote huge parts of the code; let me know how it works.
Also removed an obsolete command since it's not build inside the regular search

fe9lix:
Hey, I've quickly tested the extensions with the changes and have a couple of remarks:

**Search:**
Does the changed version not use typeahead search any more but rather filters the local list? I've got around two thousand bookmarks on Pinboard and it seems that all of them are now loaded initially. With caching, this still performs okayish on my fast machine but I guess it doesn't scale for users with even more bookmarks. Unfortunately, the pinboard API is somewhat limited, which is why the typeahead search has been limited to tags so far. (Pinboard's newer beta API would allow for searches in titles but back then I wasn't able to get access to it and it's state is generally somewhat unclear...)

**Tags:**
The tags show up funky when bookmarks have multiple tags that don't fit – with the new rendering they take up more space now: 
CleanShot 2023-02-19 at 19 29 10@2x
I'd suggest to limit displaying tags to the first 5 maybe. 

**Errors:**
When opening "Search Bookmarks with Constant Tags", I receive the following error toast:
"Failed to fetch latest data"

```
TypeError: Cannot read properties of undefined (reading 'split')
    at Object.parseResponse [as current] (...searchBookmarksWithConstantTags.js:44:1350)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async Object.current (...searchBookmarksWithConstantTags.js:19:314913)
```

kory-smith:
> **Search:** Does the changed version not use typeahead search any more but rather filters the local list? I've got around two thousand bookmarks on Pinboard and it seems that all of them are now loaded initially. With caching, this still performs okayish on my fast machine but I guess it doesn't scale for users with even more bookmarks. Unfortunately, the pinboard API is somewhat limited, which is why the typeahead search has been limited to tags so far. (Pinboard's newer beta API would allow for searches in titles but back then I wasn't able to get access to it and it's state is generally somewhat unclear...)

Yes. Right now, every time you run the "Search Pinboard", the "posts/all" endpoint is called. It's set up currently so that the results of the previous search are cached so that results are _immediately_ visible. This probably has a few problems, the most obvious is that the /all/ API route is rate limited to once every 5 minutes (though testing on my part shows that this isn't always enforced).

My previous version of the code handled things a bit differently: It would call posts/all once, cache the response, and only re-call posts/all if the code determined that the users bookmarks had changed since the last fetch (as determined by the last-updated time provided by Pinboard). If you think it's preferable, we could go back to that?

Sidebar: I also tried to get access to the new version of the API and was unsuccessful.

> Tags:
The tags show up funky when bookmarks have multiple tags that don't fit – with the new rendering they take up more space now:
CleanShot 2023-02-19 at 19 29 10@2x
I'd suggest to limit displaying tags to the first 5 maybe.

This is a good idea and I'll make the change.

I'll also look into that error.

kory-smith:
Let's unstale this bad boy. I moved to Brooklyn and had zero time to work on this. Now I have non-zero time!

@fe9lix, there is a design decision to be made here. We can continue the current method where the previous search results are returned instantly and a fetch is called whenever the search bookmarks endpoint is called.

Alternatively, we can go back to an earlier implementation of mine. We can do some magic with caching such that new fetches are only performed if we've determined that a user's pins have been modified since our last fetch.

What do you think?

kory-smith:
> How about adding the caching for the last X most recent entries so that there's immediately a list when you enter the command, but still doing throttled type-ahead search against the endpoint as the user enters the query? The downside is that (like before) we can only search tags via the API – but I guess that's preferable to potentially slowing down the retrieval and list rendering when there are a lot of bookmarks.

The ability to search for more than just tags is really, really important to me. I have the implementation locally, so i'm good, but I would be willing to bet that other people want to search with tag titles too. If we're slowing things down too much, I think I would prefer if we went back to the slightly cumbersome caching strategy instead of removing the ability to search for tags.

To summarize, I can bring the code to a state where all your bookmarks are available instantly (via a local cache) and no network requests are done unless the code determines that your bookmarks have changed since the last cache time. How does that sound?

fe9lix:
Hey @kory-smith, searching titles is indeed a great improvement. I think we talk about different things re performance, though: what I mean is the the actual list performance and search-as-you-type and not the networking call. I have >2k bookmarks, which are all loaded into the list; the list then gets re-rendered with each keypress. It does perform okayish on my machine but I guess there would be a performance issue on slower machines or when you load even more bookmarks (which, at some point, will also result in an out-of-memory exception when the heap limit is reached). 

Here's the ideal solution from my perspective: load all bookmarks and store them in a local SQLite database; each time the command gets opened, fire a networking call to see whether something has changed; if so, load the current data and update the database; whenever the user searches, query the local database and only render the results into the list. This would obviously be some work, so I wonder what the best path forward is for now. We *could* leave your current version as-is and wait until somebody really runs into the mentioned performance issues. What I'd suggest to still tackle is not loading *all* bookmark entries whenever the command is opened (in my case ~2k entries in JSON format each time) but check for changes and only then load all. If I understand it correctly, that's also what you were suggesting above.

PR #4335, opened 2023-01-19
[code-stash] Import action

ben-greenwood:
## Description
Adds a new action to allow users to import code stashes. Can either be a single .txt file or a folder containing multiple files. It will ignore any files that are not .txt extension

## Screencast

https://user-images.githubusercontent.com/32043941/213522638-e045ad1e-a467-4594-92b9-af804f31c3df.mp4

## Checklist

### DIFF
--- a/extensions/code-stash/src/actions/useImport.ts
+++ b/extensions/code-stash/src/actions/useImport.ts
@@ -5,13 +5,13 @@ const useImport = () => {
   const handleValidationError = async () => {
     await showToast({
       title: "There was an error",
-      message: "Please only select .txt files",
+      message: "Please only select non-media files",
       style: Toast.Style.Failure,
     });
   };
 
-  const isTxtFile = (file: string) => {
-    return fs.lstatSync(file).isFile() && file.endsWith(".txt");
+  const isValidFile = (file: string) => {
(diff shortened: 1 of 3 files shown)


pernielsentikaer:
Is there a reason why only .txt files can be imported?

Would it maybe be better to make a filter with known types?

ben-greenwood:
> Is there a reason why only .txt files can be imported?
> 
> Would it maybe be better to make a filter with known types?

Mostly because thats what the export action exports to, but I can see value in being able to import a code file directly. Is there a way to limit the files that users can select directly from the FilePicker API? Or would I need to manually filter them?

pernielsentikaer:
For now, it's something you'll have to do yourself - could be a preference with comma-separated extensions

ben-greenwood:
> For now, it's something you'll have to do yourself - could be a preference with comma-separated extensions

I've updated the logic to check that its a non-media file. Think it would be cumbersome to keep a list of all valid file extensions since there would be many.

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

timcastillogill:
Nive one @ben-greenwood ! Thanks for implementing. Love this extension

PR #4410, opened 2023-01-24
Hacker News: Improve UX

thomaspaulmann:
## Description

- Change icon for upvotes to use built-in icon
- Use front page as default topic
- Use utils for caching and networking

## Screencast

See updated metadata

## Checklist

### DIFF
--- a/extensions/hacker-news/src/frontpage.tsx
+++ b/extensions/hacker-news/src/frontpage.tsx
@@ -3,13 +3,13 @@ import { startCase } from "lodash";
 import { getStories } from "./hackernews";
 import { Topic } from "./types";
 import { useState } from "react";
-import { useCachedPromise } from "@raycast/utils";
+import { usePromise } from "@raycast/utils";
 import Parser from "rss-parser";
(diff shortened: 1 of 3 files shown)


sxn:
thanks for the changes! The code looks better now, but it looks like you removed the 5 minute caching that was added [here](https://github.com/raycast/extensions/pull/2591), which will lead to API calls being made needlessly (the API only updates every 5 minutes), but more importantly, to us getting rate limited again. 😅

thomaspaulmann:
For now, how about I bring back the caching as part of `getStories()` and leverage the `usePromise` hook instead. Later, when we updated `useCachedPromise` and/or `Cache` we can update the extension. I can leave a TODO for now to avoid similar refactorings.

thomaspaulmann:
As discussed, I have brought back the caching logic. It should work as before, but now leverages our utilities. Later, we can migrate to the `useCachedPromise` hook when we have added support for stale/cache times.

sxn (approved):

PR #4426, opened 2023-01-25
New Extension: ESV Bible

cpenned:
## Description

The extension searches and looks up Bible verses from the English Standard Version using the ESV API. It contains includes two commands:

1) Lookup Verse: type a reference and see returned results
2) Search Bible: type a query and see a list of matching verses

## Screencast

[YouTube Link to Screencast](https://youtu.be/0qZ-nSUkwqU)

## Checklist

### DIFF
--- a/extensions/esv-bible/package-lock.json
+++ b/extensions/esv-bible/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "esv-bible",
-  "lockfileVersion": 2,
+  "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
@@ -8,8 +8,7 @@
       "license": "MIT",
       "dependencies": {
         "@raycast/api": "^1.46.0",
-        "@raycast/utils": "^1.4.16",
-        "node-html-markdown": "^1.3.0"
+        "@raycast/utils": "^1.4.16"
       },
       "devDependencies": {
         "@typescript-eslint/eslint-plugin": "^5.48.2",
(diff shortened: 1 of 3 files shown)


pernielsentikaer (changes requested):
Hi 👋 

* Would it be possible to ask for an application token instead so it's just using that one (so the end user doesn't have to get a token)

* Is it possible to show a failure toast if the wrong credentials are used?

* You could add the compliance to `README.md`

cpenned:
Thanks for the quick review, Peter! 

In response:

1. Unfortunately, the way they've set up the API, there are a few limitations on the key, the most relevant is one of scale. "You may only perform 5,000 queries per day, with no more than 1,000 requests in an hour and no more than 60 requests per minute. If you exceed these limits, your application will be throttled." With that limit, it wouldn't take many users to hit throttle limit. They also make a fairly big point about each key being for personal use only. More at https://api.esv.org/. Let me know if I'm misreading that?
2. Yes, you should get a toast failure message if you don't have a key or if it isn't correct. (see attached screenshot)
forbidden api message
3. Yes, added compliance to the README along with more details on how to get the key and instructions on customizing the extension.

pernielsentikaer (changes requested):
Thanks for the update 🔥 

For the two commands I suggest you to have a look at [List.EmptyView](https://developers.raycast.com/api-reference/user-interface/list#list.emptyview)

Instead of have it look like this

Raycast-2Pb90WSj@2x

It could look like this (which also supports custom image in ~128x128)

image

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

Issue #4451, opened 2023-01-26
[Toggl Track] Migrate to API v9 before April 1 2023, when v8 will be disabled


rogatty:
### Extension

https://www.raycast.com/franzwilhelm/toggl-track

### Description

API v8 which is being used by the extension is deprecated since Oct 1 2022: https://developers.track.toggl.com/changes/2022/10/01/api-v8-deprecation

During March 2023 there will be brownouts:
> During brownout periods, endpoints will be temporarily disabled and then re-enabled after a few hours. The goal is to help API clients to notice the scheduled unavailability and plan their migration accordingly.

On April 1 2023 the v8 will be disabled and the extension will stop working. I fill it as a feature request, but if we don't fix it before, then it'll become a major bug. 😉 

Here is the migration guide: https://developers.track.toggl.com/docs/additional/v8_migration_guide

### Anything else?

https://github.com/raycast/extensions/blob/2de651f199d9c46583b3510f72e24b1c42313a9b/extensions/toggl-track/src/toggl/index.ts#L6

franzwilhelm:
Will try to look at this within march. Any PRs on the extension is welcom @rogatty

PR #4476, opened 2023-01-28
Toggl Track: Upgrade to API v9

rogatty:
## Description

Fixes https://github.com/raycast/extensions/issues/4451.

I decided to remove unused fields instead of adding all the new ones. I also decided to use the default workspace when creating a new time entry when using "No project". [v8 docs](https://github.com/toggl/toggl_api_docs/blob/master/chapters/time_entries.md) say "wid: workspace ID (integer, required if pid or tid not supplied)", but the old code didn't pass `wid` and it was working fine. I'm guessing there was an undocumented fallback to the default workspace, so it should keep working the same way.

I couldn't test the billable feature, as I use a free Toggl account.

## Checklist

### DIFF
--- a/extensions/toggl-track/CHANGELOG.md
+++ b/extensions/toggl-track/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Toggl Track Changelog
 
+## [Update] - 2023-02-07
+
+- Upgrade to Toggle API v9
+
 ## [Enhancements] - 2023-01-27
 
 -  Sort time entries starting from latest


extensions/toggl-track/src/components/CreateTimeEntryForm.tsx
```diff
@@ -16,6 +16,7 @@ function CreateTimeEntryForm({ project, description }: { project?: Project; desc
     try {
       await toggl.createTimeEntry({
         projectId: selectedProject?.id,
+        workspaceId: selectedProject?.workspace_id || (me as Me).default_workspace_id,
```
rogatty, line 19: I'm not 100% sure about this cast. Is it possible that user can get here without `me` loaded correctly? If so, there will be an error thrown here. Should I verify it and use `showToast` instead?

bkeys818: The `/me` request would have to fail for some weird reason, but it's always possible. I'd add a `showToast` just in case it does fail.

extensions/toggl-track/src/storage/index.ts
```diff
@@ -55,6 +60,8 @@ export const storage: TogglStorage = {
 };
 
 export type StorageValues = {
+  // It's null only before the first API fetch
```
franzwilhelm, line 63: Remove comment

extensions/toggl-track/src/components/CreateTimeEntryForm.tsx
franzwilhelm: This is already typed as `Me | null`. Change accordingly:
```suggestion
        workspaceId: selectedProject?.workspace_id || me?.default_workspace_id,
```

extensions/toggl-track/src/toggl/auth.ts
```diff
@@ -34,9 +34,9 @@ export const authenticatedFetch = (token: string, baseUrl: string) => ({
     }
     return (await response.json()) as T;
   },
-  put: async <T>(endpoint: string, body: unknown): Promise<T> => {
+  patch: async <T>(endpoint: string, body: unknown): Promise<T> => {
     const response = await fetch(baseUrl + endpoint, {
-      method: "PUT",
+      method: "PATCH",
```
franzwilhelm, line 39: Did they change the api to do patch requests instead of put requests, or did we change the logic of this extension?

extensions/toggl-track/src/toggl/index.ts
```diff
@@ -1,72 +1,72 @@
...
-          tags,
-          created_with: "raycast-toggl-track",
-        },
+      const now = new Date();
+
+      return api.post<{ data: TimeEntry }>(`/workspaces/${workspaceId}/time_entries`, {
+        billable,
+        created_with: "raycast-toggl-track",
+        description,
+        // For running entries should be -1 * (Unix start time)
```
franzwilhelm, line 47:
```suggestion
        // For running entries should be -1 * (Unix start time). See https://developers.track.toggl.com/docs/tracking
```

extensions/toggl-track/src/toggl/index.ts
```diff
@@ -1,72 +1,72 @@
...
+    getMe: (): Promise<Me> => {
+      return api.get<Me>("/me");
+    },
     getWorkspaces: (): Promise<Workspace[]> => {
       return api.get<Workspace[]>("/workspaces");
     },
     getWorkspaceProjects: async (workspaceId: number): Promise<Project[] | null> => {
       const projects = (await api.get<Project[] | null>(`/workspaces/${workspaceId}/projects`)) || [];
       projects.push({
         id: -1,
```
franzwilhelm, line 19: Could we change this to `undefined` as the api changed?

extensions/toggl-track/src/toggl/index.ts
```diff
@@ -1,72 +1,72 @@
...
-        wid: workspaceId,
-        cid: -1,
+        workspace_id: workspaceId,
+        client_id: -1,
```
franzwilhelm, line 21: Could we change this to `undefined` as the api changed?

extensions/toggl-track/src/toggl/index.ts
```diff
@@ -1,72 +1,72 @@
...
-        },
+      const now = new Date();
+
+      return api.post<{ data: TimeEntry }>(`/workspaces/${workspaceId}/time_entries`, {
+        billable,
+        created_with: "raycast-toggl-track",
+        description,
+        // For running entries should be -1 * (Unix start time)
+        duration: Math.floor((-1 * now.getTime()) / 1000),
+        project_id: projectId !== -1 ? projectId : undefined,
```
franzwilhelm, line 49: If you change according to comments about `-1` elsewhere, we can change this to:
```suggestion
        project_id: projectId,
```

extensions/toggl-track/src/toggl/types.ts
```diff
@@ -1,44 +1,29 @@
...
-  is_private: boolean;
-  active: boolean;
-  template: boolean;
-  at: Date;
-  created_at: Date;
+  client_id: number;
   color: string;
-  auto_estimates: boolean;
-  actual_hours: number;
-  hex_color: string;
```
franzwilhelm, line 30: Did all of these get removed?

franzwilhelm (changes requested):
Great work @rogatty, really appreciate your hard work here! I have some code suggestions. When fixed, I'll pull and test it myself, then we can merge

extensions/toggl-track/src/toggl/auth.ts
bkeys818: The API was changed from `PUT` to `PATCH`

extensions/toggl-track/src/toggl/index.ts
bkeys818: Out of curiosity, why?

extensions/toggl-track/src/toggl/types.ts
bkeys818: They still exist, but @rogatty removed unused properties to clean up types.

extensions/toggl-track/src/components/CreateTimeEntryForm.tsx
bkeys818: The value can’t be undefined

extensions/toggl-track/src/toggl/index.ts
rogatty: If I understand correctly `-1` is used here to identify the "no project" item and it doesn't have anything to do with the API itself. If we changed it to `undefined` then we'd have to guard against it in all places this field is used. I think the current approach makes more sense.

extensions/toggl-track/src/toggl/types.ts
rogatty: Yup, as I said in the description:
> I decided to remove unused fields instead of adding all the new ones.

I could add _all_ the new fields, but it feels like a waste of time. I can revert and leave it as is, but I wouldn't want to go through all unused fields and verify if they're still (?) correct.

I believe unused fields only add noise, and because they're not verified in action, they can be wrong sometimes. For example `at: Date` I've removed were incorrect, as they should be strings.

bkeys818 (approved):
I just finished testing, and everything looks good to me!

extensions/toggl-track/src/components/CreateTimeEntryForm.tsx
franzwilhelm: @bkeys818 it can before it's fetched

extensions/toggl-track/src/toggl/types.ts
franzwilhelm: I'm fine with this as is @rogatty @bkeys818 :)

franzwilhelm (approved):
🚀 🚀 🚀  Great work  🚀 🚀 🚀 

Let's get this merged :D

franzwilhelm:
@pernielsentikaer let's go ;)

pernielsentikaer (approved):

Issue #4439, opened 2023-01-25
Promote how to partially download the repository


garrucho:
### Description

This protip to avoid pulling all the extensions repository is valuable: https://developers.raycast.com/basics/publish-an-extension#alternative-way

But it's not presented there: https://developers.raycast.com/basics/contribute-to-an-extension#:~:text=of%20an%20extension-,Then%2C%20you%20need%20to,-fork%20our%20repository

So, sadly I didn't knew about the tip before cloning the repository, which took a really long time just to work in a simple PR. 😁

mathieudutour:
In 1.47.0, we added a new action on the extensions in the root search: `Fork Extension`! This will create a folder containing only the files related to that extension, again, in a matter of seconds. Once you want to publish your contribution, just run `npx ray publish`!

More info here: [raycastapp.notion.site/ray-publish-c7209e1bd00a4858a68f13d9095910ea](https://raycastapp.notion.site/ray-publish-c7209e1bd00a4858a68f13d9095910ea)

_This is a beta feature and things might break (hopefully gracefully). We aren't updating the developer docs just yet but would love your feedback!_

garrucho:
Wow, nice! Both are great ways, but for sure, guiding through "npx" is even better. 😄 
Then, having that in the Docs will be enough – which I've got as something for the future.

Congrats on the awesome work, guys! I truly love it. ✨

PR #4508, opened 2023-01-31
Add sparse checkout option to Collaboration guide

d0utone:
## Description

Sparse Checkout is a pretty good hint, so it might make sense to add it to the docs as proposed here: https://github.com/raycast/extensions/issues/4439


mathieudutour:
I'm going to close this because we are working on a new flow that makes all of this more accessible!

We added a new action on the extensions in the root search: `Fork Extension`! This will create a folder containing only the files related to that extension, again, in a matter of seconds. Once you want to publish your contribution, just run `npx ray publish`!

More info here: [raycastapp.notion.site/ray-publish-c7209e1bd00a4858a68f13d9095910ea](https://raycastapp.notion.site/ray-publish-c7209e1bd00a4858a68f13d9095910ea)

_This is a beta feature and things might break (hopefully gracefully). We aren't updating the developer docs just yet but would love your feedback!_

PR #4533, opened 2023-02-01
[Xcode] Add command to open URL in Simulator

qeude:
## Description

This modification adds a command to the Xcode extension  to open URLs such as deeplinks in the iOS Simulator.

## Screencast

Screen Recording 2023-02-01 at 23 49 11

## Checklist


pernielsentikaer:
Thanks for the update @qeude 🔥 

Do you midn checking it @SvenTiigi

extensions/xcode/package.json
```diff
@@ -2,7 +2,7 @@
   "$schema": "https://www.raycast.com/schemas/extension.json",
   "name": "xcode",
   "title": "Xcode",
-  "version": "2.1.2",
+  "version": "2.1.3",
```
SvenTiigi, line 5: As a new command has been added it would be good to bump the minor version number and set the version to `2.2.0`

extensions/xcode/src/open-url-in-simulator.command.tsx
```diff
@@ -0,0 +1,37 @@
+import { XcodeSimulatorService } from "./services/xcode-simulator.service";
+import { Clipboard, closeMainWindow, showToast, Toast } from "@raycast/api";
+
+interface OpenSimulatorArguments {
+  url?: string;
+}
+
+export default async (props: { arguments: OpenSimulatorArguments }) => {
```
SvenTiigi, line 8: Type of props must be `LaunchProps` otherwise the command will never receive the URL which the user has entered in the argument placeholder.

```ts
export default async (props: LaunchProps<{ arguments: { url?: string } }>) => { ...}
```

extensions/xcode/src/open-url-in-simulator.command.tsx
```diff
@@ -0,0 +1,37 @@
...
+  const toast = await showToast({
+    style: Toast.Style.Animated,
+    title: "Opening URL in Simulator...",
+  });
+  try {
+    const url = props.arguments.url ?? (await Clipboard.readText());
+    if (!url) throw Error("Please provide or copy a valid URL.");
```
SvenTiigi, line 15: Instead of Try/Catch it would be good to split up the error paths and show a different error e.g. in this case where no URL is available.

extensions/xcode/src/open-url-in-simulator.command.tsx
```diff
@@ -0,0 +1,37 @@
...
+    const result = await XcodeSimulatorService.openUrl(url);
```
SvenTiigi, line 16: What happens if no Simulator is booted? Maybe it would be an option to either check for a booted simulator and show a dedicated error to the user that no simulator is booted or redirect the user to the `Manage Simulators` command to boot a simulator.

Additionally, I think it would be great option for a user to also open URL in the Actions Menu of a Simulator in the `Manage Simulators` command.

extensions/xcode/package.json
```diff
@@ -239,6 +239,21 @@
...
     },
+    {
+      "name": "open-url-in-simulator.command",
+      "title": "Open URL in Simulator",
+      "subtitle": "Xcode",
+      "description": "Open the given URL in the currently booted simualtor.",
+      "mode": "no-view",
+      "arguments": [
+        {
+          "name": "URL",
```
SvenTiigi, line 250: The `name` must match with the property name inside the `LaunchProps` argument. So changing `URL` to `url` should do the job.

extensions/xcode/src/open-url-in-simulator.command.tsx
```diff
@@ -0,0 +1,37 @@
...
+    toast.style = Toast.Style.Success;
+    toast.title = "URL opened in Simulator.";
+    XcodeSimulatorService.showSimulator();
+    await new Promise((r) => setTimeout(r, 1500));
```
SvenTiigi, line 20: Is this timeout mandatory?

extensions/xcode/src/open-url-in-simulator.command.tsx
```diff
@@ -0,0 +1,37 @@
...
+      isSuccess: true,
+      result: result,
+      toast: toast,
+    };
+  } catch (error) {
+    console.error(error);
+    toast.style = Toast.Style.Failure;
+    toast.title = "Error while opening URL in Simulator.";
+    toast.message = `${error}`;
+    return {
```
SvenTiigi, line 32: The return statement can be omitted

extensions/xcode/src/open-url-in-simulator.command.tsx
```diff
@@ -0,0 +1,37 @@
...
+  try {
+    const url = props.arguments.url ?? (await Clipboard.readText());
+    if (!url) throw Error("Please provide or copy a valid URL.");
+    const result = await XcodeSimulatorService.openUrl(url);
+    toast.style = Toast.Style.Success;
+    toast.title = "URL opened in Simulator.";
+    XcodeSimulatorService.showSimulator();
+    await new Promise((r) => setTimeout(r, 1500));
+    await closeMainWindow();
+    return {
```
SvenTiigi, line 22: The return statement can be omitted

extensions/xcode/src/open-url-in-simulator.command.tsx
```diff
@@ -0,0 +1,37 @@
+import { XcodeSimulatorService } from "./services/xcode-simulator.service";
+import { Clipboard, closeMainWindow, showToast, Toast } from "@raycast/api";
+
+interface OpenSimulatorArguments {
+  url?: string;
+}
+
+export default async (props: { arguments: OpenSimulatorArguments }) => {
+  const toast = await showToast({
```
SvenTiigi, line 9: Before showing a toast it would be good to check if an Xcode installation is available as the subsequent invocations are relying on `xcrun`. If Xcode is not installed a dedicated error message would be suitable to inform the user.

extensions/xcode/package.json
```diff
@@ -239,6 +239,21 @@
...
+          "placeholder": "URL to open...",
```
SvenTiigi, line 251: I think other commands which are making use of the Arguments API are simply displaying the name 🤔 so maybe removing the dots would be a good idea.

Bildschirm­foto 2023-02-03 um 17 45 25

SvenTiigi (changes requested):

extensions/xcode/src/open-url-in-simulator.command.tsx
qeude: It's here to avoid dismissing the Raycast window right away without having the success message, so might not be mandatory since you have a direct feedback with the simulator showing up

extensions/xcode/src/open-url-in-simulator.command.tsx
qeude: I did use this at first but for some reason I get a warning on this `LaunchProps` -> `Type 'LaunchProps' is not generic`

extensions/xcode/src/open-url-in-simulator.command.tsx
qeude: Why splitting instead of just throwing an `Error` with a different message?

qeude:
👋🏼  Hey @SvenTiigi, thanks for the very constructive review!

I've made changes you requested and added an Action to the `Manage Simulators` extension so we can open an url from clipboard in a specific Simulator.
For a more convenient usage of this, I did add a filter dropdown as well in this extension.

SvenTiigi:
@qeude thanks for the changes 👍
In addition I've created a PR on your fork https://github.com/qeude/extensions/pull/1 with some improvements regarding this PR.

SvenTiigi (approved):

pernielsentikaer (approved):
Hi 👋

Approved; let's 🚢 it!

PR #4557, opened 2023-02-02
[New Extension] LastPass

alex-popov-tech:
## Description

Hello!

This pr adds new extension for LastPass password manager tool. It allows to search through credentials, copy/paste them, open detailed view.

## Screencast

lastpass-1
lastpass-2
lastpass-3

## Checklist


extensions/lastpass/src/utils/cli.ts
```diff
@@ -0,0 +1,80 @@
...
+      } as Account)
+  );
+  return res;
+};
+
+const execute = async (
+  command: string,
+  opts: { log: { stdout: boolean; stderr: boolean } } = { log: { stdout: false, stderr: false } }
+) => {
+  const wrappedCommand = `zsh -l -c '${command}'`;
```
sxn, line 33: is it necessary to wrap the command? Why not invoke `lpass` directly?

extensions/lastpass/src/accountDetail.tsx
```diff
@@ -0,0 +1,50 @@
...
+    <Detail
+      isLoading={!data}
+      markdown={data ? toMarkdown(data) : ""}
+      metadata={
+        data && (
+          <Detail.Metadata>
+            <Detail.Metadata.Label title={"ID"} text={data?.id} />
+            <Detail.Metadata.Label title={"Name"} text={data.name} />
+            <Detail.Metadata.Label title={"Group"} text={data.group} />
+            <Detail.Metadata.Label title={"Fullname"} text={data.fullname} />
```
sxn, line 36:
```suggestion
            <Detail.Metadata.Label title="Full Name" text={data.fullname} />
```

extensions/lastpass/src/accountDetail.tsx
```diff
@@ -0,0 +1,50 @@
...
+  return error ? (
+    <ErrorDetail error={error} />
+  ) : (
+    <Detail
+      isLoading={!data}
+      markdown={data ? toMarkdown(data) : ""}
+      metadata={
+        data && (
+          <Detail.Metadata>
+            <Detail.Metadata.Label title={"ID"} text={data?.id} />
```
sxn, line 33:
```suggestion
            <Detail.Metadata.Label title="ID" text={data?.id} />
```

extensions/lastpass/src/accountDetail.tsx
```diff
@@ -0,0 +1,50 @@
...
+            <Detail.Metadata.Label title={"Name"} text={data.name} />
```
sxn, line 34:
```suggestion
            <Detail.Metadata.Label title="Name" text={data.name} />
```

extensions/lastpass/src/accountDetail.tsx
```diff
@@ -0,0 +1,50 @@
...
+            <Detail.Metadata.Label title={"Group"} text={data.group} />
```
sxn, line 35:
```suggestion
            <Detail.Metadata.Label title="Group" text={data.group} />
```

extensions/lastpass/src/accountDetail.tsx
```diff
@@ -0,0 +1,50 @@
...
+            <Detail.Metadata.Label title={"Fullname"} text={data.fullname} />
+            <Detail.Metadata.Label
+              title={"Last modified"}
```
sxn, line 38:
```suggestion
              title="Last Modified"
```

extensions/lastpass/src/accountDetail.tsx
```diff
@@ -0,0 +1,50 @@
...
+              text={`${data.lastModified.toLocaleTimeString()} - ${data.lastModified.toLocaleDateString()}`}
+            />
+            <Detail.Metadata.Label
+              title={"Last touched"}
```
sxn, line 42:
```suggestion
              title="Last Touched"
```

extensions/lastpass/src/index.tsx
```diff
@@ -0,0 +1,116 @@
...
+                  content={password}
+                />
+                <Action.CopyToClipboard
+                  icon={Icon.Clipboard}
+                  title={"Copy Username"}
+                  shortcut={{ modifiers: ["cmd"], key: "u" }}
+                  content={username}
+                />
+                <Action.Push
+                  title="Show details"
```
sxn, line 105:
```suggestion
                  title="Show Details"
```

extensions/lastpass/src/index.tsx
```diff
@@ -0,0 +1,116 @@
...
+import { AccountDetail } from "./accountDetail";
+import { ErrorDetail } from "./errorDetail";
+import { lastPass } from "./utils/cli";
+
+interface Preferences {
+  email: string;
+  password: string;
+}
+
+const faviconForDomain = (url: string) => {
```
sxn, line 12: you can use the [getFavicon](https://developers.raycast.com/utilities/icons/getfavicon) util for this 😸

sxn (changes requested):
hi @alex-popov-tech 👋 thanks for the PR!

Unfortunately I couldn't get the extension to work 😓

image

Funny enough, running the echoed commands in the terminal _did_ work as expected – i.e. I was logged in etc. However, the command isn't detected in Raycast. It would also be great to handle the case when the CLI is not detected and instruct users how they can install the command – or offer to install it for them how e.g. [BitWarden](https://github.com/raycast/extensions/blob/main/extensions/bitwarden/src/components.tsx#L12-L23) does it. 

I've also left a couple suggestions to address.

extensions/lastpass/src/utils/cli.ts
alex-popov-tech: because by default it takes `/bin/sh`, which is not people usually use 
CleanShot 2023-02-13 at 23 06 38@2x

extensions/lastpass/src/accountDetail.tsx
alex-popov-tech: thanks, i will commit that myself ✅

extensions/lastpass/src/accountDetail.tsx
alex-popov-tech: thanks, i will commit that myself ✅

extensions/lastpass/src/accountDetail.tsx
alex-popov-tech: thanks, i will commit that myself ✅

extensions/lastpass/src/accountDetail.tsx
alex-popov-tech: thanks, i will commit that myself ✅

extensions/lastpass/src/index.tsx
alex-popov-tech: thanks, i will commit that myself ✅

extensions/lastpass/src/index.tsx
alex-popov-tech: done, thank you ✅

extensions/lastpass/src/utils/cli.ts
alex-popov-tech: I've steal that from `displayplacer` extension

alex-popov-tech:
@sxn thanks for review, i've pushed changes!

Question - following instructions on error view helps to solve your issue?

extensions/lastpass/src/utils/cli.ts
sxn: got it. The problem is, I don't use zsh, but [fish](https://fishshell.com/). 😅 

Unfortunately I can't get it to work without modifying extension code. What _does_ work is either:

- changing `execute` s.t. it uses `fish -l -c <>` instead of `zsh -l -c <>`
- changing occurrences of `lpass` to `/opt/homebrew/bin/lpass`

image

In general though, I'd say it's probably better to add a `cliPath` preference (similar to [BitWarden](https://github.com/raycast/extensions/blob/main/extensions/bitwarden/package.json#L103-L110)), which could also remove the need for dropping into `zsh` (or whatever other shell). 😸

extensions/lastpass/src/index.tsx
```diff
@@ -0,0 +1,122 @@
...
+        }
+      }
+    })();
+  }, []);
+
+  return error ? (
+    <ErrorDetails error={error} />
+  ) : (
+    <List
+      isLoading={accounts.length === 0}
```
sxn, line 62: this isn't a great idea, as it'll lead to infinite loading if users have no accounts in LastPass :D an edge case, I know, but I just ran into (as I don't use LastPass, I needed to create a new account 😸)

extensions/lastpass/src/utils/cli.ts
alex-popov-tech: i think using `zsh` in this case is easier from development point of view ( we don't need to consider lots of shells with their specifics ) and since its in macos by default ( so it should work for everyone )...

thanks for input, it helped me to moreless reproduce issue and find a good solution which should suite everyone 🌻

extensions/lastpass/src/index.tsx
alex-popov-tech: true, thanks!
i will add one more details view for such case

sxn: nice! btw, you can also use an [EmptyView](https://developers.raycast.com/api-reference/user-interface/list#list.emptyview) for similar results.

sxn:
thanks for the changes – the binary is found now, so I can authenticate etc. I did run into one edge case, though:

image

The reason this is happening seems to be the fact that I have an entry that has this checkbox enabled:

image

Toggling it off makes the command work again, but I'm not sure if that's an option for everyone who'd use the extension. 😓 

A couple other remarks on the UX of the extension:

- Please make sure [all actions have an icon, or none do](https://developers.raycast.com/basics/prepare-an-extension-for-store#action-panel) 
image

- all items on the list screen that does not have a URL has a `Key` icon, although none of them are actually passwords 😅 
image

- the detail screen:
  - every entry seems to have a `URL` even though I didn't specify one
  - every entry has a `last touched` label, but it's…1970? 😅 seems like a date being created from an invalid timestamp
  - the `Note` section formatting could be a bit nicer (`Company: Company Name` instead of `Company:Company Name` for example)
  - there's no Action Panel, so I can't copy any information (unless I use the mouse), even though on the previous screen I could
image

You could probably make it a [List.Item.Detail](https://developers.raycast.com/api-reference/user-interface/list#list.item.detail) component instead, since you wouldn't lose context by navigating, and you would still be able to use the same action panel as on the list page.

alex-popov-tech:
> Toggling it off makes the command work again, but I'm not sure if that's an option for everyone who'd use the extension. 😓

Thanks for sharing! I am using lastpass for few years, and i did not know that checkbox exists :D Anyway, I manage to fix that, please confirm it is not visible anymore ( i tested locally too ).

> Please make sure [all actions have an icon, or none do](https://developers.raycast.com/basics/prepare-an-extension-for-store#action-panel)

Thanks, done.

> - it would be nice to have an action that lets me open a particular entry in LastPass so I can edit it.

It sounds super complicated...i would leave this feature for next release :)

> all items on the list screen that does not have a URL has a Key icon, although none of them are actually passwords 😅

Well, LastPass is a password manager tool, and its complicated to diff entities from cli, so i thought its good enough :)

> every entry seems to have a URL even though I didn't specify one

It comes from LastPass cli... I just checked and my notes have url as your - `http://sn` - i can put a guard for this, do you think it is needed? ( maybe its 'special' url for non-password entities, i don't know

> every entry has a last touched label, but it's…1970? 😅 seems like a date being created from an invalid timestamp

Maybe, but it comes from CLI, and i just consume it :) maybe that timestamp is for 'default' entities, which were created automatically

> the Note section formatting could be a bit nicer (Company: Company Name instead of Company:Company Name for example)

All text in Note is not formatted by me at all, i just put some hacks there to replace \n with real linebreaks for example :) 

> there's no Action Panel, so I can't copy any information (unless I use the mouse), even though on the previous screen I could

true, it was intended...i thought it as read-only view on something specific - like notes - where any formatting could be applied, and it makes little sense to copy it as a whole
Not sure I understand about 

> You could probably make it a [List.Item.Detail](https://developers.raycast.com/api-reference/user-interface/list#list.item.detail) component instead

What exactly? I can put there url, but notes and timestamps won't fit there

sxn:
> It sounds super complicated...i would leave this feature for next release :)

Alright, makes sense

> 

> I just checked and my notes have url as your - http://sn - i can put a guard for this, do you think it is needed? 

Yeah it looks useless, so I think it makes sense to just remove it.

> What exactly? I can put there url, but notes and timestamps won't fit there

You have access to `Metadata` for `List.Item.Detail`, the same way you do for `Detail`, so it shouldn't be any problem. They might end up being too large, but it's definitely worth exploring :) but most importantly, if we keep `Detail`, we should provide actions to e.g. copy the password/username from the `Detail`.

sxn (changes requested):

extensions/lastpass/src/index.tsx
```diff
@@ -0,0 +1,130 @@
...
+              title="Manual Sync"
+              shortcut={{ modifiers: ["cmd"], key: "s" }}
+              onAction={() => api.list({ sync: "now" }).then(setAccounts, setError)}
+            />
+          </ActionPanel.Section>
+        </ActionPanel>
+      }
+    >
+      {!accounts.length ? (
+        <List.EmptyView icon={{ source: "pepe.png" }} title="Sorry, you have no accounts" />
```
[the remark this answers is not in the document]
alex-popov-tech: okay 😿

alex-popov-tech:
@sxn thanks for review!

I did url thing you said.

About metadata and details - i would like to try out list details, but the problem is `lpass` cli cannot return `notes` aka `extra` properly - they only return first line :/ 

Which means that for getting this ( for list details ) i would need to fire another cli command for each account, like `lpass show ${id}`, and that means if i have 100 accounts - on list view i will run 100 subprocesses for these details, even tho i might not want it :( 

Is there a way to load list details more lazy? Like, when i first see list i want to call `lpass show` only on first row ( selected one ), and then call `lpas show` only when selected row changes

alex-popov-tech:
@sxn thanks for advices! I was able to refactor things as you suggested!

extensions/lastpass/src/index.tsx
```diff
@@ -0,0 +1,129 @@
...
+                    icon={Icon.Clipboard}
+                    title="Copy Username"
+                    shortcut={{ modifiers: ["cmd"], key: "u" }}
+                    content={username}
+                  />
+                  <Action.CopyToClipboard
+                    icon={Icon.Clipboard}
+                    title="Copy Url"
+                    shortcut={{ modifiers: ["cmd"], key: "l" }}
+                    content={username}
```
sxn, line 119: 🤔 this seems wrong? 😅

alex-popov-tech: indeed, sorry, will push fix now

alex-popov-tech:
@sxn it should, yes.
I want to have your overall approval before it :)

extensions/lastpass/package.json
```diff
@@ -0,0 +1,60 @@
...
+  "author": "alex-popov-tech",
+  "categories": [
+    "Productivity",
+    "Security"
+  ],
+  "license": "MIT",
+  "commands": [
+    {
+      "name": "index",
+      "title": "Search",
```
pernielsentikaer, line 16:
```suggestion
      "title": "Search Vault",
```

pernielsentikaer (changes requested):
Hi 👋 

I have one last thing before you can make the metadata images; when the suggestion and new images are added, then we're ready to go 🔥 

alex-popov-tech:
Hey guys, I've updated code with your comments, updated metadata screenshots, and PR description with new screenshots

pernielsentikaer (approved):
Hi 👋

Looks good to me, approved 🔥

alex-popov-tech:
@pernielsentikaer hey! thanks for approving! i wanted to squash commits before merging, because i was updating screenshots two times :|

pernielsentikaer:
I squah before merge, all should be fine ☺️

PR #4605, opened 2023-02-04
[Wikipedia] Chinese language support with variant

justin0u0:
## Description

Support **Chinese** language with configurable language variant.

## Screencast

wikipedia-1

Preference to select the **Chinese Language Variants**:
image

For example:
- zh-tw
    wikipedia-1
- zh-cn
    wikipedia-2

## Checklist

### DIFF
--- a/extensions/wikipedia/src/wikipedia.ts
+++ b/extensions/wikipedia/src/wikipedia.ts
@@ -16,7 +16,7 @@ export interface PageSummary {
 }
 
 interface Preferences {
-  zhVariant: string;
+  zhVariant?: string;
 }
 
 const preferences = getPreferenceValues<Preferences>();
@@ -26,7 +26,7 @@ function getVariant(language: string) {
     return preferences.zhVariant;
   }
 
-  return "";
+  return undefined;
 }
 
 const getApiUrl = (language = "en") => `https://${language}.wikipedia.org/`;
@@ -69,9 +69,10 @@ export async function getPageData(title: string, language: string) {
   const variant = getVariant(language);
 
   const response = await got
-    .get(`${getApiUrl(language)}api/rest_v1/page/summary/${encodeURIComponent(title)}`, {
-      ...(variant && { headers: { "Accept-Language": variant } }),
-    })
+    .get(
+      `${getApiUrl(language)}api/rest_v1/page/summary/${encodeURIComponent(title)}`,
+      variant ? { headers: { "Accept-Language": variant } } : undefined
+    )
     .json<PageSummary>();
   return response;
 }


vimtor (approved):

NikosEfthias:
👍

extensions/wikipedia/src/wikipedia.ts
```diff
@@ -14,6 +15,20 @@ export interface PageSummary {
   };
 }
 
+interface Preferences {
+  zhVariant: string;
```
JonathanWbn, line 19: As this is an optional preference, I believe we should type it as such.

```suggestion
  zhVariant?: string;
```

extensions/wikipedia/src/wikipedia.ts
```diff
@@ -14,6 +15,20 @@ export interface PageSummary {
...
+
+const preferences = getPreferenceValues<Preferences>();
+
+function getVariant(language: string) {
+  if (language === "zh") {
+    return preferences.zhVariant;
+  }
+
+  return "";
+}
```
JonathanWbn, line 30: As we won't always have a variant, I suggest explicitly returning `undefined` in this case.

```suggestion
function getVariant(language: string) {
  if (language === "zh") {
    return preferences.zhVariant;
  }

  return undefined;
}
```

extensions/wikipedia/src/wikipedia.ts
```diff
@@ -51,8 +66,12 @@ export async function findPagesByTitle(search: string, language: string) {
 }
 
 export async function getPageData(title: string, language: string) {
+  const variant = getVariant(language);
+
   const response = await got
-    .get(`${getApiUrl(language)}api/rest_v1/page/summary/${encodeURIComponent(title)}`)
+    .get(`${getApiUrl(language)}api/rest_v1/page/summary/${encodeURIComponent(title)}`, {
+      ...(variant && { headers: { "Accept-Language": variant } }),
+    })
```
JonathanWbn, line 74: Suggestion for an alternative way that I find a bit more readable:

```suggestion
    .get(
      `${getApiUrl(language)}api/rest_v1/page/summary/${encodeURIComponent(title)}`,
      variant ? { headers: { "Accept-Language": variant } } : undefined
    )
```

JonathanWbn (commented):
Great addition! 👏 
The code looks good!
I added some further suggestions.

justin0u0:
Hi @JonathanWbn, thanks for the advices, I have made changes according to the suggestions!

extensions/wikipedia/package.json
```diff
@@ -40,9 +41,44 @@
...
+  "preferences": [
+    {
+      "name": "zhVariant",
+      "title": "Chinese Language Variants",
+      "description": "Select the default language variant for the Chinese Wikipedia",
+      "required": false,
+      "type": "dropdown",
+      "data": [
+        {
+          "title": "臺灣正體",
```
pernielsentikaer, line 53: At the moment, Raycast doesn't support localization and only supports US English. Therefore, please avoid introducing your custom way to localize your extension.

extensions/wikipedia/package.json
```diff
@@ -40,9 +41,44 @@
...
+        {
+          "title": "澳門繁體",
+          "value": "zh-mo"
+        },
+        {
+          "title": "大马简体",
+          "value": "zh-my"
+        },
+        {
+          "title": "新加坡简体",
```
pernielsentikaer, line 73: At the moment, Raycast doesn't support localization and only supports US English. Therefore, please avoid introducing your custom way to localize your extension.

extensions/wikipedia/package.json
```diff
@@ -40,9 +41,44 @@
...
+        {
+          "title": "香港繁體",
+          "value": "zh-hk"
+        },
+        {
+          "title": "澳門繁體",
+          "value": "zh-mo"
+        },
+        {
+          "title": "大马简体",
```
pernielsentikaer, line 69: At the moment, Raycast doesn't support localization and only supports US English. Therefore, please avoid introducing your custom way to localize your extension.

extensions/wikipedia/package.json
```diff
@@ -40,9 +41,44 @@
...
+        {
+          "title": "大陆简体",
+          "value": "zh-cn"
+        },
+        {
+          "title": "香港繁體",
+          "value": "zh-hk"
+        },
+        {
+          "title": "澳門繁體",
```
pernielsentikaer, line 65: At the moment, Raycast doesn't support localization and only supports US English. Therefore, please avoid introducing your custom way to localize your extension.

extensions/wikipedia/package.json
```diff
@@ -40,9 +41,44 @@
...
+        {
+          "title": "臺灣正體",
+          "value": "zh-tw"
+        },
+        {
+          "title": "大陆简体",
+          "value": "zh-cn"
+        },
+        {
+          "title": "香港繁體",
```
pernielsentikaer, line 61: At the moment, Raycast doesn't support localization and only supports US English. Therefore, please avoid introducing your custom way to localize your extension.

extensions/wikipedia/package.json
```diff
@@ -40,9 +41,44 @@
...
+      "description": "Select the default language variant for the Chinese Wikipedia",
+      "required": false,
+      "type": "dropdown",
+      "data": [
+        {
+          "title": "臺灣正體",
+          "value": "zh-tw"
+        },
+        {
+          "title": "大陆简体",
```
pernielsentikaer, line 57: At the moment, Raycast doesn't support localization and only supports US English. Therefore, please avoid introducing your custom way to localize your extension.

pernielsentikaer (changes requested):

extensions/wikipedia/package.json
justin0u0: Hi, @pernielsentikaer, I think the intention of this preference is not to localize the extension, but just to support the original behavior of Wikipedia.

image

As you can see, there is a dropdown at the top of the Chinese Wikipedia to support language variants. The Wikipedia extension has a dropdown to support searching in different language already, but I can't think of any easy way to add language variants support inside that dropdown.

And I think just like the [**translate** extension](https://github.com/raycast/extensions/blob/main/extensions/google-translate/package.json#L34), although Raycast doesn't support localization, there can be preferences allowing the user to select their language to use this extension.

Or do you have any other suggestions to support the language variants feature of the Wikipedia extension?

JonathanWbn (approved):

extensions/wikipedia/package.json
pernielsentikaer: You can have the preferences just fine, but you need their `title`'s to be in US English 😊

PR #4611, opened 2023-02-05
[Update Extension] add hex to string conversions

Saafo:
## Description

add hex to string and string to hex conversions

## Screencast

image

## Checklist

### DIFF
--- a/extensions/toolbox/package.json
+++ b/extensions/toolbox/package.json
@@ -6,7 +6,8 @@
   "icon": "icon.png",
   "author": "Kang",
   "contributors": [
-    "Saafo"
+    "Saafo",
+    "pernielsentikaer"
   ],
   "license": "MIT",
   "categories": [


Saafo:
Besides this PR, I also wonder why Toolbox doesn't export all of its scripts in `package.json`'s commands, so that users can directly use scripts from the main page of Raycast.

pernielsentikaer (changes requested):
Hi @Saafo 

Thanks for this update, [Buffer](https://nodejs.org/api/buffer.html) is builtin with node, so I don't think you need to import it.

It's up to the extension author to decide, but it could have worked, too 😊

Saafo:
@pernielsentikaer ok I see, I'm quite a newbie to js / ts, so.. up to @kkange to decide😂, both are ok to me. XD

Saafo (commented):
Thanks a lot!

pernielsentikaer (approved):
Hi 👋

Approved; let's 🚢 it!

kkange:
@Saafo

The added functionality looks great. 👍

The reason we didn't export everything as a command is a matter of personal taste.
I wanted to support different types of command behavior while avoiding command duplication.

But I'll think about adding a command export feature if people want it.

PR #4656, opened 2023-02-08
Add Jupiter Aggregator Extension

stevedylandev:
## Description

This extension is pretty simple and has just two commands: 

Menu Bar Token Price - This simply displays the value of SOL in USD in the menu bar 
Compare Token Prices - This lets the user compare the value between two SPL tokens using Jupiter

## Screencast

[Link to Video](https://stevedsimkins.mypinata.cloud/ipfs/QmXExnf5JSNpFW6BTTgrGY5G8pZ8LGyguVBvQ77zQs5Rjt?stream=true)

## Checklist

### DIFF
--- a/extensions/jupiter-aggregator/src/api.ts
+++ b/extensions/jupiter-aggregator/src/api.ts
@@ -34,7 +34,7 @@ export interface TokenListResponse {
   tags: string;
 }
 
-export function fetchPrice <T extends string>(token: T) {
+export function fetchPrice<T extends string>(token: T) {
   return useFetch<PriceResponse<T>>(`https://price.jup.ag/v4/price?ids=${token}`, {
     onError: (error: Error) => {
       showToast({
@@ -72,7 +72,7 @@ export function fetchTokenList() {
   }
 }; */
 
-export function compareTokens <T extends string>(token1: T, token2: string) {
+export function compareTokens<T extends string>(token1: T, token2: string) {
   return useFetch<PriceResponse<T>>(`https://price.jup.ag/v4/price?ids=${token1}&vsToken=${token2}`, {
     onError: (error: Error) => {
       showToast({
--- a/extensions/jupiter-aggregator/src/compare-token-prices.tsx
+++ b/extensions/jupiter-aggregator/src/compare-token-prices.tsx
@@ -28,14 +28,14 @@ export default function Command() {
   const { data: tokenData, isLoading: tokenDataLoading, revalidate } = compareTokens(token1, token2);
 
   const submitHander = async (amount: any) => {
-    revalidate()
-      console.log(tokenData.data[token1].price)
+    revalidate();
+    console.log(tokenData.data[token1].price);
 
-      const total = Number(tokenData.data[token1].price) * Number(amount)
-      console.log(total)
+    const total = Number(tokenData.data[token1].price) * Number(amount);
+    console.log(total);
 
-      push(<Result price={tokenData.data[token1].price} token2={token2} />);
-};
+    push(<Result price={tokenData.data[token1].price} token2={token2} />);
+  };
 
   return (
     <Form


stevedsimkins:
Just starting this PR, will be adding screenshots too and a screencast for review 🫡

stevedsimkins:
Ok should be good to review! :)

extensions/jupiter-aggregator/CHANGELOG.md
```diff
@@ -0,0 +1,3 @@
+# Solana Price Changelog
```
peduarte, line 1: Should we change the title to Jupiter Aggregator?

extensions/jupiter-aggregator/src/api.ts
```diff
@@ -0,0 +1,80 @@
...
+
+export function fetchPrice(token) {
+  return useFetch<PriceResponse>(`https://price.jup.ag/v4/price?ids=${token}`, {
+    onError: (error: Error) => {
+      showToast({
+        style: Toast.Style.Failure,
+        title: "Price Fetch Failed",
+        message: "Check network connection",
+      });
+      console.log(error);
```
peduarte, line 49: Should we remove the log?

extensions/jupiter-aggregator/src/api.ts
```diff
@@ -0,0 +1,80 @@
...
+
+export function fetchTokenList() {
+  return useFetch<TokenListResponse>(`https://token.jup.ag/${list}`, {
+    onError: (error: Error) => {
+      showToast({
+        style: Toast.Style.Failure,
+        title: "Failed to fetch token list",
+        message: "Check network connection",
+      });
+      console.log(error);
```
peduarte, line 62: Same here

extensions/jupiter-aggregator/src/api.ts
```diff
@@ -0,0 +1,80 @@
...
+  });
+}
+
+export const compareTokens = async (token1: string, amount: any, token2: string) => {
+  try {
+    const res = await axios.get(`https://price.jup.ag/v4/price?ids=${token1}&vsToken=${token2}`);
+    const result = amount * res.data.data[token1].price;
+    return result;
+  } catch (error) {
+    console.log(error);
```
peduarte, line 73: Same here

extensions/jupiter-aggregator/src/api.ts
```diff
@@ -0,0 +1,80 @@
...
+        message: "Check network connection",
+      });
+      console.log(error);
+    },
+  });
+}
+
+export const compareTokens = async (token1: string, amount: any, token2: string) => {
+  try {
+    const res = await axios.get(`https://price.jup.ag/v4/price?ids=${token1}&vsToken=${token2}`);
```
peduarte, line 69: Did you consider using `useFetch`? https://developers.raycast.com/utilities/react-hooks/usefetch

extensions/jupiter-aggregator/src/api.ts
peduarte, line 73: Same here

extensions/jupiter-aggregator/src/compare-token-prices.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+export default function Command() {
+  const { data, isLoading } = fetchTokenList();
+  const [token1, setToken1] = useState("SOL");
+  const [token2, setToken2] = useState("USDC");
+  const [amount, setAmount] = useState("1");
+
+  const submitHander = async () => {
+    const price = await compareTokens(token1, amount, token2);
+    await showToast({
```
peduarte, line 13: I found that showing a Toast on success was a bit confusing when I was testing this. For a second, I thought the extension didnt work, because the UI didnt change. 

I think we need to improve the UX expectation here. Im thinking that there could be two:
1. Show the value
2. Copy the value

If 1, we chould update the view to a `Detail`, where we show the value in markdown, with a `Action` that allows users to copy it if they want

If 2, we should use a (`showHUD`) upon copying, as it automatically closes Raycast, and we should ensure the primary action is "Calculate and Copy", or something that manages users expectation better

Thanks!

extensions/jupiter-aggregator/src/menu-bar-token-price.tsx
```diff
@@ -0,0 +1,22 @@
...
+    <MenuBarExtra
+      icon={{
+        source: "solana.svg",
+        tintColor: {
+          light: "black",
+          dark: "white",
+        },
+      }}
+      tooltip="gm"
+      title={data ? `$${price}` : "loading"}
```
peduarte, line 18: While it's loading, I think we should do an early `null` return, instead of changing the title to `loading`. It can be quite jarring otherwise to see things flicking in the Menu Bar

peduarte (changes requested):
Thanks for your extension! Love it! ❤️ 

Reviewed, and looking forward to checking it again

extensions/jupiter-aggregator/CHANGELOG.md
stevedsimkins: Good catch! Thank you!

extensions/jupiter-aggregator/src/api.ts
stevedsimkins: I was actually curious if you knew of a good way to do that. I'm already using fetchTokenList in the compare-token-prices command and I couldn't figure out how to implement two useFetch hooks in the same command. It would error out because there would be two sets of data:
```javascript
const { data, isLoading } = fetchTokenList()
const { data, isLoading } = compareTokens()
```
Any thoughts?

extensions/jupiter-aggregator/src/compare-token-prices.tsx
stevedsimkins: Good thoughts here! I'll work on implementing these

stevedsimkins: Added!

stevedsimkins:
Hey @peduarte! I added the changes suggested, would you mine taking a look and providing further feedback? Also let me know if you have any info on how I could use two useFetch hooks in the same command!

extensions/jupiter-aggregator/src/compare-token-prices.tsx
```diff
@@ -0,0 +1,80 @@
+import { Detail, useNavigation, Clipboard, Action, ActionPanel, Form, showToast, Toast } from "@raycast/api";
+import { fetchTokenList, compareTokens } from "./api";
+import { useState } from "react";
+
+function Result({ price }) {
+  const markdown = `
+# Result:
+# ${price}
```
peduarte, line 8: For extra context, can we add the token name in the result? Eg: `25.5 USDC`

stevedsimkins: Excellent idea!! Done!

extensions/jupiter-aggregator/src/api.ts
stevedsimkins: Awesome!! Thank you!! I've started implementing it but got a little stuck. The main problem is preventing the data from executing before submitting the form / revalidating with the new inputs. I tried the `execute: false` in the fetch options and tried adding a `revalidate()` in the submit handler but it doesn't seem to be working. Any thoughts on how I could fix it? I submitted the changes so you can see what I'm doing

stevedsimkins: Ok so I think I solved some problems by updating the interface for the result to target the right dataset. However now my problem is that I'm having trouble multiplying the data result with the amount. This was working before and I've tried a few things with no luck. Maybe I need to adjust my interface? I was using "string" for price before and it was working, but even if I parse them both as numbers still no luck :/ 

New Interface 
```javascript
export interface PriceResponse<T extends string> {
  data: Record<T, SolItem>;
  timeTaken: number;
  contextSlot: number;
}

interface SolItem {
  id: string;
  mintSymbol: string;
  vsToken: string;
  vsTokenSymbol: string;
  price: string;
}
```
Function trying to calculate 
```javascript
  const { data: tokenData, isLoading: tokenDataLoading, revalidate } = compareTokens(token1, token2);

  const submitHander = async (amount: any) => {
    revalidate()
      console.log(tokenData.data[token1].price)

      const total = Number(tokenData.data[token1].price) * Number(amount)
      console.log(total)

      push(<Result price={tokenData.data[token1].price} token2={token2} />);
};
```

stevedsimkins:
@peduarte Made some more comments about some problems when using useFetch

peduarte (approved):

peduarte:
Nice one!

stevedsimkins:
Oh no!! It's not actually ready, I haven't been able to integrate useFetch successfully 😬😬 I was saying I had some progress but still had some questions in that review haha

stevedsimkins:
@peduarte https://github.com/raycast/extensions/pull/4656#discussion_r1112332781

stevedsimkins:
If it's ok I can just go back to using the old axios request method since it was technically working ¯\_(ツ)_/ lol

stevedsimkins:
https://github.com/raycast/extensions/pull/4892

PR #4738, opened 2023-02-11
Add extension for controlling the Logitech Litra Glow and Logitech Litra Beam lights

timrogers:
This adds a new Raycast extension for turning on and off Logitech Litra Glow and Logitech Litra Beam lights.

This is a second attempt at creation this extension after https://github.com/raycast/extensions/pull/3019 at the end of last year. 

Previously, I was depending on my own [`litra` npm package](https://github.com/timrogers/litra) in the extension itself - but this didn't work because it required native extensions.

This takes a different approach: ask the user to download and install the `litra` npm package separately, and then configure the extension with the path where its CLI binaries are available.

## Checklist

- [x] I ran npm run build and [tested this distribution build in Raycast](https://developers.raycast.com/basics/prepare-an-extension-for-store#metadata-and-configuration)
- [x] I checked that assets used by the README are placed outside of the metadata folder


timrogers:
That push was just me checking off "I checked that assets used by the README are placed outside of the metadata folder". Ready for review 🥳

extensions/logitech-litra/package.json
```diff
@@ -0,0 +1,49 @@
...
+      "description": "Turn your Logitech Litra devices on or off",
+      "mode": "view"
+    }
+  ],
+  "preferences": [
+    {
+      "name": "cliDirectory",
+      "type": "textfield",
+      "required": true,
+      "title": "CLI directory",
```
pernielsentikaer, line 25: Hi @timrogers 👋 

Do you think it makes sense to make this not required?

Maybe set this a not required and only set it if needed; in the code, could we check for the default path (shouldn't it be the same almost always)

timrogers: I don't think it will be the same across machines - it depends on your Node installation. In my case, it's in `/Users/timrogers/.asdf/shims`, which I suspect is different to most machines!

timrogers: It'd be easiest if I could just check `PATH`, but I don't think that's going to work in Raycast given different shells people might be using with different config files.

timrogers:
@pernielsentikaer 👋🏻 Sure you're very busy, but just a reminder to come back to this 🙏🏻

pernielsentikaer (approved):
Hi 👋

PR #4740, opened 2023-02-11
[New Extension] MDN Docs

bobbymannino:
## Description

Search the MDN docs from your keyboard. Next feature is saving recent searches, more to come after too!

### DIFF
--- a/extensions/mdn-docs/CHANGELOG.md
+++ b/extensions/mdn-docs/CHANGELOG.md
@@ -7,4 +7,8 @@
 ## [History & Clipboard] - 2023-02-12
 
 -   [feat] Copy url to clipboard on a search result
--   [feat] Recently searched results are saved to history, can clear all recently searched items at once or by item
+-   [feat] Recently searched results are saved to history
+    -   You can clear all recently searched items at once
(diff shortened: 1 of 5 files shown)


pernielsentikaer:
How is this different than [this one](https://www.raycast.com/krzysztofzuraw/search-mdn)?

bobbymannino:
Same as but quicker speeds

pernielsentikaer:
Would be more nice to enhance the current version, what do you think here @krzysztofzuraw

krzysztofzuraw:
Yes - it will be awesome 🎉

bobbymannino:
not sure how to contribute to an existing one, I have looked on your docs but there's no option to 'fork'

pernielsentikaer:
There are two options as I see it:

1.  This supersedes the current and brings features that are not included in this
2. Change the current version directly from this repo instead

What do you both think?

krzysztofzuraw:
I'm open for both but it seems like 2) don't require any work from end users and avoid confusion with two extensions doing the same.

pernielsentikaer:
2) would just be an update that all users that have the current version installed get without any interference 😊

PR #4748, opened 2023-02-12
Glyph Search

gbougakov:
## Description

Opinionated glyph picker that allows you to paste pretty arrows (←→), special keys (⌃⌘⇧⌫) and other cool stuff (❖◇♡★☆) without opening [Character Viewer](https://support.apple.com/guide/mac-help/use-emoji-and-symbols-on-mac-mchlp1560/mac). The goal of this extension is not to replace the Character Viewer altogether, but to provide a faster alternative for frequently used characters that probably did not make it onto your keyboard

## Screencast

https://user-images.githubusercontent.com/11464592/218311230-b97c1c25-b093-4c16-9019-f9c2f4afe2e5.mp4

## Checklist


gbougakov:
PSA: Due to the lack of better options, I had to include basically every character from Inter as an SVG, so there's about 8mb of SVGs in the `assets` folder. While reviewing, you can filter them out using the `File Filter` dropdown :)

CleanShot 2023-02-12 at 15 38 52@2x

pernielsentikaer:
Just a question, did you [see this extension](https://www.raycast.com/mmazzarolo/unicode-symbols) before making yours?

gbougakov:
Hi! I did, and tried integrating it into my workflow, but I have found myself going back to the Inter charset page and copying the glyphs from there. Usually, it’s a lot easier for me to visually pick one arrow from ten different options than to try and find it using search, and for that a grid layout works a lot better than a list layout, where the glyph is ten times smaller.At first, I wanted to contribute to the extension you mentioned, but then quickly realised that Raycast API doesn’t allow devs to show glyphs in a grid, only vector images, and that would mean I would need to bundle 40 thousand images with the extension. Another approach would be to only use a subset of characters (like I did in my extension), and it will probably work for most people, but for everyone else it will probably break their workflow (https://xkcd.com/1172/)Hopefully this makes sense, but I will understand if my extension will be considered a duplicate On 14 Feb 2023, at 21:47, Per Nielsen Tikær ***@***.***> wrote:﻿
Just a question, did you see this extension before making yours?

pernielsentikaer:
Is there a way to download the icons from a secure path, like the GitHub repo, on opening the command? Would be more nice and save more than 8mb in the repo 😊

gbougakov:
Trying to figure out how to do it right now, will maybe release it next week. I got the SVGs from [here](https://github.com/rsms/inter/blob/master/docs/glyphs/index.html), so I'll probably try to parse them out of this page at runtime.

pernielsentikaer:
You can maybe see how it's done in https://www.raycast.com/yug2005/react-icons

basically: react-icons has a bunch of JSONs to specify what icons exist, and when the extension loads, the data in those JSONs is used to create URLs to GitHub

gbougakov:
Done! Moved the SVGs to a [separate repo](https://github.com/gbougakov/inter-svgs), now the extension loads the glyphs from there

pernielsentikaer (approved):

PR #4790, opened 2023-02-14
[Arc] Add the ability to create a quicklink from a tab

pomdtr:
## Description

Add the ability to create a new quicklink from a tab.

## Screencast

arc 2023-02-14 at 16 01 49

## Checklist

### DIFF
--- a/extensions/arc/CHANGELOG.md
+++ b/extensions/arc/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Arc Changelog
 
-## [Quicklinks] - 2023-02-14
+## [Quicklinks] - 2023-02-15
 
 - Added a new action to quickly create a quicklink from a page.
 


extensions/arc/src/actions.tsx
```diff
@@ -313,3 +313,14 @@ export function EditTabActionSection(props: { tab: Tab; mutate: MutatePromise<Ta
     </ActionPanel.Section>
   );
 }
+
+export function CreateQuickLinkActionSection(props: { url: string; title?: string }) {
+  return (
+    <ActionPanel.Section>
+      <Action.CreateQuicklink
+        quicklink={{ link: props.url, name: props.title, application: "Arc" }}
+        shortcut={{ modifiers: ["cmd"], key: "d" }}
```
thomaspaulmann, line 322:
```suggestion
        shortcut={{ modifiers: ["cmd"], key: "s" }}
```

thomaspaulmann (approved):
Nice addition! I would suggest to use a different shortcut than `⌘ + D` which is often used for duplicating an item (which you could do in theory with tabs). How about using `⌘ + S` for **S**aving?

extensions/arc/src/actions.tsx
pomdtr: I was inspired by the pin/bookmarks tab action of most browser: 
image

pernielsentikaer: I committed the suggestion since it's not really related to pinning tabs (and for making some similarities to how other extensions are doing it)

pomdtr: sounds good, let's merge this !

pernielsentikaer (approved):

PR #4821, opened 2023-02-15
Adding support for more than one Elgato Key Light

omarshahine:
## Description

I updated the core parts of the extension to remove the assumption that there is a single Elgato Key Light. To do this I needed to add an array of Key Lights and then modify each function to loop through the array. This also required that the user tell us how many lights they have.

## Checklist

### DIFF
--- a/extensions/elgato-key-light/src/decreaseBrightness.ts
+++ b/extensions/elgato-key-light/src/decreaseBrightness.ts
@@ -4,8 +4,10 @@ import { run } from "./utils";
 const command = async () => {
   const keyLight = await KeyLight.discover();
   const brightness = await keyLight.decreaseBrightness();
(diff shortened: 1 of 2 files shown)


thomaspaulmann:
This is a great addition @omarshahine! While looking through the code, I was wondering if we could cache the discovered lights? Then we don't need to wait for the discovery to complete every time you change some light. Though, I'm not sure if the information is static or could change often.

omarshahine:
We could store the lights in the preferences storage and add a setting to check say once a week or if there is an error sending a command. Or we could add a setting that lets people discover each time or perhaps on some other cadence.

thomaspaulmann:
@omarshahine sorry for the delay. Yeah, probably best to merge it as is and then we can improve further. 

Thanks again for your contribution. This has been one of the most requested features for the extension. I'm glad we sorted it.

Issue #4816, opened 2023-02-15
[NotePlan 3] doesn't look to see any of my notes


pr1mal:
### Extension

https://www.raycast.com/shakedlokits/noteplan-3

### Description

When I try to view today's note or search for other notes in my NotePlan, it just shows empty list. Given I have NotePlan 3 flavour which is distributed through SetApp, I think this issue is caused by different directory names for regular and SetApp flavours of NotePlan. I have notes stored in following directory:

`$HOME/Library/Containers/co.noteplan.NotePlan-setapp/Data/Library/Application Support/co.noteplan.NotePlan-setapp`

### Steps To Reproduce

Environment: MacOS 13.1, NotePlan 3.8 (SetApp flavor), latest Raycast. Noteplan is configured to use CloudKit for cross-device synchronization.

1. invoke Raycast with hotkey
2. type "Daily Plan" (or "Search notes"), hit Return

### Current Behaviour

Raycast displays empty space, no single line of text

### Expected Behaviour

Raycast displays today's note content (or list of notes relevant to search)

PR #4828, opened 2023-02-16
Fix for #4816: add support for SetApp flavour of NotePlan3

pr1mal:
## Description

NotePlan3 is distributed through AppStore and SetApp. These distribution methods result in different naming of directories where NotePlan3 stores notes.
This fix tests for distribution flavor of NotePlan: if SetApp directory exists then it is used as a location to search for notes, and falls back to AppStore location otherwise

## Checklist

### DIFF
--- a/extensions/noteplan-3/package.json
+++ b/extensions/noteplan-3/package.json
@@ -5,7 +5,10 @@
   "description": "Allows you to see your daily plan, search notes and launch straight into NotePlan.",
   "icon": "command-icon.png",
   "author": "shakedlokits",
-  "contributors": ["lukerandall", "pr1mal"],
+  "contributors": [
+    "lukerandall",
+    "pr1mal"
+  ],
   "categories": [
     "Productivity",
     "Applications",


extensions/noteplan-3/src/lib/note-utilities.ts
```diff
@@ -1,14 +1,22 @@
...
 import { enGB } from "date-fns/locale";
 import { capitalize } from "lodash";
 import { Icon } from "@raycast/api";
 
-const NOTE_PLAN_URI = `${homedir()}/Library/Containers/co.noteplan.NotePlan3/Data/Library/Application Support/co.noteplan.NotePlan3`;
+function getNotePlan3URI() {
+  const appstoreFlavorPath = `${homedir()}/Library/Containers/co.noteplan.NotePlan3/Data/Library/Application Support/co.noteplan.NotePlan3`;
+  const setappFlavorPath = `${homedir()}/Library/Containers/co.noteplan.NotePlan-setapp/Data/Library/Application Support/co.noteplan.NotePlan-setapp`;
+
+  const useSetappVer: boolean = existsSync(joinPath(setappFlavorPath));
```
shakedlokits, line 15: What is the purpose of the `join`?

extensions/noteplan-3/src/lib/note-utilities.ts
```diff
@@ -1,14 +1,22 @@
...
+
+  return useSetappVer ? setappFlavorPath : appstoreFlavorPath;
```
shakedlokits, line 17: Is there an option where a user installed the app through `SetApp` in the past and since then moved to the `AppStore`?

extensions/noteplan-3/src/lib/note-utilities.ts
pr1mal: It's nothing atm. Will remove it, thanks 😊

extensions/noteplan-3/src/lib/note-utilities.ts
pr1mal: I guess this may happen. I also guess it's usually otherwise, because there is no difference in features/functionality between them.
In any case, we can only reliably handle this via some configuration option where user explicitly tells us which version they use. Would you like to add it?

pernielsentikaer:
@shakedlokits could you check it once more?

pr1mal:
Hi there! Please let me know if this PR should be improved in any way - I just don't feel I have right to decide it on my own )

extensions/noteplan-3/src/lib/note-utilities.ts
shakedlokits: Whoops, sorry, missed this notification.
Yes, I think we should add this configuration. 
As much as I don't like adding the need for user intervention, it's better than implicit behaviour without any other way to intervene.

shakedlokits (commented):
Sorry, I missed the first notification, thanks for the heads up (:

extensions/noteplan-3/src/lib/note-utilities.ts
pr1mal: Added configuration: now user can select flavour using dropdown, which is set to "AppStore" flavour by default
Please check )

pr1mal:
@shakedlokits I have added configuration: now user can select flavour using dropdown, which is set to "AppStore" flavour by default
Please check )

extensions/noteplan-3/package.json
```diff
@@ -37,6 +37,24 @@
       "title": "File Extension",
       "default": "txt",
       "description": "The file extension you use for notes"
+    },
+    {
+      "name": "noteplanFlavour",
```
shakedlokits, line 42: Why flavour?
I'm not sure people would follow the language.
Let's try something that communicates to non-savvy users that this is where their app came from.
More on the lines of "Installation Source" or "Store" or "Distributor", etc.
What do you think?

shakedlokits (commented):
Looks good, small comment about the language.

extensions/noteplan-3/package.json
shakedlokits: @pernielsentikaer how would you call the application source? Having seen many extensions by now

pr1mal: Makes sense. And, I guess, it would be good to rename variables in code too

pr1mal: @shakedlokits @pernielsentikaer I have renamed 'flavour' to "installation source" in user-facing text and also variable names, please have a look

shakedlokits (approved):
Looks good, great job! Thank you for the patience 🙏🏽
amaze-gif

pr1mal:
Hi @pernielsentikaer, is there anything else to be done before this PR can be merged?

pernielsentikaer (approved):
Hi 👋

PR #4840, opened 2023-02-17
Color picker: Fix error message if menu bar component disabled

EmersonEmerson:
## Description

When using the color picker extension, and having the 'Menu bar' component disabled like so:

image

When using "Pick color", the functionality works but the output wrongly displays an error:

e9b0c1c792ac58be14cb5827652e9835675d50686681484baafb40e29f53e52e copy

Added a rescue in case menu bar widget doesn't exist (Forgive me but I can't find anything in your API docs to achieve something cleaner)

### DIFF
--- a/extensions/color-picker/package.json
+++ b/extensions/color-picker/package.json
@@ -6,7 +6,9 @@
   "icon": "icon.png",
   "author": "thomas",
   "contributors": [
-    "otakustay"
+    "otakustay",
+    "EmersonEmerson",
+    "pernielsentikaer"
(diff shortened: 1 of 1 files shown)


EmersonEmerson:
(the error log)
image

pernielsentikaer:
I removed the duplicated code @EmersonEmerson - I think you forgot that 😊

pernielsentikaer (approved):
Hi 👋

Approved; let's 🚢 it!

Issue #4814, opened 2023-02-15
[Linear] Views such as Active Cycle do not support paginated results


joel-u410:
### Extension

https://www.raycast.com/linear/linear

### Description

The Linear GraphQL API [paginates](https://developers.linear.app/docs/graphql/working-with-the-graphql-api/pagination) results by 50 issues at a time by default. As far as I can tell, the [API client in the Linear extension](https://github.com/raycast/extensions/blob/528e717591494b099704a434c0974d4a167f6583/extensions/linear/src/api/getIssues.ts) does not support fetching paginated results at all. This results in incomplete data, for example, in the "Current Cycle" view anytime there are more than 50 issues in the cycle. This is particularly a problem in a team context, where one cycle is used by a large team, resulting in more issues -- filtering by my name to view only _my_ issues in the active cycle results in a very incomplete list, because many of the issues were never loaded into Raycast in the first place.

### Steps To Reproduce

1. Have more than 50 issues in the active cycle in Linear
2. Open the Active Cycle view in the Linear Raycast extension
3. Count the issues

### Current Behaviour

Only 50 issues are displayed

### Expected Behaviour

All issues should be displayed

thomaspaulmann:
Hey, that is indeed a limitation of the extension right now. Our extensions API doesn't provide an easy way to implement pagination yet. It's something that we had in mind to implement at some point. Could there be any "work-around" for now to help with your problem, f.e. showing some list item that there are more issues that would then fetch them or open Linear?

joel-u410:
Maybe an easy (though imperfect) fix could be to just increase the requested items to be more than the default --

e.g. on this line: https://github.com/raycast/extensions/blob/main/extensions/linear/src/api/getIssues.ts#L211

changing

```
      query($cycleId: String!) {
```

to

```
      query(first: 500, $cycleId: String!) {
```

joel-u410:
I went ahead and implemented pagination for this -- it doesn't change anything about the extensions API, but instead makes multiple requests to the Linear GraphQL API until all pages have been retrieved. I wrote it in a generic way so you could easily use it for any other Linear GraphQL queries too if you'd like.

The code is in https://github.com/raycast/extensions/pull/4856. Would be great to merge this into the official extension! (Otherwise, if you don't want to merge it, is there a way I can install my fork locally instead of the official one?)

PR #4856, opened 2023-02-17
Implement pagination for getActiveCycleIssues.

joel-u410:
## Description

Implementing pagination support for the Linear GraphQL API. Initially it is used only in `getActiveCycleIssues` but could be easily added to any other GraphQL request. This resolves issue #4814.

## Checklist

### DIFF
--- a/extensions/linear/package.json
+++ b/extensions/linear/package.json
@@ -213,6 +213,14 @@
       "title": "Issue Custom Copy Action",
       "description": "Add a custom copy action to the issue's actions following the format you want.\nAvailable keys: {ISSUE_TITLE}, {ISSUE_ID}, {ISSUE_URL}, {ISSUE_BRANCH_NAME}\nExample: '{ISSUE_ID}: {ISSUE_TITLE} ({ISSUE_URL})'",
       "required": false
+    },
+    {
+        "name": "limit",
+        "title": "API Issue Limit",
+        "description": "The maximum number of issues to fetch from the Linear API at once. Be careful when changing this, as too large a number can cause you to exceed your API rate limit and/or cause heap memory errors.",
+        "type": "textfield",
+        "default": "50",
+        "required": false
     }
   ],
   "dependencies": {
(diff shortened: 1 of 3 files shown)


thomaslombart (changes requested):
Thanks for this PR. I've reviewed it and have some feedback:

I can't retrieve any issues in the `Active Cycle` command. I get this specific error:

linear 2023-02-21 at 10 20 48

Could you please take another look at the implementation and hopefully fix it? 😊

Please, request a new review when you are ready. Feel free to contact me here or in the [Slack community](https://www.raycast.com/community) if you have any questions.

joel-u410:
Yikes! I guess it should have some limit to how many pages it will retrieve. Out of curiosity, how many issues do you have in your active cycle?

joel-u410:
I added a limit so it will stop after fetching 4 pages. I wonder if this should be a configurable option? (Seems like it would be an "advanced" config though, maybe too low level for extension preferences.)

joel-u410:
Update - still looking into this whenever I have time.

joel-u410:
OK @thomaslombart -- I've _sort of_ punted on this by introducing a preference for "API Issue Limit" that allows the user to set a larger-than-default value in order to fetch multiple pages. The default is `50` which will maintain current behavior, and _should_ prevent heap errors from occurring. But it will easily allow for use cases like mine, where I want to fetch more issues. (With a warning in the preference that setting it too large can result in hitting API rate limits and/or heap errors.)

If this is acceptable, can you test again and see if you still get memory errors with the default settings? (And, perhaps, try some larger settings and see if you can trigger the heap error by setting the limit too high.)

thomaslombart (approved):
I don't have the error anymore. Thanks for the contribution!

thomaslombart:
That being said, what's weird though is that setting a ridiculously high number did not make the error show up again. I tried setting the API size limit to 50, 100, 1000 and 10000. None of them caused the error 🤷

thomaslombart:
Should I go ahead and merge it?

joel-u410:
> Should I go ahead and merge it @joel-u410?

Yes, please do, thanks!

joel-u410:
> That being said, what's weird though is that setting a ridiculously high number did not make the error show up again. I tried setting the API size limit to 50, 100, 1000 and 10000. None of them caused the error 🤷

That is weird indeed. It probably also depends on how many issues are actually returned by your query (i.e. how many are in your active cycle). In my testing, I found the error to be somewhat intermittent, but definitely triggered more by larger queries.

(I was able to trigger the error most consistently by adding pagination to the "search issues" query, and setting it to a large number.)

PR #4867, opened 2023-02-18
Add strip text extension.

desduggan:
## Description

A very simply extension which takes the latest item in your clipboard, removes formatting and trims it, and puts it back on the clipboard. 

This is useful when you copy some rich text from a website, but when you paste into an app (like Gmail) you don't want font size 26! 

## Screencast

https://user-images.githubusercontent.com/3515525/219883617-46bca37b-3b37-43bd-864c-897f8f31c567.mp4

## Checklist

### DIFF
--- a/extensions/strip-text/src/index.ts
+++ b/extensions/strip-text/src/index.ts
@@ -1,8 +1,8 @@
 import { showHUD, Clipboard } from "@raycast/api";
 
 export default async function main() {
-  const { text } = await Clipboard.read();  
-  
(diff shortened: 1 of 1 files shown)


pernielsentikaer:
How does this differ from [Paste as Plain Text](https://www.raycast.com/koinzhang/paste-as-plain-text) except that this copies instead of pasting

desduggan:
Ah, does the same thing. Couldn't find that in the store. We can close the
PR as it would be a duplicate.

PR #4947, opened 2023-02-24
1Password: Add 'sort by' option for 1Password items

EmersonEmerson:
## Description

The 1password extension currently sorts by category by default. Even when within a category, sort is by title. Suggestion is to sort by updated_at date instead, similar to how the real app does it. Increases usability greatly imo. 

Note: I implemented changes for v7 but did not test v7 functionality. v8 functionality tested.

### DIFF
--- a/extensions/1password/src/v7/utils.ts
+++ b/extensions/1password/src/v7/utils.ts
@@ -1,4 +1,4 @@
-import { Cache, Icon } from "@raycast/api";
+import { Cache, Icon, getPreferenceValues } from "@raycast/api";
 
 import fg from "fast-glob";
 import { homedir } from "os";


extensions/1password/src/v7/utils.ts
```diff
@@ -9,6 +9,7 @@ import { readFileSync } from "fs";
 const cache = new Cache();
 
 const ITEMS_CACHE_NAME = "@items";
+export const SORTING_METHOD = getPreferenceValues<Preferences>().sortingMethod || "title";
```
pernielsentikaer, line 12: Where do you get `Preferences` from?

EmersonEmerson:
I redownloaded 1Password7 so I could test that portion as well...

I receive this error when I attempt to run the extension with 1Password 7 set as the version:

```
TypeError: Cannot read properties of undefined (reading 'localeCompare')

ji:Items.tsx:12:57
    at Zi
wr:index.js:5:2490
    at ray-navigation-stack 
```

The 1Password CLI is also a requirement here, and it's requirements only list 1Password 8:
CleanShot 2023-02-28 at 22 36 03@2x

as such, would suggest removing 1Password 7 option.

khasbilegt:
We're planning to keep the 1Password 7 support for a little while, so the changes should be made accordingly.

extensions/1password/src/v7/utils.ts
pernielsentikaer: I don't think this is solved yet? 

Raycast-gpjHxwza@2x

This defines the type, but there is no type?

Issue #4953, opened 2023-02-24
[Amazon AWS] Not getting AWS credentials


andreacavagna01:
### Extension

https://www.raycast.com/Falcon/aws

### Description

The Amazon AWS Extension of Raycast does not work with aws credentials in the credentials file

### Steps To Reproduce

1 - start Leapp to generate a set of short-lived AWS credentials inside `~/aws/credentials` folder
2 - searching for s3 bucked fo any other command
3 - no result found always

JonathanWbn:
relates to #4941

I will check this over the weekend.

boringContributor:
Hi @JonathanWbn,

I created your mentioned issue and already found the root problem. 

https://github.com/raycast/extensions/pull/4984

I created a PR here which fixes the problem. Feel free to collaborate with me on this issue :)

JonathanWbn:
I opened https://github.com/raycast/extensions/pull/5141 which I expect fixes this. @andreacavagna01 could you check this?

andreacavagna01:
Sorry for being late, by checking the repo, I see that the extension only checks for ENV variables, which is not the default method of credentials generation of AWS. This is why the app is not working with Leapp. Can we re-open this?

Issue #4941, opened 2023-02-23
Amazon AWS Extension can't figure out credentials with session tokens


boringContributor:
### Extension

https://www.raycast.com/Falcon/aws

### Description

I want to use the AWS extension and figured out that it has problems to pick up temp. session tokens from the `.aws` folder. 

Running any commands with the `awscli` and `aws sts get-caller-identity` returns the correct data so my local configs do work.

I'm creating those session tokens with Leapp (https://www.leapp.cloud/) which is a pretty common thing.

However the extension fails with 
```
CredentialsProviderError: Could not load credentials from any providers
    at /Users/sebastian/.config/raycast/extensions/cdcb4b69-7df7-4f7e-977b-1406967eb3e7/cloudwatch.js:23:6960
    at /Users/sebastian/.config/raycast/extensions/cdcb4b69-7df7-4f7e-977b-1406967eb3e7/cloudwatch.js:1:57606
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async d (/Users/sebastian/.config/raycast/extensions/cdcb4b69-7df7-4f7e-977b-1406967eb3e7/cloudwatch.js:1:57950)
    at async Um.credentialProvider (/Users/sebastian/.config/raycast/extensions/cdcb4b69-7df7-4f7e-977b-1406967eb3e7/cloudwatch.js:1:58099)
    at async Um.signRequest (/Users/sebastian/.config/raycast/extensions/cdcb4b69-7df7-4f7e-977b-1406967eb3e7/cloudwatch.js:2:520)
    at async /Users/sebastian/.config/raycast/extensions/cdcb4b69-7df7-4f7e-977b-1406967eb3e7/cloudwatch.js:12:4923
    at async bm.retry (/Users/sebastian/.config/raycast/extensions/cdcb4b69-7df7-4f7e-977b-1406967eb3e7/cloudwatch.js:1:51762)
    at async /Users/sebastian/.config/raycast/extensions/cdcb4b69-7df7-4f7e-977b-1406967eb3e7/cloudwatch.js:1:35639
    at async Object.eA [as current] (/Users/sebastian/.config/raycast/extensions/cdcb4b69-7df7-4f7e-977b-1406967eb3e7/cloudwatch.js:48:17204)
```

### Steps To Reproduce

1. Create session tokens / temporary credentials with Leapp
2. Open Raycast and go to the AWS extension
3. Try any service

### Current Behaviour

[...]

PR #4984, opened 2023-02-25
fix(aws): check for default named profile

boringContributor:
## Description

The AWS extension does not properly handle default named profile. 

## Checklist

### DIFF
--- a/extensions/amazon-aws/CHANGELOG.md
+++ b/extensions/amazon-aws/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Amazon AWS Changelog
 
+## [Fix] - 2023-03-01
+
+- Handles default named profile correctly
+
 ## [Renamed command] - 2023-02-03
 
 - Renamed Elasticsearch Service to OpenSearch Service


boringContributor:
relates to: https://github.com/raycast/extensions/issues/4941 and https://github.com/raycast/extensions/issues/4953

JonathanWbn:
Thank you for tackling this! 👏 

I could reproduce your issue by using Leapp. Your changes seem to fix the issue. 👍 

However, I believe these changes will cause different problems, as the `isDefaultProfile ? 'default' : selectedProfile` basically overwrites the profile selection setting.

I think the problem rather laid in the `const profileOptions` definition. This was not accounting for the fact that `configFile` is an empty object when it doesn't exist (as opposed to `undefined`).

So I suggest reverting the `isDefaultProfile`-changes and introducing a change like this:
```ts
const { data: configs = { configFile: {}, credentialsFile: {} } } = useCachedPromise(loadSharedConfigFiles);
const { configFile, credentialsFile } = configs;
const profileOptions = Object.keys(configFile).length > 0 ? Object.keys(configFile) : Object.keys(credentialsFile);
```

This will let `profileOptions` be a list of profiles even when the configFile is empty or non-existent. In the case of a simple Leapp-setup, that would be `['default']`.

Unrelated to the bug, I think we could also make the use effect more robust, like so:
```ts
useEffect(() => {
  const isSelectedProfileInvalid = selectedProfile && !profileOptions.includes(selectedProfile);

  if (!selectedProfile || isSelectedProfileInvalid) {
    setSelectedProfile(profileOptions[0]);
  }
}, [profileOptions]);
```

Looking forward to hear what you think and see if these changes work for you.

pernielsentikaer:
Hi @boringContributor 👋 

Did you see comments from @JonathanWbn

boringContributor:
Hi @pernielsentikaer ,

jip I saw the comments thanks for the input 🙏🏽 However, this won't fix the use of default profiles. I'm pretty busy this week and i'll have a look this weekend though :)

PR #4988, opened 2023-02-26
Spotify Player: Add "My Playlists" functionality

dfangshuo:
## Description

Users wishing to use playlists-related functionality currently have 2 (imo unideal) options:
1. Search through a list of featured playlists
2. Search through all playlists

As a user, I find myself wanting to see a list of my own playlists, and have the ability to search through (only) that list. This PR introduces that behavior

## Checklist

```
$ npm install && npm run build
up to date, audited 381 packages in 442ms

50 packages are looking for funding
  run `npm fund` for details

10 vulnerabilities (9 moderate, 1 high)

To address issues that do not require attention, run:
  npm audit fix

Some issues need review, and may require choosing
a different dependency.

Run `npm audit` for details.

> build
> ray build -e dist

info  -  entry points [src/justPlay.ts src/nowPlaying.tsx src/nowPlayingMenuBar.tsx src/playSimilar.ts src/searchTracks.tsx src/searchAlbums.tsx src/searchPlaylists.tsx src/searchArtists.tsx src/myPlaylists.tsx src/featuredPlaylists.tsx src/browseAll.tsx src/likeCurrentSong.ts src/dislikeCurrentSong.ts]
info  -  checked TypeScript
ready -  built extension successfully
```

### DIFF
--- a/extensions/spotify-player/src/spotify/client.ts
+++ b/extensions/spotify-player/src/spotify/client.ts
@@ -534,33 +534,27 @@ export function usePlaylistSearch(query: string | undefined): Response<SpotifyAp
   return response;
 }
 
-
-async function getUserPlaylistsPage(
-  offset: number, limit: number
-) {
+async function getUserPlaylistsPage(offset: number, limit: number) {
   return await spotifyApi
-      .getUserPlaylists({ offset: offset, limit: limit })
-      .then(
-        (response: { body: any }) => response.body as SpotifyApi.ListOfUsersPlaylistsResponse
-      )
+    .getUserPlaylists({ offset: offset, limit: limit })
+    .then((response: { body: any }) => response.body as SpotifyApi.ListOfUsersPlaylistsResponse);
 }
 
 async function getAllUserPlaylists(
-  offset:number, items: SpotifyApi.PlaylistObjectSimplified[]
+  offset: number,
+  items: SpotifyApi.PlaylistObjectSimplified[]
 ): Promise<SpotifyApi.PlaylistObjectSimplified[]> {
   const step = 50;
   const resp = await getUserPlaylistsPage(offset, step);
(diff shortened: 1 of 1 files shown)


pernielsentikaer:

@peduarte what do you think about merging the three playlist commands into one, would that make sense (you already did something)

peduarte:
Hey 👋 Thanks for your contribution!

Once we introduce "My Playlists", users will want "My artists", "My Albums", "My podcasts", etc... I think we should avoid having granular commands like this. We currently have this issue with "Search Artists", "Search Albums", "Search Tracks", "Search Playlists". The high number of commands starts to become overwhelming, and we've received a few users feedback about this.

We're in the process of refactoring this extension and re-thinking its entire UX. A lot of work has been done internally already.

We'll have a **"Your Library" Command**, which will list all your Playlists, Artists and Albums. It includes a search filter so you can then filter it by a specific category. 

I already started chatting with @mattisssa about some of these changes, as he's the original author.

Unfortunately, for the reasons mentioned, we won't be able to accept your PR, but we hope that you'll enjoy the upcoming improvements to the Spotify Extension!

Thanks!

dfangshuo:
Sounds like there's a bunch of exciting changes underway. Is there an estimate of when these changes will go live?

You all have probably heard a bunch of feedback regarding this, but just to chime in w my 2c: I think the biggest thing I was looking for with a spotify extension was functionality to search through "my" things - i didnt care so much about featured playlists, or all artists, as much as the one that I had "saved". I would imagine many other users would share such inclinations (but this is a reach here), and I really hope the new functionality would show that.

Additionally, I personally really appreciate the UI display a list and having input filter them down, which is what the current playlist search lacks, and something I would love to see in the new version

peduarte:
Thanks for the feedback! Taking it into account 👍

Issue #3948, opened 2022-12-18
[Quit Applications] ...


mattie-g:
### Extension

https://www.raycast.com/mackopes/quit-applications

### Description

The Icon of the application, inline before the name.

### Who will benefit from this feature?

I imagine the majority, it would help with quicker ID of any apps to quit.

mackopes:
Just for visibility: this is a duplicate of the [Issue 4007](https://github.com/raycast/extensions/issues/4007). 
I will look into this in the upcoming weeks when I'll have some spare time

mackopes:
@mattie-g I merged a PR adding this functionality. feel free to check it out and close this task if you feel like this issue has been resolved :)

PR #4995, opened 2023-02-26
[Quit-Applications] Add app icons

mackopes:
## Description

I added the app icon for every app in the list (see updated screenshots).
I've already seen a couple of is issues raised requesting this feature (e.g. https://github.com/raycast/extensions/issues/3948).

I also added "Quit" word in front of every app as it is suddenly too similar to the app launcher and can be easily confused when you are trying to open an app but you are in the Quit Applications extension. 

## Screencast

See updated screenshots in this PR

## Checklist


pernielsentikaer:
Hi @mackopes 👋 

I don't think it makes sense to have `quit` in front since it's the primary action; also, if you use the secondary restart action, it's the wrong context - could we leave the word `quit` from the title?

We could also add a preference to let the user control if they want a `confirmAlert` before quitting/restarting.

mackopes:
@pernielsentikaer Fair point you had. I removed "Quit" from the title. It indeed doesn't make sense with the secondary restart action!

I'll consider adding the preference to ask for the confirmation in a separate PR if that's ok.

pernielsentikaer (approved):
Hi 👋

Approved; let's 🚢 it!

PR #5014, opened 2023-02-27
[npm] Improve startup speed

TimonVS:
## Description

Improves (perceived) startup speed of npm extension by skipping initial fetch (which otherwise results in an empty list), this results in the loader being visible for a shorter period of time.

## Checklist

### DIFF
--- a/extensions/search-npm/src/index.tsx
+++ b/extensions/search-npm/src/index.tsx
@@ -11,6 +11,7 @@ export default function PackageList() {
       /\s/g,
       '+',
     )}`,
+    { execute: searchTerm.trim() !== '' },
   )
 
   return (
--- a/extensions/search-npm/package-lock.json
+++ b/extensions/search-npm/package-lock.json
@@ -7,8 +7,8 @@
       "name": "search-npm",
       "license": "MIT",
       "dependencies": {
-        "@raycast/api": "^1.25.0",
-        "@raycast/utils": "^1.4.8",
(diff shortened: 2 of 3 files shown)


TimonVS:
For some added context, the loader that is initially visible puts up a mental block for me so that I cannot type until the loader is gone. This improvement helps, but the loader is still visible for a few moments, is that because the extension is still initializing/rendering/...?

CleanShot 2023-02-27 at 15 14 09@2x

extensions/search-npm/src/index.tsx
```diff
@@ -6,16 +6,18 @@ import { NpmSearchFetchResponse } from './npmResponse.model'
...
       /\s/g,
       '+',
     )}`,
+    { execute: shouldFetch },
   )
 
   return (
     <List
-      isLoading={isLoading}
+      isLoading={shouldFetch && isLoading}
```
TimonVS, line 20: `isLoading` is set to `true` despite `execute` being set to `false`, it's no big deal, but I think `isLoading` should be `false` when `execute` is set to `false`? @mathieudutour 

(As a side note, I can't help but wonder: why are `@raycast/api` and `@raycast/utils` private?)

TimonVS: That's fair! Thanks, looking forward to it :)

TimonVS: Wonderful! I updated the PR to incorporate this change :)

PR #5028, opened 2023-02-28
add get ssh key

tc9011:
## Description

get your local SSH key quickly

## Screencast

https://user-images.githubusercontent.com/18380140/221825412-6a71a5a8-858d-477f-a11e-bb78463a52b5.mov

## Checklist

### DIFF
--- a/extensions/get-ssh-key/CHANGELOG.md
+++ b/extensions/get-ssh-key/CHANGELOG.md
@@ -1,3 +1,3 @@
 # Get SSH Key Changelog
 
-## [Initial Version] - 2023-02-27
+## [Initial Version] - 2023-03-03


pernielsentikaer:
Potentially this will `cat` multiple keys, is this as intended?

tc9011:
hi, @pernielsentikaer, thanks for reviewing. As this article says ( [Checking for existing SSH keys - GitHub Docs](https://docs.github.com/en/authentication/connecting-to-github-with-ssh/checking-for-existing-ssh-keys)), there are three kinds of public SSH key names,  so I changed `cat ~/.ssh/id_rsa.pub` to `cat ~/.ssh/id_*.pub`.  The `*` is used for compatibility with the key name that may appear later.

pernielsentikaer:
But the extension is called `Get SSH Key`, which might give you a lot of keys `cat'ed` into one string?

tc9011:
Oh, I know your point, and most developers may only have one public key in their local computer, and for those who have multi kinds of public keys, how about just copying the first `.pub` file?  like `cat $(find ~/.ssh/ -maxdepth 1 -name "*.pub" -print -quit) | pbcopy`

pernielsentikaer:
One crazy idea could also be if there is more than one, show them in a list and let the user pick. 🚀

tc9011:
Great idea!

pernielsentikaer:
Hi @tc9011 👋 

What do you think about this einhancement?

tc9011:
wow! It looks great! Thanks for helping me enhance this extension!

pernielsentikaer:
I'm glad you like the enhancement @tc9011 🔥

pernielsentikaer (approved):
Hello 👋

PR #5049, opened 2023-02-28
Toggl Track: Display total tracked time for current day

michaelfaisst:
## Description

Calculate the total time tracked for the current day and display it in the main windows navigation title.

CleanShot 2023-02-28 at 21 41 24@2x

Time entries for the current day only include entries that also start on the current day, which is from what I can tell the same logic as Toggl does it as well.

Also updated the `@raycast/api` package to be able to use `npx ray publish`, didn't run into any issues but I can also revert back and do a manual PR.

## Checklist

### DIFF
--- a/extensions/toggl-track/src/index.tsx
+++ b/extensions/toggl-track/src/index.tsx
@@ -25,6 +25,21 @@ function ListView() {
       [] as TimeEntry[]
     );
 
+  const totalDurationToday =
+    timeEntries
+      .filter((timeEntry) => dayjs(timeEntry.start).isSame(dayjs(), "day"))
+      .reduce((acc, timeEntry) => acc + timeEntry.duration, 0) +
+    (runningTimeEntry ? dayjs().diff(runningTimeEntry.start, "second") : 0);
+
+  function formatSeconds(seconds: number) {
+    const h = Math.floor(seconds / 3600);
+    seconds %= 3600;
+    const m = Math.floor(seconds / 60);
+    const s = seconds % 60;
+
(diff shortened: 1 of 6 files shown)


extensions/toggl-track/src/index.tsx
```diff
@@ -25,6 +25,21 @@ function ListView() {
       [] as TimeEntry[]
     );
 
+  const totalDurationToday =
+    timeEntries
+      .filter((timeEntry) => dayjs(timeEntry.start).isSame(dayjs(), "day"))
```
rogatty, line 30: This is a tricky one, isn't it? Current approach can show today's duration longer than today is, e.g. a couple of hours at 00:01.

I'd say it's good enough, and the complexity of splitting it at midnight might be not worth the trouble.

rogatty (approved):
I gave it a spin and seems to be working just fine, looks good to me.

extensions/toggl-track/src/index.tsx
michaelfaisst: Actually the running entry is always added, even if didn't start today 🤔 Maybe would be better to also check for the start date there as well, what do you think?

franzwilhelm (approved):
🙌

pernielsentikaer (approved):
Hello 👋

extensions/toggl-track/src/index.tsx
rogatty: I think it depends on the specific use case but both approaches are fine.

Issue #4577, opened 2023-02-03
[Timeular] Show total time completed


BennEntterprise:
### Extension

https://github.com/raycast/extensions/tree/main/extensions/timeular

### Description

# As a USER I want TO SEE MY TOTAL TRACKED TIME  so that I CAN KNOW WHEN I AM NEARING THE END OF MY WORK DAY. 

Currently the Status shows a summation of each category, but does not show the total of all completed trackings.

The `status` command should prominently show the cumulative total of all tasks.

### Who will benefit from this feature?

This feature will be useful for anyone who want's to keep track of their total hours. 

I work an 8 hour day. Once I get to 8 hours I try to ease up and be with my family, go for a walk etc. 

Others may have similar quotas they are trying to hit per day.

### Anything else?

The Timeular app UI has some summarized stats by day, and for the entire week. 

<img width="821" alt="image" src="https://user-images.githubusercontent.com/29560043/216705527-2f254ead-0b36-4b92-9799-69efec618c41.png">

This feature will not go THAT far, but instead just show the daily total. If a user wants to see totals over the week they can use the full featured app. For now I think this feature gives the MVP needed to not break focus.

BennEntterprise:
I am actively  working on this myself and will have a PR sometime this weekend.

PR #5056, opened 2023-03-01
Timeular total daily time

BennEntterprise:
This will close [Issue 4577](https://github.com/raycast/extensions/issues/4577)

## Description

I added a way to view the daily total of time accumulated by the items in the report AND any active timers. 

## Screencast

Status Command when there are not tasks being actively tracked.
Status Command when there are not tasks being actively tracked.

Status Command when there is a task that is being tracked. 
Status Command when there is a task that is being tracked.

## Checklist

### DIFF
--- a/extensions/timeular/CHANGELOG.md
+++ b/extensions/timeular/CHANGELOG.md
@@ -1,10 +1,10 @@
 # Timeular Changelog
 
-## [Update "Edit Note" View] - 2023-02-03 @BennEntterprise
+## [Update "Edit Note" View] - 2023-02-03
 
 - The Status command will now show how much time has been tracked today.
 
-## [Update "Edit Note" View] - 2023-02-03 @BennEntterprise
+## [Update "Edit Note" View] - 2023-02-03
 
 - The "edit" view now supports a textarea, to better capture long form notes in a more friendly way
 


BennEntterprise:
I failed the first build test because of package-lock.json file so I deleted that to see what happens. 🤷‍♂️

If it fails again I'll take a look tomorrow.

kudrykv:
@BennEntterprise `package-lock.json` should be committed. If there are any weird dependency issues too cumbersome to resolve manually, you can delete the file, and after that, you should recreate it (e.g., `npm i`) and commit it back again.

extensions/timeular/src/useTodayReport.ts
```diff
@@ -23,18 +23,37 @@ export const useTodayReport = (tracking: Tracking | null, activity?: Activity) =
...
     }
+    const ms = Date.now() - new Date(tracking?.startedAt + "Z").getTime();
 
     const markStart = Date.now();
 
+    const totalBookedTime = entries.reduce((acc, val) => {
+      if (!val.duration.stoppedAt || !val.duration.startedAt) {
+        // Skip entries that are still running (no stoppedAt)
+        return acc;
+      }
```
kudrykv, line 34: In which scenarios `val.duration.startedAt` isn't there?

extensions/timeular/src/useTodayReport.ts
```diff
@@ -23,18 +23,37 @@ export const useTodayReport = (tracking: Tracking | null, activity?: Activity) =
...
+
+      const [startedAt, stoppedAt] = [
+        new Date(val.duration.startedAt + "Z").getTime(),
+        new Date(val.duration.stoppedAt + "Z").getTime(),
+      ];
```
kudrykv, line 39: Let's make it a tad simpler
```ts
const startedAt = new Date(val.duration.startedAt + "Z").getTime();
const stoppedAt = new Date(val.duration.stoppedAt + "Z").getTime();

```

extensions/timeular/src/useTodayReport.ts
```diff
@@ -23,18 +23,37 @@ export const useTodayReport = (tracking: Tracking | null, activity?: Activity) =
...
       .then(entries => entries.map(entry => `* ${entry.activity.name} for ${humanizeDuration(entry.duration)}`))
       .then(list => list.join("\n\n"))
       .then(markdown =>
         markdown
-          ? setReportMarkdown("Time entries you have tracked today:\n" + markdown)
+          ? setReportMarkdown(
+              `Time entries you have tracked today (including current)[**${humanizeDuration(
+                totalBookedTime + (ms || 0)
+              )}**]:\n` + markdown
+            )
```
kudrykv, line 56: I would suggest converting this syntax into a basic anonymous function (`() => {}`), extracting things to consts, etc., to make this thing easier to read&comprehend.

Also, what do you think about this word change: switch from `[HH:mm]` to `... is HH:mm`, so it reads as:
```
Time entries you have tracked today (including current) is 3h:14m: # time can be bold if
- ...                                                              # that works better
                                                                   # for you
```

BennEntterprise:
All great suggestions I’ll address them in the coming days.  Thanks for
taking a look.

extensions/timeular/src/useTodayReport.ts
BennEntterprise: There probably is not a scenario where `val.duration.startedAt` would be missing. I supposed I just threw it in there __in case__ 🤷🏻‍♂️. 

I think it can probably be removed. 

We should still leave the `val.duration.stoppedAt`. We need to ensure that it exists, otherwise  it is the currently running task.

extensions/timeular/src/useTodayReport.ts
BennEntterprise: Yeah you're right. no need to get fancy haha. I'll split them out.

extensions/timeular/src/useTodayReport.ts
BennEntterprise: I like that suggestion. I have extracted it to `buildHeadlineWithTotal`. I utilized a useCallback. 

I like the content suggestion as well.

BennEntterprise:
I have run both `npm run lint` and `npm run build` locally and both exit cleanly... 

Going to think on this for a bit. Open to suggestions. I'm thinking that perhaps I need to make sure my global preferences aren't using the wrong linter and overriding the workspace...

image

BennEntterprise:
@kudrykv I made the changes you suggested, and was able to run prettier `npx prettier . --fix` to fix my build issue. Am I ok to squash/merge this ?

kudrykv:
@BennEntterprise, you'll need to ping someone from the Raycast team, they ultimately have the rights to merge.

Oh also, you can add yourself to contributors (it's somewhere in the `package.json` file).

@pernielsentikaer can you please review the PR?

kudrykv (approved):

pernielsentikaer (approved):
Hi 👋

Issue #5050, opened 2023-02-28
[Font Awesome] Add Copy Glyph action


gtg922r:
### Extension

https://www.raycast.com/dutzi/font-awesome

### Description

Request: Please add "Copy Glyph" as a user action for a specific Font Awesome symbol. 

Current actions are "Copy SVG" and "Copy Slug". It would be great to be able to copy the font glyph itself for use with the Font Awesome font.

Action should work the same as the "Copy Glyph" action on the Font Awesome webpage. For example: [Calendar Classic Solid Icon | Font Awesome](https://fontawesome.com/icons/calendar?s=solid&f=classic) gives the options to quickly: Copy unicode code, copy glyph, copy svg, download. 

### Who will benefit from this feature?

Anyone who uses Font Awesome ttf fonts to incorporate Font Awesome fonts (for example in text-based notes)

gtg922r:
Implemented in #5065

PR #5061, opened 2023-03-01
Add font-awesome extension

gtg922r:
## Description

Add a new action to copy the font awesome glyph to the clipboard (#5050)

## Screencast

https://user-images.githubusercontent.com/1195174/222089887-16f4f733-2d7f-4061-bd9c-99db99b71735.mov

## Checklist


gtg922r:
Not quite sure what I did wrong to get the label "new extension", this is supposed to be an update to the existing fontawesome extension

gtg922r:
Looks like `npx ray publish` uses the `name` of the extension as set in `package.json` to determine the folder. Which in this case is not the correct folder name. Will manually re-submit

PR #5119, opened 2023-03-03
Add aws-profiles extension

frese:
## Description

AWS Profiles lists all your AWS profiles from $HOME/.aws/config and let you chose one to execute a predefined command/script. This allows you to eg. open a terminal authenticating using this profile or (the default) use granted.dev assumes command to open an AWS console with selected profile.

## Checklist

### DIFF
--- a/extensions/aws-profiles/README.md
+++ b/extensions/aws-profiles/README.md
@@ -9,14 +9,14 @@ You will probably need to write a script or prepare a command that you will like
 
 Examples :
 
-- Open AWS console using https://granted.dev/ _assume_ command (default)
+### Open AWS console using https://granted.dev/ _assume_ command (default)
 
-     /bin/zsh -l -c "assume -c <profile>"
+    /bin/zsh -l -c "assume -c <profile>"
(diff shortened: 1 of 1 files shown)


pernielsentikaer:
Would this make sense to have under the [AWS extension](https://www.raycast.com/Falcon/aws)?

pernielsentikaer:
CC @JonathanWbn 😊

frese:
Not sure about this - my extension is very different in nature - the AWS profiles is the key of the search here.
So far you can have only one "command" to execute on the searched profile, but it would be relatively easy to extend to multiple commands, but the profile would still be the key element. I guess my extension targets people with a lot of profiles.

JonathanWbn:
@frese I love this idea! I could use that myself at my place of work. 🎉 

I **do** think this could work well as part of the existing Amazon AWS extension. I opened #5357 to show I think we could include it. Let me know what you think!

pernielsentikaer:
Let's close this for now, we can always open it again

Issue #5101, opened 2023-03-02
[Logitech Litra] ... More commands


zalewskigrzegorz:
### Extension

https://www.raycast.com/timrogers/logitech-litra

### Description

First! 
Awesome work.

It will be great to have more commands available to use
litra-brightness-lm
or 
litra-temperature-k

### Who will benefit from this feature?

Everyone. :D

timrogers:
@zalewskigrzegorz Thanks for reaching out! Do you have any thoughts on how you'd like the interface to work?

zalewskigrzegorz:
Personally, I will prefer command to increase or decrease bright by 10% and similar for a color changes. But maybe it will be quicker to use some presets what user can configure in a menu.

PR #5144, opened 2023-03-04
[Logitech Litra] Allow managing Litra devices' temperature and brightness using presets

timrogers:
With this update, you can now set the brightness and temperature of your Litra devices from Raycast - thanks to [@zalewskigrzegorz](https://github.com/zalewskigrzegorz) for the [suggestion](https://github.com/raycast/extensions/issues/5101).

It isn't possible to get the current brightness or temperature, so we can't have a nice "increase by 10%" or "decrease by 10%" user experience. Instead, you can configure presets using the new "Manage Temperature Presets" and "Manage Brightness Presets" commands, and then those presets are available for each device in the "Manage Devices" command.

Closes https://github.com/raycast/extensions/issues/5101.

### DIFF
--- a/extensions/logitech-litra/CHANGELOG.md
+++ b/extensions/logitech-litra/CHANGELOG.md
@@ -1,3 +1,7 @@
 # Logitech Litra Changelog
 
+## [Brightness and temperature support] - 2023-03-04
+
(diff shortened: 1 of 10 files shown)


timrogers:
FYI @zalewskigrzegorz

pernielsentikaer (approved):
Hello 👋

zalewskigrzegorz (commented):
Works great Thanks

Issue #4921, opened 2023-02-22
[Arc] Add command for New Little Arc Window


sdalmonte:
### Extension

https://www.raycast.com/the-browser-company/arc

### Description

Adding the New Little Arc Window command to Raycast would let users open a page in Little Arc without having to switch to their main Arc window.

### Who will benefit from this feature?

all users of the Arc extension

thomaspaulmann:
Need to see if that's possible. Want to give it a try yourself?

PR #5191, opened 2023-03-07
Add new little arc command

AiroPi:
## Description

#4921 
Add a new command to open little arc.
Don't know if it is possible to add an argument with an URL ? Currently it will display the default page, same as a new tab.

## Checklist

- [ x I ran `npm run build` and [tested this distribution build in Raycast](https://developers.raycast.com/basics/prepare-an-extension-for-store#metadata-and-configuration)

### DIFF
--- a/extensions/arc/CHANGELOG.md
+++ b/extensions/arc/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Arc Changelog
 
-## [Improvements] - 2023-03-07
+## [Improvements] - 2023-03-08
 
 - Add a new command to open little arc.
 


pernielsentikaer:
Hi @AiroPi 👋 

I updated the code a bit, now you can pass an URL as an argument, if not, use the preference

Could you give it a quick spin 😊

AiroPi:
This argument works great, thanks!
But, if the url provided is something like "google.com" instead of "https://google.com/" it throw an error.
I think it is not an important problem though, and it could be improved in the futur.

But I have a question, by throwing an error, the error message appear in the console while being in debug mode. How will it act if used in "production" ? Will the error appear in the raycast window as with showHUD, or need I replace this error throw with showHUD ?

Also I'm new to the "contribution world of open source", so I don't know for who is the "contributors" key in package.json
Should I add myself or not?

Finally, I think an optional argument could be add to other commands such a "Open New Tab" and "Open New Window". Should I open a 2nd PR for this purpose or is it ok if I add them in this PR, while it is not the main goal of this PR ?

Thanks 😄

pernielsentikaer:
Could be something like this 😊

extensions/arc/package.json
```diff
@@ -33,6 +25,31 @@
...
+          "description": "The URL to open in the new tab",
+          "type": "textfield",
+          "required": false,
+          "default": "https://google.com"
+        }
+      ],
+      "arguments": [
+        {
+          "name": "url",
+          "placeholder": "Optional URL",
```
thomaspaulmann, line 47:
```suggestion
          "placeholder": "URL",
```

extensions/arc/src/types.ts
```diff
@@ -38,3 +38,7 @@ export type Suggestion = {
   query: string;
   url: string;
 };
+
+export type Arcguments = {
```
thomaspaulmann, line 42: Probably best to call this something like `NewLittleArcArguments` to make it more explicit but can be done with the next codechange.

thomaspaulmann:
I think we're good here, aren't we? This looks great. Thanks for your contribution @AiroPi - I leave the merging to @pernielsentikaer given that he pushed some little adjustments.

PR #5301, opened 2023-03-12
[Brew] Enhance search result sorting

aveplen:
## Description

I am fixing the issue of not showing search result that exactly matches query on the top of the list. 

Searching for `node` without fix:

Screenshot 2023-03-12 at 04 54 24

Searching for `node` with fix:

Screenshot 2023-03-12 at 04 54 36

## What I did

Added fast-sort library for faster search results sorting. 
Also I reordered those results so that:
1. Exact match is topmost
2. Results that include query in the beginning of the word have higher priority

## Checklist

_This is not a new extension, I don't think that checklist is related to this scenario, anyway:_

### DIFF
--- a/extensions/brew/package-lock.json
+++ b/extensions/brew/package-lock.json
@@ -10,6 +10,7 @@
         "@raycast/api": "^1.43.2",
         "@raycast/utils": "^1.4.11",
         "@types/stream-json": "^1.7.2",
+        "fast-sort": "^3.2.1",
         "node-fetch": "^2.6.7",
         "run-applescript": "^6.0.0",
         "stream-chain": "^2.2.5",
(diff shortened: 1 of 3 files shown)


nhojb (changes requested):
Thanks for the contribution. Code looks good 😄 

I've had a quick play and it definitely improves search results when there is a match on the formulae/cask name.

However, there are a couple of changes to search result order which I wonder if we can improve:

* The sort order changes for a partial name match compared to a full name match. This wasn't the case with the original code. For example compare the two top matches for "typescri" compared with "typescript".

Screenshot 2023-03-13 at 14 52 20
Screenshot 2023-03-13 at 14 52 22
Screenshot 2023-03-13 at 15 08 35



* When the search term partially matches the description the resulting sort order / ranking seems less predictable than before. Previously matches on description were sorted alphabetically by formulae/cask name, now it's hard to detect a pattern, which makes it more difficult to find the item you want without further refining the search term.

Screenshot 2023-03-13 at 14 53 13
Screenshot 2023-03-13 at 14 53 16

Screenshot 2023-03-13 at 14 51 14
Screenshot 2023-03-13 at 14 51 17

PR #5361, opened 2023-03-15
feat(Timers): choice of alert sound for custom timer

lokot0k:
## Description

Implementation of alert sound choice dropdown when creating/starting custom timer

## Checklist

### DIFF
--- a/extensions/timers/CHANGELOG.md
+++ b/extensions/timers/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Timers Changelog
 
+## [Enhancement] - 2023-03-15
+
+- Implementation of alert sound choice dropdown when creating/starting custom timer
+
 ## [Alert sound volume customization and pausable stopwatches] - 2023-03-10
 
 - Fix bug where timer completion notification would not send until after timer was dismissed when "Ring Continuously" setting was turned on


lokot0k:
@ThatNerdSquared, please check. Also, there is one bug in extension with these changes. If custom timer is created while it's still running,  this created timer will use default sound. I made this because I can't explicitly access selected sound from code. I have 2 solutions for this, but they're requiring some architect changes or just look bad:
1. Take name of sound from timer file, so it'll needed to write not just timer's name, but name of sound in timer file too.
2. Replace all of timers with custom timers, because in fact user can't create non-custom timer😅  
So, it's up to you. You can write which option you prefer or maybe suggest new solution and I'll implement it

ThatNerdSquared (commented):
Hey @lokot0k, thanks for contributing! The code looks good to me in general.

> Also, there is one bug in extension with these changes. If custom timer is created while it's still running, this created timer will use default sound.

I'm not entirely sure what you mean here - do you mean when an existing running timer is saved as a preset, the running version of it will still use the default sound? Other than that, the only problem I see is that there's no way to change the sound of a running timer or existing preset. You're right that this would require substantial changes though, both in terms of architecture and UI.

I think your changes are ready to ship - these problems don't seem like too big of a deal for now. I'm also in the middle of working on major changes to how the timers work entirely (in preparation for the Alarms feature), so I don't think it's worth investing in additional code to handle functionality that I am currently rewriting 😅 

If you think the flaws mentioned above are a blocker for merging this PR, I'd advise you to PR your changes to [ThatNerdSquared/timers-for-raycast](https://github.com/ThatNerdSquared/timers-for-raycast) instead so I can merge with my refactors before we ship. But, as mentioned before, I think your PR looks good.

Also, please add yourself to `contributors` in the `package.json`!

lokot0k:
Thanks for quick review!
> existing running timer is saved as a preset, the running version of it will still use the default sound

Almost that. Running version'll use selected sound, but saved preset of this timer will use default sound.
But it's not a big deal for the moment, I guess🤔 Will wait for the new version!
> add yourself to contributors in the package.json

With great pleasure!

pernielsentikaer:
Are we ready to merge or awaiting extra fixes?

ThatNerdSquared:
It's good to go @pernielsentikaer :)

PR #5384, opened 2023-03-15
Update Replicate Extension

cbh123:
## Description

Updates @KevinBatdorf's Replicate extension. Now users can:
- Run models on Replicate
- View prediction data
- Explore models / open replicate dashboard 

## Screencast

https://www.loom.com/share/7afb734f7a814829af8655ecb5470108

## Checklist

### DIFF
--- a/extensions/replicate/package.json
+++ b/extensions/replicate/package.json
@@ -4,7 +4,6 @@
   "description": "Replicate is a tool for creating and sharing machine learning models. This extension allows you to create and view your predictions in Raycast.",
   "$schema": "https://www.raycast.com/schemas/extension.json",
   "icon": "replicate.png",
-  "owner": "replicate",
   "author": "KevinBatdorf",
   "contributors": [
     "bfirsh",


zeke:
Replicate employee here. 👋🏼 

Took this for a spin and it works like a charm. Looks good to me!

cc @pernielsentikaer @mathieudutour

extensions/replicate/package.json
```diff
@@ -16,10 +17,10 @@
   "license": "MIT",
   "commands": [
     {
-      "name": "replicate",
-      "title": "View Predictions",
-      "subtitle": "Replicate",
-      "description": "View your Replicate predictions",
+      "name": "index",
```
pernielsentikaer, line 20: Could we NOT change this if it's not really important? When this is changed people loses their hotkey/aliases/ranking for the command

KevinBatdorf: @cbh123 Do you think this new feature should be a separate command? So users would have one command for viewing past predictions and another to create new ones?

cbh123: Sure! Just pushed a commit to keep the command consistent. I think having one command is good for now so we can get this merged in!

zeke:
Excited to see that land! 🙏🏼

pernielsentikaer (approved):
Hello 👋

I gave it a quick spin, works as expected, and approved 🔥

PR #5413, opened 2023-03-16
[Visual Studio Code - Project Manager] Fixed a problem with the remote path when there is a "." in the path

MarkusLanger:
## Description

Fixed a problem with the remote path when there is a "." in the path


pernielsentikaer (approved):
Hello 👋

Issue #5502, opened 2023-03-22
[Bear Notes] Bear 2.0 search node error


poycc:
### Extension

https://www.raycast.com/hmarr/bear

### Description

<img width="862" alt="image" src="https://user-images.githubusercontent.com/24654957/226816303-ce6b6c18-a0ff-432b-9232-6cfd4c79d382.png">

### Steps To Reproduce

1. install Bear 2.0 beta
2. search note

raffeyang:
I fixed this problem, this problem is caused by bear2 changing the database structure.since bear2 is currently in beta, I'm not going to commit code and merge branches for now. You can simply copy and clone my fork, then open the `extension/bear` and run `npm install && npx run develop`

[fork](https://github.com/raffeyang/extensions)


This is the code to fix bear2 error:

```javascript
const BEAR_DB_PATH =
  homedir() + "/Library/Group Containers/9K33E3U3T4.net.shinyfrog.bear/Application Data/database.sqlite";
const BEAR_EPOCH = 978307200; // Start of 2001 as a timestamp
const SEARCH_NOTES_QUERY = `
SELECT
  notes.ZUNIQUEIDENTIFIER AS id,
  notes.ZTITLE AS title,
  notes.ZTEXT AS text,
  notes.ZMODIFICATIONDATE AS modified_at,
  notes.ZCREATIONDATE AS created_at,
  group_concat(tags.ZTITLE) AS tags,
  notes.ZENCRYPTED AS encrypted
FROM
  ZSFNOTE AS notes
LEFT OUTER JOIN
  Z_5TAGS AS notes_to_tags ON notes.Z_PK = notes_to_tags.Z_5NOTES
LEFT OUTER JOIN
  ZSFNOTETAG AS tags ON notes_to_tags.Z_13TAGS = tags.Z_PK
WHERE
  -- When there is a query, filter the body by that query, otherwise
  -- ignore the query
  (
    lower(notes.ZTITLE) LIKE lower('%' || :query || '%')
    OR lower(notes.ZTEXT) LIKE lower('%' || :query || '%')
    OR :query = ''
    OR lower(notes.ZUNIQUEIDENTIFIER) LIKE lower('%' || :query || '%')
  )
  -- Ignore trashed, archived, and empty notes
  AND notes.ZARCHIVED = 0
  AND notes.ZTRASHED = 0
  AND (
    notes.ZTEXT IS NOT NULL
    OR notes.ZENCRYPTED = 1
  )
GROUP BY
[...]

Issue #5407, opened 2023-03-16
[Bear Notes] Raycast no longer working with updated Bear (Bear 2 - released today)


apg-dev:
### Extension

https://www.raycast.com/hmarr/bear

### Description

<img width="738" alt="Raycast-Bear" src="https://user-images.githubusercontent.com/55888411/225668832-27b24439-26a0-46c1-b9f3-fdd9997a955f.png">

### Steps To Reproduce

Update Bear with today's update from Shiny Frog. Try using Raycast to find notes.

### Expected Behaviour

I'm not sure how much work it will be to allow Raycast to work with the updated Bear, but unfortunately the upgrade to Bear 2 is one-way (can't go back). So we're stuck without Raycast access to Bear notes until compatibility can be achieved :-(.

apg-dev:
It's in Testflight for both macOS and iOS.

And sadly, I didn't realize it would break Raycast.

>>
Hey there! The time you knew was coming is now here: the Bear 2 beta is
ready for you *with sync* on Mac and iOS! The next section is really
important, so please read it:

[image: :warning:] *Once you go Bear 2 beta, you can’t go back*. This new
beta will migrate and upgrade your old Bear 1 database and use it from now
on. The /Application Data v2/ folder is no longer used. Be sure to *export
all new and edited notes since Bear 1*, then import them into this new beta.

lucaschultz:
I've got everything up and running locally. I'll submit a pull request tomorrow.

gacharles23:
Hello. Is there an estimate on when this plugin might be able to support Bear 2? Thank you.

lucaschultz:
Currently we're just waiting for the MR beeing merged 🙂

program247365:
Was fixed in the #5536 PR. @raycastbot close this issue

PR #5536, opened 2023-03-23
Add support for Bear 2 to the bear extension

lucaschultz:
## Description

I added support for the new Bear 2 (see #5407, #5502 and #5510) which is [currently in beta](https://beta.bear.app/t/the-bear-2-beta-is-here/9115). If the extension detects a certain error (missing table when reading the bear database) the extension shows an empty layout which prompts the user to select the the correct bear version in the extension preferences.

I also added a _Search Query_ argument to the _Search Notes_ command and migrated to the lates API version.

### The New Argument

Raycast 2023-03-23 at 15 40 28

### The Prompt To Select The Bear Version

bear-4

## Checklist

### DIFF
--- a/src/bear-db.ts
+++ b/src/bear-db.ts
@@ -1,8 +1,8 @@
-import { Cache, environment } from "@raycast/api";
+import { environment } from "@raycast/api";
 import { readFileSync } from "fs";
 import { homedir } from "os";
 import path from "path";
-import initSqlJs, { Database, ParamsObject, Statement } from "sql.js";
+import initSqlJs, { Database, ParamsObject } from "sql.js";
 import {
   SEARCH_BACKLINKS_V1,
   SEARCH_BACKLINKS_V2,
@@ -10,6 +10,7 @@ import {
   SEARCH_LINKS_V2,
   SEARCH_NOTES_V1,
   SEARCH_NOTES_V2,
+  TABLE_EXISTS,
 } from "./db-queries";
 
 export interface Note {
@@ -53,7 +54,6 @@ function formatTags(tags: string[]): string {
 
 export class BearDb {
   database: Database;
-  cache: Cache;
 
   static searchNotesQueries = {
     v1: SEARCH_NOTES_V1,
@@ -72,7 +72,6 @@ export class BearDb {
 
   constructor(database: Database) {
     this.database = database;
-    this.cache = new Cache();
   }
 
   close() {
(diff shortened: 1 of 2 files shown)


lucaschultz:
@hmarr Hmm hadn't thought about that to be honest. Should be possible by just running the V1 queries and switching to the V2 versions if the `no such table` error is thrown.

hmarr:
@lucaschultz you should be able to read the schema programmatically by querying the [schema table](https://www.sqlite.org/schematab.html). As long as there are schema differences (e.g. tables or columns that have changed) you could decide which query to use based on that.

lucaschultz:
Hmm since Bear 2 is a beta I think it would be safe to assume most people are still using V1. If we do something like this:

```ts
  getNotes(searchQuery: string): Note[] {
    let statement: Statement;
    try {
      const bearVersion = this.cache.get("bearVersion");
      statement = this.database.prepare(bearVersion === "2" ? SEARCH_NOTES_V2 : SEARCH_NOTES_V1);
    } catch (error) {
      if (error instanceof Error && error.message === "no such table: Z_7TAGS") {
        statement = this.database.prepare(SEARCH_NOTES_V2);
        this.cache.set("bearVersion", "2");
      } else if (error instanceof Error && error.message === "no such table: Z_5TAGS") {
        statement = this.database.prepare(SEARCH_NOTES_V1);
        this.cache.set("bearVersion", "1");
      } else throw error;
    }
    statement.bind({ ":query": searchQuery });
    const results: Note[] = [];
...
```

We avoid an extra query. Since we cache the last used Bear version using the Raycast cache the error should happen only once per version change. What do you think @hmarr?

hmarr:
That approach also looks fine, as long as the error messages are stable and predictable. We'll need to be slightly careful if we tweak the queries in the future—e.g. you could imagine that if we joined against another table, that table's name might appear in the error message rather than the tags table.

lucaschultz:
> That approach also looks good, as long as the error messages are stable and predictable. We'll need to be slightly careful if we tweak the queries in the future—e.g. you could imagine that if we joined against another table, that table's name might appear in the error message rather than the tags table.

That's a great point! I implemented the solution using a query to the `sqlite_master` table! 🙂

lucaschultz:
@raycastbot please open this issue. @pernielsentikaer if @hmarr agrees this is ready to merge.

pernielsentikaer:
Looks like I can't reopen it 😞

Raycast-0eMiEnPm@2x

ttuygun:
@pernielsentikaer can you try?
https://stackoverflow.com/questions/62479813/github-how-to-re-open-a-pull-request-when-the-branch-has-no-more-history-in-com

lucaschultz:
> Looks like I can't reopen it 😞
> 
> Raycast-0eMiEnPm@2x

Hmm, I force pushed from my local repository to `ext/bear` on my branch. Seems that doesn't work. What is the correct workflow for this?

I created a new fork of the extension, deleted my fork of the repository and copied the files over. However I still can't `npx ray publish`:

CleanShot 2023-04-01 at 11 48 01

lucaschultz:
Okay, I ended up creating a new fork, cloning the fork, copy and pasting my changes in the new clone, committing everything and creating a new pull request #5706 ... maybe someone of the Raycast team could explain what wen't wrong here? 

I read the in the GitHub UI that I could add further changes by pushing to the `ext/bear` branch of the `lucaschultz/raycast-extensions` fork which was created after forking the extension in Raycast. I added the fork as remote to the local repository and tried pushing my changes to the branch. Since that didn't work I force pushed. This immediately closed this pull request because of unrelated histories. What would have been the correct workflow?

pernielsentikaer:
Looks like GitHub was down at the moment, did you ever experience this again?

PR #5623, opened 2023-03-28
Add new Laravel shift extension

denniseilander:
## Description

Running [Laravel Shift ](https://laravelshift.com) makes upgrading Laravel applications a breeze.
If needed, there is a possibility to run the shifts locally instead of in the cloud, due to policies or agreements which prevent you from sharing code it in any way, which applies to me.

To run a Shift locally, a command with some parameters, a SHIFT_TOKEN and a SHIFT_CODE is needed.
It can be hard to remember all the possible shift codes and you always have to look up the shift code.

That's where this extension comes in: easily generating Laravel Shift commands for Docker

## Screencast
laravel-shift-1
laravel-shift-2
laravel-shift-6

## Checklist

### DIFF
--- a/extensions/laravel-shift/CHANGELOG.md
+++ b/extensions/laravel-shift/CHANGELOG.md
@@ -1,5 +1,5 @@
 # Laravel Shift Changelog
 
-## [Initial Version] - {PR_MERGE_DATE}
+## [Initial Version] - 2023-04-17
 
 - Initial release of Laravel Shift


pernielsentikaer (commented):
Hi 👋 

Why didn't you go with the toast instead of the hud? 😊

denniseilander:
Hi @pernielsentikaer,

Thanks for your reply.

At first I had the toast, while the goal of this command is just copying the command, I thought it would make more sense if you just close the form after copying.

But I can Imagine, people will make additions to the generated command (which is why I made the output field editable), so maybe a toast is more user friendly?

I can also made it configurable in the extension settings.

What do you think?

pernielsentikaer:
Thanks for the corrections, much appreciated ❤️‍🔥

Could you check my corrections and make sure I didn't break anything

denniseilander:
@pernielsentikaer many thanks for the corrections, everything still works fine! 😊

pernielsentikaer:
It still feels like the UX should be improved, like when it's showing the command and you focus the field (if you want to modify it) it just copies it and you'll have to blur and focus again to copy it

I'm not sure what the best way to do this is, what do you think?

denniseilander:
I agree with that. What if we change the behaviour of the `onChange` trigger on the TextArea to something like this?
```
// onChange={setShiftCommand}
onChange={(shiftCommand: string) => {
    setShiftCommand(shiftCommand);
    copyToClipboard(shiftCommand).then(() => showToast(Toast.Style.Success, "Command copied to clipboard"));
}}
```

This wil update the command while you're typing, and at the same time copying it to the clipboard.
It may need a timeout of a few ms to prevent a lot of trigger events?

https://user-images.githubusercontent.com/3907144/231433311-650c2827-08d4-4512-ac50-a4bedde62801.mov

denniseilander:
@pernielsentikaer So you mean that only the dropdown and textfield (project path) will trigger the text area being filled and copied to the clipboard. That means that the manual call for Generating Shift Command with the command+enter will be removed.

Wat to do if you manually change the contents of the textarea? Just let the user copy the contents manually?

pernielsentikaer:
Ideally, I would just make the action into a copy that will copy whatever is in the text field and make the two other fields update it 😊

Then you can both copy manually or use the action

denniseilander:
Ah great, thanks, I can build that I guess. I will try to fix this today (afternoon). If I have any questions I will let you know.

Thanks for so far and taking the time to improve the extension, I really appreciate that 😊

pernielsentikaer:
That would nice, and yes, let me know if you need anything at all, I'm here to help you out

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #5643, opened 2023-03-29
Create `Youtube Downloader` extension

vimtor:
## Description

This extension attempts to download YouTube videos from Raycast. It has been attempted twice (#138 and #173), but many issues arose. I think I've solved them all, but before adding new features, I want to make sure my tricks comply with the Raycast team.

For it to work I had to do 2 things:
1. Require that the user install `ffmpeg` using `brew install ffmpeg`. Maybe this could be installed by the extension but I will need some guidance on how to properly do this.
2. Use [`patch-package`](https://www.npmjs.com/package/patch-package) to instruct [`fluent-ffmpeg`](https://www.npmjs.com/package/fluent-ffmpeg) to use the correct library. This happened because (don't ask me why) the package was requesting a `lib-cov` library instead of `lib` because the `FLUENTFFMPEG_COV` is set to `true` on build and development.

## Screencast

https://user-images.githubusercontent.com/36263538/228520241-08c854bb-706a-4775-ad18-b2bbe45dd7a6.mp4

## Checklist

### DIFF
--- a/extensions/youtube-downloader/src/index.tsx
+++ b/extensions/youtube-downloader/src/index.tsx
@@ -1,10 +1,11 @@
-import { Action, ActionPanel, Clipboard, Form, Icon, popToRoot, showToast, Toast } from "@raycast/api";
+import { Action, ActionPanel, Clipboard, Detail, Form, Icon, showToast, Toast } from "@raycast/api";
 import ytdl, { videoFormat } from "ytdl-core";
-import { useEffect, useState } from "react";
+import React, { useEffect, useState } from "react";
 import { FormValidation, useForm } from "@raycast/utils";
 import prettyBytes from "pretty-bytes";
 import { downloadAudio, downloadVideo } from "./utils";
 import fs from "fs";
+import { execSync } from "child_process";
 
 type Values = {
   url: string;
@@ -16,6 +17,7 @@ export default function DownloadVideo() {
   const [loading, setLoading] = useState(false);
   const [title, setTitle] = useState("");
   const [formats, setFormats] = useState<videoFormat[]>([]);
+  const [, setError] = useState(true);
 
   const { handleSubmit, values, itemProps, setValue } = useForm<Values>({
     onSubmit: async (values) => {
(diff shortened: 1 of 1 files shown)


sxn:
hey @vimtor – nice idea! Regarding your questions:

> Require that the user install `ffmpeg` using `brew install ffmpeg`. Maybe this could be installed by the extension but I will need some guidance on how to properly do this.

You could check if the user has ffmpeg, and if they don't:
1. if they have brew installed, show the user the command, and maybe offer to install it for them
2. if they don't have brew installed, point them to installation instructions for ffmpeg (go to http://www.ffmpeg.org/download.html etc)

Would that work? 🤔 

> Use [patch-package](https://www.npmjs.com/package/patch-package) to instruct [fluent-ffmpeg](https://www.npmjs.com/package/fluent-ffmpeg) to use the correct library. This happened because (don't ask me why) the package was requesting a lib-cov library instead of lib because the FLUENTFFMPEG_COV is set to true on build and development.

It sounds weird that `FLUENTFFMPEG_COV` would be set during build. Do you know who's setting it? Ideally we wouldn't need to patch the library 😬 if there's absolutely no way to go around it, we could live with it as long as the library version is pinned in `package.json`.

vimtor:
hey @sxn (cool username btw)

To avoid complicating the extension code, I added the necessary documentation for installing `ffmpeg` and an appropriate error message if it is not installed.

Regarding `patch-package`, I'm not sure what is setting the environment variable; it could be something related to the operating system set by `ffmpeg`. I pinned the dependency as you suggested.

If you think it is ready we can publish the extension and see how to improve it in future versions :)

sxn:
hey @vimtor – thanks for the change, and sorry I'm only getting back to this PR now 😅
Regarding the change – I actually meant you should show somewhere inside the extension, something like [what this extension does](https://github.com/raycast/extensions/blob/revert-masscode-paste-on-enter/extensions/displayplacer/src/components/not-installed.tsx#L4-L24). That way the extension would acknowledge some prerequisite is missing, and offer to help the user fix the issue. 😺

vimtor:
Thanks for the feedback @sxn! I added the code you suggested and tried it to check it is working. `ffmpeg` is quite heavy so it takes up 2 minutes to complete the installation. I added some notes regarding that.

0xOsprey:
Looking forward to this one as I'm currently using a hacked together python script via https://github.com/ytdl-org/youtube-dl

vimtor:
> This issue has been automatically closed due to inactivity.
> 
> Feel free to comment in the thread when you're ready to continue working on it 🙂
> 
> You can also catch us in [Slack](https://www.raycast.com/community) if you want to discuss this.

😢

sxn:
hey @vimtor, very sorry that it took so long to get back 🥲 I finally managed to have a go and it works great! 🎉 :shipit:

PR #5646, opened 2023-03-29
Add token-list extension

regohiro:
## Description

Search cryptocurrency token contract addresses on multiple chains. Data is provided by [CoinGecko](https://www.coingecko.com/) and cached for 24 hours. 

## Screencast

https://user-images.githubusercontent.com/62818999/228423783-7175669a-10c5-4c8c-a284-6ee70a47faa1.mp4

## Checklist

### DIFF
--- a/extensions/token-list/src/search-token-address.tsx
+++ b/extensions/token-list/src/search-token-address.tsx
@@ -93,36 +93,40 @@ export default function Command() {
       onSearchTextChange={doSearch}
       isLoading={loading}
     >
-      {items.map((token, index) => (
-        <List.Item
-          key={index}
-          title={token.name}
-          subtitle={token.symbol.toUpperCase()}
-          accessories={[
-            ...Object.keys(token.platforms)
-              .slice(0, 3)
-              .map<List.Item.Accessory>((platform) => ({
-                tag: platform,
-              })),
-            {
-              text: `${Object.keys(token.platforms).length} chains`,
-            },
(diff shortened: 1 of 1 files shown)


sxn:
hi @regohiro – thanks for the extension! A couple of things:
- is this different from the existing [CoinGecko extension](https://www.raycast.com/destiner/coingecko)? if so, would it make sense to improve that extension instead?
- what about adding a custom [EmptyView](https://developers.raycast.com/api-reference/user-interface/list#list.emptyview) so users aren't greeted with just a plain empty screen?
- could be nice to use a [List.Item.Detail](https://developers.raycast.com/api-reference/user-interface/list#list.item.detail) in order to display the token addresses instead of navigating to a new page altogether

regohiro:
Hello 👋

- Although this extension uses the same Coingecko API, this is more focused on searching token addresses. The existing one is for getting token price and ranking. 
- Added!
Screenshot 2023-04-06 at 21 30 54

- That was my initial approach, but using a list view would allow users to easily filter networks using the search bar and copy the selected address using the keyboard shortcut.

sxn:
> Although this extension uses the same Coingecko API, this is more focused on searching token addresses. The existing one is for getting token price and ranking.

Thanks for the explanation & the changes. To me, it still sounds like this could be an improvement of the CoinGecko extension – either the current command, or a new command that focuses on searching token addresses. 😺

PR #5647, opened 2023-03-29
[Hue] Fixes and improvements

### LINKED ISSUE #4929: [Hue] Too many requests
### Extension

https://www.raycast.com/pindab0ter/hue

### Description

Even tho I managed to connect successfully after a while it throws this error, it usually work only the first time 

CleanShot 2023-02-23 at 11 34 31@2x

@pindab0ter 

### Steps To Reproduce

1. Open Raycast
2. Open one of the commands ( Control Groups, Lights, Set Scene ) 

### Current Behaviour

_No response_

### Expected Behaviour

_No response_

pindab0ter:
## Description

Fix #4929, update dependencies and improve code quality.

## Checklist

### DIFF
--- a/extensions/hue/src/controlGroups.tsx
+++ b/extensions/hue/src/controlGroups.tsx
@@ -10,6 +10,7 @@ import {
   turnGroupOff,
   turnGroupOn,
 } from "./lib/hue";
+import { Api } from "node-hue-api/dist/esm/api/Api";
 import { MutatePromise } from "@raycast/utils";
 import { CssColor, Group, Room, Scene, SendHueMessage } from "./lib/types";
 import { getIconForColor, getLightIcon } from "./lib/utils";
@@ -17,11 +18,11 @@ import { BRIGHTNESS_MAX, BRIGHTNESS_MIN, BRIGHTNESSES, COLOR_TEMP_MAX, COLOR_TEM
 import { hexToXy } from "./lib/colors";
 import ManageHueBridge from "./components/ManageHueBridge";
 import UnlinkAction from "./components/UnlinkAction";
-import Style = Toast.Style;
 import { useHue } from "./lib/useHue";
+import Style = Toast.Style;
 
 export default function Command() {
-  const { hueBridgeState, sendHueMessage, isLoading, groups, mutateGroups, scenes } = useHue();
+  const { hueBridgeState, sendHueMessage, apiPromise, isLoading, groups, mutateGroups, scenes } = useHue();
 
   const manageHueBridgeElement: JSX.Element | null = ManageHueBridge(hueBridgeState, sendHueMessage);
   if (manageHueBridgeElement !== null) return manageHueBridgeElement;
@@ -38,6 +39,7 @@ export default function Command() {
             const roomScenes = scenes.filter((scene: Scene) => scene.group == room.id);
             return (
               <Group
+                apiPromise={apiPromise}
                 key={room.id}
                 group={room}
                 mutateGroups={mutateGroups}
@@ -54,6 +56,7 @@ export default function Command() {
             const entertainmentAreaScenes = scenes.filter((scene: Scene) => scene.group == entertainmentArea.id);
             return (
               <Group
+                apiPromise={apiPromise}
                 key={entertainmentArea.id}
                 group={entertainmentArea}
                 mutateGroups={mutateGroups}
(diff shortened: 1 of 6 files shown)


pernielsentikaer:
Awesome update @pindab0ter 🔥 

I tested it and it works quite better than before 🔥

pernielsentikaer:
Do you want tot test it @thomaspaulmann @robertvanhoesel

pindab0ter:
> Awesome update @pindab0ter 🔥 
> 
> 
> 
> I tested it and it works quite better than before 🔥 

Glad to hear!

Can you or anyone give me pointers on how to best memoize/cache/reuse an [API instance](https://github.com/raycast/extensions/blob/3d66076dbab2adcfa0a52c5b39a69da3cd5969d7/extensions/hue/src/lib/getAuthenticatedApi.ts#L7)?

I'm not very well versed in React. Apparently `useMemo` is not an option in this context and storing the reference in a local variable does nothing since the variable doesn't live on.

thomaslombart:
Hey @pindab0ter, we have a few extensions that do that, like Linear. We have a function called [withLinearClient.tsx](https://github.com/raycast/extensions/blob/main/extensions/linear/src/helpers/withLinearClient.tsx) which takes care of initializing a client. It's then possible to share it using a variable that you get through the `getLinearClient` call. We also have another function called [View.tsx](https://github.com/raycast/extensions/blob/main/extensions/linear/src/components/View.tsx) that uses `withLinearClient` and is then used to wrap any commands using the Linear client. You can probably follow a similar approach here. Let me know if you have any further questions.

pindab0ter:
Hi @thomaslombart, thanks for pointing me to those resources. I gave it a good shake, however I got stuck with the following problem:

* Whether I can connect to the API depends on the credentials being present, which is determined by whether the Hue Bridge is set up, which is when the state machine reaches the `connected` state.
* There are three `useCachedPromise`s to "stale while revalidate" the lights, groups and scenes. Those can't be instantiated until there is a valid API reference to define them with (the `useCachedPromise` need to know what to do while it returns the cached data).
* It's not possible to wait to invoke the `useHue` hook until I have that reference, since that results in a `Error: Rendered more hooks than during the previous render.`.

The only way I see right now to solve this is to drop the "stale while revalidate" `useCachedPromise` and just show a loading Detail view until a connection is established. That takes about 0.5 - 1 second though, and that's something I want to avoid.

My question is: How do I combine a `useCachedPromise` with an API reference that doesn't exist (is `null` or `undefined`) the moment that the `useCachedPromise` is defined? Or do I not use `useCachedPromise` but some other caching mechanism?

pindab0ter:
I ended up connecting twice: 
* Once in the state machine to verify that we're successfully connected to a Hue Bridge
* Once for the rest of the extension, which is now reused throughout, meaning that every command usage no longer results in a new connection being established.

This could theoretically be less robust, since the connection could be lost in between establishing and the invocation of a command. However, I don't think this would happen often enough and is easily remedied by starting the command again.

pindab0ter:
In rewriting large parts of the extension, I found out the solution to my problem: `useCachedPromise` supports an `execute` configuration option that I set to whether the connection is defined or not.

That way, the hook isn't executed until the connection is set up. Perfect!

That won't make it into this PR, however. This works as it is. It's good to go on my end.

thomaslombart (commented):
Looks good to go on my end. @thomaspaulmann do you want to review and test it?

thomaslombart (approved):

thomaspaulmann (commented):
CleanShot 2023-04-06 at 2 26 30

I triet to authenticate and got the above failure toast. Pressing continue worked but the list of lights was empty. Exiting the command and re-entering then showed the lights. Can we fix this?

Instead of passing around the `apiPromise` everywhere, it might be cleaner to create some sort of client that gets the promise injected once, resolves it and then uses the resolved one. This way your code stays a bit cleaner. But up to you.

thomaspaulmann:
Given that this was merged, you can do a fix in another PR ;-)

pindab0ter:
> CleanShot 2023-04-06 at 2 26 30
> 
> I triet to authenticate and got the above failure toast. Pressing continue worked but the list of lights was empty. Exiting the command and re-entering then showed the lights. Can we fix this?
> 
> Instead of passing around the `apiPromise` everywhere, it might be cleaner to create some sort of client that gets the promise injected once, resolves it and then uses the resolved one. This way your code stays a bit cleaner. But up to you.

Thanks for the suggestion! Always happy to learn more about how to improve!

I'm basically completely rewriting the entire extension. One of the things I've done is use one single 'Hue Client' that I reuse everywhere, like you suggested.

I'll detail everything I've done in a new PR. If you're curious, check out https://github.com/pindab0ter/raycast-extensions/tree/feature/hue-v2-api.

PR #5673, opened 2023-03-30
[YouTube] Cache and Aspect Ratio

yug2005:
## Description

Major Bug Fix : Instead of storing objects in the cache (for recent and pinned items), the extension now stores the IDs. The objects are fetched using the ID when the command is launched. This ensures that the objects are up to date. For example, before this change, the view count of the pinned objects would not update. 

Other improvements : 
* Using cache instead of local storage
* Using a 4/3 aspect ratio for videos

I also refactored the code a lot but other than what was mentioned above, everything else works the same. 

## Screencast

youtube 2023-03-30 at 14 17 27

## Checklist

### DIFF
--- a/extensions/youtube/CHANGELOG.md
+++ b/extensions/youtube/CHANGELOG.md
@@ -1,5 +1,11 @@
 # YouTube Changelog
 
+## [Enhancements] - 2023-04-13
+
+- Optimized the experience
+- Using cache instead of local storage
+- Using a 4/3 aspect ratio for videos
+
 ## [Search From Root] - 2023-02-08
 - Add query parameter to Search Videos and Search Channels commands
 


extensions/youtube/src/components/popular_video.tsx
```diff
@@ -1,23 +1,38 @@
...
 import { getErrorMessage } from "../lib/utils";
 import { getPopularVideos, useRefresher, Video } from "../lib/youtubeapi";
 import { VideoItem } from "./video";
-import { ListOrGrid, getViewLayout, getGridItemSize } from "./listgrid";
-import { PinVideo } from "./recent_videos";
+import { ListOrGrid } from "./listgrid";
+import { Preferences } from "../lib/types";
+import { addPinnedVideo } from "./recent_videos";
+
+const { griditemsize } = getPreferenceValues<Preferences>();
```
tonka3000, line 9: What is the reason to call it outside of the function body?

extensions/youtube/package.json
```diff
@@ -80,19 +80,19 @@
       "type": "dropdown",
       "title": "Grid Item Size",
       "required": false,
-      "default": "medium",
+      "default": "5",
```
tonka3000, line 83: The downside here is that everyody who has already a non default have a wrong value, any maybe crash. Can you check that upfront

tonka3000 (commented):
Check the comments

extensions/youtube/src/components/popular_video.tsx
yug2005: No reason. I moved them inside the function calls.

extensions/youtube/package.json
yug2005: Not sure what you mean here. The extension will not crash but I believe that the preference will be reset to the default value (which is 5 now) for everyone. They will have to reselect their preference.

tonka3000: Yeah. I check the the code again. Ignore this comment.

tonka3000 (approved):

PR #5674, opened 2023-03-30
#4485 (tech): Fix render for corrupted OTPs and update deps

### LINKED ISSUE #4485: [Authy] Couldn't decrypt service tokens.
### Extension

https://www.raycast.com/guga4ka/authy

### Description

Hi,

I am trying to use the Authy search token plugin. The setup went well, and I got the approval process from Authy, and the plugin continues after the approval. Though when trying to find a token, it shows the "Couldn't decrypt service token. Seems like Backup Password is wrong.".

I have changed the password to make sure it is correct in both Authy itself as well as in the plugin. The error messages remain. I have even tried to use a less complex password that only exists out of letters and numbers to rule out that it might be an unusual character that causes the problem. 

I've also tried uninstalling and reinstalling the plugin, only to the same effect.

My other devices picked up that the password changed and asked to verify the backup password. After inputting the new password they sync with no problems.

Before Raycast I used Alfred, which also has an Authy plugin, which is using a command line tool. This tool connects to Authy as well and has no problems after setting up the account and using the same backup password...

I'm not sure what might be wrong, or how I might assist in hunting down the problem.

Gerard.

### Steps To Reproduce

_No response_

### Current Behaviour

Starting Raycast and triggering the search token action from the Authy plugin returns an error "Couldn't decrypt service tokens. Seems like Backup Password is wrong."

### Expected Behaviour

It should be able to decrypt the service tokens as I am sure the password is correct.

guga4ka:
## Description

fix #4485

## Checklist

### DIFF
--- a/extensions/authy/src/component/OtpList.tsx
+++ b/extensions/authy/src/component/OtpList.tsx
@@ -88,7 +88,7 @@ export function OtpList(props: { isLogin: boolean | undefined; setLogin: (login:
           issuer: i.issuer,
           logo: i.logo,
           digits: i.digits,
-          generate: () => seed ? generateTOTP(seed, { digits: i.digits, period: 30 }) : CORRUPTED,
+          generate: () => (seed ? generateTOTP(seed, { digits: i.digits, period: 30 }) : CORRUPTED),
         };
       });
       const apps: Otp[] = appsResponse.apps.map((i) => {


pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

guga4ka:
@pernielsentikaer thx for lint fix. 🍻

PR #5685, opened 2023-03-31
Slack status deeplinks

thomaspaulmann:
## Description

@sxn as discussed, this adds an option to show the form to create a status from a deeplink. Ideally, I could also just set the status via a deeplink without opening the form first. This could be super helpful for some automations.

I also migrated the API to the latest version.

## Checklist

### DIFF
--- a/extensions/slack-status/CHANGELOG.md
+++ b/extensions/slack-status/CHANGELOG.md
@@ -1,8 +1,8 @@
 # Slack Status Changelog
 
-## [Prefill Form from Launch Context] - 2023-04-06
+## [Add `Copy Deeplink` Preset Action] - 2023-04-06
 
-If the command is launched with `status`, `emoji` or `duration` context variables, it will directly show the form, prefilled.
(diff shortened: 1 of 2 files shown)


sxn:
thanks @thomaspaulmann!
I've pushed a bunch of changes as well:
- added a `Copy Deeplink` action
- when the command starts, checks if the command passed via context (if any) can be exactly mapped to a preset. If it can, the status is directly set

This allows directly setting a status via deeplink/command launching. It's not perfect though, as it introduces flickering:

https://user-images.githubusercontent.com/1155589/230378781-4025e3da-dddd-41f3-b1b7-d05c6594eb15.mov

PitNikola (approved):
LGTM

thomaspaulmann:
@sxn shall we merge it or hold off until we find a way to avoid the flickering?

sxn:
@thomaspaulmann I'd hold off until we either find how to avoid the flickering or come up with a better way of doing this.

Issue #5323, opened 2023-03-13
[Deepcast] Support more languages


9bow:
### Extension

https://www.raycast.com/mooxl/deepcast

### Description

Recently [DeepL announced that they're supporting some more languages like Korean and Norwegian](https://www.deepl.com/en/blog/welcome-korean-and-norwegian).

Actually, DeepL is now supporting 29 languages, but only Deepcast only supports 25 languages.

* [Languages currently DeepL supports](https://www.deepl.com/docs-api/translate-text/translate-text/) (from `source_lang` param)
```
BG - Bulgarian
CS - Czech
DA - Danish
DE - German
EL - Greek
EN - English
ES - Spanish
ET - Estonian
FI - Finnish
FR - French
HU - Hungarian
ID - Indonesian
IT - Italian
JA - Japanese
KO - Korean
LT - Lithuanian
LV - Latvian
NB - Norwegian (Bokmål)
NL - Dutch
PL - Polish
PT - Portuguese (all Portuguese varieties mixed)
RO - Romanian
RU - Russian
SK - Slovak
SL - Slovenian
SV - Swedish
TR - Turkish
UK - Ukrainian
ZH - Chinese
```

* [Languages currently Deepcast supports](https://github.com/raycast/extensions/blob/7025e822b8e0bda761c2af13b2e2b4ce6e00bb53/extensions/deepcast/src/utils.ts#L35-L59)
```
  BG: "Bulgarian",
  ZH: "Chinese",
  CS: "Czech",
  DA: "Danish",
  NL: "Dutch",
  EN: "English",
  ET: "Estonian",
  FI: "Finnish",
  FR: "French",
  DE: "German",
  EL: "Greek",
  HU: "Hungarian",
  IT: "Italian",
  JA: "Japanese",
  LV: "Latvian",
  LT: "Lithuanian",
  PL: "Polish",
  PT: "Portuguese",
  RO: "Romanian",
  RU: "Russian",
  SK: "Slovak",
  SL: "Slovenian",
  ES: "Spanish",
  SV: "Swedish",
  UK: "Ukrainian",
```

If you have a reason for excluding these languages, please let me know.
```
ID - Indonesian
KO - Korean
NB - Norwegian (Bokmål)
TR - Turkish
```

If not, I suggest adding them. :)

### Who will benefit from this feature?

Those who use unsupported languages - Indonesian, Korean, Norwegian, Turkish.

[...]

PR #5717, opened 2023-04-02
Update deepcast extension

MonsterDeveloper:
## Checklist

### DIFF
--- a/extensions/deepcast/CHANGELOG.md
+++ b/extensions/deepcast/CHANGELOG.md
@@ -1,0 +1,5 @@
+# Deepcast Changelog
+
+## [Enhancement] - 2023-04-14
+
+- Added Translate to Turkish, Korean, Indonesian, and Norwegian 


MonsterDeveloper (commented):
Fixed formatting issue.

MonsterDeveloper:
> Hey @MonsterDeveloper, thanks for the PR! Can you give us a description what you want to change and what these changes do?

Sure!

While using deepcast, I've encountered an error when translating text containing symbols, which should be URL-encoded (like %). However, the source code have shown that deepcast simply inserts the source text in the URL without URL-encoding it.

I decided to do a little refactoring (e.g. making one function to send a translation request to DeepL API) and change GET request with all params in the URL string to a proper POST utilizing a request body.

Thus code became a little bit more uniform and the bug with symbols containing text was fixed.

extensions/deepcast/src/utils.ts
```diff
@@ -5,21 +5,45 @@ interface Preferences {
   key: string;
   pro: boolean;
 }
-const translate = async (target: string) => {
-  const preferences: Preferences = getPreferenceValues();
-  const key = preferences.key;
-  const pro = preferences.pro;
+
+export function getPreferences() {
```
mooxl, line 9: Where is the point of defining it in an extra function when you use it only once?

extensions/deepcast/src/utils.ts
```diff
@@ -5,21 +5,45 @@ interface Preferences {
...
+    const text = initialText || (await getSelectedText());
+
+    const { key, pro } = getPreferences();
+
     try {
-      const response = await got(
-        `https://api${pro ? "" : "-free"}.deepl.com/v2/translate?auth_key=${key}&text=${text}&target_lang=${target}`
-      );
-      const translation = JSON.parse(response.body).translations[0].text;
+      const {
```
mooxl, line 28: If we do a rewrite of that logic, could we use https://developers.raycast.com/utilities/react-hooks/usefetch or classic fetch?

mooxl (commented):
Nice work! Left some comments:)
Maybe we could incorporate https://github.com/raycast/extensions/issues/5323 into this PR.

extensions/deepcast/src/utils.ts
MonsterDeveloper: Absolutely valid claim. My idea was to keep a unified function which would provide typed preferences, in case some other parts of deepcast would need to access them.

Since only 1 function utilizes preferences now, it may be redundant to keep a dedicated function for retrieving them. Up to you, whether to keep `getPreferences` or remove it.

extensions/deepcast/src/utils.ts
MonsterDeveloper: `useFetch` doesn't seem to be an option, as this function is called as a command as well (not in React component).

`fetch` also requires a third party library (e.g. `node-fetch`), as global fetch was deprecated (see [Raycast Changelog 1.46.0 - 2023-01-18](https://developers.raycast.com/changelog#1.46.0-2023-01-18))

pernielsentikaer:
@mooxl do you mind checking this again?

extensions/deepcast/src/utils.ts
mooxl: Ahh didn't saw that Raycast deprecated the experimental fetch in Node. We can switch to it after it's stable release.

extensions/deepcast/src/utils.ts
mooxl: Makes sense.

mooxl (approved):
LGTM!

pernielsentikaer (approved):

PR #5755, opened 2023-04-03
Update imessage-2fa extension

### LINKED ISSUE #5523: [iMessage 2FA] support code format 123-456
### Extension

https://www.raycast.com/yuercl/imessage-2fa

### Description

CashApp codes are sent like `123-456` 

It only seems to read the `123` part here.

image

arronhunt:
## Description

1. Resolves #5523 by making the check case-insensitive
2. Modified the logic to no longer use `if/else`. This is because sometimes we identify a match and only capture part of the code.

For 2. here's an example:

```
Your code: 123-456
```

This string satisfies the check 

```
// "code" OR "is" followed by an optional ":" + optional whitespace, then 3-8 consecutive digits
```

Even though we need to keep going and capture the entire string. 

## Screencast

N/A

## Checklist

### DIFF
--- a/extensions/imessage-2fa/CHANGELOG.md
+++ b/extensions/imessage-2fa/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Messages 2FA Changelog
 
+## [Update] - 2023-04-13
+
+- Resolves [iMessage 2FA] support code format 123-456 #5523 by making the check case-insensitive
+- Modified the logic to no longer use if/else. This is because sometimes we identify a match and only capture part of the code.
+
 ## [Update] - 2023-01-16
 
 - Add polling of 2FA codes


extensions/imessage-2fa/package.json
```diff
@@ -21,8 +22,18 @@
...
+  "preferences": [
+    {
+      "default": "1",
+      "description": "For searching iMessage from N days ago and now",
+      "name": "lookBackDays",
+      "required": false,
+      "title": "Search last days for 2FA",
+      "type": "textfield"
+    }
+  ],
```
arronhunt, line 34: The fork action automatically changed this. No idea why.

yuercl (approved):
Thanks for your commit.

extensions/imessage-2fa/package.json
pernielsentikaer: It just moved the position, should be fine

PR #5855, opened 2023-04-10
added new extension sun-moon-times

DannySeidel:
## Description

- Search for a city and display detailed info for the sun and moon
- Add and remove cities from your favorites list.
- Change the order of the cities in your favorites list.

## Screencast

sun-moon-times-1

sun-moon-times-2

## Checklist

### DIFF
--- a/extensions/sun-moon-times/src/components/CityListItemView.tsx
+++ b/extensions/sun-moon-times/src/components/CityListItemView.tsx
@@ -49,13 +49,13 @@ export const CityListItemView = ({
                     {isFavorite ? (
                         <>
                             <Action
-                                title="Remove From Favorites"
+                                title="Remove from Favorites"
                                 style={Action.Style.Destructive}
                                 icon={Icon.Trash}
                                 onAction={async () => {
                                     if (
                                         await confirmAlert({
-                                            title: "Remove From Favorites?",
+                                            title: "Remove from favorites?",
                                             message: `${countryList[countryCode].flag} ${name}`,
                                             primaryAction: { title: "Remove", style: Alert.ActionStyle.Destructive },
                                         })


extensions/sun-moon-times/assets/countryList.ts
sxn, line 1: kind of weird to see `.ts` files in the assets. Assets get [packaged so they can be referenced at runtime](https://developers.raycast.com/information/file-structure#assets), which you don't need in this case. As such, it's better to move them somewhere in `src/`. 😸

extensions/sun-moon-times/assets/ressourceUrl.ts
sxn, line 1: same as with `countryList.ts` above: it's better to move this file somewhere under `src/`, as it doesn't need to be available at runtime.

extensions/sun-moon-times/src/components/CityListView.tsx
```diff
@@ -0,0 +1,51 @@
...
+import { useFavorites } from "./FavoritesProvider"
+
+export const CityListView = () => {
+    const [searchText, setSearchText] = useState<string>("")
+    const { isLoading, data } = useFetch<ResponseData>(ressourceUrl(searchText))
+
+    const { favorites } = useFavorites()
+
+    return (
+        <List isLoading={isLoading} onSearchTextChange={setSearchText}>
```
sxn, line 16: search is kind of jumpy – what about using `throttle`?
```suggestion
        <List isLoading={isLoading} onSearchTextChange={setSearchText} trottle>
```

extensions/sun-moon-times/src/components/CityListItemView.tsx
```diff
@@ -0,0 +1,102 @@
...
+                },
+                { icon: Icon.Clock, text: dayDuration },
+            ]}
+            actions={
+                <ActionPanel>
+                    <Action
+                        title="Show More Info"
+                        icon={Icon.Info}
+                        onAction={() =>
+                            push(
```
sxn, line 57: how about using a [List.Item.Detail](https://developers.raycast.com/api-reference/user-interface/list#list.item.detail) instead of `Detail` here? 🤔

sxn (changes requested):
hey @DannySeidel – neat extension! I've left a couple of comments, let me know once you've had some time to go over them. 😺

extensions/sun-moon-times/src/components/CityListItemView.tsx
DannySeidel: @sxn I changed the Detail Info View to List.Item.Detail. I think this was a great idea because now you have all info in one view.

pernielsentikaer (approved):

PR #5866, opened 2023-04-10
[Arc] Improve new little Arc window usability

danpalmer:
## Description

This improves the general usability of the Arc browser `Open New Little Arc Window` command.

- By default, open `arc://newtab` instead of `https://google.com/`. Better for those who don't use Google search, better for non-US users.
- Separate the default and fallback, now resetting the setting doesn't require the user to put back in the default value themselves. Reduces edge cases/state management.
- Only add `https` when there is no protocol on the provided URL already.
- Allow non-HTTP(s) URLs to be entered.
- Allow URLs without a TLD to be entered.

## Checklist

### DIFF
--- a/extensions/arc/package.json
+++ b/extensions/arc/package.json
@@ -7,9 +7,19 @@
   "author": "thomas",
   "owner": "the-browser-company",
   "access": "public",
-  "categories": ["Web", "Applications", "Productivity"],
+  "categories": [
+    "Web",
+    "Applications",
+    "Productivity"
+  ],
   "license": "MIT",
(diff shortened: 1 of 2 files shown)


thomaspaulmann:
This is great @danpalmer! I didn't know about Arc's URL scheme for the new tab. How did you find that one out?!

danpalmer:
I was mostly aiming to remove the default page as I felt it wasn't the best UX. I knew Chrome used URLs for things like the new tab page, then saw that Arc redirects chrome:// to arc:// and pretty much supports the same URLs. Then lastly found a listing of the internal URLs at chrome://about, and chose the one that I felt had the best UX in Arc.

PR #5867, opened 2023-04-10
[Transmission] Usability improvements

danpalmer:
## Description

Use the password type for a field storing a password.

## Checklist

### DIFF
--- a/extensions/transmission/CHANGELOG.md
+++ b/extensions/transmission/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Transmission Changelog
 
+## [Usability improvements] - 2023-04-30
+
+- Use the password type for a field storing a password.
+
 ## [Error Handling and Pieces Graph] - 2022-03-25
 
 - Torrents with an error state are now displayed with a different icon and have their error visible on the details panel;


FezVrasta (approved):

extensions/transmission/src/utils/renderCells.tsx
```diff
@@ -1,6 +1,5 @@
 import { splitEvery } from "ramda";
 import BitField from "bitfield";
-import { renderToString } from "react-dom/server";
```
FezVrasta, line 3: personally I liked the renderToString approach more, less chances of messing up the resulting output. Any reason to change?

danpalmer: Yes, it wasn't compatible with the Raycast requirements anymore. Raycast requires cross-platform compatibility, but react-dom isn't fully cross-platform.

danpalmer:
@FezVrasta Ready for another look. I've bumped dependencies in order to fix a bunch of type-checking issues, the Raycast API is a bit more flexible on the types for example. I've made some UI tweaks as well, open to any feedback you have.

Screenshot 2023-04-10 at 21 39 04

Note with the above that the partial markdown field is necessary for rendering the progress SVG (nice one with that by the way!), and the divider can't be set at a custom height.

Screenshot 2023-04-10 at 22 36 49

And here, I've just pulled the instructions out of the placeholder so that the placeholders are more accurate. Alternatively the descriptions could be put in the `info` fields, but that requires a hover to reveal the details.

extensions/transmission/src/addNewTorrent.tsx
```diff
@@ -84,17 +88,20 @@ export default function AddNewTorrent() {
       <Form.TextField
         title="Magnet Link or Torrent File"
         id="input"
-        placeholder="magnet: (or ⌘ O to select from Finder)"
+        placeholder="magnet:..."
         value={input}
         onChange={(value) => setInput(value)}
       />
+      <Form.Description text="⌘ O to select from Finder" />
```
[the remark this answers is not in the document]
danpalmer: It does, but that would prevent magnet links from being used. 

It would be great to overhaul this form with the file picker OR a field for the magnet link, fix the clipboard detection, and maybe improve the UX and validation of the file path, but I didn't have time to do that yesterday. 

If this bothers me I might try to add it soon, but let's consider these changes as is for now and try to get a new release out if you're happy with them.

FezVrasta (changes requested):
Please revert the react-dom changes. renderToString is fully cross platform, it's designed to be used from node.js

danpalmer:
Strange, it was failing the CI check.

Is it just that this package was using an old version that was not cross platform?

FezVrasta:
@pernielsentikaer could you advise please?

grzegorzkrukowski:
Hey @FezVrasta @danpalmer 

We had to take a closer look at this issue - apparently, our check on RayCLI was a bit too aggressive, which was the main reason for the react DOM failing.

We are deploying a new CLI version right now and we would like to test it against your changes and make it work with the newer react-dom dependency. 

Do you mind "breaking it" agaig using the newest react-dom as it was originally, so we can make sure it works properly? :) Thanks in advance

grzegorzkrukowski:
Hey @FezVrasta @danpalmer 

We had to take a closer look at this issue - apparently, our check on RayCLI was too aggressive, which was the main reason for the react DOM failing. Sorry for the inconvenience.

We just deployed a new CLI version right now and we would like to test it against your changes and make it work with the newer react-dom dependency.
You can see [here](https://github.com/raycast/extensions/actions/runs/4660176036) it worked for the one that was failing.

Do you mind “breaking it” again and use the newest react-dom as originally, so we can ensure our fix works properly? :)

Thanks in advance

danpalmer:
@grzegorzkrukowski I just tried this and despite bumping versions for the Raycast CLI it didn't work, and actually failed earlier in the flow as I could no longer build successfully locally.

grzegorzkrukowski:
I see that checks are passing correctly now - what type of error do you get locally ?

danpalmer:
> I see that checks are passing correctly now - what type of error do you get locally ?

The checks are passing correctly because I backed out the change again. Locally I got the same error as I was seeing in CI before:

> build failed (node_modules/react-dom/server.js:3:25 module.exports = require('./server.node');): native bindings (e.g., .node files) are not supported because they are neither cross-platform nor cross-architecture; for example, native bindings built on a M1 mac would not work on an Intel mac

grzegorzkrukowski:
@danpalmer I think it may be related to the fact that you are using the CLI embedded into Raycast API.
The fix will be shipped to the official CLI on the next release, meanwhile, you can try downloading the version with the fix that is used on CI from [here](https://cli.raycast.com/1.50.0-alpha.21/arm64/ray).

If you don't want to mess up with this - just push the locally breaking version to the CI - and we can iterate on it if something will still be wrong.

danpalmer:
@grzegorzkrukowski re-pushed the changes restoring react-dom.

FezVrasta (approved):

extensions/transmission/src/modules/client.ts
```diff
@@ -127,3 +127,7 @@ export const useSessionStats = ({ refreshInterval = INTERVAL }: { refreshInterva
     { refreshInterval }
   );
 };
+
+export const isLocalTransmission = () => {
+  return preferences.host === "localhost" || preferences.host === "127.0.0.1";
```
FezVrasta, line 132: I think some users reported localhost wasn't working and they were required to use 127.0.0.1 (or vice versa?), it may make sense to enforce the correct format maybe. Not blocking for this PR of course.

danpalmer: Yeah that might be good, I think I had a similar issue. This particular function was just defensive coding to make sure that any expectedly local value will work, not necessarily sufficient validation on the input or anything like that.

pernielsentikaer: Should we wait for a push, or just continue with the merge?

FezVrasta: I would address that in a separate pr

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #5869, opened 2023-04-10
Add tempmail extension

Joshlucpoll:
## Description

A simple built-in temporary mailbox for Raycast using the [Mail.tm API](https://mail.tm).

### Features

- ✉️ Auto generate email addresses or create custom usernames
- ⏱️ Decide how often to auto-regenerate email addresses (from 5 minutes to never)
- 👀 View email natively in Raycast, in the Mail app or in the browser
- 📁 Download attachments

## Checklist


Joshlucpoll:
Hey @pernielsentikaer !

Thanks for taking a look at my pull request and for your feedback - I really appreciate it!

While I understand that there’s already an extension called Mailsy that uses the mail.tm API, I believe that TempMail offers some unique features and improvements that aren’t currently available in other temporary mail extensions. For example, with TempMail, users can download email attachments; view email details; and also set email addresses to automatically expire.

Regarding your suggestion to enhance the existing integration, I think that the differences in features and interface between Mailsy and TempMail are significant enough that it would require considerable time and effort to add TempMail features to Mailsy.

Again, thanks again for your feedback, and I hope you’ll consider merging my pull request!

danpalmer:
+1, I think an extension per temporary email service that people want to use is better than one extension for all temporary email service use-cases, and I think this would fit with precedent in other extensions.

BalliAsghar:
> +1, I think an extension per temporary email service that people want to use is better than one extension for all temporary email service use-cases, and I think this would fit with precedent in other extensions.

While offering separate extensions for different temporary email services may provide users with more options, it could also lead to clutter and confusion in the Raycast extensions. Maintaining multiple extensions for every temporary email service that users may want to use may not be sustainable in the long run. Consolidating similar functionalities into a single extension could lead to a more streamlined and user-friendly experience for users. Additionally, enhancing the existing [Mailsy](https://www.raycast.com/BalliAsghar/mailsy) extension could be a more efficient use of resources than creating a separate extension from scratch. It’s worth noting that for the average user, the specific temporary email service being used may not matter as much as the ability to quickly and easily generate a throwaway email address. Therefore, having multiple extensions for different temporary email services may not be necessary if they all provide similar basic functionality.

Joshlucpoll:
I'm not sure what the policy/guidance is for having similar extensions but there's already another temporary email extension in the store: [Temporary Email](https://www.raycast.com/natterstefan/temporary-email). So if similar extensions aren't allowed, I think should be said in the Raycast API docs.

I would disagree that enhancing Mailsy would be less effort compared to just accepting a pull request -- seeing as Mailsy's UI is built completely differently to TempMail it would require significant refactoring. 

I don't have the resources necessary to add TempMail's features to Mailsy -- therefore I won't be doing it. If I'm honest I didn't see the Mailsy extension until I finished TempMail -- however I think they're different enough to be separate extensions. TempMail is a more fully featured temporary mail extension, so I think users should get the best experience.

BalliAsghar:
The temporary email you mentioned is unique. It opens a just [maildrop](https://maildrop.cc/) in the browser. The noteworthy feature you added to view emails natively, which Mailsy does, but only in Markdown. This feature can be enhanced with the library you used, without requiring substantial refactoring. Features such as creating custom usernames, which TempMail provides, may not be necessary for a disposable email extension.

Ultimately, @pernielsentikaer has to make the definitive decision on whether to accept and integrate TempMail or enhance the existing Mailsy extension.

danpalmer:
Mailsy and Tempmail are different services with different feature sets, different terms of service, and potentially suitable for different users and use-cases. I think the fact the _services_ just so happen to do similar things is confusing things here more than is necessary.

I don't think anyone would debate adding Dropbox when Google Drive is already available, yet they fulfil largely the same purpose. See also: Apple Music and Spotify, GitHub and BitBucket, search engines, browsers, and so on.

I would strongly advocate for the Raycast policy to be one extension _per service_ (if any limitation at all), and not per feature, where curation and opinion becomes harder, and will lead to power users being neglected.

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #5881, opened 2023-04-11
[Update Extension] System Monitor - Fix Memory Stat

### LINKED ISSUE #5285: [System Monitor] Memory Stat is broken
### Extension

https://www.raycast.com/hossammourad/raycast-system-monitor

### Description

Stats show Memory at 0% (0gb) on a 16gb box with some memory still available. The "Free Ram" may fluctate between 0 and 1%, but is mostly 0. 

This may be a bug, or just an interpretation.  That is, Activity Monitor shows Physical 16, Used as 13.7 and Cached Files as 2.2.  If we assume that Used + cached is really real ram, then yeah, I have none.  But isn't a better solution to show 1-(Used/Physical)?  Cached Files memory is treated as available to apps, from what I understand 

### Steps To Reproduce

1) System Monitor is a "command" (not an extension?  Is there a diff?), install from Store
2) Run System Monitor, see CPU, Memory, Power, Network.  Look at live changing stats
3) My memory shows 0% consistently.

### Current Behaviour

Memory free appears to be 0%, but should probably be a larger number

### Expected Behaviour

Expected to see available ram % of my 16GB; 0 may be an artifact of the calculation used.

lucasmezencio:
## Description

Fix Memory Stat by using the same approach as `neofecth` ([here](https://github.com/dylanaraps/neofetch/blob/master/neofetch#L2706)).

This PR closes #5285.

## Screencast

N/A

## Checklist

### DIFF
--- a/extensions/system-monitor/src/Memory/MemoryMonitor.tsx
+++ b/extensions/system-monitor/src/Memory/MemoryMonitor.tsx
@@ -22,7 +22,6 @@ export default function MemoryMonitor() {
       .then((newTopProcess) => {
         getFreeDiskSpace()
           .then((newFreeDisk) => {
-                topProcess: newTopProcess,
             getMemoryUsage().then((memoryUsage) => {
               const memTotal: number = memoryUsage.memTotal;
               const memUsed: number = memoryUsage.memUsed;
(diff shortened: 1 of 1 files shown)


pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #5941, opened 2023-04-14
Add clear-download-folder extension

alcibiadesc:
## Checklist


thomaspaulmann:
Hey, great addition. I was wondering if we should add a new command to the [Downloads Manager extension](https://www.raycast.com/thomas/downloads-manager) instead of a new extension? I'm sure most users of the Downloads Manager extension would appreciate this new command.

As to the command itself, a few hints:
- Set the `mode` to `no-view` in `package.json`. Then we don't render a view and you can just have a simple JS function
- Show an alert before clearing the downloads folder.
- Instead of using `rm` to remove the files, use `trash` to move them to the trash. Than a user can recover it.

PR #5982, opened 2023-04-16
Add htmlquickstart extension

89801:
## Checklist


pernielsentikaer (changes requested):
Hi 👋 

Thanks for your contribution, I like the idea but maybe we need to work a bit more with it, here are some thoughts I had:

* Add so the user can add `templates` where it's possible to add the content you want in the files and let the user add as many as possible.
* Add so you can select a directory, maybe with a preference for a default location
* Missing an icon, you can make one here: https://icon.ray.so
* Missing [metadata images](https://developers.raycast.com/basics/prepare-an-extension-for-store#screenshots)

I don't think that in the current form, it makes a lot of sense to add to the store, do you want to extend the functionality to something bigger?

PR #5993, opened 2023-04-17
Update recent-downloads extension

alcibiadesc:
## Checklist

### DIFF
--- a/extensions/recent-downloads/src/index.tsx
+++ b/extensions/recent-downloads/src/index.tsx
@@ -88,6 +88,8 @@ export default function RecentDownloads() {
     }
   }
 
+  const hiddenFilesInfo = getHiddenFilesInfo();
+
   return (
     <List isShowingDetail searchBarPlaceholder="Filter files...">
       {downloads.map((download) => {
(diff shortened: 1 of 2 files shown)


pernielsentikaer:
Hi 👋 

How is this compared with [Download Manager](https://www.raycast.com/thomas/downloads-manager) - sounds like they have the same purpose?

alcibiadesc:
> Hi 👋
> 
> How is this compared with [Download Manager](https://www.raycast.com/thomas/downloads-manager) - sounds like they have the same purpose?

Hello @pernielsentikaer 

It is quite similar, the main difference is Grid with image preview instead of list  and some more direct shortcuts.

I am also looking at adding a document preview as with File Search and that the file search is location-agnostic, i.e. it shows all downloaded files.


image

pernielsentikaer:
Hi @alcibiadesc 👋 

It sounds like some nice updates for sure but it doesn't make sense to have two identical extensions with the same purpose, I would suggest that we enhance the current one build by @thomaspaulmann instead

alcibiadesc:
> Hi @alcibiadesc 👋
> 
> It sounds like some nice updates for sure but it doesn't make sense to have two identical ext

You are right, :) Thank you very much for your time and giving me an insight that I had not previously taken into account :)

pernielsentikaer:
You could still enhance the current one with many of the features you had here if you'd like to, that would be helpful for a lot of users

thomaspaulmann:
This is indeed nice. We could allow an option to show the detail view for the Downloads Manager. Feel free to contribute to it.

PR #6003, opened 2023-04-18
Add bluesky extension

dkapila:
## Description

Browse through your Bluesky timeline. Create new posts, find people, and view your notifications.

Features:

- Create posts
- Like, Reply, Repost or Quote existing posts
- Follow, unfollow, mute, unmute users
- Search users
- See your notifications
- Mark notifications as read
- View your Bluesky timeline
- See your own posts

## Checklist


peduarte:
Hey, thanks for your contribution. We'll review it asap :)

peduarte:
Hey! Thanks so much for your contribution. Love to see a Bluesky extension. I took some time to use it, and collect some thoughts. 

Some of these thoughts are mean to ensure a consistent UX across similar types of extensions as well as to ensure some things follow our UX guidelines. 

Let me know if you have any questions, looking forward to checking it out again 😄 

---

**Home**
Im not sure we need this Command. I understand what it does, but it seems unnecessary. The beauty of Raycast is that everything is available at root, but this paradigm means everything is nested within "Home". I see you also use this as a global view, based on the Dropdown value, but the Dropdown is supposed to be used as some sort of filter, rather than a navigation UI. 

If you want, you can continue using this global view as a way to host all the other views, but I'd personally not expose a "Home" command. It feels like too much overhead, and too many instructions. I think the app and the UX is quite self-explanatory. 

_About the dropdown:_
As mentioned above, I think each view should be its own command, rather than using the dropdown for this. It doesnt play well with how the navigation stack works. Eg: If you Select timeline, then notifications, then people, then press Escape, it will keep going back in the history stack. It's a bit confusing, since normally the expectation is just to go back to the root.

But if you really want to keep it, can we:
- Can you remove the "Sign out" option from there, and make it it's own command? 
- Can you rename "People" to "Search"
- Can you rename "Pedro Duarte" (name of logged in user) to "My Recent Posts" so it's consistent with the dedicated command
- Don't think we need an About page, this is typically what you'd want to put in the README. 

**View Notifications**

Can we call this just "Notifications"? It's what we do for Linear and GiHub.

Whenever I open the View Notifications Command, I see a Toast saying there are 2 new notifications. Couple of thoughts:
- I'd like to see which ones are the new ones. Perhaps some indicator in the list item's `icon` to communicate that?
- How do I mark them as "viewed"? 

(ಠ ›ಠ) — @peduarte — vL8Si1VO001651@2x


**Menu Bar**
We normally use menu bar commands for notification systems. Things that need a background refresh and can self-update. For example, in the Linear and GitHub extension, we use it to show your unread notifications. For the Spotify extension, we use it to show you what you're currently listening to. But in this case, it's used as a navigation menu, to navigate users to a 
specific view. I'd see more value in using it as a notification system. If you agree, then can we call this "Unread Notifications" as we do for Linear and GitHub?

**Create a new Post**
I dont think we need to include instructions in the label/action that users need to press ⌘ ⏎ to submit. This is how things work in Raycast, and in the Primary Action, we already display the shortcut keys. 

(ಠ ›ಠ) — @peduarte — U9OEDnuZ001652@2x

**Search People**
Can we call this "Search"? When this view loads, I see a successful Toast, why is this? Toast should be used to communicate an async action that has succeeded. If you're waiting for the "Search to be ready", consider using the `isLoading` prop on the `List` 

(ಠ ›ಠ) — @peduarte — YdNMmBKO001653@2x

**View my Recent Posts**
Can we call this "Recent Posts"?

dkapila:
Thanks a lot for the detailed feedback @peduarte. I resolved the comments in here, and as clarified discussed on Slack. Please let me know if there is anything I've missed out on, can improve further on :)

> If you want, you can continue using this global view as a way to host all the other views, but I'd personally not expose a "Home" command.
- Removed Home command from the dropdown.

> Can you remove the "Sign out" option from there, and make it it's own command?
- Done - moved sign out from the dropdown to it's own command.

> Can you rename "People" to "Search"
- Renamed people to search.

> Can you rename "Pedro Duarte" (name of logged in user) to "My Recent Posts" so it's consistent with the dedicated command
- Renamed to My Recent Posts.

> Change View Notifications: Can we call this just "Notifications"? It's what we do for Linear and GiHub.
- Changed View Notification to Notifications

> Whenever I open the View Notifications Command, I see a Toast saying there are 2 new notifications. Couple of thoughts: I'd like to see which ones are the new ones. Perhaps some indicator in the list item's icon to communicate that?
- Added an accessory to show that the notification is not read. Also, unread and read notifications are now divided into 2 list sections to make it even more readable. 

> Menu Bar → If you agree, then can we call this "Unread Notifications" as we do for Linear and GitHub?
- Renamed to "Unread Notifications"

> Create a new Post: I dont think we need to include instructions in the label/action that users need to press ⌘ ⏎ to submit. This is how things work in Raycast, and in the Primary Action, we already display the shortcut keys.
- ⌘ + Enter to "New Post". 
- And renamed ⌘ + Enter to Publish on the Action panel.

> Search People: Can we call this "Search"?
- Fixed

> When this view loads, I see a successful Toast, why is this? Toast should be used to communicate an async action that has succeeded. If you're waiting for the "Search to be ready", consider using the isLoading prop on the List
- removed the success Toast.

> View my Recent Posts: Can we call this "Recent Posts"?
- Fixed

peduarte (changes requested):
Thanks for the changes. Took some time to re-test it again.

A few comments
- Remove the confirmation dialog from New Post
- Remove the confirmation dialog from Mark Notifications as Read

Typically we'll want to use confirmation dialogs for destructive actions so I dont think its needed for the above

I've also been thinking about the Sign Out command. The reality is that doesn't actually sign the user out, since it's not currently possible to programatically update the extension's preferences. 

Two options:
- Can we sign in via OAuth?
- If not, let's remove the Sign out command, and add a note in the Readme that to sign out people need to clear the preferences 

Thanks, looking forward to testing it again :-)

extensions/bluesky/README.md
```diff
@@ -1,3 +1,9 @@
 # Bluesky
 
-Browse through your Bluesky timeline. Create new posts, find people, and view your notifications. All from Raycast.
\ No newline at end of file
+Browse through your Bluesky timeline. Create new posts, find people, and view your notifications. All from Raycast.
+
+# Sign Out
```
peduarte, line 5: Let's use H2 (`##`) here, we already have a H1

dkapila:
Thanks, yes I have now updated to new screenshots :)

PR #6117, opened 2023-04-23
Added a Mastodon Search extension

daveverwer:
## Description

This is a new extension that allows quick access people and hashtag search. It is not intended to be a full Mastodon client, just a quick and easy way to get glanceable information such as follower count and bot status and to navigate to pages on the Mastodon website. It intentionally does not include posting, browsing timelines, mentions, or even searching user posts.

## Screencast
metadata-1
metadata-2

## Checklist


pernielsentikaer:
@BasixKOR has been working on #5001 for some time and it seems like it’s almost ready.

Now we also have #6117 and #6156 in the queue, we need to decide what to do with all three, are you up for selecting the best version and going with that and improving with the feature respectably from the other PRs?

daveverwer:
Thanks for taking a look, @pernielsentikaer 

We did search for Mastodon extensions before developing this but only found #5001 just before submitting this PR. We never thought to search the pull requests!

We just checked out @BasixKOR's client in #5001 and while the search functionality is similar, I think there's a case to make for a pure search extension that does not require an access token. In order to test out the other extension we had to go into Mastodon, make a new API app and get an access token. That's obviously necessary for a fully featured client which requires account access, but seems like overkill for someone who just wants search (like we do!)

There are also other features like our indication of bot accounts and follower count, but as you say we could spend time integrating those (if @BasixKOR agrees that they are the right way to go)

However, I think, instead, if you don't think this is valid as an extension please feel free to close it and we'll use our local copy of it rather than having it on the store.

BasixKOR:
The ability to search Mastodon without any credentials could be useful, and I don't think this extension clashes with the existing (or more appropriately, pending) Mastodon extension.

daveverwer:
I'd still like to merge it. I've been using it regularly over the last few weeks and it's working well.

> We just need to make sure it's clearly documented so the end user knows what it is

The title is currently "Mastodon Search - Search for People or Hashtags on Mastodon." which feels clear to me, but I'm happy to make refinements or changes if there's feedback on the title (or other fields that could make it clearer).

Thanks both @pernielsentikaer and @BasixKOR for your feedback 👍

pernielsentikaer:
So you think we should just have one extension for Mastodon, or?

daveverwer:
> So you think we should just have one extension for Mastodon, or?

No, sorry, I was talking about the title for this extension being clear. I think there is an argument for having a search-focused extension, which is this one.

I'm happy to make refinements to the metadata if you think the purpose of this extension could be clearer, but I was saying that it feels quite clear to me being "Mastodon Search - Search for People or Hashtags on Mastodon."

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

daveverwer:
Thank you!

PR #6120, opened 2023-04-23
Refactor ClearStatusAction component

ugobok:
Use destructuring to extract props
Use arrow function shorthand syntax
Add a confirmation dialog to prevent accidental clearing of the status

## Checklist

### DIFF
--- a/extensions/slack-status/src/setStatus.tsx
+++ b/extensions/slack-status/src/setStatus.tsx
@@ -217,14 +217,24 @@ function SetCustomStatusAction(props: {
   );
 }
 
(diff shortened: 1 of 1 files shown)


PitNikola:
Hi @ugobok! Thanks for this addition. I'm not sure I'm convinced that "Clear Status" should have a confirmation alert. It's fairly lightweight action. Even if you accidentally clear your status, it's not a problem to set it back. By introducing an alert, it will break muscle memory for a lot of existing users. If you feel strongly about this problem, I'm okay with you adding a preference "Show Confirmation Alert on Clear Status" with it being off by default. What do you think?

PR #6124, opened 2023-04-24
[brew] Add option to uninstall casks with `--zap`

urwrstkn8mare:
## Description

Just adds a simple option which only when checked runs cask uninstall commands with `--zap`. Which can be quite useful when you want to get rid of *all* of the program. (https://docs.brew.sh/Cask-Cookbook#zap-purpose)

## Checklist


extensions/brew/package.json
```diff
@@ -56,6 +56,14 @@
       "required": false,
       "title": "Custom Brew Executable Path",
       "description": "Set this if your brew executable is in a non-standard location"
+    },
+    {
+      "name": "zapCask",
+      "type": "checkbox",
+      "required": false,
+      "title": "Zap Casks",
+      "label": "Add --zap to brew uninstall --cask",
```
nhojb, line 65: I'd suggest using a similar summary as `man brew` here and using a label like "Uninstall all files associated with a cask". This is clearer for non-technical users. The description text already explains the preference in more detail.

Probably worth also adding the warning "May remove files which are shared between applications" to the description?

nhojb (changes requested):
Thanks for the contribution. Works as expected here.

Just one minor suggested improvement for the preference's label text.

extensions/brew/package.json
urwrstkn8mare: > I'd suggest using a similar summary as `man brew` here and using a label like "Uninstall all files associated with a cask". This is clearer for non-technical users. The description text already explains the preference in more detail.

```suggestion
      "label": "Remove all files associated with a cask.",
```
Hi good suggestion! I'll make the change you can see here.

urwrstkn8mare: I've also updated the description with what I hope is clear enough.

extensions/brew/package.json
```diff
@@ -56,6 +56,14 @@
...
       "title": "Custom Brew Executable Path",
       "description": "Set this if your brew executable is in a non-standard location"
+    },
+    {
+      "name": "zapCask",
+      "type": "checkbox",
+      "required": false,
+      "title": "Zap Casks",
+      "label": "Remove all files associated with a cask.",
+      "description": "Run cask uninstall with the --zap option for a more complete uninstallation of files associated with cask. WARNING: This may remove files shared with other applications."
```
nhojb, line 66: I'd suggest adding a newline before the warning and removing the all caps.

nhojb (changes requested):
One minor suggested change to the preference description. Otherwise good to go.

Thanks 😄

extensions/brew/package.json
urwrstkn8mare:
```suggestion
      "description": "Run cask uninstall with the --zap option for a more complete uninstallation of files associated with cask.\nWARNING: This may remove files shared with other applications."
```
Oh I didn't realise you could add new lines to tooltips. I'll make sure that's done.

I think maybe its better to keep it in caps just have a higher chance of people noticing it? If you disagree feel free to make it commit the change, make a suggested change or just lmk and it'll be done.

nhojb (approved):

PR #6127, opened 2023-04-24
[Visual Studio Code Project Manager] Update build identifier for Codium.

theherk:
## Description

Based on changes in VSCodium/vscodium#1227, this updates to the correct `darwinBuildIdentifier` for VS Codium. Without this change Finder is opened when trying to open in Codium.

- Relates to: VSCodium/vscodium#1227

## Checklist


pernielsentikaer:
We should maybe ensure that older versions of VSCodium still work. This obliterates the old build name. 

Should we consider adding it as an array instead so there can be multiple bundles?

theherk:
> Should we consider adding it as an array instead so there can be multiple bundles?

I considered that, but couldn't settle. I think what may be best is add another option. So, in lieu of `VSCodium`, have both `VSCodium < 1.70` and `VSCodium` with the new value.

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #6201, opened 2023-04-27
[brew] should fix #6198

### LINKED ISSUE #6198: [brew] Add capability to change brew path (Apple Silicone)
### Extension

https://www.raycast.com/nhojb/brew

### Description

On Macs with Apple Silicone/arm64 the default brew path commonly used is `/opt/homebrew/bin/brew`.

Note that many have a second brew installed for rosetta that uses the old `/usr/local/bin/brew`, however, not as the main brew.

Ever brew user with Apple Silicone

### Anything else?

_No response_

urwrstkn8mare:
## Description

See https://github.com/raycast/extensions/issues/6198

## Checklist

### DIFF
--- a/extensions/brew/src/brew.ts
+++ b/extensions/brew/src/brew.ts
@@ -100,11 +100,9 @@ export interface OutdatedResults {
 
 /// Paths
 
-export const customBrewPath =
-  preferences.customBrewPath && preferences.customBrewPath.length > 0 ? preferences.customBrewPath : undefined;
-
-export const brewPrefix: string = (() => {
-  if (customBrewPath) return path_join(customBrewPath, "..", "..");
+export const brewPrefix = (() => {
+  if (preferences.customBrewPath && preferences.customBrewPath.length > 0)
+    return path_join(preferences.customBrewPath, "..", "..");
   try {
     return execSync("brew --prefix", { encoding: "utf8" }).trim();
   } catch {
@@ -112,13 +110,8 @@ export const brewPrefix: string = (() => {
   }
 })();
 
-export function brewPath(suffix: string): string {
-  return path_join(brewPrefix, suffix);
-}
-
-export function brewExecutable(): string {
-  return customBrewPath ?? path_join(brewPrefix, "bin/brew");
-}
+export const brewPath = (suffix: string) => path_join(brewPrefix, suffix);
+export const brewExecutable = () => brewPath("bin/brew");
 
 /// Fetching
 


extensions/brew/src/brew.ts
```diff
@@ -100,7 +100,11 @@ export interface OutdatedResults {
 
 /// Paths
 
+export const customBrewPath =
+  preferences.customBrewPath && preferences.customBrewPath.length > 0 ? preferences.customBrewPath : undefined;
+
 export const brewPrefix: string = (() => {
+  if (customBrewPath) return path_join(customBrewPath, "..", "..");
```
nhojb, line 107: I think something like this would potentially be more robust and future proof:

```javascript
const brew = if (customBrewPath) ? customBrewPath : "brew";
try {
  return execSync(`${brew} --prefix`, { encoding: "utf8" }).trim();
} catch { ... }

**Review by nhojb — CHANGES_REQUESTED** [review]
Hey, thanks for the quick fix!

Just one suggested improvement.

File: extensions/brew/src/brew.ts (hunk 1 above)
Comment (nhojb) [reply extensions/brew/src/brew.ts:107]: Good point.

In that case I'd suggest this:

```typescript
export const brewPrefix: string = (() => {
  const customBrewPath = preferences.customBrewPath && preferences.customBrewPath.length > 0 ? preferences.customBrewPath : undefined;
  if (customBrewPath) {
    return execSync(`${customBrewPath} --prefix`, { encoding: "utf8" }).trim();
  }
  try {
    return execSync("brew --prefix", { encoding: "utf8" }).trim();
  } catch { ... }
}
```

extensions/brew/src/brew.ts
```diff
@@ -113,11 +117,7 @@ export function brewPath(suffix: string): string {
 }
 
 export function brewExecutable(): string {
-  if (preferences.customBrewPath && preferences.customBrewPath.length > 0) {
-    return preferences.customBrewPath;
-  } else {
-    return path_join(brewPrefix, "bin/brew");
-  }
+  return customBrewPath ?? path_join(brewPrefix, "bin/brew");
```
nhojb, line 120: Just use `return brewPath("bin/brew");` here. The `brewPrefix` will have the custom path if necessary.

extensions/brew/src/brew.ts
```diff
@@ -100,7 +100,11 @@ export interface OutdatedResults {
 
 /// Paths
 
+export const customBrewPath =
+  preferences.customBrewPath && preferences.customBrewPath.length > 0 ? preferences.customBrewPath : undefined;
```
nhojb, line 104: Remove this since it can be moved into the `brewPrefix` closure.

nhojb (changes requested):

extensions/brew/src/brew.ts
nhojb: In retrospect it's a shame that the custom path preference did not specify the brew directory (e.g. `/opt/homebrew`) and not the binary path. But it's a bit late to change that now 😢

urwrstkn8mare:
@nhojb I've made the changes and also cleaned up the section a little. How does it look now?

nhojb:
Looks great. Thanks!

nhojb (approved):

PR #6204, opened 2023-04-27
Replace QuickActions with Disabled Root Commands

peduarte:
As of Raycast 1.50.0, commands can now be disabled by default, via a `disabledByDefault` property. 

This flag was introduced to help deal with extensions that expose too many commands. Spotify is one of them. The idea is that users can tailor their experience by optionally choosing to enable some of these commands, if they wish to have quicker access to them. 

Note: If you've been using Quicklinks via the previous Quick Actions command, please delete them and enable the commands you need in Preferences.


sxn (approved):
🚢

PR #6220, opened 2023-04-28
[AWS] Support GetSessionToken with aws-vault

JonathanWbn:
## Description

When using `aws-vault` there is an option to use AssumeRole or GetSessionToken API calls. The current solution only supports the former. This PR adds support for the latter.

This has been pointed out by @themanontv in https://github.com/raycast/extensions/issues/3022#issuecomment-1527157975.


pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #6230, opened 2023-04-28
Add totp-generator extension

lacherogwu:
## Description

Generate time-based one-time passwords

This lets you easily generate a time-based one-time password (TOTP).

## Screencast
totp-generator-1
totp-generator-2

## Checklist


lacherogwu:
@thomaslombart Please to support the extension store

Aayush9029:
🎉 Hurray! I'll definitely use this.

*Ignorable Suggestion:*
It may be worth considering using `Toast` and `accessories` or even `showHUD + Copy` instead of a large detail view for a length of only 6 text. Additionally, since secrets are not something users will need to view every time they want to view totp code, dismissal or a copy action should suffice if necessary.

Unless there's something fancy you're planning to add later on 😉.


Anyways great work! Definitely useful.

lacherogwu:
@Aayush9029 
I guess you are right in what you say, well I've gave it a thought and we can take it to 2 directions
1. Google Authenticator like
2. What it is right now, which is just a tool to play with, or in case you need 2fa for a service you don't really care and you don't want to go to Google authenticator or some other platform to add it and save it, you need something quick to generate this passcode. 

And there is an Action already to copy straight away without passing through this display window

Let me know what do you think

pernielsentikaer:
@Aayush9029 are you available regarding options

Aayush9029:
I strongly favor Option 1 😊, @lacherogwu. It might be beneficial to include a new command that displays a list of "saved tokens" and their codes.

lacherogwu:
Okay then, Will change it 😌

lacherogwu:
@Aayush9029 I will close PR
I've changed the extension completely, hope you will like it!

here is the new pr #6338

PR #6241, opened 2023-04-29
[Tailscale] Improve UI for disconnecting exit node

itsmingjie:
## Description

Right now, the "Turn off exit node" action has a "connected" indicator on it. This is confusing to look at, especially it is already clear which exit node you are currently using. This PR removes that indicator.

Before

Comparing raycastmain itsmingjiepatch-1 · raycastextensions - Arc - 2023-04-29 at 15 04 16@2x

After

Comparing raycastmain itsmingjiemain · raycastextensions - Arc - 2023-04-29 at 16 32 05@2x

## Checklist

### DIFF
--- a/extensions/tailscale/src/exit.tsx
+++ b/extensions/tailscale/src/exit.tsx
@@ -67,7 +67,7 @@ function ExitNodeList() {
           title="Turn off exit node"
           actions={
             <ActionPanel>
-              <Action title="Turn Off Exit Node" onAction={() => setExitNode("", false)} />
+              <Action title="Turn off Exit Node" onAction={() => setExitNode("", false)} />
             </ActionPanel>
           }
         />


pernielsentikaer:
Thanks for the update @itsmingjie 🔥 

@samlinville do you mind checking this?

extensions/tailscale/src/exit.tsx
```diff
@@ -80,6 +61,25 @@ function ExitNodeList() {
 
   return (
     <List isLoading={exitNodes === undefined}>
+      {isActive && (
+        <List.Item
+          key="_disable"
+          title="Turn off exit node"
+          icon={{
```
samlinville, line 68: I would prefer if you removed the icon from this `List.Item` altogether rather than introduce a red dot icon; the dot indicators are for showing the online/offline status of nodes. We don't currently use a red dot in any of our other UIs.

extensions/tailscale/package.json
```diff
@@ -6,17 +6,18 @@
   "icon": "command-icon.png",
   "author": "samlinville",
   "owner": "tailscale",
-  "access": "public",
```
samlinville, line 9: Please don't change the order of fields here, I'd like for `access` to remain under `owner`

samlinville (commented):
Thanks @itsmingjie for the contribution! Left a few notes, once we resolve those I think we'll be in good shape.

extensions/tailscale/src/exit.tsx
samlinville: Perhaps use the `XMarkCircle` icon instead then? https://developers.raycast.com/api-reference/user-interface/icons-and-images

itsmingjie: Resolved in 3070902 -- it doesn't look terrible.
Tailscale  Improve UI for disconnecting exit node #6241 - Arc - 2023-05-01 at 11 43 48@2x

samlinville (approved):
Thanks!

PR #6255, opened 2023-04-30
Add tmuxswitch extension

NitroCao:
## Checklist


extensions/tmuxswitch/package.json
```diff
@@ -0,0 +1,61 @@
...
+        {
+          "name": "sessionName",
+          "description": "Session name where tmux locates in iTerm2",
+          "type": "textfield",
+          "required": false,
+          "title": "Session name where tmux locates in iTerm2"
+        },
+        {
+          "name": "socketPath",
+          "description": "Socket path of tmux",
```
pernielsentikaer, line 35: Maybe add a placeholder (and update README) with how to get the socket path

pernielsentikaer (commented):
Would it make sense to support more terminal like warp/terminal too

pernielsentikaer:
Seems like we already have an tmux extension in the queue (#6200) - I think we should focus on only having one but you two could maybe take the best from each and make one?

NitroCao:
> Would it make sense to support more terminal like warp/terminal too

It's ok and easy to support more popular terminals.



> Seems like we already have an tmux extension in the queue (#6200) - I think we should focus on only having one but you two could maybe take the best from each and make one?

I just checked the extension you mentioned, that extension has no help for me. I don't want to port simple commands of tux into Raycast, it doesn't make sense cause it won't improve efficiency. My extension aims to solve my problem that is to quickly locate a window among around 10+ windows.

pernielsentikaer:
There is no reason why a `tmux` extension can't have both features in one instead of having it as two, that would also move some of the confusion that might happen 😊

NitroCao:
> There is no reason why a `tmux` extension can't have both features in one instead of having it as two, that would also move some of the confusion that might happen 😊

It can indeed have both features, but it's not necessary in my opinion 😅. It's your turn then to decide which one to choose 🤣

louishuyng:
Hi @pernielsentikaer @NitroCao i think we can have two extension separetly. What i foucs not like what @NitroCao focus to build (i focus more on session manage and @NitroCao focus on window manage). May be we can change the name of each extension to avoid confusion. @pernielsentikaer let me know if that ok with you and you can suggest us a suitable name

pernielsentikaer:
Why not make one extension with two commands, so the user has one for `tmux` where he can use both? The two codebases are not significant, so it shouldn't be a problem.

louishuyng:
That"s ok for me @pernielsentikaer we can have one extension for both commands. 

However, @NitroCao feel my tmux session commands not be useful much for his usercase so i have no idea how to group into one @pernielsentikaer let me know what i need to do in this scenario

NitroCao:
> There is no reason why a `tmux` extension can't have both features in one instead of having it as two, that would also move some of the confusion that might happen 😊

According to @pernielsentikaer 's opinion, I'd prefer to revert my PR and commit my code to louishuyng's extension, cause his extension has more features.

pernielsentikaer:
That would be the best option for the end user too, so we're closing this and after #6200 is released you make a PR with yours, or?

NitroCao:
> That would be the best option for the end user too, so we're closing this and after #6200 is released you make a PR with yours, or?

Yeah I'll, and will also communicate with @pernielsentikaer the details in the future when I have spare time.

PR #6277, opened 2023-05-02
audio-device: update system effects when output changes

### LINKED ISSUE #4706: [Set Audio Device] Changing audio device does not affect sound effects
### Extension

https://www.raycast.com/benvp/audio-device

### Description

Changing the audio device through the extension changes the "Play sound effects through" setting from "Selected Sound Output Device" (sound effects play through the active audio device) to whatever the previous audio output was before switching. This results in sound effects always playing through the first device that was in use before using the extension until the setting is changed back to "Selected Sound Output Device" in System Settings.

### Steps To Reproduce

1. Switch audio device using the extension
2. Observe that the "Play sound effects through" setting in System Settings has changed from "Selected Sound Output Device" to the previous audio device in use before switching

### Current Behaviour

"Play sound effects through" setting is affected by the extension

### Expected Behaviour

"Play sound effects through" setting remains untouched

rcruzper:
## Description

Add an option to update "Play sound effects through" in Sound Preferences when output device changes

## Screencast

Screenshot 2023-05-10 at 18 56 21

## Checklist

resolves #4706

### DIFF
--- a/extensions/audio-device/src/helpers.tsx
+++ b/extensions/audio-device/src/helpers.tsx
@@ -133,9 +133,9 @@ function SetAudioDeviceAction({ device, type }: SetAudioDeviceActionProps) {
 }
 
 async function setOutputAndSystemDevice(deviceId: string) {
-  const { systemoutput } = getPreferenceValues();
+  const { systemOutput } = getPreferenceValues();
   await setDefaultOutputDevice(deviceId);
-  if (systemoutput) {
+  if (systemOutput) {
     await setDefaultSystemDevice(deviceId);
   }
 }
--- a/extensions/audio-device/package.json
+++ b/extensions/audio-device/package.json
@@ -52,7 +52,7 @@
       "default": false
     },
     {
-      "name": "systemoutput",
+      "name": "systemOutput",
       "required": false,
       "title": "Play sound effects through current output",
       "description": "Play system sounds through current output.",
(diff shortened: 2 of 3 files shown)


extensions/audio-device/src/helpers.tsx
```diff
@@ -130,6 +131,11 @@ function SetAudioDeviceAction({ device, type }: SetAudioDeviceActionProps) {
   );
 }
 
+function setOutputAndSystemDevice(deviceId: string) {
+  setDefaultOutputDevice(deviceId);
+  setDefaultSystemDevice(deviceId);
+}
```
mathieudutour, line 137:
```suggestion
async function setOutputAndSystemDevice(deviceId: string) {
  await setDefaultOutputDevice(deviceId);
  await setDefaultSystemDevice(deviceId);
}
```

FezVrasta:
Should this be a user preference?

rcruzper:
> Should this be a user preference?

in my use case the system sounds should always be listened from the current output device. Which use cases do you think would need that user preference?

FezVrasta:
I switch the audio to the headphones during calls but I don't want the system sounds there. Same when I listen to music on my HomePod.

rcruzper:
Makes sense, I will add the preference.

On the other hand, the current audio-device binary asset used to change output and input device is only compatible with intel chips and I have a M2 chip so every time I want to make tests or used the extension I need to compile as universal binary, I want to change the package.json to use another library which compiles with an universal binary [here](https://github.com/spotxyz/macos-audio-devices/blob/master/package.json#L25). Do you think I should create another PR o can I change it here.

rcruzper:
updated with the user preference, do you think is there any better naming for title and description?

FezVrasta (approved):

mathieudutour (approved):
Looks good, thanks!

Issue #6302, opened 2023-05-03
[Bitwarden Vault] ...


Tigeredgar:
### Extension

https://www.raycast.com/jomifepe/bitwarden

### Description

Keeps logging me out right after logging in.

### Steps To Reproduce

Login with master password. 
Wait 2 seconds
Logs me out and prompts for master password again

jomifepe:
Hello, thanks for reporting, I'm taking a look into the issue. In the meantime, please try to reinstall the extension.

trm217:
Experiencing the same issue currently.

jomifepe:
Can you confirm that the issue still persists are the last update?

GumbyXGames:
Upgraded to Raycast v1.50.4 and issue persits.

trm217:
Can confirm, issue persists.

jomifepe:
Please verify if the issue still persists with the latest update

trm217:
@jomifepe It's working for me now. Thank you for your effort 🙌🏻

jomifepe:
@Tigeredgar can you confirm that it's working correctly?

GumbyXGames:
Bitwarden extension seems to be working fine after installing the update. Thanks for the quick fix!

jomifepe:
I will assume that this issue is resolved, given that two people already confirmed that it's working. 

If not, be sure to comment, thanks 🙂

@raycastbot close this issue

PR #6324, opened 2023-05-04
[Bitwarden] Use session token on every bitwarden cli command

jomifepe:
### What was done
- The session token is now saved on the environment object passed to **every** `Bitwarden CLI` command. The CLI itself also recommends that you export the session token once you unlock your vault, which is what is essentially being done here. Also, this way, you avoid passing the token around to the functions.

## Screencast

No visual changes, everything should look the same. It's possible that this might [fix this issue](https://github.com/raycast/extensions/issues/6302#issuecomment-1534196902), but I'm not entirely sure if this is the cause. Nonetheless, this is something that needed to be done, in order to make sure that every bitwarden command is using the same session.

### Manual sanity check

https://user-images.githubusercontent.com/7235666/236197145-336cb840-dfb9-4c4f-ac8a-6f979801ab86.mp4

## Checklist

- [X]  I read the [documentation about publishing](https://developers.raycast.com/basics/publish-an-extension)

### DIFF
--- a/extensions/bitwarden/CHANGELOG.md
+++ b/extensions/bitwarden/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Bitwarden Changelog
 
+## [Improvement] - 2023-05-04
+
+- Use session token on every bitwarden cli command
+
 ## [Performance Improvement] - 2023-04-03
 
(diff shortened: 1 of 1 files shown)


pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #6370, opened 2023-05-06
[Laravel Valet] - New features

KevinBatdorf:
## Description

This adds new features to the Laravel Valet extension

- Start/Stop/Restart Valet
- View log files

Also updated raycast libs, tweaked some formatting issues here and there, etc

## Screenshots

Screen Shot 2023-05-06 at 11 59 39 AM

Screen Shot 2023-05-06 at 12 09 17 PM

## Checklist


KevinBatdorf:
@denniseilander feel free to push any updates or changes!

denniseilander:
Hi @KevinBatdorf,
Thanks for these features, really like them! The log viewer was on my list, nice to see that you took care of it.

Do you mind if I look at it tomorrow?

Thanks again! 😀

KevinBatdorf:
Yeah sure no rush at all.

Also, Raycast can accept command arguments now and was thinking of piping them through valet, so it would just run `valet ${arg}`. But in a follow-up PR.

https://developers.raycast.com/information/lifecycle/arguments

So this way you could invoke Raycast, highlight the Valet command, press tab, then type 'use php@8.1' and it will run the command.

If you have any thoughts on that let me know.

denniseilander:
Hi @KevinBatdorf,

I've tested the additions, I really like the log viewer! 🔥

The idea of grouping the secure/unsecure commands is a great addition IMO, would like to add this change to this PR?

I also like the idea of passing the commands through valet, I think this makes the extension much more powerful! 💪
To keep this PR manageable, I think it would be good to open a separate PR for this addition.

I'm not sure about the way to access the stop/restart command via the cmd+k option.
Perhaps a separate command would suit this better, what do you think?


Furthermore, I think this PR contains very nice additions, for which I thank you!

KevinBatdorf:
I added the grouping of the secure/unsecure and added commands to the top of the list in their own section. I tend to think multiple commands here might fee like clutter since there are only a few things to do with the extension.

laravel-valet-7

What do you think?

I also did some more refactoring if you could just check it over when you have some free time.

> I also like the idea of passing the commands through valet, I think this makes the extension much more powerful! 💪
> To keep this PR manageable, I think it would be good to open a separate PR for this addition.

I tried to MVP this but it borked my install. So maybe needs some more careful attention. 

I often switch php versions in Valet though, so maybe a feature to do that would be useful. I can see it being a list item under "View Logs" that opens a view that gets the most recent 5 php versions and lists them with "Switch to php 7.4". But yeah... for another PR. I'll mess around with it when I have time.

KevinBatdorf:
Also, I guess we could update the screenshots too. Let me know if you are OK with this direction and I can prepare them. I'll need to set up dummy sites or configure the plugin with some fake data I guess. (or if you have it already set up, can you share a handful of them?)

pernielsentikaer:
Are we ready to merge, or not yet @KevinBatdorf

KevinBatdorf:
@denniseilander What do you think? Have you had a chance to look at the changes?

denniseilander:
Hi @KevinBatdorf,

Sorry, I've been a bit busy the last few weeks, but I've been looking into it. It looks good!
I'm still not sure about the separate commands in the list view, I would see them as separate commands instead, so you have a list "command" and the start/stop, views commands.

I imagine this is a bit more work than just adding them to the list, but these are commands you shouldn't need that often, so I think it's better not to show them all the time. What do you think?

I think the other improvements are OK.

KevinBatdorf:
> I imagine this is a bit more work than just adding them to the list, but these are commands you shouldn't need that often, so I think it's better not to show them all the time. What do you think?

I don't think it's much more work. It wouldn't be able to dynamically show start/stop but that probably doesn't matter much anyway.

It's just whether the extension should be a separate command for every Valet feature or a single command that includes them all. I'd hate to add 10+ commands just for Valet. 

What if the sites list was nested one level deeper? So the initial list is all the commands plus one that says "View Sites"?

If you feel strongly about it though Im ok separating them. I'm of two minds here and am ok with both approaches.

How are you using the extension (and valet) now? Personally I regularly either restart valet or switch PHP versions but I don't find checking on specific sites ever, really.

denniseilander:
@KevinBatdorf thanks for the reply.

I agree with you that 10+ commands is too much.
I think I can live with these 3 top level commands for now.
When a search is triggered, the commands immediately disappear, so that's nice.

Personally, I hardly ever restart Valet. Since Valet has a built-in option to support multiple PHP versions per project (site), I no longer have to manually switch PHP versions. The only time I restart valet is when I've broken something, installing some other php versions lol.

If the idea of passing the valet command is still alive, we could remove these visual commands and maybe let the search bar handle them, but that's for the next part.

@pernielsentikaer You can merge the PR ✅

denniseilander:
@KevinBatdorf Yeah, this feature is on my list too, but it will be easier to integrate with native Valet commands when we can pipe them.

For your info, using the `valet isolate php@[version]` command, you can set the version that specific project is running. So your machine can run the latest php version as default and your projects can run whatever php version you'd like.

Before this feature, I needed to restart valet multiple times a week.

denniseilander:
@KevinBatdorf, @pernielsentikaer, I just tested this PR again and it looks like the search is broken. The list of sites is not filtered by the inserted query, can you take a look at it to be sure?

Thanks!

denniseilander:
@KevinBatdorf Tested it again and it works! Thanks 👍

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #6387, opened 2023-05-07
[Bitwarden] Improve locked vault handling and separate concerns

jomifepe:
### What was done:
- Previously a `useEffect` to check the vault status was introduced on the `SessionProvider`, which ran when the extension initialized. This might be the cause of [this issue](https://github.com/raycast/extensions/issues/6302). This PR removes this code in favor of checking every time if a command asks for the master password, if it does, the vault is locked and the appropriate actions are performed.
- Some separation of concerns, now the `SessionProvider` component doesn't call as many `bitwarden` class commands, and is not called anymore for actions like locking and unlocking, instead session-specific code for these actions is called as a callback from the `bitwarden` class actions. This way, calls are made to the `bitwarden` class directly, and the session manages itself.

## Screencast

No major visual changes, everything should look the same. 

### What will happen if the vault is for some reason locked when accessing it

As you can see, the items are loaded first because of caching but, in the background, when getting the updated items, the command fails because the vault is locked, and the unlock form appears.

https://user-images.githubusercontent.com/7235666/236708213-5590ab26-89d5-4f0d-9b80-60a3a0d62789.mp4

### Manual sanity check

https://user-images.githubusercontent.com/7235666/236707364-a1711288-0259-400e-bd51-650a33a9efcd.mp4

## Checklist

### DIFF
--- a/extensions/bitwarden/CHANGELOG.md
+++ b/extensions/bitwarden/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Bitwarden Changelog
 
+## [Fix] - 2023-05-08
+
+- Hopefully fixed a bug that caused the extnesion to keep logging out right after logging in.
+
 ## [Improvement] - 2023-05-04
 
 - Use session token on every bitwarden cli command


pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #6477, opened 2023-05-11
[Steam] Add ai game recommendation

KevinBatdorf:
## Description

This adds a basic game recommendation from the AI interface recommending a game the user might want to check out.

Notes:
- If you don't have AI (non-pro user) then the experience doesn't change
- If you are not logged in, this won't run
- If you are logged in but haven't played any games, it will just send back a random popular game (or so it seems)
- Prompt sends your previously played games (ordered by length of play time)
- Prompt does not include games played less than 5 hours
- Prompt includes only 20 most played games
- Prompt includes a mention to not recommend a game you recently viewed
- Creativity is set to 0 (because I don't think this would have a positive effect?)

Screen Shot 2023-05-11 at 3 21 26 AM

This is a very simple implementation for now but I would like to eventually have a view that includes recommendation based on various things, like games you viewed, games you've played, recent games you added to your library, varying genres, and more. But I need to upgrade various parts of the extension and API, so this is just v1 to get something going and maybe get feedback

## Checklist


pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #6577, opened 2023-05-16
Update typefully extension

albertogiunta:
## Description

The Typefully backend expects the "schedule_date" field to be "schedule-date". With this change it will work correctly.

Disclaimer: I'm a Typefully Team member, I tested it and it now works as expected.

## Checklist

### DIFF
--- a/extensions/typefully/CHANGELOG.md
+++ b/extensions/typefully/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Typefully Changelog
 
+## [Fixed draft scheduling] - 2023-05-16
+
+- Fixed an issue where creating a new scheduled draft wouldn't take into account the scheduling option.
+
 ## [Fixes and enhancements] - 2023-02-03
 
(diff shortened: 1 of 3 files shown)


pernielsentikaer:
Hi @albertogiunta 👋 

According to the Typefully API Documentation, it is with an underscore, is that is wrong then? (Just to make sure when its deviating from the documentation) 

Raycast-3QDkuE3L@2x

albertogiunta:
Hey @pernielsentikaer, thanks for the quick reply. You're absolutely correct, there seems to be an inconsistency between our code and docs and we're discussing how to fix it in the most straightforward way. Will let you know if we opt to proceed with this change or change our backend (and in that case I will close the PR).

Thanks again, will let you asap 🙏

albertogiunta:
@pernielsentikaer I confirm we're proceeding with this change for Raycast (I already took care of updating the endpoint to also accept the legacy parameter to not break old clients), and we're updating the API docs.

From our side it's all good to be merged!

pernielsentikaer:
Thanks - I'll wait with the merge here then 🙂

PR #6610, opened 2023-05-17
Cast character to lowercase first, since it shouldn't matter if letters are entered upper- or lowercase.

sailingdeveloper:


jesse-c:
I can't test this locally right now but the change looks reasonable! Thanks for the fix.

jesse-c (approved):

jesse-c (approved):

pernielsentikaer:
Raycast-81edAaOc@2x

Confirmed 🙂

Issue #6145, opened 2023-04-25
[Google Translate] usually gives me different and worse results than google translate web


danielo515:
### Extension

https://www.raycast.com/gebeto/translate

### Description

This is not properly a bug, just something I wanted to raise, because it defeats a bit the purpose of the extension. Many times the extension gives me results that I'm not happy with, so I go and test the same query in the web version and then I get much better results there.

gebeto:
This one is in progress, hope will merge it soon

gebeto:
https://github.com/raycast/extensions/pull/6666

gebeto:
This issue can be closed, it was fixed in #6666

PR #6666, opened 2023-05-19
Google Translate: replace translate library, add error handling

gebeto:
## Description

- Added error handling
- Updated the Raycast libraries to the latest version.
- Looks like I’ve found a library that uses the same API as translate.google.com, so now all translations are identical to translate.google.com 🎉
- Ideally, there should be no more TooManyRequests errors now! 🚀

#### Errors handling
translate 2023-05-19 at 15 39 56

#### The same translation as on translate.google.com
Screenshot 2023-05-19 at 16 09 35
translate 2023-05-19 at 16 09 19

## Checklist

### DIFF
--- a/extensions/google-translate/CHANGELOG.md
+++ b/extensions/google-translate/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Google Translate Changelog
 
+## [Enhancement] - 2023-05-22
+
+- Added error handling
+- Updated the Raycast libraries to the latest version.
+- Translations are now identical to translate.google.com 🎉
+- Ideally, there should be no more TooManyRequests errors now! 🚀
+
 ## [Enhancement] - 2023-02-10
 
 - Add Languages Set manager to be able to configure multiple language presets and easy switch between them


extensions/google-translate/src/translate.tsx
```diff
@@ -84,17 +84,10 @@ export default function Command(): ReactElement {
         }
       })
       .catch((error) => {
-        console.error("translate error", error);
-        // HACK: BAD_NETWORK errors are caused by too many requests in quick succession;
-        // ignore and retry – this should be fixed systemically.
-        if (error?.code === "BAD_NETWORK") {
-          setRetryCount(retryCount + 1);
-        } else {
```
gebeto, line 92: @fe9lix how this one helped to fix TooManyRequests error? As I understand it is trying to call translate api infinitely while it will not return some result, but in this way it will be just returning TooManyRequests, or is it return success response after a bunch of requests?

gebeto: Anyway, I've changed translate library, so now it should be irrelevant

gebeto:
https://github.com/raycast/extensions/issues/6145

extensions/google-translate/.gitignore
```diff
@@ -1,2 +1,10 @@
+# See https://help.github.com/articles/ignoring-files/ for more about ignoring files.
+
+# dependencies
+/node_modules
```
gebeto, line 4: I'm not sure that this is correct one, it is pointing to the root with `/`

gebeto:
Thanks for updating change log 🙌

PR #6689, opened 2023-05-20
Add open-path-with extension

ProAlexUSC:
## Description

Open a path in a custom application of your choice. Visual Studio Code & Xcode provided by default (if installed). It also includes features such as history preservation and clipboard reading.

## Screencast
open-path-1
open-path-2
open-path-3

## Checklist


ProAlexUSC:
> I'm not sure I get what the use case for this extension is, can you try to explain me a bit more?

In my daily development work, I use various command-line tools that output the paths of certain logs. I need to open these logs using certain applications and also need to compare different log information. Therefore, I developed this extension. With it, I can customize application configurations and manage a history of previously opened logs. This has streamlined my workflow and made comparing logs much easier.

pernielsentikaer:
Maybe you should add the path as a required argument instead and then you could actually use a no-view for the actions instead, don't you think that would be faster?

Raycast 2023-06-06 at 14 28 16

PR #6700, opened 2023-05-21
Update ticktick extension

Alireza29675:
## Description

When creating a new task, the extension will guess which project the task belongs to. For non-pro users it just acts as before.

## Screencast

screencapture

## Checklist

### DIFF
--- a/extensions/ticktick/CHANGELOG.md
+++ b/extensions/ticktick/CHANGELOG.md
@@ -1,6 +1,6 @@
 # TickTick Changelog
 
-## [Added] - 2023-05-23
+## [Added] - 2023-05-26
 
 - When creating a new task, the extension will guess which project the task belongs to. For non-pro users it just acts as before.
 


pernielsentikaer:
Seems to work pretty good @Alireza29675, well done! 🔥 

@viduycheung do you mind checking this?

extensions/ticktick/src/create.tsx
```diff
@@ -74,6 +77,16 @@ export default function TickTickCreate() {
 
   const onListChange: NonNullable<Form.Dropdown.Props["onChange"]> = useCallback((newValue) => {
     LocalStorage.setItem("defaultAddList", newValue);
+    setProjectId(newValue);
+  }, []);
+
+  const onTitleBlur: NonNullable<Form.TextField.Props["onBlur"]> = useCallback(async (newValue) => {
+    const title = newValue.target.value;
+    if (title && environment.canAccess(AI)) {
+      const toast = await showToast(Toast.Style.Animated, "🧠 Guessing project...");
```
viduycheung, line 86: debounce guessing project when on change and cancel debounce on blur and guessing project immediately will be better

extensions/ticktick/src/create.tsx
```diff
@@ -89,14 +102,14 @@ export default function TickTickCreate() {
       }
       isLoading={isLoading}
     >
-      <Form.TextField ref={titleRef} id="title" autoFocus title="Title" placeholder="No Title" />
-      <Form.Dropdown ref={listPickerRef} id="list" title="List" onChange={onListChange} defaultValue={defaultAddList}>
+      <Form.TextField ref={titleRef} id="title" onBlur={onTitleBlur} autoFocus title="Title" placeholder="No Title" />
+      <Form.TextArea ref={descRef} id="desc" title="Description" placeholder="" />
+      <Form.DatePicker ref={dueDatePickerRef} id="dueDate" title="Due Date" type={Form.DatePicker.Type.DateTime} />
```
viduycheung, line 107: restore the original order

viduycheung (changes requested):
It's a great idea, but there are a few details to improve.
I think this feature should be disabled by default and can manually open by the user.

extensions/ticktick/src/create.tsx
pernielsentikaer: That would make sense, I like the idea 🙂

Alireza29675: Great idea! I put a 500ms debounce. We might hit the rate limit if we type slowly and add multiple tasks in a row, but in normal usage it works fine to me. 🙌

extensions/ticktick/src/create.tsx
Alireza29675: My main motivation for that was that if it's being done by autocomplete, then it probably shouldn't be in the next tab-index. What do you think?

(I've put the original order back in place, so we can keep it as it was or change it we decided to)

Alireza29675:
@pernielsentikaer Settings have been added and are disabled by default 🙌

Should we enable it by default or wait until we add more features like auto-filling for date and time?

extensions/ticktick/src/create.tsx
viduycheung: I get your point. However, my concern is that the majority of users may find it inconvenient to have to go to the bottom of the form to change the project.

viduycheung:
@Alireza29675 I think we can add an action to manually parse, and keep the auto-parsing feature disabled by default.

pernielsentikaer:
@viduycheung Sounds like a plan for a start 🙂

viduycheung (approved):
Looks good to me 🔥 @Alireza29675 @pernielsentikaer

pernielsentikaer:
Could be cool to have an action to Guess Project, something like:

```
{environment.canAccess(AI) && (
  <Action
    onAction={() => guessProjectByTitle}
    icon={Icon.Wand}
    title="Guess Project"
    shortcut={{ modifiers: ["cmd"], key: "g" }}
  />
)}
```

Do you think we could add that

extensions/ticktick/src/create.tsx
Alireza29675: I guess we can resolve this

extensions/ticktick/src/create.tsx
Alireza29675: I agree that it is wiser decision to keep it as it is 👍 I assume this thread is also resolved

Alireza29675:
@viduycheung @pernielsentikaer I added an action called "Fill the Form with AI" with the hotkey "Command+F". I'm okay with other options; just change the title or hotkey if you feel like it 🙌

viduycheung (approved):

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

Issue #2710, opened 2022-08-28
[Bear] Search query should be tokenized


jrk:
# Extension – [Bear](https://github.com/raycast/extensions/tree/main/extensions/bear)

Author: @hmarr @lucaschultz

**Describe the feature and the current behavior/state.**
Bear note search currently treats the entire query string as a single token. Coming from most modern search interfaces (including Bear's own search), this is surprising and limiting behavior, as it prevents searching for things based on multiple tokens which occur independently, or not necessarily in the same order.

The search should be tokenized according to common practice (e.g., with support for `"quoted strings"` where a longer substring should be treated as a single token).

(There is currently a workaround where one can explicitly search for `foo%bar`, relying on the SQL wildcard being passed directly to the underlying SQLite query, but this is likely not expected or discoverable for most users.)

jrk:
I should clarify: even the wildcard workaround isn't a full tokenized search, as it still constrains the order of the tokens.

hmarr:
Yeah this occasionally bugs me too. Unfortunately it's not trivial to solve. I haven't put the time into improving it yet because I'm mostly searching by title, so I don't find it a huge problem day to day.

Anyway, I did have a quick think about this when I was originally building the extension, and here are a few options I came up with.

### Use the sqlite [FTS5](https://www.sqlite.org/fts5.html) module

This builds full text search right into sqlite, so on the surface is the most straightforward option available to us.

However, the Bear sqlite database doesn't have a full text search table already, so we'd need to create it. So far we've been strict about avoiding any writes to the Bear sqlite database. The worst thing this extension could do is lose someone's notes by corrupting the database. Reading the database is unlikely to cause any harm, but as soon as we start modifying the database we're in uncharted territory.

To work around this, we could periodically copy the database to another place in the filesystem, and create the search table there. Periodically recreating the search table from scratch might not be great for performance, but I'd want to try it out before ruling it out.

Additionally, the sqlite WASM library we're using (sql.js) doesn't appear to come with a distribution that bundles the FTS5 module, but it does look like other extensions (craftdocs, ensk-is) have found a distribution that does include the module.

### Build our own index

Alternatively we could extract all the notes from the database, and load them into our own search index, perhaps using something like [tinysearch](https://github.com/tinysearch/tinysearch) or [stork](https://github.com/jameslittle230/stork). The sqlite database does include modification dates, so we could probably use them to keep the index up to date incrementally.

jrk:
[...]

PR #6729, opened 2023-05-22
Update bear extension

juanbanchero:
## Description

- Adding a filter in search notes to filter by tag. It lists all your tags and it narrows the search base on the selected tag.
- Change the command `Cmd+p` from preview to filter in the search bar in order to follow Raycast conventions
- Add `Cmd+shift+p` to trigger the preview

## Screencast

https://github.com/raycast/extensions/assets/26557565/bc31d6dd-1ffc-4684-89b0-6af5f8ec061f

## Checklist

### DIFF
--- a/extensions/bear/src/db-queries.ts
+++ b/extensions/bear/src/db-queries.ts
@@ -32,7 +32,7 @@ WHERE
 GROUP BY
   notes.ZUNIQUEIDENTIFIER
 HAVING
-(:tag IS NULL OR group_concat('#' || tags.ZTITLE || '#') LIKE '%#' || :tag || '#%')
+  (:tag IS NULL OR group_concat('#' || tags.ZTITLE || '#') LIKE '%#' || :tag || '#%')
 ORDER BY
   -- Sort title matches ahead of body matches
   CASE WHEN (


pernielsentikaer:
Looks like a nice update @juanbanchero 🔥 

@hmarr do you mind checking this?

extensions/bear/src/index.tsx
```diff
@@ -25,6 +26,7 @@ export default function SearchNotes(props: LaunchProps<{ arguments: SearchNotesA
   }
 
   const showDetail = (notes ?? []).length > 0 && getPreferenceValues().showPreviewInListView;
+  const handleTagChange = (tag: string) => setNotes(db?.getNotes(tag));
```
hmarr, line 29: I think you'll want the `#` prefix / suffix to filter _only_ by the tag, not just for all occurrences of the tag's name. E.g.:

```suggestion
  const handleTagChange = (tag: string) => {
    const query = tag.includes(" ") ? `#${tag}#` : `#${tag}`;
    setNotes(db?.getNotes(query));
  }
```

hmarr: Additionally, calling `setNotes` here will mean that when the user modifies the search, the tag filter will no longer apply.

One solution would be to track the tag filter in a new `useState` variable, add that to the dependency array for the `useEffect` call further up this component, then modify the `getNotes` function to accept a tag that we inner join against. Unfortunately simply appending the tag to the existing query won't work because we match the entire query string rather than tokenizing the query (see https://github.com/raycast/extensions/issues/2710).

Alternatively we could just disable the search box when you're filtering, but that's not ideal!

hmarr (commented):
Nice work @juanbanchero! I've spotted a couple of issues with the way the tag filtering that we'll need to resolve before merging, but it looks good apart from that.

extensions/bear/src/index.tsx
juanbanchero: Hey @hmarr! Thanks for taking a look and spotting some issues. I'm fixing all of them right now. The only problem that I have is that the when no tag is selected, let's say, in the `useEffect` at the beginning, instead of showing me all the notes, it shows me the untagged notes.

This is the modified version of the getNotes 
```ts
  getNotes(searchQuery: string, tag?: string): Note[] {
    const statement = this.database.prepare(
      this.getBearVersion() === 2 ? BearDb.searchNotesQueries.v2 : BearDb.searchNotesQueries.v1
    );
    if (tag) {
      statement.bind({ ":query": `${searchQuery}`, ":tag": `${tag}` });
    } else {
      statement.bind({ ":query": `${searchQuery}` });
    }
    const results: Note[] = [];
    while (statement.step()) {
      const row = statement.getAsObject();
      results.push(this.toNote(row));
    }

    statement.free();
    return results;
  }
```
And this is the modified query 
```sql
SELECT
  notes.ZUNIQUEIDENTIFIER AS id,
  notes.ZTITLE AS title,
  notes.ZTEXT AS text,
  notes.ZMODIFICATIONDATE AS modified_at,
  notes.ZCREATIONDATE AS created_at,
  group_concat(tags.ZTITLE) AS tags,
  notes.ZENCRYPTED AS encrypted
FROM
  ZSFNOTE AS notes
LEFT OUTER JOIN
  Z_5TAGS AS notes_to_tags ON notes.Z_PK = notes_to_tags.Z_5NOTES
LEFT OUTER JOIN
  ZSFNOTETAG AS tags ON notes_to_tags.Z_13TAGS = tags.Z_PK
WHERE
  -- When there is a query, filter the body by that query, otherwise
  -- ignore the query
  (
    lower(notes.ZTITLE) LIKE lower('%' || :query || '%')
    OR lower(notes.ZTEXT) LIKE lower('%' || :query || '%')
    OR :query = ''
    OR lower(notes.ZUNIQUEIDENTIFIER) LIKE lower('%' || :query || '%')
  )
  -- When there is a tag query, filter the tags by that query, otherwise
  -- ignore the tag query
  AND (:tag = '' OR tags.ZTITLE LIKE lower('%' || :tag || '%'))  
  -- Ignore trashed, archived, and empty notes
  AND notes.ZARCHIVED = 0
  AND notes.ZTRASHED = 0
  AND (
    notes.ZTEXT IS NOT NULL
    OR notes.ZENCRYPTED = 1
  )
GROUP BY
  notes.ZUNIQUEIDENTIFIER
ORDER BY
  -- Sort title matches ahead of body matches
  CASE WHEN (
    lower(notes.ZTITLE) like lower('%' || :query || '%')
    OR :query = ''
  ) THEN 0 ELSE 1 END,
  -- When there are multiple title matches, sort by last modified
  notes.ZMODIFICATIONDATE DESC
LIMIT
  20
```
I don't know if there is any solution to this query or we just need to make a new query for this situation

hmarr: Could it be that you're not passing the `:tag` parameter in at all if it's `undefined`, meaning it won't be equal to the empty string? If that's not it, I'm happy to take a look at locally if you want to push an update to the branch.

Additionally, I just realised that if we filter by tag using the `WHERE` clause, we'll only include the tag we're filtering by in the `group_concat(tags.ZTITLE)` column, which means the note preview won't show other tags the note has. Given this, it might be better to move the tag filter to a `HAVING` clause, or to use a CTE to separate the joining on tags from the tag filtering?

extensions/bear/src/index.tsx
```diff
@@ -13,18 +14,26 @@ export default function SearchNotes(props: LaunchProps<{ arguments: SearchNotesA
...
   }
 
   const showDetail = (notes ?? []).length > 0 && getPreferenceValues().showPreviewInListView;
+  const handleTagChange = (tag: string) => {
+    let formattedTag = "";
+    if (tag !== formattedTag) {
+      formattedTag = tag.includes(" ") ? `#${tag}#` : `#${tag}`;
+    }
+    setSelectedTag(formattedTag);
+  };
```
hmarr, line 36: Given we've switched to joining against the tags table rather than searching within the note's content, we can go back to using the tag name without the `#` signs.

```suggestion
  const handleTagChange = (tag: string) => setSelectedTag(tag);
```

extensions/bear/src/search-dropdown.tsx
```diff
@@ -0,0 +1,49 @@
...
+  if (error) {
+    showToast(Toast.Style.Failure, "Something went wrong", error.message);
+  }
+  return (
+    <List.Dropdown
+      tooltip="Select a tag"
+      storeValue={true}
+      onChange={(newValue) => {
+        if (newValue === "All tags") {
+          onTagChange("");
```
hmarr, line 30: It might be clearer to explicitly use `null` to represent the absence of a tag, rather than an empty string? You'd need to change the type definition to `string | null`.

juanbanchero:
@hmarr I believe all the changes were implemented. Now is working properly and the code is updated. I tested on Bear 2 and it's working correctly 😄
Let me know if you want to change something else, but I believe that we are ready to launch 🚀

extensions/bear/src/db-queries.ts
```diff
@@ -76,6 +78,8 @@ WHERE
   )
 GROUP BY
   notes.ZUNIQUEIDENTIFIER
+HAVING
+(:tag IS NULL OR group_concat('#' || tags.ZTITLE || '#') LIKE '%#' || :tag || '#%')
```
hmarr, line 82:
```suggestion
  (:tag IS NULL OR group_concat('#' || tags.ZTITLE || '#') LIKE '%#' || :tag || '#%')
```

extensions/bear/src/db-queries.ts
```diff
@@ -31,6 +31,8 @@ WHERE
...
```
hmarr, line 35:
```suggestion
  (:tag IS NULL OR group_concat('#' || tags.ZTITLE || '#') LIKE '%#' || :tag || '#%')
```

hmarr (approved):
One more super minor indentation fix, but apart from that this looks great!

Nice work @juanbanchero 🚀

pernielsentikaer:
Thanks for the review @hmarr 🔥

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #6745, opened 2023-05-23
[OpenAI Translator] Raycast AI and other LLM provider support

### LINKED ISSUE #5738: [OpenAI Translator]  It would be great to add a two-way translation feature.
### Extension

https://www.raycast.com/douo/openai-translator

### Description

For example, as a native Chinese speaker, I can set my primary language to be Chinese and also set a target language. For instance, if I want my input to be translated into English, then I can set the target language as English. When I type in a non-Chinese language, the system will automatically recognize the language and translate it into the target language which is English in this case. If I type in my native language which is Chinese, then the system will default to translating it into my chosen target language which is English.

I think the majority of users use translation tools mainly to translate unfamiliar languages. With this feature, users can input in their familiar language and get the another translation result. This issue was translated from Chinese to English through ChatGPT. If this feature can be added, it will allow users like me to achieve this goal more efficiently. 

### Anything else?

_No response_

douo:
## Description

- Raycast AI support by @Fatpandac
- Azure OpenAI Service  support
- PaLM 2 limited support
- Using build-in NLLanguageRegcondizer for language detection
- Auto exchange target and source when translate English and Chinese. 
  - Fixed #5738 
- Fixed incomplete display of translation

## Screencast
configuration

## Checklist


extensions/openai-translator/src/providers/prompt.ts
```diff
@@ -0,0 +1,245 @@
...
+    const rolePrompt = "You are a professional text summarizer, you can only summarize the text, don't interpret it.";
+    const commandPrompt = `summarize this text in the most concise language and must use ${prompt.meta.targetLang} language!`;
+    return { ...prompt, rolePrompt, commandPrompt };
+  },
+
+  // What
+  ["what"]: (prompt: Prompt) => {
+    const rolePrompt = "You are a identifier, you can only response on markdown format.";
+    const commandPrompt = prompt.meta.toChinese
+      ? `请按照 markdown 的格式回答，Section有Maybe和Desc，Maybe回答他最可能是的东西（要求精确些），Desc回答这个东西的描述; 答案应该使用中文。`
```
pernielsentikaer, line 241: At the moment, Raycast doesn't support localization and only supports US English. Therefore, please avoid introducing your custom way to localize your extension.

extensions/openai-translator/src/providers/prompt.ts
```diff
@@ -0,0 +1,245 @@
...
+        commandPrompt = `夹在${quoteProcessor.quoteStart}和${quoteProcessor.quoteEnd}之间的内容是原文，请您将原文内容翻译成《红楼梦》风格的近代白话文`;
+      } else if (content.length < 5 && toChinese) {
+        // 当用户的默认语言为中文时，查询中文词组（不超过5个字），展示多种翻译结果，并阐述适用语境。
+        rolePrompt = `你是一个翻译引擎，请将给到的文本翻译成${targetLang}。请列出3种（如果有）最常用翻译结果：单词或短语，并列出对应的适用语境（用中文阐述）、音标、词性、双语示例。按照下面格式用中文阐述：
+                    <序号><单词或短语> · /<音标>
+                    [<词性缩写>] <适用语境（用中文阐述）>
+                    例句：<例句>(例句翻译)`;
+        commandPrompt = "";
+      }
+    }
```
pernielsentikaer, line 218: At the moment, Raycast doesn't support localization and only supports US English. Therefore, please avoid introducing your custom way to localize your extension.

pernielsentikaer (changes requested):

extensions/openai-translator/src/providers/prompt.ts
douo: Hi @pernielsentikaer,  This is not a localization . Instead, it is a prompt to optimize the quality of Chinese translation.  It is invisible to users.

pernielsentikaer (approved):

PR #6764, opened 2023-05-25
[Bing Search]fixed: Chinese garbled characters in the search list

marlkiller:
## Description

[Bing Search]fixed: Chinese garbled characters in the search list

There's still something wrong with the first default item in the list, but I'm not sure how to fix it

## Screencast

WechatIMG521


pernielsentikaer:
@marlkiller Thanks for the update 🔥 Do you mind adding a changelog entry?

is this something you should look into?

Raycast-ULfz0E9Y@2x

PR #6799, opened 2023-05-26
[kill-process] feat: sort by memory & aggregate apps

### LINKED ISSUE #6431: [Kill Process] display process ordered by memory usage
### Extension

https://www.raycast.com/rolandleth/kill-process

### Description

Sometimes we not only need to kill process that cost too much cpu, but also too much memory, such as memory leak(and in these cases, process do not often cost too much cpu). I hope we can display memory usage per process and ordered like cpu usage.

all users using this extension

### Anything else?

Not only list the process in activity monitor, but also provide a mode to aggregate process that is actually from a same app, like open multi tabs in chrome, do not show multi "chrome helper", instead show a "chrome" with the sum up cpu or memory usage.

Saafo:
## Description

closes #6431 

## Checklist


Saafo:
@rolandleth @crazyones110 Sorry for the interruption, I've add some features to `kill-process` as #6431 described, please have a review.

kill-process-1
kill-process-2
kill-process-3

Saafo:
This pr is ready for review.

pernielsentikaer:
Dop you have time to check this @rolandleth

Saafo:
Currently aggregating apps by path is not always accurate, I'm working on a more robust way to aggregating and killing apps.

extensions/kill-process/src/index.tsx
```diff
@@ -117,20 +172,46 @@ export default function ProcessList() {
...
                     shortcut={{ key: "r", modifiers: ["cmd"] }}
                     onAction={() => fetchProcesses()}
                   />
+                  <Action
+                    title={`Sort by ${sortByMem ? "CPU" : "memory"} usage`}
+                    icon={Icon.Filter}
+                    shortcut={{ key: "tab", modifiers: [] }}
+                    onAction={() => {
+                      setSortByMem(!sortByMem);
+                      showToast({
```
rolandleth, line 199: I'm not 100% sure about changing these flags from actions. The reason for my uncertainty is that these don't actually change the extension's preference, they just update flags for this session, which might be confusing to the user. @pernielsentikaer, what do you folks think?

extensions/kill-process/src/index.tsx
```diff
@@ -88,23 +141,25 @@ export default function ProcessList() {
     >
       {state
         .filter((process) => {
-          if (query == null) {
+          if (query == "" || query == undefined) {
```
rolandleth, line 144: I think `== null` checks for `undefined` as well, but `== undefined` doesn't check for `null` too.

extensions/kill-process/src/index.tsx
```diff
@@ -117,20 +172,46 @@ export default function ProcessList() {
           return (
             <List.Item
               key={index}
-              title={process.name}
+              title={(process.type == "aggregatedApp" ? process.appName : process.processName) ?? ""}
```
rolandleth, line 175: I don't understand how would this fallback on `?? ""`, isn't `process.type == "aggregatedApp" ? process.appName : process.processName` enough?

rolandleth:
Hey, @Saafo, I looked over the PR and the improvements look really nice, great ideas!

extensions/kill-process/src/index.tsx
pernielsentikaer: I can see why it could be nice to be able to change this on the fly but then I don't think there should be anything in preferences and then maybe use `useCachedState` to cache the values between loads

Saafo:
Hey @rolandleth , I've change the aggregating algorithm to grouping by pid and ppid, instead of by path, because there are cases that:

* There are multiple independent processes inside a same app, like Xcode process and Simulator process
* Cannot always kill Apps by their names, like using Xcodes to download a `Xcode-14.2.0.app` and a `Xcode-14.3.0.app` to /Application, the app name is `Xcode-14.2.0` and `Xcode-14.3.0 but the process name are both `Xcode`.

Currently, the algorithm is:

* Filter processes which path includes `.app/`.
* build a tree by pid and ppid.
* get the processes directly connect with launchd(pid=1) as the root processes, and aggregating child processes into them.

And finally I've decide to keep `processName` and `appName` because there are cases like `Xcode-14.2.0.app` and a `Xcode-14.3.0.app`, that both `processName` are `Xcode`, you have to differentiate them by `appName`.

Saafo:
@rolandleth @pernielsentikaer hello again! Would you like to review this pr? It's ready!

And @zhenpewu please review the changes I've made to `useInterval`, I'm a newbie to react, I'm not sure whether I changed it correctly.

extensions/kill-process/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Kill Process Changelog
 
+## [Sort by memory and App aggregation] - 2023-06-15
+
+- Now can sort processes by CPU or memory usage
+- Now can aggregate processes of the same app
```
zhenpewu, line 6: nit: add new line

extensions/kill-process/src/index.tsx
```diff
@@ -14,43 +15,61 @@ export default function ProcessList() {
   const shouldShowPID = preferences.shouldShowPID ?? false;
   const shouldShowPath = preferences.shouldShowPath ?? false;
   const refreshDuration = +preferences.refreshDuration;
+  const [sortByMem, setSortByMem] = useState<boolean>(preferences.sortByMem ?? false);
+  const [aggregateApps, setAggregateApps] = useState<boolean>(preferences.aggregateApps ?? false);
 
   const fetchProcesses = () => {
-    exec(`ps -eo pid,pcpu,comm | sort -nrk 2,3`, (err, stdout) => {
+    exec(`ps -eo pid,ppid,pcpu,rss,comm`, async (err, stdout) => {
```
zhenpewu, line 22: there is no await in the following fn, don't think this async is needed.

extensions/kill-process/src/index.tsx
```diff
@@ -14,43 +15,61 @@ export default function ProcessList() {
...
+          return a.cpu > b.cpu ? -1 : 1;
+        }
+      });
 
       setState(processes);
     });
   };
 
-  useInterval(fetchProcesses, refreshDuration);
+  useInterval(fetchProcesses, refreshDuration, [sortByMem, aggregateApps]);
```
zhenpewu, line 65: I don't think this kind of aggregate operation need to trigger refresh.

I mean it could aggregate based on the current result only, while the current change is updated, we just aggregate again.

extensions/kill-process/CHANGELOG.md
Saafo: Ok! My fault in resolving conflicts lol

extensions/kill-process/src/index.tsx
```diff
@@ -61,25 +80,99 @@ export default function ProcessList() {
...
+  };
 
-    return subtitle;
+  const aggregate = (processes: Process[]): Process[] => {
+    const result = Array<Process>();
+    type ProcessNode = {
+      process: Process | undefined;
+      childNodes: ProcessNode[];
+    };
+    const appMap = new Map<number, ProcessNode>();
```
zhenpewu, line 106: nit: maybe lodash groupby might be helpful?

Saafo:
I'm not sure whether to add icon
image

extensions/kill-process/src/index.tsx
Saafo: I'll get to know that

extensions/kill-process/src/index.tsx
Saafo: It can be done in this way, but I think it's unnecessary to remember the old value, since aggregating and sortByMem is not perform too often, and we can get the latest result, and refreshing seems not cost much time. @rolandleth what do you think?

extensions/kill-process/src/index.tsx
Saafo: I forgot to delete it, will remove it soon!

extensions/kill-process/src/index.tsx
zhenpewu: I don't think we need keep old value, I think we just need one more useMemo/useEffect like

```
useEffect( aggregate Fn, [ execResult ])
```

this execResult is only  a reference, whenever this ref changed, it will trigger aggregate Fn.

Saafo: Got it, I'll have a look in it.

extensions/kill-process/src/index.tsx
Saafo: It seems `groupby` will have less help to the aggregating algorithm, because it will not only aggregate by parent id, but also grandparent and grand-grandparent id... It's recursive and using `groupby` seems will have much more complexity with less benefit.

extensions/kill-process/src/index.tsx
Saafo: I'm not sure what do you mean by `useEffect( aggregateFn, [ execResult ])`, what is the `execResult`? The `aggregateFn` is irreversible, without keeping the old value, how do we dis-aggregate by the already aggregated value?

zhenpewu: First, in the first, usePromise / useCachePromise (exec( ps ...)), will return an execResult, which is basically pure list of process.

Then, if aggregation flag is true, aggregation is performed and render aggregated results, if not, render the list of process.

I think you can consider execResult as "the old value".

My main concern is that

```
  useEffect(() => {
    savedCallback.current();
    const refreshEnabled = (delay ?? 0) > 0;
    if (refreshEnabled) {
      const interval = Math.max(delay ?? 0, 1000);
      const id = setInterval(() => savedCallback.current(), interval);
      return () => clearInterval(id);
    }
  }, [delay]);

```

as you can see, previously, the only deps in delay, which won't be changed per command invocation, which means this useEffect is triggered once, and this useEffect is to create a scheduled refresh. 

If more deps are introduced, more scheduled process will be introduced as well, and most importantly, which is not necessary.

Saafo: I want to make sure that you aware that `aggregateApps` may changed during per command invocation by the action menu? If I use `useEffect( aggregateFn, [ execResult ])` and `aggregateApps` changed, the `state` won't update. If you mean `useEffect( aggregateFnIfNeeded, [setIntervalResult, aggregateApps, sortByMem])`, then I have to store both `setIntervalResult` and the result in aggregateFnIfNeeded.

If I introduce deps to `useInterval`, I think there will be only one interval loop at same time because I've tested it, there will still refresh every 3 seconds after I changed the `aggregateApps` or `sortByMem` value, and in my understanding, the `() => clearInterval(id);` will be executed after changes in [delay, aggregateApps, sortByMem] happened, which will clear last interval loop.

zhenpewu: I know it is triggered by action.

```

const execResult;
const isAggregateEnableld = false

useInterval((), [delay]);

const aggregateResult = useMemo ( aggregateFn, [execResult, isAggregateEnableld])
```

Saafo: > then I have to store both setIntervalResult and the result in aggregateFnIfNeeded.

Yes I thought there were some way not storing two value

extensions/kill-process/src/index.tsx
Saafo: Changed to `if (!query)`, is it acceptable?

extensions/kill-process/src/index.tsx
Saafo: Resolved by adding `fetchResult` property to not refresh on change of sortByMem or aggregateApps

extensions/kill-process/src/index.tsx
```diff
@@ -88,23 +141,25 @@ export default function ProcessList() {
     >
       {state
         .filter((process) => {
-          if (query == null) {
+          if (query == "" || query == undefined) {
             return true;
           }
 
-          const nameMatches = process.name.toLowerCase().includes(query.toLowerCase());
+          const name = process.type == "aggregatedApp" ? process.appName : process.processName;
```
rolandleth: When comparing, it's usually preferred to use strict compare.

extensions/kill-process/src/index.tsx
Saafo: Sounds reasonable, I'll fix it soon!

extensions/kill-process/src/index.tsx
Saafo: Got it, will fix it soon!

zhenpewu (commented):
otherwise, lgtm.

extensions/kill-process/package.json
```diff
@@ -70,7 +87,9 @@
     }
   ],
   "dependencies": {
-    "@raycast/api": "^1.52.1"
+    "@raycast/api": "^1.53.3",
+    "node-fetch": "^3.0.0",
```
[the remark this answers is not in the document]
Saafo: Removed it!

zhenpewu (approved):

rolandleth:
One last comment and then it's good to go, from my point of view: I might have explained it poorly, but shallow compares should be done against `null`, not `undefined`.

```ts
const x = null
const y = undefined

x == null // true
x == undefined // true

y == null // false -> this is the problem, basically
y == undefined // true
```

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #6912, opened 2023-06-02
Add 12ftio Extension

GregorMcC:
## Description

A simple extension that allows you to generate 12ft.io links for removing paywalls from websites. Designed to persist a list of generated links that can be searched later on.

## Screencast
12ft-1
12ft-2
12ft-3

## Checklist


pernielsentikaer (changes requested):
I might not be the right audience for this extension but is there a reason to keep the list of previously entered URLs?

You could maybe see how I do in [this extension](https://www.raycast.com/pernielsentikaer/wayback-machine) where I check clipboard/selected text too and doing an action 🙂

pernielsentikaer:
Could also be a possibility to let #6925 superseed this

GregorMcC:
> Could also be a possibility to let #6925 superseed this

💯 if we feel that the single action/command is preferred I'd be OK to either add that in here or let #6925 go ahead 👍

pernielsentikaer:
Are you still around @GregorMcC

GregorMcC:
> Are you still around @GregorMcC

@pernielsentikaer the extension here is just a starting point - I'd like to add tagging/grouping by domain in a future addition, for example allowing me to see all the `medium.com` articles at a glance.

I thought it'd be nice to be able to easily + quickly filter through all of the articles I have found interesting enough to warrant removing the paywall!

tegola:
I feel this extension is presented to be much more than a paywall remover, and instead act like a database with links without a paywall. I'm not into this kind of siloization of content, but also not against it. My concern is that it's a completely different approach to my extension, where you just remove the paywall from a URL (actually just from the selected text, but I plan to add clipboard and url from the current browser tab).

Also, I did avoid _branding_ the extension with the 12ft name (since the service can be selected by the user), instead opting for a more generic "remove paywall" paired with a open lock icon.

So, if we agree to remove the 12ft branding, then the two extensions could be merged into a single, co-authored one with multiple commands.

pernielsentikaer:
I think that would be the best solution, having one extension with a quick way and the "database" way if it makes sense 🙂

GregorMcC:
@pernielsentikaer @tegola appreciate the feedback on this. I will close this PR and allow #6925 to supersede, I can then open a PR to add the "database" logic to that extension at a later date when I have some spare time again.

GregorMcC:
See above

PR #6923, opened 2023-06-03
Add auto-work extension

weitanai:
## Screencast
summary

improve

## Checklist

### DIFF
--- a/extensions/auto-work/README.md
+++ b/extensions/auto-work/README.md
@@ -2,9 +2,7 @@
 
 # Motivation
 
-- Clear and concise writing
-- Attention to detail
-- Creative problem solving
+- Clear and concise writing, Attention to detail
 - Collaborative approach to work
 - Adaptable and flexible mindset
 - Time management and prioritization skills
--- a/extensions/auto-work/package.json
+++ b/extensions/auto-work/package.json
@@ -2,7 +2,7 @@
   "$schema": "https://www.raycast.com/schemas/extension.json",
   "name": "auto-work",
   "title": "Auto work",
-  "description": "Interact with OpenAI's auto-work API directly from your command bar",
+  "description": "Interact with OpenAI's API  directly from your command bar",
   "icon": "icon.png",
   "author": "weitanai",
   "homepage": "https://github.com/weitanai/auto-work#readme",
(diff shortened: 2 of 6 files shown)


pernielsentikaer (commented):
Hi 👋 

We do already have a lot of [GPT extensions](https://www.raycast.com/search?q=openai) in the store which also cover what this does (I think)

Could you check them out, it's also possible to enhance the current store versions

weitanai:
what did I do
Provide a better product experience
Find the balance between the minimum key value and the cache hit rate

pernielsentikaer:
You didn't do anything wrong, we're just trying to limit extension that does almost the same, there a few more that does nearly the same as this on [this list](https://www.raycast.com/search?q=chatgpt) too 🙂

weitanai:
I hope that my project can help others, and I am also recommending my classmates to use the extension I developed, which is of great significance for me to continue to study computer science

pernielsentikaer:
It would still be better to enhance one of the current ones, like the one @alanzchen made (https://www.raycast.com/alanzchen/chatgpt-quick-actions)

What do you think @alanzchen, could this be a good idea to add?

alanzchen:
Yeah i'd well come any PR that improves the performance of the extension.

I'm not sure if I'm missing something, but it seems both extension is targeted towards one-off actions that are not likely to be repeated again in the future. So I'm not sure how caching would be useful.

weitanai:
> Yeah i'd well come any PR that improves the performance of the extension.
> 
> I'm not sure if I'm missing something, but it seems both extension is targeted towards one-off actions that are not likely to be repeated again in the future. So I'm not sure how caching would be useful.

In my own use case
Sometimes: I have already summarized an article, but something happened temporarily, and this operation was interrupted, so I did not copy the good results. If there is no cache, I rewrite and summarize an article. Here, the waiting time is related to the repeated consumption of tokens.

pernielsentikaer:
I guess a simple "History" command would do the same, right?

weitanai:
> I guess a simple "History" command would do the same, right?

Yep，

Sometimes, Users don't want to search in the history, they want  straight forward   interaction

pernielsentikaer:
Besides that one feature it doesn't look like there are any differences between them 😞 I would still say that it would be better to enhance the current if it makes sense instead of having two identical extensions since it will confuse the end user

Issue #6559, opened 2023-05-15
[Folder Search] ... No folders are displayed


MichaelHHeng:
### Extension

https://www.raycast.com/GastroGeek/folder-search

### Description

Whenever I enter more than one character in the search, “No results” is displayed, and if I enter only one character, no matter which one, I always get the same list of files deep from the system. Permissions for Raycast have been granted, “Search Files” works normally, also finds Folder, but I would like to use the extended capabilities (Copy, Move of files selected in Finder) of Folder Search.
![cleanshot_2023-05-04_at_12 06 37_2x](https://github.com/raycast/extensions/assets/122787238/a19179fe-dfb3-455c-969d-7b10c29aa8cb)
![CleanShot 2023-05-15 at 15 46 38@2x](https://github.com/raycast/extensions/assets/122787238/1cea0b1e-9fcd-4a9c-ae39-5fced0202b11)

### Steps To Reproduce

MacBook Pro M1 Pro, Venture 13.31.1 (a), Raycast 1.51.1
Start folder-search in raycast and type one or two character.

MichaelHHeng:
There is another topic with this issue, maybe this entry should be deleted?

pernielsentikaer:
Merged and closed, please give it a shot @MichaelHHeng  - you can run `Check for extension Updates` to ensure you have the newest update

MichaelHHeng:
Tested and working! Thank you very much!

omarshahine:
@MichaelHHeng Wohoo! So happy to hear!

Issue #6333, opened 2023-05-04
[Folder Search] Query string is duplicated when Folder Search is opened as a fallback command.


nukesean:
### Extension

https://www.raycast.com/GastroGeek/folder-search

### Description

When opening Folder Search as a fallback command for a query, it duplicates the query text. [See example below](https://imgur.com/a/kOghQ8t):
<p>
<a href="https://imgur.com/a/kOghQ8t"><img src="https://i.imgur.com/7C9rc8D.png"></a>

### Steps To Reproduce

1. Enter a search query in Raycast
2. Select Folder Search as the fallback command
3. Folder Search opens with the query duplicated

### Current Behaviour

The query string is duplicated when opening Folder Search as a fallback command.

### Expected Behaviour

Folder Search should not open a search with the initial query duplicated.

omarshahine:
@nukesean  cannot repro this (but saw this happen exactly once today, and am not sure how or why it happened).

Is this consistent repro?

nukesean:
@omarshahine Really? That's surprising, because yes, it is consistent for me. It happens 100% of the time on all three of my computers: an Intel iMac, an M1 Mac Mini, and an M1 MacBook Air, all running macOS Ventura.

omarshahine:
So I was able to get it to repro only when there are "no results".

question, do you get any results at all from Folder Search when searching (non fallback)? 

Did you try and change your search scope from This Mac to your use profile?

It seems to be an issue on first use.

nukesean:
Yes, the command works perfectly fine except for this issue; i.e., it returns the desired results if I (1) launch it directly and then enter my search, or (2) launch it as a fallback command and then manually delete the duplicated text.

The duplication from fallback happens regardless of whether there are "no results." For example, I just searched for "dropbox" as a fallback, and it once again produced "dropboxdropbox."

Yes, I've switched the scope back and forth from This Mac to my user profile several times since I first installed it almost a year ago.

omarshahine:
I have a bug fix for this that I am testing. It's now happening to me a lot, the fix is a hack but it seems to work.

Is anyone here willing to test it out?

pernielsentikaer:
Merged and closed, please give it a shot @nukesean - you can run `Check for extension Updates` to ensure you have the newest update

pernielsentikaer:
I can see that, same happens here - we'll take a look 🙂

omarshahine:
@pernielsentikaer 

[...]

Issue #4391, opened 2023-01-23
[Folder Search] Extension not working


andreaguida:
### Extension

https://www.raycast.com/GastroGeek/folder-search

### Description

Whetever I look for, there are no results.

https://user-images.githubusercontent.com/54178949/214118392-856eeff4-e6b9-4687-a1ec-6ba8c2666e12.mp4

omarshahine:
I wonder if this is a spotlight issue. Does File search work for you?

andreaguida:
@omarshahine yeah, Spotlight and the Raycast File Search feature are working perfectly fine. Thanks!
![CleanShot 2023-01-23 at 22 57 33@2x](https://user-images.githubusercontent.com/54178949/214159113-f7e5c2ae-ef85-4d9d-9531-69490fcc1c5e.png)

omarshahine:
@GastroGeek might have ideas as he authored 99% of the extension including this piece.

andreaguida:
FYI: extension started working after I've set macOS in English language (instead of Italian).

MichaelHHeng:
I have the same Problem, German language, same as @flegfleg i don't want to change to english. Thanx @GastroGeek and @omarshahine 😊

omarshahine:
Are any of willing to install https://github.com/derhuerst/node-spotlight and see if you have the same problem?

The extension is just a wrapper for node-spotlight.

MichaelHHeng:
I have looked at this and honestly have no idea what it is, how to install it and what will happen after installation and/or how to test it. Even less do I know what a wrapper for node-spotlight is. Sorry, I am only a user and unfortunately completely lost 🤷‍♂️

omarshahine:
I found the problem, now I have to fix it.

@raycastbot can you merge this issue with https://github.com/raycast/extensions/issues/6559

omarshahine:
I have a fix and am now testing it out.

pernielsentikaer:
Merged and closed, please give it a shot @andreaguida  - you can run `Check for extension Updates` to ensure you have the newest update

PR #7032, opened 2023-06-11
[Folder-Search] Fix for issues 4391, 6333, 6333

omarshahine:
## Description

Fixing Three Issues / Two bugs
- Extension only worked in English. Fixed up code to be language agnostic
- Sometimes the fallback invocation resulted in the search query being duplicated so "Query" would appear "QueryQuery". This is not consistent, but I added some code to deal with this issue.

https://github.com/raycast/extensions/issues/6333
https://github.com/raycast/extensions/issues/6559
https://github.com/raycast/extensions/issues/4391

## Checklist

### DIFF
--- a/extensions/folder-search/CHANGELOG.md
+++ b/extensions/folder-search/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Folder Search Changelog
 
-## [Command rename] - {PR_MERGE_DATE}
+## [Command rename] - 2023-06-13
 Fixing Three Issues / Two bugs
 - Extension only worked in English. Fixed up code to be language agnostic
 - Sometimes the fallback invocation resulted in the search query being duplicated so "Query" would appear "QueryQuery". This is not consistent, but I added some code to deal with this issue.


omarshahine:
@raycastbot I think @GastroGeek is no longer active. He has not piped up for many months.

extensions/folder-search/src/search.tsx
```diff
@@ -55,6 +56,14 @@ export default function Command() {
   const [isShowingDetail, setIsShowingDetail] = useState<boolean>(true);
   const [results, setResults] = useState<SpotlightSearchResult[]>([]);
 
+  // hack to fix annoying double text during fallback. Typing helloworld results in helloworldhelloworld
```
fengtianze, line 59: I also found this bug, but don't know how to reproduce it stably. If you know, I can check if it is a problem caused by the React component state.

fengtianze: The callback parameter below looks a bit suspicious, as it will cause searchSpotlight to be executed every time the UI is updated.

```
usePromise(
    searchSpotlight,
    [
      searchText,
      searchScope,
      abortable,
      (result: SpotlightSearchResult) => {
        setResults((results) => [result, ...results].sort(lastUsedSort));
      },
    ],
   ...
)
```
--->
```
const callback = useCallback(
  (result: SpotlightSearchResult) => {
        setResults((results) => [result, ...results].sort(lastUsedSort));
   },
)

usePromise(
    searchSpotlight,
    [
      searchText,
      searchScope,
      abortable,
      callback
    ],
   ...
)
```

omarshahine: That is the original code though. I just patched it up to fix these two bugs.

pernielsentikaer: Is this something we could patch too, sounds like something that would be a good idea?

omarshahine: Oh I was looking at this on my phone and did not realize there was a proposed change from @fengtianze. Testing this now and if it works I'll update the PR.

omarshahine: No luck with this code. Note the code as is did not work:

  ```
const searchCallback = useCallback(
    (result: SpotlightSearchResult) => {
          setResults((results) => [result, ...results].sort(lastUsedSort));
     },
     [] //I needed to add this as I was getting an error
  )
```

fengtianze: I just tried debugging, but couldn't reproduce the issue. (I've encountered this a few times before.)

I found that the rendering performance of this extension has optimization space, it re-renders the UI every time it searches a directory. Sadly, I am a frontend developer and I don't know how to optimize `node-spotlight`.

pernielsentikaer:
Should we just release as is and see later if we can track down the bug, are you two okay with that @fengtianze @omarshahine

fengtianze (approved):

omarshahine:
> **Published to the Raycast Store:**
> https://raycast.com/GastroGeek/folder-search

Thanks. I’ll keep an eye out and see if there is an opportunity to improve this down the road. I don’t quite know how to go about addressing the issue but at least it has fewer bugs now.

PR #7038, opened 2023-06-11
Add laravel-tips extension

godruoyi:
## Checklist

### DIFF
--- a/extensions/laravel-tips/package.json
+++ b/extensions/laravel-tips/package.json
@@ -14,19 +14,19 @@
   "commands": [
     {
       "name": "random",
-      "title": "Random",
+      "title": "Random Tip",
       "description": "Random a lucky tip",
       "mode": "view"
     },
     {
       "name": "search",
-      "title": "Search",
+      "title": "Search Tips",
       "description": "Search tips by keyword",
       "mode": "view"
     },
     {
       "name": "sync",
-      "title": "Sync",
+      "title": "Sync Tips",
       "description": "Sync all tips from remote",
       "mode": "no-view"
     }


pernielsentikaer (changes requested):
Hi 👋

I'm not into Rust, how can I compile it myself, also, is there a way to only have one binary - I don't think both are necessary, are they?

Wouldn't it be simpler to have it in a JSON file somewhere?

godruoyi:
> is there a way to only have one binary - I don't think both are necessary, are they?

Hi @pernielsentikaer thank your reply, Yes, it would be better to keep only one binary file in this extension. I plan to optimize it in the next version. Can I keep it as is for this initial version?

> Wouldn't it be simpler to have it in a JSON file somewhere?

I don't quite understand what this means, could you  please be more specific? That would be even better if you could give me an example.

pernielsentikaer:
> I plan to optimize it in the next version. Can I keep it as is for this initial version?
It will just add 10+ MB to the repository which could be only half of it and we can't really make it disappear from the history so users will need to download it while checking the repository out.

> I don't quite understand what this means, could you please be more specific? That would be even better if you could give me an example.
I was thinking that a simple JSON API would be more sufficient for this job instead of having a binary with the data. 

Can also be that I'm missing something 🙂

godruoyi:
Hi @pernielsentikaer Now there is only one binary file, can you help take a look?

pernielsentikaer:
That's a lot better @godruoyi 🔥 

Is there a way I can compile it from the source myself?

godruoyi:
@pernielsentikaer hi buddy can you please take a look?

godruoyi:
hi @pernielsentikaer This pull request has been blocked for several days. Is there anything I need to know?

PR #7142, opened 2023-06-16
[Obsidian] Fix menu bar icon colors

ashway83:
## Description

After changes introduced with #7034 menu bar icon is always black.

In recent versions of macOS, the menu bar color adapts to the background brightness instead of the current appearance setting. When the background is light, the menu bar switches to dark mode, and when the background is darker, the menu bar switches to light mode.

It seems that the `Color.PrimaryText` constant is not being updated for the menu bar, which is independent of the system appearance.

image
image
image

## Screencast

With my simple fix, the icon color now changes as expected:

image
image
image

## Checklist

### DIFF
--- a/extensions/obsidian/CHANGELOG.md
+++ b/extensions/obsidian/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Obsidian Changelog
 
+## [Fix] - 2023-06-19
+
+- Fixed menubar icon for light backgrounds
+
 ## [Updated icons] - 2023-06-15
 
 - Update extension icons to match the new [logo and brand guidelines](https://obsidian.md/brand)


ofalvai:
Whoops, I overlooked that the menu bar colors change based on the wallpaper. I'm okay with this change, but maybe @pernielsentikaer knows of a better API to handle this automatically instead of hardcoding the colors?

ashway83:
I have also tried [macOS color constants](https://developer.apple.com/design/human-interface-guidelines/color). They work but don't change dynamically based on the menu bar appearance.

ashway83:
@ofalvai I was looking for a more elegant way, but haven't found anything that would work as expected.
`MenuText` [CSS system color ](https://developer.mozilla.org/en-US/docs/Web/CSS/system-color) doesn't reflect the menu bar appearance.
I have also tried [macOS color constants](https://developer.apple.com/design/human-interface-guidelines/color). They work but don't change dynamically based on the menu bar appearance.

pernielsentikaer:
Yes, the image in menubar is based on the background 🙂

This was is fine, otherwise you'll need two png's

PR #7176, opened 2023-06-19
BitBucket: Remove usage of deprecated `render` function.

sxn:
The deprecated `render` function is not working as expected in the current version of Raycast. While the next release fixes it, instead of waiting for that version to be released, I thought it's better to update the extension to not rely on deprecated behaviour.

Closes https://github.com/raycast/extensions/issues/6987


sxn:
@Lp-Francois I hope you don't mind if I'll just go ahead and merge this fix. 🙏

PR #7272, opened 2023-06-25
Update json-format extension

AlexIsMaking:
## Checklist

### DIFF
--- a/extensions/json-format/CHANGELOG.md
+++ b/extensions/json-format/CHANGELOG.md
@@ -1,9 +1,8 @@
 # Changelog
 
-## [Option to Parse & Format Stringified JSON] - 2023-06-29
+## [Adds support for parsing stringified JSON] - 2023-06-29
 
-- Adds Parse & Format action to the Format JSON command, which parses stringified JSON, before formatting it
-- Reorientates the actions for the Format JSON command, to display a preview of the formatted JSON by default, which the user can then copy, using the Copy Result action
+- If the JSON is stringified, it will be parsed before formatting
 
 ## [Allow to Preview Formatted JSON] - 2023-04-04
 
--- a/extensions/json-format/package.json
+++ b/extensions/json-format/package.json
@@ -2,7 +2,7 @@
   "$schema": "https://www.raycast.com/schemas/extension.json",
   "name": "json-format",
   "title": "Format JSON",
-  "description": "Formats a JSON file with a selected indentation. Also supports parsing a JSON string before formatting.",
+  "description": "Formats a JSON file with a selected indentation. If the JSON is stringified, it will be parsed before formatting.",
   "icon": "command-icon.png",
   "author": "destiner",
   "contributors": [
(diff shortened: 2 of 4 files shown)


devloos:
Hey 🎉, @AlexIsMaking looks good thank you for contributing. I am trying to figure out what value this provides. I might be missing something but I see no difference than just running format command instead.

AlexIsMaking:
Hey @devloos thanks for checking this out! At the moment, if you try to format stringified JSON like:

```
"{\n  \"product\": \"Live JSON generator\",\n  \"version\": 3.1,\n  \"releaseDate\": \"2014-06-25T00:00:00.000Z\",\n  \"demo\": true,\n  \"person\": {\n    \"id\": 12345,\n    \"name\": \"John Doe\",\n    \"phones\": {\n      \"home\": \"800-123-4567\",\n      \"mobile\": \"877-123-1234\"\n    },\n    \"email\": [\n      \"jd@example.com\",\n      \"jd@example.org\"\n    ],\n    \"dateOfBirth\": \"1980-01-02T00:00:00.000Z\",\n    \"registered\": true,\n    \"emergencyContacts\": [\n      {\n        \"name\": \"Jane Doe\",\n        \"phone\": \"888-555-1212\",\n        \"relationship\": \"spouse\"\n      },\n      {\n        \"name\": \"Justin Doe\",\n        \"phone\": \"877-123-1212\",\n        \"relationship\": \"parent\"\n      }\n    ]\n  }\n}"
```

then the format command won't work. This gives the user the option to parse stringified JSON and then format it, if they need to.

devloos:
Hey @AlexIsMaking, I see thanks for the clarification. Maybe it would play nicer if we didn't make another command and used your existing logic and make it work with `formatJS`. That way users don't have another command to think about.

This could be accomplished by using your `parse` function in formatJS then check if successful give `formatJS` a stringified json else give trimmed text.

let me know if it works I did not test

mil3na:
Hi @AlexIsMaking ! Thank you for your contribution and thanks @devloos for checking it out!
I'm just a bit confused: I don't see a new command being added, just a parser, which is exactly what this change is about. Seems to be transparent for the user. Am I wrong? @AlexIsMaking do you need any help? @devloos please tell us if you are ok if those changes are transparent to the user :)

mil3na:
It would be nice to update the CHANGELOG ;)

devloos:
Hey, @AlexIsMaking and @mil3na I fumbled my wording what I meant to say was instead of adding a new action toss it all under the same action. It could be cleaner since the end goal of both actions is to format the JSON given. It's not confusing to the user just removes friction points.

AlexIsMaking:
@devloos no worries, nice idea. I've just updated the commit to add the parsing functionality to the existing actions and commands so it's integrated seamlessly now.

mil3na (approved):
It looks good to me! 
If @devloos is also ok with it, we can merge it :)

AlexIsMaking:
Best,

Alex

devloos:
Looks good!

PR #7283, opened 2023-06-26
feat: add "Copy FA Class" action

sitek94:
## Description

Added "Copy FA Class" action which copies Font Awesome classes that are ready to use as class. For example, for Chrome icon it copies `fa-brands fa-chrome`.

## Screencast

https://github.com/raycast/extensions/assets/58401630/077d2962-a8dc-4138-8e90-35279491a54d

## Checklist

### DIFF
--- a/extensions/fontawesome/CHANGELOG.md
+++ b/extensions/fontawesome/CHANGELOG.md
@@ -1,6 +1,11 @@
 # Font Awesome Changelog
 
+## [Added action] - 2023-06-26
+
+- Added "Copy FA Class" action which copies Font Awesome classes, e.g. "fa-brands fa-chrome".
+
 ## [Added action] - 2023-02-28
+
 - Added "Copy FA Glyph" action which copies the unicode font-awesome glyph to clipboard for use with font-awesome font files
 
 ## [Removed HTTP Proxy] - 2023-02-28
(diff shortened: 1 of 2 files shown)


mil3na:
My name is Milena and I will be reviewing your change today :)

It looks good to me! I'm just trying to find an example where the icon doesn't have the class. Do you have an example? I wonder if it will break anything.

sitek94:
Hey @mil3na, thanks for reviewing! 

I'm not sure if I understand what do you mean by icon without a class? Do you mean that the request to fetch icons failed?

mil3na:
I am asking if there is any icon that doesn't have a class. And, if so, what happens when you try to copy the class?
Or are all the icons expected to have a class?
I'm just wondering about error handling.

sitek94:
I see. The icons don't have classes. I construct the class using `icon.id`, so it becomes, e.g. `fa-user`. 

Existing method `copyFASlugToClipboard` is also using `icon.id`, so if that has been working so far, I think that it should be safe:

```ts
const copyFASlugToClipboard = async (icon: Icon) => {
    // Copy icon name to clipboard
    await Clipboard.copy(icon.id);
    // Notify the user
    await showHUD('Copied Slug to clipboard!');
  };
```

PR #7307, opened 2023-06-28
Update jira extension

jesperorb:
## Description

This PR adds support for persisting the last selected project in the _"Active Sprint"_ Command. So you don't have to use `CMD+P` and manually search for your project every time you use the command. This will instantly load all issues for the last selected project.

## Checklist

### DIFF
--- a/extensions/jira/src/active-sprints.tsx
+++ b/extensions/jira/src/active-sprints.tsx
@@ -1,65 +1,62 @@
 import { List } from "@raycast/api";
 import { useCachedPromise, useCachedState } from "@raycast/utils";
-import { useRef, useState } from "react";
+import { useState } from "react";
 
-import { getProjects } from "./api/projects";
+import { Project, getProjects } from "./api/projects";
 import StatusIssueList from "./components/StatusIssueList";
 import { getProjectAvatar } from "./helpers/avatars";
 import { withJiraCredentials } from "./helpers/withJiraCredentials";
 import useIssues from "./hooks/useIssues";
 
 export function ActiveSprints() {
-  const [projectKey, setProjectKey] = useCachedState("active-sprint-project", "");
-  const cachedProjectQuery = useRef(projectKey);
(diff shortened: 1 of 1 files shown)


thomaslombart (commented):
Hey @jesperorb, thanks a lot for this PR. I have some feedback on it:

- `useCachedState` is already supposed to persist the value, so how is `usePersistedState` useful here? After testing it locally, it seems the behavior is the same only if I change the `projectQuery` to have `projectKey` as a default value.
- As the `projectQuery` is initially modified when remembering a value, it causes the dropdown to only display the project key in the initial list of results. Ideally, the last project should be in the list among initial results. For that, I think you'd need to retrieve the associated project on the initial render, put it as the first item of the dropdown and display the initial results (but without this project to avoid any duplicate results).

jesperorb:
> Hey @jesperorb, thanks a lot for this PR. I have some feedback on it:
> 
>     * `useCachedState` is already supposed to persist the value, so how is `usePersistedState` useful here? After testing it locally, it seems the behavior is the same only if I change the `projectQuery` to have `projectKey` as a default value.
> 
>     * As the `projectQuery` is initially modified when remembering a value, it causes the dropdown to only display the project key in the initial list of results. Ideally, the last project should be in the list among initial results. For that, I think you'd need to retrieve the associated project on the initial render, put it as the first item of the dropdown and display the initial results (but without this project to avoid any duplicate results).

Hey Thomas, 

- Yes, you are right. I was mistakenly fooled by the key being reset all the time which made me think the cache was not working. But it was actually the order of operations that was wrong.
- I updated my code to try to reflect this if I understood you correctly. So what I did now was to make a separate query for the save project and add that to the list or projects and filter it out from the whole search of all projects. Except when searching as only the search result should be displayed.

I pushed some new edits. Do you mind taking a look at it again?

jesperorb:
@thomaslombart Yes, that looks much better. I changed it to be closer to your example.

thomaslombart (approved):

PR #7310, opened 2023-06-28
Update one-time-secret extension

LarsaSolidor:
## Description

- Sends secret and metadata in form body which is encrypted and sent securely over HTTPS instead of sending this data in the URL of the request, which could be intercepted. 
- Adds support for drafts, allowing users to come back to a secret they are working on.

## Screencast

https://github.com/raycast/extensions/assets/16529675/ff2989ef-fd94-4c3a-99c5-cead54100d10

## Checklist


mil3na:
My name is Milena and I will be reviewing your change today :)

It seems to be working fine and I don't see any major issue with the code, but I'm wondering if the Draft feature makes sense. If it's a secret, should it be visible, even if partially? Maybe it's sensitive information you don't want other's eyes on it.

LarsaSolidor:
Hi @mil3na 

Thanks for looking at my PR! 

The draft feature was a direct request from a colleague who was using Raycast to fetch the things they wanted to share.  When they left the one-time-secret extension, it would lose whatever they'd collated so far, so they had to build up their secret's contents in a note file, and then paste it into the one-time-secret extension when they were finished.  

I couldn't see a way to expire a draft and thought this was the more useful solution, rather than not saving anything at all.

mil3na (approved):
Well, I don't have a strong opinion here, it just looked strange to me. But there's an option to discard the draft, so it should be fine. Thanks for adding more features to your extension

PR #7320, opened 2023-06-29
Bard: Upgrade Bard version to address API BC breaks.

sxn:
## Description

This PR includes all of the changes from https://github.com/raycast/extensions/pull/7285 that are _not_ related to the Chat feature.


EvanZhouDev:
It's `bard-ai` version v1.4.2, but other than that it should be good :)

EvanZhouDev:
Thank you so much! People at Raycast are wonderful :)

PR #7365, opened 2023-07-02
Add harpoon extension

brrian:
## Description

Add any application to a list and use the commands to quickly navigate between them. Inspired by ThePrimeagen's [Harpoon](https://github.com/ThePrimeagen/harpoon) plugin.

## Screencast

In this screencast you can see me adding 3 different applications to my list and using the different commands to quickly switch between them.

https://github.com/raycast/extensions/assets/6050972/06dffc84-acb0-4e84-ad1d-577362a9af46

## Checklist


mil3na:

I'm trying to understand the value this extension is adding when you already have the app search on Raycast and `cmd+tab` from macOS. Do you mind explaining your background? Or how this extension is useful to you?

mil3na:
So, like a workspace? I get it. Have you considered adding the "multiple workspaces" functionality? So you could change depending on the task you are working on.

brrian:
> Have you considered adding the "multiple workspaces" functionality?

I hadn't considered it since in my mind the list is very fluid and is constantly changing depending, but I do think that's an interesting idea and would be quite valuable to some users. I'm definitely open to adding this feature in the future and in the meantime can make some updates to better support something like this down the road.

> Moreover, it would be interesting to see the application's name, instead of the Application number.

I love this idea, I think it would be super helpful to see the app name in the `subtitle`. Is this possible with the current API? The commands are registered in the `package.json` so I'm not sure how to dynamically update it. 

> And why limit to 5 Apps?

No reason, it's just an arbitrary number that I decided on. 😅  Do you feel more would be more valuable? I can also add another command to `Open application at position {x}` and accept an argument to make it more dynamic. To me, the value of the app heavily depends on attaching a hotkey to the command, and having too many felt overwhelming. Adding more commands for more positions is easy so I'm happy to do it if you feel it adds value.

mil3na:
I'm thinking about how to improve the UI of your extension, as I'm not so happy with having an arbitrary number of apps. Nor with not knowing which app is associated with each number, unless I go to the list.

Unfortunately, there's no way to add dynamic command descriptions, but I like the idea of having an `Open application at position {x}` command. Maybe adding the list of apps on the `ActionPanel`? 

I don't like the idea of having a pre-defined command that does nothing. That's why I'm concerned about the arbitrary 5. What if the user only needs 3 apps? 

Let me know what you think.

brrian:
> Maybe adding the list of apps on the ActionPanel?

Sure! Let me make the changes when I get the time.

> I don't like the idea of having a pre-defined command that does nothing. That's why I'm concerned about the arbitrary 5. What if the user only needs 3 apps?

Hrm, yeah I see your point. Is there a way to dynamically create a command that launches another command with arguments? (For example, something similar to "Create quicklink")

To me, it's critical to be able to launch an app at a specific position via a hotkey as that was my primary motivation for creating this extension. Any ideas or suggestions would be appreciated!

mil3na:
> _Edit: I went to add this and I think I'm a little confused. I was imagining that Open Application at Position {x} would just be a no-view command with a single argument. I can't add an ActionPanel to this, right?_

I was thinking about having this `ActionPanel` on the same view that you have a list of Apps, in order to avoid the need to have a command for each app in the main list and make it a more concise user experience.

Deeplinking Quicklinks might be a way to go! Let me know if I can help. Ping me here or on Slack if you want to exchange more ideas.

brrian:
Hi Milena,

I've simplified the commands by removing all the `Open Application at Position 1-5` in favor of a single `Open Application at Position {x}`.

I've also added a new action in the "Show Applications List" to create a Quicklink with a deep link URL at the proper position.

I think I've address all of your concerns, if not please let me know. Thank you!

mil3na:
Hey @brrian ! Thanks so much for your effort.
The extension seems to be working fine and with a much better UI!
Could you please just update your screenshots? Also, if you do not intend to add anything to your `README` just remove it :)

brrian:
@mil3na Hi Milena, apologies for the delay. I've removed the README and updated the screenshots!

mil3na (approved):
Thank you! Looks good to me now. 🚀

PR #7381, opened 2023-07-03
Update authy extension

gdsmith:
## Description

@guga4ka hoping this fixes the errors people are seeing. The fix is in the [one liner here](https://github.com/raycast/extensions/pull/7381/files#diff-68aef4d8b7ebd8b1f6e53cec91a51d6e10f3d394e13e736e481e84160afac139R20). As far as I can tell, the main time the problem occurs is when the password is wrong and then all the OTP are corrupted, so I added some checks to determine if that is the case and show an alert to take them to settings or then to the Q&A page.

## Checklist


mil3na:
My name is Milena and I will be reviewing your code today :)

It does seem to fix the issue on my side. What do you think about the solution, @guga4ka? It would be lovely to hear from @vladfrangu if this solves the issue for them.

vladfrangu:
I'm not sure how to locally test this change 😅, but the constant mention of a wrong password did make me realize that I missed 1 character from it (-.-) which explained why passwords couldn't load in!

mil3na:
Yeah, inputting a wrong password was the way I was able to reproduce this issue. I wonder if a session termination or error-throwing would be a better solution here.

gdsmith:
It's an awesome extension that I use all the time so I'm happy to help. I
just saw a lot of the same type of issues popping up and tried to figure
out what might be happening.

mil3na (approved):

PR #7387, opened 2023-07-04
feat(Bilibili): support notifications

Fatpandac:
## Description

support notifications

## Screencast

image

## Checklist

### DIFF
--- a/extensions/Bilibili/package.json
+++ b/extensions/Bilibili/package.json
@@ -50,7 +50,7 @@
     }
   ],
   "dependencies": {
-    "@raycast/api": "^1.39.3",
+    "@raycast/api": "^1.55.1",
     "run-applescript": "^6.1.0"
   },
   "devDependencies": {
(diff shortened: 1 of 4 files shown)


extensions/Bilibili/package.json
```diff
@@ -43,7 +50,8 @@
     }
   ],
   "dependencies": {
-    "@raycast/api": "^1.39.3"
+    "@raycast/api": "^1.39.3",
```
mil3na, line 53: Do you mind updating to the latest Raycast API version?

Fatpandac: I do not mind.

mil3na:
Is there any other way to login to your extension that is not by the QR code? I managed to create an account, but the app is not available in my App Store's region.

Fatpandac:
There is no option to log in without using the QR code. Moreover, the Bilibili app is only available on the Chinese App Store.

mil3na:
Okay, would it be possible for you to make a video of the change in action? Just to make sure we are not introducing something that doesn't work :)

Fatpandac:
录屏2023-07-13 16 36 50

Check here.

PR #7398, opened 2023-07-05
Update menubar-weather extension

fuksman:
## Description

 You can choose to display either the forecasted or the "feels-like" temperature in the menubar, as well add UVI, pressure, humidity or wind info.

## Screencast
image
<br/>
image

## Checklist

### DIFF
--- a/extensions/menubar-weather/package.json
+++ b/extensions/menubar-weather/package.json
@@ -110,8 +110,8 @@
     {
       "name": "tempType",
       "type": "dropdown",
-      "title": "Display in menubar",
-      "description": "Forecast of feels-like",
+      "title": "Show in menubar",
+      "description": "Temperature type",
       "required": false,
       "default": "temperature",
       "data": [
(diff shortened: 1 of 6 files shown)


extensions/menubar-weather/package.json
```diff
@@ -107,6 +107,24 @@
         }
       ]
     },
+    {
+      "name": "tempType",
+      "type": "dropdown",
+      "title": "Display in menubar",
+      "description": "Forecast of feels-like",
+      "required": true,
```
mil3na, line 115: Hi! Thank you so much for your contribution to Raycast!
My name is Milena and I will be reviewing your update today :)

Adding `required` preferences might break your extension in case an older user doesn't have it set. Do you think it would be ok to remove the `required` as you already have a default value and the user input is controlled by a dropdown of options? And, of course, if the user has not selected anything, always fallback to the default value.

mil3na (changes requested):

extensions/menubar-weather/package.json
mil3na: Not really. This is a known issue right now.

PR #7412, opened 2023-07-06
Update numi extension

viandwi24:
## Description
adding option for "use numi-cli", so don't need to download the desktop version of Numi and run it, just download numi-cli and then can directly use this extension.

## Screencast

https://github.com/raycast/extensions/assets/41994283/c0db5607-7c6f-4b81-a073-7504df0adaab

## Checklist

### DIFF
--- a/extensions/numi/README.md
+++ b/extensions/numi/README.md
@@ -18,3 +18,10 @@ This command allows for quick search on Numi
 - `enter` will copy the result at the clipboard
 
 ![Query on Numi Command](https://raw.githubusercontent.com/raycast/extensions/711c5d9d36f3ed872fc29616945d33b137802c3f/extensions/numi/metadata/screenshot-02.png)
+
+## Use Numi CLI
+
+You can make this extension use the Numi CLI instead of the Numi API if you want
+
+- Install [numi-cli](https://github.com/nikolaeu/numi#numi-cli)
+- Open Extension Preferences and Check 'Use Numi CLI'


mil3na:
My name is Milena and I will be reviewing your update today :)

It looks great for me, thanks for adding this to the existing extension! @AndresMorelos what do you think?

extensions/numi/src/services/checkinstall.ts
```diff
@@ -10,21 +15,54 @@ async function isNumiInstalled() {
...
+      await showToast(options);
+    }
+  }
+}
 
-    await showToast(options);
+export async function isNumiCliInstalled(path?: string) {
+  try {
+    const res = await execp(`${path || DefaultPreferences.numi_cli_binary_path} --version`, { shell: "/bin/bash" });
+    console.log(res);
```
AndresMorelos, line 59: Remove `console.log`

extensions/numi/src/services/checkinstall.ts
```diff
@@ -10,21 +15,54 @@ async function isNumiInstalled() {
...
+
+      await showToast(options);
+    }
+  }
+}
 
-    await showToast(options);
+export async function isNumiCliInstalled(path?: string) {
+  try {
+    const res = await execp(`${path || DefaultPreferences.numi_cli_binary_path} --version`, { shell: "/bin/bash" });
```
AndresMorelos, line 58: Always use `DefaultPreferences.numi_cli_binary_path`; if there is any security flaw that permits the call of this function, modifying the path parameter could handle a big problem if the user has failures. They need to set up the correct path on the preferences.

extensions/numi/src/services/requests.tsx
```diff
@@ -31,3 +36,16 @@ export const query = (q?: string): Promise<string[]> => {
     }
   });
 };
+
+export const queryWithNumiCli = async (q?: string, path?: string): Promise<string[]> => {
+  const command = `${path || DefaultPreferences.numi_cli_binary_path} "${q}"`;
```
AndresMorelos, line 41: Always use `DefaultPreferences.numi_cli_binary_path`; if there is any security flaw that permits the call of this function, modifying the path parameter could handle a big problem if the user has failures. **They need to set up the correct path on the preferences.**

extensions/numi/src/index.tsx
```diff
@@ -56,39 +63,72 @@ export default function Command(props: LaunchProps<{ arguments: NumiArguments }>
...
-        setState((oldState) => ({ ...oldState, results, isLoading: false }));
-
-        handleUpdateHistory(q, results);
-      })
-      .catch((err) => {
-        if (err.message.includes("ECONNREFUSED")) {
-          setApiStatus(false);
-          setState((oldState) => ({ ...oldState, isLoading: false }));
-        }
+    console.log("Querying", q);
```
AndresMorelos, line 66: Remove `console.log`

extensions/numi/src/index.tsx
```diff
@@ -56,39 +63,72 @@ export default function Command(props: LaunchProps<{ arguments: NumiArguments }>
...
+            title: "Something went wrong",
+            message: "Please make sure Numi is running",
+          })
+            .then((toast) => setToast)
+            .catch((err) => console.error("Error Creating Toast"));
+        });
+    } else {
+      queryWithNumiCli(q, numi_cli_binary_path)
+        .then((results) => {
+          console.log("Results", results);
```
AndresMorelos, line 94: Remove `console.log`

extensions/numi/src/index.tsx
```diff
@@ -56,39 +63,72 @@ export default function Command(props: LaunchProps<{ arguments: NumiArguments }>
...
+      query("1")
+        .then(() => setApiStatus(true))
+        .catch((err) => {
+          if (err.message.includes("ECONNREFUSED")) {
+            setApiStatus(false);
+          }
+        });
+    } else {
+      let path: string | undefined = numi_cli_binary_path;
+      if (path.trim() === "") path = undefined;
```
AndresMorelos, line 127: Let the `isNumiCliInstalled` handle the error. If there is no bin path, the exec will fail, or check this on the `isNumiCliInstalled` function.
CleanShot 2023-07-06 at 23 15 07

AndresMorelos (changes requested):
Some Feedbacks given

AndresMorelos:
@mil3na @viandwi24 I left some little feedback here.

extensions/numi/src/services/checkinstall.ts
viandwi24: I provide the path as an optional parameter if the user installs it outside of the options recommended by Numi-Cli, at this case using the homebrew bin path. is it better to use this default binary path without providing an option for the user to change it?

extensions/numi/src/index.tsx
viandwi24: ok, I have tried to handle all the errors it gets from `exec` it will return `false` value on `isNumiCliInstalled` function. ensure the output of the path check

extensions/numi/src/services/checkinstall.ts
AndresMorelos: Ok, I got it. Always use the **numi_cli_binary_path** from the preferences.

`extensions/numi/src/types`
```ts

export interface Preferences {
  max_history_elemets: string;
  use_numi_cli: boolean;
  numi_cli_binary_path: string;
}

```
At `extensions/numi/src/services/checkinstall.ts`
```ts
import { getPreferenceValues } from "@raycast/api";
import { Preferences } from "../types/index";
...

const {  numi_cli_binary_path } = getPreferenceValues<Preferences>();

export async function isNumiCliInstalled(path?: string) {
  if (path && path.trim() === "") path = undefined;
  try {
    const res = await execp(`${numi_cli_binary_path} --version`, { shell: "/bin/bash" });

   if (res.stderr) return false;
    if (!res.stdout.includes("numi-cli")) return false;
    return res.stdout;
  } catch (error) {
    return false;
  }
}

```

At `extensions/numi/src/services/requests.tsx`
```ts
import { getPreferenceValues } from "@raycast/api";
import { Preferences } from "../types/index";

const {  numi_cli_binary_path } = getPreferenceValues<Preferences>();

export const queryWithNumiCli = async (q?: string, path?: string): Promise<string[]> => {
  q = q?.replace(/"/g, '\\"') || ""; // safe " in query
  const command = `${numi_cli_binary_path} "${q}"`;

  const res = await execp(command, { shell: "/bin/bash" });
  if (res.stderr) {
    console.error(res.stderr);
    return [];
  }
  let result = res.stdout;
  if (result.endsWith("\n")) result = result.slice(0, -1);
  result = Buffer.from(result).toString("utf8");
  return [result];
};

```

viandwi24: yes sure, I use the preferences of the user via the argument parameter which is passed via the function `isNumiCliIstalled` and function `queryWithNumiCli` parameters (path?: string).

in current code, i use `numi_cli_binary_path` from user prefs and if user not provided (preference undefined/null) so use a default value from `DefaultPreferences.numi_cli_binary_path`

like 
```ts
const {  numi_cli_binary_path } = getPreferenceValues<Preferences>();
const command = `${numi_cli_binary_path || DefaultPreferences.numi_cli_binary_path} "${q}"`;
```

or you want to me keep use from user pref only? without default value if user not setted? 
```ts
const {  numi_cli_binary_path } = getPreferenceValues<Preferences>();
const command = `${numi_cli_binary_path} "${q}"`;
```

AndresMorelos: Yep, always use the preference bin path.

viandwi24: sure, changed on https://github.com/raycast/extensions/pull/7412/commits/d3b9beb95d0b16d7031726c9026db5e2373b8301

viandwi24:
I think that's all I can do from @AndresMorelos  feedback

AndresMorelos (approved):
LGTM

mil3na (approved):
Looks good to me as well! Thanks for your effort :D

PR #7451, opened 2023-07-10
Add colorify extension

EvanZhouDev:
## Description
Introducing `colorify` for Raycast: Say goodbye to the days of manually crafting themes to match your wallpaper! With `colorify`, simply upload an image, tweak a few settings, and let the magic happen.

Key features:
-  Raycast AI-powered theme generator
-  Support for both light and dark themes
-  AI-generated theme titles for a personalized touch

## Screencast
colorify-1

## Checklist

### DIFF
--- a/extensions/colorify/package-lock.json
+++ b/extensions/colorify/package-lock.json
@@ -11,6 +11,7 @@
         "@raycast/api": "^1.55.1",
         "file-type": "^18.5.0",
         "get-pixels": "^3.3.3",
+        "image-size": "^1.0.2",
         "jimp": "^0.22.8"
       },
       "devDependencies": {
@@ -2021,6 +2022,20 @@
       "resolved": "https://registry.npmjs.org/@types/node/-/node-16.9.1.tgz",
       "integrity": "sha512-QpLcX9ZSsq3YYUUnD3nFDY8H7wctAhQj/TFKL8Ya8v5fMm3CFXxo8zStsLAl780ltoYoo1WvKUVGBQK+1ifr7g=="
     },
+    "node_modules/image-size": {
+      "version": "1.0.2",
+      "resolved": "https://registry.npmjs.org/image-size/-/image-size-1.0.2.tgz",
+      "integrity": "sha512-xfOoWjceHntRb3qFCrh5ZFORYH8XCdYpASltMhZ/Q0KZiOwjdE/Yl2QCiWdwD+lygV5bMCvauzgu5PxBX/Yerg==",
+      "dependencies": {
+        "queue": "6.0.2"
+      },
+      "bin": {
+        "image-size": "bin/image-size.js"
+      },
+      "engines": {
+        "node": ">=14.0.0"
+      }
+    },
     "node_modules/import-fresh": {
       "version": "3.3.0",
       "resolved": "https://registry.npmjs.org/import-fresh/-/import-fresh-3.3.0.tgz",
@@ -2654,6 +2669,14 @@
       "integrity": "sha512-qxXIEh4pCGfHICj1mAJQ2/2XVZkjCDTcEgfoSQxc/fYivUZxTkk7L3bDBJSoNrEzXI17oUO5Dp07ktqE5KzczA==",
       "engines": {
         "node": ">=0.6"
+      }
+    },
+    "node_modules/queue": {
+      "version": "6.0.2",
+      "resolved": "https://registry.npmjs.org/queue/-/queue-6.0.2.tgz",
+      "integrity": "sha512-iHZWu+q3IdFZFX36ro/lKBkSvfkztY5Y7HMiPlOUjhupPcG2JMfst2KKEpu5XndviX/3UhFbRngUPNKtgvtZiA==",
+      "dependencies": {
+        "inherits": "~2.0.3"
       }
     },
     "node_modules/queue-microtask": {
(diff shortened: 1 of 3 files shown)


AaronDavidSchneider:
I love the idea of this PR and tried it out. I get the following error:
```
07:21:20.594

NodeError:errors:399:5
[kOnExit]:worker:287:26
Worker.<computed>.onexit:worker:209:20
07:22:27.210

NodeError:errors:399:5
[kOnExit]:worker:287:26
Worker.<computed>.onexit:worker:209:20
```
It works on some  images and not on others.  Here are some of the images I tried it with:
20131112_PIA17172
main_image_star-forming_region_carina_nircam_final-5mb

It works on the carina nebula (wich has 5mb) but not on the image of saturn.

Asking for a title has worked in some cases but not in all cases.

EvanZhouDev:
It doesn’t work with images that are too large. Going to try to find a workaround but essentially you can just reduce image quality with something else and plug it in.

What do you mean by AI title not working, though? It has worked for me nearly every time?

AaronDavidSchneider:
> It doesn’t work with images that are too large. Going to try to find a workaround, but essentially you can just reduce image quality with something else and plug it in.

I had the same thought, but it seems that the large image works (5mb) and the small one doesn't (0.5mb).

> What do you mean by AI title not working, though… It has worked for me nearly every time?

I can't reproduce it at the moment, but there were certain title that didn't work

EvanZhouDev:
Well I’ll try to figure it out the image problem when I get back to my computer… it’s probably a resolution thing in the end.
What do you mean by “Doesn’t work” though for the title? Does it not give you a title, error out, or something else?

AaronDavidSchneider:
> Well I’ll try to figure it out the image problem when I get back to my computer… it’s probably a resolution thing in the end. What do you mean by “Doesn’t work” though for the title? Does it not give you a title, error out, or something else?

I believe it doesn't give a title. But there is no direct error thrown. I think you could use a fallback title, when the AI command returns something weird.

EvanZhouDev:
Okay, so I have implemented a system in which images are first resolution-reduced, then processed with `color-thief`, but it still doesn't work with the 9k Saturn image...

AaronDavidSchneider:
Thats fair! Sorry, I only now realised it was 9K. Its black most of it and thus can be compressed to 0.5mb, which confused me.

Here is a warning that I now get:
```
14:38:46.424 (node:39491) [DEP0005] DeprecationWarning: Buffer() is deprecated due to security and usability issues. Please use the Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() methods instead.
(Use `node --trace-deprecation ...` to show where the warning was created)
```

The warning is a good idea, but maybe you could catch the exception? And print a proper toast?

EvanZhouDev:
> Here is a warning that I now get:
> 
> ```
> 14:38:46.424 (node:39491) [DEP0005] DeprecationWarning: Buffer() is deprecated due to security and usability issues. Please use the Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() methods instead.
> (Use `node --trace-deprecation ...` to show where the warning was created)
> ```

That's from a library, can't necessarily fix it easily...

> The warning is a good idea, but maybe you could catch the exception? And print a proper toast?

It's hard to tell the difference between that exception and the rest of the exceptions (i.e. in AI, JSON.parse(), etc.), because you can't know when _exactly_ JS heap runs out of memory...

EvanZhouDev:
On second thought I may as well wrap the whole thing inside a `try/catch`

EvanZhouDev:
It doesn't work 😅. I think it's an error on the entire Raycast `node` process and not a JS thing...

mil3na:

First of all, I love the idea and it will be so useful to me! Thanks again.
About the image size problem, maybe you could do some validation on the form and limit the image size? I think it's worth doing that if it's a known issue / unsupported feature.

Also, for the dark themes, codeblocks are not very color-friendly. Anything that can be done about that?

CleanShot 2023-07-14 at 15 06 53@2x

EvanZhouDev:
Hello! Thank you so much for your kind words.

I will definitely try to get the size thing down, but the code blocks are a bit harder to fix. I am currently extracting colors with `colorthief`, and then using AI to improve contrast and build it into a theme. The codeblocks look bad because I am not editing the "Support Colors," but I fear that providing the AI with too much will just cause it to be confused. Do you have any ideas about how to fix that?

mil3na:
I believe that the way to go is to match the support colors with the ones you get from AI. You don't need to generate those, but re-use the ones you got or maybe re-use them in another tone/contrast/brightness so it would match the theme. Give it a go, if it doesn't work, I don't think it's a blocker, as the user can edit the theme later.

EvanZhouDev:
I'm just curious, how do you suggest I modify the colors? I just don't see a way to measure the "contrast" level of the colors...

mil3na:
Yeah, I know you can do some mathematics and find different tones of the same color. But generating color based on pictures that also look good/visible is a very hard problem. My suggestion was to use the colors you already have.
Again, I don't think it's a blocker, just let me know if you want to solve this problem later.

EvanZhouDev:
Yes, I will try to implement that some time in the future. I am currently working on another project, so I may get back to this later.
With the image size checking, even checking the size of really large photos drives JS heap out of memory, so there's really nothing I can do about it... I did implement the code though.

mil3na (approved):
I think it should be fine. We will see on the community if people are encountering errors with it and we can come back to this then. Thanks for your effort, I will be using this right now :)

PR #7469, opened 2023-07-11
Add markdown-image-to-html extension

qeude:
## Description

This is an extension to easily transform Markdown formatted images to HTML formatted images. This is pretty useful when writing Markdown and needing a specific fixed size for images or to avoid big images such as in Github PRs.

There is a preference to set the base width wanted, `300` by default (just because this is the one I use) that can be modified and that also can be overridden for a one-shot thanks to arguments.

## Screencast

https://github.com/raycast/extensions/assets/19294297/e243efd3-ecd8-4633-9953-62e15be5a049

## Checklist

### DIFF
--- a/extensions/markdown-image-to-html/src/index.ts
+++ b/extensions/markdown-image-to-html/src/index.ts
@@ -33,7 +33,7 @@ export default async function main(
 }
 
 async function getBestMatch(regex: RegExp, param: string | undefined): Promise<string | undefined> {
-  if (param) {
+  if (param && regex.test(param)) {
     return param;
   }
   try {


daniellacoste:
wow! what are the odds that I created an extension ( #7462 ) that does the exact same thing less than 24 hours before this one 😂  admittedly, mine needs a more Raycast-like experience — well done @qeude 😛 great minds think alike

daniellacoste:
To be honest, this is the first time I've actually attempted to contribute to an OSS platform... so now I feel extremely validated to see that someone had the exact same problem as me and wanted to solve it, too! 😂 I think your implementation is more aligned with Raycast usage in that it accepts a parameter, whereas mine simply fetches the most recent content on the clipboard. At a quick glance I see that our markdown parsing logic is somewhat different, but I am not strongly opinionated enough to know which works "better", or if they could simply be combined. I'll have to circle back, so please weigh in 😅 

In summary, I have no preference—but I would love to help make our extensions better, together!

qeude:
Well, sorry to bother your first contribution haha, your implementation is great, you def should keep contributing to such OSS projects!
I tried to cover as many types of usage as possible that why I set `preferences`, `parameter`, and so on so users could use this extension the way that fits their needs.
IMO for the parsing logic it really depends if the purpose of the extension remains to transform only a few specific Markdown tags. If so, I preferred regex over external lib for maintainability. Having an external lib means that it needs to be updated at some point while regexes don't 😁.

At the end of the day, I don't think there is really one that it's "better" it's just two implementations that fix the same problem.

mil3na (approved):

It looks good to me! Thanks for your effort :)

daniellacoste:
@qeude totally, I agree with your points — I'll be happy to use the new extension! thank you 👍

qeude:
@daniellacoste well, I forgot that when accepted the PR is automatically merged, feel free to open a PR to add you as contributor 🙏

PR #7498, opened 2023-07-13
Add format-code extension

raszi:
## Checklist

### DIFF
--- a/extensions/format-code/README.md
+++ b/extensions/format-code/README.md
@@ -1,3 +1,3 @@
 # format-code
 
-Format Code
+Formats source code of various languages from your clipboard and copies that back to the clipboard in the formatted way.
--- a/extensions/format-code/package.json
+++ b/extensions/format-code/package.json
@@ -96,8 +96,8 @@
     },
     {
       "name": "format-vue",
-      "title": "Format VUE",
-      "description": "Formats VUE on the clipboard",
+      "title": "Format Vue",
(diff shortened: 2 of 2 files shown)


mil3na:

We already have an extension for Prettier on our store: https://www.raycast.com/lin/prettier
Do you think your extension provides any value that the one in our store doesn't? Or that could not be added to them?

mil3na:
I understand that your extension has more Languages/ Frameworks options, but still, you are only using Prettier. I see your point about Prettier being just a way of implementing things, but so far, that's the only way you are doing it.

We want to avoid having duplications in our store and I still believe that your extension would be better lived as an addition to the current extension :)

raszi:
Okay, then let me play the devil's advocate for a minute.

1. I would do the changes to the existing Prettier extension as you suggested.
2. Imagine it gets merged.
3. Then I would need to add yet another format that Prettier does not support but there is a great formatter that does.
4. Then, I guess the logical next step is to pull that dependency in and implement the functionality.
5. And then we would need to change the name of the prettier extension to something else.

Could you please explain to me how that would work? How would the existing users get the new functionality without uninstalling and installing the newly named extension?

mil3na:
Extension updates are automatic and transparent to the user. Including when the extension changes names. There are a couple of cautionary actions to be taken by the reviewer on our side, but nothing that should block the dev/author to change their extension name if needed.

PR #7571, opened 2023-07-19
[1password] support multiple 1password accounts. fixes #6713

### LINKED ISSUE #6713: [1Password] - Can only login and use passwords from 1 account
### Extension

### Description

Many of us will have 2 accounts -  if you use 1password for work, 1password actually give everyone a free account to use for their personal passwords. It would obviously be much more useful if I could access passwords from both my personal and work account via Raycast.

Everyone with 1password for work and home

### Anything else?

_No response_

jedelson-pagerduty:
## Description

As described in #6713, the current 1Password extension only supports a single account. Switching accounts requires clearing cache. With this change, Account selection is now a `List` that enables navigation into the items for that account.

## Screencast

https://github.com/raycast/extensions/assets/103767938/06687a72-434f-40e8-b10f-030fa67d4afd

## Checklist

### DIFF
--- a/extensions/1password/.eslintrc.json
+++ b/extensions/1password/.eslintrc.json
@@ -6,5 +6,8 @@
   },
   "parser": "@typescript-eslint/parser",
   "plugins": ["@typescript-eslint"],
-  "extends": ["eslint:recommended", "plugin:@typescript-eslint/recommended", "prettier"]
+  "extends": ["eslint:recommended", "plugin:@typescript-eslint/recommended", "prettier"],
+  "rules": {
+    "@typescript-eslint/no-non-null-assertion": "off"
+  }
 }
--- a/extensions/1password/package.json
+++ b/extensions/1password/package.json
@@ -12,7 +12,8 @@
     "cb372",
     "Davidwalser",
     "henrik-dmg",
-    "spacedog"
+    "spacedog",
+    "justinedelson"
   ],
   "categories": [
     "Security"
--- a/extensions/1password/src/v8/components/AccountForm.tsx
+++ b/extensions/1password/src/v8/components/AccountForm.tsx
@@ -1,65 +1,89 @@
-import { Action, ActionPanel, Clipboard, Form, Icon, showToast, Toast } from "@raycast/api";
+import { Action, ActionPanel, Clipboard, Icon, showToast, Toast, List } from "@raycast/api";
 import { useState } from "react";
 
 import { Items } from "./Items";
 import { Guide } from "./Guide";
 import { User } from "../types";
-import { op, ACCOUNT_CACHE_NAME, useOp, cache } from "../utils";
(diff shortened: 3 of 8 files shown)


pernielsentikaer:
I haven't played with this yet, out of curiosity; is it possible to have only one list and use a [List.Dropdown](https://developers.raycast.com/api-reference/user-interface/list#list.dropdown) where you can select either one (or all)?

jedelson-pagerduty:
@pernielsentikaer I don't think so (at least if I'm understanding you correctly). The `List` used for the actual 1Password items already has a [`searchBarAccessory`](https://github.com/raycast/extensions/blob/e0d7d84d228586a7547747d4b568b9a50486807e/extensions/1password/src/v8/components/Items.tsx#L35)

Combining the two selections (account and category) into one dropdown seems wrong IMO.

jedelson-pagerduty:
I'm going to turn this into a draft. After testing this on my personal machine (which only has one account), the UX for single-account users is worse. I'll work up an alternate solution.

pernielsentikaer:
Sounds like a plan to skip that step if you only have one account

jedelson-pagerduty:
I've updated this change so that the selected account is persisted in cache across sessions. For a single account system, this results in a UX that is essentially the same as the existing experience. For a multi-account system, by default you are taken to the _last used_ account and then can navigate to the account selection screen.

@MGHollander @oliver-platt curious for your feedback on this approach

MGHollander:
@jedelson-pagerduty I can't review the code, but it work great!

jedelson-pagerduty:
@pernielsentikaer FWIW, I also played around with adding an "All Accounts" option on the Account List screen. It seems doable but IMO belongs as a separate PR from this one. If you disagree, please let me know and I'll clean the code up and include it.

pernielsentikaer:
I think it makes sense to keep out of this PR 🙂

@khasbilegt @dteare do you have time to check this out?

pernielsentikaer:
Opening again

pernielsentikaer:
@thomaslombart could you help with checking this? I checked it and it works as I think it should but would love some extra eyes on it

thomaslombart:
@pernielsentikaer I'm not familiar with the 1Password extension, what should I check in particular?

craigharman:
Looks good! I'd definitely like to see a PR that allows for all accounts to be available at the one time.

vitorf7:
> Looks good! I'd definitely like to see a PR that allows for all accounts to be available at the one time.

@craigharman similar to what the 1Password Quick Access or even alfred would allow which goes through all your available passwords from all accounts.

craigharman:
> > Looks good! I'd definitely like to see a PR that allows for all accounts to be available at the one time.
> 
> @craigharman similar to what the 1Password Quick Access or even alfred would allow which goes through all your available passwords from all accounts.

Exactly!

PR #7614, opened 2023-07-22
Update ssh-manager extension

RtiM0:
## Description

This change adds the ability to view/modify ssh hosts from `~/.ssh/config file` in addition to localStorage

## Checklist

### DIFF
--- a/extensions/ssh-manager/CHANGELOG.md
+++ b/extensions/ssh-manager/CHANGELOG.md
@@ -1,5 +1,9 @@
 # ssh-manager Changelog
 
+## [Modify hosts from  ~/.ssh/config] - 2023-09-20
+
+- This change adds the ability to view/modify ssh hosts from ~/.ssh/config file in addition to localStorage
+
 ## [Make user field optional] - 2023-09-06
 
 - Make user field optional in connection config - to support user-less hosts (e.g. pseudo hosts that are configured in `~/.ssh/config`)


pernielsentikaer:
@RtiM0 thanks for your contribution 🔥 

@scopsy could you give this a spin

RtiM0:
Bump

RtiM0:
Hey @scopsy 
Sorry for the ping, but can we escalate this. I think this is a useful feature and I want to be able to use it on multiple devices with raycast

scopsy (approved):

scopsy:
@RtiM0 sorry, just was really busy at work. Haven't had a chance to test the functionality on my machine, but couldn't find any issues with the code proposed 🙏

RtiM0:
Hey @pernielsentikaer 
Can this be escalated 😬

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

yaronuliel:
@RtiM0 @pernielsentikaer Given that I've authored a different PR https://github.com/raycast/extensions/pull/8343 to address a very similar issue - there is a number of problems with this current implementation

1. It doesn't support the `Include`([info](https://man.openbsd.org/ssh_config#Include)) directive in `~/.ssh/config`
2. Since you only support a very small subset of `~/.ssh/config` options - writing to the file from raycast might cause a catastrophe and override user's valid configurations, and this is not a very good idea
3. The script doesn't support case-insensitive directive, while ssh-config do

I believe that the solution should be that records from `~/.ssh/config` are read only, while the user can still add additional records to localStorage, the 2 should co-exist and not be alternatives

RtiM0:
@yaronuliel these are some good points regarding parsing the ssh config file.
The parsing functionality should at least be whole.

Although I would very much prefer keeping localhost and ssh config as an alternative.
In my opinion, all the changes made through raycast should also reflect in the ssh config so that your workflow is consistent if you have to use ssh config in some other usecase.

yaronuliel:
@RtiM0 [sshconfig options](https://linux.die.net/man/5/ssh_config) are WAY to broad for raycast to support all of them, moreover - it will require a constant upgrade with relation to feature release in openssh, and conditional support based on the installed version of openssh (not that openssh update too often, but still...)
in my version - I treat sshconfig as a dictionary of predefined hosts that the user has configured for personal use, and nothing more
I don't aim to manage it from a tool that (IMHO) should give me shortcut for common tasks in my Mac - such as connecting to a SSH server

Also note one of my points - that ssh_config supports splitting the config across multiple files - meaning that when defining a new host - you will need to chose where to write the custom "local" hosts to

Other tools have selected different approaches to these issues, I like Fig's approach:
they parse ssh config files as readonly for typeahead purposes (including recursive search over included files) + they generated a completely separate config file, fully managed by them to configure the hosts that the user configure through their UI, and they add reference to it with an `Include` statement in the main sshconfig file. 
This way - they don't risk damaging existing config files, while still persisting the user's custom hosts in the starndard ssh_config way

Regardless of what approach is selected for raycast - the current behavior should be limited to readonly ASAP to prevent risk damage !

Issue #7475, opened 2023-07-12
[Pomodoro] ... menu bar timer does not update


theomjones:
### Extension

https://www.raycast.com/asubbotin/pomodoro

### Description

Menu bar timer does not update unless it is interacted with (clicked on)

### Steps To Reproduce

1. Enable menu bar item
2. Start a timer
3. See that it does not update

pernielsentikaer:
It's only possible to update the menubar every 10th seconds as it is right now, 1 second won't work

PR #7624, opened 2023-07-24
[pomodoro] Update menubar interval 10s -> 1s in package.json

erictheswift:
Fix addresses https://github.com/raycast/extensions/issues/7475


pernielsentikaer:
It's only possible to update the menubar every 10th seconds as it is right now, 1 second won't work

pernielsentikaer:
No, but it can lead to unexpected behavior if 5+ menubar commands request data every 1 second 🙂

PR #7647, opened 2023-07-25
Update xcode extension

qeude:
## Description

Changed the data source for the feature in this [PR](https://github.com/raycast/extensions/pull/7613) from a txt gist to a proper API.
Addition of macs as well.

## Checklist

### DIFF
--- a/extensions/xcode/CHANGELOG.md
+++ b/extensions/xcode/CHANGELOG.md
@@ -1,8 +1,8 @@
 # Xcode Changelog
 
-## [Change Data Source for List Apple Devices] - 2023-07-26
+## [Change Data Source for List Apple Devices] - 2023-07-28
 
-- Changed the data source for the "List Apple Devices" command so it doesn't rely on a txt file anymore.
+- Changed the data source for the "List Apple Devices" command so it doesn't rely on a text file anymore.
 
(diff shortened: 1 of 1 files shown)


pernielsentikaer:
Thanks for the correction @qeude 🚀 

Do you mind checking @SvenTiigi 🔥

SvenTiigi:
LGTM 👍
One little suggestion @qeude: 
Would it be possible to add a fallback in case of an error or unavailability of the `https://api.ipsw.me/v4/devices` HTTP endpoint to use the contents of the gist from the previous PR https://github.com/raycast/extensions/pull/7613

qeude:
It would be possible, but I'm not sure we want to do so. 
Let's say that for some reason someone is using the extension with low connectivity. They could launch the command two times in a row, but have two different results which looks confusing to me. I think this is better for the UX to have a consistent and predictable result, even though it means that HTTP errors could happen at some point.
Adding to that, let's say that `https://api.ipsw.me/v4/devices` just shutdown and is not available anymore, the architecture of the extension with the service allows us to easily change the data source for this list so not a big deal to fix this.
WDYT @SvenTiigi?

SvenTiigi:
Alright, sounds good 👍

pernielsentikaer (approved):

PR #7666, opened 2023-07-26
Update youtube-downloader extension

darmiel:
## Description

- updated `ytdl-core` dependency from `^4.11.4` to `^4.11.5`
- added an option to enable `.webm` for higher quality downloads

## Screencast

https://github.com/raycast/extensions/assets/71837281/85fde443-c113-4fce-8257-9ea8580f605d

## Checklist

### DIFF
--- a/extensions/youtube-downloader/CHANGELOG.md
+++ b/extensions/youtube-downloader/CHANGELOG.md
@@ -1,5 +1,10 @@
 # YouTube Downloader Changelog
 
+## [Added new format] - 2023-08-05
+
+- Updated ytdl-core dependency from ^4.11.4 to ^4.11.5
+- Added an option to enable .webm for higher quality downloads
+
 ## [Custom `ffmpeg` path] - 2023-07-08
 
 - Added a preference so users can configure the `ffmpeg` executable path


vimtor:
Hey @darmiel! Thank you for the contribution.

I think this is a great addition. The only thing I'm wondering is if webm support should be hidden behind a preference. I wouldn't mind having the `webm` options shuffled with the `mp4` options or putting them, as you did, in a separate section shown by when available.

What do you think?

darmiel:
Hi @vimtor,

> I wouldn't mind having the `webm` options shuffled with the `mp4` options or putting them, as you did, in a separate section shown by when available.

I think it might be beneficial to clearly differentiate between MP4 and WebM formats in separate sections to prevent accidental WebM downloads, as WebM is primarily used for web stuff. Most video-related software, like Adobe Premiere or Final Cut, lack the capability to import WebM files.

> The only thing I'm wondering is if webm support should be hidden behind a preference

Currently, the default setting for WebM downloads is Opt-Out. I thought this might be helpful if you need the compatibility and don't want to download them by accident. But we can of course remove the option, whatever you think is the best 😁

(btw. maybe in another PR I can try implement webm to mp4 conversion using ffmpeg, wdyt?)

darmiel:
> (btw. maybe in another PR I can try implement webm to mp4 conversion using ffmpeg, wdyt?)

I just saw I can just leave the container to be `mp4` so webm downloads are converted anyways:

```ts
    ffmpeg()
      .input(videoTempFile)
      .input(audioTempFile)
      .videoCodec("copy")
      .audioCodec("copy")
      .format("mp4") // <--
      .outputOptions("-strict", "-2")
      .save(filePath)
```

I will update the PR later to convert webm to mp4 :)

darmiel:
@vimtor what do you think?

image

It displays webm and mp4 downloads together but indicates that the webm downloads have to be converted. When you press download, it downloads the video in webm and then using ffmpeg converts it to mp4. You can set the codec in the preferences:

image

And you can also download the original webm file:

image

pernielsentikaer:
Could you check it again @vimtor 🙂

vimtor:
I don't thing adding an extra action is the way to go. I would prefer if we could place the new format in a new section, as @darmiel did in the demo, but without it being behind a preference.

darmiel:
> I don't think adding an extra action is the way to go. I would prefer if we could place the new format in a new section, as @darmiel did in the demo, but without it being behind a preference.

I have removed the preference :)

vimtor (approved):
Looks great! Thank you @darmiel for the contribution ❤️

PR #7773, opened 2023-08-02

efmmoncada:
## Checklist

### DIFF
--- a/extensions/spotify-player/src/addToQueue.ts
+++ b/extensions/spotify-player/src/addToQueue.ts
@@ -1,8 +1,8 @@
-import { showHUD } from '@raycast/api';
-import { searchTracks } from './api/searchTracks';
-import { setSpotifyClient } from './helpers/withSpotifyClient';
-import { addToQueue } from './api/addTrackToQueue';
-import { getErrorMessage } from './helpers/getError';
+import { showHUD } from "@raycast/api";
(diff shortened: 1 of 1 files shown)


peduarte:
Hi, we already have an Add to Queue Action, which is more contextual to the song/artist/playlist you want to listen to next

(ಠ ›ಠ) — @peduarte — MDp2jeeq002592@2x

I'm not sure this command offers a good UX, you'll be adding a potentially unexpected item to your queue, since there is no search.

peduarte:
Also, just a heads up for the future, before creating a PR, it may be worth creating an issue so we can debate, or at least provide enough context in the description. It's hard to understand why we need this otherwise. Thanks!

efmmoncada:
Thanks for the heads up! I'll note to raise an issue next time. I totally hear you on the issue where you potentially add something unexpected to the queue. I guess the functionality I had in mind was similar to the "Just Play" action, but instead of overriding what you have playing, just add it to the queue instead, keeping the behavior of selecting the first/most popular result matching the query, just like "Just Play" does

peduarte:
I hear you, but I think adding something to a playlist is different than "just playing". When you play something that's not what you expect, you can easily skip, play something else, etc. When you add something to the playlist, its not as easy to remove it. It needs to be a fully conscious decision

PR #7848, opened 2023-08-08
feat: support qrcode Clipboard

chenpx976:
## Description

support qrcode copy the image to the clipboard

## Screencast
image

## Checklist

- [x] I read the [documentation about publishing](https:developers.raycast.com/basics/publish-an-extension)
- [x] I ran `npm run build` and [tested this distribution build in Raycast](https:developers.raycast.com/basics/prepare-an-extension-for-store#metadata-and-configuration)


extensions/qrcode-generator/src/clipboard.tsx
```diff
@@ -1,40 +1,90 @@
...
+    }
+    return name + "-" + index + "." + extension;
+  }
+}
 
 export default function Command() {
   const [qrData, setQrData] = useState<string>();
-
+  const [clipboardContent, setClipboardContent] = useState<string>();
+  const hasCopy = useRef(false);
```
darmiel, line 28: Is `hasCopy` necessary? The QR code should be generated only once because of the empty `useEffect` anyways, shouldn't it?

darmiel (changes requested):
Thank you for the contribution! I think it's a really good addition to the clipboard QR Code generation 🙂. However, I think saving & copying the QR Code to the clipboard should be an extra action (e. g. pressing enter) because the main reason for this command was to quickly transfer a link / text contents to a mobile phone and most of the time you probably don't want to override the clipboard.

extensions/qrcode-generator/src/clipboard.tsx
pernielsentikaer: Did you see this @chenpx976

extensions/qrcode-generator/src/clipboard.tsx
```diff
@@ -1,40 +1,89 @@
...
+      const downloadedPath = homedir() + "/Downloads/qrcode/";
+      fse.ensureDirSync(downloadedPath);
+      const fileName = buildFileName(downloadedPath, "qrcode", "png");
+      const filePath = `${downloadedPath}/${fileName}`;
+      fse.writeFileSync(filePath, file);
+      const fileContent: Clipboard.Content = { file: filePath };
+      await Clipboard.copy(fileContent);
+    } catch (error) {
+      console.log(`Could not copy file. Reason: ${error}`);
+    }
```
Melvynx, line 66: Can you extract this in a separate function to have a clean and readble code ?

e-volusian:
Does anybody know if this feature was ever implemented? I would also like to copy the generated QR code to the clipboard, but don't currently see any way to do that.

PR #7858, opened 2023-08-09
Add get-path extension

nguyenvanduocit:
## Checklist

### DIFF
--- a/extensions/get-path/package.json
+++ b/extensions/get-path/package.json
@@ -12,7 +12,7 @@
   "commands": [
     {
       "name": "get-path",
-      "title": "Get path",
+      "title": "Get Path",
       "subtitle": "Get path of current selected item",
       "description": "You need to select a file first",
       "mode": "no-view"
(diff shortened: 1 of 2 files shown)


pernielsentikaer:
Hi 👋 

Thanks for your contribution 🔥 Did you see [this extension](https://www.raycast.com/koinzhang/copy-path) before creating this one?

nguyenvanduocit:
i saw it right after submitting the ext, but i think it is good to have
multiple ext for that.

pernielsentikaer:
If they do exactly the same it only confuses the end user to have multiple extensions doing the same purpose - if you have something in yours it might be better to enhance the current extension instead 🙂

nguyenvanduocit:
I think it is exactly the same. no enhance.

pernielsentikaer:
Closing as dublicate

nguyenvanduocit:
I believe that Raycast should allow users to describe the extensions they want to use, rather than imposing the team's opinion. If duplicate extensions are not allowed, there will be no competition, which will make it difficult for the platform to grow.

For example, let's say I create a small extension that is later rejected. If the extension had been approved, I would have had the opportunity to improve it over time by adding new features. This would have made my extension better and different from other extensions, which would have forced them to adapt. However, since my extension was rejected, I have no incentive to improve it.

I think this is a shortsighted approach. By allowing users to create and use the extensions they want, Raycast can create a more vibrant and competitive ecosystem. This will benefit everyone, including the platform itself.

PR #7874, opened 2023-08-09
Update arc extension

### LINKED ISSUE #4576: [Arc] `Search Arc` command opens wrong pinned tab if URLs are the same
### Extension

https://www.raycast.com/the-browser-company/arc

### Description

If you have 2 pinned tabs which have the same URL, for example, in different spaces, and search one tab by using `Search Arc` command - the wrong tab opens. In my example, I have 2 different `127.0.0.1:5173` pinned tabs in 2 different spaces: https://share.cleanshot.com/PVVpLSYm

### Steps To Reproduce

1. Pin 2 tabs with the same URL, rename them so you can distinguish
2. Run `Search Arc` command, search for the tab
3. The wrong tab opens

### Current Behaviour

If I run `Search Arc` command and search for a pinned tab which has the same URL with the another pinned tab - the another one opens instead.

### Expected Behaviour

I expect that the pinned tab I was searching for will be opened.

danulqua:
## Description

Fixes #4576. Had to pass a whole tab instead of just its URL into `OpenLinkActionSections` component so we can know exactly which tab to open if the tab was provided.

## Screencast

Not required.

## Checklist

### DIFF
--- a/extensions/arc/CHANGELOG.md
+++ b/extensions/arc/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Arc Changelog
 
-## [Fixes] - {PR_MERGE_DATE}
+## [Fixes] - 2023-08-11
 
 - Fixed an issue where, when searching for tabs with the same URL but different titles, the wrong tab would be opened.
 


extensions/arc/src/types.ts
```diff
@@ -15,6 +15,14 @@ export type Tab = {
 
 export type TabLocation = "topApp" | "pinned" | "unpinned";
 
+export function isTab(tab: Tab): tab is Tab {
```
hugivar, line 18: I don't think this function should be located in a file strictly created to house types

I'd recommend moving it to the utils.ts file

danulqua: @hugivar done ✅

danulqua:
Sure, one moment

danulqua:
@pernielsentikaer Updated changelog using `{PR_MERGE_DATE}`. Also, fixed the previous changelog entry as the placeholder was not replaced.

pernielsentikaer (approved):

PR #7935, opened 2023-08-14
Add phone-call extension

karbassi:
## Description

Make a phone call with default telephone app. The default is usually FaceTime.

Code is based on [Ron Myers' Facetime extension](https://www.raycast.com/ron-myers/facetime).

## Screencast

phone-call-1

## Checklist

### DIFF
--- a/extensions/quick-call/package.json
+++ b/extensions/quick-call/package.json
@@ -2,9 +2,12 @@
   "$schema": "https://www.raycast.com/schemas/extension.json",
   "name": "quick-call",
   "title": "Quick Phone Call",
-  "description": "Make a phone call with default telephone app.",
+  "description": "Make a phone call with the default telephone app.",
   "icon": "phone-icon.png",
   "author": "karbassi",
+  "contributors": [
+    "pernielsentikaer"
+  ],
   "categories": [
     "Communication"
   ],


pernielsentikaer:
We already had this in the [queue](https://github.com/raycast/extensions/pull/7250) some time ago where it stalled, it was pretty close to done, maybe we could use that?

I don't think there should be too many rules about the number and formatting, it needs to work all over the globe 🙂

karbassi:
> We already had this in the [queue](https://github.com/raycast/extensions/pull/7250) some time ago where it stalled, it was pretty close to done, maybe we could use that?

Thanks for sharing that. The PR had it's source file deleted. This PR is working. Could we use this moving forward?

> 
> I don't think there should be too many rules about the number and formatting, it needs to work all over the globe 🙂

I could remove the US-centric checks. Let me know if the team wants me to do that.

pernielsentikaer:
Sure, I still think it has some great features that I also implemented - this way you can use it with selected text and arguments and fallback

```
import { LaunchProps, getSelectedText, open, showHUD } from "@raycast/api";

export default async function Command(props: LaunchProps<{ arguments: Arguments.QuickDial }>) {
  const { fallbackText } = props;
  const { number } = props.arguments;

  let dialNumber = fallbackText || number;
  if (dialNumber === "") {
    try {
      const selectedText = await getSelectedText();
      if (selectedText !== "") {
        dialNumber = selectedText;
      }
    } catch (error) {
      console.error(error);
    }
  }

  dialNumber = dialNumber.replace(/\s/g, "");

  try {
    await open(`tel://${dialNumber.trim()}`);
    await showHUD(`Opening facetime calling ${dialNumber}...`);
  } catch (error) {
    console.error(error);
    if (error instanceof Error) {
      await showHUD(error.message);
    }
  }
}
```
What do you think about that?

Yeah, I think we should move the US-centric checks since I it would trigger some scenarios in e.g. Denmark 🙂

pernielsentikaer (approved):
Awesome @karbassi 🔥 Let's get it in the store

PR #8068, opened 2023-08-24
Add unix-timestamp-conversion extension

loerise:
## Screencast

unix-timestamp-conversion-1

## Checklist


mil3na:
Hi!
Thanks for your contribution. My name is Milena and I will be reviewing your extension today.
I installed it and it works fine, however, I wonder if it adds anything more than what our native calculator does:
CleanShot 2023-08-28 at 14 21 21@2x

It looks like repeated functionality.

loerise:
Hi Milena,

Thank you for your review. Let me introduce my extension to you.

Firstly, I suggest adding a shortcut for **Unix Timestamp Conversion**, such as Alt+D. This will enhance its performance.

Here are the key points about how the extension functions:

By default, if you haven't selected any text or if there is no text in your clipboard, the extension will display the current datetime in four formats (datetime, date, time, timestamp). You can easily copy them.

If you have selected some text or copied text to your clipboard, using the shortcut will trigger automated conversion. You could input the text by yourself too.

**Unix Timestamp Conversion** is built on Day.js, which means it supports [many formats](https://day.js.org/docs/en/parse/parse).

To be honest, I have tried all the other extensions in the store and the calculator you mentioned. **Unix Timestamp Conversion** provides a simpler and faster way to convert timestamps with dates.

Best regards,

Lester Lau

mil3na:
Hi!
Thanks for the clarification, but I still don't see how it adds more functionality than being able to call it via hotkey. Am I wrong?
I'm sorry, but I don't think it adds more value to what we have natively or in other extensions, to be its own extension.

loerise:
Hi!

About your questions:
### 1. adds anything more than what our native calculator does?
I have to say there is no **additional functionality**. The functionality have indeed been provided by other extensions or the native calculator now.

### 2. how it adds more functionality than being able to call it via hotkey?
My extension workflow's best practice is base on hotkey, but you don't have to，and here is how it different from native calculator (both list only the simplest and most way):

**Native calculator**
1. Select the timestamp text.
2. Copy the selected text.
3. Call the native calculator using a shortcut key (suppose there is a hot key).
4. Paste the copied text.
5. Type " in date" if you copied a timestamp, or " in timestamp" if you copied a date string.
6. Copy the result (And there are no more formats to select from).

**My extension**
1. Select the timestamp text.
2. Call the extension using a shortcut key (get the text from the selection or clipboard first, and finally from the user input, default is current timestamp) or just open raycast and enter extension (only get the text from the user input, default is current timestamp).
3. Select and copy the formatted result.

As I mentioned before, my extension offers a simpler and faster way to convert timestamps with dates. There is no new additional functionality, it's only simple the workflow and saves me times, and maybe other user's too, just like most other extensions do.

I hope that could answer your question, thanks.

mil3na:
Yes, it does answer my question.
Thanks for your effort, but as this adds no extra functionality, I will be closing this PR.
If you want to discuss further, feel free to find me on our [Slack](https://www.raycast.com/community) community.

PR #8076, opened 2023-08-24
[JIRA] Add setting to search issues by default project key

iltimasd:
## Description

Modifies the extension to include a setting for the Search Issues command. If a project key is provided in setting, and a users search only using numbers then the JQL will be modified to search with the project key concatenated to the initial query. 

I opted to use the issue key JQL field as less restrictive search than strictly by project key. Maybe it should be more restrictive? Open to thoughts

## Screencast

*I can't include screen recording due to sensitive info

Image 1: Setting with project key 
Image 2: Example query
Image 3: Another example query

CleanShot 2023-08-23 at 16 14 32
CleanShot 2023-08-23 at 16 15 01
CleanShot 2023-08-23 at 16 15 09

## Checklist

### DIFF
--- a/extensions/jira/src/search-issues.tsx
+++ b/extensions/jira/src/search-issues.tsx
@@ -14,11 +14,11 @@ export function SearchIssues({ query: initialQuery }: SearchIssuesProps) {
   const [query, setQuery] = useState(() => {
     return initialQuery ?? "";
   });
-  
-  let prefix = getPreferenceValues()["search-issues-prefix"]
+
+  let prefix = getPreferenceValues()["search-issues-prefix"];
   //Normalize project key to without dash
   if (prefix.at(-1) === "-") {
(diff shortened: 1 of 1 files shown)


teziovsky:
Maybe a good option would be to have an accessory to filter by a specific project, rather than opening settings and changing the project key every time you want to change the current project.

I mean something like that on the screenshot. A dropdown with the list of projects to select (with an option to show all) and the selected option could be saved in the cache to persist between opens of that extension.

height 2023-08-25 at 19 30 53

Just leaving that suggestion, you can think about it! Cool that you decided to implement it 🔥

iltimasd:
Its a good idea, I'll see how viable/realistic it is to implement with JIRA's APIs

> That could be a neat idea, but I don't use JIRA myself so have no idea about it's impact - would it make sense @iltimasd

PR #8115, opened 2023-08-28
Add conventional-commits extension

tinybighuman:
## Description

Initial release of the extension. It's aimed to provide a list and description of conventional commit prefixes that can be inserted into the active window.

Aimed at developers to provide them with an easy cheat sheet of all the CC prefixes and their definitions.

## Screencast

cc-screenshot-2

## Checklist

### DIFF
--- a/extensions/conventional-commits/README.md
+++ b/extensions/conventional-commits/README.md
@@ -1,7 +1,7 @@
-# Conventional Commits - Extension for Raycast
+# Conventional Commits
 
 ![Command Screenshot](./screenshots/cc-screenshot.png)
 
-Provides a list and description of conventional commit prefixes that can be inserted into the active window. This is an extension written for [Raycast](https://www.raycast.com).
+Provides a list and description of conventional commit prefixes that can be inserted into the active window. Aimed at developers to provide an easy cheat-sheet of all the Conventional Commit prefixes and their meanings.
 
 ![Example of the list of commands available](./screenshots/cc-screenshot-2.png)


mil3na:
Hi!

Thank you so much for your contribution! My name is Milena and I will be reviewing your extension today.
Do you have any data to back up saying that these are the most common prefixes for commit messages? I think that this kind of thing varies a lot between orgs and devs. So much that I believe an individual would benefit more by having theirs setup as a snippet. 

Do you mind explaining what would be the advantage of your extension over snippets? Would the values change over time? Do you have any plans on make it dynamic? Otherwise, I don't think this is a good candidate to our store, I'm afraid.

niemtec:
Hi, I'll try to answer each of your points below 😄 

> Do you have any data to back up saying that these are the most common prefixes for commit messages? I think that this kind of thing varies a lot between orgs and devs.

Yes, these are sourced directly from the [Conventional Commits Specification](https://www.conventionalcommits.org/en/v1.0.0/#specification) and enriched using the referenced [Angular convention](https://github.com/angular/angular/blob/22b96b9/CONTRIBUTING.md#-commit-message-guidelines) for additional usecases.

I agree that some prefixes might vary between companies/individuals, but these are defined by the aforementioned specifications and would therefore likely be present in any variation. 

> So much that I believe an individual would benefit more by having theirs setup as a snippet.... Do you mind explaining what would be the advantage of your extension over snippets?

I disagree with your point about using snippets for this. A snippet would expand or replace a placeholder text, this extension is meant to insert the prefix after the user determines the most appropriate choice for their current situation.

It provides a reference point that is informative (since it defines what each prefix means) and allows the user to make a selection from the list instead of memorising expansion strings.
It is both a cheat sheet and a learning/reference tool to familiarise oneself with the convention.

> Would the values change over time? Do you have any plans on make it dynamic?

The prefix values are defined by the convention and are not likely to change, though improvements would be implemented if that should ever happen.

Hope that answers your questions 😄

PR #8154, opened 2023-08-30
[Hue] Update and bugfixes

pindab0ter:
## Description

* Hide dimming actions when a light does not support them
* Fix bug where toggling a group on will set its brightness to the lowest possible setting
* Fix “Toggle All Lights” behaviour
  * When no lights are on, all lights will turn on
  * When some, but not all lights are on, the preference will be used (default: Turn all off)
  * When all lights are on, all lights will turn off

### Known issues

* Transition Time preference does not apply to toggling groups, as this causes the group to turn on with minimal
  brightness due to a bug in the Hue V2 API.

## Checklist

### DIFF
--- a/extensions/hue/CHANGELOG.md
+++ b/extensions/hue/CHANGELOG.md
@@ -9,7 +9,13 @@
   * When some, but not all lights are on, the preference will be used (default: Turn all off)
   * When all lights are on, all lights will turn off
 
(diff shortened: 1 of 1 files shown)


pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #8206, opened 2023-09-03
Add perchance-generator extension

Visual-Studio-Coder:
## Screencast
image
So here's what you do. Go to perchance.org, and click on "generators" and then get the endpoint of that generator URL. Now, you just input that into Perchance Generator for Raycast, and you're pretty much done.

## Checklist


pernielsentikaer (commented):
Hi 👋 

How do I use this? I'm not sure what it does, could you create a screencast of the flow?

Visual-Studio-Coder:
try inputting `plotgenertor` to this generator's argument

pernielsentikaer:
Is there a reason why it takes to long time to generate? It take more than 15 seconds here

Visual-Studio-Coder:
It's just how their api works unfortunately. It's faster on future requests

pernielsentikaer:
Sounds like the content should be shown in a `<Detail>` view instead of a toast?

Visual-Studio-Coder:
Well no. The reason is in a toast is because it was designed so that people can spam generations and only copy when they find something good. Sometimes it's also just for a quick laugh, so I wouldn't want a detail view

pernielsentikaer:
You could use useFetch's revalidate inside the `<Detail>`, like I did with [Dad jokes](https://www.raycast.com/pernielsentikaer/dad-jokes)

Visual-Studio-Coder:
Cmon. I really liked the toast messages style. Even the dice rolling extension uses it. Could I still keep the toast message layout or do I HAVE to change it to detail.

pernielsentikaer:
There is way too much text for the toast, it's not making the best UI/UX, also, from a detail pane you can generate it again with just a hit on enter 🙂

Raycast 2023-09-15 at 11 29 01

Visual-Studio-Coder:
Ok that makes sense. I'll work on fixing this

Visual-Studio-Coder:
Ok. I just copied your interface and added some actions and an image.

extensions/perchance-generator/package.json
```diff
@@ -0,0 +1,73 @@
+{
+  "$schema": "https://www.raycast.com/schemas/extension.json",
+  "name": "perchance-generator",
+  "title": "Perchance Generator",
+  "description": "Input the project name and receive the generated text.",
```
pernielsentikaer, line 5: Project name?

extensions/perchance-generator/package.json
```diff
@@ -0,0 +1,73 @@
...
+    "random-text-generator",
+    "keyboard",
+    "words"
+  ],
+  "license": "MIT",
+  "commands": [
+    {
+      "name": "index",
+      "title": "Generate Using Perchance",
+      "description": "Template with a plain detail view",
```
pernielsentikaer, line 28: Should be changed

pernielsentikaer (changes requested):
Thanks for the update 🔥 

Right when entering the extension it shows the success toast, after waiting around 10 seconds the data is presented, could be good with isLoading and a toast saying loading instead

Is that possible?

extensions/perchance-generator/package.json
Visual-Studio-Coder: Yeah, I believe that's what they're called . . .

extensions/perchance-generator/package.json
Visual-Studio-Coder: Ok, I'll fix that.

extensions/perchance-generator/package.json
Visual-Studio-Coder: Hmmm. I did a little more digging and found that they also called them _generator names_ OR project names, so will change to first one

Visual-Studio-Coder:
You spelled it wrong...

Visual-Studio-Coder:
I think it might be an internet issue.

Visual-Studio-Coder:
Wait sorry, its that "undefined" thing that I forgot to put in quotes. I fixed it

pernielsentikaer:
I have made a few minor corrections. Could you please review to ensure that nothing has been broken? :)

Visual-Studio-Coder:
Ok yeah, it works. Thanks for adding the corrections.

Visual-Studio-Coder:
@pernielsentikaer

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #8259, opened 2023-09-06
Add emoticons extension

blekmus:
## Description

The command `Show Emoticons` displays a list of categories featuring emoticons.

## Screencast

emoticons-1

## Checklist

### DIFF
--- a/extensions/emoticons/src/index.tsx
+++ b/extensions/emoticons/src/index.tsx
@@ -29,17 +29,27 @@ export default function Command() {
       ],
     },
     {
-      id: 3,
       type: "Surprise",
       items: ["(・o・)", "(⊙_⊙)", "＼(°o°)／", "(@_@)", "(●_●)", "(⁠‘⁠◉⁠⌓⁠◉⁠’⁠)", "(((;ꏿ_ꏿ;)))", "(ﾉﾟ0ﾟ)⁠ﾉ"],
     },
     {
-      id: 4,
       type: "Dissapproval",
-      items: ["ಠ_ಠ", "ಠ⁠︵⁠ಠ", "ರ_ರ", "ಠಿ_ಠಿ", "ಠಿ_ಠ", "ಠ⁠﹏⁠ಠ", "(￣～￣)", "ಠ╭╮ಠ"],
+      items: [
+        "ಠ_ಠ",
+        "ಠ⁠︵⁠ಠ",
+        "ರ_ರ",
+        "ಠಿ_ಠಿ",
+        "ಠಿ_ಠ",
+        "ಠ⁠﹏⁠ಠ",
+        "(￣～￣)",
+        "ಠ╭╮ಠ",
+        "(¬､¬)",
+        "(⩺_⩹)",
+        "（＞д＜）",
+        "⋋_⋌",
+      ],
     },
     {
-      id: 5,
       type: "Indifference",
       items: [
         "(￢_￢)",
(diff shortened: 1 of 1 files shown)


pernielsentikaer:
It feels a bit like this one: https://www.raycast.com/yalishanda/kaomoji-search - is it kinda the same?

blekmus:
@pernielsentikaer I made this because I feel that organizing them this way is better than what yalishanda had done. However, the downside is that the emojis can't be copied by clicking on them in my implementation.

I acknowledge that they are similar so include this only if you feel that it's enough of an improvement from kaomoji-search.

pernielsentikaer:
Maybe we can enhance @yalishanda42's version if yours containing something that could be nice to have?

yalishanda42:
I can see why @blekmus has chosen not to add a command to my extension. 
The `kaomoji-search` extension uses [`asciilib`](https://github.com/iansinnott/asciilib) which in theory provides a category for each kaomoji, but in practice those categories are either undefined or not expressive/specific enough for the purposes of this new extension. 
If one finds another library that would fit our common purpose that would be great! I however have no knowledge of such, and apparently @blekmus doesn't have as well, so having that in mind I can easily see why he has decided to add specific kaomoji by hand to those hand-picked categories.
That, of course, doesn't mean it should definitely live in a separate extension - if this new proposal conveys the same idea of listing kaomoji, then it can just be a separate command of the original extension, regardless of whether it uses the same data source or not. After all, to the end user everything is just a collection of kaomojis - it would be more confusing to separate them.
Regarding this, I have a question about the name - *emoticon* means also symbols like `:D`, `:)` etc, but all I can see here is only kaomoji. @blekmus  Do you plan on adding different emoticons? This could be a point in which our two extensions start to actually differ from the user perspective and then it would be make much more sense to have this be a separate extension and named like this.

pernielsentikaer:
There could also be a manually updated JSON file that could make the binding between the kaomoji and the category. I'm not sure if that's too much work?

blekmus:
I primarily based this extension on the Emoticons collection found on Google's keyboard application GBoard. Thus I followed their name as well. I plan to extend this by adding more simpler emoticons such as `:D` as well. Which as @yalishanda42 says might be a point of difference. I was planning on keeping this as a highly curated list of emoticons that primarily express emotions rather than animals and the like.

But I think adding this as a separate command into the already existing kaomoji extension would be more practical in that sense. Because even if simpler ones were to be added, this would primarily be a collection of emoticons that are handpicked rather than brought in from a database like asciilib. So it would be more practical to exist as a command under a umbrella `emoticon` or `kaomoji` extension. Where one command is hand picked and the other is to search from the asciilib database. What do you think @yalishanda42?

@pernielsentikaer I believe that is quite a lot of work considering the number of kaomojis in that listing.

yalishanda42:
> .

Yeah, maybe naming the extension "Emoticons" and the commands "Search Kaomoji" and "List Emoticons" for example would be better? Or something like that

blekmus:
I agree. I'm willing to do that. Is it fine if I just extend this pull request with code brought in from yours @yalishanda42?

yalishanda42:
> I agree. I'm willing to do that. Is it fine if I just extend this pull request with code brought in from yours @yalishanda42?

Hm but if we delete the existing kaomoji-search extension, what about the existing users of it?

blekmus:
Yeah, that's a problem. Same issue if we rename the extension as well. @pernielsentikaer thoughts?

pernielsentikaer:
If you keep the folder structure and name, then there is nothing wrong with a new title - so we're going to mix up one extension?

Issue #8272, opened 2023-09-07
[My Daily Log] ...


bakari45:
### Extension

https://www.raycast.com/frugoman/my-daily-log

#### Error:
```
Error: ENOENT: no such file or directory, open '/Users/bakari/Documents/DailyLog/2023-09-07.json'

Object.openSync:node:fs:601:3
Object.writeFileSync:node:fs:2249:35
Re.saveLogs:createLogCommand.js:30:2617
Re.create:createLogCommand.js:30:2824
```

### Steps To Reproduce

Opened the log window and wrote a log. 

### Current Behaviour

I'm trying Daily Log for the first time, but I get an error message when I try to submit a log. 

### Expected Behaviour

Expect it to save the log.

frugoman:
@bakari45 I've fixed this issue! Please give it a try!!

Thanks for reporting it! Feel free to close it if it's resolved for you!

bakari45:
Yes, thanks, it works now. 
Do you think you will ever enlarge the input text window to make it easier to edit text or write longer logs?

PR #8297, opened 2023-09-08
fixing non existing folder bug

frugoman:
This pull request fixes a bug in the my-daily-log extension that caused it to crash when the folder for logs did not exist. The bug was reported in issue #8272.

The main changes are:

* Added a new interface DataStorage to abstract the data storage operations
* Implemented a concrete class LocalFilesDataStorage that uses the file system to store and read data
* Injected an instance of LocalFilesDataStorage into the JsonDailyLogRepository constructor
* Refactored the JsonDailyLogRepository methods to use the * DataStorage instance instead of directly accessing the file system
* Added a check for the existence of the folder before writing or deleting data
* Updated the changelog with the fix details


pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #8315, opened 2023-09-10
[Bitwarden] Download own CLI binary instead of requiring local installation

jomifepe:
## Description

For a while, this extension has required the installation of the Bitwarden CLI, which is a step that may lead to issues (e.g. incorrect installations), doesn't allow the developers to control when the users have the latest CLI features, and according to [Raycast's recommendations](https://developers.raycast.com/basics/prepare-an-extension-for-store#binary-dependencies-and-additional-configuration), is something that should be avoided. This PR aims to remove this step, in favor of using its own CLI binary.

### What was done
- Download the Bitwarden CLI binary when opening any extension command. The binary is downloaded from the official  [bitwarden clients github](https://github.com/bitwarden/clients/releases)
- Check the integrity of the downloaded file by using a SHA256 hash (also available alongside the binary release)
- Update and delete old CLI binaries.

✅ Tested on `arm64` and `x64` macOS.

Local installations are still supported, and users can still specify the path to the local binary if they so desire.

### Updating
A version is associated with the binary, so if the version specified in the extension is changed, a new binary will be downloaded automatically once the user gets the update (as seen in the screencast).

### Cleanup
Because CLI binaries are 100MB+ in size, when downloading a new binary, we check if there is any `bw-*` file present under the support directory and delete them all before downloading a new one (as seen in the screencast).

## Screencast

https://github.com/raycast/extensions/assets/7235666/10a8e931-d4a0-4658-825d-887ff0f4d62f

## Checklist

### DIFF
--- a/extensions/bitwarden/CHANGELOG.md
+++ b/extensions/bitwarden/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Bitwarden Changelog
 
+## [Improvements] - 2023-09-11
+
+- Download, if needed, the official CLI binary on extension start, instead of depending on the user to install it
+- Cleanup possible existing CLI binaries when downloading a new one, to save space
+
 ## [Fix] - 2023-08-29
 
 - Fixed crash in Logout command


pernielsentikaer (commented):
Do we need to update the extension every time a new build is made from Bitwarden, or doesn't that happen that often?

jomifepe:
> Do we need to update the extension every time a new build is made from Bitwarden, or doesn't that happen that often?

It's not very often, at least for major features, the CLI isn't as updated as the other clients. 

The issue with the current setup is that it's very hard to react to new CLI features because people will have old versions of the CLI installed on their machines.

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

Issue #8164, opened 2023-08-31
[SSH Connection Manager] Read connection list from ~/.ssh/config


yaronuliel:
### Extension

https://www.raycast.com/dimagrossman/ssh-manager

### Description

It would be great to have the app (at least optionally) - read ssh hosts lists from `~/.ssh/config` which is the standard place for defining remote sever, identity keys, etc. with openssh

This feature will make the extension 10x faster to getting start with, and will remove a lot of hassle from managing the connection list

### Who will benefit from this feature?

Users that already have a list of ssh server defined for openssh

### Anything else?

To fully support the `Host` definition in openssh - the `user` property of the connection should also be considered optional.
It is very common to set the connection username in the openssh config, so that connecting to a server doesn't require reentering the information on any connection

Today when the username field is left blank - the extension tries to connect to `@ssh.example.com` - which lead to an error

pernielsentikaer:
Did you see this @scopsy

silas229:
The extension does not recognise the Hostname.
I think it would be easier if it would just insert the alias (`ssh <alias>`)

In sshconfig:
<img width="253" alt="SCR-20231205-nbab" src="https://github.com/raycast/extensions/assets/23215125/f14853d1-7575-43af-ac85-16055e0bdb13">

In Raycast:
<img width="284" alt="SCR-20231205-nati" src="https://github.com/raycast/extensions/assets/23215125/f3d43064-f767-4f8a-bb8b-8a1e22066a41">

In iTerm:
<img width="188" alt="SCR-20231205-nbib" src="https://github.com/raycast/extensions/assets/23215125/17d5e5ba-753e-43f3-8f30-66317e81e93b">

silas229:
unstall

pernielsentikaer:
Can you look into that @solid-pixel

solid-pixel:
> Can you look into that @solid-pixel

Sorry I can't, I don't have the time nor the energy to contribute at the moment.
Doesn't the extension already do what OP is asking though?

silas229:
@solid-pixel no, it has bugs:
https://github.com/raycast/extensions/issues/8164#issuecomment-1840823197

silas229:
.

silas229:
.

silas229:
.

dongnak:
please fix https://github.com/raycast/extensions/issues/8164#issuecomment-1840823197

pernielsentikaer:
@scopsy are you still around?

PR #8343, opened 2023-09-11
feat(ssh-manager) #8164 Read ssh hosts from openssh config

yaronuliel:
## Description
🎉 New Feature in SSH Manager Extension 
This PR adds the option to auto populate the list of ssh connections from the user's openssh config 

For example, if the user's `~/.ssh/config` file contains the following content:

```
Host project-prod-on-aws
    HostName realhost.example.com
    User ec2-user
    IdentityFile ~/.ssh/keys/my-key
    Port 4441

Host project-staging
   User ubuntu
   IdentityFile ~/.ssh/keys/other

```

The `Show connections` screen of the extension will list these host, without needing the user to manually insert anything:

Screenshot 2023-09-12 at 0 24 22

In case the user for some reason want to cancel this feature - it can be done in the `Configure extension` page

## Screencast

Screenshot 2023-09-12 at 0 24 22

Screenshot 2023-09-12 at 0 28 15

## Checklist

### DIFF
--- a/extensions/ssh-manager/CHANGELOG.md
+++ b/extensions/ssh-manager/CHANGELOG.md
@@ -1,5 +1,9 @@
 # ssh-manager Changelog
 
+## [Read SSH Hosts from ~/.ssh/config] - 2023-09-11
+
+- Allow auto suggest of SSH hosts defined in `~/.ssh/config` file
+
 ## [Make user field optional] - 2023-09-06
 
 - Make user field optional in connection config - to support user-less hosts (e.g. pseudo hosts that are configured in `~/.ssh/config`)
--- a/extensions/ssh-manager/package-lock.json
+++ b/extensions/ssh-manager/package-lock.json
@@ -8,8 +8,11 @@
       "license": "MIT",
       "dependencies": {
         "@raycast/api": "^1.31.0",
+        "@raycast/utils": "^1.9.1",
+        "glob": "^10.3.4",
         "nanoid": "^3.3.2",
-        "run-applescript": "^6.0.0"
+        "run-applescript": "^6.0.0",
+        "ssh-config": "^4.4.0"
       },
       "devDependencies": {
         "@types/node": "~16.10.0",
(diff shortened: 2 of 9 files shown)


khurshid-alam:
Sometimes I just want to copy hostname or ip address  from action with shortcut ^c. Can this be added ?

pernielsentikaer:
I think we just released an older pull request from the queue, adding this functionality. Could you check that, @yaronuliel?

yaronuliel:
@pernielsentikaer I've commented on the [other PR](https://github.com/raycast/extensions/pull/7614#issuecomment-1728498776) and raised some concerns about the solution that was implemented and merged there... it is similar, but not exactly the same. I personally won't use it since [it might overwrite configuration](https://github.com/raycast/extensions/pull/7614/files#diff-17c94cfe646c9eb1599d9752a9114c5d611cf7c175633cc354057ae33b60c683R79) that I have on my machine


> @RtiM0 @pernielsentikaer Given that I've authored a different PR #8343 to address a very similar issue - there is a number of problems with this current implementation that I think worth discussion or reference
> 
> 1. It doesn't support the `Include`([info](https://man.openbsd.org/ssh_config#Include)) directive in `~/.ssh/config`
> 2. Since you only support a very small subset of `~/.ssh/config` options, and parsing the file in a naive approach - writing to the file from raycast might cause a catastrophe and override user's valid configurations, and this is not a very good idea
> 3. The script doesn't support case-insensitive directive, while ssh-config do
> 
> I believe that the solution should be that records from `~/.ssh/config` are read only, while the user can still add additional records to localStorage, the 2 should co-exist and not be alternatives

I really hope that my concerns will be addressed

pernielsentikaer:
So does does this PR take of the things your mentioning @yaronuliel, or do you think @RtiM0 should?

Could you also look into the merge conflicts?

RtiM0:
@pernielsentikaer I think it's up to raycast to choose between:

- A mixture of localStorage and ssh-config where ssh config is read-only and new config will be only added to localStorage.

- Option to choose between either localStorage and ssh-config. Both sources being completely separate and support modification (add/remove).

The 2nd approach is currently merged, @yaronuliel has raised the issues that surround its current implementation.

I have no opinion here, we can revert that PR and maybe merge this one.

@yaronuliel please correct me if I didn't summarise your approach correctly.

PR #8373, opened 2023-09-13
Update keepassxc extension

pabroux:
## Description

- New feature/action "Paste TOTP" so as to be more productive;
- Respecting the KeePassXC shortcut logic as much as possible since `cmd + P` isn't an available custom shortcut anymore for extensions;
- <del>Giving a tempory solution for the "Copy Password" (why? when the password contains a backslash, that action fails due to AppleScript in the `protectedCopy` function);</del>
- Updating the README.md accordingly;
- Fixing the `protectedCopy` function to handle passwords with backslash.

## Checklist

### DIFF
--- a/extensions/keepassxc/CHANGELOG.md
+++ b/extensions/keepassxc/CHANGELOG.md
@@ -1,5 +1,13 @@
 # KeepassXC Extension Changelog
 
+## [Fixes] - 2023-10-03
+
+- New feature/action "Paste TOTP" so as to be more productive
+- Respecting the KeePassXC shortcut logic as much as possible since cmd + P isn't an available custom shortcut anymore for extensions
+- Giving a tempory solution for the "Copy Password" (why? when the password contains a backslash, that action fails due to AppleScript in the protectedCopy function)
+- Updating the README.md accordingly
+- Using the build in Clipboard.Copy's concealed type 
+
 ## [Fix] - 2023-06-20
 
 - Fixed a logic bug


Sulfyderz:
@zkytech , by the way I think `protectedCopy` is quite useless. The only case where I see it as interesting is when you are away of your computer and someone checks your Raycast copy history. Feel free to share your agenda about that function.

zkytech:
**Shortcut in KeepassXC**:

- Copy Password: `CMD + C`

**About protectedCopy**: 

If you can see those password you have copied in the `Clipboard History` extension, I think it is unsafe. 

My suggestion is to remove backslash from your password. Backslash is too spetial and rare to use in password, it is not worth removing protectedCopy for this.

Sulfyderz:
Hello @zkytech 

> **Shortcut in KeepassXC**:
> 
> * Copy Password: `CMD + C`

I tried to apply it but it doesn't work. Raycast doesn't allow that custom shortcut as well (anymore ?). The current one `CMD + P` doesn't work anymore. That's why I suggest `CMD + G`.

> **About protectedCopy**:
> 
> If you can see those password you have copied in the `Clipboard History` extension, I think it is unsafe.
> 
> My suggestion is to remove backslash from your password. Backslash is too spetial and rare to use in password, it is not worth removing protectedCopy for this.

I have just fixed `protectedCopy`. Now, it works with backslash 😉. I have also put it back for copy TOTP and copy Password.

Feel free to share your feedback !

zkytech (approved):
OK

zkytech:
Thanks for your contribution! @Sulfyderz

pernielsentikaer:
In 1.59.0 we just released `concealed` which you can find [here](https://developers.raycast.com/api-reference/user-interface/actions#props-1) - you should use that instead 🙂

Copy Password could also be cmd+shift+c or cmd+. as many things in Raycast use

Sulfyderz:
Hello @pernielsentikaer,

> In 1.59.0 we just released `concealed` which you can find [here](https://developers.raycast.com/api-reference/user-interface/actions#props-1) - you should use that instead 🙂

Done. Feel free to share your feedback (@zkytech you too 😉).

⚠️ By the way, one of your CI fails because of `Clipboard.copy` saying `src/utils/keepassLoader.ts(140,37): error TS2554: Expected 1 arguments, but got 2..` According to your [doc](https://developers.raycast.com/api-reference/clipboard#clipboard.copyoptions), it can have two parameters. On my side, everything works. I have no error message.

> Copy Password could also be cmd+shift+c or cmd+. as many things in Raycast use

Thank you. I'd rather stay with `cmd+G` since it's less complex, especially if you have another daemon listener for scripts (e.g. [skhd](https://github.com/koekeishiya/skhd) with [yabai](https://github.com/koekeishiya/yabai)).

zkytech:
> ⚠️ By the way, one of your CI/GitHub Action fails because of `Clipboard.copy` saying `src/utils/keepassLoader.ts(140,37): error TS2554: Expected 1 arguments, but got 2.`. According to your [doc](https://developers.raycast.com/api-reference/clipboard#clipboard.copy), it can have two parameters. On my side, everything works. I have no error message.


Upgrade the `raycast/api` in `package.json` to `1.59.0`.

Sulfyderz:
> > ⚠️ By the way, one of your CI/GitHub Action fails because of `Clipboard.copy` saying `src/utils/keepassLoader.ts(140,37): error TS2554: Expected 1 arguments, but got 2.`. According to your [doc](https://developers.raycast.com/api-reference/clipboard#clipboard.copy), it can have two parameters. On my side, everything works. I have no error message.
> 
> Upgrade the `raycast/api` in `package.json` to `1.59.0`.

Thank you. My bad.

zkytech:
Looks good to me. @pernielsentikaer

PR #8375, opened 2023-09-13
Update hacker-news extension

hakob8956:
## Description
Add functionality to summarize articles with OpenAI GPT-3.5 Turbo. 
The OpenAI key is optional. If the user does not provide it, the summarize function will be unavailable.
image
image

## Checklist

### DIFF
--- a/extensions/hacker-news/src/frontpage.tsx
+++ b/extensions/hacker-news/src/frontpage.tsx
@@ -69,10 +69,12 @@ function setTitle(title: string, topic: Topic) {
   return title;
 }
 
-function getWindowContent(title: string, topic: Topic) {
+function getWindowContent(title: string, topic: Topic, isLoading: boolean) {
   const _title = setTitle(title, topic);
   const _informationContent = "To summarize the article, please press *'Cmd(⌘) + ['*.";
-  return "# " + _title + "  \n" + _informationContent;
+  const _isLoading = "Loading...";
+  if (isLoading) return "# " + _title + "  \n" + _isLoading;
+  else return "# " + _title + "  \n" + _informationContent;
 }
 
 function StoryListItem(props: {
@@ -86,7 +88,7 @@ function StoryListItem(props: {
   const [isLoading, setIsLoading] = useState(false);
 
   const fetchSummary = async () => {
-    if (!openAI) {
+    if (!isValidKey) {
       throw new Error("openAI is not initialized");
     }
     if (props.item.link) {
@@ -107,9 +109,7 @@ function StoryListItem(props: {
       detail={
         <List.Item.Detail
           markdown={
-            isLoading
-              ? "*Loading...*"
-              : summary || getWindowContent(props.item.title || "No Title", props.topic || Topic.FrontPage)
+            summary || getWindowContent(props.item.title || "No Title", props.topic || Topic.FrontPage, isLoading)
           }
         />
       }
@@ -156,7 +156,7 @@ function Actions(props: {
           title="Toggle Summarize Window"
           shortcut={{ modifiers: ["cmd"], key: "]" }}
           onAction={async () => {
-            if (openAI != null) props.setIsShowingDetail((prev) => !prev);
+            if (openAI != null || !isValidKey) props.setIsShowingDetail((prev) => !prev);
             else
               await showToast({
                 style: Toast.Style.Failure,
@@ -169,7 +169,15 @@ function Actions(props: {
           <Action
             title="Summarize Article"
             onAction={async () => {
-              props.fetchSummary();
+              if (!isValidKey) {
+                showToast({
+                  style: Toast.Style.Failure,
+                  title: "Cannot execute command",
+                  message: String("You don't have Raycast Pro subscrition you need to provide openAI key"),
+                });
+              } else {
+                props.fetchSummary();
+              }
             }}
             shortcut={{ modifiers: ["cmd"], key: "[" }}
           />


extensions/hacker-news/package.json
```diff
@@ -22,10 +23,22 @@
...
   ],
+  "preferences": [
+    {
+      "default": "",
+      "description": "Enter the OpenAI access token for the summarize function.(BETA)",
+      "name": "openaiAccessToken",
+      "placeholder": "[Enter Your Access Token]",
+      "required": false,
+      "title": "OpenAI access token (Summarizer function).",
+      "type": "textfield"
```
pernielsentikaer, line 34:
```suggestion
      "type": "password"
```

extensions/hacker-news/src/frontpage.tsx
```diff
@@ -75,6 +142,37 @@ function Actions(props: { item: Parser.Item }) {
           />
         )}
       </ActionPanel.Section>
+
+      <ActionPanel.Section>
+        <Action
+          title="Toggle summarize window"
```
pernielsentikaer, line 148:
```suggestion
          title="Toggle Summarize Window"
```

extensions/hacker-news/src/frontpage.tsx
```diff
@@ -75,6 +142,37 @@ function Actions(props: { item: Parser.Item }) {
...
+            } else {
+              await showToast({
+                style: Toast.Style.Failure,
+                title: "Cannot execute command",
+                message: "You should enable summarize window first",
+              });
+            }
+          }}
+          shortcut={{ modifiers: ["cmd"], key: "]" }}
+        />
```
pernielsentikaer, line 174: Maybe only show this `Action` if the window is open? Also, remember the icon.

extensions/hacker-news/src/frontpage.tsx
```diff
@@ -75,6 +142,37 @@ function Actions(props: { item: Parser.Item }) {
...
+            else
+              await showToast({
+                style: Toast.Style.Failure,
+                title: "Cannot execute command",
+                message: "To enable the summarize function, you should provide an openAI key on the preferences page.",
+              });
+          }}
+        />
+        <Action
+          title="Summarize article"
```
pernielsentikaer, line 161:
```suggestion
          title="Summarize Article"
```

extensions/hacker-news/src/frontpage.tsx
pernielsentikaer: Also, remember the icon.

pernielsentikaer (changes requested):
Hi 👋 

- Add a changelog entry

- Maybe also allow access for Raycast AI users? 

extensions/hacker-news/src/frontpage.tsx
```diff
@@ -75,6 +142,31 @@ function Actions(props: { item: Parser.Item }) {
           />
         )}
       </ActionPanel.Section>
+
+      <ActionPanel.Section>
+        <Action
+          title="Toggle Summarize Window"
```
pernielsentikaer, line 148: Only show if you have OpenAI/Raycast AI

pernielsentikaer:
Seems like it doesn't work if you have Raycast AI, I didn't enter a OpenAI key and getting this

hacker-news 2023-10-03 at 11 18 43

hakob8956:
> Seems like it doesn't work if you have Raycast AI, I didn't enter a OpenAI key and getting this
> 
> hacker-news 2023-10-03 at 11 18 43

It should work now.

pernielsentikaer:
Still facing the same issue, I have Raycast AI and it still just gives me the error - it should work if I only had Raycast AI, right?

Raycast-wAVyLlOk@2x

hakob8956:
> Collaborator

Hi, I'm checking the rayCast AI subscription with environment.canAccess(AI). I don't have a subscription, but in my case, I just debug it by changing it to a fixed value like 'true' and it works for me. So please check that branch is updated and could you check if environment.canAccess(AI) returns true in your case? 
Thanks.

extensions/hacker-news/src/frontpage.tsx
```diff
@@ -75,6 +142,31 @@ function Actions(props: { item: Parser.Item }) {
...
+          shortcut={{ modifiers: ["cmd"], key: "]" }}
+          onAction={async () => {
+            if (openAI != null) props.setIsShowingDetail((prev) => !prev);
```
pernielsentikaer, line 151: There is no check for Raycast AI here, also, is there a way to just show the result in the next screen without having to run another action panel item?

extensions/hacker-news/src/frontpage.tsx
```diff
@@ -47,19 +61,72 @@ function setTitle(title: string, topic: Topic) {
...
+  index: number;
+  topic: Topic | null;
+  setIsShowingDetail: React.Dispatch<React.SetStateAction<boolean>>;
+  isShowingDetail: boolean;
+}) {
+  const [summary, setSummary] = useState<string | null>(null);
+  const [isLoading, setIsLoading] = useState(false);
+
+  const fetchSummary = async () => {
+    if (!openAI) {
```
pernielsentikaer, line 81: Same here

pernielsentikaer (changes requested):
Try to look around here 🙂

extensions/hacker-news/src/frontpage.tsx
hakob8956: Hi, in line 22 `const openAI = new OpenAIModule(openAIKey);` actually, it shouldn't be null because it initializes it. In OpenAIModule, there is a logic that will check if you are using RaycastAI or not.
In OpenAIModule class you can see
`if (environment.canAccess(AI)) return await this.raycastAiRequest(prompt);
    else if (this.openai != null) return await this.gptRequest(prompt);
    else throw new Error("AI module not initialized");`

hakob8956: Actually, it was tough because there should be some action that will toggle that window. One action can toggle the window and call AI, but if you go to the next item, it will automatically make an AI request that costs money. So, in this way, I'm making sure that the user actually wants to call AI to summarize the article.

hakob8956:
@pernielsentikaer  I just realized that PR wasn't up-to-date. I thought it was. I'm getting nothing new to publish, but actually, there is something to publish.
`ready -  validate git repository
ready -  getting fork
ready -  preparing clone
ready -  checking for new contributions
error -  preparing extension
seems like there is nothing new to publish`

extensions/hacker-news/src/frontpage.tsx
```diff
@@ -1,19 +1,34 @@
...
+
+const isValidKey = openAIKey || environment.canAccess(AI);
+if (!isValidKey) {
+  showToast({
+    style: Toast.Style.Failure,
+    title: "Cannot execute command",
+    message: String("You don't have Raycast Pro subscrition you need to provide openAI key"),
+  });
+}
+const openAI = new OpenAIModule(openAIKey);
```
hakob8956, line 22: @pernielsentikaer 
In here it should be created and it cannot be null

hakob8956:
Please check it again 
Hope this time will work :)

pernielsentikaer:
Hi @hakob8956! 👋 

It works as expected now, but I think we can improve the user experience (UX). It feels wrong to have two hotkeys to retrieve the data, and it would also consume a lot of credits if it only summarizes the next item in the list.

Here's my suggestion: let's switch to a new markdown view where it performs its magic. This way, it will feel more like Raycast.

Do you need my assistance with this, or can you handle it on your own?

hakob8956:
Hi @pern

> Hi @hakob8956! 👋
> 
> It works as expected now, but I think we can improve the user experience (UX). It feels wrong to have two hotkeys to retrieve the data, and it would also consume a lot of credits if it only summarizes the next item in the list.
> 
> Here's my suggestion: let's switch to a new markdown view where it performs its magic. This way, it will feel more like Raycast.
> 
> Do you need my assistance with this, or can you handle it on your own?

Hi @pernielsentikaer , I would like to understand the concept of markdown view more. Will it no longer list items? What exactly do you mean by that? Could you give an example of a plugin or some mocks?

pernielsentikaer:
Sorry for the delay here @hakob8956 - Somehow I missed the notification regarding this PR

First opening the detail panel on an item and then running another command feels a bit heavy and is not a good UX, what would be is to push to a <detail /> pane with the data instead

PR #8389, opened 2023-09-14
Slack Status: Add OAuth + AI

thomaspaulmann:
## Description

I revamped the extension with the following:
- Migrate to latest API
- Integrate OAuth for easier login, maintain backwards comaptibility for access token
- Add AI to set custom status
- Add optimistic updates to make extension feel snappier
- Revamp README.md to reflect new authentication

Remaining steps to publish the extension:
- [x] Finalize proxy server
- [x] Publish Slack OAuth app
- [x] Decide on keeping/removing access token

New contribution points for afterwards:
- Menu bar command to see current status and change it
- Integrate do-not-disturb with actions to toggle it and include in presets
- New action to save current status as preset

## Screencast

N/A - Nothing changed in the UI

## Checklist

### DIFF
--- a/extensions/slack-status/README.md
+++ b/extensions/slack-status/README.md
@@ -2,6 +2,10 @@
 
 Extension for quickly changing your Slack status. Set up presets for common statuses or set a custom one. Make sure to explore available actions in the action panel (`⌘ + K`). And if you want to spice it up, use Raycast AI to set your status. Simply type what you are doing and hit return to let AI pick an emoji, title, and duration of your status.
 
+## Automatic Setup
+
+Simply open the extension and authenticate with OAuth to your workspace. Afterwards use the extension to quickly set your Slack status right from Raycast.
+
 ## Manual Setup
 
 The extension supports OAuth, which makes it easy to connect to your Slack workspace. If you prefer a manual setup, follow those steps to create your own Slack app and use it's access token to access your Slack workspace.


PitNikola:
Thanks for adding OAuth capability @thomaspaulmann! It's sooo much better

You have my approval with whatever you do with this extension 😅

extensions/slack-status/src/actions.tsx
```diff
@@ -0,0 +1,383 @@
...
+              `You help a Slack user set their status.
+              
+              Respond with the following JSON for the Slack status:
+              {
+                "text": <string for status text, should be short and sweet, no punctuation at the end, e.g. "Working out", "Listening to Drake's new album", "Coffe break">,
+                "emoji": <string for single emoji matching the text of the status>,
+                "duration": <optional integer for duration of status in seconds, only include if user specified duration or end of status in their description>
+              }
+
+              Current time of users status: ${new Date().toLocaleTimeString()}. User's description of their status: ${
```
thomaslombart, line 76: Should the user's description of their status be sent to the AI If it turns out to be empty?

extensions/slack-status/src/oauth.tsx
```diff
@@ -0,0 +1,52 @@
...
+  }, []);
+
+  if (!accessToken) {
+    if (environment.commandMode === "view") {
+      // Using the <List /> component makes the placeholder buggy
+      return <Detail isLoading />;
+    } else if (environment.commandMode === "menu-bar") {
+      return <MenuBarExtra isLoading />;
+    } else {
+      console.error("`withLinearClient` is only supported in `view` and `menu-bar` mode");
```
thomaslombart, line 28:
```suggestion
      console.error("`withOAuthSession` is only supported in `view` and `menu-bar` mode");
```

extensions/slack-status/src/oauth.tsx
```diff
@@ -0,0 +1,52 @@
...
+  return (props: P): ReactElement => {
+    return (
+      <OAuthSessionProvider config={config}>
+        <Component {...props} />
+      </OAuthSessionProvider>
+    );
+  };
+}
+
+export function useOAuthSession() {
```
thomaslombart, line 46: `useOAuthSession` may be confusing here since it's not a hook but simply a getter returning the access token.

mil3na:
Hi Thomas! 
Thank you for making the OAuth happen! It makes things much easier!
The extension seems to be working fine. I love the new AI feature to add new status. I wonder, maybe those could be persisted? It can also be a preference thing, if the user wants it to be persisted or not.
Also, do you mind creating a CHANGELOG?

extensions/slack-status/src/actions.tsx
thomaspaulmann: The action is only available if the `searchText` is not empty. Therefore the user's description will also not be empty.

thomaspaulmann:
> Hi Thomas! Thank you for making the OAuth happen! It makes things much easier! The extension seems to be working fine. I love the new AI feature to add new status. I wonder, maybe those could be persisted? It can also be a preference thing, if the user wants it to be persisted or not. Also, do you mind creating a CHANGELOG?

Added a Changelog entry, thanks for the reminder! I didn't persist the AI-generated status for now. Not sure if we should store them as presets. Another alternative would be adding some sort of history to search previous status but that sounded a bit too complicated. Maybe a good middle ground would be adding a new action to the current status to save it as a preset in case it isn't one yet?

thomaspaulmann:
@mil3na and @thomaslombart I updated the exension. Do you want to give it another spin? Then we can merge it 🚀

mil3na:
> Maybe a good middle ground would be adding a new action to the current status to save it as a preset in case it isn't one yet?

I like this idea. It will be nice for a next update :)

PR #8444, opened 2023-09-18
Add fixlayoutenuk extension

canab:
## Checklist

### DIFF
--- a/extensions/fixlayoutenuk/.prettierrc
+++ b/extensions/fixlayoutenuk/.prettierrc
@@ -1,0 +1,4 @@
+{
+  "printWidth": 120,
+  "singleQuote": false
+}
--- a/extensions/fixlayoutenuk/package.json
+++ b/extensions/fixlayoutenuk/package.json
@@ -4,7 +4,7 @@
   "title": "Fix Layout EnUk",
   "description": "Fixes typed text with wong keyboard layout.\nEnglish <-> Ukrainian",
   "icon": "command-icon.png",
-  "author": "canab",
+  "author": "yaroslav_kihtan",
   "categories": [
     "Productivity"
   ],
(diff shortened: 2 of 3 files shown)


pernielsentikaer:
This seems to be the same as [this extension](https://www.raycast.com/danulqua/input-source-switcher) by @danulqua

canab:
> This seems to be the same as [this extension](https://www.raycast.com/danulqua/input-source-switcher) by @danulqua

Yes, it is similar but it have different behavior. Mentioned extension transforms the text only in one chosen direction. This extension transforms text in any direction depending on typed symbols, requires to select the text being transformed and also switches input language to continue typing in a proper lang.

danulqua:
> This seems to be the same as [this extension](https://www.raycast.com/danulqua/input-source-switcher) by @danulqua

Definitely 🙂

danulqua:
Hi @canab,

Still, your extension solves the same problem. But mine was built in a way that it's easy to add new languages for people who want to contribute, and it features all characters hidden behind additional layers (when holding ⌥ and ⇧).

Also, your extension has a huge problem:

```applescript
  keystroke space using {control down}
end tell
```

The shortcut is hardcoded, but people use different shortcuts for switching languages (I use Caps Lock). Additionally, a user could switch the input source after finishing typing. In that case, the language would be switched to the wrong one because there is no checking.

I didn't implement language auto-detection because of the problems you encountered, and there are many others. It's not that easy 🙂

canab:
Hi @danulqua, this is the first version and I am going to improve language detection before switching. I created it to fit my (and coworkers) needs. Bidirectional transformation is on of them. BTW, I am OK to cancel this submission if it does not fit in the reycast requirements.

danulqua:
Raycast has awesome feature called "Organizations". 

CleanShot 2023-09-22 at 12 56 26@2x

You can create internal extensions and all users in your organization will have access to it.

Check out how the Raycast team benefits from it (link with the time code): https://youtu.be/xRnMXJcH9Pg?t=270

canab:
@danulqua thanks, I will look into it. It maybe a better solution for us.

pernielsentikaer:
Sounds like an organizational extension might be a better match here. I agree.

PR #8510, opened 2023-09-22
Update: Home Assistant `⌘` + `=` instead of `⌘` + `+`

samuelthng:
## Description

Fixes unusable shortcut with lights and fans by replacing:
<kbd>⌘</kbd> + <kbd>+</kbd> with <kbd>⌘</kbd> + <kbd>=</kbd>.

## Screencast

Not a new extension, just updating shortcut keys on an action.

## Checklist

### DIFF
--- a/extensions/homeassistant/src/components/light.tsx
+++ b/extensions/homeassistant/src/components/light.tsx
@@ -19,11 +19,7 @@ function getBrightnessValues(): number[] {
 
 function hasBrightnessSupport(state: State): boolean {
   const modes = state.attributes.supported_color_modes;
-  if (
-    modes &&
-    Array.isArray(modes) &&
-    (modes.includes("brightness") || modes.includes("color_temp"))
-  ) {
+  if (modes && Array.isArray(modes) && (modes.includes("brightness") || modes.includes("color_temp"))) {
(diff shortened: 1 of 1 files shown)


tonka3000:
@samuelthng Why should `CMD` + `+` not work? I checked it and they work normally.

samuelthng:
Thanks for the reply!

Doesn't seem to work for me on a MacBook en-US keyboard where there is no dedicated <kbd>+</kbd> key.

You'd have to hit <kbd>shift</kbd> to get the + keystroke, and in that case you'd be adding another modifier, which also doesn't trigger the action.

Can you verify if this is the case for you as well @tonka3000?

tonka3000:
Can not(!) confirm that. In my case (de-at) layout `+` is a "direct" key and `=` would require shift.

The main problem here that I would be good to have support for alternative shortcuts in raycast, because not every layout is the same.

samuelthng:
Agree with your opinion. Either that or allow customised shortcut overrides for actions.

In the mean time, since there might not be much demand for this, I'll close the PR and just make the change locally.

Thank you for taking a look at the PR anyway @tonka3000, appreciate it.

PR #8543, opened 2023-09-23
[Mozilla Firefox] Add nightly support as fallback

kud:
Support Firefox Nightly as fallback.

## Description

I only use Firefox Nightly as default browser. This extension crashed. It does not anymore. I did not want to break the previous behaviour so I only use Nightly as fallback and not as first profile. People using both release and nightly will be able to still use release as default.

## Screencast

Private.

## Checklist

### DIFF
--- a/extensions/mozilla-firefox/CHANGELOG.md
+++ b/extensions/mozilla-firefox/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Mozilla Firefox Changelog
 
+## [Add support to Firefox Nightly] - 2023-10-25
+
+- Add support as a fallback for Mozilla Firefox Nightly when only this version exists.
+
 ## [Initial Version] - 2022-12-20
 
 - Open new tabs in Mozilla Firefox
--- a/extensions/mozilla-firefox/src/util/index.ts
+++ b/extensions/mozilla-firefox/src/util/index.ts
@@ -11,7 +11,17 @@ const userDataDirectoryPath = () => {
 
 const getProfileName = (userDirectoryPath: string) => {
   const profiles = fs.readdirSync(userDirectoryPath);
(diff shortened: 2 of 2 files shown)


pernielsentikaer:
What if a user uses both, maybe a preference would be a good thing to add?

kud:
Hahaha, I knew it! One at a time, @pernielsentikaer. We struggle to have multiple accounts/apps in Raycast in general, so yeah, let's say for now that the normal behaviour is to use classic Firefox. Some people, like me, only use Nightly, and it's my choice. Usually, when people use both, it's because they want to try a futuristic version of Firefox for testing. But they could also use classic Firefox for web development testing and use Nightly for regular business.

Oh well, I suggest accepting this PR in the short term to fix the issue because it's a blocker as it causes a crash. And in the future, we can add a feature to choose the version of Firefox people want to use in the settings of this extension.

Does that sound fair?

pernielsentikaer:
Sure that works fine for me, @crisboarna can you also check this

kud:
> Sure that works fine for me, @crisboarna can you also check this

Amazing. Thanks! At some point I could create a PR to be able to choose in the settings what browser you want to use. 😌 I note this in my todo list.

kud:
Hellooooooooooooooooo 👋

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

pernielsentikaer:
Could you add a [CHANGELOG](https://developers.raycast.com/basics/prepare-an-extension-for-store#version-history) entry @kud

kud:
I can definitely do it yeah

kud:
Taaaaaaaaadaaaaaa 🎉

PR #8571, opened 2023-09-26
Add china-ip-address extension

ErlichLiu:
## Checklist


pernielsentikaer (commented):
Hi 👋

How is this compared to https://www.raycast.com/koinzhang/ip-geolocation and https://www.raycast.com/Kang/myip

Don't you see the same data?

ErlichLiu:
Hello,

Firstly, I'd like to clarify that my plugin, "China IP Address", is
distinct from the two you've mentioned. They return different public IP
addresses, so there's a unique value to what my plugin offers. Let me delve
deeper into the reasons:

Due to the unique network environment in China, most Chinese internet users
don't possess a dedicated public IP address. In fact, many of us aren't
even aware of our public IP as we are primarily operating within an
internal network.
However, in certain scenarios, such as when accessing restricted databases
or APIs, it's crucial to identify our public IP. While the two plugins you
mentioned do provide this service, the manner in which my plugin operates
is distinct.
To accurately determine the public IP of a Chinese user, one must utilize
APIs provided by Chinese companies. This ensures you get the actual Chinese
public IP, not one from a proxy server.
The two plugins you referenced tend to return proxy IPs for Chinese users,
which isn't the genuine Chinese IP.
I understand that you might not fully grasp the difference between my
plugin and the others since you haven't lived in China. However, I assure
you, for Chinese users, my plugin holds tangible value and significance.

Thank you for your feedback. I hope that my plugin can serve and benefit
many more users in China.

Per Nielsen Tikær ***@***.***> 于2023年10月10日周二 16:43写道：

pernielsentikaer:
Thanks for the clarification @ErlichLiu 🔥

PR #8683, opened 2023-10-04
feat(deepcast): more input and output options

erics118:
## Description

- add ability to choose where to get text from, as a preferred source, eg get from selected text first, but then fallback to clipboard, or the other way around.
- add ability to paste text when done translating (best when preferred source is selected text)

## Screencast

todo

## Checklist

### DIFF
--- a/extensions/deepcast/package.json
+++ b/extensions/deepcast/package.json
@@ -248,7 +248,7 @@
     {
       "description": "Your DeepL API key",
       "name": "key",
-      "placeholder": "Enter your free DeepL API key",
+      "placeholder": "Enter your DeepL API key",
       "required": true,
       "title": "API key",
       "type": "password"


extensions/deepcast/tsconfig.json
```diff
@@ -1,16 +1,19 @@
...
     "strict": true,
     "isolatedModules": true,
     "esModuleInterop": true,
     "skipLibCheck": true,
     "forceConsistentCasingInFileNames": true,
-    "jsx": "react-jsx"
+    "jsx": "react-jsx",
+    "resolveJsonModule": true,
+    "typeRoots": ["."],
+    "types": ["raycast-env"]
```
erics118, line 17: This was added to take advantage of the autogenerated types in `raycast-env.d.ts`. `ray dev` uses them just fine, but the editor needs to know that it exists by adding these lines.

extensions/deepcast/src/utils.ts
```diff
@@ -6,19 +6,11 @@ import {
...
   LaunchType,
+  closeMainWindow,
 } from "@raycast/api";
 import got from "got";
 import { StatusCodes, getReasonPhrase } from "http-status-codes";
 
-interface Preferences {
-  key: string;
-  onTranslateAction: "clipboard" | "view";
-}
```
erics118, line 16: The addition to `tsconfig.json` makes this interfaace unnecessary.

extensions/deepcast/package.json
```diff
@@ -266,25 +267,49 @@
...
+          "title": "Selected",
+          "value": "selected"
+        },
+        {
+          "title": "Clipboard",
+          "value": "clipboard"
+        }
+      ],
+      "default": "selected",
+      "required": false
```
erics118, line 293: This was copied over from my `change-case` extension

erics118:
I was thinking about removing all the individual commands and instead using quicklinks. This would make the codebase a lot easier, but a migration script probably would need to be written to automatically convert existing enabled commands to quicklinks, and we need a good UX for making these quicklinks

erics118:
CI fails, but idk what to say except that it works on my computer 😅


CleanShot 2023-10-04 at 18 09 56

extensions/deepcast/tsconfig.json
pernielsentikaer: I'm not sure I understand why you need this.

If something is not found, you should be able to run `> ESLint: Restart ESLint Server` in code.

pernielsentikaer:
@erics118 do you mind adding a CHANGELOG entry too?

@mooxl do you mind checking this?

pernielsentikaer:
@mooxl could you check this?

mooxl:
LGTM!

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #8685, opened 2023-10-04
feat: ✨ Added Better Commit extension

BrieucCaillot:
## Description

Be able to create better commits with :prefix & gitmojis.

## Screencast

better-commit-prefixes
better-commit-gitmojis

## Checklist


pernielsentikaer:
Hi there! 👋 

Thank you for your contribution. 🔥 

What is the difference between this and https://www.raycast.com/niemtec/conventional-commits made by @niemtec?

niemtec:
I'd very much welcome some of these improvements if you'd like to contribute to my extension 😊

BrieucCaillot:
The main difference is that Better Commit will append the emoji after the prefix : ex: feat: 🔥
Conventional Commits only allows to add prefixes, not emojis : ex: feat:

pernielsentikaer:
Would it indeed be better for the end user to have only one extension? Otherwise, they end up being confused, as I am. 🙂

Would that be something you, @BrieucCaillot, would like to contribute to the current extension?

whitespacecode:
> Would it indeed be better for the end user to have only one extension? Otherwise, they end up being confused, as I am. 🙂
> 
> Would that be something you, @BrieucCaillot, would like to contribute to the current extension?

There is already a extension for this https://www.raycast.com/krzysztofzuraw/conventional-comments
Maybe the easy list @niemtec created could be a new command ?

pernielsentikaer:
I didn't even remember that one. Thanks for mentioning @whitespacecode. It makes sense to only have one extension. We can always enhance the current to make a better user experience

whitespacecode:
I like the older one because it gives you the possibility to add a title and description. Only downside is that when submitting the form you get something like this: 
image
It should be `fix: This is a test `

PR #8688, opened 2023-10-05
Toothpick: Mark  `bluetoothBackend` preference as optional.

sxn:
## Description

A previous extension update introduced the option to choose between `AppleScript` and `blueutil` as a backend. While there's nothing wrong with the option itself being present, we generally try to avoid adding required preferences to an already-published extension. The reasoning behind that is: it can be jarring when, as a previous user of the extension, I want to run a command, only to seemingly out-of-the-blue (I might not know the extension was updated) be asked to choose between two backends, one of them marked as experimental.
Since AppleScript was already marked as default in the manifest file, this PR updates the manifest so that `bluetoothBackend` is optional instead of required. Hope that's fine! 😺

### DIFF
--- a/extensions/toothpick/CHANGELOG.md
+++ b/extensions/toothpick/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Toothpick Changelog
 
+## [QoL Improvement] - 2023-10-05
+
+- Make bluetooth backend preference an optional preference.
+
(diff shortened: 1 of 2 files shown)


pernielsentikaer (approved):
That makes sense, thanks for the PR

VladCuciureanu:
As Per said, this completely makes sense. Thanks for your contribution!

Issue #8692, opened 2023-10-05
[Visual Studio Code - Project Manager] Url parsing breaks projects that have remote url


joshuata:
### Extension

https://www.raycast.com/MarkusLanger/vscode-project-manager

### Description

If a project points to a code-workspace file, the `parseRemoteURL` function appends an extra slash. Since a saved workspace in project-manager points to the `*.code-workspace` file directly, this causes code to attempt to open a non-existent directory with the error `Workspace does not exist - Please select another workspace to open.` This issue can be avoided by removing the trailing slash "`/`" from `parseRemoteURL`. This also mirrors the behavior of the extension, hopefully avoiding other similar problems in the future

### Steps To Reproduce

(In vscode)
1. Create a remote project
2. Create a `code-workspace` file for the project
3. Open the workspace
4. Save the project in the project manager extension
5. Close vscode
6. Open raycast -> vscode project manager -> the project you just created 

### Current Behaviour

Error `Workspace does not exist - Please select another workspace to open.`

### Expected Behaviour

Opening the project

PR #8693, opened 2023-10-05
Update search-project-manager-projects.tsx

joshuata:
## Description

Fixes issue #8692 by removing the trailing slash

## Checklist

### DIFF
--- a/extensions/visual-studio-code-project-manager/CHANGELOG.md
+++ b/extensions/visual-studio-code-project-manager/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Visual Studio Code - Project Manager Changelog
 
+## [Update] - 2023-10-11
+
+- Fixes opening a remote workspace file (#8692)
+
 ## [Update] - 2023-04-24
 
 - Update build identifier for VSCodium. (VSCodium/vscodium#1227)


pernielsentikaer:
Thanks for your contribution @joshuata 🔥 Do you mind adding a [CHANGELOG](https://developers.raycast.com/basics/prepare-an-extension-for-store#version-history) entry too 

@MarkusLanger can you check this?

joshuata:
@pernielsentikaer Done!

pernielsentikaer:
Then we just need to see what @MarkusLanger says 🙂

MarkusLanger:
> Then we just need to see what @MarkusLanger says 🙂

We added the slash because of this problem => [979292](https://github.com/raycast/extensions/commit/9333270cf29e94da945caa49cf684b21fddc9fb5)

So I don't know if it is a good idea to remove the slash or is the problem with the "." in the path also fixed?

joshuata:
The slash causes an error when the remote project is a workspace file.

For example, if the project is
```json
{
"name": "application",
"rootPath": "vscode-remote://dev-container+7b22...ef/workspace/application.code-workspace",
"paths": [],
"tags": [],
"enabled": true
}
```
Then adding the trailing slash causes it to try and open a directory named `/workspace/application.code-workspace/` which does not exist.

Unfortunately I can no longer compare the functionality of the raycast plugin with the upstream vscode plugin as it lloks like the vscode-project-manager-core repository that contained all of the logic was deleted at some point in August.

pernielsentikaer:
What do you think of that @MarkusLanger

MarkusLanger:
@pernielsentikaer For me it's okay to merge the PR.

joshuata:
I can also write a check for a period in the pathname that caused the earlier issue. Would that be preferred?

pernielsentikaer:
That would be a good idea @joshuata

joshuata:
I am working on this, it just isn't done

joshuata:
@MarkusLanger I have tried to replicate the issue you referenced from project manager but I can't reproduce it. I believe the version of the project manager they were using was bugged, but I cannot verify this since the extension manager has deleted the repos that implement the behavior.

pernielsentikaer:
So what do we do here, just close the PR?

PR #8695, opened 2023-10-05
Update reclaim-ai extension

### LINKED ISSUE #8119: [Reclaim] ...
### Extension

https://www.raycast.com/reclaim-ai/reclaim-ai

### Description

The Raycast integration only takes into account the events and tasks of one calendar. If we connected (without syncing the events) multiple calendars, the events from other calendars are not taken into account. 

### Steps To Reproduce

1. Install Raycast extension
2. Have multiple calendars that Reclaim watches
3. Use the "My Calendar" command to see events, and only have the ones from the main calendar displayed. 

### Current Behaviour

_No response_

### Expected Behaviour

_No response_

shkreios:
## Description

**1. Show All Connected Calendar Events:** Updated the system to display events from all calendars connected via Reclaim.
**2. Duplicate Event Filtering:** Noticed that using the Reclaim calendar sync across multiple calendars sometimes results in duplicate events. Implemented a simple filter logic to weed out these duplicates and ensure a cleaner event display.

Fixes #8119 

## Checklist

### DIFF
--- a/extensions/reclaim-ai/CHANGELOG.md
+++ b/extensions/reclaim-ai/CHANGELOG.md
@@ -1,5 +1,10 @@
 # reclaim Changelog
 
+## [Update] - 2023-10-11
+
+- Show All Connected Calendar Events: Updated the system to display events from all calendars connected via Reclaim.
+- Duplicate Event Filtering: Noticed that using the Reclaim calendar sync across multiple calendars sometimes results in duplicate events. Implemented a simple filter logic to weed out these duplicates and ensure a cleaner event display.
+
 ## [Update] - 2023-09-02
 
 **My Calendar**


shkreios:
With this change the ui & commands data deviates some times since the `join meeting` & `calendar menu bar` command both use the `/moment/next` endpoint which seems to not support the `calendarids` parameter. If its not possible to update the endpoint, it would be an option to do the filter in js from the `events` endpoint. 

For example something like this
```ts
const now = new Date();
const events = eventData
  ?.filter((event) => {
    return showDeclinedEvents ? true : event.rsvpStatus !== "Declined" && event.rsvpStatus !== "NotResponded";
  })
  .filter((event) => {
    return event.reclaimEventType !== "CONF_BUFFER" && event.reclaimEventType !== "TRAVEL_BUFFER";
  })
  .filter((event) => isAfter(new Date(event.eventEnd), now))
  .filter((event) => {
    return !(differenceInHours(new Date(event.eventEnd), new Date(event.eventStart)) >= 24);
  });

return {
  event: events?.at(0),
  nextEvent: events?.at(1),
}
```

maximilianzuern:
Thank you for your contribution! 🥳🙏
I've reviewed your code and it looks good to me!

webergnr (approved):

shkreios:
As mentioned in this [comment](https://github.com/raycast/extensions/pull/8695#issuecomment-1749333404) if we merged this pr the ui deviates in some places where the `/moment/next` endpoint is used, i opened another pr with fix for that, but i don't know if my logic is correct, maybe you can have a look at #8714

pernielsentikaer (approved):

ajohnson1200:
@shkreios for this part:

> how All Connected Calendar Events: Updated the system to display events from all calendars connected via Reclaim.

First, thanks for making this change, super helpful!

Second, we've gotten a couple of support tickets from folks over the last couple weeks after this change where people aren't seeing anything in the menu bar, it just says "No upcoming events" until they click into the menu bar... and then it updates. I'm pretty sure that there's some kind of interaction between the loading of accounts and the loading of events... but...

Third, there's an easier way of getting events from all connected accounts.  The Events API supports a URL param called allConnected, which is false by default, but when set to true, fetches events from the primary and all connected accounts.  I could do this work, but didn't want to overwrite (again, apologies if I stepped on your toes on the previous update) what you here. 

Would you want to do incorporate this allConnected param and then refactor out the fetching of accounts?

Thanks for your contributions on this!

AJ

shkreios:
Hey @ajohnson1200
> Third, there's an easier way of getting events from all connected accounts. The Events API supports a URL param called allConnected, which is false by default, but when set to true, fetches events from the primary and all connected accounts. I could do this work, but didn't want to overwrite (again, apologies if I stepped on your toes on the previous update) what you here.

Indeed simplifies fetching events from all connected accounts. I initially followed the method used by the Reclaim website, where I observed that events were fetched by passing the account IDs. 

Speaking of which, is there an OpenAPI/Swagger documentation available for the Reclaim API? Such a resource would be invaluable for identifying features like these.

> Would you want to do incorporate this allConnected param and then refactor out the fetching of accounts?

I'm inclined to proceed with it. However, a potential concern is the duplication of events, especially if the connected calendars sync with each other. Do you have any insights or recommendations on what kind of filtering logic should be applied to such duplicates?

shkreios:
> Second, we've gotten a couple of support tickets from folks over the last couple weeks after this change where people aren't seeing anything in the menu bar,

@ajohnson1200, The issue originates from [this line](https://github.com/raycast/extensions/blob/main/extensions/reclaim-ai/src/notifications.tsx#L95). I was under the impression that setting `keepPreviousData` to `true` would cache both the accounts and events endpoints, ensuring the menu bar always displayed content. However, I've realized that events are only fetched if accounts are available. Additionally, when Raycast triggers the menubar command in the background, it seems to go through just one React render cycle, therefore never calling the events endpoint.

ajohnson1200:
> Speaking of which, is there an OpenAPI/Swagger documentation available for the Reclaim API? Such a resource would be invaluable for identifying features like these.

Apologies that we don't have that published yet, we've talked about it internally but haven't pulled the trigger yet.  

> However, a potential concern is the duplication of events, especially if the connected calendars sync with each other. Do you have any insights or recommendations on what kind of filtering logic should be applied to such duplicates?

There's some logic that exists today where if a connected calendar is involved in a sync, it becomes an implicitly connected calendar and as a result, the event (or both of them, because of the sync) gets de-duplicated by Reclaim properly.  I'm 99.9% sure that allConnected=true and sourceDetails=true (which is already there) will handle this appropriately, but give it a try and let's chat if you don't like what you're seeing.

ajohnson1200:
> The issue originates from this line.
sounds good, seems like the allConnected=true negates the need for fetching the accounts separately but either way, definitely want the end user experience to be that 100% of the time Raycast is showing an up to the minute view of what's on their calendar.

shkreios:
@ajohnson1200 

> There's some logic that exists today where if a connected calendar is involved in a sync, it becomes an implicitly connected calendar and as a result, the event (or both of them, because of the sync) gets de-duplicated by Reclaim properly. I'm 99.9% sure that allConnected=true and sourceDetails=true (which is already there) will handle this appropriately, but give it a try and let's chat if you don't like what you're seeing.

My current setup includes 2 calendars syncs

- university calendar (external & imorted into google via link) --> sync to work calendar with Event Details setting
- personal calendar (main account) --> sync to work calendar with Personal commitment setting

For the first sync the events endpoint correctly dedupes the events but for the second sync all events are duplicated, which is especially polluting the menu bar view for when travel time is active.

> sounds good, seems like the allConnected=true negates the need for fetching the accounts separately but either way, definitely want the end user experience to be that 100% of the time Raycast is showing an up to the minute view of what's on their calendar

do you want me to open pr without any filtering just for this hotfix?

ajohnson1200:
> do you want me to open pr without any filtering just for this hotfix?

yeah, that'd be great to start and then let's tackle the duplicates.

shkreios:
@ajohnson1200 Done https://github.com/raycast/extensions/pull/9037

PR #8705, opened 2023-10-06
[word-search] Capitalize output when search text is capitalize

bkeys818:
## Description

When the search text is capitalized, the copied/pasted word will also be capitalized.

## Checklist

### DIFF
--- a/extensions/word-search/package.json
+++ b/extensions/word-search/package.json
@@ -25,6 +25,15 @@
         {"title": "Paste Word in Active App", "value": "paste"},
         {"title": "Copy Word to Clipboard", "value": "copy"}
       ]
+    },
+    {
+      "name": "capitalizeResults",
+      "description": "Whether or not a capitalized search will cause the results to be capitalized.",
+      "required": false,
+      "title": "Search Results",
+      "label": "Capitalize results when search is capitalized.",
+      "type": "checkbox",
+      "default": false
     }
   ],
   "commands": [
--- a/extensions/word-search/src/api.ts
+++ b/extensions/word-search/src/api.ts
@@ -1,6 +1,6 @@
 import fetch from "node-fetch";
-import { SearchType, Word } from "./types";
-import { showToast, Toast } from "@raycast/api";
+import { SearchType, Word, Preferences } from "./types";
+import { showToast, Toast, getPreferenceValues } from "@raycast/api";
 import { URL, URLSearchParams } from "url";
 
 export async function searchWords(wordToSearch: string, type: SearchType): Promise<Word[]> {
@@ -10,6 +10,8 @@ export async function searchWords(wordToSearch: string, type: SearchType): Promi
     max: "50",
     [type]: wordToSearch,
   });
+
+  const { capitalizeResults } = getPreferenceValues<Preferences>();
 
   const url = new URL(`/words?${searchParams}`, "https://api.datamuse.com/words").toString();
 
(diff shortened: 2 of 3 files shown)


pernielsentikaer:
Hi @bkeys818 👋 

Wouldn't it make sense to uppercase in the list too, now that it's done in the actions?

@RishabSwift do you mind checking this

bkeys818:
@pernielsentikaer, Can't hurt

pernielsentikaer:
@RishabSwift  do you have time to check?

RishabSwift:
Thanks for the feature! Will check it out in a few hours

RishabSwift:
@bkeys818 I just tried it out.

Doesn't seem to work?

Screenshot 2023-10-15 at 18 04 31@2x

It is not capitalizing the list nor the actions for me (like when I copy or paste in active app).

extensions/word-search/src/search-results.tsx
```diff
@@ -35,26 +37,26 @@ export default function SearchResults(
...
-          <List.Item
-            icon={"command-icon.png"}
-            key={word.word}
-            title={word.word}
-            subtitle={word.defs !== undefined ? word.defs[0] : ""}
-            actions={<Actions {...{ word, search }} defaultAction={defaultAction} />}
-          />
-        ))
+        words.map((word) => {
+          const wordString = searchIsCapitalized ? word.word[0] + word.word.slice(1) : word.word;
```
RishabSwift, line 41: This logic is not capitalizing the word.

Try this instead:

    const wordString = searchIsCapitalized ? word.word.charAt(0).toUpperCase() + word.word.slice(1) : word.word;

RishabSwift (changes requested):

extensions/word-search/src/search-results.tsx
RishabSwift:
```suggestion
          const wordString = searchIsCapitalized ? word.word.charAt(0).toUpperCase() + word.word.slice(1) : word.word;
```

frntline-dev:
@bkeys818, I think it's preferable to capitalize the words as they are extracted from the request response in the API file rather than in the search results. This approach keeps everything consolidated in one location and avoids the need to manipulate the render method. 

What do you think?

bkeys818:
@frntline-dev, I'd probably help to explain why I suggested this change.

I use to use [Alfred 5](https://www.alfredapp.com), which had "[spell](https://www.alfredapp.com/help/features/dictionary/#spell)", a similar functionally to this extension. One of my favorite features was that it would match the case of the original word.

When I use this extension, I'll copy a word I'm unsure is spelled right and call the extension. Then, I select the correct word which replaces the original. When the misspelled is the first word in a sentence (aka. capitalized), the corrected spelling always appears all lowercase. That means I have to go back and re-capitalize it every time.

I acknowledge that it's a minor annoyance, but I think the fix is justified since it's insignificant. That's just my opinion. I'd love to hear what other people think.

pernielsentikaer:
We could also make it a preference if it should happen or not and then lease default behavior as today and make a toggle for it 🙂

RishabSwift:
@bkeys818 I think there's a better way to achieve the capitalization of the first letter.

For example, in the `api.ts`, you could simply do the following at the last block of code that parses the definition of each word search result (`line 28`)

```
words.forEach((word) => {
   ...

    // if wordToSearch is capitalized, capitalize the first letter of the word
    if (wordToSearch[0] === wordToSearch[0].toUpperCase()) {
        word.word = word.word[0].toUpperCase() + word.word.slice(1);
    }

   ...
});
```

This centralizes the logic, ensuring the `tsx` file remains clean and focused on rendering.

I think this is a better way to go about it, don't you think?

bkeys818:
@RishabSwift, that is an excellent idea. I'll get it done right now.

pernielsentikaer:
Can you check it again @RishabSwift

RishabSwift:
Looks great @bkeys818! Thanks for taking the time 💯

RishabSwift:
> Can you check it again @RishabSwift

All looks well!

pernielsentikaer:
Awesome, thanks 🔥

PR #8763, opened 2023-10-11
GitHub: Add "Clone and Open" action

daquinoaldo:
## Description

Add a `Clone and Open` command in the `Search Repositories` results, with shortcut `⌘⌥C`.
The command clones the repository in a configurable path (by default, `~/GitHub`). Then, it opens the cloned folder using a configurable IDE (by default, VS Code).
The clone is skipped if the repository already exists. When cloning, it uses the `git` binary (pre-installed on MacOS) and HTTPS with the extension's OAuth token to avoid any user configuration requirement.

## Checklist

### DIFF
--- a/extensions/github/src/generated/graphql.ts
+++ b/extensions/github/src/generated/graphql.ts
@@ -21351,6 +21351,8 @@ export enum RepositoryLockReason {
   Rename = "RENAME",
   /** The repository is locked due to a trade controls related reason. */
   TradeRestriction = "TRADE_RESTRICTION",
+  /** The repository is locked due to an ownership transfer. */
+  TransferringOwnership = "TRANSFERRING_OWNERSHIP",
 }
 
 /** A GitHub Enterprise Importer (GEI) repository migration. */
@@ -21656,6 +21658,8 @@ export enum RepositoryRuleType {
   TagNamePattern = "TAG_NAME_PATTERN",
   /** Only allow users with bypass permission to update matching refs. */
   Update = "UPDATE",
+  /** Require all changes made to a targeted branch to pass the specified workflows before they can be merged. */
+  Workflows = "WORKFLOWS",
   /** Workflow files cannot be modified. */
   WorkflowUpdates = "WORKFLOW_UPDATES",
 }
@@ -22466,7 +22470,8 @@ export type RuleParameters =
   | RequiredDeploymentsParameters
   | RequiredStatusChecksParameters
   | TagNamePatternParameters
-  | UpdateParameters;
+  | UpdateParameters
+  | WorkflowsParameters;
 
 /** Specifies the parameters for a `RepositoryRule` object. Only one of the fields should be specified. */
 export type RuleParametersInput = {
@@ -22488,6 +22493,8 @@ export type RuleParametersInput = {
   tagNamePattern?: InputMaybe<TagNamePatternParametersInput>;
   /** Parameters used for the `update` rule type */
   update?: InputMaybe<UpdateParametersInput>;
+  /** Parameters used for the `workflows` rule type */
+  workflows?: InputMaybe<WorkflowsParametersInput>;
 };
 
 /** Types which can have `RepositoryRule` objects. */
@@ -23039,6 +23046,8 @@ export enum SocialAccountProvider {
   Linkedin = "LINKEDIN",
   /** Open-source federated microblogging service. */
   Mastodon = "MASTODON",
+  /** JavaScript package registry. */
+  Npm = "NPM",
   /** Social news aggregation and discussion website. */
   Reddit = "REDDIT",
   /** Live-streaming service. */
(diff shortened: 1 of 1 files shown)


extensions/github/src/helpers/repository.ts
```diff
@@ -34,15 +34,61 @@ export const WEB_IDES = [
...
+
+  const toast = await showToast({
+    title: `Opening ${repository.nameWithOwner}`,
+    message: `at ${clonePath}`,
+    style: Toast.Style.Animated,
+  });
+
+  if (!existsSync(clonePath.replace("~", homedir()))) {
+    const gitHubToken = getGitHubToken();
+    const cloneUrl = `https://oauth:${gitHubToken}@github.com/${repository.nameWithOwner}`;
```
daquinoaldo, line 67: I tried many commands, but in the end, I found this the most robust.

- SSH should be preferred but requires the user to set it up, which may lead to unpleasant failures. Also, when the user uses [1Password CLI](https://developer.1password.com/docs/ssh/) or other password manager integrations to protect their SSH keys, the SSH clone won't work because the user doesn't get the authentication prompt, and the command fails as unauthenticated.
- The `gh` CLI is an alternative method but forces the user to install the CLI. We could install it for them, but I don't like forcing such dependency when it's not needed and doesn't provide much additional value. In any case, it also requires more complex error-handling logic, making this command unnecessarily cumbersome. Finally, if the user protects the `gh` CLI token with password managers (e.g., [1Password](https://developer.1password.com/docs/cli/shell-plugins/github/)), it won't work for the same reason as SSH.

We already store and use a GitHub token, either user-generated or from OAuth, and it already has the permissions required for the clone, so it's pointless to ask the user for further authentications.

extensions/github/src/components/RepositoryActions.tsx
```diff
@@ -107,6 +107,13 @@ export default function RepositoryActions({ repository, mutateList, onVisit }: R
...
+        <Action
+          icon={Icon.Terminal}
+          title="Clone and Open"
+          onAction={() => cloneAndOpen(repository)}
+          shortcut={{ modifiers: ["cmd", "opt"], key: "c" }}
+        />
+
         <Action.OpenInBrowser
           icon="vscode.svg"
           title="Clone in VSCode"
```
daquinoaldo, line 119: I know there was already this command available. However, I think it's very limited and nowhere near as powerful as the new one.
It allows you to open VS Code with the clone prompt but doesn't let you open an already-cloned repository from the search results. That's why I added the Clone and Open: you clone once and open as many times as you want.
In my opinion, we could deprecate the Clone in VSCode command. However, it would be a breaking change, and I don't know what's the best practice in this case. Personally, I don't have any preference in this regard, and we could leave it there.

extensions/github/package.json
```diff
@@ -193,6 +194,22 @@
...
+      "placeholder": "/Applications/Visual Studio Code.app"
+    },
+    {
+      "name": "baseClonePath",
+      "type": "directory",
+      "required": false,
+      "title": "Clone Path",
+      "description": "Path in which the repositories will be cloned.",
+      "placeholder": "~/GitHub"
+    },
```
thomaslombart, line 212: You can use the `default` property over here instead.

extensions/github/src/components/RepositoryActions.tsx
thomaslombart: I think it's fine having them both. The VS Code action always prompt you to select the repository destination so it's slightly different.

extensions/github/package.json
```diff
@@ -193,6 +194,22 @@
     }
   ],
   "preferences": [
+    {
```
thomaslombart, line 197: Could you move both preferences in the `Search Repositories` command preferences? That way it doesn't pollute the global preferences for these not using it.

thomaslombart (commented):
Great work!

daquinoaldo:
Thanks for the review, @thomaslombart! It's my first contribution, so I'm not yet familiar with the SDK. 😄 I should have addressed all the comments, can you have a second look?

extensions/github/src/helpers/repository.ts
```diff
@@ -34,15 +34,59 @@ export const WEB_IDES = [
...
+    message: `at ${clonePath}`,
+    style: Toast.Style.Animated,
+  });
+
+  if (!existsSync(clonePath.replace("~", homedir()))) {
+    const cloneUrl = `https://github.com/${repository.nameWithOwner}`;
+    const cloneCommand = `git clone ${cloneUrl} ${clonePath}`;
+
+    try {
+      execSync(cloneCommand);
```
thomaslombart, line 69: Missed that one, but you probably want to show a success toast when it's done. Otherwise, you'll get the animated toast infinitely.

daquinoaldo: Updated, thanks.

thomaslombart (approved):

thomaslombart:
Great! 🚀

PR #8783, opened 2023-10-12
Add installed-extensions extension

pernielsentikaer:
## Checklist


mil3na (changes requested):
Hey Per! 

I'm reviewing your extension here. I have a few notes:

- When I first load it and don't have jq installed, after the error happens and the note on the bottom disappears, I cannot see why it doesn't work. Do you think it would be worth putting a note on the empty state like "it's empty or you don't have jq installed"?
- Would it be possible to open the settings when selecting an extension?
- Maybe for a future update: sorting!

Other than that, it looks good to me!

pernielsentikaer:
> Do you think it would be worth putting a note on the empty state like "it's empty or you don't have jq installed"?

Good idea! I'll implement that!

> Would it be possible to open the settings when selecting an extension?

That's not possible (yet) in the API, but would indeed be cool, it would make sense if it opened extensions preferences for sure 🙂

> Maybe for a future update: sorting!

What kind of sorting do you want to have?

mil3na:
Sorting alphabetically and by installed/update date would be a nice start!
By usage would be awesome, but not sure if it is possible.

pernielsentikaer:
This version only sorts alphabetically, I have now added an empty view 🙂

mil3na (approved):
LGTM!

PR #8798, opened 2023-10-13
feat(bookmarks): open the browser where the bookmark came from

iloveitaly:
- fix: frecency => frequency
- build: major package bump
- feat(bookmarks): open the browser where the bookmark came from

## Description
Open the browser where the bookmark was sourced from.

In my case, and I imagine most users, different browsers have different login state. For instance, one work, one personal.

If you are trying to open a personal google doc in Chrome when the account is logged in on Safari, the bookmarks plugin is useless.

This fixes that issue.

## Screencast
NA

## Checklist

### DIFF
--- a/extensions/browser-bookmarks/src/hooks/useChromiumBookmarks.ts
+++ b/extensions/browser-bookmarks/src/hooks/useChromiumBookmarks.ts
@@ -77,6 +77,7 @@ async function getChromiumProfiles(path: string) {
   const file = await read(`${path}/Local State`, "utf-8");
   const localState = JSON.parse(file);
 
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
   const profileInfoCache: Record<string, any> = localState.profile.info_cache;
 
   const profiles = Object.entries(profileInfoCache)


iloveitaly:
I'm confused what is going on with the typescript errors—any ideas?

extensions/browser-bookmarks/src/index.tsx
```diff
@@ -63,12 +63,13 @@ export default function Command() {
         return [browsers[0].bundleId as string];
       }
 
+      // TODO I don't understand why pulling the default is useful here
```
thomaslombart, line 66: We pull the default browser to enable it until the user manually selects the browsers from which they want to pull the bookmarks. If we remove that, then the user would see an empty state and would have to select browsers the very first time. To be honest, that would be okay from a UX standpoint, but having it already pre-selected is better since the user doesn't have to think about it.

thomaslombart (commented):
Thanks for the PR! Regarding the TS errors, it's probably related to the TypeScript upgrade. I remembered seeing that on another PR too, downgrade to 5.1.6 and you should be fine.

Frecency can seem like a typo, but it isn't. It's a measure that combines frequency and recency: https://wiki.mozilla.org/User:Jesse/NewFrecency.

iloveitaly:
Thanks for the comments! Reverted the frecency change, dropped ts version.

erics118:
The CI errors were partially a Raycast API issue as well. Updating to latest `@raycast/api` fixes the issues. You can take a look at #8340 for an example

iloveitaly:
@erics118 fixed!

PR #8801, opened 2023-10-14
feat: upload rust-documentation extension

waynexia:
## Description

Search rust crates' document easily

## Screencast

screenshot 2023-02-08 19 00 47
screenshot 2023-02-08 19 01 57

## Checklist

### DIFF
--- a/extensions/rust-documentation/src/index.tsx
+++ b/extensions/rust-documentation/src/index.tsx
@@ -1,6 +1,6 @@
 import { ActionPanel, List, Action } from "@raycast/api";
 import got from "got";
-import { useState, React, useEffect } from "react";
+import React, { useState, useEffect } from "react";
 import { parse } from "node-html-parser";
 
 const INPUT_DELAY_MS = 500;


extensions/rust-documentation/src/index.tsx
```diff
@@ -0,0 +1,195 @@
+import { ActionPanel, List, Action } from "@raycast/api";
+import got from "got";
+import { useState, React, useEffect } from "react";
```
waynexia, line 3:
```suggestion
import React, { useState, useEffect } from "react";
```

pernielsentikaer:
Hi 👋 

What is the difference between this and [this one](https://www.raycast.com/matt/cratecast) 🙂

pernielsentikaer:
What do you think here @BalliAsghar, would it make sense to supersede yours with this if it has more features?

BalliAsghar:
> What do you think here @BalliAsghar, would it make sense to supersede yours with this if it has more features?

I like the symbol search feature, but I disagree with its current implementation. I believe it's more logical for users to view all available symbols. Regarding the extension's performance, it seems slower compared to [cratecast](https://www.raycast.com/matt/cratecast) and has some quirks. Adding the symbol search feature to [cratecast](https://www.raycast.com/matt/cratecast) with a more refined implementation makes sense to me.

BalliAsghar:
i was able to make implament it [cratecast](https://www.raycast.com/matt/cratecast). 
@waynexia what do you think about this implementation. take a look #8871 

Metadata

waynexia:
Greeting @BalliAsghar 

>would it make sense to supersede yours with this if it has more features?

First of all, I don't see any reason to "supersede" one over another. Our difference is told in the naming, crates v.s. document. It doesn't hurt to have both, right?

>I like the symbol search feature

I love it, too. And I've been using it for over half a year [from Feb](https://github.com/waynexia/raycast-rust-document/commits/main). During that time, there isn't an extension that could search this for me. 

>I believe it's more logical for users to view all available symbols.

You are correct, so I also provide this functionality. If you have installed it, try `tokio#` and here is my result.

image

>i was able to make implament it

I apologize in advance, but, to be honest, I don't like this behavior. Of course you are able to. And you implement it way more faster than me. It took me quite a while to find this workaround to fetch the symbol list. If this is what raycast community encourages, I don't want to waste any time arguing this.

BalliAsghar:
> First of all, I don't see any reason to "supersede" one over another. Our difference is told in the naming, crates v.s. document. It doesn't hurt to have both, right?

I believe it makes more sense to consolidate these into a single extension. The differences between them are minor, and it would be more user-friendly to have everything in one comprehensive extension.

> I love it, too. And I've been using it for over half a year [from Feb](https://github.com/waynexia/raycast-rust-document/commits/main). During that time, there isn't an extension that could search this for me.

It's a great approach to check the Raycast store for existing extensions before creating a new one. I've had a similar experience where I developed an extension, only to discover that there were already similar ones available. It can save you time and effort in the long run. 

Additionally, if there are similar extensions, consider building on top of them rather than creating a separate one. This way, users won't have to jump around between multiple extensions, providing a more streamlined experience.

> I apologize in advance, but, to be honest, I don't like this behavior. Of course you are able to. And you implement it way more faster than me. It took me quite a while to find this workaround to fetch the symbol list. If this is what raycast community encourages, I don't want to waste any time arguing this.

I understand your concerns, and I appreciate the effort you put into it. However, it's essential to maintain a positive and collaborative atmosphere in the Raycast community. t's always good to keep an open dialogue and share your perspective with the community. It can lead to a more constructive and inclusive development environment.

waynexia:
>However, it's essential to maintain a positive and collaborative atmosphere in the Raycast community.

By positive and collaborative, do you mean just put the logic from other's code into yours? I don't think so. You could ask me to contribute, or at least ask before you do this. But I can't blame you as mine is unlicensed after all.

>It's a great approach to check the Raycast store for existing extensions before creating a new one.

I've searched this before I started. I believe back to the time I wrote, there was no such extension, right?


>I believe it makes more sense to consolidate these into a single extension.

I don't think this matters. Or why doesn't Raycast join Alfred, why doesn't MySQL join PG. How can you submit a rust extension (with a few functionalities) and then everything comes after should be "consolidated" into that one, or "get consolidated" into that one?

BalliAsghar:
> By positive and collaborative, do you mean just put the logic from other's code into yours? I don't think so. You could ask me to contribute, or at least ask before you do this. But I can't blame you as mine is unlicensed after all.

I want to clarify that I didn't review your code before implementing mine. If you compare my code with yours, you'll notice that the implementation logic is quite different. My intention was to provide a demonstration of how it could be easily implemented. In the spirit of collaboration, I didn't intend to take away from your work.

> I've searched this before I started. I believe back to the time I wrote, there was no such extension, right?

cratecast is nearly 2 years old extension. first available in October 2021.

> I don't think this matters. Or why doesn't Raycast join Alfred, why doesn't MySQL join PG. How can you submit a rust extension (with a few functionalities) and then everything comes after should be "consolidated" into that one, or "get consolidated" into that one?

I'm not going to comment on that!

BalliAsghar:
@pernielsentikaer I've shared my perspective, and now I'll leave the rest to you.

waynexia:
>I want to clarify that I didn't review your code before implementing mine.

I apologize 🙇‍♂️ Should not say that.

Then we are implementing the same thing independently. Let's see what @pernielsentikaer says.

pernielsentikaer:
It always makes sense from a user's perspective to have only one extension with a purpose, in order to avoid confusing the user. In this case, they would be quite similar.

We always ask the original author of the matching extension for their opinion. They can either supersede or incorporate the implemented features or add them themselves. In all cases, we include the author of the new extension as a contributor to the original extension.

I hope the above makes sense, feel free to contact me here or at [Slack](https://www.raycast.com/community) if you have any questions.

waynexia:
I agree. Thank you both. I'm going to close this thread.

waynexia:
Thanks for inviting me. But not this time as I didn't actually contribute to cratecast. Wish I could become one in the future 🤗

PR #8820, opened 2023-10-16
[Val Town] Add val town

KevinBatdorf:
## Description

Run, view, and search Vals from https://www.val.town/

## Screencast

https://github.com/raycast/extensions/assets/1478421/51f601be-1f14-42e9-8f76-78e022b71140

## Checklist


KevinBatdorf:
Some upcoming api changes required I disable the "Recent Runs" you see in the screencast. I'll re-add them after the api is updated and public.

pernielsentikaer (changes requested):
Hi @KevinBatdorf 👋 

Thanks for an awesome extension! 🔥 

Would it make sense to hide the items in the list when there is an error?

Raycast-Ml1kKaOJ@2x

KevinBatdorf:
Ah yeah right. When I had the "recent runs" stuff there it would push an EmptyView if it would error like that. I'll clean that up thanks

KevinBatdorf:
I re-implimented the "recent runs" feature I mentioned removing above and confirmed the empty view shows.

CleanShot 2023-10-18 at 12 37 31

Users can still search other's content, but it's not really the primary use for the extension (and arguably not very useful without an account anyway) so I think it's still better to make the api token required even though some things still work.

pernielsentikaer:
Thanks for the update 🔥 

It seems like both `isLoading` and `isSearching` is always false so it makes a weird flicker 🙂

https://github.com/raycast/extensions/assets/38033107/f1c09c40-3e53-4574-8cb0-7030f4ffd3d3

KevinBatdorf:
I pushed an update now if you can check it again. Thanks. 

I guess it's because `isLoading` is false if the previous data is already there, and it takes time for the error to come in. Makes sense. I updated it to check whether profile is undefined instead.

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #8905, opened 2023-10-21
[Git Repos] Bug fix for #8887

### LINKED ISSUE #8887: [Git Repos] Search does not work anymore since last update
### Extension

https://raycast.com/moored/git-repos

#### Error:
```
Error: Invalid binary plist. Expected 'bplist' at offset 0.
    at a1.parseBuffer (/Users/tms/.config/raycast/extensions/e0a4c081-81fa-44ca-a689-08e4920ee819/list.js:30:23308)
    at qa (/Users/tms/.config/raycast/extensions/e0a4c081-81fa-44ca-a689-08e4920ee819/list.js:39:2630)
    at async _2 (/Users/tms/.config/raycast/extensions/e0a4c081-81fa-44ca-a689-08e4920ee819/list.js:46:3981)
    at async u8 (/Users/tms/.config/raycast/extensions/e0a4c081-81fa-44ca-a689-08e4920ee819/list.js:49:396)
    at async Object.gitRepos [as current] (/Users/tms/.config/raycast/extensions/e0a4c081-81fa-44ca-a689-08e4920ee819/list.js:46:4648)
```

### Steps To Reproduce

1. select 'list git repos'
2. start search ... error appears

does not matter if i search for names or paths (new config item)

### Current Behaviour

_No response_

### Expected Behaviour

_No response_

nitishpdmn:
## Description

Resolves #8887: getting default browser throws an exception when the plist is not in binary format.

The default browser bundleId is used when opening an app to decide whether to open the path to the file, or the remote url. The proposed fix is to just catch and ignore the error when fetching bundleId, and instead also maintain a list of all installed common browsers (list maintained by [get-installed-browsers](https://www.npmjs.com/package/get-installed-browsers) package). If the default browser bundleId is unset (due to the binary plist exception), then for most browsers, the app name will be matched against the installed browsers list, allowing proper remote url opening.

Additionally, the "Open in All Applications" action wasn't using the same target logic, so it would always open the file path, even for browsers. This is now fixed.

#### Considerations

Checking against the list of browsers will happen for every single app (not just browsers) in the user's list of preferences, so there is theoretically a minor performance hit. However there's a maximum of 5 apps and unlikely to be more than say 5 browsers installed for most people, so this shouldn't be a huge issue.

The only advantage I see, besides the browser opening option continuing to work when the plist parsing fails, is that the user would be able to set a non-default browser to open their repos (unlikely scenario).

## Screencast

Reproducing what I think is the bug on the published build:
Screenshot 2023-10-21 at 11 50 38 AM

And the same thing with the new patch (nothing exciting):
Screenshot 2023-10-21 at 11 50 47 AM

## Checklist

### DIFF
--- a/extensions/git-repos/CHANGELOG.md
+++ b/extensions/git-repos/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Git Repos Changelog
 
+## [Bug Fix] - 2023-10-21
+
+- Resolved issue with binary plist for reading default browser. Now checks against all known browser paths as a fallback.
+- Fixed "Open in All Applications" opening browser not as a url
+
 ## [Enhancement] - 2023-10-18
 
 - Added option to filter search with the entire repo path.
--- a/extensions/git-repos/package-lock.json
+++ b/extensions/git-repos/package-lock.json
@@ -10,6 +10,7 @@
         "@raycast/api": "^1.59.0",
         "@raycast/utils": "^1.5.2",
         "default-browser": "^4.0.0",
+        "get-installed-browsers": "^0.1.7",
         "parse-git-config": "^3.0.0",
         "parse-github-url": "^1.0.2"
       },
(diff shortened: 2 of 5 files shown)


pernielsentikaer:
Thanks for your contribution @nitishpdmn 🔥 

@moored do you mind checking this?

moored:
Looks good

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #8967, opened 2023-10-25
Update github extension

aeither:
## Description

Added My Latest Repositories Command to show my latest Github respositories in descending order.

## Screencast

CleanShot 2023-10-25 at 23 33 53@2x

## Checklist

### DIFF
--- a/extensions/github/CHANGELOG.md
+++ b/extensions/github/CHANGELOG.md
@@ -1,5 +1,9 @@
 # GitHub Changelog
 
+## [Add My Latest Repositories command] - 2023-11-06
+
+- Add a command to show my latest repositories in descending order.
+
 ## [Add Clone and Open command] - 2023-10-11
 
 - Add a command to clone (unless already cloned) and open in the preferred code editor a repository from the seacrh repository results.
--- a/extensions/github/src/search-repositories.tsx
+++ b/extensions/github/src/search-repositories.tsx
@@ -1,6 +1,6 @@
 import { List, getPreferenceValues } from "@raycast/api";
 import { useCachedPromise } from "@raycast/utils";
-import { useState, useMemo } from "react";
+import { useMemo, useState } from "react";
 
 import { getBoundedPreferenceNumber } from "./components/Menu";
 import RepositoryListEmptyView from "./components/RepositoryListEmptyView";


aeither:
> @aeither : Some considerations:
> 
> I tried the code, the current repository what is displayed in the list on raycast does not correspond to what I see by going to my profile on repositories.
> 
> Sorting should not be based on the last update which could be a commit as well as a description change.
> 
> I also tried editing on myLatestRepositories using `PUSHED_AT` and `CREATED_AT`, changing `DESC` to `ASC`, but the result never changes, it just stays the same.
> 
> Some dates of the last update on some repositories do not seem to match, between those on raycast and those displayed on github repositories.

Thanks for noticing the difference between the Github GraphQL query and github profile repositories.
So I just tried with several more queries 

As you can find under profile repositories filter. It is sort by last updated by default.

CleanShot 2023-10-26 at 18 42 13@2x

That's why I tried with updatedAt. Actually my first try was with search.

```
query {
  search(
    type:REPOSITORY,
    query:"sort:updated owner:YOUR_USERNAME",
    first: 10
  ) {
    nodes{
        ... on Repository {
          name,
					# ...THE REST OF THE VALUES
        }
    },
  }
}
```

But it is a long query, I wanted something where I can just hit enter to see my latest repositories. I have noticed that under profile repositories it include repos like those that are updated not by you. for example: updated by dependabot and forked repositories also as you mentioned there is mismatch in date.

Regarding the dates, In Raycast the date comes from repository `updatedAt`.

After few more attempts. To reflect better raycast my latest repos with profile repositories from github.com I changed to PUSHED_AT and also updated the the date to pushedAt. In this way, it also shows repositories where is updated by dependabot and forks which weren't showing with UPDATED_AT. Also the dates should be also the same now that it comes from pushedAt.

aeither:
@Angelk90 updated

aeither:
I am on raycast slack but I don't open it often as I don't use Slack.

thomaslombart:
@aeither I get the need for this command but there's already a command called `Search Repositories` to find repositories. There you can select `My Repositories` in the dropdown to see your repositories. It may makes sense to add the "latest" option there.

aeither:
@thomaslombart Hey Thomas.

First, It serves a different purpose. with search repositories is for when I already know what I am looking which is why there is a search bar where I can filter and look for the term I already have in mind. With latest, I don't know what is the name but I know I commit to it recently so I can find the repo with the command without any term in mind.

Putting my latest inside search make it less intuitive for new users as the name search repositories doesn't indicate immediately it can also list your latest repositories. 

I also use search repositories. Putting my latest repos inside make me have to move back and forth in the dropdown so after I searched for something I have change in the dropdown back to my latest repos and viceversa.

Putting latest option inside search also means adding technical complexity. The graphQL query for search repos is different from my latest repos.

thomaslombart (commented):
Fair enough! 🙂 

Could you add an entry in the `CHANGELOG.md` file though? Here's our guidelines about that: [Version History](https://developers.raycast.com/basics/prepare-an-extension-for-store#version-history).

PR #9037, opened 2023-10-30
Update reclaim-ai extension

shkreios:
## Description
In this pull request, we've made significant improvements to the event fetching mechanism:

1. **Refactor of Event Fetching Process**: Previously, to retrieve events from all connected calendars, the process involved an API call to the accounts endpoint, followed by passing the obtained account IDs as the `calendarIds` parameter to the events endpoint. We've streamlined this process. Instead of making two separate API calls, we now simply pass the `allConnected` flag directly to the events endpoint. This simplifies the code and reduces the overall number of API calls, leading to improved performance.

2. **Bug Fix in the Menu Bar Component**: There was an issue where the menu bar component wasn't displaying any data until the user clicked on it. This was due to the component executing only one React render cycle when rendered in the background. At this point, the `accountsData` was undefined/null, which was part of the check if the fetch events hook should be called. This behavior has been corrected, ensuring consistent and expected data rendering.

## Checklist

### DIFF
--- a/extensions/reclaim-ai/CHANGELOG.md
+++ b/extensions/reclaim-ai/CHANGELOG.md
@@ -1,5 +1,10 @@
 # reclaim Changelog
 
+## [Update] - 2023-10-31
+
+- Refactor of Event Fetching Process
+- Bug Fix in the Menu Bar Component
+
 ## [Update] - 2023-10-24
 
 - Revert filter logic


ajohnson1200:
👍 from us at Reclaim. I worked with Simon on this. We'd like this merged ASAP.

pernielsentikaer:
Done 🙂

PR #9061, opened 2023-11-01
[Installed Extension] Various enhancements

pernielsentikaer:
installed-extensions 2023-11-01 at 15 15 22

Added a few enhancements @erics118 - do you mind checking 🙂

### DIFF
--- a/extensions/installed-extensions/src/index.tsx
+++ b/extensions/installed-extensions/src/index.tsx
@@ -150,7 +150,7 @@ export default function IndexCommand() {
               accessories.push({
                 tag: { color: Color.Red, value: "Private" },
                 icon: Icon.EyeDisabled,
-                tooltip: "Private Extensions",
+                tooltip: "Private Extension",
               });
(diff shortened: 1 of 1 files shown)


extensions/installed-extensions/src/index.tsx
```diff
@@ -52,13 +52,14 @@ export default function IndexCommand() {
     let result = await Promise.all(
       files.map(async (file) => {
         const content = await fs.readFile(file, "utf-8");
+        const stats = await fs.stat(file);
         const json = await JSON.parse(content);
```
erics118, line 56: This `await` for the `JSON.parse` is extraneous

erics118:
Looks good, just one small nitpick that I didn't notice before

extensions/installed-extensions/src/index.tsx
pernielsentikaer: Thanks for noticing 👏

PR #9101, opened 2023-11-04
[Espanso]: Add ability to display match labels to Search Matches command

### LINKED ISSUE #8235: [Espanso] Copies "Label" field instead of "Replace" field.
### Extension

https://www.raycast.com/kud/espanso

### Description

First of all thanks for this extension. The idea is very neat. I hope when the issue is fixed, it will be part of my workflow and will boost my productivity.

The expectation is that when I am searching for a snippet, the search should be both on the trigger, and on the label field. The output of the search should ideally show the trigger, and the replace parameter but copy the replace parameter by default. Currently, it is copying the label field which is useless as it is a description of the snippet.

A sample snippet:

```
    - label: Bug emoji
      replace: 🐛
      trigger: "!ebug"
```

### Steps To Reproduce

Make a new snippet in your espanso config file as provided in description of this bug. 

Case 1: Try to search for espanso config with text "emoji". 
Case 2: Try to search for espanso config with text "bug".

### Current Behaviour

Case 1: There is no match found.
Case 2: The match is found. However, the right window shows the snippet label text "Bug Emoji" instead of replace value. Pressing "Enter" copies the label text.

### Expected Behaviour

Case 1: The match should correctly filter the snippets based either on label or on the trigger field.
Case 2: The matched entity should copy the trigger by default. The right window can be enhanced to show full snippet details but pressing enter should copy the replace value.

@kud please have a look.

nbbaier:
# Description

- Changes the method of getting matches for the Search Matches command, directly checking YAML files instead of using Espanso's shell command
- Introduces the ability to display [labels](https://espanso.org/docs/matches/basics/#search-labels) in search results and therefore resolves #8235 (labels overriding the replace of an Espanso match).
- Improves the fetching of matches and match files, accounting for potential differences in configs that was not there before

## Checklist

### DIFF
--- a/extensions/espanso/src/index.tsx
+++ b/extensions/espanso/src/index.tsx
@@ -25,11 +25,11 @@ export default function Command() {
         const { packages: packageFilesDirectory, match: matchFilesDirectory } = await getEspansoConfig();
         const packageMatches = getMatches(packageFilesDirectory, { packagePath: true });
         const userMatches = getMatches(matchFilesDirectory);
-        let matches: NormalizedEspansoMatch[] = userMatches.concat(packageMatches);
+        const combinedMatches: NormalizedEspansoMatch[] = userMatches.concat(packageMatches);
 
-        matches = sortMatches(matches);
+        const sortedMatches = sortMatches(combinedMatches);
 
-        setItems(matches);
+        setItems(sortedMatches);
         setIsLoading(false);
       } catch (err) {
         setError(err instanceof ProcessOutput ? err : null);


pernielsentikaer:
@nbbaier thanks for your contribution 🔥 Do you mind adding a [CHANGELOG](https://developers.raycast.com/basics/prepare-an-extension-for-store#version-history) entry 🙂

@kud do you mind checking this 🙂

nbbaier:
> @nbbaier thanks for your contribution 🔥 Do you mind adding a CHANGELOG entry 🙂

Done!

kud:
Hey, thanks for the PR. Could you please update the title of the PR to make it meaningful? 🙏

Regarding the PR itself, to understand it better, instead of using the shell command, are you checking the espanso YAML files? It seems legitimate to me. However, I didn't notice initially that the previous PR added "Add match," but it's broken. The issue lies in assuming that people use the default folders. For some reason that I haven't explained (maybe a legacy one), I don't use them, but I use `~/Library/Preferences/espanso`.

I will change it to use the default ones now.

However, this means that the code is not secure enough. Could you please use something like `espanso path` to determine the location of the config? It would be safer. (Unless you know a better way to find the config files). Thanks.

nbbaier:
> Hey, thanks for the PR. Could you please update the title of the PR to make it meaningful? 🙏

Done!

> Regarding the PR itself, to understand it better, instead of using the shell command, are you checking the espanso YAML files? It seems legitimate to me. 

Yes, exactly. I intended this both as a method to get the labels from the YAML files and to also allow for future addition of the displaying of variables in the list item detail view (since neither variables, forms, labels, etc are returned by `espanso match list -j`. 

> However, I didn't notice initially that the previous PR added "Add match," but it's broken. The issue lies in assuming that people use the default folders. For some reason that I haven't explained (maybe a legacy one), I don't use them, but I use `~/Library/Preferences/espanso`. However, this means that the code is not secure enough. Could you please use something like `espanso path` to determine the location of the config? It would be safer. (Unless you know a better way to find the config files). Thanks.

Will do! I forgot about `espanso path` and this will be a nice way to do it.

nbbaier:
@kud, those fixes have been made

extensions/espanso/src/addMatch.tsx
```diff
@@ -1,19 +1,50 @@
...
+    }
+    getMatchFilesDirectory();
+  }, []);
+
+  useEffect(() => {
+    async function fetchMatchFiles() {
+      try {
+        const { match: matchFilesDirectory } = await getEspansoConfig();
+        const sortedMatchFiles: { file: string; mtime: number }[] = getAndSortTargetFiles(matchFilesDirectory);
+        console.log("hello");
```
kud, line 35: Hello to you too :D

extensions/espanso/src/addMatch.tsx
```diff
@@ -1,19 +1,50 @@
...
   matchFile: string;
 }
 
 export default function Command() {
+  const [matchFilesDirectory, setMatchFilesDirectory] = useState<string>("null");
+  const [sortedMatchFiles, setSortedMatchFiles] = useState<{ file: string; mtime: number }[]>([]);
+  const [error, setError] = useState<Error | null>(null);
+
+  useEffect(() => {
+    async function getMatchFilesDirectory() {
```
kud, line 19: I prefer `const getMatchFilesDirectory = aynsc () =>` but it's just a preference.

kud: I wonder if you could export this function from the main command and pass functions as props.

extensions/espanso/src/index.tsx
```diff
@@ -21,9 +22,22 @@ export default function Command() {
...
     async function fetchData() {
       try {
-        const { stdout: result } = await $`espanso match list -j`;
-        let matches: EspansoMatch[] = JSON.parse(result);
-        matches = matches.sort((a, b) => a.triggers[0].localeCompare(b.triggers[0]));
+        const { packages: packageFilesDirectory, match: matchFilesDirectory } = await getEspansoConfig();
+        const packageMatches = getMatches(packageFilesDirectory, { packagePath: true });
+        const userMatches = getMatches(matchFilesDirectory);
+        const matches: NormalizedEspansoMatch[] = userMatches.concat(packageMatches);
+        matches.sort((a, b) => {
```
kud, line 29: Export this function please 😌

extensions/espanso/src/utils.ts
```diff
@@ -33,3 +34,71 @@ export function appendMatchToFile(fileContent: string, fileName: string, espanso
   fse.appendFileSync(filePath, fileContent);
   return { fileName, filePath };
 }
+
+export function getMatches(espansoMatchDir: string, options?: { packagePath: boolean }): NormalizedEspansoMatch[] {
+  let matchFiles: string[];
```
kud, line 39:
```js
fse.readdirSync(espansoMatchDir)
    .filter(fileName => options?.packagePath ? true : path.extname(fileName).toLowerCase() === '.yml') // Filter based on options
    .filter(fileName => fileName !== '.DS_Store')
    .map(fileName => {
      const filePath = path.join(espansoMatchDir, fileName);
      return options?.packagePath ? path.join(filePath, 'package.yml') : filePath;
    })
    .filter(filePath => !options?.packagePath || fse.statSync(filePath).isFile());
```

kud (changes requested):

extensions/espanso/src/addMatch.tsx
kud: `async function getMatchFilesDirectory(matchFilesDirectory)`

Something like this. But maybe it's too much optimisation.

nbbaier: I can try it out!

extensions/espanso/src/index.tsx
nbbaier: Could you say a little more about what you mean here as well?

extensions/espanso/src/addMatch.tsx
nbbaier: Ended up going this direction:

```ts
  useEffect(() => {
    const fetchData = async () => {
      try {
        const { match: matchFilesDirectory } = await getEspansoConfig();
        const sortedMatchFiles: { file: string; mtime: number }[] = getAndSortTargetFiles(matchFilesDirectory);
        setMatchFilesDirectory(matchFilesDirectory);
        setSortedMatchFiles(sortedMatchFiles);
      } catch (err) {
        setError(err instanceof Error ? err : null);
      }
    };
    fetchData();
  }, []);
  ```
  Collapsing the two `useEffect()`s into one requiring only one call to `getEspansoConfig()`.

File: extensions/espanso/src/index.tsx (hunk 3 above)
Comment (kud) [reply extensions/espanso/src/index.tsx:29]: Yeah sorry, I was a bit busy haha.

It's actually a generic function, it could be extracted from the main function and being on the top or in utils.ts.

Comment (nbbaier) [reply extensions/espanso/src/index.tsx:29]: Got it, will do.

**Review by kud — APPROVED** [review]

File: extensions/espanso/src/index.tsx [hunk 5]
```diff
@@ -21,9 +22,13 @@ export default function Command() {
...
       try {
-        const { stdout: result } = await $`espanso match list -j`;
-        let matches: EspansoMatch[] = JSON.parse(result);
-        matches = matches.sort((a, b) => a.triggers[0].localeCompare(b.triggers[0]));
+        const { packages: packageFilesDirectory, match: matchFilesDirectory } = await getEspansoConfig();
+        const packageMatches = getMatches(packageFilesDirectory, { packagePath: true });
+        const userMatches = getMatches(matchFilesDirectory);
+        let matches: NormalizedEspansoMatch[] = userMatches.concat(packageMatches);
+
+        matches = sortMatches(matches);
```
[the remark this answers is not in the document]
nbbaier: I changed this to avoid the manipulation of `matches`.

kud:
Thanks. It's really neat now!

nbbaier:
> Thanks. It's really neat now!

Thanks for your suggestions! Still getting the hang of writing neat code haha

kud:
Hehe congrats!

pernielsentikaer:
Thanks for handling @kud, and thanks @nbbaier for the enhancement

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #9185, opened 2023-11-09
feat(kill-process): allow killing multiple processes

erics118:
## Description

pretty simple. it just adds a preference to not close the main window after killing a process

## Screencast

https://github.com/raycast/extensions/assets/52634785/ffc52fb1-f209-43f9-87d3-7608847cab4d

## Checklist

### DIFF
--- a/extensions/kill-process/package.json
+++ b/extensions/kill-process/package.json
@@ -23,7 +23,7 @@
   "preferences": [
     {
       "description": "Some processes are named differently than the app itself. This flag determines whether\nthe path of the process should be searched for the query, or only the process name.",
-      "label": "Filtering should search process paths",
+      "label": "Search Process Paths",
       "name": "shouldSearchInPaths",
       "required": false,
       "default": false,
@@ -32,7 +32,7 @@
     },
     {
       "description": "Determines whether the pid of the process should be searched for the query, or only the process name.",
-      "label": "Filtering should search process pid",
+      "label": "Search Process PID",
       "name": "shouldSearchInPid",
       "required": false,
       "default": false,
@@ -41,7 +41,7 @@
     },
     {
       "description": "Determines whether apps are listed before other results when filtering.",
-      "label": "Prioritize apps when filtering",
+      "label": "Prioritize Apps",
       "name": "shouldPrioritizeAppsWhenFiltering",
       "required": false,
       "default": false,
@@ -50,7 +50,7 @@
     },
     {
       "description": "Determines whether processes are sorted by CPU or memory usage.",
-      "label": "Sort processes by memory usage",
+      "label": "Sort by Memory Usage",
       "name": "sortByMem",
       "required": false,
       "default": false,
@@ -59,7 +59,7 @@
     },
     {
       "description": "Determines whether aggregate processes belong to the same app.",
-      "label": "Aggregate apps",
+      "label": "Aggregate Apps",
       "name": "aggregateApps",
       "required": false,
       "default": false,
@@ -76,7 +76,7 @@
       "type": "checkbox"
     },
     {
-      "description": "Show process Path.",
+      "description": "Show process path.",
       "label": "Show Process Path",
       "name": "shouldShowPath",
       "required": false,
@@ -86,7 +86,7 @@
     },
     {
       "description": "Close window after killing a process",
-      "label": "Close window after kill",
+      "label": "Close Window",
       "name": "closeWindowAfterKill",
       "required": false,
       "default": true,
@@ -95,7 +95,7 @@
     },
     {
       "description": "Clear search bar after killing a process",
-      "label": "Clear search bar after kill",
+      "label": "Clear Search Bar",
       "name": "clearSearchBarAfterKill",
       "required": false,
       "default": true,
@@ -108,7 +108,8 @@
       "required": false,
       "title": "Refresh Duration",
       "type": "textfield",
-      "default": "3000"
+      "default": "3000",
+      "placeholder": "3000"
     }
   ],
   "dependencies": {


extensions/kill-process/src/index.tsx
```diff
@@ -82,8 +93,15 @@ export default function ProcessList() {
   const killProcess = (process: Process) => {
     exec(`kill -9 ${process.id}`);
     setFetchResult(state.filter((p) => p.id !== process.id));
-    clearSearchBar({ forceScrollToTop: true });
-    showHUD(`✅ Killed ${process.processName === "-" ? `process ${process.id}` : process.processName}`);
+    if (multipleKills) {
```
zhenpewu, line 96: i think it is not necessary to have pref for this, I think we can migrate from showHUD and showToast, while window is closed, toast will fallback to showHUD, and we can respect to the preference in "Advanced" for pop to search search.

zhenpewu: oh, for this change, you are mainly to avoid close main window or clear search bar ?

erics118: > not necessary to have pref 

My thinking was that people may be currently using the extension to just kill a single rogue process, not killing multiple. So, I don't want their workflow to be impeded

> mainly to avoid close main window or clear search bar ?

I've changed the code a bit to avoid some duplicated logic

extensions/kill-process/src/index.tsx
```diff
@@ -82,8 +93,14 @@ export default function ProcessList() {
   const killProcess = (process: Process) => {
     exec(`kill -9 ${process.id}`);
     setFetchResult(state.filter((p) => p.id !== process.id));
-    clearSearchBar({ forceScrollToTop: true });
-    showHUD(`✅ Killed ${process.processName === "-" ? `process ${process.id}` : process.processName}`);
+    if (!multipleKills) {
+      closeMainWindow();
+      popToRoot({ clearSearchBar: true })
```
zhenpewu, line 98: I think it is not necessary to add popToRoot, and this should respect to the the config in the "Advance" settings.

extensions/kill-process/src/index.tsx
```diff
@@ -82,8 +92,13 @@ export default function ProcessList() {
   const killProcess = (process: Process) => {
     exec(`kill -9 ${process.id}`);
     setFetchResult(state.filter((p) => p.id !== process.id));
-    clearSearchBar({ forceScrollToTop: true });
-    showHUD(`✅ Killed ${process.processName === "-" ? `process ${process.id}` : process.processName}`);
+    if (!multipleKills) {
+      closeMainWindow();
```
zhenpewu, line 96: regarding close main window, personally, I think it is not needed, but other reviewers might have different thoughts.

rolandleth: I agree with this as well.

rolandleth: I don't think it's needed either, for me it already closes it, so I think it just respects Raycast's settings.

extensions/kill-process/src/index.tsx
rolandleth: I also agree with this, no need to pop.

extensions/kill-process/package.json
```diff
@@ -76,6 +77,14 @@
       "title": "",
       "type": "checkbox"
     },
+    {
+      "description": "Allow killing multiple processes at once. This keeps the Raycast window open after killing a process",
+      "label": "Allow multiple kills",
+      "name": "allowMultipleKills",
+      "required": false,
+      "title": "Multiple Processes",
```
rolandleth, line 85: I don't think `title` is needed, for checkboxes it uses the `label`.

rolandleth (changes requested):

extensions/kill-process/src/index.tsx
```diff
@@ -82,8 +92,13 @@ export default function ProcessList() {
   const killProcess = (process: Process) => {
     exec(`kill -9 ${process.id}`);
     setFetchResult(state.filter((p) => p.id !== process.id));
-    clearSearchBar({ forceScrollToTop: true });
```
rolandleth, line 85: I would also not remove this, or customise it based on this new preference, because after you kill a process, it makes sense to clear the search bar (at least in the majority of cases).

extensions/kill-process/src/index.tsx
zhenpewu: if we all agree it is not necessary to close window, I don't think multiple kills pref is needed.

the main change for this PR is to migrate from showHUD to showToast, as showHUD will close window directly so that it will be annoying to kill multiple processes.

extensions/kill-process/src/index.tsx
zhenpewu: +1

extensions/kill-process/src/index.tsx
rolandleth: Oh, I didn't know that `showHUD` is actually what's closing the window. In that case, `allowMultipleKills` doesn't make sense anymore, no? With `showToast` it will always leave the window open and you can do multiple kills.

What would make sens would be a "Close window after kill" preference, default to `true`, because most of the time/people need to kill one process and that's it.

pernielsentikaer:
That's some good feedback, so we're switching to a toast that doesn't close the window - we could have a preference for letting the user chose to close the window (which would do as today) and let them toggle it off to leave open?

extensions/kill-process/package.json
erics118: I thought it would look better with a heading. WIth the title field, it looks like this:
CleanShot 2023-11-10 at 08 26 26

Without it, it would just get smushed into the previous category, Display:
CleanShot 2023-11-10 at 08 29 15

extensions/kill-process/src/index.tsx
erics118: I've added another preference option to customize the clearSearchBar behavior.

extensions/kill-process/src/index.tsx
erics118: Yep, I've changed the preference to `closeWindowAfterKill`, with a default to `true`

erics118:
I believe I've implemented the requested changes

extensions/kill-process/src/index.tsx
rolandleth: @pernielsentikaer Yup, that's what we're discussing here 😄

extensions/kill-process/package.json
```diff
@@ -73,9 +80,27 @@
...
       "title": "",
       "type": "checkbox"
     },
+    {
+      "description": "Close window after killing a process",
+      "label": "Close window after kill",
+      "name": "closeWindowAfterKill",
+      "required": false,
+      "default": true,
+      "title": "After Killing",
```
rolandleth, line 93: Since we have this as a title (heading), I think we can label the two actions "Close window" and "Clear search bar". @pernielsentikaer what do you think?

rolandleth (changes requested):

extensions/kill-process/package.json
erics118: I agree, as it's pretty redundant

rolandleth (approved):

zhenpewu (approved):

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #9205, opened 2023-11-11
Add Fast Chat Extension

TimonPeng:
## Description

A Raycast extension enables faster chatting with a target phone number in IM apps like WhatsApp and Telegram, eliminating the need to add contacts beforehand.

## Screencast

https://github.com/raycast/extensions/assets/8691151/6732c49b-2efe-409a-8970-d6c4a8fc59a5

## Checklist

### DIFF
--- a/extensions/phonenumber-in-im/package.json
+++ b/extensions/phonenumber-in-im/package.json
@@ -5,7 +5,6 @@
   "description": "A Raycast extension enables faster chatting with a target phone number in IM apps like WhatsApp and Telegram, eliminating the need to add contacts beforehand.",
   "icon": "logo.png",
   "author": "timonpeng",
-  "owner": "hardfury",
   "categories": [
     "Communication"
   ],


pernielsentikaer (commented):
Hi @TimonPeng 👋 

This feels a bit like [Texts extension](https://www.raycast.com/zavbala/texts) from @zavbala - correct me if wrong 🙂

TimonPeng:
> Hi @TimonPeng 👋
> 
> This feels a bit like [Texts extension](https://www.raycast.com/zavbala/texts) from @zavbala - correct me if wrong 🙂

Hi @pernielsentikaer 👋

After experiencing the Texts extension, I believe my submitted extension differs from Texts in several aspects:

1. My extension directly calls based on bundle ID and URL scheme (Texts redirects to a browser first, whereas mine directly opens the corresponding app).
2. Texts do not preprocess or format input parameters for WhatsApp.
3. Texts supports Telegram via username, not phone number (in fact, using a username makes it convenient to converse directly in Telegram).
4. Browser-based calls only invoke apps that handle URL schemes. This means when multiple release versions are installed, additional steps may be needed to make it work.
5. My extension focuses solely on `conversations with phone numbers` and it can easily support multiple release versions through set preferences. Chatting with phone numbers in IM Apps usually requires adding contacts first, so this is where my extension optimizes the workflow.

Therefore, I believe the two extensions fundamentally perform different tasks, hence I cannot directly contribute to the Texts extension either.

pernielsentikaer:
Thanks for the clarification, I'll give it another whirl 🙂

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #9241, opened 2023-11-14
Add code-saver extension

Gnosnay:
## Description

Store your code snippets and export them freely when you need it.

Pure local managed Snippet store. You should 
1. Pick one local data store first. 
2. Create one Library then
3. Create snippet that you need.
4. Brower the snippts you created.

## Screencast

code-saver-1
code-saver-2
code-saver-3
code-saver-4
code-saver-5
code-saver-6
code-saver-7
code-saver-8
code-saver-9
code-saver-10

## Checklist


pernielsentikaer:
You should be able to use `useSql` instead, you can always make it safe yourself by escaping correctly (even tho it's only deading with local databases so worse thing that can happen is a user hacking its own db...)

pernielsentikaer:
I still suggest that you use `useSql` so we don't have to deal with multiple binaries and have a way to patch things as we go. There shouldn't be SQL injections if you escape your queries correctly (even though it's only dealing with local databases, so the worst thing that can happen is a user hacking their own db...)

Gnosnay:
> I still suggest that you use `useSql` so we don't have to deal with multiple binaries and have a way to patch things as we go. There shouldn't be SQL injections if you escape your queries correctly (even though it's only dealing with local databases, so the worst thing that can happen is a user hacking their own db...)

oh i think i got what you mean... cuz i got this error today.

image

so raycast will change the version of nodejs it carries from time to time?

if so, then it will be very troublesome.....

BTW, may i know whether `useSql` can execute online DDL?

PR #9266, opened 2023-11-15
Add omnivore extension

KarolusD:
## Checklist

### DIFF
--- a/extensions/omnivore/src/omnivore-capture.tsx
+++ b/extensions/omnivore/src/omnivore-capture.tsx
@@ -2,14 +2,14 @@ import { LaunchProps, Toast, showToast } from '@raycast/api'
 import { saveUrl } from './utils'
 
 export default async function QuickCapture(props: LaunchProps<{ arguments: Arguments.OmnivoreCapture }>) {
-  const { url, labels } = props.arguments
+  const { url } = props.arguments
 
   showToast({
     style: Toast.Style.Animated,
     title: 'Saving Article',
   })
 
-  const urlSaveResponse = await saveUrl(url, labels)
+  const urlSaveResponse = await saveUrl(url)
   if (urlSaveResponse.success) {
     showToast({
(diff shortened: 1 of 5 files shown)


extensions/omnivore/src/omnivore-capture.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+  try {
+    const isUrlSaved = await saveUrl(url, labels)
+    if (isUrlSaved) {
+      showToast({
+        style: Toast.Style.Success,
+        title: 'Saved!',
+        message: 'Your URL was saved',
+      })
+    }
```
pernielsentikaer, line 20: Maybe add a failure toast here if `isUrlSaved` is `false`

pernielsentikaer (changes requested):
Hi 👋 

I have now tested your extension, and I have some feedback ready for you

extensions/omnivore/src/omnivore-capture.tsx
KarolusD: Agree, actually because how the saveUrl function is constructed (it returns true or false always) we don't need try/catch block here - if/else block should be enough and it should simplify the code.

pernielsentikaer (changes requested):
If I enter a wrong key it still says success here. Could you look into that?

Raycast 2023-11-28 at 10 44 28

pernielsentikaer:
I'm not sure if you're done yet, but still getting √ Saved 🫠

KarolusD:
> I'm not sure if you're done yet, but still getting √ Saved 🫠

What do you mean by wrong key? I wrote url safeguard by using new URL().
Are you passing random string or some incomplete URL?

KarolusD:
@pernielsentikaer there is not way to have instant feedback after saving fake url, it have to be processed on the omnivore side first. But I've added API key as required field, so you shouldn't be able to save nor search articles before providing the api key.

PR #9283, opened 2023-11-16
Update ip-geolocation extension

frankiexyz:
## Description

- Adding IPv6 support, the existing check does not include ":" and all the IPv6 query will be marked as invalid query
- Adding a check to use net library to valid if the user input is a valid IP address when the user input include 3 "." and no [a-z] character(indicating an IPv4 address query) or there is more than two ":" (indicating an IPv6 address query).
- For the domain query, added a check to remove uri like `https://www.foo.bar/123` becomes `www.foo.bar`

## Checklist


pernielsentikaer (commented):
Thanks for your contribution @frankiexyz 🔥 

@koinzhang do you mind checking this?

koinzhang:
@pernielsentikaer I tested it briefly and I think it's okay

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #9290, opened 2023-11-17
Add workspace extension

eeee0717:
## Description
This is a plugin for managing workspaces.

## Screencast
CleanShot 2023-11-17 at 11 20 02@2x
CleanShot 2023-11-17 at 11 22 06@2x

## Checklist

### DIFF
--- a/extensions/workspace/README.md
+++ b/extensions/workspace/README.md
@@ -1,1 +1,8 @@
 # Workspace
+This plugin is used to manage workspaces, you can create multiple workspaces, just need one click to open files, folders, websites, apps.
+
+# Sceen Shot
+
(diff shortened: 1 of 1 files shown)


pernielsentikaer:

This is similar to the extension [Xecutor](https://www.raycast.com/GastroGeek/xecutor), could you check that out?

eeee0717:
> 
> This is similar to the extension [Xecutor](https://www.raycast.com/GastroGeek/xecutor), could you check that out?

Thanks for your reply, but I don't think my extension is the same problem he's trying to solve. He can only open the URL, I do this extension purpose and hope to manage a workspace, for example, when coding, I can not only open Github, and Stackoverflow these sites, but also open some programming tools, such as VSCode and so on. So I think we are different in terms of fundamental needs.

pernielsentikaer:
The only thing I don't see in the other one is the support for files and folders, but that should be doable to add in the other one so we only have one extension with the same purpose - having two almost identical might confuse

xecutor 2023-11-21 at 09 19 10

Issue #9244, opened 2023-11-14
[Obsidian] Search Note failure


placer14:
### Extension

https://www.raycast.com/KevinBatdorf/obsidian

#### Error:
```
Error: Cannot get default application for file at /Users/<path_to_file>/2023-11-13.md


```

### Steps To Reproduce

0. Somehow disassociate markdown files with Obsidian
1. Attempt to Search Notes for a specific file (any)
2. Observe Raycast error view

### Current Behaviour

Shows Raycast error view

### Expected Behaviour

Shows search results from vault

placer14:
Seems to be a duplicate of https://github.com/raycast/extensions/issues/9242. Closing.

KevinBatdorf:
Opening a PR with a fix now. Thanks for reporting it.

Issue #9242, opened 2023-11-14
[Obsidian] search nog working. after entering search criteria and pressing enter, this error occurs


alexanderrkok:
### Extension

https://www.raycast.com/KevinBatdorf/obsidian

#### Error:
```
Error: Cannot get default application for file at /Users/alex/Library/Mobile Documents/iCloud~md~obsidian/Documents/Alex vault/Projects/Teamcoach opleiding/Reflectieverslagen/Opleidingsdagen/Reflectieverslag TC06 - Coachen bij stagnerende en regressieve teams.md


```

### Steps To Reproduce

Lastest version van Mac OS and Obsidian version 1.4.16 of raycast  version 1.61.2 
I will get an error when I search in raycast for search note. 
I enter search criteria in the search box and press enter
'something went wrong' is all I see.

### Current Behaviour

<img width="788" alt="image" src="https://github.com/raycast/extensions/assets/39742244/6c983a15-df9d-40b5-9b21-cdfb5dcf624c">
<img width="802" alt="image" src="https://github.com/raycast/extensions/assets/39742244/4628e9d1-508a-48c5-8b3f-a436a7c111e4">

### Expected Behaviour

search results.

KevinBatdorf:
Opening a PR with a fix now. Thanks for reporting it.

alexanderrkok:
Thnx, works like a charm now :-D

PR #9324, opened 2023-11-19
[Obsidian] ignore empty cache

KevinBatdorf:
## Description

This PR does 3 things:

1. If the cache is empty (`[]`) then ignore it
2. If the `getDefaultApplication` promise rejects, it handles that properly now
3. In the case where the user deleted the custom config location in prefs, use `.obsidian`

Closes https://github.com/raycast/extensions/issues/9242
Closes https://github.com/raycast/extensions/issues/9244


pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

Issue #9317, opened 2023-11-18
[Twitch] Add ability to browse VODs


Sheraff:
### Extension

https://www.raycast.com/dieserRobin/twitch

### Description

Currently the extension only allows browsing *live* channels. Maybe adding the ability to browse on-demand videos would be a nice addition.

### Who will benefit from this feature?

Everyone watching VODs, i mean who watches Critical Role episodes in 1 sitting?

### Anything else?

I started implementing the idea. I've got the basics working, it's just a matter of polishing now. And to me this is the least fun part so I'm opening this issue before just in case this is a hard no.

The main view for "followed channels" could display them all, even if not live at the moment, with an icon displaying whether they are live, have VODs, or nothing at all. Or it could be a separate command like "offline followed channels.
![twitch-6](https://github.com/raycast/extensions/assets/1325721/701d37f1-63ad-4797-9eaa-ac01f464d119)

List of followed channels should be searchable, because if we don't just display the live ones, the list might get big.
![twitch-7](https://github.com/raycast/extensions/assets/1325721/c2abac62-0aeb-4861-af1f-1dda21412a6c)

One of the actions (the main one?) pushes a 2nd view that lists all of the VODs (the limit is 20 for the twitch API, but we could paginate). This list contains an item for "live stream" on top if the channel is live. Actions are pretty much the same as the current behavior.
![twitch-8](https://github.com/raycast/extensions/assets/1325721/1fa28b53-79e8-457d-a900-5d8c79541a27)

Sheraff:
Here's a repo with my changes https://github.com/Sheraff/raycast-twitch-extension. I also ended up adding the official Twitch OAuth because it feels better than going through a 3rd party.

PR #9349, opened 2023-11-21
Update twitch extension

Sheraff:
This PR is a work in progress, though I'd love some feedback from anyone motivated enough to read through the changes.

Closes https://github.com/raycast/extensions/issues/9317

## Description

- add a way to search through VODs of the followed channels. This command could be a replacement for the existing one that simply lists the followed channels that are currently online.
- add a menu bar item that shows all currently online followed channels.
- add official Twitch OAuth flow, this feels safer than having to grant access to a 3rd party platform

## Screencast

https://github.com/raycast/extensions/assets/1325721/9b4d4b22-212a-4894-a1c4-64a49c3e5289

## Checklist


Sheraff:
I had a bunch of troubles with the OAuth, would love any feedback if anyone has some answers.

---

# through Implicit Grant flow using `response_type=token`

I'm trying to setup the OAuth flow for twitch, but I'm getting stuck after the redirection to raycast.com, where the app doesn't open.

The URL at that point is this
```shell
https://www.raycast.com/redirect
  ?packageName=Extension
  #access_token=[redacted]
    &scope=user:read:follows
    &state=[redacted]
    &token_type=bearer
```

where `state` decodes to
```json
{
  "id": [redacted],
  "providerName":" Twitch",
  "flavor": "release"
}
```

As far as I can tell, the logic inside the `<Redirect>` component that is rendered on the `raycast.com/redirect` route seems to indicate that there should be a `code` search param for the URL to be properly generated in the `onClick` listener of the `<button>` (or in the `useEffect` that is supposed to call the same listener instantly) and applied to `window.location.href =`.

I'm not sure why I'm missing this param (or if I'm completely wrong and missing something else entirely). 

During the OAuth flow, at `client.authorize(authRequest)`, the `authRequest.toURL()` resolves to

```shell
https://id.twitch.tv/oauth2/authorize
  ?client_id=[redacted]
  &redirect_uri=https://raycast.com/redirect
    ?packageName=Extension
    &response_type=token
    &scope=user:read:follows
    &code_challenge=[redacted]
    &code_challenge_method=S256
    &state=[redacted]
```

where `state` decodes to the same thing as above. I did try the alternative redirect URI (`https://raycast.com/redirect/extension`) and I get the same result (except hash is a search param instead).

---

At this point I'm assuming the issue is on the Twitch side, maybe they do not support PKCE

---

Just in case, I checked the `lsregister -dump URLSchemeBinding` and I do find appropriate entries for raycast URI schemes
```log
com.raycast:                com.raycast.macos (0x2778) (0x277a)
raycast:                    com.raycast.macos (0x2778) (0x277a)
```

# Through OIDC authorization code grant flow using `response_type=code`

This time I do have a `code` search param in the URL, and raycast manages to open the extension. However there is a new problem: the `scope` key in the `OAuth.TokenResponse` passed to `client.setTokens()` is not a string: twitch sends back an array.

For this to work, I have to override it with `token.scope = token.scope.join(' ')`.

# Refreshing the access token in the background of a `"mode": "menu-bar"` command

This is kind of hard to reproduce, but I'm pretty sure I saw this issue at least once: refreshing the access token from the menu bar script (`"interval": "10m"`) seems to fail. I'll update if I find a way to properly reproduce this, or at least the log of such an error.

# Using `clientSecret` in OAuth process

Since the [OIDC Authorization code grant flow](https://dev.twitch.tv/docs/authentication/getting-tokens-oidc/#oidc-authorization-code-grant-flow) is the only one I managed to make work, Twitch requires a `clientSecret` which Raycast seems to [recommend against](https://developers.raycast.com/api-reference/oauth#oauth-app). 

The way I set this up, it's up to each individual user of the extension to add their own id+secret pair through the appropriate `preferences` fields. This seems fine to me but I'd love some feedback.

pernielsentikaer:
Is this ready or are you still in touch with @thomaslombart regarding this?

Sheraff:
> Is this ready or are you still in touch with @thomaslombart regarding this?

There are 2 issues left, but which (probably) won't impact much of the 2.8k lines of code:
- oauth right now requires the user to create their own twitch app, and paste ID and secret in the extension preferences. Not sure if we can do better, this is what I asked Thomas about.
- menu bar command interrupts execution before finishing its refetch, I think it's because the previous fetch is cached, and so loading is `false` from the beginning. I can fix that soon.

extensions/twitch/src/helpers/auth.tsx
```diff
@@ -0,0 +1,109 @@
...
+    return tokenSet.accessToken;
+  }
+  const nonce = randomUUID();
+  const authRequest = await client.authorizationRequest({
+    endpoint: "https://id.twitch.tv/oauth2/authorize",
+    clientId,
+    scope: "user:read:follows",
+    extraParameters: {
+      response_type: "code",
+      redirect_uri: "https://raycast.com/redirect?packageName=Extension",
```
thomaslombart, line 52: These two parameters are automatically added by the `authorizationRequest` call. You shouldn't need to add these.

extensions/twitch/src/helpers/auth.tsx
```diff
@@ -0,0 +1,109 @@
...
+  const accessToken = await promise;
+  const headers = {
+    "Client-Id": clientId,
+    Authorization: `Bearer ${accessToken}`,
+  } as const;
+  return headers;
+}
+
+const client = new OAuth.PKCEClient({
+  redirectMethod: OAuth.RedirectMethod.AppURI,
```
thomaslombart, line 26: Why not using the `Web` method here as you're passing the web static URL?

extensions/twitch/src/helpers/auth.tsx
```diff
@@ -0,0 +1,109 @@
...
+): Promise<OAuth.TokenResponse> {
+  const response = await fetch("https://id.twitch.tv/oauth2/token", {
+    method: "POST",
+    body: new URLSearchParams({
+      client_id: clientId,
+      client_secret: clientSecret,
+      code: authorizationCode,
+      code_verifier: authRequest.codeVerifier,
+      grant_type: "authorization_code",
+      redirect_uri: "https://raycast.com/redirect?packageName=Extension",
```
thomaslombart, line 75: You can also use `authRequest.redirectURI` here.

Sheraff:
nope not stale, waiting for someone to come back from their holidays

Sheraff:
@thomaslombart 

TL;DR: could we change the .env of Raycast's Twitch oauth proxy to use `JSON_OR_FORM=form`?

--- 

Ok so with the new raycat proxy for oauth, the authorizationRequest seems to pass ok, but when fetching a token, I get one of 2 errors
```js
{
  statusCode: 415,
  code: 'FST_ERR_CTP_INVALID_MEDIA_TYPE',
  error: 'Unsupported Media Type',
  message: 'Unsupported Media Type: application/x-www-form-urlencoded;charset=UTF-8'
}
```
or
```js
{
  status: 400,
  message: 'missing client secret'
}
```

The 2nd kind seems to come from the `"json"` path of the `/token` route in the fastify server for the oauth proxy (https://github.com/mathieudutour/pkce-proxy/blob/9f7f36ccf13f4c5ba4f988b128e35dbdabba7da0/src/token.ts#L10-L48) which I'm assuming is the same code for every service, with just the .env changed.

With the same assumptions, the 1st kind of error must come from the fact that `"json"` is hardcoded (or not declared at all) the the .env file for the twitch proxy (`process.env.JSON_OR_FORM`) when we would need `"form"` to work with Twitch.

thomaslombart:
@Sheraff Done.

Sheraff:
> @Sheraff Done.

@thomaslombart Done myself too. Appart from a couple question I asked on slack, I think this is ready for review.

---

Just in case you have the time:

> Hey all, I have an extension that only requires a preference item for some commands, not all. Is there a way to trigger the nice looking "welcome, we need some info" UI of a required preference? For now I'm just falling back to a default "not available" screen, with an action that calls openExtensionPreferences, but it's not as nice...

> Other question: is there a way to trigger some form of notification from a "mode": "menu-bar", command? I know we can run applescripts, but those only allow a few params (no image, no on-click callback)

extensions/twitch/src/helpers/datetime.tsx
thomaslombart, line 1: It may be worth it taking a look at `date-ins` or `day.js` here for these kind of utilities.

extensions/twitch/src/helpers/useTwitchRequest.tsx
```diff
@@ -0,0 +1,149 @@
...
+  return true;
+}
+
+function hasObjectPrototype(o: any): boolean {
+  return Object.prototype.toString.call(o) === "[object Object]";
+}
+
+const defaultSelect = (data: any) => data.data;
+
+export function useTwitchRequest<T>({
```
thomaslombart, line 85: Note you could use `useCachedPromise` or `useFetch` from `@raycast/utils` here. That would make some things easier to handle (loading, caching, etc.)

thomaslombart:
Everything looks good to me on the actual extension behavior 😊

extensions/twitch/package.json
```diff
@@ -9,26 +9,11 @@
...
-      "description": "Twitch ClientId https://twitchapps.com/tokengen/"
-    },
-    {
-      "name": "authorization",
-      "type": "password",
-      "default": "",
-      "title": "Authorization Token",
-      "required": true,
-      "description": "Twitch Bearer Token (Must be matching with ClientId) https://twitchapps.com/tokengen/"
-    },
```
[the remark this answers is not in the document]
Sheraff: I'm open to the idea if you think it's better for the users, but I do feel against it:
- it's less secure (unknown 3rd party controls the auth)
- supporting multiple ways of doing this, seeing how easy it is to log it with the new implementation, might lead to diverging implementations in the future (or at least increase the burden of maintenance)

extensions/twitch/src/helpers/useTwitchRequest.tsx
Sheraff: I did kinda look into it but they didn't do exactly what I needed (keeping in mind that for some commands (`following`) we need to minimize re-renders as much as possible because they already render *a lot* initially.
- `useFetch` doesn't allow injecting headers as far as I could tell
- `useCachedPromise` I don't remember exactly the issue... I might take a second look at it now that I have a stable functional requirement

extensions/twitch/package.json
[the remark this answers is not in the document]
thomaslombart: The primary purpose of keeping that is to maintain backward compatibility for users. I would say it's better that existing users don't have to suddenly log in through OAuth. However, if it turns out to be too complicated maintenance-wise, I am fine with removing it.

extensions/twitch/src/helpers/useTwitchRequest.tsx
thomaslombart: > useFetch doesn't allow injecting headers as far as I could tell

You can do so in the `options` object: https://developers.raycast.com/utilities/react-hooks/usefetch#arguments.

These hooks are supposed to make both the extension's UX and DX better, so let me know if it doesn't work out to figure out how we can improve them.

Sheraff: The issue with `options` is that obtaining the access token is async, and I don't really want it in a state (to minimize renders). Maybe `options` could also take an async function returning the options?

extensions/twitch/src/helpers/datetime.tsx
Sheraff: How strongly do you feel about this one? I kinda dislike every date library 😢

extensions/twitch/.eslintrc.json
```diff
@@ -5,6 +5,14 @@
...
+    "@typescript-eslint",
+    "@raycast"
+  ],
+  "extends": [
+    "eslint:recommended",
+    "plugin:@typescript-eslint/recommended",
+    "prettier",
+    "plugin:@raycast/recommended"
+  ]
+}
```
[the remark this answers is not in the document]
Sheraff: This seems to give the following error:
```shell
Oops! Something went wrong! :(

ESLint: 7.32.0

ESLint couldn't find the config "@raycast" to extend from. Please check that the name of the config is correct.

The config "@raycast" was referenced from the config file in "/Users/Flo/Projects/raycast-extensions/twitch/.eslintrc.json".

If you still have problems, please stop by https://eslint.org/chat/help to chat with the team.
```

extensions/twitch/src/helpers/datetime.tsx
thomaslombart: No strong feelings, you can discard the comment if you want 🙂

thomaslombart (commented):
There's one last small thing I've noticed: the Twitch icon appears cropped in the menu bar extension. Is it possible to change it to make it look nicer?

CleanShot 2024-01-05 at 16 36 47@2x

Also, is it necessary to ask for the user's username? It would be better to avoid doing so to make the onboarding process even easier. No worries if it's not possible.

Sheraff:
TL;DR: done and done

---

> There's one last small thing I've noticed: the Twitch icon appears cropped in the menu bar extension. Is it possible to change it to make it look nicer?

I hadn't noticed! It's a png from the official press kit (https://brand.twitch.tv/downloads/twitch-brand-assets.zip) so I didn't even think to touch it. I cropped it to a square, it removes a bit of the coma below, but only on the end of the shadow so it's not too noticeable. Tell me if that looks ok for you.

> Also, is it necessary to ask for the user's username? It would be better to avoid doing so to make the onboarding process even easier. No worries if it's not possible.

So... I had tried without, and I got errors, and then the doc mentioned something like "some tokens don't have a user associated" and I thought that was the price to pay for going with a Raycast clientId. But I just re-tried now and it seems to work. Let's hope the errors I had before was because I was still battling with the auth.

thomaslombart (approved):
Great work! 🚀 
Sorry it took a long time.

PR #9351, opened 2023-11-21
Add mound-for-pile extension

cr34t1ve:
## Checklist


extensions/mound-for-pile/src/utils/types.ts
```diff
@@ -0,0 +1,57 @@
...
+  },
+  {
+    highlight: PostHightlightsI.DoLater,
+    highlightColor: "#4de64d",
+  },
+  {
+    highlight: PostHightlightsI.NewIdea,
+    highlightColor: "#017AFF",
+  },
+];
```
pernielsentikaer, line 52: I would rely on `Color` from `@raycast/api` instead, this should always look good if the user is having a theme, is that possible?

pernielsentikaer (changes requested):
Just one little thing  🙂

extensions/mound-for-pile/src/utils/types.ts
pernielsentikaer: You can still use the defined colors, its more a heads up that it might go a bit bad with custom themes 🙂

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

extensions/mound-for-pile/src/utils/types.ts
cr34t1ve: got it!. I'll change that up in the next build 😄

PR #9382, opened 2023-11-23
FIXED: the problem of long names

sbugzu:
## Description

When the name is longer than 30 characters, the name taken by the `scutil` command will lose the trailing double quotes.

issue: https://www.raycast.com/extension-issues/sbugzhu/wireguard/4639679654


pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

Issue #9361, opened 2023-11-22
[Notion] Open page in browser or app when searching pages


Yakitrak:
### Extension

https://www.raycast.com/notion/notion

### Description

Currently we can configure the default settings to open a page in browser or app from the extension settings but this applies to everything. It would be helpful to use the actions menu when searching a page and chose to open in app/browser specifically for that interaction.

### Who will benefit from this feature?

Any user who wants to have more control on where certain pages are opened

### Anything else?

I am happy to add a PR for functionality if you agree - would appreciate any guidance.

thomaslombart:
Thanks for the feature request. Would you like to work on it? I can provide some guidance on how to contribute to this extension so that you can open a pull request. You can contact me here or in the [Slack community](http://raycastcommunity.slack.com/) if you need further help.

Note that opening up a PR and having it merged will earn you some credits as we want to reward contributions. You’ll then be able to exchange them in a swag store that we’ll launch soon!

Yakitrak:
@thomaslombart thank you for the reply! Yes I'll be happy to work on it. Haven't looked into the extensions code too much but I have a rough idea how it would work - but any advice on best approach or things to keep in mind would be appreciated.

Yakitrak:
@thomaslombart I've created a quick PR for this: https://github.com/raycast/extensions/pull/9411

PR #9411, opened 2023-11-24
[Notion] Open in notion alternate action

Yakitrak:
## Description

Created from https://github.com/raycast/extensions/issues/9361

Adds another action to open in Notion, which is the opposite of the default open setting. e.g. If primary action is set to open in app, secondary action will open in browser and vice versa. As explained in the linked issue, I think this will be helpful to have a choice of where to open whilst still having the default a user would want. 

## Screencast

image

## Checklist

### DIFF
--- a/extensions/notion/src/components/PageListItem.tsx
+++ b/extensions/notion/src/components/PageListItem.tsx
@@ -12,7 +12,7 @@ import {
   PagePropertyType,
   User,
 } from "../utils/notion";
-import { handleOnOpenPage } from "../utils/openPage";
+import { handleOnOpenPage, isNotionInstalled } from "../utils/openPage";
 import { DatabaseView } from "../utils/types";
 
 import { DatabaseList } from "./DatabaseList";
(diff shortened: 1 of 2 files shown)


bkeys818 (approved):

extensions/notion/src/components/PageListItem.tsx
```diff
@@ -121,9 +121,21 @@ export function PageListItem({
     <Action title="Open in Notion" icon="notion-logo.png" onAction={() => handleOnOpenPage(page, setRecentPage)} />
   );
 
+  const { open_in } = getPreferenceValues<Preferences>();
+  const otherNotionActionTitle = open_in === "web" ? "Open in App" : "Open in Browser";
+  const otherNotionAction = open_in === "web" ? "app" : "web";
+  const openInNotionOtherAction = (
+    <Action
+      title={otherNotionActionTitle}
+      icon="notion-logo.png"
```
thomaslombart, line 130: It'd be nice that opening in the browser displays an `Icon.Globe` to make it clear the page will open in the web browser.

extensions/notion/src/components/PageListItem.tsx
```diff
@@ -121,9 +121,21 @@ export function PageListItem({
     <Action title="Open in Notion" icon="notion-logo.png" onAction={() => handleOnOpenPage(page, setRecentPage)} />
   );
 
+  const { open_in } = getPreferenceValues<Preferences>();
+  const otherNotionActionTitle = open_in === "web" ? "Open in App" : "Open in Browser";
+  const otherNotionAction = open_in === "web" ? "app" : "web";
+  const openInNotionOtherAction = (
```
thomaslombart, line 127: What about adding the `alternate` name to the variable as you did in the PR's title?

extensions/notion/src/components/PageListItem.tsx
Yakitrak: Good idea, done

thomaslombart (approved):
🚀

thomaslombart (commented):
Could you also hide the alternate action and only show `Open in Browser` if a user doesn't have the Notion app installed? Otherwise, opening it will fail with an error such as this one:

CleanShot 2023-11-29 at 17 23 56@2x

Yakitrak:
> Could you also hide the alternate action and only show `Open in Browser` if a user doesn't have the Notion app installed? Otherwise, opening it will fail with an error such as this one:
> 
> CleanShot 2023-11-29 at 17 23 56@2x

Good spot, thank you! Had to change the implementation slightly, but all done now 🙂

thomaslombart (approved):
🚢

PR #9438, opened 2023-11-27

the-devbear:
## Description

I added a new command to instantly add the current playing song to a playlist.

Please give a much feedback as possible. I'm still learning and this was the first time using "React".

## Checklist

### DIFF
--- a/extensions/spotify-player/CHANGELOG.md
+++ b/extensions/spotify-player/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## [Feature] - 2024-02-02
 
 - New command `Add Playing Song to Playlist` to directly add the current song to a playlist of your choice.
-- Removed filter for `Add Song to Playlist` command since Spotify doesn't update it corectly.
+- Removed the condition to filter collaborative playlists from other users, since Spotify doesn't update it correctly.
 
 ## [New Command] - 2023-01-31
 


extensions/spotify-player/src/components/PlaylistItem.tsx
```diff
@@ -24,7 +26,7 @@ export default function PlaylistItem({ type, playlist }: PlaylistItemProps) {
       subtitle={subtitle}
       content={icon}
       accessories={[{ text: `${playlist?.tracks?.total} songs` }]}
-      actions={<PlaylistActionPanel title={title} playlist={playlist} />}
+      actions={actions}
```
bjarn-hau, line 29: This should also work. I can change it if you like.
```suggestion
      actions
```

extensions/spotify-player/src/addCurrentSongToPlaylist.tsx
```diff
@@ -0,0 +1,124 @@
...
+function AddToPlaylistCommand() {
+  const { currentlyPlayingData, currentlyPlayingIsLoading, currentlyPlayingRevalidate } = useCurrentlyPlaying();
+  const [searchText, setSearchText] = useState("");
+
+  const { myPlaylistsData } = useMyPlaylists();
+  const { meData } = useMe();
+
+  if (!currentlyPlayingData || !currentlyPlayingData.item) {
+    return (
+      <List isLoading={currentlyPlayingIsLoading}>
```
bjarn-hau, line 32: If you like I can extract this into it's own component. It's using the same one as in `nowPlaying`

pernielsentikaer:
Keeping this open

extensions/spotify-player/package.json
```diff
@@ -152,6 +153,13 @@
       "description": "Toggle the playback of the current song.",
       "mode": "no-view"
     },
+    {
+      "name": "addPlayingSongToPlaylist",
+      "title": "Add Playing Song to Playlist",
+      "subtitle": "Spotify",
+      "description": "Adds the current song to a playlist of your choice.",
+      "mode": "view"
```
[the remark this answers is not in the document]
the-devbear: Sure will do!

peduarte:
Looks good, thanks for that. I noticed that with some playlists I get an error: 
image

This error also happens in the "Add to Playlist" action in the Now Playing commands, but I was wondering if maybe we could fix those as part of this PR as well?

Off the top of my head, Im not sure what the errors mean, but I'm guessing it's related to the [public](https://developer.spotify.com/documentation/web-api/reference/get-playlist) field? Here's [more info](https://developer.spotify.com/documentation/web-api/concepts/playlists)

the-devbear:
Yeah makes sense. I believe it's a problem with the collaborative flag. Spotify doesn't update correctly. I would remove now the `|| playlist.collaborative` from both my code and the one in nowPlaying. Would this work? Or I can leave it in and update the error message. 

The problem is that you can have a private Playlist which is collaborative but you can not add songs to.
image

peduarte:
Yeah if we can filter out these playlists that would be good I think

peduarte:
I think we also need in the Now Playing Actions
(ಠ ›ಠ) — @peduarte — jrQ6mJao004113@2x

the-devbear:
Ohh yeah, sorry I fixed for the menu bar. Now it should work.

extensions/spotify-player/src/components/AddToPlaylistAction.tsx
```diff
@@ -15,7 +15,7 @@ export function AddToPlaylistAction({ playlists, meData, uri }: AddToPlaylistAct
   return (
     <ActionPanel.Submenu icon={Icon.List} title="Add to Playlist">
       {playlists
-        ?.filter((playlist) => playlist.owner?.id === meData?.id || playlist.collaborative)
```
peduarte, line 18: Why did we remove this?

the-devbear: > Yeah makes sense. I believe it's a problem with the collaborative flag. Spotify doesn't update correctly. I would remove now the `|| playlist.collaborative` from both my code and the one in nowPlaying. Would this work? Or I can leave it in and update the error message.
Because of this.

peduarte: Ah ok, got it. thanks! Can we update the changelog to make a note that we removed this? Because I presume that for most users, they will see less playlists now

peduarte (approved):

peduarte:
Thanks so much

the-devbear:
Thanks for the Review 👍

PR #9448, opened 2023-11-28
fix(visual-studio-code): set terminal app preference required

### LINKED ISSUE #9446: [Visual Studio Code] ...
### Extension

#### Error:
```
TypeError: Cannot read properties of undefined (reading 'name')

E2:index.tsx:134:49

---
131: }
132: 
133: function LocalItem(props: { entry: EntryLike

### Steps To Reproduce

Open "Search Recent Projects", Panic occurs

```
Something went wrong
You can try reloading, or report the error to the extension developer, by
copying the log from the Action Panel K
```



### Current Behaviour

I cannot use "Search Recent Projects" command and function

### Expected Behaviour

fix

tmwrnr:
## Description

Set the terminal app preferences as required to prevent crashes.

Fixes #9446

## Checklist

### DIFF
--- a/extensions/visual-studio-code-recent-projects/CHANGELOG.md
+++ b/extensions/visual-studio-code-recent-projects/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## [Fix] - 2023-11-28
 
-- Fix the "Open With Terminal" action by making the terminal preference a requirement
+- Hopefully Fix the "Open With Terminal" action by using the bundle identifier and checking that it exists
 
 ## [Update] - 2023-11-28
 
--- a/extensions/visual-studio-code-recent-projects/package.json
+++ b/extensions/visual-studio-code-recent-projects/package.json
@@ -125,8 +125,8 @@
     {
       "name": "terminalApp",
       "type": "appPicker",
-      "default": "Terminal",
-      "required": true,
+      "default": "com.apple.Terminal",
+      "required": false,
       "title": "Terminal App",
       "description": "Select which Terminal App to use when opening with a terminal"
     }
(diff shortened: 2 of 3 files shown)


macbookandrew:
#9443 seemed to work fine without this, but I guess it doesn’t hurt ¯\\\_(ツ)_/¯

cy-98:
I think replace  `terminalApp.path` with `terminalApp?.path` is better .
TerminalApp maybe should keep not required.

tmwrnr:
> https://github.com/raycast/extensions/pull/9443 seemed to work fine without this, but I guess it doesn’t hurt ¯\_(ツ)_/¯

@macbookandrew  I didn't encounter any problems either, so I set it to "required" because I couldn't reproduce the issue.

> I think replace terminalApp.path with terminalApp?.path is better .
TerminalApp maybe should keep not required.

@cy-98 If the `terminalApp` is not set, I would have to hide the action. Therefore, the user needs to know about setting the preference to access this action. However, I agree with you that it is not ideal to ask the user about the preference.

I'm unsure which solution is best.

macbookandrew:
Can’t we set Terminal.app as the default and use that if the preference is not seT?

macbookandrew:
e.g., something like `terminalApp?.path ?? 'Terminal.app'` ?

macbookandrew:
> I thought by specifying a default in the preferences, it would always be there.

Yes, if you have opened the pref pane and saved it…I could be wrong, but I’m not sure that Raycast will merge in the default if you haven’t saved it yet 🤷🏻‍♂️

mathieudutour (approved):

tmwrnr:
@pernielsentikaer Thank you for looking at the issue. I also wanted to think about an alternative immediately after work.

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

Issue #9502, opened 2023-12-01
[The Movie Database] ability to copy the title id


georgesnow:
### Extension

https://www.raycast.com/Aayush9029/tmdb

### Description

After searching for a movie having an action to copy the item's tmdbs id. example

Ghostbusters URL:
https://www.themoviedb.org/movie/620-ghostbusters

The id for the movie is 620 contained in the URL

### Who will benefit from this feature?

When using JellyFin and Plex the id of the title allows for better linking for metadata. When adding the title's id to the filename.

georgesnow:
@j3lte awesome I just happen to notice when I was updating some movies this was added. thanks!

PR #9508, opened 2023-12-01
Update tmdb extension

j3lte:
## Description

- Feature request #9502 : Add the ability to directly copy the TMDB ID to your clipboard

## Checklist

### DIFF
--- a/extensions/tmdb/src/components/MovieDetail.tsx
+++ b/extensions/tmdb/src/components/MovieDetail.tsx
@@ -97,7 +97,7 @@ export default function MovieDetail({ movie }: { movie: MovieResponse }) {
           <Action.OpenInBrowser title="Open in TMDB" url={`https://www.themoviedb.org/movie/${movie.id ?? 0}`} />
           {movie.id ? (
             <Action.CopyToClipboard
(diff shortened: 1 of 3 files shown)


andreaselia:
Looks good to me, but do we really need to display the ID we are copying? Could the action simply be "Copy TMDB ID"? Just thinking if it's useful to show the ID here or better to keep it simple 🤔

Aayush9029:
LGTM

imo, It’s fine to display the user ID since the length of the string is small.

j3lte:
@andreaselia I was thinking of that, but the ID isn't shown anywhere, hence why I would add it in the Action menu. That way the user knows what it copied. It's only a few numbers, so it wouldn't mess up the Action panel. 

If the ID was displayed in the Details, I would remove it from being displayed in the Action title I guess, does that make sense?

Let me know if I should remove it from the title :)

andreaselia:
I agree, it doesn't provide additional value to the user seeing it rather than just copying it.

If we did want to display the ID somewhere, perhaps we could add it below "votes" here?

CleanShot 2023-12-03 at 22 18 04@2x

Issue #9546, opened 2023-12-05
[Library Genesis] ...


Saidelocha:
### Extension

https://www.raycast.com/yz3440/library-genesis

### Description

The search command is not working.
A typeerror appear, see screenshot.
<img width="848" alt="Capture d’écran 2023-12-05 à 17 53 23" src="https://github.com/raycast/extensions/assets/97402209/ee01f41d-4fbc-4b3b-88df-59990a2af033">

### Steps To Reproduce

Any search

j3lte:
Fixed in #9549, issue can be closed

j3lte:
@raycastbot close this issue

Issue #8803, opened 2023-10-14
[Library Genesis] Book search not working 


BitLegion:
### Extension

https://www.raycast.com/yz3440/library-genesis

### Description

It does not render a search for books on Library Genesis. Instead, I get the error "TypeError: p.replace is not a function"

### Steps To Reproduce

1. Open up Raycast
2. Navigate to the Library Genesis plugin 
3. Select the "Search with Books" option
4. Attempt to search with a book

jphastings:
This is a duplicate of #8355, there's a [quick-fix](https://github.com/raycast/extensions/issues/8355#issuecomment-1720927792) and an explanation in the (short!) thread.

j3lte:
Fixed in #9549, issue can be closed

j3lte:
@raycastbot close this issue

Issue #8778, opened 2023-10-12
[Library Genesis] TypeError: p.replace is not a function


timebooksio:
### Extension

https://www.raycast.com/yz3440/library-genesis

### Description

![CleanShot 2023-10-12 at 16 41 03@2x](https://github.com/raycast/extensions/assets/66457203/dc8ba571-111a-43f5-a110-83bb7947e57c)

### Steps To Reproduce

Searching any books name

jphastings:
This is a duplicate of #8355, there's a [quick-fix](https://github.com/raycast/extensions/issues/8355#issuecomment-1720927792) and an explanation in the (short!) thread.

j3lte:
Fixed in #9549, issue can be closed

j3lte:
@raycastbot close this issue

Issue #8355, opened 2023-09-12
[Library Genesis] Searching does not work


Mdejonge:
### Extension

https://www.raycast.com/yz3440/library-genesis

### Description

When searching for books an exception is thrown: `p.replace is not a function`

### Steps To Reproduce

1. Install the app for ray
2. Search for a book
3. Get's error `p.replace is not a function`

Mdejonge:
The problem is caused because i am unable to connect to http://gen.lib.rus.ec/. Maybe skip the url if an exception is received?

nununo:
I'm getting the same error.

Mdejonge:
@nununo A temporary solution is in the settings of the extension to add `https://libgen.is/` to the Preferred Libgen Mirror.

yz3440:
Hi folks! Thank you so much for reporting these bug. The truth is the mirror picking mechanism was not well thought out and pretty hacky. I'm working on a fix now, but I can't promise an quick updates given my recent schedule.

### Some Background

I was using a [npm package for fetching libgen url](https://www.npmjs.com/package/libgen) which doesn't really work. I'm planning to encode a few well-known libgen mirrors and ping them to choose the best mirror manually. While picking out the mirrors, I realized there is [this type of mirror](https://libgen.is/) and [that type of mirror](http://libgen.gs/), having different UIs. I want to include them both to maximize the chances of hitting a usable mirror, but my old html parser doesn't work on the latter one. I'll be tackling these problem one by one.

@Mdejonge Thank you for saying the what I was going to say! I should probably encode the instruction if none of the candidate mirrors are reachable.

Anyways, thank you all so much for using my extension! I hope you find it useful as I do. Happy reading!

PR #9549, opened 2023-12-05
Update library-genesis extension

j3lte:
## Description

- Removed [libgen](https://www.npmjs.com/package/libgen) dependency
  - The library is not maintained and had issues with old mirrors
- Removed `axios` because we can use `node-fetch` instead
- Updated all (dev) dependencies to latest version
- Added better code formatting and linting, cleaned up code
- Updated download controller, this is now cancellable
- Add Mirror List to the extension
  - This is a list of all mirrors that are currently available in the extension

This fixes the following issues:
- #9546  no results due to `libgen` dependency failing
- #8778 duplicate
- #8355 duplicate
- #8333 duplicate
- #8803 duplicate

## Checklist

### DIFF
--- a/extensions/library-genesis/CHANGELOG.md
+++ b/extensions/library-genesis/CHANGELOG.md
@@ -10,6 +10,8 @@
 - Updated download controller, this is now cancellable
 - Add Mirror List to the extension
   - This is a list of all mirrors that are currently available in the extension
+- Cached fastest mirror
+  - This is now cached for 1 hour, so we don't have to check for the fastest mirror every time
 
 ## [Feature Updates] - 2022-07-21
 
--- a/extensions/library-genesis/package-lock.json
+++ b/extensions/library-genesis/package-lock.json
@@ -11,7 +11,8 @@
         "@raycast/utils": "^1.10.1",
         "cheerio": "^1.0.0-rc.12",
         "fs-extra": "^11.2.0",
-        "node-fetch": "^3.2.8"
+        "node-fetch": "^3.2.8",
+        "swr": "^2.2.4"
       },
       "devDependencies": {
         "@raycast/eslint-config": "^1.0.6",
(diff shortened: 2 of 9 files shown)


yz3440:
First off, @j3lte thank you so much for taking care of the extension! I really appreciate it! 

I was thinking about fixing the mirror issue but I found that some mirrors have the "Library Genesis +" (eg. [this](https://libgen.gs/), as compared to [this](https://libgen.is/)) UI so the scraper I wrote would stop working. 

You seem like a much experienced TS developer than I was, so I'm curious to hear your thoughts.

j3lte:
@yz3440 thanks!

Yes, there is a list of mirrors that is using a different UI. I'm planning on implementing that one tomorrow, I already moved some of the code for parsing a page around.

The beauty of the other mirrors is that it also has a JSON Api, so I think we only have to parse the initial page in cheerio and take it from there.

I'll update the PR further tomorrow to see if we can include the 'new UI' mirrors. We could also publish this version of the extension (so we can fix the list with open issues) and make a new PR for the mirrors

yz3440:
Sounds great! Wow didn't know the JSON api existed! Is it documented somewhere? 
If you ended up making an update tmr, just ping me and I'll make sure to review it!

j3lte:
@yz3440 I checked the other API, but it seems it uses a bit of a different structure, as they are using with editions for example. The new UIs are not really compatible with the current extension. I would argue that we publish this update, as it fixes a list of issues right now, and see if we can add the new UI mirrors a later moment.
This update contains 3 mirrors that are all working fine

pernielsentikaer:
Sounds good to me, what do you think @yz3440

yz3440:
Yeah sounds good to me too. Thank you so much for all the work. Sorry for the late reply!

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #9567, opened 2023-12-06
[spotify-player] loads 300 instead of 50 albums, artists, playlists, shows

mads-hartmann:
## Description

This PR contains changes to the Your Library command provided by the Spotify Player extension.

Previously the extension would only load at most 50 playlists, albums, artists, tracks, shows, and episodes. That means if you have more than e.g. 50 albums in your library then you can't use "Your Library" to search for it.

This PR adds changes the limits so they're 300 for playlists, albums, artists, shows, and 50 for tracks and episodes. If I tried to load all my tracks the extension runs out of memory. The new limits are based on my personal needs 😉 

To accomplish this I added an async iterator that we can use paginate over the responses from Spotify.

If you have a large Spotify library then it takes a while longer to load the initial data. I want to add more user feedback while the data is being loaded but I'd prefer to do that in a follow-up PR ☺️ 

## Screencast

No screencast for this one.

## Checklist

### DIFF
--- a/extensions/spotify-player/CHANGELOG.md
+++ b/extensions/spotify-player/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Spotify Player Changelog
 
+## [Optimisation] - 2023-12-20
+
+- The `Your Library` command now loads 300 instead of 50 albums, artists, playlists, shows.
+
 ## [Feature & Optimisation] - 2023-12-02
 
 - Added `Select Device` command to select the device to play music on.
--- a/extensions/spotify-player/src/api/getFollowedArtists.ts
+++ b/extensions/spotify-player/src/api/getFollowedArtists.ts
@@ -1,17 +1,37 @@
 import { getErrorMessage } from "../helpers/getError";
 import { getSpotifyClient } from "../helpers/withSpotifyClient";
+import { ArtistObject } from "../helpers/spotify.api";
 
-type GetFollowedArtistsProps = { limit?: number };
+type GetFollowedArtistsProps = { limit: number };
 
-export async function getFollowedArtists({ limit = 50 }: GetFollowedArtistsProps = {}) {
-  const { spotifyClient } = getSpotifyClient();
-
+export async function getFollowedArtists({ limit }: GetFollowedArtistsProps) {
+  const iterator = iterate();
   try {
-    const response = await spotifyClient.getMeFollowing("artist", { limit });
-    return response.artists;
+    const artists: ArtistObject[] = [];
+    for await (const items of iterator) {
+      artists.push(...items);
+      if (artists.length >= limit) {
+        break;
+      }
+    }
+    return { items: artists };
   } catch (err) {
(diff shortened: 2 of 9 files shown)


mads-hartmann:
@peduarte I rebased on main to resolve conflicts - would you be able to review the PR? ☺️

mads-hartmann:
@peduarte Anything I can do to make the review easier? ☺️

peduarte:
So sorry for the delay. Am I right in thinking that this change will make 6 API calls (6 calls of 50 results) instead of 1? The problem is we are already having rate limiting issues with the Spotify API, so I'm quite hesitant on merging this as it'll just mean we call the API more often than we most likely need. 

The ideal solution for this would be to implement pagination.

The good news is that we're working on a pagination API right now, so as soon as it's out, we'll add it to the extension.

mads-hartmann:
> Am I right in thinking that this change will make 6 API calls (6 calls of 50 results) instead of 1?

It depends on how many albums, playlists, etc. you have. Before it would make 1 request per type (6 in total) which means it would only fetch at most 50 entries for each type.

This PR implements pagination for the Spotify API but then still has hardcoded caps for how many of each type to fetch. 

- Playlists, albums, artists, shows are capped at 300 so 6 requests for each at most.
- Tracks, Episodes are capped at 50 so 1 quest at most

So worst case you'd make 4*6+2*1 = 26 requests - not in parallel though.

> The good news is that we're working on a pagination API right now, so as soon as it's out, we'll add it to the extension.
Awesome 

---

We can close this PR and revisit once you have support for pagination and see if some of the changes in this PR still makes sense. I suspect you'll need the Spotify API pagination at least ☺️

peduarte:
Thanks for the reply.

With this implementation, it means that if I have 500 albums, it will call the API 6 times, even if I the artist I'm after is the first in the list, correct?

I think the main different with our upcoming one, is that it will only make further requests when needed, ie: user has reached the bottom of the list. 

I'm also curious about why Tracks, Episodes are capped at 50.

mads-hartmann:
>With this implementation, it means that if I have 500 albums, it will call the API 6 times, even if I the artist I'm after is the first in the list, correct?

Yeah - conceptually it's creating a local cache of your Spotify data

>I'm also curious about why Tracks, Episodes are capped at 50.

With the extensions current way of caching the results the extension runs out of memory pretty quickly. Given the cap was 50 before I decided to just not change it ☺️

---

In general the way the extension deals with caching of the results (both in this PR and on main) means it runs out of memory if it wanted to cache my entire library. Pagination at the Raycast level + context-specific-search (e.g. favourite albums) would be a nice solution to this ☺️

peduarte:
Yeah as part of the pagination work that @sxn is working on, he's also dealing with memory-related issues! 

Thanks for putting this together, we may use it as reference when we need to implement the new Pagination API, but for now I think it's best for us to wait for the first-class API. 

Thanks again!

PR #9575, opened 2023-12-06
Add youtube-subscriber-celebrator extension

imaNNeo:
## Checklist

### DIFF
--- a/extensions/youtube-subscriber-celebrator/README.md
+++ b/extensions/youtube-subscriber-celebrator/README.md
@@ -1,6 +1,6 @@
 # YouTube Subscriber Celebrator - Raycast Extension
 
-<img src="https://github.com/imaNNeo/raycast_extensions/blob/main/youtube-subscriber-celebrator/assets/command-icon.png?raw=true" alt="YouTube Subscriber Celebrator" width="200"/>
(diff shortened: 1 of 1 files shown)


pernielsentikaer:
Hi @imaNNeo 👋 

Even though I like the idea, it might not be the best to publish.

When running confetti, it steals focus from the fields you're currently working in, which will be really annoying for the user.

I tested it by running `for i in {1..5}; do open raycast://confetti && sleep 2; done` in the terminal

imaNNeo:
Hi there,
Yes, you're right, it distracts you if you get more than an amount per day,
This config (checking every minute and showing confetti every time the subscribers increase) works well for me because I just started my YouTube channel and have no more than 3 subscribers per day.
But I agree it can distract people who have famous YouTube channels

What if I add some configs for `checkInterval` and `thresholdCount`?
In this case, for example you can set it to check every one hour with the threshold of 20 subscribers

pernielsentikaer:
That sounds like a plan. What about changing the name to "YouTube Subscriber Count" and adding the required preferences for the confetti and the threshold?

Then, initially, when people install it, they can select their preferences.

imaNNeo:
Sounds good, I will close this PR and will create a new one with the better name
`YouTube Subscriber Count`

PR #9682, opened 2023-12-13

### LINKED ISSUE #9649: [Spotify Player] does not launch Spotify when asked to play
### Extension

https://www.raycast.com/mattisssa/spotify-player

### Description

Before when I typed play into raycast and clicked the command by your extension it opened Spotify and started playing. This does not work now with any function.

### Steps To Reproduce

1. Open Raycast with extension installed
2. type play and press enter
3. Does not do anything but get the playing notification

### Current Behaviour

_No response_

### Expected Behaviour

_No response_

jatindotdev:
## Description

The pull request fixes #9649 by removing the incorrect script. Previously, the script used to check for Spotify's presence relied on whether Spotify was actively running, rather than confirming its installation. 

Now it launches Spotify and show better error messages if not installed

## Screencast

https://github.com/raycast/extensions/assets/59236972/813f0a67-d8cb-48f1-9fe3-9f38933b5963

## Checklist

### DIFF
--- a/extensions/spotify-player/CHANGELOG.md
+++ b/extensions/spotify-player/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Spotify Player Changelog
 
+## [Fix] - 2024-01-31
+
+- Fixed a bug that caused launching the Spotify app to not work properly
+- Modified the device selection preference for playback
+
 ## [Improvement] - 2024-01-30
 
 - Increase the stale data time from 10 minutes to 2 hours
--- a/extensions/spotify-player/src/api/play.ts
+++ b/extensions/spotify-player/src/api/play.ts
@@ -1,9 +1,9 @@
 import { runAppleScript } from "@raycast/utils";
 import { buildScriptEnsuringSpotifyIsRunning } from "../helpers/applescript";
 import { getErrorMessage } from "../helpers/getError";
+import { checkSpotifyApp } from "../helpers/isSpotifyInstalled";
 import { getSpotifyClient } from "../helpers/withSpotifyClient";
 import { getMyDevices } from "./getMyDevices";
-import { checkSpotifyApp } from "../helpers/isSpotifyInstalled";
 
 type ContextTypes = "album" | "artist" | "playlist" | "track" | "show" | "episode";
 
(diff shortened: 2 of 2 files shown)


jatindotdev:
@peduarte can you take a look at it once this will resolve many issues getting raised

jatindotdev:
@peduarte Let me know if I can help you on getting this PI reviewed

peduarte:
Im not sure this addresses the issue fully. Read this [comment for more info](https://github.com/raycast/extensions/issues/9649#issuecomment-1917118251). 

But this highlights something that we could improve
- `checkIfSpotifyIsInstalled` and `checkSpotifyApp` look like they're supposed to do the same thing, but are yielding different results. This should probably be fixed and we should be able to rely on a single function throughout the codebase
- Deleting `checkIfSpotifyIsInstalled` causes an error as it's used somewhere else in the codebase

I like that your PR improved the error message, and I think we could use this PR to fix the notes above, but in my opinion it's not the adequate fix for #9649

peduarte:
Thanks for the reply!

> It was only used in one place afair and i have replaced it

Yep, you're right, my bad!

So this PR fixes the edge case with the check, that's good, we can keep it. 

---

Right now, we check for devices, and if there are no active devices, we pick the first one.

`play.ts#33`

```tsx
// If there is an active device, use that. Otherwise, use the first device.
const activeDevice = devices?.find((device) => device.is_active);
const deviceId = activeDevice?.id ?? devices?.[0]?.id ?? undefined;
```

However, Im not sure this is the right heuristic. In my case, I have an active device which is my "Spotify Connect"'ed speaker at home. So when I press play, I see the successful toast that something's playing, but I hear nothing. Because it's playing from my speaker at home and Im in my office. 

Does this make sense?

So what I am thinking here, is that before relying on `devices?.[0]`, we should rely on the Spotify App installed on the Mac, and only if there is no Spotify App installed, fallback to the first available device. 

Thoughts?

jatindotdev:
> So what I am thinking here, is that before relying on `devices?.[0]`, we should rely on the Spotify App installed on the Mac, and only if there is no Spotify App installed, fallback to the first available device.
> 
> Thoughts?

Yeah right, this could be annoying sometimes.

1. If there's no active device and we have Spotify app installed play it from Spotify app
2. If Spotify app is not installed, go for the first available device

I hope i didn’t miss anything here, anything else would you like me to do in this pr

Would you like me to implement this?

jatindotdev:
@peduarte i have updated the implementation to what we discussed above

do let me know if there is anything else

peduarte:
Thanks so much, lets take it for a spin!

peduarte (approved):

PR #9827, opened 2023-12-22
Add vscode-open extension

0reoo:
## Checklist


pernielsentikaer (commented):
Hi 👋

It seems like there is a lot of test data in this extension, could you look into removing that?

At the moment, Raycast doesn't support localization and only supports US English. Therefore, please avoid introducing your custom way to localize your extension.

What's the purpose of this extension, I think we have multiple extensions already doing this 🙂

PR #9875, opened 2023-12-26
Add snap-jot extension

k41531:
## Description
Customizable and very simple memos with a timestamp.
You can specify the directory and file name format.
If there is no file, create a file with the template written in it.

## Screencast

snap-jot-1
snap-jot-2

## Checklist


pernielsentikaer (commented):

We have something similar in the [store](https://www.raycast.com/frugoman/my-daily-log), could that work for you or is this for something specific?

k41531:
@pernielsentikaer 
Thank you for your response.
While there are several extensions for note-taking, many of them integrate with existing apps. I wanted to create something that allows for simple and flexible local note storage, which is why I made this.

k41531:
@pernielsentikaer
Yes, the extension linked can view logs and record locally on Raycast, but it saves in json format and is optimized for viewing on Raycast, which doesn't meet my needs.

My extension is designed to append to local files in markdown format. In the future, I want to expand it to not just be limited to Markdown, but to allow saving in any format.

Currently, I use it to append to my daily notes. I write these in Markdown format using a text editor, and I've set it up so they can also be appended from Raycast.

pernielsentikaer (changes requested):
- I find the initial preferences confusing
  - Set default values so the user doesn't have to do it
  - Instead of the time format, have a 12/24 setting to remove that input field
  - Use `tooltip` to guide the user

- Missing a `toast` when saving an entry (and maybe `popToRoot()` too)

k41531:
@pernielsentikaer
Thank you for your review!
I have made adjustments to the initial preferences:

- The only required settings now are the saving folder and a dropdown for the 12/24 hour setting.
- Optional settings such as the template file, filename format, and prefix format can be configured in the extension settings if needed.

Regarding the tooltip, the file type and directory type fields in the initial preferences don't support displaying tooltips, so I've added a brief description in the title instead.

I have addressed this issue as well.
>Missing a toast when saving an entry (and maybe popToRoot() too)

pernielsentikaer:
I think we're about to be ready, could you modify the `README` too 🙂

k41531:
@pernielsentikaer 
I have updated the README.
Please check it out.

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #9899, opened 2023-12-28
Add osint-web-check extension

gksander:
## Description

This PR adds a `osint-web-check` extension (Open Source Intelligence) motivated by [this repo](https://github.com/lissy93/web-check) that takes a URL and runs various OSINT checks against the URL (DNS, SSL, DNSSEC, HTTP Headers, etc.).

## Screencast

| | |
| --- | --- |
| Input | Example (WHOIS) |
|  |   |
| Example (robots.txt) | Info view |
|  |  |

## Checklist


pernielsentikaer (changes requested):
Hi 👋 

- You can add `tintColor` to your icons like above, that could maybe give some extra attention?
osint-web-check 2024-01-09 at 12 06 25

- You could use navigationTitle to append the domain, you're allowed to do that since you push to a new view 🙂

gksander:
Thank you for the feedback @pernielsentikaer! Added in your suggestions with `tintColor` and `navigationTitle`, looks much better 🎉 

image

pernielsentikaer:
You could also add an argument so you don't need the form maybe? Something like this

Raycast09012024-FVyo05UE png@2x

Not sure if you like that?

gksander:
Oh neat. Is there any way to do validation with arguments? Right now with the form, I'm doing a DNS `lookup` to ensure the input is a valid url to do a check against – wondering if there's any way to do custom validation with arguments.

pernielsentikaer:
No that's not possible, but it would be cool - You could use the argument and only show the form if the validation fails 🙂

pernielsentikaer:
Might be a bit too much for v1?

gksander:
Oh nice, I like that idea! Updated to do just that.

https://github.com/raycast/extensions/assets/12721310/0543d25c-6424-43e1-a1a9-accb2ee118e1

extensions/osint-web-check/src/osint-web-check.tsx
```diff
@@ -0,0 +1,72 @@
...
+    return <Detail markdown={`## Validating ${url}...`} isLoading={true} />;
+  }
+
+  if (isUrlValid) {
+    return <CheckDetails url={url} />;
+  }
+
+  return (
+    <UrlInput
+      onSubmit={(url) => {
```
pernielsentikaer, line 38:
```suggestion
      url={url}
      onSubmit={(url) => {
```

extensions/osint-web-check/src/UrlInput.tsx
```diff
@@ -0,0 +1,49 @@
+import { Action, ActionPanel, Form } from "@raycast/api";
+import { useState } from "react";
+import { addHttps } from "./utils/addHttps";
+import { checkUrl } from "./utils/checkUrl";
+
+type UrlInputValues = { url: string };
+
+type UrlInputProps = {
+  onSubmit: (url: string) => void;
+};
```
pernielsentikaer, line 10:
```suggestion
type UrlInputProps = {
  onSubmit: (url: string) => void;
  url: string;
};
```

extensions/osint-web-check/src/UrlInput.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+export function UrlInput({ onSubmit }: UrlInputProps) {
+  const [urlError, setUrlError] = useState<string | undefined>(undefined);
```
pernielsentikaer, line 13:
```suggestion
  const [urlError, setUrlError] = useState<string | undefined>("Enter a valid http(s) URL.");
```

extensions/osint-web-check/src/UrlInput.tsx
```diff
@@ -0,0 +1,49 @@
...
+            }}
+          />
+        </ActionPanel>
+      }
+    >
+      <Form.Description text="Enter a URL to view OSINT information (DNS, SSL, etc)." />
+      <Form.TextField
+        id="url"
+        title="URL"
+        placeholder="google.com"
```
pernielsentikaer, line 41:
```suggestion
        placeholder={url}
```

pernielsentikaer (changes requested):
That's a really awesome update 🔥 

Do you think something like suggested (or check `useForm`) would make sense to add?

extensions/osint-web-check/src/UrlInput.tsx
gksander: In this case, `url` will be a *bad* url, so we wouldn't want that as a placeholder. I'm instead using it as an initial value, and opting for a controlled form here – so upon entry here, we'll show the "bad" url they just entered, along with the error message/state.

gksander:
Okay, took the heart of your suggestions and incorporated them in! Of note, changed to a controlled form for `UrlInput` so we can pass the "bad" url they initially entered in as an initial value and have an error state.

https://github.com/raycast/extensions/assets/12721310/92781aee-97d1-4060-b5a5-bbb75e6accb0

gksander:
Think I'm going to push up one more commit in a little bit to get rid of the loading Detail view when the initial DNS lookup is happening. I think the List view can just pause for a short bit while we do the initial validation, feels like it'd be less jank and a smoother UX.

gksander:
Okay, pushed up another commit that cleans up a lot of the repeated code, and assumes user will take "the happy path" and enter a valid URL, so we don't need an intermediate step. If user enters invalid url, we'll then kick them back to the manual form with the error notice.

https://github.com/raycast/extensions/assets/12721310/9bf60905-1610-493e-8ac1-4fefa5686ea3

pernielsentikaer:
What a huge difference it makes! I really love the flow now. Awesome work! 🚀

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #9937, opened 2023-12-31
[Toggl Track] Refactor extension.

bkeys818:
## Description

The Toggle Track extension was written before utilities like `useCachedPromise` were available. The extension currently implements a custom method of caching and refreshing data. The method works perfectly for the extension in its current state, but I believe future changes will be easier if refactoring is done first.

### Proposed Changes

- [x] Update dependencies. (4e318eeb4f1cf7a6cc01d3a634f95f710931c222)
- [x] Rename the `toggle` directory to `api`. (a6cd0e8e1b9caf5b725d1b6d25d23189561ed472)
  - To be more consistent with other directories.
- [x] Move each API endpoint to its own file. (6ed66ebc87ee81e12a82c4b42aba8836b03cb588)
  - It will be more evident what methods rely on what other methods.
  - Code will be easier to navigate and read.
- [x] Replace uses of `persistedStorage` with `useCachedPromise`. (8f18e30e4329148d2475f67da32972adac857622)
  - Data will be refreshed every time the method is called, eliminating the need for a refresh action.
  - Any data that needs to be actively refreshed while the view is showing can be done so with `revalidate`.
- [x] Replace `AppContext` with local calls of custom hooks. (8f18e30e4329148d2475f67da32972adac857622)
  - Future commands can use the methods (with the same cached data) without having to fetch the data for all other commands.
  - The structure will look similar to that of the [Spotify Player extension](https://github.com/raycast/extensions/tree/main/extensions/spotify-player).

## Checklist

### DIFF
--- a/extensions/toggl-track/.eslintrc.json
+++ b/extensions/toggl-track/.eslintrc.json
@@ -1,10 +1,4 @@
 {
   "root": true,
-  "env": {
-    "es2020": true,
-    "node": true
-  },
-  "parser": "@typescript-eslint/parser",
-  "plugins": ["@typescript-eslint"],
-  "extends": ["eslint:recommended", "plugin:@typescript-eslint/recommended", "prettier"]
+  "extends": ["@raycast"]
 }
--- a/extensions/toggl-track/package-lock.json
+++ b/extensions/toggl-track/package-lock.json
@@ -13,12 +13,10 @@
         "node-fetch": "^3.3.2"
       },
       "devDependencies": {
+        "@raycast/eslint-config": "^1.0.8",
         "@types/node": "~20.8.10",
         "@types/react": "~18.2.27",
-        "@typescript-eslint/eslint-plugin": "^6.16.0",
-        "@typescript-eslint/parser": "^6.16.0",
         "eslint": "^8.56.0",
-        "eslint-config-prettier": "^9.1.0",
         "react-devtools": "~4.28.4",
         "typescript": "^5.3.3"
       }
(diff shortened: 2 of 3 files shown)


bkeys818:
@franzwilhelm, since you wrote the original extension, I'd love your opinion on this.

bkeys818:
So far, the only roadblock I see ahead is how to validate the API token throughout all commands without using `AppContext`.

extensions/toggl-track/src/api/index.ts
```diff
@@ -0,0 +1,7 @@
+export * from "./me";
+export * from "./workspaces";
+export * from "./projects";
+export * from "./clients";
+export * from "./tags";
+export * from "./tasks";
+export * from "./timeEntries";
```
franzwilhelm, line 7: I would change this to ensure better intellisense. import * as statements are harder to autocomplete, as the editor will default to finding named exports. Instead of this, we could use this file to export an object containing all methods like:

```ts
const api = {
  me,
  projects
  //... etc
}

export default api;

/*
Usage:

import api from './api';

api.projects.getWorkspaceProjects();
*/
```

Or we could `...projects` etc in the object, essentially giving us the same exports, but without the need for `import * as`

The issue with not having export * as, is that we're missing types. But anyways. I'll leave this one up to you. Just a personal preference

extensions/toggl-track/src/api/toggleClient.ts
```diff
@@ -0,0 +1,49 @@
+import fetch from "node-fetch";
```
franzwilhelm, line 1: Feel free to rename to `togglClient` :D

franzwilhelm (commented):
Hey thanks a lot @bkeys818 ! Great work here, and you're right. The caching was a fun little project I did before the implementation came out from the raycast team ;) I got a couple comments so far, nothing breaking tbh, everything LGTM 🚀 

Feel free to re-request review when done :D

extensions/toggl-track/src/api/index.ts
bkeys818: What editor do you use?

franzwilhelm: @bkeys818 vscode, and I'm used to imports automatically appearing as I type. With `import * as` that doesn't happen, and you have to manually type the import statement. From there I have autocomplete

bkeys818: Would this work?

```ts
export { getMe, Me } from './me'
```

franzwilhelm: Yup ! I think that's the best solution so far

bkeys818:
## Issues

- [ ] The data, especially `projects` and `clients`, is refreshing more than expected.
  - This could have something to do with the issue below.
- [ ] The cached values for `projects` and `clients` are getting mixed up.
  - I'm not sure where or why this is happening.
  - I used [`useTraceUpdate`](https://stackoverflow.com/a/51082563) to track state.

@franzwilhelm could you help me out?

franzwilhelm:
@bkeys818 sorry for the late response. I suspect the issue might be related to the allWorkspacesFetch utility function. The cache of raycast is a bit magical IMO, and probably uses the function signature to determine how to key the cache layers. So: if you try to do a `useMemo` on all your `allWorkspacesFetch` functions I think we should get unique caches, and avoid overfetching. I don't have time atm, but try it out and report back if it works or not :D Looking forward to it!

bkeys818:
@franzwilhelm, I started looking for the culprit. I think the issue is actually that I'm using the cached values as dependencies for `useEffect`.

Since `useEffect` checks for dependencies updates with `Object.is()`, [this effect](https://github.com/bkeys818/raycast-extensions/blob/35f08757838b086ffb1d84dde30370f286a2a083/extensions/toggl-track/src/context/TimeEntryContext.tsx#L73) is being run when `workspaces` is updated from the cached object to the fetched object, even when the value is the same.

bkeys818:
@franzwilhelm, you're spot on. I should've read your suggestion more closely. I spent forever looking for the culprit and ended up finding a few other bugs, but that was the main issue in the end.

bkeys818:
Neither `useMemo` nor `useCallback` are working, so I'm going to look for some help on the Slack channel.

franzwilhelm:
Ey @bkeys818 did you manage to get it working?

bkeys818:
@franzwilhelm Yup!

extensions/toggl-track/src/helpers/allWorkspacesFetch.ts
```diff
@@ -1,13 +1,14 @@
...
+  workSpaceFunction: (workspaceId: number) => Promise<T[] | null>,
+  workspaces: Workspace[],
+) {
+  const data = await Promise.all(
+    workspaces.map(async (workspace) => {
+      const workspaceData = await workSpaceFunction(workspace.id);
+      return workspaceData || [];
+    }),
+  );
+  return data.flat();
```
franzwilhelm, line 13: So not returning a function was the fix in combination with moving the function outside of the hook?

franzwilhelm: Great work man! This makes it easier to contribute with more features

bkeys818: It was actually just moving the function outside of the hook. Return a function here was just unnecessary.

franzwilhelm (approved):
@pernielsentikaer This one is good to go! 🚀

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #9949, opened 2024-01-01
Add crawldoc extension

johan-perso:
## Screencast

screen_1
screen_2

## Checklist


pernielsentikaer:
I can see that it could be handy to have a collection of docs (like [DocSearch](https://www.raycast.com/Fatpandac/docsearch)).

I have no clue about what I can search for and how precise I need to be. I'm not sure if this is the best approach.

Maybe DocSearch does a better job in general?

johan-perso:
DocSearch may do a better job on some search, my service CrawlDoc is made to have his own index of results to avoir depending on externals services. The algorithm is not quite so good at the actual time, it is better to search for page title or elements inside it. A good query would be for example "tailwindcss font size".

pernielsentikaer:
I would search for `tailwind` which would results in zero elements

johan-perso:
Thanks for issuing this, it will be fixed in the next update of the backend. For the moment, you can try using `tailwindcss` as the first word for your query (instead of `tailwind` which is currently not recognized).

pernielsentikaer:
That's exactly my point. The user doesn't have a clue about what to search for, which makes it a really bad user experience.

johan-perso:
When the service isn't recognized, the plugin tells the user that he need to firsly type the name of any service, then it will be easier for him when he know that

pernielsentikaer:
We need a better way to handle this, I tried to search for `vue`, `vuejs` and `vue3` and I was not able to find anything

There is something wrong with the UX we need to figure a way to deal with, otherwise users will get stuck almost before using it

johan-perso:
Well, the service is kinda new and by trying to have his own index it lacks of some services like vuejs. Currently it supports 17 documentations like Electron, Vite, Visual Studio Code, Spotify.

johan-perso:
> I would search for `tailwind` which would results in zero elements

The latest commit add support for sites aliases, making it easier to lowering the chance of typing a wrong name.

pernielsentikaer:
What about adding a dropdown in the upper right corner with the options instead? It feels like shooting in the dark to determine which services are supported or not. As a new user, I have no clue about what I can do, and it really depends on the wording in the end to find a match or not. Maybe we can assist end users this way 🙂

johan-perso:
I didn't thinked about this but it could be better for user experience, something like this?

image

pernielsentikaer:
YES thats much better 👏 Now I'm actually able to use it and know what I can search in 🙏 

You're missing a lot of the default files from when you start a new extension, did you make this a long time ago? Can you try to move the root files from `Create Extension` manually 🙂

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #9981, opened 2024-01-04
Add macos-utilities extension

kciarnie:
## Description

A macOS utilities extension that allows the user to clear/restore their history, search, send a ray.so link, copy the command. You can also Toggle your dock.

This is the only macOS utility right now. But, I plan to add
- Cron jobs creator/edit/deletor
- Clamshell mode toggler
- URLs to ts models

That's why the labeling is macOS Utilities.

## Screencast

https://github.com/raycast/extensions/assets/6669089/100b0d98-a361-4c95-b931-fe4e3a7fad36

## Checklist

### DIFF
--- a/extensions/macos-utilities/package.json
+++ b/extensions/macos-utilities/package.json
@@ -13,6 +13,12 @@
   "license": "MIT",
   "commands": [
     {
+      "name": "show-hide-dock",
+      "title": "Toggle Dock",
+      "description": "Show or hide the dock",
+      "mode": "no-view"
+    },
+    {
       "name": "terminal-history",
       "title": "Search History",
       "subtitle": "Terminal",
--- a/extensions/macos-utilities/src/show-hide-dock.tsx
+++ b/extensions/macos-utilities/src/show-hide-dock.tsx
@@ -1,0 +1,54 @@
+import { runAppleScript, showFailureToast } from "@raycast/utils";
+import { showHUD } from "@raycast/api";
+import { execSync } from "child_process";
+
+/// The value for the Show dock mode enabled
+const IS_DOCK_SHOWN = "1";
+
+/// A class to handle the showing dock mode
+class ShowDockMode {
+  /// Gets the state of the show dock mode
+  /// @returns the state of the clamshell mode
+  getState(): boolean | undefined {
+    try {
+      const result = execSync(`defaults read com.apple.dock "autohide"`);
+      const value = result.toString().trim().at(-1);
+      return value == IS_DOCK_SHOWN;
+    } catch (error) {
+      showFailureToast(error, {
+        title: "Could not determine the state of the dock",
+      });
+    }
+  }
+
+  async setState(state: boolean): Promise<void> {
+    try {
+      const result = state ? "true" : "false";
+      const message = `✅  ${state ? "Hiding" : "Showing"} the dock`;
+
+      await runAppleScript(`tell application "System Events" to set the autohide of the dock preferences to ${result}`);
+      await showHUD(message);
+    } catch (error) {
+      if (error instanceof Error) {
+        if (!error.message.includes("-128")) {
+          showFailureToast(error, { title: "Could not Toggle Dock" });
+        }
+      }
+    }
+  }
+
+  /// Toggles the dock mode
+  /// @returns a promise that resolves when the state is toggled
+  async toggle(): Promise<void> {
+    const state = this.getState();
+    await this.setState(!state);
+  }
+}
+
+/// The ShowDock Mode
+export const mode = new ShowDockMode();
+
+/// Toggles the clamshell mode
+export default async function main() {
+  await mode.toggle();
+}


extensions/macos-utilities/src/utils.tsx
```diff
@@ -0,0 +1,76 @@
...
+      // ZSH just places one command on each line
+      return line;
+    });
+}
+
+/// Get the path to the shell history file
+/// - Returns: The path to the shell history file
+export function shellHistoryPath() {
+  const zshHistoryPath = path.join(os.homedir(), ".zsh_history");
+  const bashHistoryPath = path.join(os.homedir(), ".bash_history");
```
danpalmer, line 33: Not sure if you've considered this, but instead of reading the files and needing explicit support for different shells, could you instead use `useExec` to run the `history` command and parse the output. The `history` command should be available in almost any major shell, and `useExec` can use the user's login shell I believe, which would then respect the configured shell.

extensions/macos-utilities/src/utils.tsx
```diff
@@ -0,0 +1,76 @@
...
+  execSync(`cp ${path} ${backupLocation}`);
+}
+/// Create a ray.so link
+/// - Parameter command: The command to encode
+/// - Returns: A ray.so link
+export function createRaySoLink(command: string): string {
+  try {
+    // eslint-disable-next-line no-undef
+    const decodedString = btoa(command);
+    return `https://ray.so/#language=shell&code=${decodedString}&padding=16&title=Terminal+History+Command&darkMode=true&background=true&theme=raindrop`;
```
danpalmer, line 71: Instead of pre-configuring this, why not leave off all the configuration to use the ray.so defaults?

extensions/macos-utilities/src/utils.tsx
```diff
@@ -0,0 +1,76 @@
...
+    });
+}
+
+/// Get the path to the shell history file
+/// - Returns: The path to the shell history file
+export function shellHistoryPath() {
+  const zshHistoryPath = path.join(os.homedir(), ".zsh_history");
+  const bashHistoryPath = path.join(os.homedir(), ".bash_history");
+
+  if (fs.existsSync(zshHistoryPath)) {
```
danpalmer, line 35: The core of this plugin seems to all be synchronous. Given that history files could be significant in size, would it be worth using the async versions of all these functions and using `usePromise` at the command level?

extensions/macos-utilities/package.json
```diff
@@ -0,0 +1,42 @@
...
+    "Applications",
+    "Developer Tools"
+  ],
+  "license": "MIT",
+  "commands": [
+    {
+      "name": "terminal-history",
+      "title": "Search Terminal History",
+      "description": "Search your terminal history, copy and paste the command into your terminal",
+      "mode": "view"
```
danpalmer, line 19: Just my opinion, but terminal history management (great idea!) doesn't seem particularly macOS specific, for a "macOS Utilities" plugin.

Conversely, future features like toggling clamshell mode absolutely are macOS utility style tools, and I could see users wanting that functionality and not caring about terminal history.

Do you think this is perhaps two separate extensions? One for terminal history manipulation, and one for macOS utilities/tweaks/etc? They don't feel particularly related to me, and both feel like very useful extensions in their own right!

extensions/macos-utilities/src/terminal-history.tsx
```diff
@@ -0,0 +1,133 @@
...
+let history: string[] = [];
+
+export default function Command() {
+  const [loading, setLoading] = useState(true);
+
+  useEffect(() => {
+    setLoading(false);
+  }, [loading]);
+
+  history = getHistory();
```
danpalmer, line 16: I don't think this does quite what might be intended.

Firstly, the `loading` variable isn't hooked up to the list's `isLoading` prop, so I don't _think_ it's actually doing anything.

Secondly though, because `getHistory` is synchronous, it'll run on every render of this component (which is typically very often). The loading state will start in `true` for the first render. The first render will then kick off the effect, which will set it to `false`, which will _coincidentally_ only happen after the history has been read, and then the component will be re-rendered with the `false` state. This will then reload the history again, but without the loading state.

I'd suggest a) investigating `usePromise` and the `isLoading` parameter from that, and b) making `getHistory` async so that it doesn't block the JS execution thread and lock up Raycast. These two things should improve the responsiveness, ensure the history is only loaded once rather than the current two or more times, and also ensure the loading state is actually correct.

extensions/macos-utilities/src/utils.tsx
danpalmer: The `history` command also supports clearing in at least Fish and Bash.

It seems that `history` is implemented on a per-shell basis, so there might be some different support needed here, but using it rather than files would at least reduce the need to support custom configuration locations, and between bash/zsh/fish you'd probably have almost all setups covered.

extensions/macos-utilities/src/terminal-history.tsx
```diff
@@ -0,0 +1,133 @@
...
+
+  const length = history.length;
+
+  if (length === 0) {
+    return (
+      <List>
+        <List.EmptyView
+          icon={{ source: "https://placekitten.com/500/500" }}
```
danpalmer, line 24: Nit: I think the kitten is a placeholder from the examples, not actually intended for production. How about using `icon={Icon.Terminal}` instead?

https://developers.raycast.com/api-reference/user-interface/icons-and-images

danpalmer (commented):
Hey, not a Raycast team member, but saw your extension in the PRs and thought I'd leave a few thoughts, I hope they help.

Nice work, hope to see it ship soon!

extensions/macos-utilities/src/utils.tsx
kciarnie: https://github.com/raycast/extensions/pull/9981/commits/6142c027e0c299f47f6554f6615583060a1b8628

Thanks for the input, I have decided to change it to the default

extensions/macos-utilities/src/utils.tsx
kciarnie: Thanks for the feedback here. Yes, the core was syncronous. I've moved this extension towards an asyncronous approach in https://github.com/raycast/extensions/pull/9981/commits/6142c027e0c299f47f6554f6615583060a1b8628

extensions/macos-utilities/src/terminal-history.tsx
kciarnie: i went with the approach of using `useState` variables that way we don't have an overly abundant usecase of doing the syncronous approach of `getHistory`.

extensions/macos-utilities/src/terminal-history.tsx
kciarnie: https://github.com/raycast/extensions/pull/9981/commits/6142c027e0c299f47f6554f6615583060a1b8628

Thanks for the suggestion, I have updated the icon to the following as suggested.

extensions/macos-utilities/src/utils.tsx
kciarnie: That was my first though as far as parsing out the history command, however, history acts on an non-interactive shell and therefore, I was unable to ever get any results from this approach. My best way to do this was to obtain which shell it was with the `process.env.SHELL` variable and then from there that can determine which file I grab. I will not be able to use the `useExec` command. Unless you've found a way, I wasn't able to.

I took a different approach and looked into the `shell-history` module and that uses the same approach as I have. If you have a solution for this, I am more than willing to adopt it.

extensions/macos-utilities/package.json
kciarnie: Right now, I am planning on adding those other tools. I could easily keep this as one extension, but maybe @pernielsentikaer  can chime in here.

The thing is that I want to make sure it gets proper exposure in the right fields. I'm totally indifferent to where the extension goes.

pernielsentikaer: I might agree with @danpalmer since we already have [Shell](https://www.raycast.com/asubbotin/shell) in the Store which also reads the history, maybe it would be better to try an enhancement of that instead?

kciarnie: Shall I still keep this extension here and remove the shell-history portion from here and add in the show/hide dock, toggle clamshell mode and other macOS specific items?

pernielsentikaer: It's always hard to figure out what the best approach is since we also have [Dock Tinker](https://www.raycast.com/koinzhang/dock-tinker) and [Dock](https://www.raycast.com/pradeepb28/dock) in the store

kciarnie: Ahh okay, yeah, I noticed a bug with the show/hide toggle dock in `Dock Tinker`. In this case, I can update the `Dock` extension.

pernielsentikaer:
Closing as requested 🙂

PR #9995, opened 2024-01-04
Update punto extension

archibasov:
## Checklist


extensions/punto/src/Dict.ts
```diff
@@ -25,7 +25,7 @@ export const ru_en = new Map([
   ["№", "#"],
   ["%", "$"],
   [":", "%"],
-  [",", "^"],
+  [",", "б"],
```
red-avtovo, line 28: The former value was correct

extensions/punto/src/Dict.ts
```diff
@@ -161,7 +161,7 @@ export const en_ru = new Map([
   ["b", "и"],
   ["n", "т"],
   ["m", "ь"],
-  [",", "б"],
+  [",", "?"],
```
red-avtovo, line 164: Same here

extensions/punto/src/index.ts
```diff
@@ -66,11 +73,11 @@ function detectLayout(input: string): string {
 function switchCharacterLayout(char: string): string {
   if (en_ru.has(char)) {
     // console.log(char + " detected in en dict")
-    switchLayout("Russian");
```
red-avtovo, line 69: I believe here should be something like `switchLayout(layoutRu.value);`

extensions/punto/src/index.ts
```diff
@@ -66,11 +73,11 @@ function detectLayout(input: string): string {
...
+    // switchLayout("Russian");
     return en_ru.get(char) ?? char;
   } else {
     // console.log(char + " is probably detected in ru dict")
-    switchLayout("ABC");
```
red-avtovo, line 73: And here `layoutEn.value`

extensions/punto/src/index.ts
```diff
@@ -66,11 +73,11 @@ function detectLayout(input: string): string {
...
+    // switchLayout("ABC");
     return ru_en.get(char) ?? char;
```
red-avtovo, line 81: If you want to support an additional layout, you need to create a separate mapping

red-avtovo (changes requested):

extensions/punto/src/index.ts
archibasov: This is a duplicate switch, the main one is already in line 34.

extensions/punto/src/Dict.ts
archibasov: Yes, perhaps should add setting for a different keyboard.

extensions/punto/src/index.ts
red-avtovo: Good call

red-avtovo:
tbh, I would rather create multiple static maps and configure switching between then instead of one dynamic generation of a required map. That would simplify the setup and make it easy to extend

PR #9997, opened 2024-01-05
Update Splitwise extension: fix "Add Expense" command

### LINKED ISSUE #9880: (Splitwise) 404 error when adding expenses
### Description

I've connected splitwise to my raycast following the instructions. I can see and navigate through my expenses just fine but when I try to add a new expenses, it always leads to a 404 error.

Could you kindly advise? Thanks!

### Steps To Reproduce

_No response_

### Current Behaviour

404 error when trying to add a new expense via Raycast

### Expected Behaviour

_No response_

djpowers:
## Description

This fixes #9880. The previous endpoint used for natural language processing of expenses is no longer available, so this command is currently non-functional. This pull request uses the `/create_expense` API endpoint that is still supported.

My goal was to get the command in a working state. There could likely be some future work done to extract some of the shared form components, but I didn't tackle that now as there was some different functionality between them that complicated the process. Right now it defaults to an [equal group split](https://dev.splitwise.com/#tag/expenses/paths/~1create_expense/post) with the authenticated user as the payer.

I didn't realize at first the other "Lists Expenses" command had a "Change values" action, but I've since updated it to match that form structure for a consistent experience.

This also adds a way to select the currency when creating an expense.

My suggestion/request would be to merge this now to restore the functionality, and some of the above items could be completed in a followup pull request.

## Screenshot
splitwise-4

## Checklist

### DIFF
--- a/extensions/Splitwise/src/add-expense.tsx
+++ b/extensions/Splitwise/src/add-expense.tsx
@@ -136,7 +136,7 @@ function FillForm(props: FriendOrGroupProps) {
         title={`${props.friend ? "Friend" : "Group"}`}
         text={props.friend ? [props.friend.first_name, props.friend.last_name].join(" ") : props.group.name}
       />
-      <Form.TextField title="Description" placeholder="Grocery run" {...itemProps.description} />
+      <Form.TextField title="Description" {...itemProps.description} />
       <Form.DatePicker title="Date of Expense" {...itemProps.date} />
       <Form.Dropdown title="Currency Code" {...itemProps.currency_code}>
         <Form.Dropdown.Item value="USD" title={`USD (${getCurrency_code("USD")})`} icon="💵" />


mathieudutour (approved):
That makes sense, thanks for fixing it :+1:

maximilianzuern:
Thank you for your contribution @djpowers 

Unfortunately, this only works for group expenses. 
For individual expenses, the 'by_shares' schema is required, as implemented in the [list-expenses command](https://github.com/raycast/extensions/blob/main/extensions/Splitwise/src/list-expenses.tsx#L235).

Here's a schema that works for non-group expenses:
```
{
	"cost" : 10,
	"description" : "Test",
 	"currency_code" : "EUR",
 	"users__0__user_id" : <user's id>,
 	"users__0__paid_share" : 10,
 	"users__0__owed_share" : 5,
	"users__1__owed_share" : 5,
	"users__1__user_id" : <friend's id>
}
```

Keep in mind, direct division of costs may not work due to Splitwise's rounding. 
For example, a cost of 10.01 divided by 2 gives 5.005, but Splitwise rounds this to 5.01, resulting in a total of 10.02. This difference can prevent the creation of the expense, due to a mismatch between the total owed shares (€10.02) and the actual cost (€10.01).
Even though a 200 API response, the response text would be:
```
{'expenses': [],
 'errors': {'base': ["The total of everyone's owed shares (€10.02) is different than the total cost (€10.01)"]}}
```

Therefore, I used the [this logic](https://github.com/raycast/extensions/blob/f82b8bbc1be0fffa0d90c1a6d24f35a026e124d1/extensions/Splitwise/src/list-expenses.tsx#L228C7-L228C12) to edit expenses.

PR #10017, opened 2024-01-05
Update arc extension (Fix #8189)

### LINKED ISSUE #8189: [Arc] Searching Tabs doesn't open always open the correct tab
### Extension

https://www.raycast.com/the-browser-company/arc

### Description

For example, Arc opens tab 1, tab 2 and tab 3. This plugin properly display them in the Raycast. 

However, when I select tab 1, it may open tab 2. The behavior is consistent for a certain selection. If I select tab 1 again, it will always open tab 2.

I also observe certain cases where it will open the correct tab.

### Steps To Reproduce

Mentioned in the description. Not sure how easy it is to reproduce. I don't know whether this depends on my space and tabs I have.

### Current Behaviour

_No response_

### Expected Behaviour

_No response_

loris:
## Description

Fix #8189

Previous implementation is flawed because we fetch the tabs
and their "indexes". However, those indexes will change if some tabs
are added (or reordered, removed…) from Arc. Since we cache the indexes
in Raycast between command execution, we could keep outdated value in 
cache and thus open the wrong tab.

Naive solution would be to remove the cache but it makes the extension
very slow (4 to 5 seconds to load tabs on each invokation in my setup)

New implementation will cache the tabs without their "indexes" and will
search for the tab index (which is required by current Arc AppleScript
implementation to open a tab) once selected. It properly fixes the issue
(will always open the correct tab), however there is now a slight delay
when opening a tab (delay will vary based on the number of tabs you have)

## Checklist

### DIFF
--- a/extensions/arc/CHANGELOG.md
+++ b/extensions/arc/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Arc Changelog
 
+## [Fixes] - 2024-01-05
+
+- Fixed an issue (#8189) where a wrong tab could be randomly opened instead of the one selected
+
 ## [Improvements & New Features] - 2023-12-08
 
 - Improve behavior when opening a new tab, if no window is open, a new window is opened


thomaslombart (approved):
Thanks for the contribution! 🚀

PR #10045, opened 2024-01-08
Fix/swift evolution keys

dehesa:
## Description

The following is a list of the changes applied to the [Swift Evolution](https://www.raycast.com/BProg/swift-evolution) extension by @BProg:
- Migrated to the new Raycast API version.
- Fix a typo in the `package.json`.
- Fix React warnings about list items without keys in `map` loop.
- Fix issue when the Swift version is not specified.
- Used _cached state_ to remember if you use simple or detail list last time a user opened the command.
- Add standard keyboard shortcuts to open in browser and copy proposal URL.

## Checklist

### DIFF
--- a/extensions/swift-evolution/package.json
+++ b/extensions/swift-evolution/package.json
@@ -5,6 +5,10 @@
   "description": "Search and see details for the Swift language proposals from swift evolution web page https://apple.github.io/swift-evolution/",
   "icon": "swift.png",
   "author": "BProg",
+  "contributors": [
+    "dehesa",
+    "pernielsentikaer"
+  ],
   "license": "MIT",
   "categories": [
     "Developer Tools",


pernielsentikaer:
Hi @dehesa 👋 

It works as expected and it's some nice enhancements, @BProg do you want to check it before merging?

BProg:
Let me take a look

BProg:
Works great 🚀  
however I found another place where there are duplicate items: 
```
13:52:08.885 Warning: Encountered two children with the same key, `https://github.com/apple/swift/pull/62652`. Keys should be unique so that components maintain their identity across updates. Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported and could change in a future version.
```

I could fix this separately

pernielsentikaer:
You can use make a suggestion in the code @BProg 🙂

dehesa:
I am not sure whether you can directly push to this branch, but if you can feel free to do it. If not, you can do as @pernielsentikaer suggests and leave me some comments. I will then write them right away.

BProg:
@dehesa following method needs an update

New implementation for `getImplementations` from the `domain.ts` file
```
function getImplementations(proposal: ProposalJson): { title: string; url: string }[] {
  const implementations = proposal.implementation ?? [];
  const initialValue: { title: string; url: string }[] = [];
  return implementations
    .reduce((accumulator, x) => {
      const mapped = {
        title: `${x.repository}#${x.id}`,
        url: `https://github.com/${x.account}/${x.repository}/${x.type}/${x.id}`,
      }
      // Ignore duplications
      if (accumulator.find(impl => impl.url == mapped.url)) {
        return accumulator
      }
      return accumulator.concat(mapped);
    }, initialValue)
}
```

dehesa:
Done! We are done @pernielsentikaer !

By the way, @BProg, it would be amazing to add ⌘+P filter for the status of the proposals (i.e. accepted, returned, implemented, etc.) if you have some time. Bonus points if we can filter for Swift version implementations. Sadly, my TypeScript/React knowledge is not that great.

BProg:
@dehesa The existing search is looking at status and swift version as well. It may not be very intuitive, I do like your suggestion, would look at it 👍🏽 

Existing search functionality:
Screenshot 2024-01-12 at 08 54 58
Screenshot 2024-01-12 at 08 54 48

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #10081, opened 2024-01-10
feat: add command to open issue from clipboard to jira

mheidinger:
## Description

During daily work I often find myself in a situation that a coworker just sent me a Jira issue key but without a link to directly open it. This PR adds a helper to the existing jira extension that opens the Jira issue with just copying the key and triggering the "Open Issue from Clipboard" command.

It supports custom base URLs so that if provided the issue is opened in the browser instead of in Raycast.

## Screencast

https://github.com/raycast/extensions/assets/8899636/342ad5b5-43a7-4b83-ae0d-ebb2a395007e

Invalid Jira Key and Custom Base URL example:

https://github.com/raycast/extensions/assets/8899636/ed52d7ed-26f6-4b8c-a1f9-21feae3f15a3

## Checklist

### DIFF
--- a/extensions/jira/CHANGELOG.md
+++ b/extensions/jira/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Jira Changelog
 
+## [Add command to open issue from clipboard] - 2024-01-19
+
+- Add a new command to open issue by key from the clipboard.
+
 ## [Fix issue with JQL reserved keywords in project names] - 2023-12-18
 
 - Fix a bug causing queries to fail when project names included JQL reserved keywords.
--- a/extensions/jira/package.json
+++ b/extensions/jira/package.json
@@ -13,7 +13,8 @@
     "rifont",
     "gavinroderick",
     "lusit",
-    "literallyjustroy"
+    "literallyjustroy",
+    "mheidinger"
   ],
   "owner": "raycast",
   "access": "public",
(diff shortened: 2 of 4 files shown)


pernielsentikaer:
@thomaslombart would that fit into the current implementation we have, or do you think it works better as a standalone?

thomaslombart:
We provide a similar experience in the current Jira extension. By enabling `Search Issues` as a fallback command, users can copy the key, launch `Search Issues` and open the issue in Raycast or the browser from there. However, this adds an extra step, so it could also be a separate command in the current Jira extension.

mheidinger:
I'd be happy to refactor this into the existing implementation.

For me it would be important that:
1. It works not only with the Atlassian cloud version but also on-premise installations (don't think the current implementation supports that)
2. It's a one click action
3. Best case it works without authentication as that _could_ be an issue with locked down on-premise installations (but haven't tested that yet against "my" instance)

thomaslombart:
I think the simpler would be to add a new command in the existing Jira implementation. That way, you can leverage the Atlassian cloud authentication but also add support for on-premise installations in the command preferences. What do you think?

mheidinger:
Sure I'll have a look into that, just not sure when I'll find the time for that :)

mheidinger:
@thomaslombart I updated the PR accordingly (incl. screencasts), would be great if you could have another look. Thanks!

FezVrasta (approved):

thomaslombart (approved):

PR #10087, opened 2024-01-10
Add my-gitlab-dashboard extension

mikybars:
## Description

Track merge requests and pipelines from all your projects in one single place

## Screencast

image

## Checklist

### DIFF
--- a/extensions/my-gitlab-dashboard/package.json
+++ b/extensions/my-gitlab-dashboard/package.json
@@ -81,7 +81,8 @@
       "required": false,
       "title": "Soft warning threshold for dates",
       "description": "Creation and modification times that are past this value will be displayed in a different colour",
-      "placeholder": "1d"
+      "placeholder": "1d",
+      "default": "1d"
     },
     {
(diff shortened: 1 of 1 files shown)


pernielsentikaer:
Hi 👋

Would that fit into the [GitLab](https://www.raycast.com/tonka3000/gitlab) extension by @tonka3000 instead of being a standalone extension?

mikybars:
As I see it the point of a dashboard is making it easy to manage multiple projects by hiding all the moving parts in one single place (i.e. command). There'll be power users for which this simplified view will just not be enough and prefer the fine grained control offered by the GitLab extension. Therefore those users will simply dismiss this addition. 
On the flip side some users just want to get things done and are overwhelmed by all the bells and whistles of a thorough solution like that.
I think there may be some kind of overlap between these two groups of users but I'm not entirely sure if that's worth of an integration effort. It would be nice to have @tonka3000 's opinion on this :)

mikybars:
I then guess this PR can be closed as an integration is preferred.

PR #10108, opened 2024-01-11
Add simple-icons extension

LitoMore:
## Description

This extension is used to copy SVG icons from https://simpleicons.org.

### References

- [Simple Icons repository](https://github.com/simple-icons/simple-icons)

## Screencast

|First view|Detail page|
|:-:|:-:|
|simple-icons-1|simple-icons-2|

## Checklist


pernielsentikaer:
Hi 👋

Is there a way to stream them directly from an external source and just having a json file pointing on the path instead of having them all in the PR?

LitoMore:
@pernielsentikaer It's possible. You could try with jsDelivr resource, which is for the copy action.

But using image URLs can cause loading throttle when users scroll the page quickly.

pernielsentikaer:
That might still be an idea to explore, we'll cache the images when loading so it shouldn't be that bad but ofcause some throttle will happen

Another thing could be a way to unpack from external source to local somehow?

LitoMore:
> That might still be an idea to explore, we'll cache the images when loading so it shouldn't be that bad but of cause some throttle will happen

I just tried, and the user experience is terrible. It takes very long time to load those stucked image requests.

> Another thing could be a way to unpack from external source to local somehow?

But I can only use image assets from `assets/`. Do you mean I use some scripts to download the image pack and unpack them to `assets/`? This means the script directly accesses the `.config/raycast` folder. Is this consistent with Raycast's philosophy?

We have the SVG image path in the object, but Raycast's `Image.ImageLike` type does not support SVG tag. For example:


```jsx
<Grid.Item content={{value: <svg><path d={`icon.path`} /></svg>}} />
```

pernielsentikaer:
We have [environment.supportPath](https://developers.raycast.com/api-reference/environment) which is meant for exactly that purpose - could that work at all?

pernielsentikaer:
I think extensions like https://www.raycast.com/destiner/iconify does something like that too, if I'm not mistaken, github is having a hard time here :D

LitoMore:
> We have [environment.supportPath](https://developers.raycast.com/api-reference/environment) which is meant for exactly that purpose - could that work at all?

I'm afraid we can't use `supportPath`. According to the [Icons & Images](https://developers.raycast.com/api-reference/user-interface/icons-and-images) documentation, images can only be used from `assets/`. The [`environment.assetsPath`](https://developers.raycast.com/api-reference/environment) returns the correct asset path, but it's not the recommended directory for read & write.

Even if we can download the icon pack file, we still need to execute an "unpack" command to unpack them to `assets/`. This is inconvenient for future version updates.

So, I think committing asset files to our repository is the best solution based on the current Raycast API design.

pernielsentikaer:
There is also this extension doing something similar, is there any inspiration to checkout? https://github.com/raycast/extensions/pull/10144/files

LitoMore:
> There is also this extension doing something similar, is there any inspiration to checkout? [#10144 (files)](https://github.com/raycast/extensions/pull/10144/files)

They just cached icon data from their API. For simple-icons, we already have icon data from the npm package. So, they will still encounter throttle problems when loading images from the CDN.

pernielsentikaer:
Since Raycast caches them first time it sees it I don't think it's that big of a deal - do you?

LitoMore:
Reopen

PR #10110, opened 2024-01-11
Add typofixer extension

archibasov:
## Checklist


pernielsentikaer:

What's the difference between this and [Punto Switcher](https://www.raycast.com/red.avtovo/punto) from the store?

archibasov:
> What's the difference between this and [Punto Switcher](https://www.raycast.com/red.avtovo/punto) from the store?

Typofixer:
- supports other keyboard layouts (it's main difference)
- has layouts settings
- can convert each word separately
- has a more correct algorithm for encoding detection

pernielsentikaer:
Basically, they do the same, right? Another solution would be to let this one supersede the other one so we don't end up with multiple extensions doing the same 🙂

archibasov:
Sounds great. I can improve typofixer to support a different layout.

pernielsentikaer:
What do you think @red-avtovo since you're the author of Punto 🙂

red-avtovo:
@pernielsentikaer thanks for reviewing this one. I was already reviewing the very same change, but as a PR to puntoswitcher extension. I was a bit surprised, that the [PR](https://github.com/raycast/extensions/pull/9995) was closed and now I understand why.
I had no intention to "gatekeep" the extension, but was rather offering a clearer way to define immutable parts of it. I still think, that the original PR was a great contribution and I would love to see it merged

archibasov:
Yes, I've already tried making PR to Punto, with a sincere desire to do better. But:
- To make it work correctly, I have to completely rewrite all the code of Punto.
- @red-avtovo didn't like my approach. Although it is more universal, correct and with elegant code.
I can improve typofixer to support layout of Punto.

pernielsentikaer:
It would be better for all users to only have one agreed-upon extension. Otherwise, there will be a lot of confusion, and we are really trying not to make that happen 🙂

red-avtovo:
In all fairness, I merely suggested some changes and without a comment you've closed your PR, which was at least a surprise to me

pernielsentikaer:
Does this cover everything in Punto @red-avtovo? Then, this could also supersede it if it's doing the job better. I'm not sure what this really does, so I can't judge here.

red-avtovo:
While it implements much more than Punto does I still think, that it is only because there was no intention in that logic in the first place
- supports other keyboard layouts (it's main difference) -> skimming through the code gives me an impression, that it supports only 2 layouts (as mine extension as well), so either it is not fully implemented or I'm missing something
- has layouts settings -> Fair feature and I would be happy to introduce that in Punto, but I had no use case
- can convert each word separately -> not a feature that I had ever a use case, but yes, there it no such thing implemented in punto
- has a more correct algorithm for encoding detection -> that is subjective. This extension has just a regex to detect the layout

While we're selecting the extension to choose, I would remind, that we're talking about mostly 1 function only and about 15 lines of code in difference. I would like to highlight again, that it was not my intention to gatekeep the project and I'm more than happy to accept any feature request. Only thing that I was asking was a dialogue

PR #10181, opened 2024-01-16
Update ext/waktu solat

sayuthisobri:
- Fix some issue and logic on menubar
- Monotone icon in menubar


pernielsentikaer (changes requested):
Hi 👋

Wouldn't it be better to always show the icon instead of only when there is not recent prayers?

sayuthisobri:
> Hi 👋
> 
> Wouldn't it be better to always show the icon instead of only when there is not recent prayers?

Hi 👋,

I've make changes to address that, but is it ok that checking based on user preference that one still remain? mean if icon enabled in preference we always display the icon otherwise we just display text based notification on event similar to Calendar extension 👉👈.

pernielsentikaer:
That should be fine 🙂

pernielsentikaer:
Is this ready for review then?

PR #10186, opened 2024-01-16
feat(arc): "new little arc" command searches highlighted text or opens highlighted URLs

bdsqqq:
## Description

Enables the "New Little Arc" command from the Arc extension to open selected URLs and search selected text. The search engine used for searching with this command can be changed with a preference, Google is the default.

## Screencast

https://github.com/raycast/extensions/assets/37847523/f6cd4d99-6425-4c72-9468-7afc02a8baeb

## Checklist

### DIFF
--- a/extensions/arc/src/new-little-arc.ts
+++ b/extensions/arc/src/new-little-arc.ts
@@ -17,7 +17,7 @@ export default async function command(props: LaunchProps<{ arguments: URLArgumen
   const { fallbackText } = props;
   const selectedText = await getSelectedText();
 
-  const selectedTextAsSearch = `${config[newLittleArcPreferences.engine].search}${encodeURIComponent(selectedText)}`;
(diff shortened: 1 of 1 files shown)


extensions/arc/src/types.ts
```diff
@@ -19,6 +19,18 @@ export type Space = {
   title?: string;
 };
 
+export type NewTabSearchConfig = {
+  search: string;
+};
+
+export type NewTabSearchConfigs = {
+  google: NewTabSearchConfig;
```
peduarte, line 27: curious why use `NewTabSearchConfig` with a single `search` key rather than just `string`?

bdsqqq: good catch, this was mostly copy pasted from below, then trimmed to only include what was needed.

Updating to use a string.

bdsqqq:
🚢🚢🚢

PR #10218, opened 2024-01-18
[ssh-manager] fix parsing remote command in ssh config

gdebrauwer:
When the RemoteCommand option in a SSH config file contains multiple spaces ...
```
Host Foo
  HostName example.com
  RemoteCommand cd current; bash -l
```
... then that remote command is parsed incorrectly:
```
ssh -t example.com "cd" 
```
This PR fixes that so it parsed correctly:
```
ssh -t example.com "cd current; bash -l" 
```


pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #10242, opened 2024-01-19
Update homeassistant extension

amuelli:
## Description

I'm using the "Lights" command quite often and it would be nice to have the list of lights show up immediately without having to wait for the first state update. I therefore suggest using a cache for the states.

## Checklist

### DIFF
--- a/extensions/homeassistant/src/components/hooks.tsx
+++ b/extensions/homeassistant/src/components/hooks.tsx
@@ -2,6 +2,7 @@ import { Connection, entitiesColl, subscribeEntities } from "home-assistant-js-w
 import { useEffect, useRef, useState } from "react";
 import { getHAWSConnection } from "../lib/common";
 import { State } from "@lib/haapi";
+import { useCachedState } from "@raycast/utils";
 
 interface EntityRegistryEntry {
   device_id?: string | null;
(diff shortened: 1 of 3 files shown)


extensions/homeassistant/src/components/hooks.tsx
```diff
@@ -41,21 +42,13 @@ export function useHAStates(): {
   error?: Error;
   isLoading: boolean;
 } {
-  const [states, setStates] = useState<State[]>();
+  const [states, setStates] = useCachedState<State[]>("states");
```
amuelli, line 45: I suggest to use a cache for storing states. This way, when we execute a command like "Lights," the list won't be empty until we receive the first state update. Instead, it should already be populated with the entries from the last time we ran any command.

extensions/homeassistant/src/components/hooks.tsx
```diff
@@ -41,21 +42,13 @@ export function useHAStates(): {
...
   const [error, setError] = useState<Error>();
   const [isLoading, setIsLoading] = useState<boolean>(true);
   const hawsRef = useRef<Connection>();
 
   useEffect(() => {
-    // FIXME In the future version, we don't need didUnmount checking
-    // https://github.com/facebook/react/pull/22114
-    let didUnmount = false;
```
amuelli, line 52: Looks like this is no longer needed.

extensions/homeassistant/src/components/hooks.tsx
tonka3000: Works nice. A little snappier

extensions/homeassistant/src/components/hooks.tsx
tonka3000: Seems so.

tonka3000 (commented):
What/Why did you change in the png files?

amuelli:
> What/Why did you change in the png files?

That was not intentional. Somehow that diff got added when I ran the "Fork Extension" command and then after adding my changes ran `npx @raycast/api@latest publish`.
I'll try to remove them.

tonka3000 (approved):
Thanks for removing the png changes.

Looks great, thanks for the contribution 🎉

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #10243, opened 2024-01-19
Quit Applications: Update extension to use `fileIcon` instead of getting it from the app's `.plist`

sxn:
## Description

- Updates extension to not parse each application's `.plist` in order to find its icon, but to instead make use of the [fileIcon](https://developers.raycast.com/utilities/functions/runapplescript) API. This should clean up the code, and as a side effect _should_ also make the extension load faster.
- Updates `@raycast/api` to the latest version
- Replaces `run-applescript` dependency with `@raycast/utils`' [runApplescript](https://developers.raycast.com/utilities/functions/runapplescript)

## Screencast

https://github.com/raycast/extensions/assets/1155589/d2316d3f-c754-4b8a-853d-a8f1ef962d7d

### DIFF
--- a/extensions/quit-applications/package.json
+++ b/extensions/quit-applications/package.json
@@ -5,8 +5,15 @@
   "description": "Quit appplications directly from Raycast",
   "icon": "quit-app.png",
   "author": "mackopes",
-  "contributors": ["jrnxf", "3llomi", "sxn"],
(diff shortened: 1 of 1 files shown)


pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #10244, opened 2024-01-19
Tmux Sessioner: Launch `Change Default Terminal` if there is none currently set.

sxn:
## Description

I installed the extension and tried to see my sessions, only to be greeted with an error screen telling me I didn't select a default terminal. I also missed that there's an action to `push` the terminal view, so it was a bit annoying.

This PR makes it s.t. instead of showing an error view, the `Change Default Terminal` command is directly launched, and on success, the previous command is launched again, to make onboarding slightly more user friendly.

In the process I also updated `@raycast/api` to the latest version, replaced the `run-applescript` dependency with `@raycast/util`'s [runApplescript](https://developers.raycast.com/utilities/functions/runapplescript), and removed the `applicationIconFromPath` utility function because Raycast can already fetch [fileIcons](https://developers.raycast.com/api-reference/user-interface/icons-and-images#fileicon) 😸

## Screencast

https://github.com/raycast/extensions/assets/1155589/1f0ff384-3acb-40d1-83ee-39bf5c133b92


louishuyng (approved):
Awesome @sxn thanks for your contribution. I really like that approach 🎉

PR #10253, opened 2024-01-20
Add timecamp extension

iAmKevinCruz:
## Checklist


extensions/timecamp/package.json
```diff
@@ -0,0 +1,48 @@
...
+      "name": "index",
+      "title": "Start/Stop Tasks",
+      "description": "Start/Stop tasks and edit active timer's notes",
+      "mode": "view"
+    }
+  ],
+  "preferences": [
+    {
+      "name": "timecamp_api_token",
+      "description": "Get your programming API token from https://app.timecamp.com/app#/settings/users/me",
```
pernielsentikaer, line 24: Could you add a placeholder to show a mock api key 🙂

extensions/timecamp/src/index.tsx
```diff
@@ -0,0 +1,374 @@
...
+        setActiveTask(taskFetch);
+        setSelectedItemId(taskFetch.task_id.toString());
+      }
+    } catch (error) {
+      console.error(error);
+    }
+  }
+
+  return (
+    <List filtering={{ keepSectionOrder: true }} selectedItemId={selectedItemId} searchBarPlaceholder="Search Task">
```
pernielsentikaer, line 347: Missing `isLoading` property

pernielsentikaer (changes requested):
Hi 👋 

- You can maybe use [useFetch](https://developers.raycast.com/utilities/react-hooks/usefetch) to fetch which gives you cache, error handling and isLoading 🙂

extensions/timecamp/package.json
iAmKevinCruz: Taken care of :)

extensions/timecamp/src/index.tsx
iAmKevinCruz: Added that in. Good catch!

pernielsentikaer (commented):
Maybe we can add an action called "Start Task" and another one where it's possible to set a note. Currently, it's a bit confusing because when the task is started, it goes to a different view where it seems like we should be able to add a note. However, when we try to insert a note, the window just closes. It would be helpful to have a HUD or toast notification that guides the user back to the main view after adding a note.

iAmKevinCruz:
Thank you for your feedback, and so quick!! This is a feature. After
testing it with a few coworkers, we determined that the natural flow after
starting a task is to want to add a note and then close raycast to continue
working. We got the inspiration to do that way from Timeular’s Mac app. But
I do see how this can be confusing for those just starting.

How does making it your way, as in when you start a task, you have to enter
it to edit the note, and upon saving the note it tasks you back to the task
list. But I can add some preferences that will give the user the
flexibility to make it work how it currently works. What do you think?

iAmKevinCruz:
@pernielsentikaer I added new commands. 

**List - Tasks**
- Now there are `Start Task & Edit Note` on `ENTER`, and `Start Task` on `CMD + S`.

**Form - Edit Note**
- Now there is `Save Edits & Close Window` on `CMD + ENTER`, and `Save Edits` on `CMD + S`.

Let me know if that is satisfactory.

I had an idea to add an `ENTER` keyboard listener in the form that will submit the form and go through that flow. Will that pass review?

pernielsentikaer (approved):
All other forms uses cmd+enter so we want to stick to that 🙂

PR #10262, opened 2024-01-21
[Twitter] Restore some functionality which now use the browser

### LINKED ISSUE #10252: [Twitter] Error 403
### Extension

https://www.raycast.com/tonka3000/twitter

### Description

I gave permission to the extension but still getting Error 403.

Screenshot 2024-01-20 at 13 37 05

### Steps To Reproduce

Run any command except "Search User".

### Current Behaviour

Error 403

### Expected Behaviour

See the normal output.

tonka3000:
## Description

Try to restore some functionally after X.com smashed the API.

- Use X icon instead of the blue bird

  I can not change the title to X because at least 3 characters are required 😆

- Settings for v1 API are removed

  Kept v1 components if it maybe work some day again.

- `Send Tweet` and `Search User` open the browser when no OAuth key is set
- Removed the old OAuth key because it is no longer working
- `Recent Tweets` and `My Recent Tweets` now have a warning when no OAuth key is configured
- OAuth key is now optional. By default the extension use the website when hit a command

I don't have a X subscription, so maybe somebody can help in the future to check if the old API stuff is working and maybe add some fixes.

Closes #10252

## Checklist


xilopaint:
What do you mean by subscription? Does this require a Premium subscription?

tonka3000:
> What do you mean by subscription? Does this require a Premium subscription?

Yes. Musk changed requirements of the API. Basically on Send does work in the free version very limited.

CleanShot 2024-01-21 at 13 51 10@2x

xilopaint:
Well, I don't have a paid user or dev account. I don't believe anyone will pay $100/month for personal API access, so the new requirements seem to have killed this extension.

tonka3000:
@xilopaint No really killed, but limited the option. With paid sub stuff should be able to work.
I now fallback to web-version for `Send Tweet` and `Search User` which work quite well for me.

pernielsentikaer:
Is there any other alternatives to this @tonka3000 - if we don't do this, what would then happen with above changes?

pernielsentikaer (changes requested):
Could we maybe make the List.EmptyView image 128x128 or less, it's a bit too big at the moment 🙂

tonka3000:
@pernielsentikaer 

> Could we maybe make the List.EmptyView image 128x128 or less, it's a bit too big at the moment 🙂

I use now a built-in icon which is much smaller.

> Is there any other alternatives to this @tonka3000

In my opinion not much options. The old API should still work. Now the OAuth key is just optional.

If no key is set the Send Tweet will work as before but open the a new web tweet with the content. No perfect but it works not that bad.

> if we don't do this, what would then happen with above changes?

At least the OAuth key should be removed, because it does not work and users will login but will always see a 403.
I personally switch to this new mode for myself, because I don't pay for the API key that much.

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

Issue #8844, opened 2023-10-17
[Google Translate] Translate to multiple languages at the same time


rasitayaz:
### Extension

https://www.raycast.com/gebeto/translate

### Description

Currently, using the 'Translate' command, we can add multiple language sets and switch between them. But even when we want to translate interchangeably between two languages (like A to B, and then B to A), we need to change the language set.

With this new feature, user will not need to constantly change language sets. Let me explain what I have in my mind with some example images I made:
<img width="862" alt="example" src="https://github.com/raycast/extensions/assets/45076094/04a264f7-6299-4f25-bd59-c972017a25b6">
<img width="862" alt="example2" src="https://github.com/raycast/extensions/assets/45076094/d9e79278-d524-4063-9be4-d9a0f1e6f019">

### How this feature should work in my opinion;

- Command should ask the user for two main languages for the first run (like the current 'Translate' command)
- User should be able to add and delete languages in the list (min two languages should be mandatory)
- Command should automatically hide the detected language in the result list (since it will display the input text as is)

We can either change the existing command with this, or maybe add another command for this feature like "Translate Multiple Languages". What do you think?

### Who will benefit from this feature?

Users who are regularly making translations between multiple languages, or making translations interchangeably between two languages.

gaojunran:
if you strongly need the feature, you can try [antfu's raycast extension](https://github.com/antfu/raycast-multi-translate), which meets your requirement. However, as he does not like the management way of extensions of raycast, you need to follow the steps in his README and install it locally.

gebeto:
I'm going to add ability to add multiple target languages to the translation sets

gebeto:
Hey @rasitayaz , I've just added this ability here, now you will be able to create language set with as much target languages as you want:

https://github.com/raycast/extensions/pull/16240

PR #10264, opened 2024-01-21
Add new "Quick Translate" command to "Google Translate" extension

rasitayaz:
## Description

Google Translate extension had two commands, "Translate" and "Translate Form". Using "Translate", we can define multiple language sets to translate from/to. With "Translate Form", we can select single language set (from/to) and directly translate a text.

With this new "Quick Translate" command, the users are not restricted to language sets, and can quickly translate from the selected language to multiple languages at the same time. #8844 

By default, auto-detect is used as source language, primary & secondary languages are used as target languages. Users are allowed to add as many target languages as they want.

Apart from the new command, I changed preference titles like following:

* First Language -> Primary Language
* Second Language -> Secondary Language

In my opinion, it's better to replace the original "Translate" command with this one instead of adding this as a new command because I think this is easier to use compared to language set approach. Take a look at the command and its ux, if you agree with me, request changes for the pr and I'll gladly replace the original one with this implementation.

## Screencast

image
image
image
image
image

## Checklist

### DIFF
--- a/extensions/google-translate/CHANGELOG.md
+++ b/extensions/google-translate/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## [Enhancement] - 2024-01-26
 
- - Added "Autofill Input" preference to allow users to disable the autofill selected text feature.
+ - Added "Autofill Input" preference to allow users to disable the autofill selected text feature
 
 ## [Feature] - 2024-01-20
 


nirtamir2 (approved):

gebeto (approved):
It looks great! Thanks for your contribution!

Yeah, this one definitely have better UX then original `Translate`, but let's have it as separate command for now, and maybe in the future we will replace `Translate` with `Quick Translate`

extensions/google-translate/src/simple-translate.ts
```diff
@@ -96,3 +96,18 @@ export async function doubleWayTranslate(text: string, options: LanguageCodeSet)
...
+
+  return await Promise.all(
+    targetLangs.map((lang) =>
+      simpleTranslate(text, {
+        langFrom: sourceLang,
+        langTo: lang,
+      }),
+    ),
+  );
+}
```
gebeto, line 113: I have a suggestion about translations. So as we have ability to reorder target languages, I think we can utilize it. And instead of translating to all the targets in one Promise, we can translate every language in own `List.Item`. And show some loader during translation on each item

So it can looks like this:

```
{targetLangs.map(t => (
  <TranslationListItem sourceLang={sourceLang} targetLang={t} text={text} />
)}
```

rasitayaz: done! waiting for your feedback. :)


https://github.com/raycast/extensions/assets/45076094/6278f480-23f9-4fb3-8efe-755fe0484ce0

rasitayaz:
How do you guys feel about giving users the option to disable autofill input like below? Translate commands always try to get the selected text and sometimes make that annoying 'ding' sound when they fail. Personally, I'd keep it disabled.

image

pernielsentikaer:
If the default behavior is the same as today, then I don't see a problem, do you @gebeto

gebeto:
Yep, I agree with @pernielsentikaer , if it will be checked by default, so it will not affect old users UX and don't force them to go to the settings and change settings, we can do it

extensions/google-translate/src/simple-translate.ts
gebeto: Looks awesome!

rasitayaz:
okay, working on it

gebeto (approved):

rasitayaz:
done, can you check if you're able to use it as is and the option is enabled by default?

gebeto:
Yep, according to the api we can set `default` value for the preference: https://developers.raycast.com/information/manifest#preference-properties

gebeto:
one sec, I'll check it

gebeto:
@rasitayaz can you please add this latest change to the Changelog?

gebeto:
> done, can you check if you're able to use it as is and the option is enabled by default?

yep, it works!
Screenshot 2024-01-26 at 11 27 21

rasitayaz:
> @rasitayaz can you please add this latest change to the Changelog?

updated

gebeto (approved):

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

Issue #8870, opened 2023-10-18
[Microsoft Teams Meeting] The new Microsoft teams app is not supported


Richigeht:
### Extension

https://www.raycast.com/sven/microsoft-teams-calling

### Description

Hello,
I updated to the "new" Microsoft teams 2.0 app and the shortcts don't work no more. 

It's not only you - also OBS doesn'T work with it anymore or Teams doesn't with both of you. But it would be great to get back my beloved shortcuts :-) thanks!

### Steps To Reproduce

1. Update to the NEW teams 2.0 app.
2. Be happy about the quicker loading times.
3. Notice in your meetings, you cannot control the mute button anymore....
4. Wonder about less successful meetings :( 
5. Make a bug at github.... oh wait! ;) 

### Current Behaviour

Error message within Teams that it cannot control the app. 

### Expected Behaviour

Mute/unmute to work

svenwiegand:
Hi @Richigeht, I can confirm that the new Teams 2.0 client currently doesn't provide the API used by my extension. 

I'm using an undocumented API here, which is the same as used by the Microsoft Teams plugin for the Streamdeck (see [Microsoft's blog post](https://techcommunity.microsoft.com/t5/microsoft-teams-blog/delivering-new-webinar-experiences-with-microsoft-teams/ba-p/3725145)). On the [plug-in's page](https://marketplace.elgato.com/product/microsoft-teams-da5e2bbc-197c-4afe-8a85-a9941bf52697) they say, the latest version of the plugin is working with the new Microsoft Teams, but for me, it does not. 

As this API isn't documented anywhere, at the moment I don't see a chance to get this fixed. Any hints are welcome.

raffiloch:
@svenwiegand - First, thank you for the excellent extension! It is incredibly useful and greatly simplifies things.

I have come across various Reddit posts mentioning that the Streamdeck plugin for Teams now works with the New Teams. However, the API key registration flow has changed slightly. The New Teams now supports third-party app APIs, but there is no longer an option to generate a new API key. According to the Microsoft support page, once the feature is enabled, you simply need to activate the plugin on the third party device. A confirmation window will then appear to connect the plugin with Teams. I'm really curious if something similar can be achieved with the Raycast extension. :-)

In any case, thank you for your outstanding work!

Links:
- https://support.microsoft.com/en-us/office/connect-to-third-party-devices-in-microsoft-teams-aabca9f2-47bb-407f-9f9b-81a104a883d6?storagetype=live
[...]

PR #10307, opened 2024-01-23
[MS Teams] Add support for new Teams API (2.0.0)

philippreissing:
## Description

The current extension for MS Teams only supports the Classic version of Teams. The New Teams has a slightly different API and doesn't work with the current extension. (#8870)
This PR includes a `MeetingClientV2` that supports version 2 of the Teams API with the same functionalities as for the classic teams.

Important changes on the API:
- The token mechanism works differently. While the token had to be looked up in the Teams client and then provided in the preferences of the extension, the new API first works without token. Once an action is requested, the user has to confirm this in the Teams client and the API then provides the token directly via the Websocket. The extension then stores the token in Raycast storage.
- Some commands have changed (e.g. reactions now use a single action `send-reaction` with an additional parameter to specify the type of reaction). 
- The `service` parameter doesn't seem to be needed anymore
- Also, the action to request an update of the meeting status has changed and I couldn't find out the new action. I worked around this in the code.

This code works fine with my Teams client and I tested all the functionalities. I also tried to keep the existing Classic API working by introducing a V2 client and keeping the V1 client. This was tricky sometimes because the APIs differ, but I tried to keep it as consistent as possible.

A switch in the preferences now allows to select the version. The token can be left empty for the New Teams since it will be provided by the API automatically.
Screenshot 2024-01-23 at 22 24 06

I feel this could still use some testing with both Teams versions. There's probably also ways to improve the code itself.

## Checklist

### DIFF
--- a/extensions/microsoft-teams-calling/src/controlMeeting.tsx
+++ b/extensions/microsoft-teams-calling/src/controlMeeting.tsx
@@ -142,7 +142,8 @@ const sections: Section[] = [
 ];
 const allItems = sections.map((section) => section.items).flat();
 
-const isItemEnabled = (item: Item, state: State) => (state.state.isInMeeting || state.permissions.canPair) && state.permissions[item.permission];
+const isItemEnabled = (item: Item, state: State) =>
+  (state.state.isInMeeting || state.permissions.canPair) && state.permissions[item.permission];
 const itemContent = (item: Item, state: State): Image.ImageLike => {
   const iconName = item.state && !state.state[item.state] && item.iconInactive ? item.iconInactive : item.icon;
   return { source: iconName };
--- a/extensions/microsoft-teams-calling/src/teams/meetingClientV2.ts
+++ b/extensions/microsoft-teams-calling/src/teams/meetingClientV2.ts
@@ -149,10 +149,12 @@ export class MeetingClientV2 implements MeetingClient {
   }
 
   private convertUpdateMessageToV1(msg: UpdateMessageV2): UpdateMessage {
-    const meetingState: MeetingState | undefined = msg.meetingUpdate.meetingState ? {
-      ...msg.meetingUpdate.meetingState,
-      isCameraOn: msg.meetingUpdate.meetingState.isVideoOn,
-    } : undefined;
+    const meetingState: MeetingState | undefined = msg.meetingUpdate.meetingState
+      ? {
+          ...msg.meetingUpdate.meetingState,
+          isCameraOn: msg.meetingUpdate.meetingState.isVideoOn,
+        }
+      : undefined;
     return {
       ...msg,
       meetingUpdate: {


pernielsentikaer:
@philippreissing Could you add an entry in the `CHANGELOG.md` file? Here's our guidelines about that: [Version History](https://developers.raycast.com/basics/prepare-an-extension-for-store#version-history).

@svenwiegand do you have the possibility to add this?

philippreissing:
@pernielsentikaer : Changelog is added. Thanks for the hint.

svenwiegand:
Sorry for the delay @philippreissing and @pernielsentikaer, I'll take a look on this today.

svenwiegand:
Great work @philippreissing! I only recognized one crash, one missing option to pair and some missing documentation. Additionally, I've found one opportunity for better type safety.

# Crash
The crash occurs, when you start the **Control Meeting** action, while in a meeting in the new Teams and you are not yet paired. This is due to the fact that Teams will send a message where the `meetingState` isn't set, which isn't supported by our types and thus, not handled by our code. To resolve this, we only need a few changes:

In `meetingClient.ts`:

```TypeScript
export interface UpdateMessage {
  apiVersion: string;
  meetingUpdate: {
    meetingState?: MeetingState; // <<-- lets make this optional
    meetingPermissions: MeetingPermissions;
  };
}
```

In `meetingClientV2.ts`:

```TypeScript
...

interface UpdateMessageV2 {
  apiVersion: string;
  meetingUpdate: {
    meetingState?: MeetingStateV2; // <<-- same here
    meetingPermissions: MeetingPermissions;
  };
}

...

  // a typesafe version of the function respecting the new optional meetingState
  private convertUpdateMessageToV1(msg: UpdateMessageV2): UpdateMessage {
    const meetingState: MeetingState | undefined = msg.meetingUpdate.meetingState ? {
      ...msg.meetingUpdate.meetingState,
      isCameraOn: msg.meetingUpdate.meetingState.isVideoOn,
    } : undefined;
    return {
      ...msg,
      meetingUpdate: {
        ...msg.meetingUpdate,
        meetingState,
      },
    };
  }
```

In `controlMeeting.tsx`:

```TypeScript
  ...
  const updateState = (msg: UpdateMessage) =>
    setMeetingState({
      state: msg.meetingUpdate.meetingState ?? emptyState.state, // <<-- Use the empty state if undefined
      permissions: msg.meetingUpdate.meetingPermissions,
    });

  ...

  
```

# Support pairing from "Control Meeting" action
Now that we are able to bring up the control panel when not paired, there is no option to start the pairing process from the panel. The new Teams sends us an indicator we can utilize here:

In `meetingClient.ts`:

```TypeScript
export interface MeetingPermissions {
  canToggleMute: boolean;
  canToggleVideo: boolean;
  canToggleHand: boolean;
  canToggleBlur: boolean;
  canToggleRecord: boolean;
  canLeave: boolean;
  canReact: boolean;
  canToggleShareTray: boolean;
  canToggleChat: boolean;
  canStopSharing: boolean;
  canPair?: boolean; // <<-- This one is new in v2
}
```

In `controlMeeting.tsx` lets show all the actions when pairing is supported by teams, so that the user can select an action to initiate the pairing process:

```TypeScript
const isItemEnabled = (item: Item, state: State) => (state.state.isInMeeting || state.permissions.canPair) && state.permissions[item.permission];
```

# `README.md`
We need to update the setup instructions in the `README.md` beginning at line 26:

```Markdown
## Setup for new Teams
Follow these steps to use the extension with the new Microsoft Teams:

1. Go to the extension's settings in Raycast and set **Teams Version** to **New Teams**
2. In Microsoft Teams go to **Settings** → **Privacy** → **Manage API** and ensure the **Enable API** switch is turned on
3. When executing one of the extension's actions for the first time, while in a meeting, Microsoft Teams will bring up a prompt **New connection request** for the extension. **Allow** the extension and you're ready to go!

## Setup for Teams classic
...
```

pernielsentikaer:
Thanks for the extended review @svenwiegand 🔥

philippreissing:
thanks for the great review @svenwiegand ! I updated the code with all the changes

svenwiegand:
The PR looks good to me now.

Thanks @philippreissing for bringing in New Teams support. I’ve never found the time to do this. Thanks to your work I’m finally using the new teams since last Tuesdas :)

pernielsentikaer (approved):
🫶 Approved

PR #10323, opened 2024-01-24
[Day One] catch "sync issue" and show an error for it

AntonNiklasson:
## Description

Sometimes the Day One CLI gets out of sync somehow with the storage used by the desktop application. I haven't understood yet when or why this happens. This change just adds an extra layer of protection and a specific error message.

This is the error message I get: `The model used to open the store is incompatible with the one used to create the store`. Seems to be something in the iOS standard library related to storage, but I haven't found anything more specific.

Instead of saying "couldn't find the CLI" the extension now says "launch to desktop application to resolve this" instead. So hopefully less confusing for the users.

Screenshot 2024-01-26 at 11 49 27

### DIFF
--- a/extensions/day-one/CHANGELOG.md
+++ b/extensions/day-one/CHANGELOG.md
@@ -1,10 +1,10 @@
 # Day One Changelog
 
-## [Bug fix]: 2024-01-24
+## [Bug fix] - 2024-01-24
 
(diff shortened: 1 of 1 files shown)


pernielsentikaer (approved):
✅ Approved

PR #10333, opened 2024-01-25
Add wise-accounts extension

robsel118:
## Checklist


pernielsentikaer:
Hi 👋

I don't really know that much about wise, but we should maybe consider merging it into the current [Wise](https://www.raycast.com/hugoliconv/wise-quotes) extension by @HugoLiconV instead and make it a Wise extension so we don't confuse the end users using it?

Does that make sense?

robsel118:
> Hi 👋
> 
> 
> I don't really know that much about wise, but we should maybe consider merging it into the current [Wise](https://www.raycast.com/hugoliconv/wise-quotes) extension by @HugoLiconV instead and make it a Wise extension so we don't confuse the end users using it?
> 
> Does that make sense?

Hi, I did a separate one because the features are not overlapping and his code was not usable for my extension. To me they serve different purpose which is why I made a different one. But lmk your what you think is best

pernielsentikaer:
It would be nice to hear what @HugoLiconV thinks; it would be nice with one unified extension for Wise if possible - I guess most users would gain from all commands?

HugoLiconV:
Hi guys! Honestly, I have no idea what's the best approach. As @robsel118 said, both extensions have different purposes:

**Current Wise Extension**
- Get quotes to see the exchange rate and fees Wise offers.
- Get the current rate between USD and a target currency

**Wise accounts extensions**
- Get profiles
- Get balances
- Get transactions

As you can see the features don't overlap but I don't mind merging the extensions. My only concern is not making any of the extensions harder to use.

pernielsentikaer:
So let's keep them separated if that makes sense, thanks for the chip in @HugoLiconV 🙂

pernielsentikaer (approved):
✅ Approved

PR #10350, opened 2024-01-25
Add website-blocker extension

gersomvg:
## Description

This extension allows you to manage a list of distracting websites that you want to block. It stores these websites in LocalStorage and allows you to toggle blocking on and off by updating `/etc/hosts`.

Updating `/etc/hosts` requires an admin password, which is triggered by running the command through apple script with administrator privileges. The password is an inconvenience but it's necessary and similar to how tools like [SelfControl](https://github.com/SelfControlApp/selfcontrol/) handle it.

## Screencast

https://github.com/raycast/extensions/assets/8234283/5a6f29b0-3256-45f8-b90a-276ee460ec93

## Checklist


pernielsentikaer:
Hi @gersomvg 👋 

I update the extension a bit, I hope you don't mind 🙂

Could you please check that I didn't break something while doing it 🙂

gersomvg:
@pernielsentikaer I agree on the changes and I've verified that everything still works.

The only change that I (initially) disliked is the removal of the default action "Remove Host" on a selected host in the list.

I made it this way so that you could both add and remove hosts without using a shortcut, but by simply pressing <kbd>Enter</kbd>: if you type a non-existing host into search then <kbd>Enter</kbd> would add this host but when you select an existing host you could also simply press <kbd>Enter</kbd>.

I gave it a few minutes of thought and I can see why your change of default action is maybe a better default for the general public. It's probably unwise to have a destructive action under <kbd>Enter</kbd> while it can also be confusing that <kbd>Enter</kbd> does completely different things under context.

gersomvg:
I think having it secondary is enough for now. I would dislike having a confirmAlert on every removal. I sometimes quickly remove a few websites that I actually need to access for work (such as hacker news or youtube) and having the password prompt in order to sync is already more than enough friction :)

pernielsentikaer (approved):
Approved

gersomvg:
@pernielsentikaer How frequently is the online store synced with `main`?

pernielsentikaer:
@gersomvg Raycast clients asks for updates every fourth hour, the store is updated just after the update is merged

gersomvg:
@pernielsentikaer Maybe something went wrong then, since I can't yet find it through the Raycast website or the Raycast store extension?

pernielsentikaer:
Trying to push it again, it should almost be instant

pernielsentikaer:
Should be good now 🙂

gersomvg:
Yes, that fixed it 🙌🏻 Thanks!

pernielsentikaer:
Sorry for the inconvenience 🙂

PR #10351, opened 2024-01-25
GitHub: Add category preferences to select which pull requests to show in the menu bar

qeude:
## Description

I think most of us have some preferences on which pull requests we would like to have in the menu bar, that's why I added the possibility to select categories we want to include.

## Screencast

image
image
image

## Checklist

### DIFF
--- a/extensions/github/src/open-pull-request-menu.tsx
+++ b/extensions/github/src/open-pull-request-menu.tsx
@@ -105,6 +105,7 @@ function OpenPullRequestMenu() {
       {filteredSections().map((section) => {
         return (
           <MenuBarSection
+            key={section.type}
             title={section.type}
             maxChildren={getMaxPullRequestsPreference()}
             moreElement={(hidden) => (


thomaslombart (commented):
Thanks for the PR. It's a great addition. I don't see how the icon count categories are useful. I think the count should simply reflect the number of items you have in your menu bar.

For example, when I tested it, I only had 2 PRs for the icon count, but when I saw the list, I noticed 12 PRs, which was confusing. It would also simplify the preferences and the code.

qeude:
Thanks for your review @thomaslombart! 🙏
I think that the icon count depends a lot on individual needs. 
Personally, I lean towards having a count of PRs that require my attention—be it for a review, assignments, or mentions. It serves more like a notification badge for me, meaning I don't want notifications about the PRs I've opened. 
Yet, I still like the convenience of quickly accessing these PRs from the menu bar list. 
I understand, though, that preferences can differ; some folks might prefer a broader view, encompassing all ongoing PRs, regardless of whether they've been opened or are awaiting review. 
That's precisely why I've made it fully customizable.

qeude:
Is there anything else I can help you with so we can merge that PR @thomaslombart?

thomaslombart:
I was on vacation last week, so I couldn't reply. We can keep both preferences but they should at least have the same defaults so that it's not initially confusing. What do you think?

thomaslombart (approved):
🚀

PR #10372, opened 2024-01-26
add GitLab search groups command

macbookandrew:
## Description

This adds a new command to search GitLab groups. Previously you could view and browse through groups hierarchically, but not search them.

## Screencast

https://github.com/raycast/extensions/assets/784333/d5de6570-b7bb-4b72-be4d-c86d2defb3cb

## Checklist


tonka3000:
Hey @macbookandrew,

I tried the new command, but for me I get basically the exact same results, because I can already search subgroups. I think I missed something here or maybe my setup is to simple. Can you give me some hints.

macbookandrew:
@tonka3000 we have a self-hosted GitLab instance with multiple levels of groups:

- org-name
   - clients
     - client A
     - client B
     - client C
   - internal
- demo
- etc.

If I want to search for “client C”:

- Currently I’d have to navigate through org-name > clients before I can search
- With the new command, I can search for any group regardless of its level

tonka3000:
Did you turn on the `Flat List View` option?

CleanShot 2024-01-28 at 11 09 10@2x

macbookandrew:
@tonka3000 no, I didn’t realize that was an option! That works; thanks!

macbookandrew:
actually, I’m seeing a fairly significant performance difference between the two…it takes quite a while to load “My Groups” when not cached (roughly 10 seconds), whereas “Search Groups” is < 1s given a search term

tonka3000:
Interesting, I had the opposite 🤔.
I think this depends on which instance do you use

macbookandrew:
I’m an admin on a self-hosted instance running enterprise edition; currently our max project ID is in the 1500 range and max group ID is 2000+ (some of those are archived/moved).

Sorry…I was trying the “My Projects” command and _that_ took 10s. The “My Groups” still took about 8s just now.

Using the new “Search Groups” adds the search term to the API request so returns much more quickly compared to a cold cache for “My Groups,” where Raycast has to load everything.

Basically this is adding server-side filtering vs. the client-side filtering in “My Groups”

pernielsentikaer:
8 seconds sounds a bit long. If the new command can speed up the time, it might be a good thing. What do you think, @tonka3000? Will it break something?

tonka3000:
@pernielsentikaer It depends. We talking here about the initial speed of an command (first run). `Search Groups` should be more or less the same speed, maybe depend a little bit how fast the instance react and also send a lot less data.

`My Groups` will be a little slower in the first run because it fetch all groups and store them locally but starting with the second run it is much faster than `Search Groups` because it search locally and fetch data in the background, but users don't need to wait for it.

>Will it break something?

No. It are more or less the same commands and it is just how the data is fetched but they always end up in the same result.

So @macbookandrew case here is the initial load time for the first command. Not quite sure how to continue, fully understand @macbookandrew usecase. Maybe deactivate the caching via a user preference for `My Groups` could also be an alternative option.

I personally prefer a fast local search but I have no problem with the `Search Groups`. Can be confusing for user when it do the same thing, but maybe we adapt the command to be more different.

macbookandrew:
@tonka3000 looks like the [current cache TTL](https://github.com/raycast/extensions/blob/dc20b0c0f4650ef2f91e12ebeeac8bf162ce0ece/extensions/gitlab/src/components/groups.tsx#L184) for groups is 15 min, correct? Whereas the [project cache](https://github.com/raycast/extensions/blob/dc20b0c0f4650ef2f91e12ebeeac8bf162ce0ece/extensions/gitlab/src/components/project.tsx#L114) is 7 days.

I don’t think deactivating caching would help…but a significantly longer cache TTL would.

macbookandrew:
For what it’s worth, the extension already has distinct “My Projects” and “Search Project” commands, so this addition brings the set of commands for groups a bit closer to the set of commands for projects.

tonka3000:
@macbookandrew good point.

@pernielsentikaer Let's merge it

tonka3000 (approved):

Issue #10359, opened 2024-01-26
[Arc] Ability to search using Kagi


multivac61:
### Extension

https://www.raycast.com/the-browser-company/arc

### Description

[Arc] Ability to search using Kagi

bdsqqq:
looking into it!

PR #10373, opened 2024-01-26
feat(arc): Add The Kagi search engine as an option to Arc commands

bdsqqq:
## Description

Adds the Kagi search engine as an option to Arc commands that can search.
Feature requested in #10359 

## Checklist


pernielsentikaer:
I guess we can't do much kagi since it's behind API, right @bdsqqq

bdsqqq:
> I guess we can't do much kagi since it's behind API, right @bdsqqq 

I think so, tried finding something we could do for suggestions in their docs but no luck.

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

Let's go with this and see how it goes 🙂

extensions/arc/src/suggestions.ts
```diff
@@ -61,6 +61,11 @@ const config: SearchConfigs = {
       });
     },
   },
+  kagi: {
+    search: "https://kagi.com/search?q=",
+    suggestions: null, //Note: Unknown
```
dvanoni, line 66: Can this use `https://kagi.com/api/autosuggest?q=%s` as mentioned on https://kagi.com/assets?

bdsqqq: Looks like it can! don't have the time to look into it rn but will try in a few days. 
For now I opened #10461 so we can track it.

ty so much

PR #10383, opened 2024-01-27
Correct tranposition of "parent" and "the"

### LINKED ISSUE #10366: Typo: Transposed words "parent the action"
### Description

I believe the words "parent" and "the" are transposed on https://developers.raycast.com/basics/prepare-an-extension-for-store

image

> Don't repeat parent the action name ...

should be

```
Don't repeat the parent action name ...
```

https://github.com/raycast/extensions/blob/cc0ba2508ce384b9821376242fd3cca2d08d4c27/docs/basics/prepare-an-extension-for-store.md?plain=1#L248

salcode:
Change the text

"Don't repeat parent the action name ..."

to

"Don't repeat the parent action name ..."

Resolves #10366


pernielsentikaer (approved):
✅ Well spotted

PR #10398, opened 2024-01-28
feat: Add launch context support to slack-status extension

foobarnes:
## Description

Adds the ability to launch the `slack-status` extension with `launchContext` props. Specifically, this PR adds a param called `presetName` to allow users to open Quicklinks specifying which of their presets to launch, without having to navigate through the commands actions.

## Screencast

https://www.loom.com/share/780e2f3d9cce49cd98462a0fe0131bc3?sid=25190fd5-c844-4b5d-af6d-aa4007888a05

## Checklist

### DIFF
--- a/extensions/slack-status/src/setStatus.tsx
+++ b/extensions/slack-status/src/setStatus.tsx
@@ -34,21 +34,21 @@ function StatusList(props: LaunchProps<{ launchContext: CommandLinkParams }>) {
   const slack = useSlack();
 
   useEffect(() => {
-    if (props.launchContext?.presetId) {
-      const presetId = props.launchContext.presetId;
-      const presetToLaunch = presets.find((p) => p.id === presetId);
+    (async () => {
+      if (props.launchContext?.presetId) {
+        const presetId = props.launchContext.presetId;
+        const presetToLaunch = presets.find((p) => p.id === presetId);
 
-      if (!presetToLaunch) {
-        console.error("No preset found with id: ", presetId);
-        showFailureToast(new Error(`Could not find ID: "${presetId}" preset`), { title: "No preset found" });
-      } else {
-        setStatusToPreset({ preset: presetToLaunch, slack, mutate });
-        popToRoot();
-        closeMainWindow();
+        if (!presetToLaunch) {
+          console.error("No preset found with id: ", presetId);
+          showFailureToast(new Error(`Could not find ID: "${presetId}" preset`), { title: "No preset found" });
+        } else {
+          await setStatusToPreset({ preset: presetToLaunch, slack, mutate });
+          popToRoot();
+          closeMainWindow();
+        }
       }
-
-      return;
-    }
+    })();
   }, [slack]);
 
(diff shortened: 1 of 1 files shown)


foobarnes:
I can't seem to reproduce the `prettier` failure locally. If someone could let me know the issues, I'll fix them right away.

```sh
> npx ray lint                
ready -  validate package.json file
ready -  validate package-lock.json file
ready -  validate extension icons
ready -  validate extension metadata
ready -  run ESLint
ready -  run Prettier 3.0.3
ready -  no linting issues found
```

PitNikola:
@foobarnes thanks for the addition!

@pernielsentikaer, could you please take a look if everything works as expected. Otherwise, I'm fine.

pernielsentikaer:
Hi @foobarnes 👋

It seems like an awesome update 🔥

The only thing I can think of that might be a problem is if two presets have the same name, then it will just pick the first one, right? Do you have any good ideas regarding that?

foobarnes:
@pernielsentikaer Great callout! I didn't have an elegant solution for that issue. I couldn't think of a more-specific identifier that didn't involve adding complexity to the preset data structure / preset creation process.

I'm open to implementing a better solution here, but wasn't sure how common/useful presets with the same name were to other users. Open to discuss!

foobarnes:
If it helps you organize your enhancements, my next goals are:
- implement the presence API to go along with presets (e.g. set as away)
- pass `duration` as launchContext to override preset (e.g. `lunch 90` to set "Lunch" status for 90 minutes instead of saved 45 minutes

pernielsentikaer:
With the new actions, it might be easier to make the "Quicklink" and "Deeplink" - we could even add the full "preset" to ensure it's the right one. I'm not sure how often you'll have two with the same title, though. I guess it could happen.

pernielsentikaer:
So now we just need to figure out how much we want to do; we could serialize the object or just first add a block in the README about the naming (only one item with each name) - do you have any ideas?

foobarnes:
I think the sufficient first pass would be to add the note to the readme / instructions for the extension. If other users find that they need multiple statuses with the same name, the serialize approach (or some other unique-ify strategy) would be a good next step.

I'm battling my own perfectionism here, but I imagine we'll make bigger changes that impact the preset structure before we get feedback asking for same-name presets.

pernielsentikaer:
Hi @foobarnes 👋

I gave it another stab; could you check out this modification? Passing the launchContext is not always working for me (nothing happens), but it always works when you do it from the list. Do you experience the same and maybe know what's causing it, it feels like something is not correctly awaiting

pernielsentikaer:
Would like to hear some input and maybe also if you have the same issue as me? Sometimes it works, but it doesn't seem related to the code I changed - want to catch up on Slack @foobarnes

foobarnes:
Thank you for handling @pernielsentikaer! My startup team and I have been riding some major waves closing our fundraising round, so I've been MIA from some of my other passions.

PR #10414, opened 2024-01-28
Update resend extension

AlexIsMaking:
## Checklist


extensions/resend/package.json
```diff
@@ -41,6 +44,13 @@
       "subtitle": "Resend",
       "description": "View Logged Emails and Send A New Email",
       "mode": "view"
+    },
+    {
+      "name": "contacts",
+      "title": "Contacts",
+      "subtitle": "Resend",
+      "description": "View, create, update and delete contacts from your audiences.",
+      "mode": "view"
```
xmok, line 53: Maybe it would make more sense if we rename this to `Audiences and Contacts`? Currently, you are dealing only with Contacts but the API also supports Audiences now e.g. [Add Audience](https://resend.com/docs/api-reference/audiences/create-audience) so eventually we'll want to support Audience endpoints as well so easier to do it right now rather than change later.

extensions/resend/src/contacts.tsx
```diff
@@ -0,0 +1,258 @@
...
+
+  return error ? (
+    <ErrorComponent error={error} />
+  ) : (
+    <List
+      isLoading={isLoadingAudience || isLoadingContacts}
+      searchBarAccessory={<AudienceDropdown audiences={audiences} setAudience={setAudience} />}
+      actions={
+        <ActionPanel>
+          {audience && <Action.Push title="Create Contact" target={<CreateContact audience={audience} />} />}
```
xmok, line 105: Missing Icon.

extensions/resend/src/contacts.tsx
```diff
@@ -0,0 +1,258 @@
...
+          }`}
+          icon={
+            contact.unsubscribed
+              ? { source: Icon.Circle, tintColor: Color.Red }
+              : { source: Icon.CheckCircle, tintColor: Color.Green }
+          }
+          actions={
+            <ActionPanel>
+              {audience && (
+                <Action.Push title="Create Contact" icon={Icon.Plus} target={<CreateContact audience={audience} />} />
```
xmok, line 124: Missing Icon.

extensions/resend/src/contacts.tsx
```diff
@@ -0,0 +1,258 @@
...
+      actions={
+        <ActionPanel>
+          <Action.SubmitForm icon={Icon.Check} onSubmit={handleSubmit} />
+        </ActionPanel>
+      }
+    >
+      <Form.TextField title="Email" {...itemProps.email} />
+      <Form.TextField title="First Name" {...itemProps.first_name} />
+      <Form.TextField title="Last Name" {...itemProps.last_name} />
+      <Form.Checkbox label="Unsubscribed" {...itemProps.unsubscribed} />
```
xmok, line 216: Missing placeholders.

extensions/resend/src/contacts.tsx
xmok: Also, for consistency, maybe you could follow the way I've done it and automatically reload the Contacts list after a new Contact is added? e.g. [Line 98 from api_keys.tsx](https://github.com/raycast/extensions/blob/5bb43b534e7ab265cdfe370c09d88f68c142c5a5/extensions/resend/src/api_keys.tsx#L98). Right now, I need to re-run the entire command, fetch the Audiences and Contact again, in order to see the new contact.

xmok (commented):
Hi 👋.

Thank you for this update 🚀.

In addition to the feedback I've outlined, could you also:

- modify README to mention the new Commands
- update the screenshots to show your new command (I think we could replace `resend-4-add-new-domain` and `resend-6-view-logged-emails`)

I look forward to reviewing the new changes 💪.

Also

extensions/resend/package.json
AlexIsMaking: It might be better if we added a separate Audiences command, rather than adding more functionality to this one? That command could allow you to create, update and delete audiences, which would be a lot to fit into a single command. The only purpose of audiences for this command is allowing you to group / filter contacts.

AlexIsMaking:
I'm guessing that check failing is due to something not being installed correctly because it's affecting files that I haven't edited but I'm not sure what the issue is / how to solve it 😬

AlexIsMaking:
@xmok alright so I've just made all of the changes that you mentioned, apart from the one that I commented on + updating to the latest version of the Raycast API. I'm not sure why but changing the API version causes tons of linting issues so I'll leave that for you to take a look at, if that's ok?

pernielsentikaer:
We can always update the API  in a new PR - we might need new metadata images for everything, or can either one of you make new metadata images with the same background?

AlexIsMaking:
@xmok I might be able to use the same background for my metadata images, did you use one of the Raycast wallpapers?

pernielsentikaer:
I think it's [this one](https://www.raycast.com/uploads/wallpapers/floss.png) 🙂

AlexIsMaking:
@pernielsentikaer thanks! I wasn't sure where to look for that. Should be all set now 🤞

pernielsentikaer:
Awesome @AlexIsMaking 🔥 

@xmok do you have anything else we need to address?

xmok:
Looks great! LGTM 🚀

pernielsentikaer (approved):

Issue #10426, opened 2024-01-29
[Obsidian] Numerical Year/Month/Date templates


j-peeters:
### Extension

https://www.raycast.com/KevinBatdorf/obsidian

### Description

Hi, I was trying to store tasks in my Daily note with the 'Append task' function. 
I have structured my Daily Note folder like this:
`Vault/Timestamped/2024/01/20240129` for a Daily note for the 29th of January in 2024
When I use the current placeholders it translates to this
`Vault/Timestamped/2024/Jan/2024JanMon`
It would be great if there was a setting to use numerical instead of textual templates for date formats.
Maybe I'm missing something, then I would be glad if someone could elaborate.

### Who will benefit from this feature?

To be more flexible in having tasks stored in a specific daily note.

FServais:
I would suggest support of `{MM}` or `{DD}` could be added to templates?

I can take care of it if this approach seems fine.

KevinBatdorf:
Sure thing @FServais 🙌

The code for that is here: https://github.com/raycast/extensions/blob/46b603905f374e904bb2612830569488a6fb0a9c/extensions/obsidian/src/utils/utils.tsx#L284

j-peeters:
That addition would be really great. Thanks

AdamAdamsMusic:
Just a thought - what do you guys think about potential Luxon/Moment.js integration for support for any kind of date/time templates? Maybe this could be the solution for date/time template parsing?

KevinBatdorf:
Yeah that would probably work too. [date-fns](https://date-fns.org/docs/format) is another alternative

FServais:
I was planning to use `luxon` *in addition* to the templates already supported today.

I have the diff but my workspace is in a messed up state seems like, build dist fails and publish fails as well, I'll try to troubleshoot this so we can have code review 👍

FServais:
@raycastbot close this issue

PR #10452, opened 2024-01-30
Support Luxon formatting tokens in templates (yyyy, MM, dd)

FServais:
## Description

Related to issue https://github.com/raycast/extensions/issues/10426.

This PR adds the support of date and time formats as defined by [Luxon](https://moment.github.io/luxon/#/formatting?id=table-of-tokens).

As such, this change refactos the way templating is done:
- Replace `replaceAll` functions by `.replace(regex, matchFun)`.
- `regex` will check all text within `{` and `}`.
- The match function will return the same return values for keywords already supported (`day`, `year`, `timestamp`,...).
- If `regex` matches format from Luxon (`dd`, `MM`,...), it will defer the call to `dateTime.toFormat(key);`.
 
## Screencast

Screenshot 2024-01-29 at 23 56 39

Screenshot 2024-01-29 at 23 57 05

Screenshot 2024-01-30 at 16 03 56

## Checklist

### DIFF
--- a/extensions/obsidian/src/utils/utils.tsx
+++ b/extensions/obsidian/src/utils/utils.tsx
@@ -284,111 +284,113 @@ export async function applyTemplates(content: string, template = "") {
     ? template // Has {content} e.g. | {hour}:{minute} | {content} |
     : template + content; // Does not have {content}, then add it to the end
 
-  return preprocessed.replace(/{.*?}/g, (match) => {
-    const key = match.slice(1, -1);
-    switch (key) {
-      case "S":
-      case "u":
-      case "SSS":
-      case "s":
-      case "ss":
-      case "uu":
-      case "uuu":
-      case "m":
-      case "mm":
-      case "h":
-      case "hh":
-      case "H":
-      case "HH":
-      case "Z":
-      case "ZZ":
-      case "ZZZ":
-      case "ZZZZ":
-      case "ZZZZZ":
-      case "z":
-      case "a":
-      case "d":
-      case "dd":
-      case "c":
-      case "ccc":
-      case "cccc":
-      case "ccccc":
-      case "E":
-      case "EEE":
-      case "EEEE":
-      case "EEEEE":
-      case "L":
-      case "LL":
-      case "LLL":
-      case "LLLL":
-      case "LLLLL":
-      case "M":
-      case "MM":
-      case "MMM":
-      case "MMMM":
-      case "MMMMM":
-      case "y":
-      case "yy":
-      case "yyyy":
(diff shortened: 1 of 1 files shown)


KevinBatdorf:
I think they all need to run after `{content}` is processed so that one needs to be broken out. 

Was it not possible to just add another replaceAll with a regex after content (or at the end) that just runs format? Maybe like: 


```js
  return preprocessed
    .replaceAll("{content}", content)
    .replaceAll(/{(.*?)}/g, (_, c) => dateTime.toFormat(c))
    // rest
```

FServais:
Oh sure, I didn't get `{content}` could be broken out. I'll add it.

Regarding `replaceAll`, I am unsure about `replace` vs `replaceAll` while using a regex. From what I got, they are similar in this case.

Regarding all the different `case`, I wanted to have a safeguard before hitting `dateTime.toFormat(c)`, as to be able to control what we give to `luxon`, and also such that anything not matching will be kept. For instance, I think using `{mytasks}` should be left intact and not sent to `toFormat`.

With those reasons, I thought we could refacto altogether to not to keep a mix of `.replaceAll({...})` and `case`, might as well all put in the `switch` for coherence.

Let me know what you think, I can adapt this of course.

KevinBatdorf:
Just might be easier to maintain if it's a one liner. Could put it at the end I guess? What happens if it captures an unsupported value like {foo}? Does it throw or just move on?

But outside of {content}, what you have now is fine too. I'm not sure those values ever change anyway. It's your call.

FServais:
> What happens if it captures an unsupported value like {foo}? Does it throw or just move on?

From the few tests, it parses it in unexpected ways, but does not throw an error. My thought of the expectation was that if a user wants to uses any text in `{...}` that is not defined in the documentation (extension + luxon), then it should not be parsed. 

I agree with maintaining, although I anticipate that should be really rare. 

I'll keep that switch as it is for this commit, but let me know if you still think the use case above is not relevant. 👍

KevinBatdorf:
> My thought of the expectation was that if a user wants to uses any text in {...} that is not defined in the documentation (extension + luxon), then it should not be parsed.

Yeah if it encountered {foo} I was assuming it would just ignore it. I guess it breaks it apart and matches characters within it. 

> I'll keep that switch as it is for this commit

Sounds good.

pernielsentikaer:
Is this ready for review before shipping, or do we still await something?

pernielsentikaer (approved):
🟢 Approved

j-peeters:
Thanks for implementing my request @FServais and to the others for their critical eye!

Issue #10162, opened 2024-01-15
[Linear] Can not search projects


nnphong1904:
### Extension

https://www.raycast.com/linear/linear

### Description

 I am unable to find any projects through my search, but I can still locate them if I have marked them as favorites on the linear.app site.
![image](https://github.com/raycast/extensions/assets/45620830/d8774b5d-093e-406c-8aa2-1f5d4dbcaa34)
![image](https://github.com/raycast/extensions/assets/45620830/0e9e0338-a082-4d1d-8c93-ce9bc6500fb4)

### Steps To Reproduce

1. Install linear extension
2. Sign in with Linear
3. Try to search any project

PR #10457, opened 2024-01-30
[Linear] Fix projects search not working

giuseppesalvo:
## Description

Fixes https://github.com/raycast/extensions/issues/10162

- If there are no roadmaps, all the projects are now shown
- Similarly, all the projects issues are now shown if there are no milestones

## Checklist

### DIFF
--- a/extensions/linear/CHANGELOG.md
+++ b/extensions/linear/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Linear Changelog
 
+## [Fix projects search not working] - 2024-01-30
+
+- The projects search was not working in case there were no roadmaps available. This is now fixed.
+
 ## [Add Favorites command] - 2024-01-05
 
 - Add `Favorites` command to browse your Linear favorites right from Raycast.


extensions/linear/src/components/ProjectList.tsx
```diff
@@ -25,7 +25,7 @@ export default function ProjectList() {
       return [];
     }
 
-    if (roadmap === "") {
+    if (!roadmap || !roadmaps || roadmaps.length < 1) {
```
giuseppesalvo, line 28: The issue was a bit subtle:
- `roadmap` can be `undefined`, so this string equality check would be evaluated to false, even if there was no `roadmap` to filter the projects.

I also added a check for the roadmaps. It seems conceptually correct to check if there are roadmaps available, instead of only relying on the `roadmap` filter being implicitly invalid if there aren't any roadmaps.

thomaslombart (approved):
Awesome! Thanks for the fix.

giuseppesalvo:
Thanks! I just linked my account 🙂

PR #10511, opened 2024-02-02
New extension: Perplexity-API

maximilianzuern:
## Description

This extension provides user access to all models through the [Perplexity API](https://docs.perplexity.ai/docs/getting-started).

There is a Raycast [perplexity extension](https://github.com/raycast/extensions/tree/main/extensions/perplexity) but it just opens query links in the browser for faster access to Perplexity web interface. 

This extension, though, gives user access to open-source models not offered on the web interface. That's why I built a new extension rather than updating the existing one because they serve different purposes.

### Selected Text Commands

- Change tone to casual
- Change tone to friendly
- Change tone to professional
- Comment code
- Compose Message/Mail responses
- Explain selected text
- Fix grammar and spelling
- Improve selected text
- Make selected text longer
- Make selected text shorter
- Summarize selected text
- Find synonyms according to the selected word
- Break down project description into tasks
- Translate selected text

### Quick AI
- Ask AI a one-shot question

### Supported Models

* codellama-34b-instruct
* codellama-70b-instruct
* llama-2-70b-chat
* mistral-7b-instruct
* mixtral-8x7b-instruct
* pplx-7b-chat
* pplx-70b-chat

## Screencast

prompt
askAI
fix
summary

Additionally, it calculates the cost per request, as displayed in the toast notification.

## Checklist


pernielsentikaer:
I really think it would make sense to only have one Perplexity extension - what do you think, @third774?

This seems like a far better way to work with perplexity

third774:
I'm hopeful that one day my extension can be deprecated/removed — especially with the upcoming work y'all are doing to directly integrate Perplexity with Raycast. That said, I would argue that the current extension is still _very_ useful in a few ways:
- No API key required!
- Can make use of Perplexity's copilot mode
- Can make use of Perplexity's focus modes

I don't know if these are available via the API access to models are they? Often, I also like to inspect the search results that are surfaced in the web UI. I think we'd be losing that as well, wouldn't we?

I'm def not opposed to merging these two extensions though if y'all think that's best! It's just that personally I think I'm going to keep using the web UI until those features are available directly in Raycast.

pernielsentikaer:
What about we merge them so we can keep both yours and all the new stuff? It will require that content of the PR is added on top of the current extension so it's "just" an update 🙂

maximilianzuern:
>     * No API key required!
> 
>     * Can make use of Perplexity's copilot mode
> 
>     * Can make use of Perplexity's focus modes
> 
> 
> I don't know if these are available via the API access to models are they? Often, I also like to inspect the search results that are surfaced in the web UI. I think we'd be losing that as well, wouldn't we?

Exactly! None of those features are accessible through the API. Moreover, the web UI uses models from OpenAI, Anthropic and Google, whereas the API relies on models from Meta, Mistral, and Perplexity.

> What about we merge them so we can keep both yours and all the new stuff? It will require that content of the PR is added on top of the current extension so it's "just" an update 🙂

The point is, the two extensions serve different purposes. Individuals seeking access to GPT3.5 or GPT4 prefer the web UI, whereas those who favor open-source models would use my extension.
Still, I get why you might consider merging them. In this case, should I modify the PR?

pernielsentikaer:
Sounds like it's better to keep it as two separate, as long as the naming clearly shows it 🙂

extensions/perplexity-api/CHANGELOG.md
```diff
@@ -0,0 +1,3 @@
+# perplexity-api Changelog
```
pernielsentikaer, line 1:
```suggestion
# Perplexity API Changelog
```

extensions/perplexity-api/README.md
```diff
@@ -0,0 +1,12 @@
+# perplexity-api
```
pernielsentikaer, line 1:
```suggestion
# Perplexity API
```

extensions/perplexity-api/package.json
```diff
@@ -0,0 +1,229 @@
...
+          "type": "text",
+          "required": false
+        }
+      ]
+    }
+  ],
+  "preferences": [
+    {
+      "description": "Perplexity API Key",
+      "name": "apiKey",
```
pernielsentikaer, line 166: Maybe add a `placeholder`

pernielsentikaer (changes requested):
Hi 👋 

I have now tested your extension, and I have some feedback ready for you 

extensions/perplexity-api/package.json
maximilianzuern: Added via 587fe7f

pernielsentikaer:
Would it make sense to make some of the commands disabled by default? Like we do in our [Spotify](https://www.raycast.com/mattisssa/spotify-player) extension 🙂

maximilianzuern:
> Would it make sense to make some of the commands disabled by default? Like we do in our [Spotify](https://www.raycast.com/mattisssa/spotify-player) extension 🙂

Thank you! I was unaware of the property.
I disabled some commands.

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #10551, opened 2024-02-04
Update hide-all-apps extension: speed up hiding of other applications

andrewsardone:
## Description

I've noticed that the extension hangs in the Finder for a second or two before hiding the other applications' windows. I believe that's because of AppleScript's iteration of `processes` and imperatively setting their visibility to `false`.

We can speed this up by issuing the native `Hide Others` command through its <kbd>⌥ ⌘ H</kbd> keyboard shortcut. The trade-off is that the command won't work for users who override Finder's <kbd>⌥ ⌘ H</kbd> shortcut to do some other command.

Apologies for not providing any specific benchmarking data via `console.time()` or something. I couldn't get any console statements to appear in my terminal. I'm not sure if that's a `ray develop` shortcoming or user error (likely the latter).

### Before

_Note that the other applications' windows stay hanging around for a second after invoking the command._

https://github.com/raycast/extensions/assets/14554/ad2a828d-8698-41c4-8b47-610e8edf1942

### After

https://github.com/raycast/extensions/assets/14554/51cd5124-8521-44f8-811e-595e9081b7d1

## Checklist

### DIFF
--- a/extensions/hide-all-apps/src/index.tsx
+++ b/extensions/hide-all-apps/src/index.tsx
@@ -10,8 +10,8 @@ export default async function Command() {
 
   await runAppleScript(`
 activate application "Finder"
-tell application "System Events"
-    keystroke "h" using {command down, option down}
+tell application "System Events" to tell process "Finder"
+    click menu item "Hide Others" of menu 1 of menu bar item "Finder" of menu bar 1
 end tell
 tell application "Finder" to set collapsed of windows to true
   `);


peduarte:
Thanks for this! Really appreciate it 🚀 

> The trade-off is that the command won't work for users who override Finder's ⌥ ⌘ H shortcut to do some other command.

Good point. Is it worth adding some sort of [preference](https://developers.raycast.com/api-reference/preferences) for this? I dont know what we could call it though, but essentially something that users can check, and internally we use either this strategy or what we were using before.

andrewsardone:
> Is it worth adding some sort of [preference](https://developers.raycast.com/api-reference/preferences) for this? I dont know what we could call it though

That's a good question, as it's one I'm struggling to answer. I feel like a preference would be overkill, and/or confusing; it's exposing an inner implementation detail. But I'm also hesitant about potentially invoking an unknown command. I'll see if there's some automatic detection that can be done. I'll think on it. Thanks!

peduarte:
> But I'm also hesitant about potentially invoking an unknown command


yeah, my thoughts too!

peduarte:
What should we do about this?

andrewsardone:
> What should we do about this?

I just pushed up a new change (https://github.com/raycast/extensions/pull/10551/commits/5ede3a643818a4a0d1ebfedb7f709cad6b9b7172), which does not rely on a consistent keyboard shortcut while maintaining the performance improvement. We're still using the native `Hide Others` command, but now invoking it through AppleScript's menu bar commands.

I think this should be good to go ✅

peduarte:
Love it! Let's take it for a spin. Thanks!

Issue #8790, opened 2023-10-13
[1Password] is not synchronized with the client.


superwhys:
### Extension

https://www.raycast.com/khasbilegt/1password

### Description

when i add a new entry in client or web page, the entry I added cannot be queried using this extension.

### Steps To Reproduce

1. add a new entry in client or web page(add a password with title "testPwd").
2. open Raycast window and select the 1password extension
3. query title "testPwd" in extension
4. can not query the entry i just added

### Current Behaviour

can not query the entry i just added

### Expected Behaviour

query the entry i just added

superwhys:
At first I thought the time was too short and it didn't sync up, but after I tried to wait for a few days, it still couldn't query

Davidwalser:
@superwhys could you resolve the issue?

radualexe32:
+1

I have the same issue. To get all my passwords synced I would need to restart the integration process from scratch

solid-pixel:
same here, acknowledgment from the maintainer would be very appreciated.

PR #10556, opened 2024-02-04
[1password] fix synchronization issues with the client in v8 version

superwhys:
## Description
- fix the problem of synchronization with the 1password client in v8 verison

## Checklist

### DIFF
--- a/extensions/1password/src/v8/utils.ts
+++ b/extensions/1password/src/v8/utils.ts
@@ -58,30 +58,50 @@ export function op(args: string[]) {
   throw Error("1Password CLI is not found!");
 }
 
+const CACHE_TIMEOUTS: { [key: string]: number } = {
+  [CATEGORIES_CACHE_NAME]: 1000 * 60 * 10,
+  [ITEMS_CACHE_NAME]: 1000 * 60 * 10,
+  [ACCOUNT_CACHE_NAME]: 1000 * 60 * 10,
+};
+
+const DEFAULT_CACHE_TIMEOUT = 1000 * 60 * 10;
+
 export function useOp<T>(args: string[], cacheKey?: string) {
   const [data, setData] = useState<T>();
   const [error, setError] = useState<unknown>();
   const [isLoading, setIsLoading] = useState<boolean>(true);
+  const lastUpdatedKey = `${cacheKey}_lastUpdated`;
 
   useEffect(() => {
-    if (cacheKey && cache.has(cacheKey)) {
-      setIsLoading(false);
-      return setData(JSON.parse(cache.get(cacheKey) as string));
-    }
-
-    try {
-      const items = op([...args, "--format=json"]);
-
-      if (cacheKey) {
-        cache.set(cacheKey, items);
-        return setData(JSON.parse(cache.get(cacheKey) as string));
+    const fetchData = async () => {
+      try {
+        const items = op([...args, "--format=json"]);
+        if (cacheKey) {
+          cache.set(cacheKey, items);
+          cache.set(lastUpdatedKey, Date.now().toString());
+        }
+        setData(JSON.parse(items));
+      } catch (error: unknown) {
+        setError(error);
+      } finally {
+        setIsLoading(false);
       }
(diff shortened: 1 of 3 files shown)


pernielsentikaer:
Hi 👋

@khasbilegt do you mind checking this?

extensions/1password/src/v8/utils.ts
```diff
@@ -58,30 +58,56 @@ export function op(args: string[]) {
...
+      } finally {
+        setIsLoading(false);
+      }
+    };
 
-      if (cacheKey) {
-        cache.set(cacheKey, items);
-        return setData(JSON.parse(cache.get(cacheKey) as string));
+    if (cacheKey && cache.has(cacheKey)) {
+      const lastUpdatedString = cache.get(`${cacheKey}_lastUpdated`) as string;
```
cb372, line 93: I don't think this will work correctly the first time it runs on an existing install, where the cache contains a `cacheKey` element but `${cacheKey}_lastUpdated` doesn't exist.

We could guard against that by checking for both keys in the line above:

```ts
if (cacheKey && cache.has(cacheKey) && cache.has(`${cacheKey}_lastUpdated`)) {
```

extensions/1password/src/v8/utils.ts
```diff
@@ -58,30 +58,56 @@ export function op(args: string[]) {
...
+      const lastUpdated = parseInt(lastUpdatedString, 10);
+      const timeSinceLastUpdate = Date.now() - lastUpdated;
+      const cacheTimeout = CACHE_TIMEOUTS[cacheKey] || DEFAULT_CACHE_TIMEOUT;
+
+      console.log(
+        `${cacheKey} last udpated is: ${lastUpdated}, time since last update: ${timeSinceLastUpdate}, specify cache timeout: ${cacheTimeout}`
```
cb372, line 99: typo

```suggestion
        `${cacheKey} last updated is: ${lastUpdated}, time since last update: ${timeSinceLastUpdate}, specify cache timeout: ${cacheTimeout}`
```

extensions/1password/src/v8/utils.ts
```diff
@@ -58,30 +58,56 @@ export function op(args: string[]) {
...
-
-    try {
-      const items = op([...args, "--format=json"]);
+    const fetchData = async () => {
+      setIsLoading(true);
+      try {
+        const items = op([...args, "--format=json"]);
+        if (cacheKey) {
+          cache.set(cacheKey, items);
+          cache.set(`${cacheKey}_lastUpdated`, Date.now().toString());
```
cb372, line 81: Worth pulling this `${cacheKey}_lastUpdated` string into a const, as it's used multiple times? Just to make sure we're consistently using the same key both for reading and writing.

cb372 (commented):
LGTM, just a few minor comments

extensions/1password/src/v8/utils.ts
superwhys: It does have some problems, but it will works correctly.
when `${cacheKey}_lastUpdated` doesn't exist, it will be `Nan`in`lastUpdated` and `timeSinceLastUpdate`.
it will always get `false` in `if (timeSinceLastUpdate < cacheTimeout)` .

but I still fix it

extensions/1password/src/v8/utils.ts
superwhys: These logs will be removed

pernielsentikaer:
I just gave this a spin and I'm not sure how I can test it? For me it seems to do the same in both this and the store version?

Is there a way I can test it?

superwhys:
> I just gave this a spin and I'm not sure how I can test it? For me it seems to do the same in both this and the store version?
> 
> Is there a way I can test it?

This `PR` is designed to solve the problem in this [issue](https://github.com/raycast/extensions/issues/8790).

you can add a new entry in client or web page. 

in older versions of this extension, it won't sync the information you just added in 1password, but in my fix, it will.

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #10599, opened 2024-02-05
Update create-your-first-extension.md

luiscobot:
## Description

Update the command name, replace `Render Markdown` with `Show Detail`.

CleanShot 2024-02-05 at 15 39 21@2x


pernielsentikaer:
That's a tricky one since in the screenshot it shows that the name is also "Hello World," which should not be changed at the root of a command, only when pushing 🫠

pernielsentikaer:
We'll soon update images and will make sure to update the text here too, thanks for tipping 🔥

PR #10615, opened 2024-02-06
[Day One] support a complete timestamp instead of just dates

### LINKED ISSUE #10604: [Day One] Add timestamp to the entry
### Extension

https://www.raycast.com/AntonNiklasson/day-one

### Description

Currently, when creating an entry for Day One from Raycast, only the date is passed to Day One. Going through the Day One CLI documentation, looks like adding timestamp should be fairly straightforward.

This will help remove the need to keep the menubar item for Day One active for quick entries. The point of using Day One from Raycast is to use it as a micro-journal with multiple short entries in a day. Without the timestamp the sequence of these entries is not maintained and makes it difficult to use the feature.

### Anything else?

_No response_

AntonNiklasson:
## Description

Instead of just capturing dates for the entries, this adds support for including the time in that date. These are the changes:

- Use `--isoDate` instead of `--date` in the underlying CLI call
- Change `type` on `Form.Datepicker` from `Date` to `DateTime`

Fixes #10604 🎈

### DIFF
--- a/extensions/day-one/CHANGELOG.md
+++ b/extensions/day-one/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Day One Changelog
 
+## [Feature] - 2024-02-06
+
+- Capture a complete timestamp when creating an entry, instead of just the date
+
 ## [Bug fix] - 2024-01-24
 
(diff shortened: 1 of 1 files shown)


mittuled:
@raycastbot Can I review this PR? Looks fairly straightforward. (New here)

mittuled (approved):
Looks good. Straightforward changes.

pernielsentikaer (approved):

PR #10619, opened 2024-02-06
Add share-my-code extension

jairaume:
## Description
This is an extension to easily share **code snippets**, links, texts between people by sharing a simple slug. 

Taking advantage of the already existing service [**ShareMyCode**](https://sharemycode.fr), this extension simplifies the process by avoiding users to open their web browsers, fiddle with the url to get the correct content, etc...

It also **tracks** recently created  snippets and stores them. Any snippet can be also added to the _recents_ list to be kept.

> ⚠️ I am not the owner of [**ShareMyCode**](https://sharemycode.fr) and am simply porting this service as an extension to Raycast.

## Checklist


jairaume:
> Hi 👋
> 
> I just get this error when I test it, does it work for you?
> 
> share-my-code 2024-02-12 at 12 38 14

Hi 👋

Sorry for the inconvenience, there indeed seems to be an error when using random slug :(
I am working on it to try and figure it out asap! 
Should i convert this pr to draft for now while I investigate ? 😁

Thanks!

pernielsentikaer:
Let's convert it, let me know if you need any help

jairaume:
Hi again 👋

I finally found the cause of the bug !
It should work fine now with some small featured I added along the way :)

Let me know if I can/have to do anything 🙂

pernielsentikaer (changes requested):
Thanks for the corrections 🔥 

I would popToRoot() with a toast after submitting the form instead of the push, when you hit `escape` you end up in the previous view where you can submit data with the same slug again

What do you think?

jairaume:
Thanks so much for the feedback ! 

Indeed hitting `escape` allows for modifications of the content which might be confusing and totally be an unexpected behavior 🥲

I think what lead me to this choice was that I thought it would be interesting to give the user a preview of what he just shared without him having to launch another command to check if everything went well!

I will implement your suggestions right-away but in the mean time I had a small concern about a choice I made and I would love to have your thoughts on it : 
Is it ok to copy some content (here the link of the *ShareMyCode* he just created) to the user's clipboard without a mindful action from him (here it's triggered on form submission)?

Thanks 🙂

pernielsentikaer:
You could always make it a preference, this way the user can disable(or enable) if they want 🙂

jairaume:
Didn't even consider that, thanks!
Changes have been made now 🙂

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #10628, opened 2024-02-07
Replace icons with hi-res ones

oasaleh:
## Description

Just swapping icons with high resolution ones.


pernielsentikaer:
An image was missing in the PR, I update the code to use `tintColor` instead 🙂

pernielsentikaer:
Could you test the corrections @oasaleh

@mooxl are you okay with the icon change? 🙂

pernielsentikaer:
You only added [three images](https://github.com/raycast/extensions/pull/10628/commits/742c3e81eb8d3f92c18aef1441010bdf469d4be8) but referenced four in this PR 🙂

You should be able to pull my changes since they're are made on `oasaleh:replace-icons`

pernielsentikaer:
I can see that; you're right 🙂 but then you would have bad contrast in dark mode, showing dark gray on, e.g., a black background.

How did you add this PR? With the publish command? If so, there is a pull-contributions command too 🙂

pernielsentikaer:
Let me check how it look again, it makes sense - I can test it 🙂

pernielsentikaer:
I can see why you wanted to do that, but what if your background image is almost 100% the same color as the icon?

I tested something similar (not the exact same colors, but I imagine that for some people, it might be almost invisible)

Raycast07022024-oPXtu5b0 png@2x
Raycast07022024-xa47qo2x png@2x

Raycast07022024-J7BHmDTg png@2x
Raycast07022024-nDCpPKKc png@2x

pernielsentikaer:
The contrast in the image you shared is really bad and might not work for color blind people

Raycast07022024-m7n5SUc2 png@2x

pernielsentikaer:
Here are three images of the original `empty` image on different backgrounds, I can't really see it 🫠

Raycast07022024-P2QzfEj6 png@2x

Raycast07022024-R7g18hUT png@2x

Raycast07022024-tHIDwvNa png@2x

pernielsentikaer:
Like this
Raycast08022024-OO6Tz9bh png@2x

Raycast08022024-4Ok2i7z9 png@2x

Raycast08022024-vuoU2Nn9 png@2x

Raycast08022024-LRC9TWkx png@2x

pernielsentikaer:
It will follow almost every other menubar item (white/black)

Raycast09022024-EmzbsgDX png@2x

Raycast09022024-xITgoDlX png@2x

pernielsentikaer (approved):

ckyoung:
Hi, I don't know if it's just me or because of my 2k monitor, but the new menu icon kinda looks like a hand grenade to me lol. I even restarted it because I thought there's something wrong with the app.

Anyway, it's not really a big of an issue. Thanks for the app, it works great!

pernielsentikaer:
What if we could make the user select a few presets of icons from preferences? Would you be up for that @oasaleh 🙂

PR #10643, opened 2024-02-07
[pipe-commands] create separate clipboard action

iloveitaly:
## Description
Right now, the text action uses the selected text and then if it can't be extracted, uses the clipboard contents.

There is no way to force the clipboard contents to be used. I find this really annoying because most of the time I don't want to use the selected text, but actually want to use the clipboard contents.

This change extracts out the clipboard logic into a separate command which can be invoked independently. The downside here is it is a breaking change because the text command now just uses the selected text and does not fall back to clipboard.

## Screencast

NA

## Checklist


iloveitaly:
Not a bad idea! I like it. Let's see what @pomdtr thinks.

extensions/pipe-commands/src/types.ts
```diff
@@ -35,5 +35,5 @@ export type ScriptArgument = {
   percentEncoded?: boolean;
 };
 
-export type InputType = "text" | "file" | "url";
+export type InputType = "text" | "file" | "url" | "clipboard";
```
pomdtr, line 38: I don't think the clipboard input is necessary. What does it add on top of text ?

pomdtr:
Thanks for your contribution!

I agree about asking for the user to select either the selection or the clipboard when triggering a pipe command, but please address my two comments before merging this :)

pernielsentikaer:
Preferences sound like the right thing. How else would you do it in a good way, @pomdtr?

pomdtr:
I would just use two commands (as you have already done):

- pipe clipboard to command
- pipe text to command

pernielsentikaer:
That's good to me, but what about adding two new ones then? So users of the current command have the same behavior as always, and then you can, if you want, explicitly choose either paste or clipboard.

pernielsentikaer:
What do you think of that @pomdtr @iloveitaly - so current users still has the current behaviour as they know

pomdtr:
I'm not in favour of it. The current behaviour is not working well, and having a bunch of command will be confusing for new users.

pernielsentikaer:
We can proceed with that. I understand that it doesn't change the name for one of the commands (which is good), so people will still have the hotkey/alias/ranking. This command, is it the one most people will use?

pomdtr:
Yeah it is the main one.

extensions/pipe-commands/src/types.ts
pernielsentikaer: Could you check this @iloveitaly

extensions/pipe-commands/src/pipe-to-command.tsx
```diff
@@ -198,20 +193,25 @@ function PipeCommand(props: { command: ScriptCommand; inputFrom?: InputType; onT
...
   chmodSync(command.path, "755");
+
   const { stdout, stderr, status } = spawnSync(command.path, args, {
     encoding: "utf-8",
     cwd: command.metadatas.currentDirectoryPath
       ? untildify(command.metadatas.currentDirectoryPath)
       : path.dirname(command.path),
     input: command.metadatas.mode === "pipe" ? input : undefined,
     env: {
+      // TODO this is a lttle scary and doesn't seem like best practice?
```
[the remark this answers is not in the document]
pernielsentikaer: Could you check this @iloveitaly

gasolin (approved):
like the idea to have dedicated clipboard pipe command

pomdtr (approved):

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #10668, opened 2024-02-09
Update calendar extension

fuksman:
## Description

Allowed view mode switching: "compact" for the old-style layout and "wide" for the new table-based layout

## Checklist

### DIFF
--- a/extensions/calendar/package.json
+++ b/extensions/calendar/package.json
@@ -5,12 +5,8 @@
   "description": "View calendar month-by-month",
   "icon": "command-icon.png",
   "author": "fuksman",
-  "contributors": [
-    "pernielsentikaer"
-  ],
-  "categories": [
-    "Productivity"
-  ],
+  "contributors": ["pernielsentikaer"],
+  "categories": ["Productivity"],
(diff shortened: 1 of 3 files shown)


extensions/calendar/package.json
```diff
@@ -47,6 +43,24 @@
       "title": "Week numbers",
       "description": "Indicated if it is necessary to show week numbers",
       "default": true
+    },
+    {
+      "name": "viewMode",
+      "type": "dropdown",
+      "required": true,
```
pernielsentikaer, line 50: Do you mind if we set this to `false` so it doesn't require the user to actively choose? I guess you're not a fan of the wide view? 🙂

fuksman: Is it okay to leave it as it is? There are already two required preferences, so new users have to deal with them anyway and will have a chance to know about different view modes.
Actually, I've just gotten used to the compact view and trying to navigate in the wide calendar requires more cognitive powers. I've also received the same feedback from other users via email.

pernielsentikaer (approved):

PR #10722, opened 2024-02-12
Update pomodoro extension

MartinGonzalez:
## Description

This pull request adds slack support to update slack status when using pomodoro intervals.
2 new commands were added to not modify existing pomodoro commands.

## Screencast

https://github.com/raycast/extensions/assets/4130008/6467d10f-894f-45fb-9590-0161b5e928e5

## Checklist

### DIFF
--- a/extensions/pomodoro/CHANGELOG.md
+++ b/extensions/pomodoro/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## [Enhancement] - 2024-02-27
 
-Added support for slack to change status during interval
+Added support for slack to change status during interval, the two new commands are disabled by default
 
 ## [Enhancement] - 2024-02-16
 


pernielsentikaer:
Thanks for adding this. I think it's adding a lot of value, and I do think that you're right about having it as separate commands works.

Could you check this, @KELiON?

MartinGonzalez:
@KELiON just pinging

KELiON (changes requested):
Thanks for contribution! I like your idea and probably I'm going to use this integration, but I'd like to keep pomodoro features as simple as possible. Slack integration won't be used by all pomodoro users, some users would like to see similar integrations with other services, so it could add more complexity to the extension.

As an option, we can add this integration but make it optional. Can you please add [disabledByDefault](https://github.com/raycast/extensions/blob/1e557701ac0a5ab40004dd333d5dccd4725e81f9/docs/information/manifest.md?plain=1#L61) to new commands? This way everyone who want to use slack integration can enable these commands, but "default" pomodoro extension won't show them

pernielsentikaer:
Thanks for the correction, let's merge it 🙂

PR #10778, opened 2024-02-15
Arc: Add Spaces argument to 'New Window' and 'New Tab' commands

jayeshbhole:
## Description

Hey folks,
I felt the need for the Arc commands such as `new-window`, `new-tab` to have a `spaces` argument. In the changes I've proposed, the browser simply focuses onto the specified space before executing the desired command.

As a validation step, the commands also now pulls the existing spaces from Arc and validates the given space title or the space id (which btw is a number and is wrongly typed as a string in the `Space` type).

I feel like this could be an awesome addition to the command quick links where an user would like to open some link in a specific space rather than the one that's currently focused.

## Issues

I had raised the previous [PR](https://github.com/raycast/extensions/pull/10776) after a round of testing out the new additions and I was pretty confident about it until I discovered a bug, rather two. Kindly pardon me closing the previous PR!

- Opening a new window with the command `new-window` while fullscreen is enabled does the following:
    1. Opens a new window with the latest space.
    2. Switches the previous window's space to the one we specified in the command.
- I am using a dual monitor setup. The first screen has Arc opened in fullscreen, and next I use `new-window` command in the second screen. Expected behaviour would be that the new window pops up in the second screen. Current behaviour is that it opens on top of the fullscreen Arc window.

I have recorded the first issue in the screencast down below with the demo of the new additions. The second issue can be reproduced with the original extension as well. I hope we can find a solution to this issue.

## Screencast

https://github.com/raycast/extensions/assets/54071350/4843c373-224b-4dad-8e4f-7b10e0ed6d0f

## Checklist

### DIFF
--- a/extensions/arc/CHANGELOG.md
+++ b/extensions/arc/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Arc Changelog
 
+## [Improvements] - 2024-02-15
+
+- Added an argument `space` to the `Open New Window` and `Open New Tab` commands to open a new window or tab in a specific space.
+- Edited the AppleScript commands to open a new window or tab to use the `space` argument.
+
 ## [Fixes] - 2024-02-26
 
 - Fixed an issue in the `Search History` command where quotation marks in the search text were causing a syntax error.


extensions/arc/src/utils.ts
```diff
@@ -21,6 +21,16 @@ export function getSpaceTitle(space: Space) {
...
 
+export function findSpaceInSpaces(spaceId: string, spaces: Space[]): string | undefined {
+  for (const s of spaces) {
+    const title = getSpaceTitle(s);
+    if (title.toLowerCase() === spaceId.toLowerCase() || s.id.toString() === spaceId) {
+      return title;
+    }
+  }
+  return undefined;
+}
```
pernielsentikaer, line 32:
```suggestion
export function findSpaceInSpaces(spaceId: string, spaces: Space[]): string | undefined {
  const space = spaces.find(s => 
     getSpaceTitle(s).toLowerCase() === spaceId.toLowerCase() || 
     s.id.toString() === spaceId
  );
  return space && getSpaceTitle(space);
}
```
Could also be written something like that, not sure if it works since though since I didn't test it 🙂

pernielsentikaer (changes requested):
Hi 👋 

- What's your Raycast handle, it seems like the one you added doesn't exists 🙂

extensions/arc/src/utils.ts
jayeshbhole: Oh, definitely. I've DM'd you on Slack.

pernielsentikaer:
Could you add an entry in the `CHANGELOG.md` file? Here's our guidelines about that: [Version History](https://developers.raycast.com/basics/prepare-an-extension-for-store#version-history) 🙂

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

Issue #7578, opened 2023-07-19
[Slack Status] Pause Notifications


thnk2wn:
### Extension

https://www.raycast.com/petr/slack-status

### Description

I'd love to be able to configure a preset so that I can optionally pause notifications. As it stands I always still show active with just text/icon changes.

### Who will benefit from this feature?

Anyone wanting a break from notifications for certain status presets

PR #10782, opened 2024-02-15
[Slack Status] - Add option to pause notifications

luizeboli:
## Description

Resolves https://github.com/raycast/extensions/issues/7578

This PRs adds the option to configure a preset which pauses notifications (activates do not disturb mode)

~As DND mode doesn't have the "Don't clear" option, I decided to add another dropdown to configure the time separately.~

## Screencast

https://github.com/raycast/extensions/assets/13091635/7cded68f-9629-404e-9531-adf4f29f0949

## Checklist

### DIFF
--- a/extensions/slack-status/src/setStatus.tsx
+++ b/extensions/slack-status/src/setStatus.tsx
@@ -1,4 +1,13 @@
-import { ActionPanel, Detail, Icon, LaunchProps, List, closeMainWindow, getPreferenceValues, popToRoot } from "@raycast/api";
+import {
+  ActionPanel,
+  Detail,
+  Icon,
+  LaunchProps,
+  List,
+  closeMainWindow,
+  getPreferenceValues,
+  popToRoot,
+} from "@raycast/api";
 import { OAuthService, useCachedState, withAccessToken, showFailureToast } from "@raycast/utils";
 import { useEffect, useState } from "react";
 import {
--- a/extensions/slack-status/src/utils.ts
+++ b/extensions/slack-status/src/utils.ts
@@ -142,12 +142,12 @@ export function setStatusToPreset({
   slack,
   preset,
   mutate,
-  mutateDnd
+  mutateDnd,
 }: {
   slack: WebClient;
   preset: SlackStatusPreset;
   mutate: MutatePromise<Profile | undefined>;
-  mutateDnd: MutatePromise<DndInfoResponse | undefined>
+  mutateDnd: MutatePromise<DndInfoResponse | undefined>;
 }) {
   return showToastWithPromise(
     async () => {


thomaspaulmann:
Nice addition and something that was outstanding for a while. Do you think we really need another dropdown for the duration of the notifications? Slack only has a checkbox in their UI to pause notifications, which I would prefer. It keeps it simple and I don't think there are any use-cases where you want to have different durations.

thomaspaulmann:
Because you add new scope for OAuth, we might need to handle the migration. Easiest would be just logging out the OAuth client. Though, it's a bit trickier how to handle the custom API token...

luizeboli:
Hey @thomaspaulmann, thanks for your comments!

The idea behind adding a new dropdown is that I couldn't think of a way to set the snooze duration when the preset has the "Don't clear" option selected for the status. My initial thought was to use the same duration for both, coming from the status. But how to handle this edge case?

You should choose the duration of the 'do not disturb,' and it seems there's no way to just 'toggle' it. Even the "custom" option requires a time (am I missing something here?)

image

Regarding the OAuth scope, yes, you're right... It will fail the request, should we maybe add a custom message in this case informing the user to re-authenticate?

thomaspaulmann:
> Hey @thomaspaulmann, thanks for your comments!
> 
> The idea behind adding a new dropdown is that I couldn't think of a way to set the snooze duration when the preset has the "Don't clear" option selected for the status. My initial thought was to use the same duration for both, coming from the status. But how to handle this edge case?
> 
> You should choose the duration of the 'do not disturb,' and it seems there's no way to just 'toggle' it. Even the "custom" option requires a time (am I missing something here?). [The slack api ](https://api.slack.com/methods/dnd.setSnooze)also says the duration is a required parameter
> 
> image
> --

I was thinking more about the Set Status form of Slack and I think this is what suits the command more. In Slack it looks like this:
CleanShot 2024-02-16 at 9  33 13

We could introduce command called Pause Notifications with an argument that has a few options. I think this would be closest to Slack's UI, which should feel familiar for users.

thomaspaulmann:
> Regarding the OAuth scope, yes, you're right... It will fail the request, should we maybe add a custom message in this case informing the user to re-authenticate? Btw, how can we logout the OAuth? Should I do something through code?

I think it's common to get logged out for OAuth when new scope is required. I would simply do some sort of migration and log out people before showing anything. You would just need to keep track of that state, like with our `LocalStorage`. You can log out the `OAuthClient` with `removeTokens()`, see our [docs](https://developers.raycast.com/api-reference/oauth#oauth.pkceclient-removetokens) for more information.

What's trickier is handling the custom API tokens. Any idea on how to handle that? Is there a way to check which scopes the token has?

luizeboli:
@thomaspaulmann that's interesting, I missed that checkbox. Anyway, the API doesn't seem to provide a property to snooze notifications in the method used to update the profile status ([reference](https://api.slack.com/methods/users.profile.set)). So using the `dnd` method is still required

Besides that, if you choose "Don't clear" it will hide the checkbox, so maybe we can do something similar? Keep showing the checkbox "Pause notifications", but when the status has the "Don't clear" option, we hide it. This way we remove the extra dropdown and make it closer to Slack look and feel

image

---

Handling the auth should be easier, we can make use of the `slack.auth.test` api method. It returns all the scopes of the current user in the `response_metadata` property:

```
{
  "response_metadata": {
    "scopes": ["identify", "emoji:read", "dnd:read", "users.profile:read", "dnd:write", "users.profile:write"]
  }
}
```

We can run this request and check if all the scopes are there, if it's missing something we logout the user, then we store something on `LocalStorage` to avoid running that every time the command is opened. What do you think?

luizeboli:
@thomaspaulmann while I was testing the scopes, I found out that we don't need the "dnd:read" and "dnd:write". I couldn't find that info in the docs, but I reckon it's not needed since we're just tweaking the snooze for the authenticated user, and it's included in the "users.profile:write" somehow. Based on that we don't need to worry about revoking tokens... 

In addition to this change, I removed the extra dropdown and considered the status expiration as the pause notifications time, as Slack does on its UI.

I think we're closer now to Slack's look and feel. What do you think?


https://github.com/raycast/extensions/assets/13091635/7cded68f-9629-404e-9531-adf4f29f0949

luizeboli:
After more testing, adding the `dnd:read dnd:write` scopes is necessary indeed. I'm not sure why I could use DND without the scopes even after recreating the app without them.

I will handle the scope checking when app is loading and then store in local storage to avoid refetching

luizeboli:
@thomaspaulmann would you mind checking this PR again? I think we're done now.

pernielsentikaer:
Hi @luizeboli 👋

There was recently an update to the extension that has now caused some conflicts in the code. Could you try to resolve them? Let me know if you need help doing so 🙂

pernielsentikaer:
I also just tested this enhancement with both `API` and `OAuth` - it works as expected 👏

luizeboli:
hey @pernielsentikaer, thanks for taking a look at these changes. All the conflicts have been solved :) 

- Quicklinks feature still working
- Pause notifications still working

pernielsentikaer:
@thomaspaulmann do you want to give it a go too?

mil3na (approved):
LGTM!
Thanks for your contribution!

luizeboli:
@thomaspaulmann @mil3na hey I just noticed the merge was reverted, if you need help to sort anything pls reach out 🙏🏽

PR #10790, opened 2024-02-16
Update to Ext/pomodoro

Nathanjms:
- Instead of replacing image, add checkbox for both, enable image by default
- Option to use a quote on timer completion, instead of an image
- Initial commit

## Description

Added option to use a quote on timer completion, and also added option to enable/disable the image on timer completion (image is enabled by default to maintain existing behaviour).

### When image is disabled and quote is enabled:

image

### When Both are enabled:
image

## Checklist

### DIFF
--- a/extensions/pomodoro/src/pomodoro-control-timer.tsx
+++ b/extensions/pomodoro/src/pomodoro-control-timer.tsx
@@ -13,7 +13,7 @@ import {
   resetInterval,
 } from "../lib/intervals";
 import { FocusText, ShortBreakText, LongBreakText } from "../lib/constants";
-import { GiphyResponse, QuoteResponse } from "../lib/types";
+import { GiphyResponse, Quote } from "../lib/types";
 
 const createAction = (action: () => void) => () => {
   action();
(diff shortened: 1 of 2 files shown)


Nathanjms:
I've just noticed an outstanding PR on this extension (#10722) which will likely cause clashes with this. If preferred, I can re-do this after that has been merged in?

extensions/pomodoro/src/pomodoro-control-timer.tsx
```diff
@@ -107,8 +107,29 @@ const ActionsList = () => {
...
+
+  function formatQuote(quote: { content: string; author: string }) {
+    // Format the quote in markdown:
+    return `> ${quote.content} \n>\n> &dash; ${quote.author}`;
+  }
+};
+
 const EndOfInterval = () => {
-  let markdownImage;
+  let markdownContent = "# Timer Completed \n\n";
```
MartinGonzalez, line 132: Perhaps would be nice to keep the same language. Instead of Timer perhaps using Interval?
What do you think?

Nathanjms: Sounds good, I'll update this now

pernielsentikaer:
@KELiON can you check this?

extensions/pomodoro/src/pomodoro-control-timer.tsx
```diff
@@ -107,8 +107,29 @@ const ActionsList = () => {
...
+    markdownContent += data.length ? formatQuote(data[0]) : formatQuote(backupQuote);
+  }
+
+  return markdownContent;
+
+  function formatQuote(quote: { content: string; author: string }) {
+    // Format the quote in markdown:
+    return `> ${quote.content} \n>\n> &dash; ${quote.author}`;
+  }
+};
```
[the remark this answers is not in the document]
Nathanjms: Thanks for taking a look and the helpful feedback! This is much sleeker than how it was, so I've just updated it with your changes. I did have two comments on it:

1. I had to change the line below in order for VS Code to stop shouting at me (it was wanting `data.data` after the change) - I'm guessing this is what was intended, and looks to be working with this:

```diff
-- const { isLoading, data } = useFetch<{data: Quote[]}>(
+ const { isLoading, data } = useFetch<{Quote[]}>(
```

2. There is a small change of behaviour based on simplifying this method as you've suggested: we'll show the `"You did it!"` quote while the api request is loading, then it will change to the quote once ready. 
    * Previously, there will be no quote until the API request has resolved, then it will use the backup if no quote is returned. Now, depending on the API response speed, we may see a 'flicker' of the quote changing, but the API response is fast so I'm happy with this if you are! (I think this is personal preference anyway, some may prefer it not to be blank).
    * See gifs below demo'ing this:

### Before Change:

Screen Recording 2024-02-20 at 16 46 53


### With this new Change:

Screen Recording 2024-02-20 at 16 42 41

PR #10995, opened 2024-02-27
Update power-management extension

bbatsche:
## Description

Using AppleScript's `with administrator privileges` feature requires the user to type in their password every time it's run. Since I have very silly password and am not always the best typist, this was timing out for me. I've fixed this in two ways:

1. Bump the timeout to 1 minute (instead of the default 10 seconds)
2. Check if the user has configured `sudo` to allow touch ID and if so, use that to execute the command instead

## Checklist

### DIFF
--- a/extensions/power-management/README.md
+++ b/extensions/power-management/README.md
@@ -1,3 +1,37 @@
 # Power Management
 
 Change Power Management Settings of the system
+
+## Authenticate with Touch ID
+
+Changing power management settings requires administrator privileges. By default this is done via username & password. If you would rather use your fingerprint to authenticate you must modify your `sudo` configuration (**note:** this change will apply to `sudo` globally, not just within Raycast).
+
+### macOS Sonoma (14) and Later
+
+1. Copy `/etc/pam.d/sudo_local.template` to `sudo_local`:
+   ```bash
+   sudo cp /etc/pam.d/sudo_local.template /etc/pam.d/sudo_local
+   ```
+2. Remove the comment character (`#`) from the `auth` line in `sudo_local`:
+   ```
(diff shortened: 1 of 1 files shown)


pernielsentikaer (changes requested):
Hi 👋

I still need to enter the password after every run, also I need to enter the password - should I do anything to test in order to use touch id?

bbatsche:
You need to configure `sudo` to use Touch ID. Depending on your version of macOS, there are [two](https://sixcolors.com/post/2020/11/quick-tip-enable-touch-id-for-sudo/) [ways](https://sixcolors.com/post/2023/08/in-macos-sonoma-touch-id-for-sudo-can-survive-updates/) to do this. This change should support both.

pernielsentikaer:
Perhaps consider adding a note in the README regarding this matter. Could you do that?

bbatsche:
Sure, I can add that! Thanks for your feedback

pernielsentikaer:
@sasivarnan can you check this?

bbatsche:
@pernielsentikaer or @sasivarnan could either of you respond so this PR doesn’t get closed?

sasivarnan:
@bbatsche Thanks for adding these enhancements! They provide a significant improvement to the user experience.

Additionally, I learned about `/etc/pam.d/sudo_local` file and its benefits. 

@pernielsentikaer I've tested the changes introduced in this PR on my M2 MacBook Air running macOS Sonoma and everything seems to be functioning as expected.

sasivarnan (approved):

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #11021, opened 2024-02-29
Update color-picker to force using srgb color space

otakustay:
## Description

macOS's color picker by default picks color in color space of the monitor, with almost all Apple monitors supporting P3, it's color code is different with the commonly used web standard.

In this patch I forced color space to be sRGB so that the color code is the same as that in css.

As an example, the [tailwind palate](https://tailwindcss.com/docs/customizing-colors) renders red-400 as `#f87171` but raycast picks it as `#e77975`, in this patch this is fixed.

A exception is that macOS's native color picker still don't pick some color exactly the same as css specified, such as the `#f87171` red-400 value is picked as `#f77171`, I checked it with macOS's built-in app, this is not a problem of RayCast or this extension.

## Checklist

### DIFF
--- a/extensions/color-picker/CHANGELOG.md
+++ b/extensions/color-picker/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Color Picker Changelog
 
+## [Fix] - 2024-02-29
+
+- Forced to use sRGB color space
+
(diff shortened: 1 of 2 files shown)


pernielsentikaer:
Closing in favor of #11024

Issue #10844, opened 2024-02-20
CHATGPT list render problem


bxb100:
### Description

list render with wrong behavior

https://github.com/raycast/extensions/assets/20685961/ad4f7913-c915-4d13-acb2-e4c653b8a9f3

https://github.com/raycast/extensions/assets/20685961/5d53146a-3d5d-410c-9fff-aa97fd241b51

### Steps To Reproduce

 1. using any model
  2. input https://gist.githubusercontent.com/KristofferEriksson/9ce429cbc33a9accb28bb1abdfecc775/raw/ef52e08fbbf76d931ede7067847fe02ae73ab6f3/useUndo.ts 
  3. and Repeat 2

mathieudutour:
The extension uses `selectedItemId` which I'm guessing is the source of the bug. Could you create a small example where the issue can be reproduced?

mathieudutour:
For now you can just remove `onSelectionChange` prop and it should work. I'll see how we can prevent the race condition

PR #11038, opened 2024-03-01
[ChatGPT]: support Groq and Ollama

bxb100:
## Description

- break change: Remove `useApiEndpoint` `useAzure` and `azureEndpoint` from the settings; use `apiEndpoint` instead `azureEndpoint`, use `apiEndpoint` empty or not instead of originally enable checkbox
- support Ollama and Groq
- remove `onSelectionChange` due to [issue](https://github.com/raycast/extensions/issues/10844)

### effect: 
- using Azure OpenAI will be affected by this PR, they need to change the `apiEndpoint`
- can't use both OpenAI and `Azure OpenAI` by checkbox
- change `provider`, the original model setting may not be suitable, but it seems not a real edge case
- this extension chatGPT will not be "OpenAI" only, but akin to OpenAI

## Screencast

image

https://github.com/raycast/extensions/assets/20685961/7b9f24ae-62aa-42d3-9857-f7c606baa6bd

## Checklist

### DIFF
--- a/extensions/chatgpt/src/hooks/useModel.tsx
+++ b/extensions/chatgpt/src/hooks/useModel.tsx
@@ -31,7 +31,7 @@ export function useModel(): ModelHook {
         .list({ httpAgent: proxy })
         .then((res) => {
           const models = res.data;
-          setOption(models.filter(x => provider === "groq" || x.id.startsWith("gpt")).map((x) => x.id));
+          setOption(models.filter((x) => provider === "groq" || x.id.startsWith("gpt")).map((x) => x.id));
         })
         .catch(async (err) => {
           console.error(err);


pernielsentikaer:
Wouldn't that be odd, given that this is an OpenAI ChatGPT extension? Llama and Groq are not from OpenAI, correct?

bxb100:
Ya, I admit it's a bad idea, and Groq/Llama isn't as useful as I expected.

reneleonhardt:
Would it make sense to fork chatgpt into a separate extension instead to change it to allow working with local models through ollama?

bxb100:
> Would it make sense to fork chatgpt into a separate extension instead to change it to allow working with local models through ollama?

The extension now can define the model name, which allows it to support Ollama in some way.

PR #11073, opened 2024-03-03
Update gcp-search extension

m-gentil:
## Description

Sort projects and products by relevance.
Relevance is computed by usage and scaled down daily by a decay factor so that usage of projects / products fades with time.

One notable change in this PR is that we aren't able to leverage Raycast's built-in fuzzy List search, since it doesn't preserve the initial order of the list, making the manual filtering process a bit more demanding and potentially less accurate—for now I integrated exact matching rather than fuzzy which seems to be internally employed by Raycast to match items (see [docs](https://developers.raycast.com/api-reference/user-interface/list#search-bar)).

LMK if you think there's a better way to achieve this.

NB: I also [migrated](https://developers.raycast.com/migration) the existing code to the newer version of the Raycast APIs, updating the React components and their properties.

## Checklist

### DIFF
--- a/extensions/google-cloud-platform-search/.gitignore
+++ b/extensions/google-cloud-platform-search/.gitignore
@@ -12,9 +12,3 @@ compiled_raycast_swift
 # misc
 .DS_Store
 .idea
-.npmrc
-raycast-env.d.ts
-
-.raycast-swift-build
-.swiftpm
-compiled_raycast_swift


klaa97 (approved):
Bomber! Seems okay to me

(I am not aware of any other way to combine fuzzy search with this... I checked a few other extensions and they all do something quite similar to your solution)

extensions/google-cloud-platform-search/.gitignore
```diff
@@ -11,3 +11,10 @@ compiled_raycast_swift
 
 # misc
 .DS_Store
+.idea
+.npmrc
+raycast-env.d.ts
+
+.raycast-swift-build
+.swiftpm
+compiled_raycast_swift
```
hisomura, line 20: These are duplicated.

hisomura (commented):
Great Job!

I made a small comment, but it's so trivial that it doesn't need to be fixed.

extensions/google-cloud-platform-search/.gitignore
m-gentil: Thank you, I missed them during conflict resolution! Removed them ✅

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #11102, opened 2024-03-05
Dedup codeowners

LitoMore:
See https://github.com/raycast/extensions/pull/11085#issuecomment-1976355134.

There are two same codeowners in the comment.

It seems the `scripts/generate-code-owners.js` needs to be fixed as well. I will keep an eye on this.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -147,7 +147,7 @@
 /extensions/charged                                  @apoorvsadana
 /extensions/charming-chatgpt                         @KittenYang
 /extensions/chatgo                                   @chuyun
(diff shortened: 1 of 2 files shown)


mathieudutour:
The issue isn't the code owners file, it's https://github.com/raycast/extensions/blob/main/.github/raycast2github.json. There are probably some users that are incorrectly matched

LitoMore:
@mathieudutour Thanks. I will take a look. In the meantime, I think we still need to dedup users in CODEOWNERS.

mathieudutour:
No, it's auto-generated from the file I linked so no point in changing it manually (as mentioned at the top of the file)

LitoMore:
Ah, I see. Closing. I will check that part.

mathieudutour:
FYI this script https://github.com/raycast/extensions/blob/main/scripts/find-wrong-raycast-to-github-mapping.js can be used to find wrong mapping - then it's a tedious process to find the correct one 😓

PR #11161, opened 2024-03-07
Update Dropshare Extension

tjosten:
## Description

Updating the existing Dropshare extension to add a new "URL shortener" action that helps accessing the configured Dropshare URL shortener.

## Checklist

### DIFF
--- a/extensions/dropshare/src/shorten-url.tsx
+++ b/extensions/dropshare/src/shorten-url.tsx
@@ -6,59 +6,59 @@ import { getUrlFromClipboard } from "./utils/clipboard";
 
 const queryClient = new QueryClient();
 export default function ShortenLinkWrapper(props: LaunchProps<{ arguments: { url: string; key?: string } }>) {
-    return (
-        <QueryClientProvider client={queryClient} >
-            <ShortenLink arguments={props.arguments} launchType={LaunchType.UserInitiated} />
-        </QueryClientProvider>
-    );
+  return (
+    <QueryClientProvider client={queryClient}>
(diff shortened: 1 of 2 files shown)


pernielsentikaer (changes requested):
Wouldn't it be better just to have it as a no-view instead 🙂

tjosten:
I oriented myself on comparable integrations (like Dub), but sure, I'll happily update it.

pernielsentikaer:
For me, it just feels like an annoying extra step. If you want to keep it as it is, that's also okay with me 🙂

tjosten:
If the extension was a no-view, the user had no chance up update or replace the URL that might have been read from the clipboard before shortening, right? If so, then I'd personally prefer it the way it is.

pernielsentikaer:
One solution could _maybe_ be that if you enters an argument you'll early `popToRoot()` and shorten in the script, that might be possible 🙂

tjosten:
> One solution could _maybe_ be that if you enters an argument you'll early `popToRoot()` and shorten in the script, that might be possible 🙂

I was rather thinking about it from a UX point of view. The current flow is:

User selects "Shorten URL" action in Raycast, and can either directly provide a URL as argument, or just hit "return" and go into the action's view. There, the input field will either be pre-filled with a URL that the extension reads from the user's clipboard, or be empty, so that the user can enter a URL manually or copy & paste it from somewhere else. That way, it's always very clear what is about to happen, and the user will never be wondering where the URL has been read from when the action was triggered.

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #11185, opened 2024-03-09
Update fantasy-premier-league-rankings extension

niall-maloney:
## Description

- Fix: Use EPL's API directly.
- Fix: Added missing dependency that was causing player data to be missing on initial render.

## Checklist

### DIFF
--- a/extensions/fantasy-premier-league-rankings/.eslintrc.json
+++ b/extensions/fantasy-premier-league-rankings/.eslintrc.json
@@ -1,10 +1,6 @@
 {
   "root": true,
-  "env": {
-    "es2020": true,
-    "node": true
-  },
-  "parser": "@typescript-eslint/parser",
-  "plugins": ["@typescript-eslint"],
-  "extends": ["eslint:recommended", "plugin:@typescript-eslint/recommended", "prettier"]
+  "extends": [
+    "@raycast"
+  ]
 }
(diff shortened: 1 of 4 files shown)


hwb15:
Hi @niall-maloney,

Thanks for your work on the extension, I think the proxy I built might have gone down recently. At the time I built the extension with the FPL API you needed to proxy the api and use a specific user agent to avoid rate limiting and CORS issues which meant you couldn't use the api directly - I can this branch adds adds the api directly, does it now work ok to call directly? 

I think previously it would work for 10 requests+ (could be 5+-), but after then it'd start to rate limit.

niall-maloney:
@hwb15 I haven't encountered any rate limiting or CORS issues in my testing so far. It would be great if you could pull this branch and give it a test to confirm.

pernielsentikaer:
Same here, tested it extensively without encountering any rate limits.

@hwb15, are you okay to proceed? If we encounter any issues, we can always find a better solution.

niall-maloney:
@pernielsentikaer should we proceed with merging since the current version is broken and we have both tested the fixes? Like you suggested, if any issues arise we can find another solution.

hwb15:
> @pernielsentikaer should we proceed with merging since the current version is broken and we have both tested the fixes? Like you suggested, if any issues arise we can find another solution.

I'll pull this down today and get it tested to check that the API doesn't rate limiting then should be good to go?

hwb15:
> Same here, tested it extensively without encountering any rate limits.
> 
> @hwb15, are you okay to proceed? If we encounter any issues, we can always find a better solution.

Tested for rate limiting and all works well on boundaries where previously using the direct FPL api would have caused issues so very positive and more than happy with @niall-maloney changes - thanks for maintaining it and getting it back up and running!! (maybe next season we should get a raycast fpl league going!)

@pernielsentikaer I'm happy proceed and go ahead pending any merge checks

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #11188, opened 2024-03-09
[ChatGPT] Allow to hide model selector when asking a question

irangarcia:
## Description

Allow to hide model selector when asking a question

## Screencast

| -  | Configuration  | UI |
|---|---|---|
|**Default behavior**| hidden-disabled | Screenshot 2024-03-09 at 15 42 47 |
| **Changed behavior** | hidden-enabled | hidden |

## Checklist


pernielsentikaer (changes requested):
Same as the other, rely on raycast-env.d.ts instead 🙂

bxb100:
@irangarcia I don't think it's a good idea because there are already a lot of configurations.

irangarcia:
> @irangarcia I don't think it's a good idea because there are already a lot of configurations.

Could you ellaborate on that? Why would be a problem?

bxb100:
IMHO it introduces a new configuration and does nothing

pernielsentikaer:
I might agree, since it's only hiding the field in a view where you won't spend many seconds.

PR #11247, opened 2024-03-12
feat(slack): add ability to set slack status

sammcj:
## Description

Slack Extension

- Add the ability to set Slack status.
- Update package libraries (minor).
- Update README to reflect changes.

## Checklist

### DIFF
--- a/extensions/slack/package-lock.json
+++ b/extensions/slack/package-lock.json
@@ -7,7 +7,7 @@
       "name": "slack",
       "license": "MIT",
       "dependencies": {
-        "@raycast/api": "^1.69.2",
+        "@raycast/api": "^1.69.3",
         "@slack/web-api": "^6.12.0",
         "date-fns": "^2.30.0",
         "lodash": "^4.17.21",
@@ -16,8 +16,8 @@
       },
       "devDependencies": {
         "@types/lodash": "^4.17.0",
-        "@types/node": "~18.19.23",
-        "@types/react": "^17.0.76",
+        "@types/node": "~18.19.25",
+        "@types/react": "^17.0.79",
         "@typescript-eslint/eslint-plugin": "^5.62.0",
         "@typescript-eslint/parser": "^5.62.0",
         "eslint": "^7.32.0",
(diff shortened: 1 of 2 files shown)


pernielsentikaer:
Hi @sammcj 👋 

Thanks for your contribution, could you correct you Raycast handle in `package.json`

@momme-rtf could you check this PR

sammcj:
Ah sorry, I thought it was the standard package.json Github username convention, fixed now :)

momme-rtf:
Thanks for your contribution @sammcj !
@pernielsentikaer I'm just thinking: Wouldn't it rather make sense to merge the [Slack Status](https://www.raycast.com/petr/slack-status) extension with this one here instead of replicating existing features? Slack Status is also already quite advanced and heavily used.

pernielsentikaer:
The other extension does already support this as far as I know, but you're right - it might not fit in here

momme-rtf:
@pernielsentikaer 
Have you had already cases where you merged two existing extensions? If so how did you manage it?
If we would merge the Slack Status extension in to this one, would the Slack Status extension stop working (after we delete it from the store) for those users that have it already installed? 
How do you feel about it in general? Would you rather want to keep them separated or would you want to have them merged?

sammcj:
Yeah I saw that status extension but it seemed quite out of place - weird to have it seperate and it doesn’t actually let you set your status - just if you’re active or not.

pernielsentikaer:
> it doesn’t actually let you set your status - just if you’re active or not.

Are you sure about that @sammcj 

slack-status 2024-03-19 at 11 25 49

sammcj:
Oh wow! Apparently not. Not sure why but when I tried it I thought it only allowed setting 'presence'.

Hmm. Maybe it could just be combined into this one extension then? Still seems odd to have two.

sammcj:
Oh that's right - I got muddled, that was this extension that could (previously) only do presence.

The other status extension requires installing a third party integration - which I couldn't get past our Slack Admins 👮 so couldn't ever try .

pernielsentikaer:
The purpose of the two extensions is not quite the same, so it still makes sense to have them as two separate entities. It's possible to merge them, but one group of users would end up with a stale extension.

Currently, there is no effective solution for multiple workspaces, but it is something we aim to address in the future.

momme-rtf:
FYI: I also wanted to migrate this extension soon to simplify the login by using the oauth flow.
Providing the support of multiple workspaces would be then a different topic/feature. Having forks for each workspace doesn't sound like a proper solution to this.

pernielsentikaer:
I agree, it's something that we at Raycast need to handle, but it's not straightforward to do so which is also why it haven't happened yet 🙂

sammcj:
Ah I see - I'm guessing it's the same/similar reason other plugins have similar limitations (e.g. With the Ollama extension being limited to a single endpoint/instance). Good to know it's something being thought about though :)

pernielsentikaer:
Do you midd anything in the other extension now that this also will use OAuth in the future

PR #11287, opened 2024-03-14
Add raycast-g4f extension.

XInTheDark:
## Description

Add the raycast-g4f ("GPT4Free") extension. The code base is derived from the raycast-gemini extension. However, for response generation, this extension uses the [g4f-ts](https://github.com/VictorMRojas/g4f-ts) module. The default provider used is https://chatgpt.ai/, with options to switch to Bing or ChatBase. The default model used is `gpt-4-32k`.

This extension follows the terms of services of the aforementioned provider, as it provides the functionality solely for personal, non-commercial use.

[Update: Image generation capabilities have also been added. It is powered by the same g4f-ts module.] 

## Screencast
raycast-g4f-1
raycast-g4f-2
raycast-g4f-3
raycast-g4f-4
raycast-g4f-5

## Checklist

### DIFF
--- a/extensions/raycast-g4f/package-lock.json
+++ b/extensions/raycast-g4f/package-lock.json
@@ -42,23 +42,24 @@
       }
     },
     "node_modules/@babel/helper-validator-identifier": {
-      "version": "7.22.5",
-      "resolved": "https://registry.npmjs.org/@babel/helper-validator-identifier/-/helper-validator-identifier-7.22.5.tgz",
-      "integrity": "sha512-aJXu+6lErq8ltp+JhkJUfk1MTGyuA4v7f3pA+BJ5HLfNC6nAQ0Cpi9uOquUj8Hehg0aUiHzWQbOVJGao6ztBAQ==",
+      "version": "7.22.20",
+      "resolved": "https://registry.npmjs.org/@babel/helper-validator-identifier/-/helper-validator-identifier-7.22.20.tgz",
+      "integrity": "sha512-Y4OZ+ytlatR8AI+8KZfKuL5urKp7qey08ha31L8b3BwewJAoJamTzyvxPR/5D+KkdJCGPq/+8TukHBlY10FX9A==",
       "dev": true,
       "engines": {
         "node": ">=6.9.0"
       }
     },
     "node_modules/@babel/highlight": {
-      "version": "7.22.5",
-      "resolved": "https://registry.npmjs.org/@babel/highlight/-/highlight-7.22.5.tgz",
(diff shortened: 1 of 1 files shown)


XInTheDark:
Extension is ready for review :-)

pernielsentikaer:
Hi 👋

Thank you for your contribution 🔥

Why does this extension not require an API key? It appears to be leveraging GPT-4 and other paid models.

XInTheDark:
Hi! The extension does not use OpenAI as the host, instead the model is hosted by http://chatgpt.ai/. The service is free for the public and it does not require an API key. This can be verified in the source code of the g4f-ts package I used.

pernielsentikaer:
I just checked that site and as a free user (after signing up) I got 6 queries in total a day 🙂

XInTheDark:
Hmm, I don't know about the restrictions on the site itself. It is possible that the provider added it in order to rate limit users, but the API I used (offered with g4f-ts) is not affected - at least I was not rate limited during the testing of my extension.

XInTheDark:
Update: please see https://github.com/VictorMRojas/g4f-ts/blob/main/src/Providers/ChatCompletion/GPT.ts#L20 for the implementation of the provider code in g4f-ts. The README specifies that the GPT provider is powered by chatgpt.ai, but the actual URL called is "https://nexra.aryahcr.cc/api/chat/gpt". Their terms of service allow for free usage of the API for any purpose: https://nexra.aryahcr.cc/terms.

So insofar as this extension is concerned, it is in fact only calling the above URL and is not really connected to chatgpt.ai directly. ToS wise, I don't think there's a problem. 🙂

pernielsentikaer:
Thanks for the explanation.

I'm still not sure about having this extension in our store.

Some of the concerns:

* I still don't see how this can be free, given that models like OpenAI’s have a significant cost
* The server isn’t public, so it’s unknown what happens to user’s prompts

XInTheDark:
> The server isn’t public, so it’s unknown what happens to user’s prompts

This applies for a lot of API based extensions, and it's not feasible for the host to open source their model, so for all LLM extensions this issue is probably unresolvable

> I still don't see how this can be free, given that models like OpenAI’s have a significant cost

I don't really know the hosting process (see above), but looking at some repo creation dates, these free chatGPT providers have been running for many months now, so their hosting is quite sustainable for them at least.

XInTheDark:
> Do you know more about this npm package?

Unfortunately not, I found this package after searching for gpt for free on npm, and it was the best choice for this extension. I'll check if there are other possibly more reputable packages that provide the required functionality, but most of them probably also utilize endpoints similar to this one.

PR #11303, opened 2024-03-15
Update Porkbun extension: fix error when opening domains in browser

djpowers:
## Description

There "Retrieve All Domains" command was failing when using the "Open in Browser" action. The domain item is missing the HTTP(S) protocol, and resulted in an error message saying "The application can’t be opened."

I updated the `url` property to include the "HTTPS" protocol. I checked the [Porkbun API documentation](https://porkbun.com/api/json/v3/documentation) and the "Domain List All" endpoint does not include the protocol. This could default to "HTTP" instead, but the `getFavicon` function a few lines above in this file also assumes HTTPS, and it I figured it should be consistent.

With this change, the domains are opened in the browser as expected.

## Checklist

### DIFF
--- a/extensions/porkbun/CHANGELOG.md
+++ b/extensions/porkbun/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Porkbun Changelog
 
+## [Fix open in browser command for domain items] - 2024-03-14
+
+### Fixes
+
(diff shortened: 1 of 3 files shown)


pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #11319, opened 2024-03-15
Update obsidian extension

casassg:
## Description

Fixes append to daily note to work as fallback command. This can make it easy to quickly append some text.

## Screencast

https://github.com/raycast/extensions/assets/6912589/97a5656b-13b6-4f61-80fa-3cd4b03bff50

## Checklist

### DIFF
--- a/extensions/obsidian/src/dailyNoteAppendCommand.tsx
+++ b/extensions/obsidian/src/dailyNoteAppendCommand.tsx
@@ -1,4 +1,13 @@
-import { Action, ActionPanel, closeMainWindow, getPreferenceValues, List, open, popToRoot, LaunchProps } from "@raycast/api";
+import {
+  Action,
+  ActionPanel,
+  closeMainWindow,
+  getPreferenceValues,
+  List,
+  open,
+  popToRoot,
+  LaunchProps,
+} from "@raycast/api";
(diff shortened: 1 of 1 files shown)


pernielsentikaer:

Could you check this @KevinBatdorf 🙂

KevinBatdorf:
Hey @casassg Thanks for the PR. I think the PR is fine since it's supporting the fallback, and if you dont pass in the `text` it will read undefined (I initially expected this to throw actually)

But based on the video you shared, it doesn't seem like you are using this change in it? I don't think it's a blocker actually, but do you want to mention how you're using it? Could be something in Raycast I'm out of date on actually, or a misunderstanding elsewhere?

By the way, if you want to use the `launchCommand` to invoke the daily note command, you could use this signature and woulnd't need the fallback text.
```
await launchCommand({
    name: "dailyNoteAppendCommand",
    type: LaunchType.UserInitiated,
    arguments: { text: "foo" },
  });
```

Overall, I don't see any reason not to support the fallback text though except maybe it should be added to every command. Or if you just didn't know how to add the arguments to the function.

pernielsentikaer:
Hi @casassg 👋 

Could you check the response from Kevin?

PR #11326, opened 2024-03-16
Update dashlane-vault extension

neo773:
## Description
This PR adds support for authenticating the vault via Touch ID and not leaving it exposed.
Do not merge this before this [PR](https://github.com/Dashlane/dashlane-cli/pull/223)

## Screencast
https://github.com/raycast/extensions/assets/62795688/f5e96acb-8d88-482f-ab10-127752b4ab50

## Checklist


extensions/dashlane-vault/assets/auth
Mikescops, line 1: What does this binary?

Mikescops:
(Dashlane developer here): I'm not a huge fan of Raycast managing secrets like the master password, I would rather implement this in the CLI.

neo773:
@Mikescops 
Hi, the binary’s from sources/auth/main.swift

can be recompiled by running nom run build-swift

Mikescops:
> @Mikescops Hi, the binary’s from sources/auth/main.swift
> 
> can be recompiled by running nom run build-swift

Thanks I figured out after a second read. It's an interesting approach, I'm wondering if I can build this in the CLI directly.

extensions/dashlane-vault/Sources/auth/main.swift
```diff
@@ -0,0 +1,88 @@
+import Foundation
+import LocalAuthentication
+
+let policy = LAPolicy.deviceOwnerAuthenticationWithBiometrics
```
Mikescops, line 4: This policy here is insufficient, it can be bypassed, you should rather use `kSecAttrAccessControl`.

I recommend you to have a read to this post: https://www.andyibanez.com/posts/ios-keychain-touch-id-face-id/

extensions/dashlane-vault/src/lib/dcli.ts
```diff
@@ -2,21 +2,34 @@ import { getPreferenceValues } from "@raycast/api";
 import { existsSync } from "fs";
 import { safeParse } from "valibot";
 
-import { execFilePromis } from "@/helper/exec";
+import { getMasterPassword } from "@/helper/master-password";
 import { VaultCredential, VaultCredentialSchema, VaultNote, VaultNoteSchema } from "@/types/dcli";
+import { exec } from "child_process";
```
tmwrnr, line 7: Why to you add execa and then still use exec?

extensions/dashlane-vault/src/lib/dcli.ts
```diff
@@ -2,21 +2,34 @@ import { getPreferenceValues } from "@raycast/api";
...
-  if (CLI_PATH) {
-    const { stdout } = await execFilePromis(CLI_PATH, args, { maxBuffer: 4096 * 1024 });
-    return stdout;
+  if (!CLI_PATH) {
+    throw new Error("Dashlane CLI is not found!");
   }
+  const masterPassword = preferences.useTouchID ? await getMasterPassword() : null;
 
-  throw Error("Dashlane CLI is not found!");
+  const { stdout, stderr } = await execa(CLI_PATH, args, {
```
tmwrnr, line 21: This is not working on my machine. The promise never resolves. I updated the Dashlane CLI and set my master password. The Search Passwords command just shows me the loading state.

extensions/dashlane-vault/src/lib/authcli.ts
```diff
@@ -0,0 +1,27 @@
+import { environment } from "@raycast/api";
+import { execa } from "execa";
+import { chmod } from "fs/promises";
+import { join } from "path";
+
+const executeCommand = async (args: string[]) => {
+  const command = join(environment.assetsPath, "auth");
```
tmwrnr, line 7: Could you move this constant outside of the function similar to `CLI_PATH` in `dcli.ts`. This path will never change during runtime.

extensions/dashlane-vault/src/lib/authcli.ts
```diff
@@ -0,0 +1,27 @@
...
+  await chmod(command, "755");
```
tmwrnr, line 8: Do we need to change this before every run or could this also be done once?

extensions/dashlane-vault/src/master-password.tsx
```diff
@@ -0,0 +1,66 @@
...
+import { getMasterPassword } from "./helper/master-password";
+import { setKeychainItem } from "./lib/authcli";
+import { encryptVault } from "./lib/dcli";
+
+export default function SetMasterPassword() {
+  const [password, setPasswordInput] = useState("");
+  const [keyName, setKeyName] = useState("");
+  const [passwordExists, setPasswordExists] = useState(false);
+
+  const fetchKeyNameAndCheckPassword = async () => {
```
tmwrnr, line 12: In this command the user wants to set his master password. Why do we need to fetch and insert the current password?

extensions/dashlane-vault/src/master-password.tsx
```diff
@@ -0,0 +1,66 @@
...
+    const existingPassword = await getMasterPassword();
+    if (existingPassword) {
+      setPasswordExists(true);
+      setPasswordInput(existingPassword);
+    } else {
+      const keyName = `dashlane-vault-master-password-${Math.random().toString(36).substring(2, 15)}`;
```
tmwrnr, line 18: Is there the need of the random key name? I thought that the keychain has a "space" for each application. If so we just could use a static `keyName`.

extensions/dashlane-vault/src/master-password.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+  const handleSetPassword = async () => {
+    if (!password) {
+      showToast({ style: Toast.Style.Failure, title: "Password cannot be empty" });
+      return;
+    }
+    try {
+      await setKeychainItem(keyName, password);
+      await encryptVault();
+      await LocalStorage.setItem(keyName, "");
```
tmwrnr, line 31: Why do you set the keyName to the LocalStorage with no value and later get the key in `findDashlaneKey` by checking all keys. If this is for security reasons, I think there is no difference to just set the keyName as the value or just use a static keychain key.

It's like hiding your key in the garden and having a note next to the front door telling you where the key is in the garden.

tmwrnr (changes requested):
Hi @neo773,

thanks for opening this pull request and contribute to this extension. I like the idea that you don't need to save the master password by the Dashlane CLI! When I created the extension I was unsure how to achieve this.

I don't know enough about Swift to really review this part of the code. Maybe someone from Raycast can do this?

I could not really test this new feature because it did not work on my machine as mentioned in one of my comments. Regardless of the implementation, I do have a general question about your approach. As of my understanding Raycast is not allowing extensions accessing the keychain as mentioned [here](https://developers.raycast.com/basics/prepare-an-extension-for-store#keychain-access). Maybe someone from Raycast can confirm this?

Furthermore I, as an user, don't like the idea to insert my password to a Raycast extension even if I know, that the extension itself handles the keychain access. It just does not feel right for me. If a user has concerns with using the Dashlane CLI with the master password saved, maybe the CLI can implement the biometrics itself? I assume Dashlane is also just saving the password within keychain @Mikescops?

Maybe we could prompt the password every time the user want to access the vault? 

Again, thank you very much for your time implementing this and I'm curious to see where this will lead to.

extensions/dashlane-vault/src/lib/dcli.ts
neo773: execa did not work for some reason when executing that particular command it threw some weird error.

extensions/dashlane-vault/src/lib/dcli.ts
neo773: Apparently my upstream commit for dashlane cli did not work, will look into this.

extensions/dashlane-vault/src/master-password.tsx
neo773: Yes, on second thoughts this might not do much seemed good at the time of writing it.
Removing it.

extensions/dashlane-vault/src/master-password.tsx
neo773: It's mainly for UI purposes.

extensions/dashlane-vault/Sources/auth/main.swift
neo773: I looked into this but couldn't get it working as `kSecAttrAccessControl` requires "Keychain Access Groups" entitlement which is not supported by CLI Apps and also requires code signing. Workaround seems to be using a separate daemon for this.

Given the scope it would make more for Dashlane to integrate this natively in a secure manner, I expect this to take quite some time assuming your team approves this feature.

However in the meantime we could still merge this since something is better than simply leaving the vault open at all times which is a huge security hazard.

mil3na:
Is this ready for a final review or does @tmwrnr and @Mikescops have any more comments?

tmwrnr:
Hi @mil3na,

there are still general questions that are not answered, mentioned in my [comment](https://github.com/raycast/extensions/pull/11326#pullrequestreview-1956316817).

E.g.: 
> I don't know enough about Swift to really review this part of the code. Maybe someone from Raycast can do this?

Could you or someone from Raycast do that?

> As of my understanding Raycast is not allowing extensions accessing the keychain as mentioned [here](https://developers.raycast.com/basics/prepare-an-extension-for-store#keychain-access). Maybe someone from Raycast can confirm this?

What is your opinion about that?

> Furthermore I, as an user, don't like the idea to insert my password to a Raycast extension even if I know, that the extension itself handles the keychain access. It just does not feel right for me. If a user has concerns with using the Dashlane CLI with the master password saved, maybe the CLI can implement the biometrics itself? I assume Dashlane is also just saving the password within keychain @Mikescops?

What is the typical way of combining the ideas of an author and the contributors?

Furthermore the dashlane CLI is not ready as mentioned [here](https://github.com/raycast/extensions/pull/11326#discussion_r1536722387). So in my opinion this pull request is on hold even if everything else would be ready.

neo773:
@tmwrnr 
> Furthermore the dashlane CLI is not ready as mentioned https://github.com/raycast/extensions/pull/11326#discussion_r1536722387. So in my opinion this pull request is on hold even if everything else would be ready.

This is fixed in the latest commit, I've tested it locally and it's working great https://github.com/Dashlane/dashlane-cli/pull/229

Mikescops:
> Is this ready for a final review or does @tmwrnr and @Mikescops have any more comments?

For me this PR is a no go, the swift does not meet proper code quality standards and the security concerns raised in my first comments are not solved.

neo773:
> > Is this ready for a final review or does @tmwrnr and @Mikescops have any more comments?
> 
> For me this PR is a no go, the swift does not meet proper code quality standards and the security concerns raised in my first comments are not solved.

Partly agree with this statement, you're basing it off from the blog post you mentioned and specifically this part 
"This is actually a very naïve approach because in theory, a jailbreak tweak could hook into your app" 
macs don't have jailbreak and stuff like this is protected by System Integrity Protection.

Current implementation literally leaves the vault open, any program can call `dcli password output --json` and send the vault dump to their malicious server.

Mikescops:
I understand the initial concern, and as I mentioned in my very first comment I'd rather have this implemented in the CLI itself rather than in a third party which won't solve what you previously mentioned about accessing the vault with a command line. 
Also by making a wrapper on how we store the master password you're excluding users who have 2FA or SSO on their account which is handled by the CLI natively.
I'm going to raise this need internally and see what we can do in the short term. 
I'm not a raycast owner so this is just my two cents, you folks are deciding on if this is good to be merged.

neo773:
Yes, I said the same thing before long term Dashlane should do this natively but short term this isn't so bad.

tmwrnr:
> I understand the initial concern, and as I mentioned in my very first comment I'd rather have this implemented in the CLI itself rather than in a third party which won't solve what you previously mentioned about accessing the vault with a command line.

@Mikescops It would be nice, if the CLI could handling this kind of stuff. In Raycast there are also extensions for Bitwarden and 1Password. 

Bitwarden works with a client ID and a client secret to access the vault, without knowing exactly the master password.

1Password is also using I think a secret which is I think just token.

> Yes, I said the same thing before long term Dashlane should do this natively but short term this isn't so bad.

@neo773 I don't want to include a refactoring and a new command for short term. Yes the current solution is not the best, but with the limitation of the CLI I think it's acceptable. If these constrains does not match your security standards, which is fine, than maybe wait to use the extension until the CLI is updated.

Until a newer version of the CLI is released I could think of a prompt for the password every time you want to access the vault. The master password would only be saved in memory during runtime of the specific command. Would that be compromise?

Mikescops:
> Bitwarden works with a client ID and a client secret to access the vault, without knowing exactly the master password.

I've checked their code, the cID/secret grants access to the CLI. The CLI still handles the encryption of the vault. The vault can be open, similar to what we do, but can also be locked and then the Master Password will be requested and they will prompt this in Raycast.

We have already started discussions internally on what we can do here, I'll keep you posted.

Mikescops:
@neo773 I added a couple of settings, more to be added in the next weeks:

- https://github.com/Dashlane/dashlane-cli/pull/231
- https://github.com/Dashlane/dashlane-cli/pull/230

Mikescops:
I was not convinced by the current nodejs implementations so I built my own native module for the macos keychain and biometrics.
It's still quite early stage for now so it's likely missing some features but I think it's good to be integrated in projects.
https://github.com/Mikescops/node-native-keychain

Mikescops:
@neo773 You can start using https://github.com/Dashlane/dashlane-cli/releases/tag/v6.2415.0 and update this PR 🎉

tmwrnr:
Hi @neo773,

the Dashlane CLI was updated again, because of an error where the complete output could not be read by nodejs.

I opened a new pull request #11950 where I added the master password as an extension preference. With this you can already use the following configuration. I also refactored some code and added improved error handling. 

```sh
dcli c save-master-password false
```

The CLI now supports biometrics so we don't have to implement this in the extension. If you still want to use biometrics to unlock the vault and want to contribute to this extension, I would suggest you to create a new branch after my pull request is merged.

Thank you again for the pull requests, as this allowed the extension and CLI to be improved! I think we should close this pull request, what do you think?

neo773:
@tmwrnr 
Agreed, closing this!

Issue #11097, opened 2024-03-04
[Search Zotero] No Data found in referred sqlite db


ddbrierton:
### Extension

https://www.raycast.com/reckoning-dev/zotero

### Description

Is this extension compatible with Zotero 7 beta?

I am quite sure I have configured the extension correctly, and I can see the copied db `zotero.sqlite.raycast` and running a `shasum` on both `zotero.sqlite` and `zotero.sqlite.raycast` returns the same checksum.

However, all searches from Raycast fail. I just get an error saying that there is no data found in the database. (Zotero is working fine.)

### Steps To Reproduce

Install Zotero 7.0.0 beta. Configure Zotero Raycast extension. Attempt to search Zotero from Raycast

### Current Behaviour

Error message: no data found.

### Expected Behaviour

Entries from Zotero to be returned.

ddbrierton:
After further investigation it seems that I can use the extension if I disable the "Use BibTeX?" configuration option. However, 
this does reduce the functionality of the extension.

I have the Zotero Better BibTeX extension installed, version 6.7.163. The path to the CSL JSON file is correctly configured, and the file appears to be well-formed (opening it in VS Code showed no warnings). As it does not contain any sensitive information I am attaching it here.

[library.json](https://github.com/raycast/extensions/files/14487064/library.json)

ddbrierton:
Resolved with https://github.com/raycast/extensions/pull/11330

PR #11330, opened 2024-03-17
Update zotero extension

chnyutao:
## Checklist

### DIFF
--- a/extensions/zotero/src/common/zoteroApi.ts
+++ b/extensions/zotero/src/common/zoteroApi.ts
@@ -184,7 +184,7 @@ async function getBibtexKey(key: string, library: string): Promise<string> {
   }
 }
 
-async function openBibtexDb(): [SQL.Database, boolean] {
+async function openBibtexDb() {
   const preferences: Preferences = getPreferenceValues();
   const f_path = resolveHome(preferences.zotero_path);
   let new_fPath = f_path.replace("zotero.sqlite", "better-bibtex.sqlite");


chnyutao:
This PR fixes #11097. Sorry about the wrong link in the commit message :)

This is a work-in-progress. I have tested this fix on my device with Zotero 7.0.0-beta.68+c31a40c74 and [zotero-better-bibtex](https://github.com/retorquere/zotero-better-bibtex) 6.7.169. Need to make sure this fix does not break on previous Zotero builds.

chnyutao:
It seems somewhere between release `v6.7.131`and `v6.7.132` ([compare](https://github.com/retorquere/zotero-better-bibtex/compare/v6.7.131...v6.7.132)), zotero-better-bibtex (bbt for short) decided to
- drop `better-bibtex-search.sqlite` in favor of `better-bibtex.sqlite` (https://github.com/retorquere/zotero-better-bibtex/commit/94c0634988ff1f1fa07e0abe992bc81cedb7b81a, https://github.com/retorquere/zotero-better-bibtex/commit/b634c04234f5a54b62f7b700860c64477dadaeea)
- use `citationKey` instead of `citekey` (https://github.com/retorquere/zotero-better-bibtex/commit/63028dcd03ada843d6f55dfd6745bc92d6587ab2)

The implication is that this fix will break for users using bbt prior to `v6.7.132`, which was released in Oct, 2023.

I assume a few users would turn off auto-update for bbt and thus still using an older version. In this case, this fix actually breaks their extension.

pernielsentikaer:
@sadanand-singh could you check this?

sadanand-singh:
Can you please find a way to support both. Seems two changes were not triggered at the same time, so we need to consider both independently. Though, I am not what is optimal way to fix it, but one possible solution I can think of is:
1. look for existence of `better-bibtex.sqlite` file, if it fails, try checking for `better-bibtex-search.sqlite`. If that fails as well, may be show error to the user.
2. Use try block when attempting `BIBTEX_SQL` query. when it fails try with older `BIBTEX_SQL_OLD` query that uses `citekeys`.

Would really appreciate if you try to find a solution. Otherwise, I can find some free time next weekend, so if needed I could try something then.

Thanks for the contribution!

chnyutao:
I agree that the straightforward solution would be trying the new one and then falling back to the old one if the new one fails.

I'll try looking into this maybe Wednesday :)

chnyutao:
I've tested this with an previous bbt version (`v6.7.117`) on my device. Seems everything works fine. PTAL.

pernielsentikaer:
Awesome @chnyutao 🔥

retorquere:
> The implication is that this fix will break for users using bbt prior to `v6.7.132`, which was released in Oct, 2023.
> 
> I assume a few users would turn off auto-update for bbt and thus still using an older version. In this case, this fix actually breaks their extension.

I really wish people wouldn't do this.

sadanand-singh:
> > The implication is that this fix will break for users using bbt prior to `v6.7.132`, which was released in Oct, 2023.
> > I assume a few users would turn off auto-update for bbt and thus still using an older version. In this case, this fix actually breaks their extension.
> 
> I really wish people wouldn't do this.

Latest update in the code will NOT suffer from this @retorquere

sadanand-singh:
> I've tested this with a previous bbt version (`v6.7.117`) on my device. Seems everything works fine. PTAL.

This looks great. Thanks @chnyutao

retorquere:
> > I really wish people wouldn't do this.
> 
> Latest update in the code will NOT suffer from this @retorquere 

As in "I really wish people wouldn't stick to older versions of BBT".

chnyutao:
Cheers!

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #11384, opened 2024-03-19
github: remove section in action panel

antonengelhardt:
## Description

The "Mark as Read" action and its shortcut was not rendered and also did not work. This is because this action was considered a primary action. Removing this section makes the action secondary and it is now rendered and works as expected.
### Before 

Screenshot-Visual Studio Code-006228

### After 

Screenshot-Notion-006234

## Checklist

### DIFF
--- a/extensions/github/src/components/NotificationActions.tsx
+++ b/extensions/github/src/components/NotificationActions.tsx
@@ -134,12 +134,7 @@ export default function NotificationActions({ notification, userId, mutateList }
       />
       {notification.unread ? (
         <>
-          <Action
-            title="Mark as Read"
-            icon={Icon.Circle}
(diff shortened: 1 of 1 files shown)


xilopaint:
I also fixed this issue in another command of the same extension at #11343. Btw, you can even remove the shortcut property for the secondary action and the ⌘⏎ shortcut is assigned by default.

antonengelhardt:
> I also fixed this issue in another command of the same extension at #11343. Btw, you can even remove the shortcut property for the secondary action and the ⌘⏎ shortcut is assigned by default.

Thanks for the hint! normally i am a fan of writing things explicitly, if it doesn't make it slow or not work, but i guess its fine here.

thomaslombart (approved):

PR #11386, opened 2024-03-19
Add workspacey extension

3llomi:
## Description

Create and manage your workspaces, execute multiple actions at once.
Add Actions to start your workspace. eg. Open Files, Open Folder, Open App, Open Links (with deeplinks support), Exexcute a Command, and Add Notes
You can create a workspace to start ALL your defined actions at once

## Screencast

https://github.com/raycast/extensions/assets/8546946/d51d3fa2-b183-4bcc-96da-e1e019da83a0

## Checklist


mil3na:
Hi!

Your extension has a lot of potential! I like it!
A few questions:
- What does the "Add Note" action do? I created a "workspace" with just a note and nothing happened.
- Terminal commands also do nothing.
- It would be nice to choose which app to open the terminal command.

Other than that, I think this extension will be a great addition to our store.

3llomi:
Hi @mil3na ,
Glad that you liked it!

> What does the "Add Note" action do? I created a "workspace" with just a note and nothing happened.

 'Add Note' will just add a note for the Workspace that you can view later. something like the 'SSH Password' which is NOT 
executable.
workspacey 2024-03-26 at 16 46 32
workspacey 2024-03-26 at 16 46 36
> Terminal commands also do nothing.

Terminal commands uses the 'exec' API, not sure why it's not working for you though.
Can you share the command that you've tried so I can take a look on my side?


> It would be nice to choose which app to open the terminal command.

As mentioned, it uses the exec API so it will not use a specific terminal app like Warp or something.

mil3na:
The note feature seems confusing. I find it hard to use it. I would suggest either making it so you could read the note in a separate "screen" or remove it for now.

I tried creating a simple command with `ls -la`. Also, now that you mention that it uses exec without any parsing from the extension side, I wonder how easy would be to just execute malicious code with that. I'm a bit wary.

3llomi:
I can remove the note for now.

and for the terminal command I do understand that it can be used to execute malicious code, perhaps I can add a warning for the user while adding a terminal command or something?

I think of it for advanced usage something like `code /path/` to open a project in VSCode and so on.

Let me know your thoughts.

3llomi:
Hi @pernielsentikaer 👋 , Welcome back!

>It feels a bit like [Xecutor](https://www.raycast.com/GastroGeek/xecutor) but with more features, did you see that one before?

I've never seen this before, but it has the same idea - kinda, except that Workspacey has more actions like opening specific folders, opening specific files, executing commands using `exec` .
Should I open a PR instead?

>It feels weird when you create a workspace that you can only add one thing, to add more you need to find the workspace and add action this way

Good point, I think I can keep the 'Add Action Form' open after adding an action and resetting the fields so the user can add another action.

>Would be nice if all actions has icons

You mean that the user should be able to pick an icon for a specific action?

pernielsentikaer:
It could indeed work in Xecutor but it might also require a lot of rewrite to add everything from this PR.

> Good point, I think I can keep the 'Add Action Form' open after adding an action and resetting the fields so the user can add another action.

I think that would be a good idea, it felt a bit weird to only add one since it's an extension for multiple actions

> You mean that the user should be able to pick an icon for a specific action?

That would also be cool but not nessesary, I'm talking about items in the <ActionPanel> - would be nice if they all had icons

image

3llomi:
I'm not sure if I should open a PR in Xecutor, if we can stick with Workspacey as a separate extension it'd save me a lot of time 😄

pernielsentikaer:
It doesn't make that much sense to have two extensions doing the same and it would fit into Xecutor just fine (maybe some tweaks in the UI) 

Would you be up for that, or should I do it? I think I'm able to do it within a few weeks 🙂

pernielsentikaer:
So you want to try the enhancement or should I?

3llomi:
Sorry I can't take this at the moment as I got busy a bit 😊

pernielsentikaer:
I can attempt to tackle it tomorrow, but it will require some time 🙂

PR #11400, opened 2024-03-20
Fix the detail view to display multiline text easily

nagauta:
## Description

I Fixed so that line breaks are displayed as new lines.

### before I fix
スクリーンショット 2024-03-20 18 33 21

### after I fix
スクリーンショット 2024-03-20 18 32 54

## Checklist

### DIFF
--- a/extensions/change-case/src/change-case.tsx
+++ b/extensions/change-case/src/change-case.tsx
@@ -56,9 +56,9 @@ function modifyCasesWrapper(input: string, case_: CaseFunction) {
   const modifiedMarkdownArr: string[] = [];
   const lines = input.split("\n");
   for (const line of lines) {
-    const modified = line.length === 0 ? "\u200B" : case_(line);
+    const modified = case_(line);
     modifiedRawArr.push(modified);
-    modifiedMarkdownArr.push(modified + "\n");
+    modifiedMarkdownArr.push((modified.length === 0 ? "\u200B" : modified) + "\n");
   }
   return { rawText: modifiedRawArr.join("\n"), markdown: modifiedMarkdownArr.join("\n") };
 }


pernielsentikaer:
Works as expected for me 🔥

@erics118 do you want to test this too?

extensions/change-case/src/change-case.tsx
```diff
@@ -52,12 +52,15 @@ async function readContent(preferredSource: string) {
...
 function modifyCasesWrapper(input: string, case_: CaseFunction) {
-  const modifiedArr: string[] = [];
+  const modifiedRawArr: string[] = [];
+  const modifiedMarkddownArr: string[] = [];
   const lines = input.split("\n");
   for (const line of lines) {
-    modifiedArr.push(case_(line));
+    const modified = case_(line);
+    modifiedRawArr.push(modified);
+    modifiedMarkddownArr.push(modified + "\n");
```
erics118, line 61: typo: `Markddown` has an extra `d`

erics118, line 61: Also, should there be a case to handle empty lines? eg:
```js
modifiedMarkdownArr.push((modified.length === 0 ? "\u200B" : modified) + "\n");
```

erics118:
Other than those two things, it looks good 👍

extensions/change-case/src/change-case.tsx
nagauta: @erics118 
thanks you, it does make sense.
plz check my update.
I modified by checking the line directly instead of checking what was modified.

pernielsentikaer:
Do you want to check it a last time @erics118

extensions/change-case/src/change-case.tsx
```diff
@@ -52,12 +52,15 @@ async function readContent(preferredSource: string) {
 }
 
 function modifyCasesWrapper(input: string, case_: CaseFunction) {
-  const modifiedArr: string[] = [];
+  const modifiedRawArr: string[] = [];
+  const modifiedMarkdownArr: string[] = [];
   const lines = input.split("\n");
   for (const line of lines) {
-    modifiedArr.push(case_(line));
+    const modified = line.length === 0 ? "\u200B" : case_(line);
```
erics118, line 59: I think it's best to only have `modifiedMarkdownArr` contain the zws. The user probably shouldn't have to copy a zws.

erics118 (changes requested):

extensions/change-case/src/change-case.tsx
nagauta: Ah, erics118, is my understanding correct that you add only a zws becasue CommonMark could be made to line feed with 2 spaces?

erics118: I used a zws to create force a newline to be made. Just having two spaces doesn't work in Raycast.

I suggest just having:
```js
  for (const line of lines) {
    let modified = case_(line);
    modifiedRawArr.push(modified);
    modifiedMarkdownArr.push((modified.length === 0 ? "\u200B" : modified) + "\n");
  }
```

nagauta: i see. i fixed again plz check this again

erics118 (approved):
Looks good now 👍.

pernielsentikaer (approved):

PR #11408, opened 2024-03-20
Update crossbell extension

hyoban:
## Description

This PR update all dependencies, and add a new command `upload-file`.
It can upload files from your system clipboard or local file to IPFS.

## Screencast

From clipboard

https://github.com/raycast/extensions/assets/38493346/8df55c39-3ab3-4a52-b5c0-38b4be614356

From finder

https://github.com/raycast/extensions/assets/38493346/bc59c44e-8a32-45fd-bf1c-1c8cec42b9af

## Checklist

### DIFF
--- a/extensions/crossbell/package.json
+++ b/extensions/crossbell/package.json
@@ -15,14 +15,14 @@
   "commands": [
     {
       "name": "upload-file-from-clipboard",
-      "title": "Upload File From Clipboard",
+      "title": "Upload File from Clipboard",
       "subtitle": "Crossbell",
       "description": "Upload file from clipboard to IPFS",
       "mode": "no-view"
     },
     {
       "name": "upload-file-from-finder",
-      "title": "Upload File From Finder",
+      "title": "Upload File from Finder",
       "subtitle": "Crossbell",
       "description": "Upload selected file items from Finder to IPFS",
       "mode": "no-view"


extensions/crossbell/src/upload-file.tsx
```diff
@@ -0,0 +1,60 @@
+import { Clipboard, showToast, getSelectedFinderItems, showHUD, closeMainWindow } from "@raycast/api";
+import fs from "fs";
+import fsp from "fs/promises";
+import { ipfsUploadFile } from "crossbell/ipfs";
+import { fetch, FormData } from "node-fetch-native";
```
Songkeys, line 5: I don't think we need this cuz node 18+ has `fetch` built-in.

extensions/crossbell/src/upload-file.tsx
```diff
@@ -0,0 +1,60 @@
...
+      const filePaths = selectedItems.map((item) => item.path).filter((path) => fs.lstatSync(path).isFile());
+      if (filePaths.length > 0) {
+        const files = await Promise.all(
+          filePaths.map(async (path) => {
+            const data = await fsp.readFile(path);
+            return { name: path, data };
+          }),
+        );
+        console.log({ files });
+        const results = await Promise.all(
```
Songkeys, line 26: 1. I'm worried that there will be a huge load of files. Would a request queue be better? For example, use `p-limit`.
2. Should we use `Promise.allSettled` to handle failed requests?

Songkeys (changes requested):

extensions/crossbell/src/upload-file.tsx
hyoban: Tested on my local machine, there is no fetch and FormData, so I added node-fetch-native.

extensions/crossbell/src/upload-file.tsx
hyoban: Oh yes, I will make a change here

Songkeys (approved):

mil3na (changes requested):
Hi!

A few comments:
- The upload file command doesn't specify where the file will be coming from. By activating it, I only get a "No file found to upload" error. Oddly enough, I also get a green status.
- It would be interesting to either split the command in two (upload from the clipboard and from finder) or add a step. Right now it is a bit confusing to the user.
- Please, add this new command to the README

mil3na (approved):
LGTM!

PR #11485, opened 2024-03-25
[Sonos] handle error on toggleplayback

AntonNiklasson:
When there's no media selected for the active group the "toggle playback" command fails. This change catches that error and shows a failure toast instead of just crashing.

Screenshot 2024-03-25 at 14 06 41


AntonNiklasson:
@pernielsentikaer would be great to get this in 🙌

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #11492, opened 2024-03-25
Add rayvops extension

Clement-Beudot:
## Description

This Raycast extension is designed to boost productivity by providing quick and easy tools for common tasks. It includes features for modifying JSON data, handling strings, and opening multiple links at once.

## Checklist

### DIFF
--- a/extensions/rayvops/src/extractJsonFieldValues.tsx
+++ b/extensions/rayvops/src/extractJsonFieldValues.tsx
@@ -14,7 +14,7 @@ export default function ExtractJsonFieldValues() {
         try {
           let parsedJson = JSON.parse(text);
           if (!Array.isArray(parsedJson)) {
-            parsedJson = [parsedJson]; 
+            parsedJson = [parsedJson];
           }
           if (Array.isArray(parsedJson) && parsedJson.length > 0) {
             setJson(parsedJson);
--- a/extensions/rayvops/src/modifyAndFilterJsonValues.tsx
+++ b/extensions/rayvops/src/modifyAndFilterJsonValues.tsx
@@ -23,7 +23,7 @@ export default function ModifyAndFilterJsonValues() {
         try {
           let json = JSON.parse(text);
           if (!Array.isArray(json)) {
-            json = [json]; 
+            json = [json];
           }
           setOriginalJson(json);
           const initialFields = Object.keys(json[0] || {});


pernielsentikaer (changes requested):
Hi 👋

Sorry for the long response time. Due to the Easter vacation, we have a slightly increased response time, but it should soon return to normal.

Maybe it would be better to focus on the command opening multiple URLs at once and the other features could maybe fit into the [Toolbox extension](https://www.raycast.com/Kang/toolbox)

Clement-Beudot:
> Hi 👋
> 
> 
> Sorry for the long response time. Due to the Easter vacation, we have a slightly increased response time, but it should soon return to normal.
> 
> Maybe it would be better to focus on the command opening multiple URLs at once and the other features could maybe fit into the [Toolbox extension](https://www.raycast.com/Kang/toolbox)

Hello @pernielsentikaer , thank you for your comment. I check the Toolbox Extension, that is nice, but there is no way to configure shortcuts for commands :/ . This extension was built to chain action (ex : Extract value from a Json attribute, then concatenate those values, etc). 

What do you think ?

pernielsentikaer:
I tried the `Extract Json Fields Valiues` by having `{"hello": "you"}` in the clipboard, but I just gets an error:

rayvops-1

Seems to be the same with other commands, can you look into that?

Clement-Beudot:
> I tried the `Extract Json Fields Valiues` by having `{"hello": "you"}` in the clipboard, but I just gets an error:
> 
> rayvops-1
> 
> Seems to be the same with other commands, can you look into that?

Hello @pernielsentikaer . Thank you for testing. Commands to extract Json attribute values or modify them in batch are made to process many lines on data (like array of object). I add a fix to check if Json is an array. 
So 
 {
    "hello": "you"
 }
will be treated in the same way as : 
[
  {
    "hello": "you"
  },
  {
    "hello": "there"
  }
]

pernielsentikaer:
That part works a lot better now, where does the name of the extension come from? 🙂

Clement-Beudot:
> That part works a lot better now, where does the name of the extension come from? 🙂

Thx, the name comes from a play on words between Raycast and Revops (Revenue Operation Team).

pernielsentikaer (commented):
It does feel like it's more an internal tool (Also by reading the description of the extension) rather than being something that would work for everyone, is that the case?

Clement-Beudot:
> It does feel like it's more an internal tool (Also by reading the description of the extension) rather than being something that would work for everyone, is that the case?

Hello, 
For sure this extension will be used a lot internally, but not only. 
In my business, it's often useful to be able to extract data from a Json result (or a CSV, but that could be for a future update), concatenate strings to use them in SQL queries, and open urls easily from these results. 

I could publish the extension locally, but people outside my organization wouldn't be able to take advantage of it. 
Is it possible to publish an extension across several organizations?

pernielsentikaer:
You can push it to all the organizations you want if you change the owner of them yes 🙂

I think it's the mix of json/string manipulation that feels a bit weird to me, also descriptions is not really saying anything besides showing love 🙂

We can do somepush to get it in the store but we might need some more polishing 🙂

PR #11511, opened 2024-03-27
[Bitwarden] Fix vault timeout bypass and improve lock/logout safety

### LINKED ISSUE #11360: [Bitwarden Vault] The vault is not locked when the 'vault timeout' is reached (even if it let you think it is) (edit: Bug introduced by commit #7899198)
### Extension

### TLDR:
- I suspect that a lot of users of this extension thought (since the introduction of this bug 4months ago) that their vault got locked while it never was (except if/when they manually call the lock vault command).
- the error is in the lock() method of the Bitwarden class (bin/bitwarden.ts)
    - we should probably replace the `!==` by `===`  is this line of code:
    - `if (result.status !== "unauthenticated") return { error: new NotLoggedInError("Not logged in") };`
- This bug was apparently introduced ~= 4months ago by this commit: https://github.com/raycast/extensions/commit/7899198909fdbec2819cce54f509496433265fef
- Since then the 'bitwarden.lock()' method did not lock anything anymore.
       
### Description of the issue:
- In the configuration of the Bitwarden extension you can specify a ‘Vault timeout’ (’immediately’, ‘1min’, ‘5min’,…)
- when this timeout is reached the vault is supposed to get locked.
- You’ll think that it does get locked because you will see the login form if you try to search your vault but it actually isn’t locked!
- You can reproduce like that:
    - in order to easily test, set the ‘vault timeout’ to ‘immediatly’
    - try to search your vault, and you will see the login form (with a vault status ‘Locked’ and the message “Vault Timed out due to inactivity”
    - Do not unlock it (do not type your password).
    - go to the configuration of the extension and change the ‘vault timeout’ to something else (’15min for example’)
    - try to search your vault again now …. and magically you can scan and access your full vault.
- This is quite critical because anybody can then ‘unlock’ your vault without having to enter any password while you did think that it was locked.
- Hopefully the ‘manual lock’ (when you call the ‘Lock Vault’ command of this extension) does lock the vault but I fear that a lot of people never use that command since they probably trust the ‘timeout lock’
- Another proof that the vault is actually not locked is that if, after a ‘timeout locked’, you do call the ‘lock vault’ command, you will see the success message: ‘Vault successfully locked’
    - but according to the code in src/lock-vault.tsx, if the `SessionStorage.getSavedSession()` call return an undefined token this command should return the ‘Already locked or not logged in” error message.
    - if after having call the ‘Lock Vault’ command, you try to call it again, now you will see that error message, which proof that when using this manual lock command, the token is actually removed from the local storage

###  Investigations results, source of the bug and solution:
  - so, we saw that after a ‘locked timeout’, the `SessionStorage.getSavedSession()` still does return a token
  - that means that the token is still in the localStorage ⇒ `LocalStorage.getItem<string>(LOCAL_STORAGE_KEY.SESSION_TOKEN)`
  - which means that the `SessionStorage.clearSession()` method was not called  (which was supposed to do: `LocalStorage.removeItem(LOCAL_STORAGE_KEY.SESSION_TOKEN)`)
  - which mean that the `handleLock()` callback of ‘bitwarden’ (useBitwarden) is apparently not called
  - there is then probably a bug in the `bitwarden.lock()` method (called in the useOnceEffect of the SessionProvider) since it does not actually lock the vault and also don’t call the callback.
      - here is the lock() method of the Bitwarden class (bin/bitwrden.ts)
      
      ```jsx
      async lock(reason?: string, shouldCheckVaultStatus?: boolean): Promise<MaybeError> {
          try {
            if (shouldCheckVaultStatus) {
              const { error, result } = await this.status();
              if (error) throw error;
              if (result.status !== "unauthenticated") return { error: new NotLoggedInError("Not logged in") };
            }
      
            if (reason) await this.setLockReason(reason);
            await this.exec(["lock"], { resetVaultTimeout: false });
            await this.callbacks.lock?.(reason);
            return { result: undefined };
          } catch (execError) {
            captureException("Failed to lock vault", execError);
            const { error } = await this.handleCommonErrors(execError);
            if (!error) throw execError;
            return { error };
          }
        }
      ```
      
      - The error is apparently because the ‘result.status’ is not “unauthenticated” and that this method then return an object with an error field but that we don’t check the response when we call this lock() method in the ‘SessionProvider.
      - That seems very strange to expect the bitwarden status to be “unauthenticated” when we want to validated that we are logged in…
      - we should probably simply replace the `!==` by `===`
      - By looking in the commit history of this file I found that the bug was introduced about 4months ago by the commit: 
      https://github.com/raycast/extensions/commit/7899198909fdbec2819cce54f509496433265fef
      - when this:
      ```jsx
     async lock(reason?: string, shouldCheckVaultStatus?: boolean): Promise<void> {
    if (shouldCheckVaultStatus) {
      const isAuthenticated = (await this.status()).status !== "unauthenticated";
      if (!isAuthenticated) return;
     ```
     - was replaced by this:
     ```jsx
       if (shouldCheckVaultStatus) {
        const { error, result } = await this.status();
        if (error) throw error;
        if (result.status !== "unauthenticated") return { error: new NotLoggedInError("Not logged in") };
      }
      ```
  - I then tried to understand why we do see the login form after a ‘lock timeout’ even though the vault is actually not locked:
      - it’s because the state (returned by the `useSessionReducer()` do return ‘locked: true’ after a ‘timeout lock’ which make the extension think that the vault is locked and thus display the login form.
      - Why is this ‘locked’ flag set to true?
      - because of the “`dispatch({ type: "loadSavedState", ...restoredSession });`” in the ‘useOnceEffect’ of the SessionProvider’
      - the ‘restoredSession’ (returned by `getSavedSession();` ) returns a ‘shouldLockVault: true’
          - this is because it compare the Time.now with the lastActivityTime and sees that we did reach the timeout. All good.
      - but the ‘loadSavedState’ action of the ‘useSessionReducer’ reducer does that ⇒
          
          ```jsx
          return {
                    ...state,
                    token: action.token,
                    passwordHash: action.passwordHash,
                    lastActivityTime: action.lastActivityTime,
                    isLoading: false,
                    isLocked: action.shouldLockVault || !hasToken,
                    isAuthenticated: hasToken,
                    lockReason: action.lockReason,
                  };
          ```
          
      - ⇒ the ‘isLocked’ flag is set to true if the action.shouldLockVault is true. which is the case (as explained above)
      - I think that this is a first mistake. the state thinks that the vault ‘isLocked’ when it actually ‘shouldLock’
      - this is bad because as we do see the login form we all think that the vault did get locked… when it didn’t.
      - changing that would at least allow us to see that the vault was actually not locked

So, again: TLDR:
- The main error is apparently in the lock() method of the Bitwarden class (bin/bitwrden.is)
    - and that we should probably replace the `!==` by `===`  is this line of code:
    - `if (result.status !== "unauthenticated") return { error: new NotLoggedInError("Not logged in") };`
- it would also maybe be good to listen to the response of the lock() method to confirm that the lock call succeeded.
- and maybe that the isLocked flag of the session state should not be set to true with the shoudlLockVault is true.

Thank you for reading me ;-)

### Steps To Reproduce

- in order to easily reproduce without having to wait, set the ‘vault timeout’ setting (in the configuration of the extension) to ‘immediatly’
- try to search your vault, and you will see the login form (with a vault status ‘Locked’ and the message “Vault Timed out due to inactivity”
- Do not unlock it (do not type your password).
- go to the configuration of the extension and change the ‘vault timeout’ to something else (’15min for example’)
- try to search your vault again now 
….. magically you can scan and access your full vault and you didn't enter any password.  

See the description above for the explanations and my investigation results and the potential solution.

### Current Behaviour

The vault get not locked when the 'lock timeout' (the option that you can set in the 'Vault Timeout' field in the configuration of this extension) is reached.
It does show the login form making you think that the vault is actually locked (but this is due to another bug masking the main one) but is it actually not locked.

### Expected Behaviour

The vault should get locked when the locked timeout is reached.

jomifepe:
## Description

Fixes #11360

### What was done
1. Fixed easy way of unlocking the vault, simply by increasing the timeout, introduced by a wrong condition statement.
2. Improve session bootstrapping logic, moved it all into one place, simplified the reducer logic, and made it more predictable
3. Start saving auxiliary vault status on local storage, to quickly know if the vault was previously locked, and not have to rely on the CLI's slow and inconsistent status. 
4. Lock and Logout commands now still "succeed" if the CLI call failed, by still clearing the session and cache.

## Screencast

Not much has visually changed.

## Checklist

### DIFF
--- a/extensions/bitwarden/CHANGELOG.md
+++ b/extensions/bitwarden/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Bitwarden Changelog
 
-## [Fixes and Improvements] - 2024-03-27
+## [Fixes and Improvements] - 2024-03-28
 
 - Fixed lock screen bypassing by increasing timeout
 - Lock and Logout commands now still succeed even if the CLI call fails


pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #11571, opened 2024-03-30
feat(spotify-player): trigger current track command automatically

erics118:
## Description

Automatically trigger the current track command to update its state when like/unlike/previous/next command is triggered

this way, the current track command is won't be delayed or require the user to manually refresh it

## Screencast

https://github.com/raycast/extensions/assets/52634785/be319d33-5920-4d17-82e7-60c6fb106f43

(in the video, showHud was replaced with console.log so that it wouldn't automatically close the Raycast window)

## Checklist


thomaslombart (commented):
I think the `launchCommand` calls need to be enclosed in a `try/catch` statement that fails silently, otherwise this may fail for users that don't have the `Current Track` command enabled.

erics118:
I thought so too, but from my testing, there were no errors outputted, so I didn't include the try/catch. Do you mind testing this to confirm?

thomaslombart:
Indeed! Let's ship it then 🙂

thomaslombart (approved):

PR #11576, opened 2024-03-31
Update obsidian extension

xzthemeofsss:
## Description

This pull request added a new command which simply creates a new Obsidian note with the selected text as the note title. 

Besides, it also changed the "append to daily note" command behavior to support selected text or clipboard text as the fallback text input. 

## Checklist

### DIFF
--- a/extensions/obsidian/CHANGELOG.md
+++ b/extensions/obsidian/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## [New features] - 2024-04-01
 
 - Add a new command which simply creates a new Obsidian note with the selected text as the note title.
-- Support selected text or clipboard content as the fallback input in `Append to daily note` command
+- Add a new command which append the selected text to the daily note.
 
 ## [Bugfix for nested bookmarks] - 2024-02-06
 
--- a/extensions/obsidian/package.json
+++ b/extensions/obsidian/package.json
@@ -266,9 +266,48 @@
           "name": "text",
           "placeholder": "Take out the trash",
           "type": "text",
-          "required": false
+          "required": true
         }
       ],
+      "preferences": [
+        {
+          "name": "appendTemplate",
+          "type": "textfield",
+          "title": "Template for Daily Note Append action",
+          "required": false,
+          "description": "Specify a template for Daily Note Append action (e.g. '- {content}')"
+        },
+        {
+          "name": "vaultName",
+          "type": "textfield",
+          "title": "Name of Obsidian vault where note is",
+          "required": false,
+          "description": "Name of Obsidian vault where note is"
+        },
+        {
+          "name": "heading",
+          "type": "textfield",
+          "title": "Name of heading in note in which to append",
+          "required": false,
+          "description": "If no heading is set, text will be appended to the end of the daily note"
+        },
+        {
+          "name": "silent",
+          "type": "checkbox",
+          "label": "Silent Mode",
+          "title": "Silent Mode",
+          "required": false,
+          "description": "Don't open daily note when appending to the daily note.",
+          "default": true
+        }
+      ]
+    },
+    {
+      "name": "dailyNoteAppendFromSelectedTextCommand",
+      "title": "Append to Daily Note from Selected Text",
+      "subtitle": "Obsidian",
+      "description": "Append selected text to your daily note",
+      "mode": "view",
       "preferences": [
         {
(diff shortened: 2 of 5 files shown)


KevinBatdorf:
@xzthemeofsss would you mind filling out the description? When I test it I want to make sure I understand the intent fully.

Also the build is failing I think because it needs a changelog entry

pernielsentikaer:
Could you test it again @KevinBatdorf

extensions/obsidian/src/dailyNoteAppendCommand.tsx
```diff
@@ -19,14 +29,21 @@ interface DailyNoteAppendArgs {
...
+      try {
+        const selectedText = await getSelectedText();
+        const { text: clipboardText } = await Clipboard.read();
+        const text = props.arguments.text || selectedText || clipboardText;
+        const content = await applyTemplates(text, appendTemplate);
+        setContent(content);
+      } catch (error) {
+        closeMainWindow();
+        console.error("Error in getting content", error);
+      }
```
KevinBatdorf, line 46: This has the potential to fail for every command if the clipboard fails. Better would be to move the content line outside of the try/catch so it always runs

Also, do you think this might be too surprising of a feature? If the user opens the daily append and presses Enter, I think I would expect it to error, or have the text input pre-filled with the selected/clipboard text, but if it could also then be selected in Raycast so it could be easily deleted (to avoid adding an extra keypress)

What do you think?

xzthemeofsss: Excellent suggestion! But I am not sure if displaying error message to users is a good choice cause if the user just wanna input something manually but they got a clipboard fails message, should they feel confusing?

KevinBatdorf: I'd have the try/catch fail silently, so have an empty catch

I don't think this feature (pre filling the input) will work though unless the input can be fully selected. Otherwise asking the user to backspace the entire message would be pretty bad.

There's also the case of having lots of text selected or in the history. For example, just copied an entire website text.

Maybe selected text needs to stay separate? 

Do you personally plan to use the clipboard fallback? Seems like the user could just press cmd v right? Maybe it's not that useful?

xzthemeofsss: The problem I just want to solve initially is to support selected text as the daily notes input. The clipboard as input feature was a complement by the way. I agree to separate into two commands to be clearer.

KevinBatdorf: Maybe just to do the one command then? Seems like a good feature and dont have to worry about side effects

pernielsentikaer:
Thanks for helping out @KevinBatdorf 🔥

KevinBatdorf:
What do you think about the feature in general @pernielsentikaer ? 

From a purely UX standpoint, rather than creating new commands for this I'd prefer if we could instead pre-populate the argument with something (selected text / clipboard text) programatically. That's not possible, however, as the command hasn't loaded yet and there's no mechanism in Raycast to configure this (I asked the other day on Slack).

My only concern is adding two new commands for something possibly not that useful. A user could just have selected text, press cmd+c to copy it, invoke Raycast, type the existing command, and press cmd+v to paste it in. So these two new commands are really just here to save two keystrokes (copy/paste) that I don't think many people are using. And copy/paste is already in most user's muscle memory.

But this is a community extension so I'm not entitled to gatekeep or deny any requested features. I'm happy to code review and give input though, of course. So maybe you want to make the call? The code itself looks fine (minus one linter error in the title)

pernielsentikaer:
@KevinBatdorf I guess you could have argument with "default action" (and the other actions) and in preferences select the one which is default - I'm not sure it would work in this context though?

KevinBatdorf:
It might work if the we just add a setting (globally I think?) like "fallback to selected text or clipboard" that's disabled by default. 

Then 
1. Make every argument optional
2. Right before insert, check the argument value and if empty, fallback

Making it off by default would avoid users with custom implimentations from getting new behavior unexpectedly

@xzthemeofsss what do you think? Does this make sense? I know you had something similar in the insert part so maybe it's easy to remove the two commands and just add this back? I'd do it but I can't push to the PR

xzthemeofsss:
Sounds good! Thank u guys for advice. Will do it later.

Issue #4744, opened 2023-02-12
[1Password] Work with multiple vaults at the same time


mgonto:
### Extension

https://www.raycast.com/khasbilegt/1password

### Description

I'd love to be able to connect to multiple vaults at the same time

### Who will benefit from this feature?

anybody who has multiple vaults

dhnaranjo:
I'd definitely like this feature.

pprotas:
This would be awesome. I have a work account and a personal account. When installing the extension I chose my personal account, and as far as I can tell there is no way for me to switch to my work account.

I want to be able to see the passwords of both accounts in the same raycast window.

vitorf7:
This definitely would be amazing, rather than having to use the 1Password Quick Access to search for a password which does allow to work with multiple accounts, I would prefer to use this with Raycast instead.

dalposenrico:
I saw more issues have been opened about this, adding my comment just to say this would be really nice

dagfinnhagen:
Same here, but would like the option to choose which vaults it should search. Right now I'm using 6 valuts, but when I'm at work I need it to search only my "work vaults".. It seems to me now that is searches all vaults so would be nice with an option to be able to disable/enable which vaults should be included in the search.

kpeatt:
@Luk4h How do you tell Raycast which account you want to use? It seems to choose (what seems like) randomly for me.

Luk4h:
Per the 1Password documentation, it auto sign-in on the lastest signed on account. If you reset the extension cache it should ask for the account.

On another note, I am working on an action that will allow to switch to another account, as well other vaults.

I am looking for input on the Vaults... Currently I think the approach should be a new command "My Vaults" where you are allowed to select the vault you want to search the passwords on.

jntme:
I'd love that feature too. The Alfred App (https://alfred.app/workflows/alfredapp/1password/) searches all vaults per default, this is one of the features I miss most about my whole Raycast experience.

James-DeLorenzo:
I just ran into this issue when adding my work account to my 1password. I am running v8. 

Raycast kept throwing an error until I ran the `op` cli command myself and it was asking for which vault to default to. Then this extension started working again. However I don't see in this extension's preferences to select which account to default.

 To fix this issue I think it would be handy to keep a list of the current accounts logged in to change what vaults to search. And on top of that which vault in said account to list as well.

Luk4h:
[...]

PR #11605, opened 2024-04-03
[1password]: Display cached Items while refreshing

luizvergennes:
## Description

Refactored the current "My Passwords" command to use Raycast's useCachedPromise hook, to work with optimistic updates. With this there are going to be way less waiting time for fetching passwords. Also I removed the cache timeout, since the updates are fetched on "background" every time the user uses the command, bringing newly added passwords, instead of waiting 10 minutes each time.

Also added a new command that is able to auto renewal the CLI authorization by itself. According to 1Password CLI documentation, every request refreshes the 10 minute cooldown time for a new manual authorization. This command simply runs each 9 minutes getting the account info. For now I left it as a new command for it to be simple and the user understand what it does, but if it is preferred I could add to the main command with no problems.

## Checklist


Luk4h:
Currently working on a command that renews the Authorization each 9 minutes to avoid the unlock screen everytime I need to use a password...

Luk4h:
Also there is a hard limit for the auto renewal of 12 hours, according to the [documentation](https://arc.net/l/quote/msmegmbi).

Luk4h:
Last added commits, fix the issue where even if the items are loaded from cache you are not able to use it's actions. I also replaced the useCachedPromise with useExec that is focused in executing cli commands (pretty nice Raycast gj). This improved a lot of the performance.

pernielsentikaer:
Thanks for the update @Luk4h 🔥 

@khasbilegt do you want to check this?

Luk4h:
@pernielsentikaer This now should solve a lot of the issues regarding the extension. After getting this one merged I plan on working on issue #4744. Which has been long over due. Should be a simple fix, but I would like to put it in another PR since this one is getting WAY bigger than I expected. I have been using the this local version on my workflow and it has been working  seamlessly.

pernielsentikaer:
Thanks for taking a stab - I tested it ok but would still like if @khasbilegt could check too

khasbilegt:
Looks good to me, awesome work @Luk4h 👍🏻

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #11617, opened 2024-04-03
Spotify: Support "Your Library" as a fallback command

ksunats:
## Checklist

### DIFF
--- a/extensions/spotify-player/package.json
+++ b/extensions/spotify-player/package.json
@@ -24,7 +24,8 @@
     "hobhouse",
     "bkeys818",
     "jatindotdev",
-    "devbear"
+    "devbear",
+    "ksunats"
   ],
   "license": "MIT",
   "preferences": [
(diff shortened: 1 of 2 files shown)


apaofficial:
Hi, this is just a minor change to allow playing specific playlist through quicklink context query - something I consider a basic usecase and something I saw multiple discussions about dating back to 2022. Hope it's OK. Thanks!

thomaslombart (commented):
Thanks for the PR! What is this solving exactly? I thought it was to support `Your Library` as a fallback command but it seems to already work without these changes.

thomaslombart:
Hey there, thanks for the PR! The extension is designed to be in line with Spotify, so for consistency reasons, we'll keep it as `Your Library` ☺️

erics118:
From my experience, it seems that Raycast automatically supports adds in support for commands that don't explicitly support being fallback commands. I've experienced this with numerous commands. Not sure if this is the correct behavior though.

thomaslombart:
> From my experience, it seems that Raycast automatically supports adds in support for commands that don't explicitly support being fallback commands. I've experienced this with numerous commands. Not sure if this is the correct behavior though.

Yeah, it's expected on our end, which is why I'm wondering what's the goal of this PR 🙂

PR #11680, opened 2024-04-07
Add brandfetch extension

yonbergman:
## Description

A simple extension that let's you quickly search for brand logos on Brandfetch

## Screencast
brandfetch-3

## Checklist


pernielsentikaer:
Hi 👋

Thanks for the contribution 🔥 

Sorry for the long response time. Due to the Easter vacation, we have a slightly increased response time, but it should soon return to normal.

We already have #11248 in the queue, maybe some features from yours could enhance that one? Currently i'm waiting for the author to respond

yonbergman:
> Hi 👋
> 
> Thanks for the contribution 🔥
> 
> Sorry for the long response time. Due to the Easter vacation, we have a slightly increased response time, but it should soon return to normal.
> 
> We already have #11248 in the queue, maybe some features from yours could enhance that one? Currently i'm waiting for the author to respond

Yeah, there's opportunity to add some of his stuff into mine I guess. (Because he's using the paid API)
How do you usually handle multiple similar extensions proposed at once?

yonbergman:
> Perhaps it makes sense to keep them separate and then we just need to ensure clarity regarding the naming, so end users can easily select the appropriate option. Does that make sense?

Sure does naming this Brand Fetch (Free) make sense?

pernielsentikaer:
Name doesn't have to be the name, could also be something like "Search Brands" and mention in description somehow

PR #11699, opened 2024-04-08
Update acqua extension

FezVrasta:
## Description

There seem to be a bug with the `useCachedState` hook, it will reset its state once in a while... I'm not sure if you guys are tracking it already.

By the way, I'm now displaying the empty drop icon when no cached state is present so users don't forget to drink if Raycast decides to forget their last drink.

## Checklist

### DIFF
--- a/extensions/acqua/CHANGELOG.md
+++ b/extensions/acqua/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Acqua Changelog
 
+## [Bug Fixes] - 2024-04-08
+
+- On first use, the icon will display the empty drop rather than the full drop.
+
 ## [Bug Fixes] - 2024-03-21
 
 - The empty drop icon would be displayed only if the time left was exactly 0, and not if it was a negative number.


pernielsentikaer:
Can you reproduce the reset of `useCachedState` somehow or give me an example?

FezVrasta:
@pernielsentikaer  once in a while the value is deleted from the persistent storage. I'm not sure how to reproduce it, just using the extension and sometimes it goes back the initial state.

pernielsentikaer:
The cache doesn't guarantee that the values will be retained. There's a limit to the size of the cache and older entries are removed.

You should use [LocalStorage](https://developers.raycast.com/api-reference/storage#api-reference) instead for values that should be retained indefinitely 🙂

pernielsentikaer:
Is this something you want to look into in this PR or later on?

FezVrasta:
Thanks for the info, I wasn't aware of this difference. I'll address it in a separate PR, this change is needed regarding.

pernielsentikaer:
Let's merge 🙂

Issue #11706, opened 2024-04-08
[Harvest] Tasks do not appear on first load


tomgenoni:
### Extension

https://www.raycast.com/eluce2/harvest

### Description

When I first open the extension there are no Tasks listed. When I switch to a different client project and then back to the original one, the Tasks do appear. Tried uninstall/reinstall to no avail and resetting my account on the Harvest side.

This started happening maybe a month ago. Had worked fine until then.

![Screenshot 2024-04-08 at 1 23 09 PM](https://github.com/raycast/extensions/assets/1171072/f145b5d6-c581-4cb1-8b2e-9b13b1b9abd5)

### Steps To Reproduce

I'm not sure how to reproduce it.

### Current Behaviour

Client Tasks do not load when the extension first starts.

### Expected Behaviour

Tasks should load.

eluce2:
Hi @tomgenoni! Thanks for reporting this. I've also noticed this behavior but haven't been sure about its cause. I think I found a fix to improve the caching of those lists which I believe should help

tomgenoni:
Thanks for looking into it. I tried revoking/adding a new access token and archiving a project but am still seeing the issue.

tomgenoni:
#11983 sounds like same issue

PR #11707, opened 2024-04-08
Update harvest extension

eluce2:
## Description

Improve caching for projects and tasks list

## Checklist

### DIFF
--- a/extensions/harvest/src/services/harvest.ts
+++ b/extensions/harvest/src/services/harvest.ts
@@ -92,29 +92,23 @@ export function useActiveClients() {
   });
 }
 
+async function fetchProjects() {
+  let project_assignments: HarvestProjectAssignment[] = [];
+  let page = 1;
+  // eslint-disable-next-line no-constant-condition
+  while (true) {
+    const resp = await harvestAPI<HarvestProjectAssignmentsResponse>({
+      url: "/users/me/project_assignments",
+      params: { page },
+    });
+    project_assignments = project_assignments.concat(resp.data.project_assignments);
+    if (resp.data.total_pages >= resp.data.page) break;
+    page += 1;
+  }
+  return project_assignments;
+}
 export function useMyProjects() {
-  const cachedProjects = JSON.parse(cache.get("projects") ?? "[]") as HarvestProjectAssignment[];
-
-  return useCachedPromise(
-    async () => {
-      let project_assignments: HarvestProjectAssignment[] = [];
-      let page = 1;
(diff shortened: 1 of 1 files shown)


extensions/harvest/src/services/harvest.ts
```diff
@@ -93,6 +93,8 @@ export function useActiveClients() {
 }
 
 export function useMyProjects() {
+  const cachedProjects = JSON.parse(cache.get("projects") ?? "[]") as HarvestProjectAssignment[];
```
eluce2, line 96: This feels like an extra, unnecessary step since I'm already using the "useCachedPromise" function, but it seems to fix the issues... Please correct me if there is a better way that I'm missing

eluce2:
Should fix #11706

extensions/harvest/src/services/harvest.ts
pernielsentikaer: Are you able to make a very simple repro case since I don't have this extension or knows how it works 🙂

eluce2: Yeah I use this extension every day and this fix does work, I just don’t understand why. I feel like I may be using the use cached promise function incorrectly. Or misunderstand it

Let me know if you need a screencast

pernielsentikaer: I'd like to view a simple screencast of the issue so I can discuss it internally :)

eluce2: https://github.com/raycast/extensions/assets/37158449/85a45e5d-b79a-4960-8c91-901e79902d18

mathieudutour: If the cache doesn't work, it means that the hash of the function changes (and so the hooks thinks it's a new function.

Can you try it as 

```js
async function fetchProjects() {
  let project_assignments: HarvestProjectAssignment[] = [];
      let page = 1;
      // eslint-disable-next-line no-constant-condition
      while (true) {
        const resp = await harvestAPI<HarvestProjectAssignmentsResponse>({
          url: "/users/me/project_assignments",
          params: { page },
        });
        project_assignments = project_assignments.concat(resp.data.project_assignments);
        if (resp.data.total_pages >= resp.data.page) break;
        page += 1;
      }
      return project_assignments;
}
export function useMyProjects() {
  return useCachedPromise(
    fetchProjects,
    [],
    { initialData: [] }
  );
}
```

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #11786, opened 2024-04-12
Remove codeowner

stevensd2m:
## Description

Remove codeowner in order not to be called to review PRs.


stevensd2m:
codeowners is generated. checking another way.

PR #11794, opened 2024-04-13
Update gif-search extension

### LINKED ISSUE #11760: [GIF Search] ... Doesn't show any gif's sometimes when searching for stuff
### Extension

https://www.raycast.com/josephschmitt/gif-search

### Description

Hello! Having an issue where when searching for gifs the results don't populate.

EDIT *Extra Tidbits*
This may be because it's trying to find an incomplete sentence but stays in that state until it searches for word again. I have noticed a few times where I would search a complete sentence and even switching providers on search platform it would still hang onto showing no results. Noticed this recently.

### Steps To Reproduce

The way it's happening for me is I would..

1) Open gif search with Raycast
2) Search
3) Delete search and retype but it can be a few characters.
4) After around 2 characters it stops showing results sometimes.

### Current Behaviour

Shows results but then says no results after the edit, would need to delete search and reopen gif search again.

### Expected Behaviour

Should show the same gifs of topic searched.

j3lte:
## Description

This update changes a couple of things, but mainly improve loading of GIFs by:
- Only load more gifs if the list isn't loading (to make sure we don't do multiple requests at once that overwrite the state)
- Implement AbortController logic

This also meant I had to copy part of the Giphy JS library to a local version, because the library doesn't support `AbortController.signal` in their API.

This should also fix #11760 and improve the user experience

## Checklist

### DIFF
--- a/extensions/gif-search/CHANGELOG.md
+++ b/extensions/gif-search/CHANGELOG.md
@@ -1,5 +1,11 @@
 # GIF Search Changelog
 
+## [Use AbortController] - 2024-04-13
+
+- Use `AbortController` to cancel fetch requests when the user searches
+  - This fixes issues where search ends up empty
+  - Part of Giphy search is copied to a local version with added AbortController support
+
 ## [Use native pagination] - 2024-03-12
 
 - Switched to using native pagination for better scrolling experience


j3lte:
Closing PR because of #11887

PR #11843, opened 2024-04-17
Linear: Add pagination to issues search

thomaslombart:
## Description

- Add pagination when searching issues
- Rename "Assigned Issues" to "My Issues"

## Checklist


mil3na (approved):
LGTM!

extensions/linear/package.json
```diff
@@ -74,10 +74,13 @@
       "mode": "view"
     },
     {
-      "name": "assigned-issues",
-      "title": "Assigned Issues",
+      "name": "my-issues",
```
pernielsentikaer, line 77: The only thing to consider when changing this value is that all hotkeys, aliases, and rankings will be removed from the users currently using it. Sometimes it might be better to keep the old name even though it's annoying

thomaslombart: Indeed, it'd be great to find a solution to that because as extensions evolve, so are the commands 😬

Let's see with the team what we can come up with.

PR #11894, opened 2024-04-19
Add yafw extension

pablopunk:
## Description

Are you tired of apps that claim to compress videos but they're just ffmpeg wrappers? Me neither, so I decided to create a Raycast extension that does exactly that

## Screencast

*(no need to say I compressed this screencast with YAFW)*

https://github.com/raycast/extensions/assets/4324982/70738465-c75c-4940-8d89-7abd1049ed83

## Checklist


pernielsentikaer (changes requested):
Hi 👋 

- You could maybe check if ffmpeg is installed by checking the default paths instead (and keep the field optional if people installed it elsewhere)

- You can maybe do something like how [Brew](https://www.raycast.com/nhojb/brew) or [Display Placer](https://www.raycast.com/eluce2/displayplacer) installs packages.


pablopunk:
Thanks for the review @pernielsentikaer! I addressed all the comments 👌🏻

extensions/yafw/src/index.tsx
```diff
@@ -0,0 +1,144 @@
...
+        title: "Select videos first",
+        style: Toast.Style.Failure,
+      });
+
+      return;
+    }
+    compressVideoFiles(files, compression).then((successfulFiles) => {
+      if (successfulFiles.length === 0) return;
+      successfulFiles.forEach((file) => Clipboard.copy({ file }));
+      navigation.push(<Videos files={successfulFiles} />);
```
pernielsentikaer, line 110:
```suggestion
try {
        popToRoot();
        launchCommand({ name: "history", type: LaunchType.UserInitiated });
      } catch {
        /* */
      }
```
Maybe something like this instead

pernielsentikaer (changes requested):
Thanks for the corrections 🔥 

- Do you have a clue why the video I have selected seems to come back on every load (and on compleation)

extensions/yafw/src/index.tsx
pablopunk: what's the advantage here? I'd rather keep the current behavior because of 2 reasons:

* Passing the `successfulFiles` prop allows me to create a decorator for the new files (see screenshot below)
* If you want to go back to compress more videos, you can just press `Esc` or the *back* arrow in the UI



WDYT?

extensions/yafw/src/index.tsx
```diff
@@ -0,0 +1,144 @@
...
+        .catch(() => setSelectedFinderItems([]));
+    };
+    lookForVideoSources();
+    /* Poll for new files every 3 seconds
+     * In case the view is left in the background (i.e Raycast loses focus)
+     * the following warning will be shown at some point:
+     * > Looks like the command is rendering a lot without any changes. This might indicate an rendering loop caused by a `setState` not being wrapped in a `useEffect`.
+     * > This will degrade the performances of Raycast and Raycast might arbitrarily decide to terminate the extension if it happens too many times.
+     */
+    setInterval(lookForVideoSources, 3000);
```
pablopunk, line 63: > Do you have a clue why the video I have selected seems to come back on every load (and on compleation)

this is probably caused by this interval, which I'm still not sure about. I made it so if you open the extension and only then you go to select/copy some videos, it regularly checks if there are more videos in the clipboard or the finder.

I could disable this behavior if you think is not a good idea

pernielsentikaer: I think it's the tradeoff that you'll need to keep it open, much better than using a lot of renders to keep it, maybe we can do something like this "warning"

yafw 2024-04-25 at 08 45 25

extensions/yafw/src/index.tsx
pernielsentikaer: That makes senes, you can maybe add a preference to let the user decide if it should popToRoot, Show compressed files(default) or just history upon completion

pablopunk: I'd prefer to keep it as it is for the first version, but I'll keep it in mind for the future 🙏🏻

extensions/yafw/src/index.tsx
pablopunk: hmmm it's not really during compression, it's right on that selection screen

user opens command -> no videos selected -> user goes to Finder to select or copy some videos -> comes back to the command -> the command auto-refreshes the list with the new files

to be honest I think I'm overthinking a bit here, I think I'll remove the setInterval and add a warning saying

> To add videos automatically, copy them or select them in the Finder and launch this command again

pablopunk: ☝🏻 this should also solve going back from the history and seeing the compressed video auto-selected

pablopunk: image
image

pablopunk:
@pernielsentikaer there's an action to copy because you might convert more than one file at the same time, so I want to give the option to copy any of those single files afterwards.

IMO the usual flow will be compressing just one file, and I'd like to keep the automatic copy-to-clipboard behavior in this case (they'll have clipboard history in raycast anyway).

Regarding the Clear All History, yeah I'm only removing the "old" files (i.e. the ones you are not converting right now). I'll change it to remove all 👍🏻 

The only problem now is that the CLI is not letting me update the PR for some reason 🤔 

image

pernielsentikaer:
Just do as it says, go to your fork (the link) and sync it 🙂

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

j3lte:
@pernielsentikaer is it me or is it not published in the Store? Can't seem to find it...

pablopunk:
@j3lte looks like the job failed https://github.com/raycast/extensions/actions/runs/8884540958/job/24393845273

not sure if i can help with that

pernielsentikaer:
Let me have a look, thanks for flagging @j3lte

pernielsentikaer:
I should be there in two minutes. I accidentally added a style property twice 🙂

PR #11917, opened 2024-04-21
Update power-management extension

bbatsche:
## Description

I added three new commands to the Power Management extension:

- Enable low power mode
- Disable low power mode
- Low power mode details

The first two simply try to set low power mode on or off (so the user doesn’t have to know what the current state is, unlike “toggle”). The third allows a user to see the current low power settings for both AC charger and battery power and then toggle them individually.

There was also some refactoring in the menubar command. In my testing I had noticed some delays in the menubar UI changing in response to low power mode being toggled from elsewhere within Raycast. This should address those.

## Checklist

### DIFF
--- a/extensions/power-management/package.json
+++ b/extensions/power-management/package.json
@@ -41,7 +41,7 @@
         },
         {
             "name": "lowpower-menubar",
-            "title": "Low Power Mode Menu",
+            "title": "Low Power Mode Menu Bar",
             "description": "See status of low power mode in the menubar",
             "mode": "menu-bar",
             "interval": "10m",


extensions/power-management/package.json
```diff
@@ -15,22 +15,40 @@
...
+            "title": "Disable Low Power Mode",
+            "name": "disable-low-power-mode",
+            "description": "Disable low power mode",
             "mode": "no-view"
         },
         {
             "name": "lowpower-menubar",
-            "title": "Menu Bar Low Power Mode",
-            "description": "See status of Low Power Mode toggle on the menubar",
+            "title": "Low Power Mode Menu",
```
erics118, line 44: Probably should still keep the `Bar` there, eg `Low Power Mode Menu Bar`, as that's the seems to be the standard with other extensions

erics118:
The refactor looks good pretty nice!

Perhaps just name the settings command as `Low Power Mode Settings`, as that's what the list nav title says, and makes sense.

It also looks like there's a bug with toggling the enabled/disabled in the settings command. After toggling it once, it seems that you need to re-open the extension before you can toggle it again:

https://github.com/raycast/extensions/assets/52634785/10b2c89b-a8b5-4954-ad1d-4970a9aad530

bbatsche:
@erics118 Good catch on the bug! Looks like it’s related to not having the menu bar command active which I’d neglected to test. I’ll get it fixed up!

pernielsentikaer:
Thanks for the contribution @bbatsche 🔥  Is this ready for another review?

bbatsche:
@pernielsentikaer yes! Thank you, it should be ready for another check.

pernielsentikaer:
@sasivarnan do you want to check it?

sasivarnan:
Apologies for the delay. I have tested the modifications, and everything functions as anticipated.

pernielsentikaer (approved):

PR #11926, opened 2024-04-21
Add text-emoticons extension

Konfuze:
## Checklist


pernielsentikaer:
Maybe it's possible to add the missing emojis in the other one, is that possible @yalishanda42

yalishanda42:
> Maybe it's possible to add the missing emojis in the other one, is that possible @yalishanda42

> Oh, I just discovered that Raycast already has something very similar: https://www.raycast.com/yalishanda/kaomoji-search.
> 
> This extension contains more emoticons with descriptive names, split into categories. It also allows you to paste a random emoticon. However, it's probably not enough to warrant creating a new extension for the store


I agree with you guys. My extension uses `asciilib` but I agree to remove it and bring in this new list of emotes because:
1. It is an external dependency whereas this new list can be more easily maintained and updated.
2. There are more stuff in this list as far as I can see
3. This list is better categorized

I, however, unfortunately do not use a Mac these days so I would be unable to update and test any changes to the Kaomoji extension. I am okay if any of you guys are willing to do that. Thanks @Konfuze for the fresh suggestions

pernielsentikaer:
What do you say @Konfuze, would you be up for that?

Konfuze:
> What do you say @Konfuze, would you be up for that?

Oh, sorry, I missed the notifications while on my vacation.

I see the following caveats if we choose this way:

- The [kaomoji-search](https://github.com/raycast/extensions/tree/82d7aa2bb6436bd3833a530e98d421172fdca96f/extensions/kaomoji-search) extension would need to be completely rewritten
- Probably, some kaomojis will not be present in the updated database
- Users would lose their recent kaomojis
- [yalishanda42](https://github.com/yalishanda42), as a maintainer, wouldn't be able to test this update.

Additionally, as the author, I would like to "own" the extension: choose pictures and texts, and be listed as the author and maintainer of the extension.

pernielsentikaer:
It doesn't make sense to have almost two identical extensions, if this is better it can also superseed the current one

@yalishanda42 what do you think? Should this superseed (move you to contributor) and let @Konfuze be author?

yalishanda42:
> It doesn't make sense to have almost two identical extensions, if this is better it can also superseed the current one
> 
> 
> 
> @yalishanda42 what do you think? Should this superseed (move you to contributor) and let @Konfuze be author?

Oh, I am sorry for the late reply. Yes, I agree to superseed this, since I find the new version/extension as something much needed and I also unfortunately am not able to be a maintainer these days.

pernielsentikaer:
@Konfuze can you handle that? Basicly we need to replace the files in the old folder and use the same name as that one

Konfuze:
@pernielsentikaer, sure, I'll try to make this transition for users as smooth as possible

pernielsentikaer:
If you just "override" the files in the current extensions folder instead and leaving the `name` field for both root and commands the same, then they should just see it as an update

Konfuze:
I'll close this PR to make changes in a new one

PR #11941, opened 2024-04-23
Update obsidian-bookmarks extension

moeyua:
## Description

Added the datePrefix option, allowing users to choose whether to add date as a prefix

## Checklist

### DIFF
--- a/extensions/obsidian-bookmarks/package.json
+++ b/extensions/obsidian-bookmarks/package.json
@@ -156,7 +156,7 @@
       "description": "The prefix to use for the date in the filename of your bookmarks.",
       "type": "checkbox",
       "default": true,
-      "required": true
+      "required": false
     }
   ]
 }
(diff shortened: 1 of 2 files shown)


extensions/obsidian-bookmarks/package.json
```diff
@@ -145,6 +148,15 @@
...
       "required": false
+    },
+    {
+      "name": "datePrefix",
+      "label": "Date Prefix",
+      "title": "Date Prefix",
+      "description": "The prefix to use for the date in the filename of your bookmarks.",
+      "type": "checkbox",
+      "default": true,
+      "required": true
```
pernielsentikaer, line 159:
```suggestion
      "default": false,
      "required": false
```
Would be better to set it like this so we don't break users flow with a new preference

moeyua: I set the default to true because in previous versions it would automatically add the time as a prefix.

And I think setting require to false is good, and I will fix it promptly.

pernielsentikaer: Ah, that makes sense, my eyes didn't catch that 🙂

pernielsentikaer:
Remember to add a changelog entry 🙂

PR #11969, opened 2024-04-24
[Sonos] handle a stale group choice better

AntonNiklasson:
## Description

These changes adresses a problem that appears when the group configuration changes in the network. If the user has previously explicitly picked a group to control, but the group no longer is available, the command would fail.

This is pretty likely to happen where merging and unmerging groups and devices is common. I do that a lot myself, that's how I realised this was an issue.

The new logic is able to detect that the stored group choice is stale, and falls back to the "fallback group choice" that was in place already.


pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #12015, opened 2024-04-26
Update http extension

negasus:
## Checklist

### DIFF
--- a/extensions/http/src/components/FieldForm.tsx
+++ b/extensions/http/src/components/FieldForm.tsx
@@ -1,6 +1,6 @@
-import {useCallback, useState} from "react";
+import { useCallback, useState } from "react";
 import { Form, Action, ActionPanel, useNavigation } from "@raycast/api";
-import {validateNotEmptyString} from "../validation";
+import { validateNotEmptyString } from "../validation";
 
 function FieldForm(props: { name: string; index: number; onCreate: (name: string) => void }) {
   const { name, index, onCreate } = props;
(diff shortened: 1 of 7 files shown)


pernielsentikaer:
Thanks for your contribution. Have you seen [this extension](https://www.raycast.com/jonathangiardino/curl), which appears similar yet possesses a lot of features? Maybe enhancing that one would be preferable, so we avoid having two similar extensions.

negasus:
This extension has slightly different goals. Calling predefined requests with substitution of variables from the profile. I also plan to add the ability to save queries in collections.

But if you want to reject this PR, I understand you.

pernielsentikaer:
I agree it's not 100% the same, but it would be something that would fit right into the other extension, what do you think @jonathangiardino - wouldn't it make sense to enhance instead of adding a new one?

jonathangiardino:
Haven't sen what the other extension does to be honest, but if there's something that can be added to cURL and can be an enhancement for all the users, I am all for it! @pernielsentikaer

pernielsentikaer:
I think it would fit right in, what I understand is that it will be possible to make profiles and have variables - but I'm not sure how easy it will be to merge, what do you think @negasus

negasus:
I'm afraid that this will require quite a lot of modification of the cURL extension. I think it would be better if the author decides for himself whether this functionality should be added.

I suppose, this PR can be closed?

jonathangiardino:
> I'm afraid that this will require quite a lot of modification of the cURL extension. I think it would be better if the author decides for himself whether this functionality should be added.
> 
> I suppose, this PR can be closed?

@negasus unfortunately I do not have time for this implementation at the moment, but since it's all open source, feel free to add features if you want and open a PR, @pernielsentikaer also contributed to it before.

pernielsentikaer:
Closing for now, let's get some of the features implemented in the current extension when there is time for it

PR #12056, opened 2024-04-28
Update Netlify extension (add Manage DNS Records)

xmok:
## Description

I have multiple domains in Netlify and I often need to manage the DNS records which is why this PR exists.

For registered domains we can create, delete and view DNS records. I have tried to match the Netlify experience as best as possible so all descriptions, placeholders and labels are copied verbatim from Netlify "Add new record" modal. Since fields can be different I have conditionally validated and toggled visibility of form items. To improve UX, Toasts are shown during and after actions. However, I have not added Toasts to any existing functions.

On a Dev Note,
 - update `api` and `utils`
 - remove `getPreferences` from helpers 
 - remove `Preferences` from interfaces.ts since Preferences is automatically exposed
 - move some variables into a new `constants.ts`

## Screencast

https://github.com/raycast/extensions/assets/30526133/9261a28e-a6d8-4c05-8dc7-955ad7b38ead

## Checklist


xmok:
It seems that Netlify  does not check for the CAA tag to be of the correct type so I have removed that check as it is redundant (plus it would have been better as a Dropdown). I have also moved the new components into a separate `dns` folder to make the components more manageable.

jasonbarry (approved):
LGTM, thanks for bringing this up to speed @xmok!

extensions/netlify/package.json
```diff
@@ -43,10 +44,10 @@
       "mode": "view"
     },
     {
-      "name": "search-domains",
-      "title": "Search Domains",
+      "name": "search-domains-and-manage-dns-records",
```
pernielsentikaer, line 47: Can we keep the name since hotkeys, aliases and ranking are using the name as ID 🙂

xmok: Hi.
I'm not really sure what you mean but yeah we can revert the name change if required. Lemme quickly make the change.

pernielsentikaer: So the name is the ID for the command, if you change that then the users will lose the ranking / alias / hotkey attached 🙂

Thanks for changing back 🙂

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #12124, opened 2024-05-02
Add ipa-translator extension

BossElijah:
## Description

A simple tool to convert text to IPA (International Phonetic Alphabet).
It currently supports English, German, Danish, Swedish, and Czech.

There is a command per language.

## Screencast

image

## Checklist


pernielsentikaer (changes requested):
The extension crashes for me due to heap size, instead of require them you might be able to use [useStreamJSON](https://developers.raycast.com/utilities/react-hooks/usestreamjson) instead 🙂

image

BossElijah:
Hi @pernielsentikaer,
Thanks for noticing! For some reason I changed my memory limit locally in the past, so that's why I didn't get the error when developing, but when I reset it, the error immediately occurred.

I have fixed it using `readFileSync` from 'fs'.

pernielsentikaer (approved):
Just out of curiosity; how would I use this? It doesn't look like this one: https://www.raycast.com/jns/nato-phonetic-alphabet but yet sounds a bit similar?

BossElijah:
There are two types of phonetic alphabets. One was created to be able to effectively communicate in radios with bad/scratching signal. The first letter in each word is the letter the word symbolises. I don't know about other countries, but Denmark's army created one that they use especially in the marine when there are loud noises like waves etc. The other extension has what seems to be NATO's alphabet.
Example with NATO's alphabet: Per --> Papa Echo Romeo

However the IPA that this extension uses is used tell how a specific word is pronounced, so it's basically an internationally accepted pronunciation guide. Look up any word on google and it will show you the IPA so even people who don't speak the language know how to pronounce it properly.
image

pernielsentikaer:
So it's meant to look like this with above word?

ipa-translator 2024-05-08 at 14 20 29

BossElijah:
Yes, also there can be multiple IPA definitions depending on accents etc

PR #12155, opened 2024-05-03
Update perplexity extension

mtrefilek:
## Checklist

### DIFF
--- a/extensions/perplexity/CHANGELOG.md
+++ b/extensions/perplexity/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Perplexity Changelog
 
+## [Pro Mode Default] - 2024-05-03
+
+Now you can set the Ask Perplexity Command to Use Pro Mode by Default
+
 ## [Fallback Improvement] - 2023-11-29
 
 Now when using Ask Perplexity as a fallback command, the form will submit automatically. It will still use the stored value of the "Copilot" checkbox.


third774:
Hey there! Thanks for the PR! We're already storing the value (link below) — I'm curious to know more about the reasoning for this change.
https://github.com/raycast/extensions/blob/main/extensions/perplexity/src/ask-perplexity.tsx#L75

We do already have a preference for the fallback command here:
https://github.com/raycast/extensions/blob/main/extensions/perplexity/package.json#L38

If there's a good reason for moving this into the settings perhaps we should just consolidate onto a single "Use Pro Mode" setting/preference?

pernielsentikaer:
@mtrefilek could you check the response from @third774

mtrefilek:
> Hey there! Thanks for the PR! We're already storing the value (link below) — I'm curious to know more about the reasoning for this change. https://github.com/raycast/extensions/blob/main/extensions/perplexity/src/ask-perplexity.tsx#L75
> 
> We do already have a preference for the fallback command here: https://github.com/raycast/extensions/blob/main/extensions/perplexity/package.json#L38
> 
> If there's a good reason for moving this into the settings perhaps we should just consolidate onto a single "Use Pro Mode" setting/preference?

Hi, sorry for my delay! I think we can close, I didn't fully understand how fallback worked, so thought we should add a default to the action selection. I think keeping the default for fallback ,but requiring a "y" for manual action use makes sense.

Issue #12106, opened 2024-04-30
[1Password] Error: 'The "file" argument must be of type string. Received undefined'


joshuahhh:
### Extension

https://www.raycast.com/khasbilegt/1password

### Description

When I use the "My Passwords" action, the search window loads but no searches yield results. An error appears in the bottom-left: 'The "file" argument must be of type string. Received undefined'. (See screenshot.)

<img width="476" alt="Screenshot 2024-04-30 at 4 02 07 PM" src="https://github.com/raycast/extensions/assets/643799/d9b68de0-dc88-48ed-abf9-f39a6580622c">

### Steps To Reproduce

See above. I don't imagine everyone else is having this same problem, but I'm not sure what to dig into to diagnose it. Thanks!

Luk4h:
Hello, please ensure that you have the 1Password CLI installed, and its path is correctly set on the extension settings (check using 'where op' on terminal).

joshuahhh:
Ah! I got that stuff set up, and then I had an "Unexpected end of JSON" error (with no passwords) until I connected 1Password CLI to the desktop app. It's working now, except I still get a (seemingly innocuous) "Unexpected end of JSON" error, as mentioned at https://github.com/raycast/extensions/issues/11826.

I might be mistaken, but I don't think I saw any message about installing the 1Password CLI during my path to using this extension – not on the store page, or in the extension itself, or in the error message. I bet it would be helpful to message this more, perhaps most usefully at the site of the error. Sorry if you're working on this and I'm telling you something you already know! Just my two cents.

Thanks a bunch for your help.

Luk4h:
As for the latest update, there is a better error handler explaining what the error was, which will help users report the error and fix it if it is a user-related issue.

@raycastbot close this issue

Issue #12102, opened 2024-04-30
[1Password] Support for two logins (business and family account)


ashkrosh:
### Extension

https://www.raycast.com/khasbilegt/1password

### Description

Currently if I login using my business account I have access to my business vaults but not my family account vaults. 

Is it possible to be logged in both? 

PS: all business plan users have access to free family accounts on 1Password.

Luk4h:
With the latest update, the extension is able to use multiple vaults, filter items per vault, and switch accounts in extension. Right now there are no support for retrieving the passwords for multiple accounts at once. I will be leaving this issue open if someone is interested in trying the hard way to cycle between the user accounts and retrieving the passwords.

cmpaul:
+1 to original issue. I just installed Raycast and this extension today. 

When I am signed into multiple 1Password accounts and open "My Vaults" via the extension, I see this error:
<img width="344" alt="image" src="https://github.com/raycast/extensions/assets/380070/de139274-df3d-489e-a8c6-e809ccb9737e">

2start:
Would be great to have this feature. Same use case businesss + private

Issue #12095, opened 2024-04-30
[1Password] "Allow Raycast to get CLI Access" prompts constantly


scottschechner:
### Extension

https://www.raycast.com/khasbilegt/1password

### Description

Raycast is asking for CLI access 10+ times per day for the 1Password extension. I don’t believe I’ve changed any extension settings.

### Steps To Reproduce

1. Install 1Password extension for Raycast
2. Wait until the CLI prompt appears
3. Give 1Password access
4. Watch the prompt reappear throughout the day

Luk4h:
Hello Scott, how are you? 1Password has a 10 minute cooldown for each unlock. There is a command that is disabled by default that extends this cooldown up to 12 hours. But you may get prompts from nowhere after waking the device on. You may disable the command to auto renew if you prefer to only use the prompts when trying to access a password. Right now it is working as intended, but I will include an improvement to its functionality to avoid these random prompts.

j-peeters:
> You may disable the command to auto renew if you prefer to only use the prompts when trying to access a password.
That would be a great addition for me.

romanr:
Happens about once every hour for me, while I never use this extension, or when I didn't use Raycast in hours.. 

I am not sure what is usefulness of extension over 1P built-in functionality, you can enable [1P "Quick access" ](https://support.1password.com/quick-access/) global shortcut `⇧`+`⌘`+`Space`, which pops its own up Raycast-like interface to copy your credentials.

Luk4h:
On the settings page, on 1Password commands, the auto renew command has the background refresh enabled? By default it should be disabled.

j-peeters:
It was 'on' for me, thanks for this, I deactivated it. I didn't change this manually so maybe this is set by the extension to 'on' by default. Might be an idea to set it to 'off'?

scottschechner:
It was on by default for me as well.

Would there be a way to prompt [1P "Quick access" ](https://support.1password.com/quick-access/)global shortcut ⇧+⌘+Space within Raycast itself using an alias?

Luk4h:
It should be off by default, I will add a fix for it on the next update. Sorry about it guys.

j-peeters:
No problem @Luk4h Thanks for clarifying and fixing!

Luk4h:
> Would there be a way to prompt [1P "Quick access" ](https://support.1password.com/quick-access/)global shortcut ⇧+⌘+Space within Raycast itself using an alias?

[...]

Issue #11826, opened 2024-04-16
Unexpected end of json input


m-j-45:
### Extension

https://www.raycast.com/khasbilegt/1password

### Description

Every time I use the extension, even though it is working fine, I get the error message "Unexpected end of json input"

### Steps To Reproduce

I just use the extension as normal.

Luk4h:
Hello @m-j-45, I am currently working on improving error handling and explanations for these errors. I will notify you as soon as possible when it is available. 🫡

m-j-45:
OK thank you, and thanks for replying so quickly. From: Luiz Gustavo Marcondes ***@***.***>Date: Tuesday, 16 April 2024 at 12:42 PMTo: raycast/extensions ***@***.***>Cc: m-j-45 ***@***.***>, Mention ***@***.***>Subject: Re: [raycast/extensions] Unexpected end of json input (Issue #11826)Hello @m-j-45, I am currently working on improving error handling and explanations for these errors. I will notify you as soon as possible when it is available. 🫡—Reply to this email directly, view it on GitHub, or unsubscribe.You are receiving this because you were mentioned.Message ID: ***@***.***>

Luk4h:
As for the latest update, there is now better error handling which should prevent these "Unexpected end of JSON input" errors and facilitate user reports and fixes.

@raycastbot close this issue

PR #12156, opened 2024-05-03
Update 1password extension

luizvergennes:
## Description

🎉 My Vaults Command
Lists user vaults, and lists items from the vault.
🩺 Solves #4744 

🎉 Account switcher
Users are now able to switch accounts.
🩺 Solves #12102 #4744 

🎉 Better Error handling
Now errors are treated better to understand user issues.
There is a smart component that indicates the user what the problem was and tips to fix it.
🩺 Solves #12106 #11826 

🐛 Fixed the duplicated auth prompt
Fixed the issue when trying to access passwords to require two authentications after the 10 minute cooldown.
🩺 Solves #12095 

🔧 Modified the auto renew command to be disabled by default
It now requires the user to enable it in the extension config.
🩺 Solves #12095 

🔧 Improved the auto renew command to renewal only after a manual sign-in.
🩺 Solves #12095 

## Checklist

### DIFF
--- a/extensions/1password/src/v8/components/Categories.tsx
+++ b/extensions/1password/src/v8/components/Categories.tsx
@@ -1,6 +1,6 @@
 import { Icon, List } from "@raycast/api";
 
-import { Guide } from "./Guide";
+import { Error as ErrorGuide } from "./Error";
 import { CategoryName } from "../types";
 import { getCategoryIcon, useCategories } from "../utils";
 
@@ -9,7 +9,7 @@ export const DEFAULT_CATEGORY = "null";
 export function Categories({ onCategoryChange }: { onCategoryChange: (newCategory: string) => void }) {
   const { data, error, isLoading } = useCategories();
 
-  if (error) return <Guide />;
+  if (error) return <ErrorGuide />;
   return (
     <List.Dropdown
       defaultValue={DEFAULT_CATEGORY}
--- a/extensions/1password/src/v8/components/Items.tsx
+++ b/extensions/1password/src/v8/components/Items.tsx
@@ -12,7 +12,7 @@ import {
   ConnectionError,
   ExtensionError,
 } from "../utils";
-import { Guide } from "./Guide";
+import { Error as ErrorGuide } from "./Error";
 import { ItemActionPanel } from "./ItemActionPanel";
 import { useMemo, useState } from "react";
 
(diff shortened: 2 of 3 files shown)


pernielsentikaer:
Thanks for the update @Luk4h 🔥 

If I install it for the first time, then I see this error and have to use my biometric twice? Do you have an idea about that?

Raycast06052024-LdwjZxHS png@2x

Luk4h:
It has to do with the 1Password CLI auth method. Since the views always execute ```op account``` and ```op item list```. It is trying to execute both commands at the same time, that's why the first time asks twice for the auth.
Currently this one is the last issue open related to 1password. I am currently working for a fix. The issue is somewhat resolved when you have auto-renew authorization on.

Luk4h:
> I am currently working for a fix. 

Found that using a context for the auth may solve the issue. Please stand by while I fix this.

Luk4h:
@pernielsentikaer trying to find anything about not closing the raycast window if the 1password prompt opens, is it possible?

Luk4h:
@pernielsentikaer I made a way so that just before requesting 1Password authentication to pop to the root of Raycast. After authentication use, you will need to search for the command again. This is not ideal, but sufficient for now.

I believe there is a bug with Raycast or React when interacting with the node exec function that causes the 1Password prompt to break it and never return the function as finished. This makes it hard to work for a more seamless workflow for getting the passwords.

pernielsentikaer:
An extension unloads when it is no longer loaded, for example, when popping to the root. Even if you do not pop to the root and Raycast loses focus, there is a high likelihood it will be unloaded as well.

Maybe it's possible to run something like `await open("raycast://");`

Luk4h:
https://github.com/raycast/extensions/assets/78048969/27da0198-fc84-4202-a5f1-a1d20ba5322b

> Maybe it's possible to run something like await open("raycast://");

Even with that, it does not matter. As you can see in the video. If the prompt is successful, it does not continue the code execution. 

Sadly when I was using the node runtime directly to test it worked just fine...  That's why my guess is something related to React or the Raycast itself not retrieving the return of this command. 

Maybe is it because the stdout of this command is null?

Luk4h:
We are ready to go @pernielsentikaer! 🎉

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

kjeldahl:
@Luk4h I get this error: 
image
I did not get a chance to select any of the two accounts I am using. Only the 1Password version. I tried to reinstall the extension, but it still fails with the error above. On all commands for 1Password.

Luk4h:
> @Luk4h I get this error: image I did not get a chance to select any of the two accounts I am using. Only the 1Password version. I tried to reinstall the extension, but it still fails with the error above. On all commands for 1Password.

Got the fix already, I will make the PR.

Luk4h:
Refer to #12491

PR #12169, opened 2024-05-04
Add brave-search-with-results extension

electerious:
## Description

Extension that lets you search the web using the Brave API. Includes a history, suggestions and results directly in Raycast.

- This extension is different to the existing Brave extension as it heavily relies on the Brave API and will get more features powered by the API.
- AFAIK the first extension that lets you search the internet without leaving Raycast.

## Screencast

https://twitter.com/electerious/status/1783869547730145520

## Checklist


electerious:
I'm getting an error when using the extension after running `npm run build`, but I'm not sure why. The StackTrace makes no sense, there's no `throw` in the shown code and [AboutErrors are already catched](https://github.com/raycast/extensions/blob/fb8b30689fa5de6f34ec4b27b4baa2b2fa29a388/extensions/brave-search-with-results/src/hooks/useFetch.ts#L36).

Screenshot 2024-05-04 at 16 00 16

pernielsentikaer:
Which extension are you referring to, [this one](https://www.raycast.com/VishwasShashidhar/brave-search)?

brave-search 2024-05-10 at 10 55 20

It seems pretty similar to yours, or?

electerious:
> Which extension are you referring to, [this one](https://www.raycast.com/VishwasShashidhar/brave-search)?
> 
> brave-search 2024-05-10 at 10 55 20
> 
> It seems pretty similar to yours, or?

Yep. But both are completely different in how they're interacting with Brave.

Brave Search: History + Suggestions. No API key needed. Similar to most search extensions.

Mine: Uses the Brave API that requires two free (or paid) API keys. This allows me to not only show a history and suggestions, but also the results itself, without opening the browser. The API also enables to search for Images, News etc. (something that's currently not a part of the extension, but already planned).

Merging the extensions would mean that you cannot use the existing extension without API keys, which makes the setup more challenging. Alternatively, there would need to be two separate code bases in one extension - one for when you skip the API key setup and another for when you enter the keys.

I therefore think it's better to have an extra extension (similar to self-hosted JIRA vs. hosted JIRA).

electerious:
Any news? :(

pernielsentikaer (approved):
Hi @electerious 👋

Sorry for the long response, notifications has been active up weird for some time now, sorry about that 🙁

Let's get the first version live

PR #12221, opened 2024-05-07
Support macOS Big Sur 11

vadim-a-yegorov:
## Checklist


pernielsentikaer:
The problem is that as soon as a new thing is introduced, it will be updated again, which could happen anytime. I don't think it makes sense to do this, also for the reasons stated by @LitoMore in #12222

maximilianzuern:
Thank you, @pernielsentikaer! 
Therefore, I would suggest closing the PR.

PR #12222, opened 2024-05-07
Support macOS Big Sur 11

vadim-a-yegorov:
## Checklist

### DIFF
--- a/extensions/perplexity/package.json
+++ b/extensions/perplexity/package.json
@@ -42,7 +42,7 @@
     }
   ],
   "dependencies": {
-    "@raycast/api": "^1.60.1",
+    "@raycast/api": "^1.50.5",
     "@raycast/utils": "^1.10.1"
   },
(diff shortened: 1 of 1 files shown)


LitoMore:
Logically speaking, the Raycast does not support macOS 12. See the download page: https://www.raycast.com.

Downgrading the version of `@raycast/api` is not a good solution. It will only reduce the extension's stability and prevent the possibility of using the new API.

On the other hand, this change won't take effect. Because the CI uses a specific version of `ray-cli` from this [GitHub Actions configuration file](https://github.com/raycast/github-actions/blob/d3a43c46b64ac36b751a03ffce82e8517fc0228a/ray/action.yml#L43).

pernielsentikaer:
The problem is that as soon as a new thing is introduced, it will be updated again, which could happen anytime. I don't think it makes sense to do this, also for the reasons stated by @LitoMore

PR #12255, opened 2024-05-09
Spotify: Don't close Raycast when changing the volume

krsntn:
## Description

Previously, whenever the volume up and down commands were executed, the Raycast window would close, necessitating repeated actions. However, this enhancement ensures that the Raycast window stays open after executing the command, greatly improving usability.

## Checklist

### DIFF
--- a/extensions/spotify-player/CHANGELOG.md
+++ b/extensions/spotify-player/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Spotify Player Changelog
 
+## [Improvement] - 2024-05-09
+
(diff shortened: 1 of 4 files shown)


thomaslombart (commented):
Thanks for the PR! I'm not sure this needs to be changed as this could impact others experiences. Some people may want to turn down the volume a bit and be done with Raycast. With this change, they would have to close it on top of that. I'm curious to have others opinion on that.

andreaselia:
Maybe this could keep the current behaviour by default, and add a command config for determining if Raycast should close afterwards?

PR #12268, opened 2024-05-09
Add ethereum-eips extension

banteg:
## Description

Look up Ethereum Proposals (both [ERCs](https://github.com/ethereum/ERCs) and [EIPs](https://github.com/ethereum/EIPs)) and read them.

to set it up, the user would need to clone both repos into a folder and specify it in the extension settings.

```sh
mkdir ethereum
cd ethereum
gh repo clone ethereum/EIPs
gh repo clone ethereum/ERCs
```

this extension in a separate repo here https://github.com/banteg/raycast-eips

## Screencast

https://github.com/raycast/extensions/assets/4562643/e60124f1-2189-4394-9884-7e7f7a575c14

## Checklist


banteg:
i've added a bunch of changes to the original repo, can i update it before it's added to the store?

pernielsentikaer:
Is there a way to fetch the data from an API instead? Many of our users might not be that technical to handle this setup to get started 🙂

banteg:
in the new version it's handled automatically on the first run, the user just specifies a working directory in the extension settings.

pernielsentikaer:
> in the new version it's handled automatically on the first run, the user just specifies a working directory in the extension settings.

But you still need to pull the data manually before using the extension, and I guess at some point you need to manually update the repositories to get new data, right?

johnletey:
@pernielsentikaer @banteg Jumping in here, as I'd personally really love to see the data be fetched automatically ...

The extension could do the following:
- Fetch the current commit hash of https://github.com/ethereum/EIPs.
  - `https://api.github.com/repos/ethereum/EIPs/commits`
- Use that commit hash to fetch current file tree.
  - `https://api.github.com/repos/ethereum/EIPs/git/trees/396253af618f50958b4b6a851fd79fea3bc2e784`
- Use that response to fetch all EIP files.
  - `https://api.github.com/repos/ethereum/EIPs/git/trees/21d8d81cc03e5bef312e7094aa64299e001009c3`

This request can be run and cached so that it isn't spamming the GitHub API.
I suspect that we will need to configure the extension with a GitHub API token, but there are worse things 😄

pernielsentikaer:
There is other extensions using oAuth, maybe we can borrow that part? Like our own [GitHub extension](https://www.raycast.com/raycast/github), would that be possible?

johnletey:
Yeah we could use some code from the other extensions, absolutely!

PR #12273, opened 2024-05-10
Update sonos extension

krsntn:
## Description

Adding the display of the volume change after executing increase and decrease volume commands.

## Checklist

### DIFF
--- a/extensions/sonos/CHANGELOG.md
+++ b/extensions/sonos/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Sonos Changelog
 
-## [ Improvement] - 2024-05-10
+## [Improvement] - 2024-05-10
 
 - Display the volume change after executing increase and decrease volume commands
 


extensions/sonos/src/decrease-volume.tsx
```diff
@@ -30,5 +30,11 @@ export default async function Command() {
     });
   } else {
     await coordinator.SetRelativeGroupVolume(-Number(preferences.volumeStep));
+    await coordinator.GetState().then((state) =>
```
AntonNiklasson, line 33: If I remember correctly, the new volume is actually returned from `SetRelativeGroupVolume`. Couls you try that?

AntonNiklasson (changes requested):

pernielsentikaer:
Thanks for the enhancement @krsntn - could you check the review from @AntonNiklasson

extensions/sonos/src/decrease-volume.tsx
krsntn: you're right. it does return the new volume.

extensions/sonos/src/decrease-volume.tsx
```diff
@@ -29,6 +29,11 @@ export default async function Command() {
       failureMessage: `Failed to launch "Set Active Group" automatically`,
     });
   } else {
-    await coordinator.SetRelativeGroupVolume(-Number(preferences.volumeStep));
+    await coordinator.SetRelativeGroupVolume(-Number(preferences.volumeStep)).then((volume) =>
```
AntonNiklasson, line 32: Sorry to be nitpicking, bit I prefer not mixing async/await and then like this. It would be nice to stick with just the former ☺️

krsntn: sure. 🫡

pernielsentikaer:
Are we good to go @AntonNiklasson 🙂

AntonNiklasson (approved):
All good! Thank you for contributing 🙌

PR #12280, opened 2024-05-10
Update tennis-standings extension

drillprop:
## Description

Follow up of https://github.com/raycast/extensions/pull/12248

## Screencast

https://github.com/raycast/extensions/assets/51168865/9d89ab14-4d56-490d-86ae-9302aed3d51e

## Checklist


xilopaint:
I still prefer the leaderboard icon over repeating `Points:`. The rest looks good to me.

xilopaint:
I still prefer the leaderboard icon over repeating `Points:`. Also, I believe it's more common to use `ATP Rankings` (in the plural form). The rest looks good to me.

xilopaint:
I still prefer the leaderboard icon over repeating Points:. Also, I believe it's more common to use `Rankings` (in the plural form), in this context, than the singular form `Ranking`. The rest looks good to me.

drillprop:
Thanks for the feedback. 
> Also, I believe it's more common to use Rankings (in the plural form), in this context, than the singular form Ranking.

Sure, I'll fix that.

> I still prefer the leaderboard icon over repeating `Points:`. 

I find the leaderboard icon a bit unclear in terms of its meaning. What about stars? 👇🏻 
Zrzut ekranu 2024-05-10 o 20 36 20

VS

Zrzut ekranu 2024-05-10 o 20 40 13

xilopaint:
I don't see how the star can be clearer than the leaderboard icon. The tennis standings are essentially leaderboard positions. An alternative is using the `pts` abbreviation after the points (e.g. `10760 pts`), but I didn't test to see how it looks.

There's also redundancy in the dropdown. You don't need the "Filter by country" section if you're already using the same message as a placeholder:

Screenshot 2024-05-10 at 16 09 19

xilopaint:
Thinking a bit more, I would also recommend against the emojis in the rankings. It's impossible to know all the flags. Look the current WTA rankings:

Screenshot 2024-05-10 at 16 19 03

With all due respect, I have no clue what the flags of Belarus and Kazakhstan look like, even though those are two top-10 players. Your emojis are not helpful for smaller countries. I much prefer the previous way of presenting the countries. Furthermore, emojis are known to impoverish the design and make it inconsistent.

drillprop:
@xilopaint 
> Also, I believe it's more common to use Rankings (in the plural form), in this context, than the singular form Ranking.

**fixed** 

> There's also redundancy in the dropdown. You don't need the "Filter by country" section if you're already using the same message as a placeholder:

**fixed**

> I don't see how the star can be clearer than the leaderboard icon. The tennis standings are essentially leaderboard positions. An alternative is using the pts abbreviation after the points (e.g. 10760 pts), but I didn't test to see how it looks.

I'm still not convinced by this. I find the icon blurry - it looks to me like some buildings and a star on top of it. It doesn't look like a podium to me. Also, the numbers on the right of the icon don't tell about the player's position but about the player's points, so the leaderboard semantically doesn't fit here IMO. 
But anyway, I won't be arguing here - I set the leaderboard icon like you want. **Fixed**

Regarding emojis and countries:

> Furthermore, emojis are known to impoverish the design and make it inconsistent.

Sure, I understand your point and partly agree. But If I would set the name of the country close to the name(using subtitle prop) as you suggest, it would be more actually confusing.  I use the subtitle to show ranking changes since the last ranking release, and If I would add a country close to it, it wouldn't be really incomprehensible. This is how it would look 👇🏻 

image


Removing ranking change is not an option - IMO this information is more important in the leaderboard extension than showing the name country. You want to know how players progress through the season. 
 
> With all due respect, I have no clue what the flags of Belarus and Kazakhstan look like, even though those are two top-10 players.

You can always check the player's country name by checking the player's details.  So you can access this information very easily, actually. 
image

Also, to convince you a little bit more, there is a very similar extension already that uses the same pattern with emojis. 

image

So in the end I left emojis and countries as it is.

After requested changes here's how the extension looks:
image

image

xilopaint:
> I find the icon blurry - it looks to me like some buildings and a star on top of it. It doesn't look like a podium to me. Also, the numbers on the right of the icon don't tell about the player's position but about the player's points, so the leaderboard semantically doesn't fit here IMO.

A podium is basically what you get when you search for "ranking" in Iconify:

Screenshot 2024-05-12 at 10 36 35

Yet, if you think `Icon.Leaderboard` can be improved, you can always talk to the Raycast design team. New icons can also be requested. Or do you think it would be a better option not to use any icon at all, as in the Formula 1 extension you showed?

> But If I would set the name of the country close to the name(using subtitle prop) as you suggest, it would be more actually confusing. I use the subtitle to show ranking changes since the last ranking release, and If I would add a country close to it, it wouldn't be really incomprehensible.

I see your point. To be honest, I can't find a good solution to accommodate everything. The emojis make the design look cheaper, no wonder you won't find them in Raycast built-in extensions. But in the absence of a better alternative, let's leave them in the list view as you suggested. I only think it's too much using the flag ahead of the country name. In the dropdown I would remove the emojis and prove with `Icon.Globe` for "All countries" and `Icon.Map` for each country.

I would also remove the emoji in the player's details because it looks ugly and unnecessary.

pernielsentikaer:
We can also move flags to accessories, I'm not sure if it makes it less cluttered? We can also make a preference to not show the flag if you don't like that @xilopaint 

tennis-standings 2024-05-15 at 05 52 29

xilopaint:
> We can also make a preference to not show the flag if you don't like that @xilopaint

Not necessary. I'm not the extension author, just a contributor, and this PR brings good improvements.

I would remove the flags from the dropdown and the player's details, but it's just my opinion. @drillprop can have the final say here.

pernielsentikaer:
I actually like the flags, for me it brings value to have it

@drillprop is this all ready to be merged?

drillprop:
> I actually like the flags, for me it brings value to have it
> 
> @drillprop is this all ready to be merged?

I was busy for the last few days and I couldn't answer.  I would also keep the flags, especially in the dropdown. Personally with all due respect, I don't like the idea of setting repetitive Map icons.

Yes, it's ready to be merged.

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #12358, opened 2024-05-15
Update empty-screenshots extension

robotdestroy:
## Checklist

### DIFF
--- a/extensions/empty-screenshots/package.json
+++ b/extensions/empty-screenshots/package.json
@@ -25,7 +25,7 @@
           "description": "Choose how you want to clean up your screenshots.",
           "type": "dropdown",
           "required": true,
-          "title": "Delete or move to trash",
+          "title": "Delete or Move to Trash",
           "default": "0",
           "data": [
             {
@@ -41,7 +41,7 @@
         {
           "type": "directory",
           "name": "screenshotFolder",
-          "title": "Screenshot folder",
+          "title": "Screenshot Folder",
           "required": true,
           "description": "Location where you save your screenshots.",
           "default": "~/Desktop"
@@ -51,7 +51,7 @@
           "description": "Prevent screenshots from being deleted if they are older than the configured days.",
           "type": "dropdown",
           "required": true,
-          "title": "Protect recent screenshots for",
+          "title": "Protect Recent Screenshots for",
           "default": "30",
           "data": [
             {


robotdestroy:
- Added preference for the user to choose to 'delete' or 'move to trash' behavior
- Added directory UI affordance for the user to choose the screenshot folder
- Added preference for the user to choose the screenshot custom prefix
- Added additional option to not protect recent screenshots

robotdestroy:
New UI
CleanShot 2024-05-14 at 21 34 41@2x

extensions/empty-screenshots/src/index.ts
```diff
@@ -2,22 +2,50 @@ import { readdir, stat, rm } from "fs/promises";
...
+        let startsWithValidPrefix = false;
+
+        if (prefixString) {
+          startsWithValidPrefix = file.startsWith(prefixString);
+        } else {
+          startsWithValidPrefix = file.startsWith("CleanShot") || file.startsWith("Screen");
+        }
+
+        return endsWithValidExtension && startsWithValidPrefix;
+      })
```
chanceaclark, line 48: Could we split these into two separate `.filter` functions?
```suggestion
      .filter((file) => file.endsWith(".png") || file.endsWith(".mov"))
      .filter((file) => prefixString ? file.startsWith(prefixString) : file.startsWith("CleanShot") || file.startsWith("Screen"))
```

extensions/empty-screenshots/package.json
```diff
@@ -50,12 +75,21 @@
...
             }
           ]
+        },
+        {
+          "name": "prefixString",
+          "type": "textfield",
+          "title": "Custom prefix (optional)",
+          "required": false,
+          "description": "You can optionally add a custom prefix for the screenshot filenames.",
+          "placeholder": "Screenshot, CleanShot, etc."
```
chanceaclark, line 85: 🍹 What do you think about this renaming?
```suggestion
          "name": "filePrefix",
          "type": "textfield",
          "title": "File prefix (optional)",
          "required": false,
          "description": "You can optionally add a custom prefix for the screenshot filenames.",
          "placeholder": "Screenshot, CleanShot, etc."
```

extensions/empty-screenshots/package.json
```diff
@@ -50,12 +75,21 @@
...
         }
       ]
     }
   ],
   "dependencies": {
-    "@raycast/api": "^1.38.3"
+    "@raycast/api": "^1.73.2",
+    "trash-cli": "^5.0.0"
```
chanceaclark, line 92: Instead of using the CLI version, could we use the underlying `node` module instead ([`trash`](https://github.com/sindresorhus/trash)). I think we'd want to avoid spinning up another child process for a simple async command.

extensions/empty-screenshots/src/index.ts
```diff
@@ -34,7 +62,11 @@ export default async function main() {
       const stats = await stat(screenshot);
 
       if (stats.birthtimeMs <= priorDate.getTime()) {
-        await rm(screenshot);
+        if (screenshotBehavior === "0") {
+          await rm(screenshot);
+        } else {
+          await moveToTrash(screenshot);
+        }
```
chanceaclark, line 69: 🍹 In case we might want to add another behavior in the future, could we augment this a little bit to provide more extensibility?
```suggestion
        switch(screenshotBehavior) {
          case '1':
            return await moveToTrash(screenshot);
          case '0':
          default:
            return await rm(screenshot);          
        }
```

extensions/empty-screenshots/src/index.ts
chanceaclark: Additionally, with the default behavior of this filter (`file.startsWith("CleanShot") || file.startsWith("Screen")`), should we assume all screenshots start with `CleanShot` or `Screen`? This might break existing installations of the extension if we assume that to be true.

I think since we are targeting an additive behavior here, we should probably default to the existing behavior of not caring about the prefix, and with configuration value take a comma separated list of prefix values. This way the config could be `CleanShot, Screen` and we could do `prefixString.replace(' ', '').split(',')` and filter by that instead.

extensions/empty-screenshots/package.json
```diff
@@ -5,6 +5,9 @@
   "description": "Empties your screenshot folder within a configurable time period.",
   "icon": "empty-screenshot-folder.png",
   "author": "chanceaclark",
+  "contributors": [
+    "charlie_deets"
```
chanceaclark, line 9: 🍹 
```suggestion
    "robotdestroy"
```

chanceaclark (commented):
Hey @robotdestroy thanks for contributing to this extension! I have a few comments regarding this change but overall I love this addition!

🍹 = Optional or discussion comments so feel free to disregard them if you feel passionately about it.

Additionally, I am about to get married/go on my honeymoon for the next two weeks so I won't be able to respond much after this message. After this review, I will defer the approval of this PR to the Raycast team or valid approver. Hopefully this doesn't impact the resolution time too much 🙏

extensions/empty-screenshots/package.json
robotdestroy: From my understanding, this is meant to be my Raycast account name, and not my GitHub name, so this one actually is correct. Thanks for caring! :)

extensions/empty-screenshots/package.json
robotdestroy: Actually this was from an earlier attempt at this implementation, from what I can tell trash-cli is not necessary. I am using osascript to do the trash movement now so I will just remove this.

extensions/empty-screenshots/package.json
pernielsentikaer: That's correct 🙂

pernielsentikaer:

robotdestroy:
Yup!

extensions/empty-screenshots/src/index.ts
```diff
@@ -2,22 +2,42 @@ import { readdir, stat, rm } from "fs/promises";
...
-  const folder = resolve(screenshotFolder.replace("~", homedir));
+  const folder = resolve(screenshotFolder.replace("~", homedir()));
 
   try {
     const files = await readdir(folder);
 
     const screenshots = files
       .filter((file) => file.endsWith(".png") || file.endsWith(".mov"))
+      .filter((file) =>
+        prefixString ? file.startsWith(prefixString) : file.startsWith("CleanShot") || file.startsWith("Screen"),
```
pernielsentikaer, line 39: I just tried to change language to danish and the name here is `Skærmbillede 2024-05-22 kl. 09.08.22.png` - this will never be removed, so maybe it's better to do as before and only filter if custom `prefixString` has been set (which we could make to a comma seperated list)

pernielsentikaer: Are you still around @robotdestroy

robotdestroy: Yup, will clean this up this week!

mil3na:
Hi @robotdestroy ! Do you still want to work on this PR and take a look at Per's review?

PR #12386, opened 2024-05-16
Add battery-optimizer extension

Qetesh:
## Description

Protect battery healthy by limiting the charging capacity.

## Screencast

image
battery-optimizer-2

## Checklist


erics118:
Would it be possible to instead require users to install BCLM via brew, rather than to copy the entire repo over? That way, updates to BCLM don't have to be manually PR-ed into the Raycast extension repo

pernielsentikaer:
How is this compared to the two we already have in [the store](https://www.raycast.com/search?q=battery)

Qetesh:
> How is this compared to the two we already have in [the store](https://www.raycast.com/search?q=battery)

These two extensions display battery information, while the extension being submitted adjusts the battery charging threshold through system settings to protect battery health.[[1]](https://www.apple.com/batteries/why-lithium-ion/)[[2]](https://support.apple.com/en-us/108055)

pernielsentikaer:
For me it feels like the built-in version is MUCH faster than the one from brew, is that expected?

Qetesh:
Not the expected outcome.

Reviewed the code and found that the main difference is that the brew version uses the which system command more frequently. This usually wouldn't cause a significant speed difference.
I tested it on two Macs and didn't reveal any noticeable speed differences. The specific reason is still under investigation.
If possible, could you help identify the cause? I am a junior programmer and appreciate your patience.

pernielsentikaer:
What execution times do you get? It might be fine to just wait a bit longer if you don't need a third-party library

Qetesh:
Add detailed logging and test BCLM in various scenarios. Verify the BCLM path via logs and confirm no significant time discrepancies
Including the time to enter the password
image
image

pernielsentikaer:
I agree with @erics118, it's a bit confusing to have multiple option and I would maybe also opt-in for the brew, since it will gets updated 🙂

Many extensions have possibility to install from brew, let me know if you need guidance

Qetesh:
I agree it was a bit confusing. I've removed the built-in binaries and kept the brew option. If you have any more suggestions, just let me know!

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #12396, opened 2024-05-16
[Say] Fix voice picker

LitoMore:
## Description

- Fix the voice dropdown

The voice picker will be always the `System Default` due to the first dropdown item is `System Default`.

The `onChange` API is weird. It will always be executed on the first render. I'm not sure it's by the design or bug.

So I added a condition to make sure `System Default` dropdown item doesn't appear on first render.


pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #12434, opened 2024-05-17
Update drupal-org extension

BossElijah:
## Description

- Added command to search for Drupal projects which include themes, modules, and distributions.
- Also did minor Typescript refactor as agreed upon with the creator of the extension.

## Screencast

drupal-org-4

## Checklist


BossElijah:
I don't know what is wrong with the GIT history, but for some reason it includes last PR's commit messages in this PR. What was actually changed in this PR was a bit of typescript refactor and added command to search projects.

extensions/drupal-org/package.json
```diff
@@ -27,6 +27,13 @@
       "subtitle": "Drupal",
       "description": "Search Drupal API documentation",
       "mode": "view"
+    },
+    {
+      "name": "search-projects",
+      "title": "Search Projects",
+      "subtitle": "Drupal",
+      "description": "Search for Drupal projects",
```
hussainweb, line 35: Minor: I'd love it that someone can type in "module" or "theme" and be able to select this command. With that intent, what do you think of this change?

```suggestion
      "description": "Search for Drupal projects (modules, themes, etc)",
```

hussainweb (commented):
I would have preferred a separate change for the refactor but that may not be a great idea here. I like how you have generalized the command state but wonder how to extend it if future commands need more data. That's not a showstopper here though.

Generally, I am wondering if the practice of having these multiple `util.ts` and `types.ts` is typical in TypeScript projects. I don't have a lot of experience with TypeScript. I am only raising a question out of curiosity and am open to either direction.

extensions/drupal-org/package.json
BossElijah: Yeah that sounds like a great idea!

BossElijah:
> I would have preferred a separate change for the refactor but that may not be a great idea here. I like how you have generalized the command state but wonder how to extend it if future commands need more data. That's not a showstopper here though.

I see what you mean there. I decided to do it this way in order to get it done quicker, however it would probably be best-practice to have something separate. I'd suggest we just keep it as-is for now, but note it for next time. Do you agree on that @hussainweb?

> Generally, I am wondering if the practice of having these multiple `util.ts` and `types.ts` is typical in TypeScript projects. I don't have a lot of experience with TypeScript. I am only raising a question out of curiosity and am open to either direction.

I mean, it's not necessarily a TypeScript related practice, but from my experience, it has always been easier to maintain projects that implemented it this way, especially with the `util.ts` files. If some day we want to expand the capacity of one of the commands, it will be easier to review and contribute. That being said, the `types.ts` files don't necessarily have to be in different files, I just think it's a matter of personal preference.

hussainweb:
Yes, I understand and agree. I see you have merged my other suggestion and this LGTM.

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #12454, opened 2024-05-19
feat: add create page for notion

lishoulong:
## Description
add notion create page for raycast extension, now we only support create database page and append to a existing page.

## Screencast
image

## Checklist

### DIFF
--- a/extensions/figma-variables/CHANGELOG.md
+++ b/extensions/figma-variables/CHANGELOG.md
@@ -1,3 +1,7 @@
 # Figma Variables Changelog
 
+## [Fix] - 2024-05-28
+
+- Update Figma URL structure in README.md
+
 ## [Initial Version] - 2024-02-20
--- a/extensions/figma-variables/README.md
+++ b/extensions/figma-variables/README.md
@@ -14,7 +14,7 @@ This Raycast extension allows users to interact with their Figma designs by fetc
 
 ### Locating Your File Key
 
-The File Key is part of your Figma file's URL: `https://www.figma.com/file/<FILE_KEY>/...`.
+The File Key is part of your Figma file's URL: `https://www.figma.com/design/<FILE_KEY>/...`.
 
 ## Using the Extension
 


bkeys818:
Could this be implemented into the `Create database page` command?

For example, you could add `My Workspace` as an option at the bottom of the database selector.

lishoulong:
> Could this be implemented into the `Create database page` command?
> 
> For example, you could add `My Workspace` as an option at the bottom of the database selector.

Get，I have considered this logic; it's possible that the code is deemed to have too much coupling.

lishoulong:
> Could this be implemented into the `Create database page` command?
> 
> For example, you could add `My Workspace` as an option at the bottom of the database selector.

I thourouly review the two parts code, 80% of  CreatePageForm.tsx logic is relative with database, I hava a idea of the following code, move action part to higher level. @bkeys818 please have a look.

```
// PageForm.tsx
<Form
      isLoading={isLoadingDatabases || isLoadingRelationPages}
      navigationTitle={initialDatabaseId ? "Create New Page" : "Create Database Page"}
      actions={
        <ActionPanel>
          <Action
                  title="Switch database page and page mode"
                  onAction={() => {
                    setMode((databaseMode) => !databaseMode);
                  }}
         />
        </ActionPanel>
      }
    >
   databaseMode && (<DatabasePage></DatabasePage>)
   !databaseMode && (<NormalPage></NormalPage>)
</Form>
```

bkeys818:
Ah, I may have misunderstood the feature you are trying to add.

Are you trying to add the ability to create a new subpage inside of the content of an existing page?

lishoulong:
> Ah, I may have misunderstood the feature you are trying to add.
> 
> Are you trying to add the ability to create a new subpage inside of the content of an existing page?

yes, that is what I want to do

bkeys818:
Hmmmm... Ok. That's going to be difficult to do without adding features that will be unnecessary for 99% of users. If we can't think of implementing it without changing functionality or adding commands, then we'll use what you have.

Can I ask what your use case is?

lishoulong:
> Hmmmm... Ok. That's going to be difficult to do without adding features that will be unnecessary for 99% of users. If we can't think of implementing it without changing functionality or adding commands, then we'll use what you have.
> 
> Can I ask what your use case is?

Ok, I usually use it for taking a note between work

bkeys818:
When are you using the Raycast Notion extension and need to create a new subpage inside a page?

I create subpages all the time, but only when I'm inside the app.

bkeys818:
@thomaslombart, can I get your input here?

thomaslombart:
I see three solutions here:
1. Merge `Create Page` in `Create Database page` meaning that the database field should return all pages, which would add complexity to this command both on a technical and user's point of view (Add pages to the `Database` field and renaming it, change the UI if we're not adding a database page, hide some actions). It's not that complex but still.
2. Create a different command as it's done here.
3. Merge that in the `Add Text to Page` with a checkbox called `Add as Page`. This could simplify things but the command should use a better name to make it clear you can both add text to a page and also create a subpage.

Solutions 2 and 3 are fine to me. What do you think?

bkeys818:
@thomaslombart, What about an action inside of `Add Text to Page` to create a subpage?

thomaslombart:
That's what I meant with the 3rd solution 🙂

lishoulong:
thanks for @bkeys818 @thomaslombart , I think third solution is making sense , ie, merge the create page into "Add Text to Page". I will have a try.

extensions/notion/src/utils/notion/page.ts
```diff
@@ -69,6 +112,7 @@ export async function search(query?: string, nextCursor?: string) {
       direction: "descending",
       timestamp: "last_edited_time",
     },
+    filter: { property: "object", value: "page" },
```
lishoulong, line 115: add page filter, @bkeys818 @thomaslombart  please  help review is this logic make sense.
In the "append text to page" scene, according to our ui form field, I think we are not will to add text to dataPage table?

PR #12456, opened 2024-05-19
Update date-converter extension

vineus:
## Description

* add 24h time format option (disabled by default)

## Screencast

image

Raycast 2024-05-19 at 14 04 28

image

Raycast 2024-05-19 at 14 04 56

## Checklist

### DIFF
--- a/extensions/date-converter/src/index.tsx
+++ b/extensions/date-converter/src/index.tsx
@@ -190,7 +190,7 @@ export default function Command() {
         <List.Item
           key={date.toISOString()}
           title={humanFormatter.format(date)}
-          subtitle={label + "﹣" + timeAgo.format(date)}
+          subtitle={`${label} - ${timeAgo.format(date)}`}
           actions={
             <ActionPanel>
               {getSortedFormats({ human }).map(({ id, title, format }) => (


asportnoy (changes requested):
Could the relative info be part of the item description maybe? Also does it need to be that detailed?

vineus:
@asportnoy 
> Could the relative info be part of the item description maybe? Also does it need to be that detailed?

Better?

date-converter 2024-05-19 at 22 50 48

asportnoy:
- I think the relative time is still too detailed. "9 minutes" should be plenty detailed for this, I think the seconds is overkill
- Can you add a prefix or suffix like "in/ago"? So "9 minutes ago"
- Separate the two parts with something like a hyphen and ditch the "relative:"

asportnoy (commented):
Marking as reviewed again - see above

vineus:
Reverted the relative time work.

I'll use my fork, I guess my needs are very specific.

pernielsentikaer:
We can make it a preferences to show seconds or not, I like the idea 🙂

vineus:
Found some time to update.

Here is a shorter version:

image

image

image

pernielsentikaer:
@asportnoy could you check it once again

extensions/date-converter/src/index.tsx
```diff
@@ -183,7 +190,7 @@ export default function Command() {
         <List.Item
           key={date.toISOString()}
           title={humanFormatter.format(date)}
-          subtitle={label}
+          subtitle={label + "﹣" + timeAgo.format(date)}
```
asportnoy, line 193:
```suggestion
          subtitle={`${label} - ${timeAgo.format(date)}`}
```

asportnoy (commented):
Much better! One minor thing and then I think it's good to go.

asportnoy (approved):
Thanks for those changes, looks good to me.

@pernielsentikaer ready to merge when you are.

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

Issue #12120, opened 2024-05-01
[Coffee] The caffeinate state does not show up in the status bar - impossible to disable


gregkoganvmm:
### Extension

https://www.raycast.com/mooxl/coffee

### Description

The caffeinate state does not show up in the status bar - it is impossible to disable the "caffeinated" state from the status bar. The only way to do it is to find and kill the `caffeinate` process

### Steps To Reproduce

Environment: macOS Sonoma 14.4.1, Raycast 1.72.1
Click the "caffeinate" button

### Current Behaviour

The `caffeinate` process is launched, but the state is not reflected in the status bar and you cannot click the "decaffeinate" button. The decaffeinate command has no effect as well.

### Expected Behaviour

Should be able to disable the caffeinated state from the status bar

xilopaint:
The issue has been resolved.

@raycastbot close this issue

Issue #11625, opened 2024-04-04
menu bar needs manual updating


Oblique82:
### Extension

https://www.raycast.com/mooxl/coffee

### Description

The caffeinate status in the menu bar doesn't update and need to be manually activated instead of truely reflecting the status 

### Steps To Reproduce

Activate caffeinate / decaffinate and watch the menubar

xilopaint:
The issue has been resolved.

@raycastbot close this issue

Issue #11559, opened 2024-03-29
[Coffee] Formatting issue caffeinateFor validation


julienmunio:
### Extension

https://www.raycast.com/mooxl/coffee

### Description

Formatting issue of validation of action Caffeinate For:
`Caffeinating your Mac for 30m2h` and not `Caffeinating your Mac for 2h30m`

### Steps To Reproduce

On action `Caffeinate For` fill Hours and Minutes

### Current Behaviour

`Caffeinating your Mac for 30m2h`

### Expected Behaviour

`Caffeinating your Mac for 2h30m`

xilopaint:
The issue has been resolved.

@raycastbot close this issue

Issue #11463, opened 2024-03-23
[Coffee] ...


PawGangsta:
### Extension

https://www.raycast.com/mooxl/coffee

### Description

Whenever i activate "caffeinate" the extension doesn't really inform whether the action has been taken or not, which can be easily fixed with some UX :)

### Steps To Reproduce

1. try to caffeinate anything

pernielsentikaer:
There should be a HUD in the bottom middle of your screen showing this, doesn't that happen?

PawGangsta:
Hello,

Thank you for getting back to me :) and no, there is no HUD in the bottom middle of my screen, no matter how many times i click Caffeinate nothing changes

Khaled Adel
Product Designer

This email and any files transmitted with it are confidential and intended solely for the use of the individual or entity to whom they are addressed. If you have received this email in error, please notify the system manager. Please consider the environment before printing this email.

pernielsentikaer:
You should see this 
![Raycast03042024-NzLvLTSo png@2x](https://github.com/raycast/extensions/assets/38033107/1b05a921-0293-470b-bb5e-86f4158cae18)

xilopaint:
@raycastbot close this issue

PR #12459, opened 2024-05-19
Update coffee extension

### LINKED ISSUE #9239: [Coffee] One click menubar toggle
### Extension

https://www.raycast.com/mooxl/coffee

### Description

Would be great if the menubar could be toggled in 1 click (like the og Caffeine app)

Users

### Anything else?

_No response_

xilopaint:
## Description

Fixed #9239 #11463 #11559 #11625 #12120

- Replaced the form in the `Caffeinate For` command with arguments in the root search.
- Replaced the emojis in the `Caffeinate Status` command with Unicode characters for enhanced design consistency.
- Updated some menu bar icons.
- Fixed an issue where it was not possible to decaffeinate the Mac via the `Caffeinate Status Menu Bar`.
- Fixed an issue where the menu bar icon was not immediately updated when the Mac was caffeinated.
- Fixed the `Caffeinate While` command not working.

## Checklist

### DIFF
--- a/extensions/coffee/src/caffeinateFor.tsx
+++ b/extensions/coffee/src/caffeinateFor.tsx
@@ -2,14 +2,13 @@ import { showToast, Toast } from "@raycast/api";
 import { startCaffeinate } from "./utils";
 
 interface Arguments {
-  hours?: string;
-  minutes?: string;
-  seconds?: string;
+  hours: string;
+  minutes: string;
+  seconds: string;
 }
 
 export default async function Command(props: { arguments: Arguments }) {
   const { hours, minutes, seconds } = props.arguments;
-
   const hasValue = hours || minutes || seconds;
 
   if (!hasValue) {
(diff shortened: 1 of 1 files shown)


extensions/coffee/src/utils.ts
```diff
@@ -32,22 +32,22 @@ async function update(updates: Updates, caffeinated: boolean) {
...
-
+  exec(`/usr/bin/caffeinate ${generateArgs(additionalArgs)} || true`);
+  await update(updates, true);
   if (hudMessage) {
     await showHUD(hudMessage);
   }
+}
 
-  execSync(`/usr/bin/caffeinate ${generateArgs(additionalArgs)} || true`);
+export function isValidInteger(value: string | undefined): boolean {
```
aliaksandrZh, line 51: Can't we use a ```Number.isInteger()```?

xilopaint: Done.

aliaksandrZh:
Nice work ! LGTM

pernielsentikaer (commented):
The only think I can come up with is that users who have selected an icon for the menubar which is now deleted might end up with an empty icon

Can we either fallback to something, or map to old names so they also "works"

xilopaint:
> The only think I can come up with is that users who have selected an icon for the menubar which is now deleted might end up with an empty icon

Are you sure? In my test the default icon is set when the value of the selected icon is changed in `package.json`.

xilopaint:
> The only think I can come up with is that users who have selected an icon for the menubar which is now deleted might end up with an empty icon

@pernielsentikaer Are you sure? In my test the default icon is set when the value of the selected icon is changed in `package.json`.

pernielsentikaer:
Should we keep a mapped array for the old names so users don't suddenly see a new icon? Could that work in any way?

xilopaint:
> Should we keep a mapped array for the old names so users don't suddenly see a new icon? Could that work in any way?

I don't think it's a necessity since the user won't be left without an icon. Additionally, users who are no longer using the default icon have already gone through the experience of having to change it once and I believe it won't be a big trouble to do it again once they notice their icon has changed. That said, I believe the small benefit of avoiding the icon change to the default does not outweigh the inconvenience of adding complexity to the code that may not be immediately clear to contributors.

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

edheltzel:
@xilopaint Love the extension, but the user experience is suffering with this update:
> "Replaced the form in the `Caffeinate For` command with arguments in the root search."

This is just a "me" use case, but I configured a keyboard shortcut (hyperkey+7) to the `Caffeinate For`. This was possible only because of the form input. Now with the arguments having to be passed through the root search, its not possible. 

Error - Screenshot 2024-05-28 at 8 01 00 AM

This Update - Screenshot 2024-05-28 at 8 04 30 AM

pernielsentikaer:
I also think the new approach is better, but I can see the issue since the parameters in the command are optional. One thing we could do is add some preferences that set the default caffeinate time, so it will use the default values if nothing is specified.

We could also add another command (which is disabled by default) to bring back the form view. Could that work? 🙂

edheltzel:
both are great solutions. I personally would enjoy the later of the two having the option to bring back the form. I definitely appreciate it.

xilopaint:
I don't like the idea of bringing back the form. The problem here is that all the arguments are individually optional but filling in at least one of them is required.

@pernielsentikaer, would it be possible to work with the team on this type of validation (at least one argument must be filled) so that the hotkey brings the focus to the first argument instead of trying to run the command?

pernielsentikaer:
I can't promise you anything, but yes, I'll try to bring it up with the team if we can figure out a way to achieve this without changing too much. This is also a niche case but I agree with @edheltzel that it doesn't work very well at the moment.

I'll get back to you 🙂

pernielsentikaer:
@xilopaint, it will stay as is at the moment. We added it to the backlog and will see if it pops up more regularly 🙂

We could do one of the following:

-  Idea 1) If no fields are filled, we show the form
-  Idea 2) Just have one argument where you can write something like `1h 30m` or `1:30` instead

Do you have any input here, @edheltzel

edheltzel:
Idea 1 is my preferred... And seems to be the least amount of work to achieve. 

Other than that I do appreciate everything.

PR #12540, opened 2024-05-23
Update cursor extension

pedrotmr:
## Checklist

### DIFF
--- a/extensions/cursor/src/preferences.ts
+++ b/extensions/cursor/src/preferences.ts
@@ -5,12 +5,6 @@ const preferences = getPreferenceValues<ExtensionPreferences>();
 
 function getBundleIdentifier() {
   switch (preferences.build) {
-    case VSCodeBuild.Code:
-      return "com.microsoft.VSCode";
-    case VSCodeBuild.Insiders:
-      return "com.microsoft.VSCodeInsiders";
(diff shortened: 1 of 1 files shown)


erics118:
I'm curious why this was separated from the vscode extension. Since Cursor basically is a based on the same [`Code - OSS`](https://github.com/Microsoft/vscode) as vscode, the internals work identically to VSCode. Currently, the implementation is just to switch the "build" to Cursor, and everything works just fine.

pernielsentikaer:
As @erics118 mentions we already have an extension for `Cursor` which works inside [this extension](https://www.raycast.com/thomas/visual-studio-code) - I would rather add some keywords to that extension instead so it can be found if you search for `cursor`, `codium` and so on

pedrotmr:
Visually it makes difference. Ok, I can launch cursor with the vscode extension, but icons are still from vscode, action names as well. It does have the functionality, but the look/interface is not what we are looking for. Unless there's a way to toggle the name of the actions and icons as well...

pernielsentikaer:
Since Cursor is built on VSCode, it makes sense to only have one extension to maintain. Otherwise, it will be impossible to keep both up to date with the exact same features

pernielsentikaer:
Closing since it doesn't make sense to add a 1:1 copy

Issue #11897, opened 2024-04-19
[Timers] No sound when timer finishes


asgeirtj:
### Extension

https://www.raycast.com/ThatNerd/timers

### Description

Alarms don't work (give no sound)

### Steps To Reproduce

Start any timer and wait for it to finish.

ThatNerdSquared:
Hi @asgeirtj, thanks for using Timers.

Are you getting a banner notification like the one below? Or does that not show up either?

![image](https://github.com/raycast/extensions/assets/72814106/e75dd9d7-9c70-45fe-9fad-9dd78e02cb36)

Note: may be linked to #11500, will be investigating this soon.

@raycastbot rename this issue to "[Timers] No sound when timer finishes"

asgeirtj:
> Hi @asgeirtj, thanks for using Timers.
> 
> Are you getting a banner notification like the one below? Or does that not show up either?
> 
[...]

Issue #11500, opened 2024-03-26
[Timers] No sound played after the time finish


ivanmauricio:
### Extension

https://www.raycast.com/ThatNerd/timers

### Description

no sound is playing when the timer is finished. may be linked to #2893

i can play a sound if i use the `preview alert sounds` command

i also ran this command in the terminal and it works: `afplay /System/Library/Sounds/Hero.aiff`

and here are the relevant settings

![CleanShot 2024-03-26 at 12 26 26](https://github.com/raycast/extensions/assets/12462448/33eedaf0-e8ef-48f8-9b90-231ab95ab66d)

(i've also tried with different selected sounds)

### Steps To Reproduce

1. set a timer off
2. wait for it to end
3. no sound plays

ThatNerdSquared:
Hi @ivanmauricio, thanks for using Timers!

That's a weird one - thank you for all the debugging information you've provided.

Do you get a banner notification like this one when your timer finishes?

![Screenshot 2024-03-27 at 22 14 11](https://github.com/raycast/extensions/assets/72814106/e642d861-4d23-4f79-8c16-7104ee75d4ca)

Also, when did you start having this problem? Did it start ever since you installed the extension, or did it work for you before?

Thanks again.

ivanmauricio:
thanks for your quick reply @ThatNerdSquared!

i'm not getting a banner notification. i've checked the notification system settings for macOS and they are allowed for `Script Editor`.

the problem started since I installed the extension on a new macbook (i.e. the end sound has never played on this laptop). i had it on my old macbook and it was working fine.

i've tried uninstalling and installing the extension and also restarting my computer.

I also tried the sleep timer extension (https://www.raycast.com/dtmzr/sleep-timer) as I saw it has similar functionality to this one and that works as intended.

ivanmauricio:
so i downloaded the repo to take a look and looks like the problem is that the command `osascript -e 'display notification "Timer \\"${timerName}\\" complete" with title "Ding!"'` doesn't execute and so all the commands after it don't run as they're chained together. i've downloaded the repo locally and removed the notification command and everything seems to be working. 

i'm happy that it's working as i was relying on this timer! though would be curious if you have any ideas why `osascript` command might not be working from `node` on a new machine. i had a look at permissions and have all the same ones as on my old laptop  🤷🏽‍♂️

here's what i changed the code to in case it's of interest

```
async function startTimer(timeInSeconds: number, timerName = "Untitled", selectedSound = "default") {
[...]

PR #12551, opened 2024-05-24
feat(Timers): add custom notification system using Swift app bundle

ThatNerdSquared:
## Description

- Add a custom notification system using a Swift app bundle to call into the system `User Notifications` framework
- Advantages of custom notification system:
    - Avoids silent permission errors when using `osascript`
    - Banner notification icon matches extension icon
    - Adds actions to banner notification for dismissing the timer or opening up "Manage Timers" command

## Screencast

Screenshot 2024-05-23 at 23 17 16

## Checklist


ThatNerdSquared:
@pernielsentikaer — for this feature to work properly, I need to include the compiled Swift app bundle somehow on the user's machine. What's the best way to go about this?

- keep a compiled bundle in `assets/`? Seems like that would take up a lot of space in the repo.
- download it on extension install from some hosted location (e.g., GitHub Releases)?
- some other method?

Also, I don't have a dev license to sign the bundle with. Not sure if that's an issue or not.

pernielsentikaer:
You can check how extensions like [keyboard-cleaner](https://github.com/raycast/extensions/pull/12449) or [color picker](https://www.raycast.com/thomas/color-picker) does it, it will auto compile it 🙂

ThatNerdSquared:
I've tried using the Swift for Raycast method, which is what those extensions do, but unfortunately it doesn't work for this use case. The API I'm using needs to be run from an app bundle, which doesn't happen using that method (i asked about it [here](https://raycastcommunity.slack.com/archives/C02HEMAF2SJ/p1715069030390119?thread_ts=1714605923.296799&cid=C02HEMAF2SJ) on the Raycast Slack).

pernielsentikaer:
You should be able to use the "old" method, like it's [done here](https://github.com/raycast/extensions/tree/36a96ef5bebbbcea270809a664fc2e2e2ca56359/extensions/ruler)

ThatNerdSquared:
Oh, this is interesting. How does this work, exactly? Is there documentation about it somewhere? I'm not sure if I understand the extension structure here.

pernielsentikaer:
I'm not sure we have that much documentation of this way, you can see how it's triggered in package.json - first version of Color Picker also used this method afaik

ThatNerdSquared:
Alright, that makes sense. Do you know what this binary file is in the `assets/` for that extension?

Screenshot 2024-05-24 at 00 19 51

pernielsentikaer:
That's the one that was build with the `build-swift` npm run script in package.json

ThatNerdSquared:
Unfortunately, this method does not work either. The executable is compiled but since it is not in an app bundle the API fails to work (i get a `NSInternalInconsistencyException` with reason `bundleProxyForCurrentProcess is nil`).

pernielsentikaer:
I asked internally and got this reply, not sure if it works at all
> This might work, though I haven't used it for a long time and I'm not sure you can set the icon: https://github.com/julienXX/terminal-notifier

ThatNerdSquared:
@pernielsentikaer I've looked at that, unfortunately it does not solve the problem as it is also distributed as an app bundle (and at that point I might as well just use the one I've written, which has more functionality).

In general, other methods of sending notifications to the user have one or more of the following problems:
1) They require an app bundle (same as my solution) or use Applescript
2) They use the deprecated `NSUserNotification` API rather than the newer `UserNotifications` framework
3) They use private methods not exposed as public APIs by Apple
4) They are missing functionality (no custom icon, no custom actions, etc)

This is because many things like the custom icon, custom actions, etc are only available through system APIs (not through Applescript, etc). Additionally the `UserNotifications` framework only works from within an app bundle.

I do not want to continue using Applescript (it has very unstable permissions, as shown [here](https://github.com/raycast/extensions/issues/11500) and [here](https://github.com/raycast/extensions/issues/11897)), and I do not want to use the deprecated API or private methods as Apple may block their usage at any time. So I think my current solution is probably the best one.

Possibly I can use a method similar to the Ruler extension, by including only the Swift file and app bundle? Then we will not have to add all the xcode files as well. But I think that's the best we can do.

ThatNerdSquared:
As I haven't found a better way to do this, I've copied in the app bundle to the `assets/` folder. Should be ready for review.

The app bundle is still not signed - is this an issue? If so, would it be possible for someone from the Raycast team to build/sign it? Or for this to be set up as a CI job somehow?

ThatNerdSquared:
Hi @pernielsentikaer, any news regarding this? Would be great if we could get it shipped :)

pernielsentikaer:
Sorry for the delayed response.

I had to check what options were available. Unfortunately, we can’t support that at the moment, but we are working on a notification API which should give options for doing as you want. However, there is no ETA yet 🫠

ThatNerdSquared:
If I get the app bundle signed, would it be acceptable? Or is the additional app bundle itself the dealbreaker?

pernielsentikaer:
It's not possible at the moment but we're going to revamp toast/hud/notification where something like this will be possible - sorry it's not as simple as it could be

ThatNerdSquared:
I see there are [other extensions](https://github.com/raycast/extensions/tree/6c8a2b516913f96205d17a7d269a5a3d1f5448eb/extensions/pins/assets/scripts/PinsLocationManager.app/Contents) that tap into system APIs via an app bundle - it is quite unfortunate that the same cannot be done for notifications.

In any case - thank you for your time reviewing this. If there are no other solutions for the time being, feel free to close the PR.

pernielsentikaer:
We would like to support this out of the box with the revamp that we're going to add - I also tried to build this myself and I was not able to get it to work, not sure if it's something with the signing 🫠

ThatNerdSquared:
> I also tried to build this myself and I was not able to get it to work

What error did you run into? Maybe we can debug it. I can also try getting it signed if that helps.

> We would like to support this out of the box

That would be great :) It would still be nice to push a solution to the users in the meantime though.

If it's not worth the trouble of debugging the build, feel free to close the PR - I'd understand if the team feels this is out of scope for the time being.

pernielsentikaer:
For now, I'm closing it, but I will keep you in the loop about this feature internally

PR #12564, opened 2024-05-24
[PM2] Initial version

LitoMore:
### Features

- View PM2 process list
- Manage PM2 processes (start, stop, restart, reload, delete)
- Run any Raycast Node.js application in PM2 through [Cross-Extension](https://github.com/LitoMore/raycast-cross-extension-conventions)

### Requirements

- Get Node.js and npm installed on your machine
- Get PM2 installed on your machine

### Principles

Due to runtime reasons, the [pm2](https://npmjs.com/pm2) package won't work in Raycast extension. We used a PM2 wrapper application under the `assets` folder to bypass this problem. The extension will run `npm install` under the `pm2-wrapper` folder during the first run.

### Folder structure

The `pm2-wrapper` is placed under the `assets` folder. It's a standalone Node.js application. You don't need to run `npm install` unless you need to modify the package related information.

Please make sure there is no `node_modules` under the `pm2-wrapper` folder. The Raycast will build all assets to `~/.config/raycast/pm2/assets`. You should verify the `node_modules` isntallation behavior there.

### Development

If you have modified `pm2-wrapper`, you will need to stop and restart the dev server (`npm run dev`) to apply the new changes.

Please note that when you run Node.js in Raycast's box, the environment variables and PATH information you get are not the same as when running directly in your Terminal.

### Play around

Use the `Start Example Process` action to initialize a Node.js application through Raycast.

image

## Checklist


extensions/pm2/src/components.tsx
```diff
@@ -0,0 +1,147 @@
...
+      })}
+      <List.Item.Detail.Metadata.Separator />
+      <List.Item.Detail.Metadata.TagList title="status">
+        {typeof p.pm2_env?.status === "string" && (
+          <List.Item.Detail.Metadata.TagList.Item
+            text={p.pm2_env?.status}
+            color={getProcessStatusColor(p.pm2_env?.status)}
+          />
+        )}
+        {typeof p.monit?.cpu === "number" && (
```
Elmona, line 55: This code could be improved, instead of typing a lot of 
```
typeof p.pm2_env?.status === "string" && `
```

It's better to check for that earlier in the function, if you do that you don't need to duplicate your logic that you do here. I mean that's better code. (No AI involved just a dev reading code)

Elmona: early return pattern, guard close.. Let me show you an article.
https://medium.com/swlh/return-early-pattern-3d18a41bba8

That's not the best to explain there is some that are better, but you get the idea.

Elmona: actually when you always go for an early return pattern you will never look back. That is a lot better.

LitoMore: @Elmona Hey, thanks. I know this. But no. I want to keep my code style consistent here for now. I plan to refactor this part of the code entirely, but I don't want to spend too much energy on so-called "best practices".

I prefer to find out if an ESLint rule could automatically check and format the code into a unified format. I don't want to spend energy on these ambiguous code styles.

In addition, if you are obsessed with code style, I recommend that you learn about [XO](https://github.com/xojs/xo) code style. That is what I expect.

Anyway, thanks for the advice.

Elmona: Fine, truth be told and most looked good, i'm not obsessed with any pattern but I always try to choose the best myself.

I don't mean this as an git blame more that here is a file you will improve on next time you type it ❤️

Elmona (approved):
LGTM 🚀

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #12600, opened 2024-05-27
[Bugfix] Set ms to 0 to cache each sec instead of each ms

klmz:
## Description

1. I ran into a lot of ratelimit when turning on the menubar item. It seems because request are cached each ms. So, I've set the ms to 0 such that they are cached by second. This reduces the amount of requests that are not cached significantly.

2. Also the tasklist is not loaded when you first go to the new task list. It seems like a promise is not awaited, so when it returns after render the list is empty. Now the useEffect also triggers when the list of projects is updated.

3. (edit, found another one) In the menubar item when the elapsed time is close to 10 minute passed a whole hour, for instance 1.16665, you'd expect it to be rendered as x:10, but it is rendered as x:010, because of a weird rounding issue. I changed the code to render that time differently which I think fixes that issue.

## Checklist

### DIFF
--- a/extensions/harvest/src/services/harvest.ts
+++ b/extensions/harvest/src/services/harvest.ts
@@ -222,10 +222,11 @@ export function formatHours(hours: string | undefined, company: HarvestCompany |
   const { timeFormat }: Preferences = getPreferenceValues();
 
   if (timeFormat === "hours_minutes" || (timeFormat === "company" && company?.time_format === "hours_minutes")) {
-    const time = hours.split(".");
-    const hour = time[0];
-    const minute = parseFloat(`0.${time[1]}`) * 60;
-    return `${hour}:${minute < 10 ? "0" : ""}${minute.toFixed(0)}`;
+    const minutes = parseFloat(hours) * 60;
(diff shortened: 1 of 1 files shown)


pernielsentikaer:
Thanks for y our contribution @klmz 🔥 

@eluce2 can you check this?

eluce2:
the useProjects is a hook, so it shouldn't be marked as an async function. The ms fix does seem like a good way to reduce the cache, although it makes me wonder if we should cache every 5 seconds to reduce the rate limiting even further.

Can I edit this PR directly, or should I try some things in my own branch?

pernielsentikaer:
> Can I edit this PR directly, or should I try some things in my own branch?
@klmz should be able to add you to the [fork](https://github.com/klmz/extensions/tree/feature/fix-cache) - let me know if you have any troubles doing so.

Another way is just to send me the code in Slack and I can correct directly since maintainers can push

klmz:
I've added @eluce2 as collaborator if he wants to go that way.

eluce2:
I made a new PR that incorporates these changes. Should be linked above

Issue #12516, opened 2024-05-23
[Toggle Audio Input (microphone)] not restoring last audio input LEVEL


NotFiliberto:
### Extension

https://www.raycast.com/Quentin23Soleil/mute-microphone

### Raycast Version

1.75.1

### macOS Version

14.5

### Description

The extension in not able to restore the previous input volume level. For example if is set to 50% and and toggle twice the extension, I should have the volume input level still to 50% and not 100%.

### Steps To Reproduce

1. Press Option+F10 to open the sound settings in macos.
2. Set the input volume level.
3. Toggle twice the extension (with hotkey or with raycast)

### Current Behaviour

The extension restore the input volume level to 100%.

### Expected Behaviour

The extension should restore the input volume level to the last one saved by the user. So if I the volume level before muting the microphone with the extension was set to 50% before toggling the extension, when the users toggles again the extension to unmute the audio input, the input volume level should be restore to 50%.

1activegeek:
Noticed very same issue on my machine today attempting to use this plugin. Also running same versions listed.

aliaksandrZh:
Hi there! @NotFiliberto @1activegeek https://github.com/raycast/extensions/pull/12614

PR #12614, opened 2024-05-28
[Toggle Audio Input (microphone)] not restoring last audio input LEVEL

aliaksandrZh:
## Description
Issue: https://github.com/raycast/extensions/issues/12516
Audio input level is maintained by default

## Screencast
[Screen Recording 2024-05-28 at 11.23.00.webm](https://github.com/raycast/extensions/assets/108681764/d883265f-b4b9-4934-85d3-5eb3767dee8a)

## Checklist

### DIFF
--- a/extensions/mute-microphone/package.json
+++ b/extensions/mute-microphone/package.json
@@ -1,7 +1,7 @@
 {
   "$schema": "https://www.raycast.com/schemas/extension.json",
   "name": "mute-microphone",
-  "title": "Toggle Audio Input (microphone)",
+  "title": "Toggle Audio Input (Microphone)",
   "description": "Extension to mute/unmute audio input (Microphone)",
   "icon": "mute-command-icon.png",
   "author": "Quentin23Soleil",


pernielsentikaer:
Thanks for your contribution @aliaksandrZh 🔥 

@quentin23soleil  do you want to check this

quentin23soleil:
@pernielsentikaer won't have time before next week so I'll pass sorry

pernielsentikaer:
All good, I'm giving it a spin 🙂

pernielsentikaer:
Hi @aliaksandrZh 👋 

I think it makes sense to just have the preferences in extension level and not in both commands, is there a reason why you would have one command to do something different than the other one?

pernielsentikaer:
And do we even need the preferences? I think a toggle should switch on/off and maintain the current level if possible. That would make the preference redundant

aliaksandrZh:
> And do we even need the preferences? I think a toggle should switch on/off and maintain the current level if possible. That would make the preference redundant

Actually I agree that the keep input volume value by default is more convenient but there are a lot of flags in the code and I didn't want to change the core logic (May be It has some hidden sense 😅).

If you don't mind I would prefer to rid of that logic and preferences, and maintain the current level

pernielsentikaer:
I would suggest that we remove the preference and clean up the code, as it's a toggle extension

NotFiliberto:
I have a problem with it, when i toggle it with keyboard shortcut it is also triggered raycast (the search bar is showed up)

1activegeek:
Same here as reported above. Good news - it is in fact saving the level and restoring it. But it is quite annoying that the raycast window opens afterward. I'm guessing this may just be a bug someone can identify quickly. 🤞🏻

iamyeizi:
and the option I had added doesn't work as expected 🙂

CleanShot 2024-05-31 at 12  27 53

aliaksandrZh:
Will test it

aliaksandrZh:
Hi, @NotFiliberto @1activegeek @iamyeizi
Thank you for the finding bugs

Could you take a look? https://github.com/raycast/extensions/pull/12685

NotFiliberto:
Hi @aliaksandrZh, seems working fine now. Last (mini)bug is that if you try to use the hotkey when you opened the menubar toggle, it does not change, but I don't know if you can do something about that or it works like this. Take a look of the screenrec

https://github.com/raycast/extensions/assets/49571243/f6cc1647-9170-439f-a9e7-6bf5aa388b25

iamyeizi:
I've tried that too, I thought he had included it. Whit a shortcut set or the toggle button from the menu bar icon it doesn't work as expected.

PR #12713, opened 2024-06-01
Add invoice-generator extension

AlexIsMaking:
## Description

A command that generates invoices, using the [invoicegenerator.com API](https://invoice-generator.com/developers).

I'd like to include an screenshot of the invoice that's generated in the API but I'm not sure if that's allowed?

## Checklist


AlexIsMaking:
Didn't mean to publish that last commit ☺️

pernielsentikaer (changes requested):
Hi 👋 

- You can put metadata/invoice-generator-6.png in the `README` instead, but you'll need to move it to another folder called `media` instead, metadata is only for Raycast screenshots.

pernielsentikaer:
A few more questions, I hope you don't mind 🙂

- Should we consider popping to root after generating the invoice?

- It looks like there is a lot more of locales and currencies, are you planning on adding that?

- Should `Invoice Number` be remembered from last invoice? Seems like it only shows the prefix when I enters

AlexIsMaking:
> A few more questions, I hope you don't mind 🙂

No worries, always good to bounce ideas around 🙂

> * Should we consider popping to root after generating the invoice?

Done!

> * It looks like there is a lot more of locales and currencies, are you planning on adding that?

I don't think there's more locales, the docs say "English, French, German, Spanish, and Thai" are the only supported options.

I'd be happy to add more currencies though, I was partly curious to see if anyone requested others. I'm also not sure if all ISO 4217 currencies are supported or not.

> * Should `Invoice Number` be remembered from last invoice? Seems like it only shows the prefix when I enters

I did think about this but one of the nice features of this extension is that if you need to edit an invoice, it'll update the invoice in your files if you use the same invoice number so you might not want it to increment automatically.

Also, there's likely to be situations when the user wants to change the next number in the sequence and managing that seems like more hassle than it's worth.

I'd prefer to wait and see if it's requested, if that makes sense?

AlexIsMaking:
Ok so I changed my mind about pop to root while testing the new currencies. It's a lot more work to recreate the invoice from scratch if it needs changing than hitting escape to get back to the root so I think it's best not to use it.

I've added every ISO 4217 currency now.

AlexIsMaking:
@pernielsentikaer I'm not sure why the checks are failing here?

AlexIsMaking:
@pernielsentikaer that was weird, the issue's fixed itself now 🤷‍♂️

pernielsentikaer:
It seems like `defaultLocale` is not used anywhere, is that a leftover or an error?

AlexIsMaking:
@pernielsentikaer that was an error, just fixed it.

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #12717, opened 2024-06-01
[PM2] Bugfixes & enhancements

LitoMore:
## Description

- Fix API command
- Add [`raycast-pm2`](https://github.com/LitoMore/raycast-pm2) example to readme
- Bump dependencies

I created a fake toast function to bypass the exception when running `showToast` in `no-view` mode. It might be a little tricky but it will be more neat than writing a lot of let-if-else.

I've created a new library [`raycast-pm2`](https://github.com/LitoMore/raycast-pm2) for extension developers.

## Checklist


pernielsentikaer:
Should we look for the default locations and make the preferences optional instead?

pm2 2024-06-04 at 07 42 04

LitoMore:
@pernielsentikaer I don't think so. There are so many ways to for installing Node.js. Like Homebrew, nvm, fnm, installers, Docker, even source code build. We can't know where their `node` and `npm` binaries are.

Also the Raycast Preference only supports constant configurations in package.json. We can't add an option to it programmatically.

pernielsentikaer:
I was thinking about checking it by looping through the known directories in the code. If none of the default directories exist, then you can set it in preferences - doesn't that make sense?

LitoMore:
I also tried to install npm to the extension. But it doesn't work. The npm programmatic API has been deprecated, and it's missing some key functionary, like specifying the working directory.

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

Issue #12676, opened 2024-05-30
[Slack] ... is it possible to have this extension configurable to open on browser instead?


ariccb:
### Extension

https://www.raycast.com/mommertf/slack

### Raycast Version

1.75.1

### macOS Version

14.5 (23F79)

#### Error:
```
Error: Command failed with exit code 1: osascript -e 
    tell application "Slack"
      if not application "Slack" is running then
        activate
        set _maxOpenWaitTimeInSeconds to 5
        set _openCounter to 0
        repeat until application "Slack" is running
          delay 0.5
          set _openCounter to _openCounter   0.5
          if _openCounter > _maxOpenWaitTimeInSeconds then exit repeat
        end repeat

        delay 6

        # Exit 'Set yourself to active?' window
        activate
        tell application "System Events"
          key code 53
        end tell
      end if
      activate
      
      tell application "System Events" to tell process "Slack" to keystroke "A" using {command down, shift down}
    
    end tell
87:95: execution error: Can’t get application "Slack". (-1728)

    tell application "Slack"
      if not application "Slack" is running then
        activate
        set _maxOpenWaitTimeInSeconds to 5

```

I'm hoping to configure the Extension to be able to open the links in slack on the browser instead of needing the desktop application. 
It's sending a link `slack://channel?team=<team-id>/&id=<channel-id>` but i want it to open `https://app.slack.com/client/<team-id>/<channel-id>` instead

### Steps To Reproduce

works when I have slack desktop installed
doesn't work when i uninstall it

### Current Behaviour

sends me an error ```Can’t get application "Slack". (-1728)```

### Expected Behaviour

open in my default browser

jfkisafk:
@raycastbot assign me

PR #12725, opened 2024-06-02
[Slack] Adds action to open slack links in the browser

jfkisafk:
## Description

- Adds action to open in browser apart from slack app
- Only show Open in Slack option in case Slack app is installed on the system

## Screencast

Screenshot 2024-06-08 at 5 31 21 AM

## Checklist

* closes https://github.com/raycast/extensions/issues/12676

### DIFF
--- a/extensions/slack/CHANGELOG.md
+++ b/extensions/slack/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Slack Changelog
 
-## [Improvements and Fixes] - {PR_MERGE_DATE}
+## [Improvements and Fixes] - 2024-06-14
 
 - Adds action to open in browser apart from slack app
 - Only show Open in Slack option in case Slack app is installed on the system


jfkisafk:
@momme-rtf @Elliot67 looking for some traction here. If you find some time can you please review the changes?

momme-rtf:
@jfkisafk Thanks for your contribution! I saw that you're enforcing the existence of the conversationId. Could we change the logic in a way that we're opening the browser when the conversationId is defined and the openInBrowser flag is true and otherwise we open the slack app as a fallback?

extensions/slack/src/shared/utils.ts
```diff
@@ -3,10 +3,15 @@ import formatDistance from "date-fns/formatDistance";
...
+    getPreferenceValues<{ closeRightSidebar: boolean; openInBrowser: boolean }>();
 
-  open(`slack://user?team=${workspaceId}&id=${userId}`);
+  open(
+    openInBrowser
+      ? `https://app.slack.com/client/${workspaceId}/${conversationId}`
+      : `slack://user?team=${workspaceId}&id=${userId}`
+  );
   closeMainWindow();
   if (closeRightSidebar) {
```
Elliot67, line 16: With the `openInBrowser` option active, closing the sidebar focus the Slack app after opening the Slack website in a browser.

We need to check for `closeRightSidebar && !openInBrowser`.

Elliot67:
1. @momme-rtf I don't know the cases where`conversationId` can be undefined, but we should probably only open the application if the user has `openInBrowser` to false.
As a user, I would prefer to have an error rather than not understanding why the `openInBrowser` option doesn't work as expected.

2. Having this new option also impacts the expected result of  the `Open unread message` command.
The best approach would probably be to open the page in the browser, but Slack doesn't seem to have a specific link for it. It's still possible to access the page with the shortcut, but I don't know if there is an easy way to run them inside the right browser.
The second option is to disable the command when the `openInBrowser` is activated.

3. Finally, if we want the option to impact the whole extension, it would probably be clearer to use a select with `Slack Application`/`Slack in the browser` rather than a boolean.

momme-rtf:
@Elliot67 

1. Hmm.. I agree. But I'd still like to get rid of the non-null assertion operator.

2. That command needs to be revisited anyways. It works quite unreliable in the current version.  

3. 👍

extensions/slack/src/shared/utils.ts
jfkisafk: Good catch! Resolved in followup.

jfkisafk:
@momme-rtf @Elliot67 Thanks for the quick review and valuable feedback. I have changed the implementation somewhat and here are the takeaways:
1. I've left the `Open unread message` command untouched to not introduce scope creep to the issue. We can open an adjacent issue to fix that.
2. Instead of a global preference, I've opted to provide an action to `Open in Browser` in addition to `Open in Slack` (only rendered in case Slack app is installed). This will be useful to remove the issue but also for folks who prefer browser.
3. `Open in Browser` for chats only appear if conversationId is not empty (which will be the case always).

Please take a look when you find some time and let me know what you guys think!

Elliot67:
@jfkisafk Regarding point 3, `conversationId` can be undefined for a user to user chat if you never talked to each other. I believe Slack doesn't create a conversation between you and each user when you join the organization. The conversation  has to be manually created.

After some testing, I couldn't find a way to link to a user profile on the browser app.
The closest I can get is with `https://<WORKSPACE_NAME>.slack.com/team/<USER_ID>` which will ask the user to open the slack app... 😢 .

I believe the only solution is to dynamically create the conversation with the API, probably [`conversations.create`](https://api.slack.com/methods/conversations.create) or [`conversations.open`](https://api.slack.com/methods/conversations.open). But it requires new write permissions, so not sure if we really want to implement this.

---

I also found an issue in "Unread message > Open in the browser" with user to user conversation.
The provided `conversationId` isn't correct and represent the `userId` instead. The problem probably comes from `SlackClient.ts:getUnreadConversations` where the `channelId` is used as `conversationId` :
```ts
conversationId: conversationInfos[index].channel!.id!,
```

The Slack app probably automatically handle the redirect of `userId` to `conversationId`, but the browser version doesn't handle it and fall back to the previous opened channel.

---

Everything else looks good to me, creating an "Open in browser" action instead of a new command is a great idea !

jfkisafk:
@Elliot67 Thanks for the feedback.

1. You're right, we might need to update the permissions for that edge case and let's not take that scope creep as part of this issue/PR.
2. Fixed the issue by using `unreadConversation.conversationId` along with other refactors.

Please take a look when you have some time.

jfkisafk:
@momme-rtf @Elliot67 can you guys please review the latest changes when you find some time?

pernielsentikaer:
Thanks for helping with the review 🫶

jfkisafk:
Just a gentle ping for the review @momme-rtf @Elliot67 🫶

jfkisafk:
Not getting traction for re-review from @momme-rtf / @Elliot67 . I've already addressed all feedbacks for prior reviews, @pernielsentikaer can you please please take a look?

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #12735, opened 2024-06-03
Add raycast-monkeytype-theme extension

ridemountainpig:
## Description

This Raycast MonkeyType Theme extension enables users to select MonkeyType themes within the Raycast application. Users can use the MonkeyType theme in Raycast, similar to the Raycast Explorer Themes.

## Screencast

https://github.com/raycast/extensions/assets/92412722/02cea7b1-36ea-4186-b2da-0ab88aa9de8f

## Checklist


pernielsentikaer:
Hi @ridemountainpig 👋

Don't you think it would be better just to add the themes directly to our [Theme Explorer](https://github.com/raycast/theme-explorer/) so everyone can use them?

ridemountainpig:
The MonkeyType theme updates frequently, so if you add a MonkeyType theme to the Raycast Theme Explorer, you may need to update it often. This extension will fetch the API for the latest theme data, so you won't need to update the extension as often. Creating an independent extension is likely a better approach.

pernielsentikaer:
That makes sense, but it seems like many of the themes doesn't really play that well with Raycast, like `Dark` as sown below

Raycast13062024-bvuBlnIm png@2x

ridemountainpig:
Hi @pernielsentikaer,

I have added light and dark appearance options for each theme. Some themes may not work well in the dark appearance, but now they can be used in the light appearance, which will make them better. Users can also select their preferred appearance.

pernielsentikaer (changes requested):
You might want to use [environment.canAccess(AI)](https://developers.raycast.com/api-reference/environment#environment.canaccess) to check if the user has access to Theme Studio in the first place and show something else if not

ridemountainpig:
> You might want to use [environment.canAccess(AI)](https://developers.raycast.com/api-reference/environment#environment.canaccess) to check if the user has access to Theme Studio in the first place and show something else if not

When the user selects the "Add to Raycast" option, the extension will open the link to the Raycast Theme Explorer page. It will then check if the user has access to the Theme Studio. If not, it will display the upgrade to pro dialog.

CleanShot 2024-06-17 at 16 52 15@2x

pernielsentikaer:
We could proactively show an error screen directly in the extension. Don't you think it could be a nice thing?

extensions/raycast-monkeytype-theme/src/raycast-monkeytype-theme.tsx
```diff
@@ -0,0 +1,285 @@
...
+              <Action
+                title="Add Dark Appearance to Raycast"
+                icon={Icon.Moon}
+                onAction={async () => {
+                  if (!(await validateIdentity(theme.name))) return;
+                  open(theme["ray.so.add.dark.url"]);
+                  closeMainWindow();
+                }}
+                shortcut={{ modifiers: ["cmd"], key: "d" }}
+              />
```
pernielsentikaer, line 199: Since we have `appearance` in the data received, wouldn't it be better to just show either light or dark

ridemountainpig: The appearance in data is determined by the background color, but this alone doesn't always reflect the best theme. Some themes are designed to look better with a light appearance, even if they are dark. Therefore, I provide both light and dark options for users to choose the one that suits them best.

pernielsentikaer (commented):
Would it be too much to add a preference to set the default appearance to show it first in the action panel

ridemountainpig:
> Would it be too much to add a preference to set the default appearance to show it first in the action panel

Added it.

raycast-monkeytype-theme 2024-06-18 at 21 02 05

pernielsentikaer:
What about simply doing something like this:

```
const renderThemeActions = (theme: Theme) => {
    const lightAction = (
      <Action
        title="Add Light Appearance to Raycast"
        icon={Icon.Sun}
        onAction={async () => {
          if (!(await validateIdentity(theme.name))) return;
          open(theme["ray.so.add.light.url"]);
          closeMainWindow();
        }}
        shortcut={{ modifiers: ["cmd"], key: "l" }}
      />
    );

    const darkAction = (
      <Action
        title="Add Dark Appearance to Raycast"
        icon={Icon.Moon}
        onAction={async () => {
          if (!(await validateIdentity(theme.name))) return;
          open(theme["ray.so.add.dark.url"]);
          closeMainWindow();
        }}
        shortcut={{ modifiers: ["cmd"], key: "d" }}
      />
    );

    return theme.appearance === "light" ? (
      <>
        {lightAction}
        {darkAction}
      </>
    ) : (
      <>
        {darkAction}
        {lightAction}
      </>
    );
  };
```

And do something like this: `{renderThemeActions(theme)}`, this could also work for the Browser variants?

ridemountainpig:
Done. Please check. Thank you.

pernielsentikaer:
✅ Approved

PR #12834, opened 2024-06-08
Add `The Lord of the Rings` extension

xmok:
## Description

This is an extension I initially worked on in April but did not submit as there was a bug.

This is a Raycast extension to get LoTR information via [The One API v2](https://the-one-api.dev/).

## Screencast

https://github.com/raycast/extensions/assets/30526133/7a9c7620-d7e2-4c1d-a55b-95f500f5262d

## Checklist


xmok:
I kept the _id as that was the only other information available. But I agree it's not meaningful so removed.
I also removed the `BUGS` section from README as I am no longer able to reproduce the bug therein. Maybe it has something to do with latest Raycast API update.

pernielsentikaer:
Sometimes it doesn't update the section title with the number of chapters, it might also happen in other views (The toast showed the right number) 

lotr 2024-06-11 at 15 25 36

xmok:
I really appreciate the thoroughness of your reviews.

I'm unable to reproduce the issue you mentioned but to tackle it, I've added `useCachedState` which updates using a `useEffect`.

The reason we aren't using the `length` of the Array directly is because the length can be < total.

pernielsentikaer:
Try to clear the cache and visit the `Books` command

Raycast 2024-06-11 at 20 44 18

There a a few weird things, see how the list is suddently empty when going back and the section title says 0 for all of them 🙂

https://github.com/raycast/extensions/assets/38033107/30ea3fae-9594-4e86-9fe0-da8a71bdd2e7

xmok:
OK so this is the same type of bug I was facing previously which is why I didn't submit this extension earlier. Idk what the reason is but I am no longer able to reproduce any of the bugs you are showing even if I clear storage or remove and re-add the extension. There is definitely some sort of issue that has just stopped for me but is still there for you. I think I will remove the `useLOTR` hook and modify the code to just utilize individual `useFetch` and then see what happens. For reference, I am on macOS Version 13.0.1 and Raycast 1.76.0.

xmok:
https://github.com/raycast/extensions/assets/30526133/b43e4dc7-fd48-463a-b070-d517b434a4ba

In this update I'm using the `useFetch` hook itself. Instead of `useEffect` we are able to update the `useCachedState` value through the `mapResult` function. I've attached a video showing that it seems to work for me fine and I clear storage at start of video. Also note I am running the `npm run build` version although `npm run dev` is also working for me.

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #12836, opened 2024-06-08
[Espanso] Better search

kud:
## Description

- Add metadata display
- Add categories and subcategories depending on the use of ` > ` in the label of the snippet

## Screencast

CleanShot 2024-06-08 at 16 41 59@2x

## Checklist

### DIFF
--- a/extensions/espanso/src/index.tsx
+++ b/extensions/espanso/src/index.tsx
@@ -1,38 +1,75 @@
-import { Action, ActionPanel, Application, Clipboard, Detail, List, getFrontmostApplication } from "@raycast/api";
+import { Action, ActionPanel, Application, Clipboard, Detail, List, getFrontmostApplication, Icon } from "@raycast/api";
 import { useEffect, useState } from "react";
 import { ProcessOutput } from "zx";
+import { capitalCase, kebabCase } from "change-case";
+
 import { commandNotFoundMd, noContentMd } from "./content/messages";
-import { NormalizedEspansoMatch } from "./lib/types";
+
+import { FormattedMatch } from "./lib/types";
 import { getEspansoConfig, getMatches, sortMatches } from "./lib/utils";
+
+import CategoryDropdown from "./components/category-dropdown";
 
 export default function Command() {
   const [isLoading, setIsLoading] = useState(true);
-  const [items, setItems] = useState<NormalizedEspansoMatch[]>([]);
+  const [items, setItems] = useState<FormattedMatch[]>([]);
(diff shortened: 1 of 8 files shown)


kud:
I wonder tho what could be the best for categories between ` > ` in the label and having a naming convention in the file path like `[work]data.yml` could create it. What do you think @nbbaier ?

erics118:
I suggest being able to disable this category feature, with a checkbox in preferences, and also being able to customize this category divider. Eg, some people might like `/` or `-` or even `.`

extensions/espanso/src/index.tsx
```diff
@@ -25,14 +45,26 @@ export default function Command() {
...
-
         const userMatches = getMatches(matchFilesDirectory);
 
         const combinedMatches: NormalizedEspansoMatch[] = [...userMatches, ...packageMatches];
-
         const sortedMatches = sortMatches(combinedMatches);
 
+        const categoriesSet = new Set<string>();
+        sortedMatches.forEach((match) => {
+          const sectionKey = match.label?.includes(" > ") ? match.label.split(" > ")[0] : "Miscellaneous";
```
erics118, line 55: I think this should split with `>` instead, but add in a `.trim`, so categories can still work if users don't want the spaces.

Also, does this handle weird cases like if the `>` is in the beginning or end of the label?

extensions/espanso/src/index.tsx
```diff
@@ -43,32 +75,138 @@ export default function Command() {
...
+          {sections[sectionKey].map(({ triggers, replace, label, filePath }, index) => {
+            let title = triggers.join(", ");
+            let subtitle: string | undefined = undefined;
+
+            let category: string | undefined = undefined;
+            let subcategory: string | undefined = undefined;
+            let remainingLabel: string | undefined = undefined;
+
+            if (label) {
+              const parts = label.split(" > ");
```
erics118, line 139: Since the `label.split(" > ")` is duplicated quite a few times, maybe create a dedicated function for this logic?

kud:
I think I actually need to work again on that PR and use folders and not label to create categories, cf https://espanso.org/docs/matches/organizing-matches/

kud:
Okay now it's based on the filesytem, it will be better and also optionnal then.

pernielsentikaer:
Do you want to check it again @erics118 now that you gave feedback 🙂

kud:
Please approve 😈

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #12855, opened 2024-06-09
Add minesweeper extension

antoj2:
## Description

Implementation of the game Minesweeper in Raycast using a Grid and a Grid.Item for each cell, of which there are 64. It has all the bells and whistles of the classic game including, flood filling and ability to open empty cells around an already opened cell.

The controls are in the README.

## Screencast

https://github.com/raycast/extensions/assets/85374170/cbbf53cc-d0d8-45ca-a237-554ce5a335c5

## Checklist

### DIFF
--- a/extensions/minesweeper/package-lock.json
+++ b/extensions/minesweeper/package-lock.json
@@ -706,12 +706,13 @@
       }
     },
     "node_modules/braces": {
-      "version": "3.0.2",
-      "resolved": "https://registry.npmjs.org/braces/-/braces-3.0.2.tgz",
-      "integrity": "sha512-b8um+L1RzM3WDSzvhm6gIz1yfTbBt6YTlcEKAvsmqCZZFw46z626lVj9j1yEPW33H5H+lBQpZMP1k8l+78Ha0A==",
+      "version": "3.0.3",
+      "resolved": "https://registry.npmjs.org/braces/-/braces-3.0.3.tgz",
+      "integrity": "sha512-yQbXgO/OSZVD2IsiLlro+7Hf6Q18EJrKSEsdoMzKePKXct3gvD8oLcOQdIzGupr5Fj+EDe8gO/lxc1BzfMpxvA==",
       "dev": true,
+      "license": "MIT",
       "dependencies": {
-        "fill-range": "^7.0.1"
+        "fill-range": "^7.1.1"
       },
       "engines": {
         "node": ">=8"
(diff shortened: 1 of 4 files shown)


pernielsentikaer (changes requested):
Hi 👋 

- I would find the icons from 1-8 as `svg` images so you can use `tintColor: Color.PrimaryText` instead, since it's pretty hard to see the dark colors with dark appereance and using `svg` should also put the content in the middle

minesweeper 2024-06-17 at 13 03 49

- When you git a mine it would be nice to see where the mines actually is, right now it's impossible to see anything

minesweeper 2024-06-17 at 13 05 00

- If you enters something in the search bar and clears a mine, then it should you "No Results" and it's not possible to get back

minesweeper 2024-06-17 at 13 07 50
``

gfdsyjkdyhkfsydkh:
Thanks for the review and your feedback! Finally got a chance to look at this 😅. I tried my best at your suggestions and think I got it down. Lmk if the delay between the red cells and showing the bombs is sufficient. 
As for the your last point, I just removed the `filtering={false}` from the `Grid` element and let Raycast's filtering go. Still returns 0 results but deleting the search brings it back.

pernielsentikaer:
That's a lot better, awesome!

- Could we add a `Restart Game` option when you're losing or winning?
- Add `filtering={false}` to the `<Grid>` element on line 50, which should disable all filtering

gfdsyjkdyhkfsydkh:
Did not mean to push this so far back, that's my bad, got a bit busy with some personal problems. 
Just pushed a change to get the new game action on the win/lose screen. The `filtering={false}` is what caused it to be filtered when pressing enter, I believe. The extension [Hidden Icons](https://www.raycast.com/koinzhang/hidden-icons) appears to have the same problem.

gfdsyjkdyhkfsydkh:
Just tested it with this, and yeah it's the same. I'm guessing it has something to do with the way the filtering prop works, cause I don't see why it would be like this. I can attest that it does the same behavior on a List element. 
```jsx
  const [items, setItems] = useState<Array<ReactNode>>([]);
  const [counter, setCounter] = useState(1);

  const action = (
    <ActionPanel>
      <Action
        title="Add new item to the list"
        onAction={() => {
          setCounter(counter + 1);
          items.push(<Grid.Item key={counter} content={""} title="Newly added cell" />);
          setItems(items);
        }}
      />
    </ActionPanel>
  );

  return (
    <Grid filtering={false}>
      <Grid.Item key={0} content={""} title="Test item" actions={action} />
      {items.map((item) => {
        return item;
      })}
    </Grid>
  );
}
```

pernielsentikaer:
In the above example, just to be sure, what is the error?

Issue #12871, opened 2024-06-10
[Raycast Wallpaper] Change automatic refresh time


Faria22:
### Extension

https://www.raycast.com/koinzhang/raycast-wallpaper

### Description

Change the automatic refresh time of the wallpapers from the default 1 hour. It would be nice to be able to change it to any time step, but just a few options is already good enough.

PR #12896, opened 2024-06-12
[Raycast Wallpaper] Preference to specify refresh intervals

jfkisafk:
## Description

- Preference to set different refresh intervals mirroring Apple wallpaper shuffle intervals
- Updates dependencies and fixed tsconfig

## Screencast

Screenshot 2024-06-11 at 11 22 24 PM

## Checklist

* closes https://github.com/raycast/extensions/issues/12871

### DIFF
--- a/extensions/raycast-wallpaper/CHANGELOG.md
+++ b/extensions/raycast-wallpaper/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Raycast Wallpaper Changelog
 
-## [Auto Switch Wallpaper command] - {PR_MERGE_DATE}
+## [Auto Switch Wallpaper command] - 2024-06-12
 
 - Preference for refresh intervals mirroring Apple wallpaper shuffle intervals
 


koinzhang:
Hi @jfkisafk  Thanks for your contribution, I tested your changes and have two questions so far
1. The first one is more serious, now if Raycast run Auto Switch command automatically, it will close the main window, if I set the interval to a shorter one, say 1m, if I open the main window of Raycast at that time, it will hide the main window very fast!
This problem has already affected the use of Raycast and needs to be optimised, the optimisation solution is also very simple, remove `await closeMainWindow();` or put it in 
```
if (environment.launchType === LaunchType.UserInitiated) {
    await showHUD("🖥️ Setting wallpaper...") ;)
}
```

2. The second question is, isn't it too fast to set the minimum refresh interval of the extension to 1m, and will it affect the performance of Raycast? I hope the Raycast developers @pernielsentikaer will evaluate this issue.
image

jfkisafk:
Hey @koinzhang thanks for the quick review. Here is what I think:
1. Nice catch, I am making the changes in the next iteration.
2. I had some concerns with `1m` time too but found a lot of usages across different extensions ([src](https://github.com/search?q=repo%3Araycast%2Fextensions+%221m%22+path%3Apackage.json&type=code&p=1)). To be safe, I can revert back to using 5m as the least denomination.

pernielsentikaer:
Sounds like a good idea to start with 5 and see how it goes and if people is requesting lower 🙂

jfkisafk:
For sure, I have made the requested changes @koinzhang @pernielsentikaer

PR #12927, opened 2024-06-13
fix(codeowners): remove duplicate entries wrongly added by the bot

jfkisafk:
## Description

I added @srikirank as contributor b/c I got a ton of feedback developing the extensions/commands. The both put my alias in CODEOWNERS and raycast2github mapping, which will cause issues if @srikirank makes contributions to the extensions. I know these files shouldn't be changed manually, but this is a one time mistake and I wasn't aware adding multi-contributors in the same change would do this 😞

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -35,10 +35,10 @@
 /extensions/algolia                                                 @valentindotxyz
 /extensions/alice-ai                                                @wojciechkrol
 /extensions/aliyun-oss                                              @yangxyo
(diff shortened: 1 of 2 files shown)


jfkisafk:
@pernielsentikaer do you think we can make this fix or leave these files as is?

pernielsentikaer:
Should be addressed here: https://github.com/raycast/extensions/commit/6ce49704c04cca9486b85cb441fa4cd1a8bcb834

PR #12929, opened 2024-06-13
Update chatgpt extension

aeither:
## Description

As per @pernielsentikaer suggestion to integrate OpenAI Vision into ChatGPT extension

Original Extension
https://github.com/raycast/extensions/pull/12710

## Checklist


pernielsentikaer (changes requested):
It works as expected for me @aeither 🔥 

Do you think we should improve this error message when you're not selecting an image?

chatgpt 2024-06-18 at 14 23 48

aeither:
> It works as expected for me @aeither 🔥
> 
> Do you think we should improve this error message when you're not selecting an image?
> 
> chatgpt 2024-06-18 at 14 23 48
> 
> And also merge the conflicts 🙂

Updated

extensions/chatgpt/src/views/vision.tsx
```diff
@@ -0,0 +1,182 @@
...
+        <Detail.Metadata>
+          <Detail.Metadata.Separator />
+          {imageMeta.size && <Detail.Metadata.Label title="Size" text={toUnit(imageMeta.size)} />}
+          {imageMeta.width && <Detail.Metadata.Label title="Width" text={String(imageMeta.width)} />}
+          {imageMeta.height && <Detail.Metadata.Label title="Height" text={String(imageMeta.height)} />}
+
+          <Detail.Metadata.Separator />
+          <Detail.Metadata.Label title="Culmulative Tokens" text={cumulative_tokens.toString()} />
+          <Detail.Metadata.Label title="Culmulative Cost" text={cumulative_cost.toFixed(4) + " ¢"} />
+        </Detail.Metadata>
```
pernielsentikaer, line 178: Can we do something here so it doesn't look weird when there is an error?

chatgpt 2024-06-20 at 11 45 36

Could we make `Culmulative token/cost` a preference? Would that make sense at all? I can't see that information anywhere else in this extension 🙂

aeither: push -> hid metadata when error. is that ok?
I used only gpt-4o for vision as other models like gpt-3 (chatgpt default model) doesn't support.
cumulative token are only calculated on gpt-4o price.
`  switch (model) {
    case "gpt-4o":
      // The input costs are $5, while the output is $15.
      price = ((prompt_token * 5) / 1_000_000 + (output_token * 15) / 1_000_000) * 100;
      break;
    default:
      break;
  }`
 because users can add any models to it it can't estimate other models.

alternatively I can just remove the whole metadata thing

Issue #12931, opened 2024-06-13
[Mite] Add comment field


bytepoets-cha:
### Extension

https://www.raycast.com/paolo_ricciuti/mite

### Description

Hi Paolo, I wanted to let you know that I really appreciate your extension! However, there's one feature that would be incredibly helpful: the display of the comment field. For my tasks, it's essential to include a ticket number with each entry, and without the ability to add comments, it's quite challenging to track my bookings accurately with Raycast.

Additionally, it would be amazing if it would be possible to start and stop time bookings in the same way that it’s done in the mite interface and not just by entering a certain amount of time. These features would streamline my process significantly.

Thank you for your great work, and I hope you consider these suggestions for future updates.

### Who will benefit from this feature?

Anyone booking on certain tickets or just going along with the task flow

ramsaysewell:
Hey,

Thanks for the feature request. I'm not a user nor contributor to this extension but as it was simply adding an extra field, I thought I'd help out.

I've setup a PR that adds in a "note" field for when you're entering tasks.

As I'm not a user of Mite, I'm unaware of what the expected behaviour for stopping and starting the trackers. Perhaps the author of the extension is more familiar with that.

paoloricciuti:
>Hi Paolo, I wanted to let you know that I really appreciate your extension! However, there's one feature that would be incredibly helpful: the display of the comment field. For my tasks, it's essential to include a ticket number with each entry, and without the ability to add comments, it's quite challenging to track my bookings accurately with Raycast.

I will take a look at @ramsaysewell PR as soon as i get back from my self imposed stop from OSS cause of burnout but it will likely be next week

>Additionally, it would be amazing if it would be possible to start and stop time bookings in the same way that it’s done in the mite interface and not just by entering a certain amount of time. These features would streamline my process significantly.

Sure...tbf i really dislike howì the mite interface start and stop a timer when you leave the hrs at 0 but i think we can have a command to start the timer. Would that fit well into your workflow?

ramsaysewell:
Merged! 🎉 

This feature is now available so I'll close the issue. If you'd like to request anything else or find an issue with the extension then you'll need to create a new issue 😄

@raycastbot close this issue

PR #12934, opened 2024-06-13
Update mite extension

ramsaysewell:
## Checklist

### DIFF
--- a/extensions/mite/src/time-entry.tsx
+++ b/extensions/mite/src/time-entry.tsx
@@ -12,26 +12,26 @@ export default function Command() {
       const body = parse(time_entry_post_schema, { time_entry });
       const res = await fetch_mite("/time_entries.json", {
         method: "POST",
-        body: JSON.stringify(body)
+        body: JSON.stringify(body),
       });
       if (res.ok) {
         showToast({ title: "Entry created", style: Toast.Style.Success });
 
         popToRoot({
-          clearSearchBar: false
+          clearSearchBar: false,
         });
         return;
       }
       showToast({
         title: "Error",
         message: "There was an error in the submission",
-        style: Toast.Style.Failure
+        style: Toast.Style.Failure,
       });
     } catch (e) {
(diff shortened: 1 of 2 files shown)


ramsaysewell:
Fixes part of this issue #12931

paoloricciuti:
Hey @ramsaysewell thanks for this...i think this is missing something but i might be wrong (i can't actually remember how the extension handle this).

Right now i'm in a self imposed stop from OSS because i kinda burned out but i will likely look at this next week 😄

extensions/mite/src/time-entry.tsx
```diff
@@ -16,6 +16,11 @@ export default function Command() {
       });
       if (res.ok) {
         showToast({ title: "Entry created", style: Toast.Style.Success });
+        setTimeout(() => {
+          popToRoot({
+            clearSearchBar: false,
+          });
+        }, 1000);
```
pernielsentikaer, line 23: Is it necessary to use `setTimeout` here?

ramsaysewell: hey @pernielsentikaer, I've removed this now in the latest commit. This was me following the UX flow found in `setup-fast-time-entry`. I've also removed the setTimeout from this too as I believe it's not necessary either.

paoloricciuti: I've added the `setTimeout` to allow the user to see the form being submitted...i didn't like to be immediately moved outside without actually have any confirmation that the form was submitted.

pernielsentikaer: I think it makes sense to remove so we're using the same flow as the rest of Raycast 🙂

mil3na:
Hi! Is this ready for another review?

paoloricciuti:
> Hi! Is this ready for another review?

Sorry completely forgot about that...good for me, we can merge!

mil3na (approved):

Issue #12986, opened 2024-06-17
[Messages] add option to close Raycast window after message sends


octokatherine:
### Extension

https://www.raycast.com/thomaslombart/messages

### Description

After sending a message, I have to escape out of the Raycast window. It would be nice if the window closed after sending, or if there was an option to have it behave like this.

thomaslombart:
Thanks for the feature request. Would you like to work on it? This feature shouldn't be too hard to implement.

ramsaysewell:
I'd like to work on this as it could be quite fun to implement.

thomaslombart:
I'm working on it at the moment.

PR #12996, opened 2024-06-17
[Messages] Added a preference to close Raycast when sending a message

ramsaysewell:
## Checklist

### DIFF
--- a/extensions/messages/CHANGELOG.md
+++ b/extensions/messages/CHANGELOG.md
@@ -1,14 +1,13 @@
 # Messages Changelog
 
-## [Successful message action] - 2024-6-17
+## [Improvements] - 2024-07-03
 
-- gives the user the option to choosing between a Toast and a HUD notification when a message is sent successfully.
-- fixes issue where "empty" contacts were being rendered in the contact list dropdown
+- Added a preference allowing users to close Raycast when sending a message.
+- Fixed an issue where "empty" contacts were appearing in the contact list dropdown.
 
 ## [New address field] - 2024-3-28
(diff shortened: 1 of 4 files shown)


ramsaysewell:
This PR adds feature requested here:  #12986

thomaslombart (changes requested):
Thanks for the PR. The preference feels confusing right now because it's not clear what it's doing. Why not having a checkbox labelled `Close main window`? Something like this:

```
        {
          "name": "shouldCloseMainWindow",
          "type": "checkbox",
          "label": "Close window immediately",
          "description": "When enabled, the Raycast window is closed immediately, allowing you to carry on with other work.",
          "default": true,
          "required": false
        }
```

Then, you can call `closeMainWindow` right before `showToast`.

extensions/messages/src/send-message.tsx
```diff
@@ -67,6 +83,10 @@ export default function Command({
       if (result === "Success") {
         const name = getName(correspondingContact);
 
+        if (shouldCloseMainWindow) {
+          await showHUD(`Sent Message to ${name}`, { clearRootSearch: true });
+        }
```
thomaslombart, line 88: You can actually call `closeMainWindow` here instead of calling `showHUD`, the toast will still be shown as a HUD.

thomaslombart:
Hey there! I'll be on vacation next week, so I'll review this PR the following week. @pernielsentikaer, feel free to handle this one if you have time.

thomaslombart (approved):

PR #13002, opened 2024-06-17
Remove Designer News extension

LitoMore:
This extension is dead.

fyi: https://solomon.io/designer-news-is-dead/


pernielsentikaer:
I think we first need to do something like it's done with https://www.raycast.com/Dammmien/associated-press - are you up for that or should I?

mil3na (approved):
LGTM!

Issue #11628, opened 2024-04-04
[One Time Password] Reorder menu items


SoulEvans07:
### Extension

https://www.raycast.com/lachero/one-time-password

### Description

Please change the order of items so that the "+ Create new" goes to the bottom instead of the top. 
Users use the OTPs way more than they are adding new ones so it's better to have those options at the top.
![image](https://github.com/raycast/extensions/assets/4013509/82af049f-68f8-4e50-b0b8-ff2515dc5fd1)

And if possible add a custom order or reorder by usage option in the settings. 
The ultimate goal would be to be able open Raycast, press the command alias, Enter, Enter (or CMD+Enter, #11627) and the last used OTP is already in your clipboard (or in the desired input)

### Who will benefit from this feature?

All users use the OTPs more than they are adding new ones so everybody should benefit from this. 
And if the custom order or reorder feature is done as well then you'll benefit from a significantly quicker and easier usage of the extension

lacherogwu:
Thanks for your suggestion, I will take a look into it.

SoulEvans07:
I realised that raycast uses React so I can just implement these so I made a PR about it #13003

Issue #11627, opened 2024-04-04
[One Time Password] Quick insert


SoulEvans07:
### Extension

https://www.raycast.com/lachero/one-time-password

### Description

Add a shortcut (cmd+enter) which automatically inserts the value in the last focused input instead of adding it to the clipboard.

### Who will benefit from this feature?

* quicker and easier usage
* everybody who don't want to change their clipboard
* clipboard readers can't access it anymore
* there is no more accidentally pasted keys

lacherogwu:
Can you elaborate? I'm not sure I got it

SoulEvans07:
Sure. So the current behavior is something like this:
1. press cmd + space (Open Raycast)
2. press o + enter (my alias for the extension)
3. select the OTP you want to insert
7. press enter to copy
8. select the input where the OTP is needed
9. press ctr+v to insert

And the feature would work something like this:
**1. open a login form and put your cursor in the input where the OTP is needed**
2. press cmd + space (Open Raycast)
3. press o + enter (my alias for the extension)
4. select the OTP you want to insert
**5. press a shortcut (my suggestion is cmd+enter but that is used now for other stuff so shift+enter is probably better) to insert**
**6. the OTP is inserted into the previously focused input**

The idea came from the Authy OTP extension but their desktop app is discontinued and it had some network request when you open it in raycast which is now blocked by my company so I can't use it anymore.

I'm guessing that the way they implemented it is to simulate keyboard presses to avoid touching the clipboard.

The speed gain while marginal still felt easier and your clipboard is safe in 2 sense. Your OTP was never there so attackers can't access it that way and your workflow is not interrupted when let's say you had something on your clipboard but the page you wanted to insert it showed a login prompt.

lacherogwu:
Got it, will take a look into it

SoulEvans07:
I realised that raycast uses React so I can just implement these so I made a PR about it #13003

PR #13003, opened 2024-06-17
Update one-time-password extension

SoulEvans07:
### Added

- #11627
  - `CMD + enter` now will paste the OTP into the last focused input
- Sort by priority
  - ability to set a priority for accounts
  - accounts are sorted by this
  - higher prio is higher in the list
- Key Icon before accounts

### Changed

- #11628
  - create new option was moved to the bottom of the list
- Shortcut: Edit Account
  - changed to `CMD + E` to avoid conflict with Paste
- Shortcut: Remove account
  - warning caused by conflict with Raycast
  - changed to `CMD + Backspace`

### Fixed

- Shortcut: Enter a Setup Key
  - warning caused by conflict with Raycast
  - still defaults to `CMD + Enter`

### New look
image

### Priority labels
- chevron up - positive priority
- chevron down - negative priority
image

### Priority in Edit
image

### Changed shortcuts
image

## Checklist

### DIFF
--- a/extensions/one-time-password/CHANGELOG.md
+++ b/extensions/one-time-password/CHANGELOG.md
@@ -1,6 +1,6 @@
 # One Time Password Changelog
 
-## {PR_MERGE_DATE}
+## 2024-07-25
 
 ### Added
 


lacherogwu:
Fantastic work! One concern I have is about the primitive way of reordering menu items using priority. I think it would be better to do the same as Raycast did with managing fallback commands.

image

extensions/one-time-password/src/config.ts
```diff
@@ -0,0 +1,15 @@
+const theme = {
+  grey: '#909090',
+  green: '#288056',
+  red: '#802828',
```
pernielsentikaer, line 4: What about using `Color.Red`, `Color.Green`, `Color.SecondaryText` instead here

SoulEvans07: green and red are not used right now but for all 3 the idea was to use something less bright no not grab too much attention.

SoulEvans07:
That indeed looks like an interesting solution. I'm not sure when yet but I will add that as well.

lacherogwu:
@pernielsentikaer, I would like to hear your opinion about the ability to add new secrets and scan QR codes.

CleanShot 2024-06-21 at 11 20 49@2x

Moving the option down could hide it from view if someone has many codes. They would then have to go to the bottom or search for "Create new" to add a new one.

Do you have a more sophisticated solution for this? What would you recommend?

extensions/one-time-password/src/config.ts
pernielsentikaer: It is possible to create your own themes in Raycast PRO. If you don't consider this, it might conflict with the theme

pernielsentikaer:
I think it's fine to keep it as the first item in the list

lacherogwu:
@pernielsentikaer, what are your thoughts on removing this and having it in a separate actions section instead?

pernielsentikaer:
That could also work, or simply moving the action to the action panel could be a way? Users are used to looking in the action panel

lacherogwu:
@pernielsentikaer, that's precisely what I meant.

pernielsentikaer:
That's how Raycast in general works, so it should be fine 🙂

lacherogwu:
@mil3na not yet, we've suggestion some changes in the comments

extensions/one-time-password/package.json
```diff
@@ -5,6 +5,10 @@
   "description": "Generate time-based one-time passwords",
   "icon": "icon.png",
   "author": "lachero",
+  "version": "1.1.0",
```
LitoMore, line 8: The `version` field is unnecessary and nonsense for Raycast extension. Let's remove it.

extensions/one-time-password/src/one-time-password.tsx
```diff
@@ -16,10 +16,19 @@ import {
 } from '@raycast/api';
 import { useEffect, useState } from 'react';
 import { getProgressIcon } from '@raycast/utils';
+import { z } from 'zod';
```
LitoMore, line 19: I think we don't have a strong need for Zod for the current plugin, and adding Zod will only make the package size even bloated.

LitoMore (changes requested):

extensions/one-time-password/package.json
SoulEvans07: then how do I refer to the update in the changelog?

LitoMore: Just update the changelog.

extensions/one-time-password/src/config.ts
SoulEvans07: I see. I removed the custom colors

extensions/one-time-password/package.json
SoulEvans07: sure

pernielsentikaer:
Is everything ready now @SoulEvans07 / @lacherogwu

SoulEvans07:
From my side, yes. I did everything mention other than removing zod. If I don't forget I'll provide new screenshot today.

pernielsentikaer:
Do you have anything else @lacherogwu  / @LitoMore

LitoMore (approved):

mil3na (approved):
LGTM!

PR #13021, opened 2024-06-18
Add telegram-raycast extension

noxlovette:
## Checklist

### DIFF
--- a/extensions/telegram-raycast/.gitignore
+++ b/extensions/telegram-raycast/.gitignore
@@ -1,0 +1,8 @@
+# dependencies
+/node_modules
+
+# misc
+.DS_Store
+.idea/
+raycast-env.d.ts
+/.vscode/
--- a/extensions/telegram-raycast/src/telegram-login.tsx
+++ b/extensions/telegram-raycast/src/telegram-login.tsx
@@ -15,6 +15,8 @@ import { TelegramClient, Api } from "telegram";
 import { StringSession } from "telegram/sessions";
 import got from "got";
 
+const URL = "https://telray-proxy.up.railway.app";
+
 interface Contact {
   id: string;
   firstName: string;
@@ -69,7 +71,7 @@ function RequestPhoneCode({
     });
 
     try {
-      const { body } = await got.post("https://telray-proxy.up.railway.app/send-code", {
+      const { body } = await got.post(`${URL}/send-code`, {
         //@ts-expect-error refactoring this code according to types breaks it
         json: {
           phoneNumber: phoneNumber,
@@ -116,7 +118,7 @@ function SendPhoneCode() {
     });
 
     try {
-      const { body } = await got.post("https://telray-proxy.up.railway.app/start-client", {
+      const { body } = await got.post(`${URL}/start-client`, {
         // @ts-expect-error refactoring this code according to types breaks it
         json: {
           phoneNumber: await LocalStorage.getItem("phone"),
--- a/extensions/telegram-raycast/src/telegram-send-message.tsx
+++ b/extensions/telegram-raycast/src/telegram-send-message.tsx
@@ -13,6 +13,7 @@ import {
 import { useCachedPromise } from "@raycast/utils";
 import { TelegramClient, Api } from "telegram";
 import { StringSession } from "telegram/sessions";
+import { useEffect } from "react";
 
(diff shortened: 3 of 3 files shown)


noxlovette:
Hello, Raycast Team! I am excited to be here.

1. I did try to adhere to the requirements specified in the UI guidelines. I wish logging into Telegram was possible simply via login/password. The phone number has to be sent to TG servers, and the user then has to submit the code on their own.
2. I also tried to define a Telegram Bot that would yield the session string directly to the chat with the user. However, Telgram blocks a connection attempt when it spots that the login code has been sent elsewhere.
3. The proxy server I have created does not store any user data. The session identifier is stored in Raycast only.
4. I chose to use proxy because a client can only be created using unique API_ID and API_HASH, secret values issues at my.telegram.org, and the only thing the proxy does is enrich received requests with these values. I have seen another pull request with Telegram that would require the user to create their own pair of these secret values – this is not optimal in my view.

koinzhang:
Add telegram extension https://github.com/raycast/extensions/pull/12477
image

noxlovette:
> Add telegram extension #12477 image

Hello! Yes, I am aware of this extension. His code relies on identifiers available only through a series of additional steps that my code does not require, and I would have to suggest rewriting his code completely to make the experience more user-friendly, which would be strange considering that the extension has not been published and is, too, under review.

I invite you to study my logic to logging into Telegram or even try for yourself to understand the difference.

The linted code is forthcoming, sorry about overseeing that.

pernielsentikaer:
It seems like both are trying to do the same job but using two different approaches, is that correct?

Maybe you should work with @SurajMeena so we end up with only one extension

SurajMeena:
> It seems like both are trying to do the same job but using two different approaches, is that correct?
> 
> Maybe you should work with @SurajMeena so we end up with only one extension

Hi @noxlovette I am up for collaborating on this extension, maybe we could connect once and discuss how the implementation and features should look like

noxlovette:
> > It seems like both are trying to do the same job but using two different approaches, is that correct?
> > Maybe you should work with @SurajMeena so we end up with only one extension
> 
> Hi @noxlovette I am up for collaborating on this extension, maybe we could connect once and discuss how the implementation and features should look like

That sounds great to me! Could you just text me on Telegram, please? https//:t.me/noxlovette? I can't wait to get our mutual extension up and running.

pernielsentikaer:
That's so nice to see, let's get the best approaches from them both and unite 🔥

SurajMeena:
> That's so nice to see, let's get the best approaches from them both and unite 🔥

Hi @pernielsentikaer , We needed your help in figuring out whether the proxy approach used in this PR aligns with raycast's standards and requirements for an extension to get published, So that we can make an informed choice for what things needs rearranging for final telegram extension

noxlovette:
> That's so nice to see, let's get the best approaches from them both and unite 🔥

Hello @pernielsentikaer! 

We have had a very productive day today and are excited to announce that we would like to see this PR request shipped to Raycast with both our names as collaborators.

Please let us know if using proxy is not a done thing, as previously asked by Suraj. It was my idea, I saw the Notion extension handling authentication similarly. The server code can be found here: https://github.com/noxlovette/telegram-proxy. It *does not store any user data,* only sends a sessionString, a unique, encrypted session identifier, to the extension over HTTPS, where it is stored in localStorage.
 
Do let us know if there are things we can further optimise.

pernielsentikaer:
If a proxy is needed, we want to host it ourselves to maintain control of it. Otherwise, we don't know what's going on. We don't want to deploy a Bun app. Can it be rewritten with Node.js and Fastify or something similar? 😊

noxlovette:
> If a proxy is needed, we want to host it ourselves to maintain control of it. Otherwise, we don't know what's going on. We don't want to deploy a Bun app. Can it be rewritten with Node.js and Fastify or something similar? 😊

Will be done. 🤝

noxlovette:
@pernielsentikaer, I have added a branch called 'express' to the proxy code. Hope this works.

The extension code needs to know how to reach the server. Could you share the URL with me somehow so that I could change the variable in the telegram-login file? Also, that would mean setting up two .env variables – API_ID and API_HASH. I guess I could share mine over Doppler?

PR #13036, opened 2024-06-19
Add natural-language-discord-timestamps extension

FarazPatankar:
## Description

The extension allows users to enter a date and/or time in natural language and responds with 3 different variations of Discord timestamps which when entered in Discord create a dynamic message. The dynamic message makes it so that the viewer sees the time in their own timezone based on the format selected by the user.

## Screencast

A demo I had shared on Twitter: https://x.com/farazpatankar/status/1802322814642393139

## Checklist

### DIFF
--- a/extensions/natural-language-discord-timestamps/src/index.tsx
+++ b/extensions/natural-language-discord-timestamps/src/index.tsx
@@ -42,11 +42,9 @@ export default function Command() {
 
   return (
     <List onSearchTextChange={setInput} searchBarPlaceholder="Enter a date and/or time">
-      <List.Section title="Options" subtitle={options.length.toString()}>
-        {options.map((option) => (
-          <ListItem key={option.displayText} option={option} />
-        ))}
-      </List.Section>
+      {options.map((option) => (
+        <ListItem key={option.displayText} option={option} />
+      ))}
     </List>
   );
 }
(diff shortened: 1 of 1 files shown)


pernielsentikaer:
Hi 👋

How is this compared to [Discord Timestamps](https://www.raycast.com/theshadowdev/discord-timestamps) by @thewilloftheshadow, it looks like they serve the same purpose but might be a slighty different approach 🙂

FarazPatankar:
Hey, thanks for the review. 

I tried the other one first and felt like it does a bit too much and iirc, you need to open the command action before you can enter natural language. 

Comparatively, the one I've created allows you to do that from the get go + is restricted to a single functionality!

pernielsentikaer:
Hi 👋

That part could be revised; perhaps it could be combined?

If we added optional arguments, they would be used if set. If not, the script would check if the selected text is a valid date and use that. Failing that, it would check the clipboard for the same thing and fallback to nothing 🙂

It would be best to have just one extension for this purpose, as it would reduce confusion among end users by preventing multiple extensions with the same function.

FarazPatankar:
Hmm, I understand. Sadly, I don't have the bandwidth/cycles to update the other extension. I basically built this at the airport on the way to my vacation which I am now back from.

My teammates would still like to use this, IIRC there's an option to publish this only to my team, correct?

pernielsentikaer:
That's correct, you can always add it to your team store, you simply need a few more fields before publishing it

```
  "author": "thomas",
  "owner": "raycast",
  "access": "private",
```

PR #13039, opened 2024-06-19
[Slack] Add Search Messages command

thomaslombart:
## Description

Add a new command to search through your Slack's workspace messages.

Also rename `Search` to `Open Channel` for consistency.

## Checklist

### DIFF
--- a/extensions/slack/CHANGELOG.md
+++ b/extensions/slack/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Slack Changelog
 
-## [Search Messages command] - 2024-06-21
+## [Search Messages command] - 2024-07-08
 
 - Add a new command to search through your Slack's workspace messages.
 - ⚠️ The previous `Search` command is renamed as `Open Channel` to differentiate from `Search Messages`


thomaslombart:
@jfkisafk You can already have a look if you want 🙂

extensions/slack/package.json
```diff
@@ -18,11 +18,21 @@
   "commands": [
     {
       "name": "search",
-      "title": "Search",
+      "title": "Switch to Channel",
```
jfkisafk, line 21: nit (non-blocking): since this includes users conversations, should we rename it to `Switch to Conversation` or `Open Conversation`?

extensions/slack/src/search-messages.tsx
```diff
@@ -0,0 +1,125 @@
...
+      const webClient = getSlackWebClient();
+      const results = await webClient.search.messages({
+        query: `${query}${selectedChannel ? ` in:${channel}` : ""}`,
+        sort,
+      });
+      return results.messages?.matches ?? [];
+    },
+    [query, selectedChannel, sortType],
+    {
+      onError(error) {
```
jfkisafk, line 32: thought (non-blocking): I am wondering if there is an error, should we render a `List.Empty` and make the error message more visible in addition to the toast?

extensions/slack/src/search-messages.tsx
```diff
@@ -0,0 +1,125 @@
...
+      isLoading={isLoading}
+      searchText={query}
+      onSearchTextChange={setQuery}
+      actions={
+        <ActionPanel>
+          <FromMeOnlyAction />
+        </ActionPanel>
+      }
+      throttle
+      isShowingDetail
```
jfkisafk, line 65: question (non-blocking): only when `data?.length > 0` right?

extensions/slack/src/search-messages.tsx
```diff
@@ -0,0 +1,125 @@
...
+        const date = convertTimestampToDate(m.ts);
+        const text = emoji.emojify(m.text);
+
+        return (
+          <List.Item
+            key={m.iid}
+            icon={{ value: user?.icon, tooltip: user?.name ?? "Unknown user" }}
+            title={text}
+            accessories={[{ date, tooltip: format(date, "EEEE dd MMMM yyyy 'at' HH:mm") }]}
+            detail={<List.Item.Detail markdown={text} />}
```
jfkisafk, line 92: question (non-blocking): should we add extra metadata to disambiguate messages from different channels when the search query is just `from:me`?

extensions/slack/src/search-messages.tsx
```diff
@@ -0,0 +1,125 @@
...
+                  <ActionPanel.Submenu
+                    title="Sort By"
+                    icon={Icon.ArrowUp}
+                    shortcut={{ modifiers: ["cmd", "shift"], key: "s" }}
+                  >
+                    {sortOptions.map((s) => (
+                      <Action
+                        key={s.sort}
+                        autoFocus={sortType === s.sort}
+                        icon={sortType === s.sort ? { source: Icon.CheckCircle, tintColor: Color.Green } : undefined}
```
jfkisafk, line 107: thought (non-blocking): I think for the unselected sortType we can use `Icon.Circle` to keep the UX consistent

extensions/slack/src/search-messages.tsx
```diff
@@ -0,0 +1,125 @@
...
+      }
+      throttle
+      isShowingDetail
+      searchBarAccessory={
+        <List.Dropdown tooltip="Search Channels" onChange={setSelectedChannel} storeValue>
+          <List.Dropdown.Item value="" title="All Channels" />
+
+          <List.Dropdown.Section>
+            {channels?.flat().map((c) => {
+              return <List.Dropdown.Item key={c.id} icon={c.icon} value={c.name} title={c.name} />;
```
jfkisafk, line 72: issue (blocking): I am not able to look up `from:me` queries when I select Users conversations from the dropdown. For channels it works fine. Are you experiencing something similar?

jfkisafk (commented):
Tested the permission update model with both `xoxp-` token and new OAuth and they are buttery smooth xp. Just left some minor comments and one blocking one (not sure if only I am experiencing it though).

extensions/slack/package.json
thomaslombart: Even Slack has a confusing name for that. On one hand, they call it `Switch to Channel` in the menu bar option:

CleanShot 2024-06-20 at 10 16 14@2x

But on the other hand, they call it `Search {Workspace}` in the UI 😄

CleanShot 2024-06-20 at 10 17 57@2x

Aren't user conversations channels as well? This could be named `Open Channel`?

extensions/slack/src/search-messages.tsx
thomaslombart: I'm not sure that's worth it here because an error could happen for subsequent searches and in this case we'd want to keep the previous search results.

extensions/slack/src/search-messages.tsx
thomaslombart: Indeed, nice catch!

extensions/slack/src/search-messages.tsx
thomaslombart: I'm not sure I got what you mean here? Which metadata should we add?

extensions/slack/src/search-messages.tsx
jfkisafk: You're right, I didn't think of that edge-case. Please feel free to resolve.

extensions/slack/package.json
jfkisafk: I was getting confused in between Users/Groups/Channels as well, and that's why thought `Conversations` could be a good comprehensive name. But `Open Channel` works too.

extensions/slack/src/search-messages.tsx
thomaslombart: Oh, I see. Yes, we could use some metadata; the channel name is a great example.

extensions/slack/src/search-messages.tsx
thomaslombart: Same thing here. I've passed all the correct parameters and I'm not sure what's wrong here to be honest…

pernielsentikaer:
@jfkisafk could you check this after the latest update 🙂

extensions/slack/src/search-messages.tsx
jfkisafk: Figured it out, the fix is:

```typescript
<List.Dropdown.Section>
  {channels.flat().map((c) => <List.Dropdown.Item key={c.id} icon={c.icon} value={'username' in c ? c.username : c.name} title={c.name} />)}
</List.Dropdown.Section>
```

jfkisafk: Hey @pernielsentikaer, the rest of these changes looks good to me and this is the only open issue. I have the above fix on my side but don't have permissions to push to this branch. I'm not sure how to proceed here and any guidance would be appreciated (or should we wait for Thomas to come back to this PR). This issue limits functionality in the new command, so this fix might be blocking.

pernielsentikaer: After testing it out I still see some empty lines in the dropdown, do you have a clue what might cause that?

slack 2024-06-25 at 08 39 02

It shows my name when I click one of them empty lines

slack 2024-06-25 at 08 39 09

jfkisafk: Hey @pernielsentikaer, for direct IMs with your own username, did you send a message to yourself in slack? If so, you will need to press enter to query "From Me Only" or type the query `from:me`. You can try that query to see results from "All Channels" as well.

Apart from the direct user conversations fix, we will also need to fix the `mpim` (multi-people IM groups) search query. For that the query goes `{query} in:{mpim-name}`. The query can be `from:me` and `mpim-name` is actually the `name` (that starts with `mpdm-`) getting replaced here and we discard it:

https://github.com/raycast/extensions/blob/40c3b7048ce7bfebbeffcadb1d5a276ff1b865c5/extensions/slack/src/shared/client/SlackClient.ts#L149

You will also see this name in the detail metadata "Channel Name" for the search result. We will need to persist that value to successfully query mpim(s).

extensions/slack/src/search-messages.tsx
```diff
@@ -0,0 +1,164 @@
...
+            icon={{ value: user?.icon, tooltip: user?.name ?? "Unknown user" }}
+            title={text}
+            accessories={[{ date, tooltip: formattedDate }]}
+            detail={
+              <List.Item.Detail
+                markdown={text}
+                metadata={
+                  <List.Item.Detail.Metadata>
+                    {m.channel ? (
+                      <List.Item.Detail.Metadata.Label title="Channel Name" icon={Icon.Hashtag} text={m.channel.name} />
```
jfkisafk, line 118: Suggestion (non-blocking): If possible, we would want to render the channel differently for users, actual channels and MPIM(s). `channel.name` works well with channels but not with users and MPIM groups.

extensions/slack/src/search-messages.tsx
thomaslombart: Thanks for the fix @jfkisafk. What do you mean with the `mpim-name`? I tried using these (e.g. `mpdm-john--thomas.lombart-1`) in the query but they don't seem to work, same thing for the display name and the IDs. Any ideas what should we use here?

jfkisafk:
I will take a look at this today.

extensions/slack/src/search-messages.tsx
jfkisafk: Hey @thomaslombart , that is a strange behavior. Here are some things we can try:

Do you have messages from yourself in groups >= 3 folks (you + 2 others)? We can log the messages while selecting _All Channels_ and querying `from:me`. We pick the `m.channel.name` (starting with `mpdm-`) for this group of >= 3 folks and try the query `from:me in:mpdm-...`, while still keeping _All Channels_ selected in dropdown. Let me know if you don't see results.

I am able to see messages with groups of 3 if I use above steps. The problem with `mpdm-john--thomas.lombart-1` might be that it's just 2 people and when you try to start a group with one other person in slack, it switches to DM (if groups of 3 works, we might need to also hide 2 people groups from dropdown).

jfkisafk (approved):
Everything works other than MPIM chats.

extensions/slack/src/search-messages.tsx
thomaslombart: Thanks for the details! It's working properly now.

extensions/slack/src/search-messages.tsx
```diff
@@ -110,7 +110,7 @@ function Search() {
       ) : null}
       {data?.map((m) => {
         if (!m.text || !m.ts) return null;
-
+        console.log(m);
```
jfkisafk, line 113:
```suggestion
```

jfkisafk (approved):
Works like a charm, we can remove the console.log though

jfkisafk (approved):

jfkisafk (approved):

PR #13053, opened 2024-06-20
Add pinch-svg extension

connorwforsyth:
## Description

This extension helps designers generate pixel spacing increments and references using SVGs.

**There are currently two commands:**
1.  Copy SVG Spacing Reference: 
This allows users to select and copy SVG squares based on predefined grid scaling systems.
2. Create SVG Rectangle:
This enables users to specify specific the width and height in pixels to generate an SVG rectangle. This is suitable if you want to produce a rectangle at a specific size or aspect ratio that's outside of the grid system.

**Problems this solves:**
- Some design and graphical tools don't allow you to specify pixel values to create and measure shapes. With this extension, you can quickly copy SVG rectangles from an 8px scaling system and paste them into your design tool of choice.

## Screencast

https://github.com/raycast/extensions/assets/62790515/9423de00-42a9-4205-b483-033bd6ce91c7

## Checklist


pernielsentikaer:
Feedback is given directly in Slack 🙂

pernielsentikaer:
Did you have time to look into the feedback yet @connorwforsyth 🙂

connorwforsyth:
@pernielsentikaer @mil3na 

Made a few updates:
1. For the Create SVG Rectangle Command:
- Removed the form UI and replaced it with arguments, meaning people can write and copy custom SVGs faster.
- Created error states for when user enters a text string instead of a number.
2. For the grid view:
- Removed the consolidated SVG from the grid list.

Cheers!

connorwforsyth:
> Hi! Thank you for your contribution!
> 
> One more suggestion: how about using `width` and `height` instead of 200 and 300? This would mean that, when the user doesn't input those values, the error would be correct telling what is missing instead of showing a number. Another option is leave those values as optional and make 200 and 300 default values. I would prefer the first option, but it is up to you. What do you think?

Hey @mil3na - I've made that change to the arguments placeholder text. Thanks for the suggestion! I think that makes more sense too.

mil3na (approved):
Thanks again!
LGTM.

PR #13129, opened 2024-06-23
Add universal-commands extension

cyxn:
## Description

Sometimes, I encounter difficulties in finding the right shortcut for the same action across different applications.
For instance, in various meeting platforms like Google Meet, Slack Huddle, and Zoom, each has a different shortcut for
muting the microphone.
Furthermore, in Slack, the ability to rebind shortcuts is limited.

To address this issue, the solution is the Universal
Commands extension for Raycast. This extension provides users with the ability to
configure commands to execute specific shortcuts based on the application in use. As the result, users can use a single
command to toggle between mute and unmute, and this action will be tailored to the current application and it will
execute corresponding shortcut.

## Screencast

https://github.com/raycast/extensions/assets/20022927/78569bf3-1567-47ec-b8b0-6e49107377d5

## Checklist


mil3na:
Hi!

Is your extension designed for working on web apps only? Also, how come it would be different in functionality than adding the option to be app/website specific to this extension? https://www.raycast.com/HelloImSteven/keyboard-shortcut-sequences

cyxn:
Hi Milena!

> Is your extension designed for working on web apps only?
It works with both web applications and mac applications.

My extension is different from the one mentioned above and aims for a different use case. The extension above can't solve my use-case.

Good example of usage of my application is toggle mute/unmute. In different apps it is different shortcut. In slack and google meet you can't remap the shortcut and both of them are different. With my extension, you can use the same Raycast command for both apps to toggle mute/unmute. Extension will execute correct shortcut under the hood depending on what app is opened right now. Mute/unmute is just an example. You can build your own universal commands.

mil3na:
Hey! I think you didn't answered my question. I saw you example, but my question is: the only difference I see from your extension and the one I pointed out was the app awareness, which I believe can be added to the other extension. Are there any other differences I am missing?

cyxn:
It is an entirely different perspective. My extension is more like if/else, depending on what app is running. 
1/ There is no sequence of shortcuts — just a single shortcut that will be executed only to that exact app. 
2/ If app is not matched, it won't execute any shortcut. `keyboard-shortcut-sequences` will always send shortcuts whatever is set up
3/ because of the different perspective and use cases: UI, commands, and naming of commands/extensions are tailored to that exact envisioned experiences.

I think the main point from me is — it is not only about the number of differences but more about how huge that single difference is. 

***

If you still believe those two extensions should be together, I can foresee that it will be quite challenging:
- to put it together in convenient UX
- position it properly in terms of naming (`keyboard-shortcut-sequences` doesn't really explain my use-case)


But before diving into that, I want to hear your thoughts.

mil3na:
I will do some thinking and will get back to this review tomorrow!

mil3na:
I guess I understand the use case of this. However I have a few comments:
- The "run command" command expects an id that I'm not sure where to get it. Can you explain to me why this command is necessary?
- I miss a "Create Command" command, it took me a while to find it on the action menu
- Maybe an option to pin or favorite commands on the list would be useful?

Please, take a look at my comments and tell me what you think :)

cyxn:
> The "run command" command expects an id that I'm not sure where to get it. Can you explain to me why this command is necessary?

The command is meant to be called by a Raycast QuickLink rather than by the user directly. When you view list of commands, you can create it as quicklink. This is the way I call my commands. 
Do you think I should enter the primary action (instead of running the command) to guide users using it the same way I do?

CleanShot 2024-07-05 at 16 06 50

cyxn:
> I miss a "Create Command" command, it took me a while to find it on the action menu

yes, it is cmd + N in "List Custom Commands".

Should I create new command just for create flow?

mil3na:
> Do you think I should enter the primary action (instead of running the command) to guide users using it the same way I do?

Yes!

>Should I create new command just for create flow?

I believe so. It would make the process easier.

>But who knows, maybe it's only me who thinks QuickLinks is the best way to call commands. This is why the primary action is to run the command and not save it as quick link — I want users to use the extension the way it works best for them.

It's up to you. I'd recommend to implement this now, but you can wait for user feedback as well. Anyway, I would explain your vision on the README file anyway, to avoid confusion.

cyxn:
I have updated my extension, @mil3na please look

cyxn:
> And I still don't understand how to get the input for the Run Custom Command command.
as I said, the command is not supposed to be used by the user directly. You can create a Quicklink to the command (for example mute/unmute) from "List Custom Commands" command. This Quickling will use `Run Custom Command`

mil3na (changes requested):
I think that can be done in a better way. And explained on your README file. Remember, not every user will have the same context as you. Let me know if you want to brainstorm this.

cyxn:
I have updated my extension. I think it should guide a user really well. If the user runs a command without an id, it will pop the list of commands with the primary action to run (by hitting enter).
When user goes to "manage" command, primary action would be to save it as quicklink

mil3na (approved):
LGTM! Thanks you so much for your effort!

PR #13176, opened 2024-06-26
Add rule-of-three-calculator extension

luarmr:
## Description

Simplify your calculations with the Rule of Three Calculator Raycast extension. Quickly determine proportional values by entering known quantities and letting the extension do the math for you. Ideal for anyone needing fast, accurate calculations in business, science, or everyday tasks.

This extension was created as a response of this thread:
https://raycastcommunity.slack.com/archives/C02HEMAF2SJ/p1718296485887319

ping: @ArttuLai

## Checklist


koinzhang:
It looks a lot like this. https://www.raycast.com/justiniscoding/ratio-calculator

luarmr:
Hi @mil3na  and @koinzhang , Thanks for reviewing the extension!

I get that it seems similar to the existing calculator, but here’s why I think it’s different and worth adding:

**Focused Functionality**: This extension is specifically for Rule of Three calculations, making it easier for those who need this exact function regularly.

**Simplified Input**: Users just enter three known values, and the extension calculates the fourth one. This matches the mental model many users have for these calculations. Personally, I didn’t think to search for ratios in the marketplace.

**Community Request**: This extension was created because users asked for it in a Slack thread. No one mentioned an existing extension that does this. Check out the full conversation [here](https://raycastcommunity.slack.com/archives/C02HEMAF2SJ/p1718296485887319).

**Practical Use Cases**: Many users find Rule of Three calculations tricky to do in their heads, and this tool simplifies it.
I thought about adding this to the Ratio Calculator after reading your comment but decided against it because the usage and interface are different.

I know the Slack thread is just one data point, and it might not show a broad need. If you think this extension doesn’t add much value, feel free to close the pull request. I also used this project to learn more about building extensions and forms in raycast (mission accomplished), so I won't be disappointed.

You may have the analytics to see what users might find useful, like how many people search for Rule of Three and end up installing the ratio one (probably 3 users  😛 ), also the knowledge of what raycast needs, so I’ll leave the decision to you.

mil3na:
Hi! 
I do believe that the features of this extension is already covered by other extensions and Raycast native features. I will close this, but please don't take this personally. I would love to see another extension by you around!

Cheers.

luarmr:
No problem, @mil3na.

PR #13190, opened 2024-06-27
Add shakespearify extension

Binary-Bytes:
## Checklist


mil3na:
Hi!

I could not get your extension to work. Are you sure that the API key you need is from https://exchange.gemini.com/settings/api and not https://ai.google.dev/aistudio ? Using a key from the first is accepted with no check. Submitting a text does nothing and shows no error. Can you please check what is going on?

Binary-Bytes:
> 
> I could not get your extension to work. Are you sure that the API key you need is from https://exchange.gemini.com/settings/api and not https://ai.google.dev/aistudio ? Using a key from the first is accepted with no check. Submitting a text does nothing and shows no error. Can you please check what is going on?

Hi, thanks for pointing out those issues. The wrong link was a massive oversight. I have updated the extension now.
Thanks

mil3na:
Trying it again, I have a few comments:
- Your extension cannot work without the API key, so it would make sense to keep it behind an authentication screen before letting the user use the extension. That would make that error screen with only text not needed. One example is this extension: https://www.raycast.com/CofCat/mandarin-chinese-dictionary
- What does "Wise History" mean in this context? It is not clear to me what to expect from this command by just looking at its name
- Your README doesn't say where to enter the key. Again, something that the first comment would address.

Could you please take a look at my comments and tell me what you think?

Binary-Bytes:
> Trying it again, I have a few comments:
> 
> * Your extension cannot work without the API key, so it would make sense to keep it behind an authentication screen before letting the user use the extension. That would make that error screen with only text not needed. One example is this extension: https://www.raycast.com/CofCat/mandarin-chinese-dictionary
> * What does "Wise History" mean in this context? It is not clear to me what to expect from this command by just looking at its name
> * Your README doesn't say where to enter the key. Again, something that the first comment would address.
> 
> Could you please take a look at my comments and tell me what you think?

Hi, are you referring to this screen as "authentication screen" from the `Mandarin Chinese Dictionary` extension? Because I already have the preferences screen set up.

mandarin-chinese-dictionary 2024-07-08 at 18 42 09

I have edited the description for "`Wise History`" to be a bit more clearer, and although its purpose might not be clear at first glance, but after running the command its pretty easily understandable.

The README file already states "`When you first run the extension, you will be asked a few configuration options. This is required to ensure proper functioning of the extension`" and then all the config options are explained in detail. The line from README "`Enter the API key generated from Google AI studio in the input box`" is referring to the above line where it mentioned "you will be asked a few configuration options." So like enter the API key where asked during the setup process.

mil3na:
You still have access to the Wise History command even before the extension is set up, that's what I mean. 
Also, how about naming it something that refers back to what the extension does? Maybe "Shakesperian Transformation History" or something like that?

Binary-Bytes:
It's because `Wise History` does not require the API Key for its functioning, I mean it would just show nothing if there's no history, and there would be no history if the user hasn't already used the `Shakespearify` command (which requires the key). Also, I just liked the name `Wise History` and it would be kind of *unique* or different.

mil3na (approved):
Sure, that's fair. LGTM.

PR #13205, opened 2024-06-28
`lorem-ipsum` extension: Add support for HTML tags

jimgroome:
## Description

I needed some paragraphs wrapped in `<p>` tags, and figured this might be useful for other users too, so I modified the extension to include this.

* Added optional `wrapHtmlTag` argument to the `paragraphs` command
* If the user doesn't include it in their request, everything happens as before
* If the user does, each paragraph is wrapped in the specified tag

## Screencast

https://github.com/raycast/extensions/assets/294900/8a2bb357-fa67-4348-a1ac-1d0ab08ccb2e

* User enters `3` to generate 3 paragraphs
* Pastes into VS Code - previous functionality works as before
* User enters `3` and `p` to generate 3 paragraphs wrapped in `<p>` tags
* Pastes into VS Code - 3 HTML paragraphs have been generated

## Checklist

### DIFF
--- a/extensions/lorem-ipsum/src/utils.tsx
+++ b/extensions/lorem-ipsum/src/utils.tsx
@@ -25,9 +25,9 @@ export const generateParagraphs = (count: number, wrapHtmlTag?: string) => {
         generator.generateSentences(
           Math.floor(
             Math.random() *
-              (loremIpsumOptions.sentencesPerParagraph.max - loremIpsumOptions.sentencesPerParagraph.min + 1)
-          ) + loremIpsumOptions.sentencesPerParagraph.min
-        )
+              (loremIpsumOptions.sentencesPerParagraph.max - loremIpsumOptions.sentencesPerParagraph.min + 1),
+          ) + loremIpsumOptions.sentencesPerParagraph.min,
+        ),
       )
       .join(`${wrapHtmlTag && `</${wrapHtmlTag}>`}\r\n\r\n${wrapHtmlTag && `<${wrapHtmlTag}>`}`) +
     (wrapHtmlTag && `</${wrapHtmlTag}>`)


AntonNiklasson (commented):
Do we really need to make this a separate command? I could see this as a setting on the paragraph command instead. What do you think?

jimgroome:
> Do we really need to make this a separate command? I could see this just being a setting on the paragraph command instead. What do you think?

Hi! This is the first time I've tried to contribute to an extension and was just following the docs on the official site. 

Clearly it's not ready to be included, so I'll make some changes.

I'm thinking of changing it so instead of it being a separate command, we include an optional textbox for a HTML element to wrap our paragraphs. That way the user can use it for `<p>`, `<li>`, etc.

extensions/lorem-ipsum/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Lorem Ipsum Changelog
 
+## [New Feature] - 2023-01-02
```
AntonNiklasson, line 3: Change this to todays date 📅

AntonNiklasson (changes requested):
You're doing great here, keep pushing and we'll get this in! 😊

I want to keep the flows as minimal and clean as possible. So I think this should be a command level preference rather than an argument. Check out the docs here: https://developers.raycast.com/api-reference/preferences

Perhaps we could also simplify further and assume `<p>` covers most cases, so the it's enough with just a boolean - "wrap in paragraph tags?"

extensions/lorem-ipsum/CHANGELOG.md
koinzhang: Change it to `{PR_MERGE_DATE}` and raycastbot will automatically replace the merged date.

AntonNiklasson: 🤯

mil3na:
Hi! Is this ready for review? If not, can you please mark it as a draft? Cheers!

PR #13216, opened 2024-06-28
Linear: Improve order of teams in the dropdown

suevalov:
## Description

- The order of teams in "Create issue" respects order of teams in the Linear native application.

- When using "Create issue for myself" without specifying a preferred team in settings, use first team from the sorted array of teams.

## Screencast

IlmxY6zDi6p6ycTX@2x

onrbsYMHvpW49FPV@2x

## Checklist

### DIFF
--- a/extensions/linear/src/api/getTeams.ts
+++ b/extensions/linear/src/api/getTeams.ts
@@ -1,5 +1,6 @@
 import { Cycle, Organization, Team } from "@linear/sdk";
 import { getLinearClient } from "../api/linearClient";
+import { sortBy } from "lodash";
 
 export type TeamResult = Pick<
   Team,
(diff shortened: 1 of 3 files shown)


extensions/linear/src/api/getTeams.ts
```diff
@@ -16,29 +17,32 @@ export type TeamResult = Pick<
...
     `
-      query {
+      query($userId: String!) {
         teams {
           nodes {
             id
             name
             icon
             color
-            projects {
```
suevalov, line 33: Wasn't used anywhere, plus it's not really cheap request.

jfkisafk: ty ty for the cleanup 🫶

extensions/linear/src/api/linearClient.ts
```diff
@@ -10,6 +10,7 @@ export const linear = OAuthService.linear({
       accessToken: token,
       headers: {
         "public-file-urls-expire-in": "60",
+        "linear-raycast-extension": "true",
```
jfkisafk, line 13: Nit (non-blocking): I believe this is to disambiguate requests on linear server-side 😎 . Should we use `extensionName` and `commandName` from [environment](https://developers.raycast.com/api-reference/environment) for more granularity? I think commandName can get finicky in case the client is being initialized only once.

jfkisafk (changes requested):
Just need to add the changelogs but otherwise lgtm. You can also add yourself as a contributor here:

https://github.com/raycast/extensions/blob/4b6e75594bea90f4dfd9164ae6e13b3db1e17b03/extensions/linear/package.json#L10-L18

Btw most of the changes are similar to the other [PR](https://github.com/raycast/extensions/pull/13132), where we are adding typeahead searches for teams, if you just want to add to that existing PR, let me know.

suevalov:
@jfkisafk ah, perfect! can you just copy this changes into the PR you've mentioned, so you don't have conflicts.

PR #13231, opened 2024-06-29
Add what-the-commit extension

pranshu314:
## Description

This is a basic extension that calls the https://whatthecommit.com/index.txt endpoint to fetch a absurd and comical commit message. Then it displays the commit message and gives the user to copy and paste the generated message.

## Screencast

what-the-commit-1

## Checklist


mil3na:
Hi!

We have an extension on the store that does the exact same: https://www.raycast.com/rokartur/commit-message-generator
Or am I missing something from your extension?

pranshu314:
Hi,
Thank you for looking into the extension.
I did not discover the extension you mentioned when searching for it. Hence the same contribution. I would be more careful next time.
Sorry for the inconveniences caused

pranshu314:
Though the extensions do the same thing, there is a subtle difference in that, the one mentioned has an array of commits, from which it selects a random one, while mine actually queries the whatthecommit.com endpoint to get the message.
Though I am not sure if both have the same commits list.

mil3na:
Well, I would encourage you to make that change to that extension instead of making a new one, if you believe querying the website is a better way to do it. 
Closing due to duplication.

Issue #12924, opened 2024-06-13
[Color Picker] ...


yakir-reznik:
### Extension

https://www.raycast.com/thomas/color-picker

### Description

Add the ability to pick a color for a color wheel or a color selection UI.
Sometimes i want to get a new color that i don't have anywhere on my screen.
I would also love for that newly created color to be saved in my color organizer.

Image for clarification: https://ibb.co/mbxS1F6

### Who will benefit from this feature?

Everyone using this extension

### Anything else?

https://ibb.co/mbxS1F6

PR #13246, opened 2024-06-30
[Color Picker] Add color wheel feature command

ridemountainpig:
## Description
Add `Color Wheel` command for issue #12924.

## Screencast

https://github.com/raycast/extensions/assets/92412722/25674063-46b4-40dd-88e5-3d7d42202588

## Checklist


LitoMore:
The color wheel PNG image you provided is too low. The picture is severely distorted. Please note that the Raycast bot will always compress PNG files before merging.

You can use other image formats like WebP for the best quality.

ridemountainpig:
Hi @LitoMore, thank you for your advice. I have changed the color wheel image to the WebP format.

LitoMore:
This image is still in low quality. You may need to find a better color wheel image source.

image

When you zoom in, you will notice that the colors in the picture are uneven and have irregular streaks. This will affect the accuracy of picking colors.

LitoMore:
Let's use this image file instead: [color-wheel.webp.zip](https://github.com/user-attachments/files/16097226/color-wheel.webp.zip). Your image doesn't need to be so large (10000x10000).

I picked the color wheel image from https://commons.wikimedia.org/wiki/File:Color_circle_(RGB).png. I also used [cwebp](https://developers.google.com/speed/webp/docs/cwebp) to convert the PNG to WebP. The CLI command is below:

```shell
cwebp -q 100 color-wheel.png -o color-wheel.webp
```

To make its meaning clearer, we can rename the file from `color-wheel.webp` to `rgb-color-wheel.webp`.

In addition, I would like to mention that I have tried using SVG vector images. Raycast has a rasterizing process on SVG files when rendering, which will severely distort the SVG images. So WebP is still the best solution.

mil3na:
Would be an option to programmatically generate this wheel? Like using https://developer.mozilla.org/en-US/docs/Web/CSS/gradient/conic-gradient

ridemountainpig:
@LitoMore, thank you for your image. I have updated it.

Also, thank you for @mil3na's suggestion, but the `conic-gradient()` seem does not appear to be usable in Raycast extensions.

mil3na:
@LitoMore are you happy with the image at the moment?
One thing that looks strange to me is that, even though it is a color wheel command, I can pick colors outside of that image, and even outside Raycast. Is that intentional? Can we fix that?

LitoMore:
@mil3na This is by far the best solution unless Raycast core makes changes, which is unlikely at the moment.

LitoMore (approved):

mil3na (approved):
LGTM!

mohammadazeemwani:
How do you folks change hue and all for a color picked using color wheel

PR #13282, opened 2024-07-01
WIP: Add name color to color picker

nagauta:
## Checklist

### DIFF
--- a/extensions/color-picker/src/name-color.ts
+++ b/extensions/color-picker/src/name-color.ts
@@ -34,7 +34,7 @@ export default async function command(props: PickColorCommandLaunchProps) {
       }
 
       try {
-        await callbackLaunchCommand(props.launchContext.callbackLaunchOptions, { colorName });
+        await callbackLaunchCommand(props.launchContext.callbackLaunchOptions, { hex, colorName });
       } catch (e) {
         await showFailureToast(e);
       }


extensions/color-picker/package.json
```diff
@@ -76,6 +76,12 @@
       "name": "convert-color",
       "title": "Convert Color",
       "description": "Convert any color to a different format.",
+      "mode": "no-view"
+    },
+    {
+      "name": "name-color",
+      "title": "Name Color",
+      "description": "name the color selected.",
```
thomaspaulmann, line 84: I would suggest a similar naming than the `Pick Color` command:
```suggestion
      "description": "Name a color on your desktop",
```

LitoMore:
I've noticed most of the code from `name-color` is the same as the `pick-color` command.

You may be able to re-organize the reused part to polish the code.

extensions/color-picker/src/name-color.ts
```diff
@@ -0,0 +1,66 @@
...
+    if (props.launchContext?.callbackLaunchOptions) {
+      if (props.launchContext?.copyToClipboard) {
+        await Clipboard.copy(colorName);
+      }
+
+      try {
+        await callbackLaunchCommand(props.launchContext.callbackLaunchOptions, { hex });
+      } catch (e) {
+        await showFailureToast(e);
+      }
```
LitoMore, line 40: We can callback with both `hex` and `colorName` here. The API documentation needs to be updated as well.

```suggestion
      try {
        await callbackLaunchCommand(props.launchContext.callbackLaunchOptions, { hex, colorName });
      } catch (e) {
        await showFailureToast(e);
      }
```

We can also add the `colorName` callback payload to `pick-color` command.

I can help if you need any help on composing documentation and `callLaunchCommand` functions.

extensions/color-picker/package.json
nagauta: fixed, thanks!

extensions/color-picker/src/name-color.ts
LitoMore: We should make the API convenient. So let's return both `hex` and `colorName`.

LitoMore:
We can make this feature available in UI. So that users can pick their favorite one by platform type:

- roygbiv
- HTML
- x11
- pantone
- ntc

The UX workflow can be:

1. Run `name color` command, let user to pick a color
2. After picking a color, return the UI with the name list

It also facilitates the addition of AI functions. So that I can simply add a new action to provide the Raycast Pro enhanced AI feature.

For this implementation, support for [Corss-Extension](https://github.com/LitoMore/cross-extension-conventions) is no longer needed. We can implement the name color API to `pick-color` command only.

LitoMore:
@nagauta Are you still working on this?

nagauta:
> @nagauta Are you still working on this?

Yes
I have to take google cloud exam today 
It's finished and passed.
I'm on doing this again

extensions/color-picker/src/name-color.ts
nagauta: fine, I mod as you advice

PR #13316, opened 2024-07-03
[Amazon AWS] feat: glue command

b0lle:
## Description

This PR adds Glue functionality to the AWS extension.
I am using the glue service very frequently and the extension should help to get the main tasks done in no time. Those tasks are:
- get an overview about all jobs
- trigger a job (when default arguments allow it)
- show the recent runs of a job
- get insights of the job and resource definition

## Screencast

Get on overview of all defined jobs
image
Status running
image
Get a list of the most recent job runs (limited to 25 executions, to get loading times in a reasonable limit)
image
Actions
image
Damn! Job Error ... Whats the message?
image
Get all details of the job definition
image

## Checklist

### DIFF
--- a/extensions/amazon-aws/src/glue.tsx
+++ b/extensions/amazon-aws/src/glue.tsx
@@ -34,12 +34,13 @@ export default function Glue() {
       searchBarAccessory={<AWSProfileDropdown onProfileSelected={mutate} />}
       isShowingDetail={!isLoading && !error}
     >
-      if (error)
-      {error ? (
-        <List.EmptyView title={error.name} description={error.message} icon={Icon.Warning} />
-      ) : (
-        glueJobRuns?.map((job) => <GlueJob job={job} mutate={mutateJobRuns} />)
+      {!isLoading && error && <List.EmptyView title={error.name} description={error.message} icon={Icon.Warning} />}
+      {!isLoading && !error && (glueJobRuns || []).length < 1 && (
+        <List.EmptyView title="No Glue jobs found!" icon={{ source: Icon.Warning, tintColor: Color.Orange }} />
       )}
+      {(glueJobRuns || []).map((job) => (
+        <GlueJob job={job} mutate={mutateJobRuns} />
+      ))}
     </List>
   );
 }
@@ -48,6 +49,7 @@ function GlueJob({ job: glueJobRun, mutate }: { job: GlueJobRun; mutate: MutateP
   return (
     <List.Item
       key={glueJobRun.JobName}
+      keywords={[glueJobRun.Id || "", glueJobRun.JobRunState || "", glueJobRun.ErrorMessage || ""]}
       icon={{ source: "aws-icons/glue.png", mask: Image.Mask.RoundedRectangle }}
       title={glueJobRun.JobName!}
       detail={<GlueJobRunDetails jobRun={glueJobRun} />}
@@ -83,11 +85,13 @@ function GlueJob({ job: glueJobRun, mutate }: { job: GlueJobRun; mutate: MutateP
         </ActionPanel>
       }
       accessories={[
-        { date: glueJobRun.CompletedOn ?? glueJobRun.StartedOn, tooltip: glueJobRun.CompletedOn ? "Completed On" : "Started On" },
         {
-          text: glueJobRun.accessoriesText,
+          date: glueJobRun.CompletedOn ?? glueJobRun.StartedOn,
+          tooltip: glueJobRun.CompletedOn ? "Completed On" : "Started On",
+        },
+        {
           icon: { source: glueJobRun.icon!, tintColor: glueJobRun.iconTintColor },
-          tooltip: "Recent Job Run State",
+          tooltip: glueJobRun.accessoriesText,
         },
       ]}
     />
@@ -101,15 +105,15 @@ function GlueJobRuns({ glueJobName: glueJobName }: { glueJobName: string }) {
       {glueJobRuns?.map((jobRun) => (
         <List.Item
           accessories={[
-            { date: jobRun.CompletedOn, tooltip: "Completed On" },
+            { date: jobRun.StartedOn, tooltip: "Started On" },
             {
-              text: jobRun.accessoriesText,
               icon: { source: jobRun.icon!, tintColor: jobRun.iconTintColor },
               tooltip: "Recent Job Run State",
             },
           ]}
-          title={jobRun.CompletedOn?.toUTCString() || ""}
+          title={jobRun.StartedOn?.toUTCString() || ""}
           detail={<GlueJobRunDetails jobRun={jobRun} />}
+          keywords={[jobRun.JobRunState || "", jobRun.ErrorMessage || "", jobRun.Id || ""]}
           actions={
             <ActionPanel>
               <AwsAction.Console url={resourceToConsoleLink(jobRun.Id, "AWS::Glue::JobRun")} />
@@ -127,6 +131,11 @@ function GlueJobRuns({ glueJobName: glueJobName }: { glueJobName: string }) {
 function GlueJobRunDetails({ jobRun: glueJobRun }: { jobRun: GlueJobRun }) {
   return (
     <List.Item.Detail
+      markdown={
+        glueJobRun.Arguments
+          ? `**Arguments**\n\`\`\`json\n${JSON.stringify(glueJobRun.Arguments, null, 2)}\n\`\`\``
+          : undefined
+      }
       metadata={
         <List.Item.Detail.Metadata>
           <List.Item.Detail.Metadata.Label title="Job Name" text={glueJobRun.JobName} />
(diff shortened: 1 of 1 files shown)


b0lle:
Hi @jfkisafk,

I nearly finished my first implementation of the glue command. For now it works like a charm except for one minor feature. I would love to refresh the list of job runs, once I triggered a job. ATM I am facing skill issues while I try to get familiar with Typescript. Do you may have any suggestions to get this done?

jfkisafk:
Hello @b0lle , I have slated to review this by end of week. The screenshots look assuming, thanks for the ownership 🫡 . 

> I would love to refresh the list of job runs, once I triggered a job.

Yes please use the `mutate` property you are getting to optimistically update the job in the list. You can find multiple examples of mutation in [this PR](https://github.com/raycast/extensions/pull/13319). For example, look at how we are updating the queue after purging the messages:

https://github.com/raycast/extensions/blob/76a9151819ecf8e5895d406c71a57fdf4874ab31/extensions/amazon-aws/src/sqs.tsx#L220-L235

**Note:** If you don't set `shouldRevalidateAfter: false` then entire list revalidation will also be performed in the background after optimistic update and you should see if that is necessary. For pipeline commands, I've set this to false and just relying on optimistic updates and next reload of command to refresh to current state.

b0lle:
I will implement your suggestions this afternoon

jfkisafk:
Hey @b0lle let me know when this is ready to review

b0lle:
Sure, this is Reese's for your review, thanks! @jfkisafk

jfkisafk:
Hey @b0lle I am in the process of reviewing this PR. Can you please fix all build issues first? I have reviewed your other S3 PR btw.

b0lle:
Oh sorry I didn't notice the build errors. I appreciate your feedback and will work on it this das or tomorrow.

b0lle:
@jfkisafk just applied the fix and would appreciate your feedback

jfkisafk:
> @jfkisafk just applied the fix and would appreciate your feedback

Hey @b0lle , sorry I got busy with stuff and since I am not an active glue user, this review is taking more time. I will take a look tomorrow and over the weekend.

pernielsentikaer:
Did you have time to check this @jfkisafk 🔥

jfkisafk:
Hey @b0lle @pernielsentikaer sorry about the delay as it dropped out of my radar. I am taking a look now.

extensions/amazon-aws/src/glue.tsx
```diff
@@ -0,0 +1,291 @@
...
+            text={{ value: glueJobRun.JobRunState || "", color: glueJobRun.iconTintColor }}
+          />
+          <List.Item.Detail.Metadata.Label title="Execution Time" text={glueJobRun.ExecutionTime?.toString() + "s"} />
+          <List.Item.Detail.Metadata.Label title="Started on" text={glueJobRun.StartedOn?.toUTCString()} />
+          <List.Item.Detail.Metadata.Label title="Completed On" text={glueJobRun.CompletedOn?.toUTCString() || ""} />
+          <List.Item.Detail.Metadata.Label
+            title="Error Message"
+            text={{ color: Color.Red, value: glueJobRun.ErrorMessage || "" }}
+          />
+          <List.Item.Detail.Metadata.Label title="Arguments" text={JSON.stringify(glueJobRun.Arguments)} />
```
jfkisafk, line 144: I think we can use this as the markdown property for `List.Item.Detail`:

```typescript
<List.Item.Detail markdown={glueJobRun.Arguments ? `**Arguments**\n\`\`\`json\n${JSON.stringify(glueJobRun.Arguments, null, 2)}\n\`\`\`` : undefined} metadata{...} />
```

extensions/amazon-aws/src/glue.tsx
```diff
@@ -0,0 +1,291 @@
...
+          <List.Item.Detail.Metadata.Label title="DPU Seconds" text={(glueJobRun.DPUSeconds?.toString() || "") + "s"} />
```
jfkisafk, line 145: For optional properties like these, please render them optionally, for example:

https://github.com/raycast/extensions/blob/b93225144bde66a7433401e2b4b6270538fa2265/extensions/amazon-aws/src/dynamodb.tsx#L167-L169

Otherwise you would be left with empty titles like in the screenshot.

extensions/amazon-aws/src/glue.tsx
```diff
@@ -0,0 +1,291 @@
...
+      navigationTitle="Glue Job Definition"
+      markdown={
+        `# ` +
+        glueJobName +
+        `\n\n` +
+        description +
+        `\n\n` +
+        `___\n## Job Definition\n` +
+        `\`\`\`json\n` +
+        JSON.stringify(glueJobDetails?.Job, null, 2) +
```
jfkisafk, line 194: Keep in mind if this definition gets large, there might be rendering delays for the `Detail` view, and in worst case scenarios might cause JS heap memory issues. I faced this problem when I was handling state machine definitions or cloud formation templates the same way (but not sure if glue job definitions will get that large). 

Can we check the response length and use a max threshold where it doesn't get slow. After a max threshold response length say 10000 characters we can just show on the Detail view that the definition was too long to display and prompt user to either copy it to clipboard or see it in console.

extensions/amazon-aws/src/glue.tsx
```diff
@@ -0,0 +1,291 @@
...
+            title="MaxConcurrentRuns"
+            text={glueJobDetails?.Job?.ExecutionProperty?.MaxConcurrentRuns?.toString()}
+          />
+          <Detail.Metadata.Label title="MaxCapacity" text={glueJobDetails?.Job?.MaxCapacity?.toString()} />
+          <Detail.Metadata.Label title="NumerOfWorkers" text={glueJobDetails?.Job?.NumberOfWorkers?.toString()} />
+          <Detail.Metadata.Label title="Timeout" text={glueJobDetails?.Job?.Timeout?.toString()} />
+          <Detail.Metadata.Separator />
+          <Detail.Metadata.Label title="CreatedOn" text={glueJobDetails?.Job?.CreatedOn?.toISOString()} />
+          <Detail.Metadata.Label title="GlueVersion" text={glueJobDetails?.Job?.GlueVersion?.toString()} />
+          <Detail.Metadata.Label title="PythonVersion" text={glueJobDetails?.Job?.Command?.PythonVersion} />
```
jfkisafk, line 224: Same as the other comment. If some properties can be optional, we will need to render the `Detail.Metadata.Label` optionally as well.

extensions/amazon-aws/src/glue.tsx
```diff
@@ -0,0 +1,291 @@
...
+          <Detail.Metadata.Label
```
jfkisafk, line 225: Can we put this in the markdown as well? Not a big fan of json to be rendered in metadata (especially if it is long, it gets redacted).

extensions/amazon-aws/src/glue.tsx
```diff
@@ -0,0 +1,291 @@
...
+  return (
+    <Detail
+      navigationTitle="Glue Job Definition"
+      markdown={
+        `# ` +
+        glueJobName +
+        `\n\n` +
+        description +
+        `\n\n` +
+        `___\n## Job Definition\n` +
```
jfkisafk, line 192:
```suggestion
        `### ` +
        glueJobName +
        glueJobDetails?.Job?.Description ? `\n\n${description}` : '' +
        `\n\n` +
        `___\n**Job Definition**\n` +
```

extensions/amazon-aws/src/glue.tsx
```diff
@@ -0,0 +1,291 @@
...
+          <List.Item.Detail.Metadata.Label title="DPU Seconds" text={(glueJobRun.DPUSeconds?.toString() || "") + "s"} />
+          <List.Item.Detail.Metadata.Label title="Attempt" text={glueJobRun.Attempt?.toString()} />
+          <List.Item.Detail.Metadata.Separator />
+          <List.Item.Detail.Metadata.Label title="Worker Type" text={glueJobRun.WorkerType} />
+          <List.Item.Detail.Metadata.Label title="Number of Workers" text={glueJobRun.NumberOfWorkers?.toString()} />
+          <List.Item.Detail.Metadata.Label title="Timeout" text={glueJobRun.Timeout?.toString() + "s"} />
+          <List.Item.Detail.Metadata.Label title="Profile Name" text={glueJobRun.ProfileName || ""} />
+          <List.Item.Detail.Metadata.Separator />
+          <List.Item.Detail.Metadata.TagList title="Versions">
+            <List.Item.Detail.Metadata.TagList.Item text={"Glue V" + glueJobRun.GlueVersion} color={"#eed535"} />
```
jfkisafk, line 154:
```suggestion
            <List.Item.Detail.Metadata.TagList.Item text={"Glue V" + glueJobRun.GlueVersion} color={Color.Yellow} />
```

jfkisafk (changes requested):
Overall an awesome change that covers a lot of footprint. We just need to do some housekeeping to make it more robust.

pernielsentikaer:
Thanks for taking a stab @jfkisafk, could you have a look @b0lle

extensions/amazon-aws/src/glue.tsx
b0lle: I dont think that this i going to cause any issues. Glue definitions are, other than statemachines and cloud formation templates, not implementation specific. So they do not grow in size and actually its not an issue.

extensions/amazon-aws/src/glue.tsx
```diff
@@ -0,0 +1,291 @@
...
+  iconTintColor?: Color;
+}
+
+export default function Glue() {
+  const { data: glueJobs, error, isLoading, mutate } = useCachedPromise(fetchJobs);
+  const {
+    data: glueJobRuns,
+    isLoading: isLoadingJobRuns,
+    mutate: mutateJobRuns,
+  } = useCachedPromise(fetchJobRuns, [1, glueJobs]);
```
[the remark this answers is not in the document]
b0lle: I am going to implement it. Need to figure out which improvement you exactly mean

extensions/amazon-aws/src/glue.tsx
b0lle: Turns out that the default arguments are already part of te glue job definition. So I deleted the property from metadata

b0lle:
I am not finished with all PR comments yes (WIP). Thanks @jfkisafk for your effort!
It turns out, that the fist detail view is impractical. Once one have larger job names, the truncation affects the overview. So I disabled the Details view.

extensions/amazon-aws/src/glue.tsx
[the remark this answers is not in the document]
jfkisafk: For sure take a look at _hooks/use-ddb.ts_ or _hooks/use-pipelines.ts_ from this PR: https://github.com/raycast/extensions/pull/13667. Essentially, 2 things:
1. There is always a call to list the pipelines/tables (or in your case jobs, `ListGlueJobs`). You can specify `MaxItems` (or corresponding limit parameter) to slow down the TPS for parallel requests in the next step.
2. After we have received the page of Jobs/Pipelines/TableNames, we make async calls to get more granular details for the job in parallel: `ListGlueJobRuns` in your case. Right now, your `ListGlueJobRuns` is sequential. I pipelines example, we made the call to `GetPipelineExecutions` similary.
3. If there are more pages for the jobs/pipelines, we repeat the process and aggregate the result.

The thing to balance is speed/latency vs TPS as explained in the pipelines issue. We don't want to make too many requests to get throttled and if we make sequential requests it might be too slow.

extensions/amazon-aws/src/glue.tsx
```diff
@@ -0,0 +1,326 @@
...
+    />
+  );
+}
+
+async function RunGlueJob(glueJobName: string, mutate: MutatePromise<GlueJobRun[] | undefined>) {
+  mutate(new GlueClient({}).send(new StartJobRunCommand({ JobName: glueJobName })), {
+    optimisticUpdate(data) {
+      return data;
+    },
+  })
```
jfkisafk, line 192: This mutation is just doing a simple _revalidation_, which is not what we intend. We shouldn't be revalidating every job in case we have started a run for a particular job. So essentially, we should `optimisticUpdate` only certain fields like `StartedOn`, `JobRunState`, icon, accessoryText, etc. (whatever makes sense) and then set `shouldRevalidateAfter` to `false`.

jfkisafk:
> It turns out, that the fist detail view is impractical.

Hey @b0lle, which view are we talking about here? I still see the `GlueJobDefinition` view in the diff

extensions/amazon-aws/src/glue.tsx
```diff
@@ -0,0 +1,291 @@
...
+            title="Refresh"
+            onAction={() =>
+              mutate(fetchJobRuns(1, [glueJobRun.JobName!]), {
+                optimisticUpdate(data) {
+                  return data;
+                },
+              })
+            }
+            icon={Icon.RotateAntiClockwise}
+          />
```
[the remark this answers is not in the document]
jfkisafk: Ahh this is a little more complicated. I see you want to just get the latest state for the job and ideally wouldn't want to revalidate all jobs. For this case, what you had before and my recommendation do not work.

I think to keep things simple here, we just need:
1. a call to fetch the latest job run for the given job, say we store it in `newJobRun = (await fetchJobRuns(1, [glueJobRun.JobName!]))[0]`
2. we just filter and override the job state in the `glueJobRuns` list we are using throughout, so `glueJobRuns = glueJobRuns.map(j => j.Id !== glueJobRun.Id ? j : {...j, ...newJobRun} as GlueJobRun)`

I think this would achieve what you want w/o requiring a full-stack mutation. If this doesn't work try to do this with mutation, with `shouldRevalidateAfter` set to `false`. What do you think?

b0lle:
Sorry for my late response. I moved my white developer basement skin in the sun and held some of those feeds ins the ocean for vacation. I will have a glance at the rest of the PR the next days

pernielsentikaer:
We can always reopen this, @b0lle, if you'd like at some point to continue

b0lle:
yes please reopen - just merged my branch with the current main

pernielsentikaer:
Did you do something specific? Seems like I can't reopen it in this state (Like we use to)

Raycast03092024-rH2JLJhA png@2x

b0lle:
I rebased on main and force pushed ;)

pernielsentikaer:
Ah, can you open a new PR instead?

PR #13332, opened 2024-07-05
Add photoroom-image-editing extension

ubuntudroid:
## Description

This PR adds the official [Photoroom](https://www.photoroom.com) extension to Raycast. It allows image resizing, removing and changing backgrounds, adding shadows and much more.

The idea is to more or less mirror what the [Photoroom API playground web interface](https://app.photoroom.com/playground-website) provides.

Please note, that the extension uses the Image Editing v2 API and needs a Photoroom API key which can be requested here: https://app.photoroom.com/api-dashboard

## Screencast

https://github.com/raycast/extensions/assets/519867/5b0c57b0-f1e3-49ea-94aa-c19f4e7868eb

## Checklist


extensions/photoroom-image-editing/README.md
```diff
@@ -0,0 +1,3 @@
+# Photoroom Image Editing
+
+Resize images, remove and change the background, add shadows and much more!
```
mil3na, line 3: I would rather a longer README, listing all the capabilities. "Much more" doesn't say much.
Also, explaining that you can submit multiple images for instance. It was not clear to me at the beginning.

extensions/photoroom-image-editing/index.tsx
```diff
@@ -0,0 +1,393 @@
...
+
+      <Form.TextField
+        title="Width x Height"
+        placeholder="1000x1000"
+        info="The width and height of the output image in pixels. Ignored if output size is not custom."
+        {...itemProps.widthxHeight}
+      />
+
+      <Form.TextField
+        title="Max Height"
```
mil3na, line 379: What would be the difference in setting max height/width and the widthxheight field? Also, I was able to insert non-numbers charatcers on that field. That can be avoided.
Maybe for the widthxheight field, we can have a few pre-defined options?

CleanShot 2024-07-05 at 15 13 50

extensions/photoroom-image-editing/src/index.tsx
```diff
@@ -0,0 +1,393 @@
...
+  outputSize: string;
+  widthxHeight: string;
+  maxWidth: string;
+  maxHeight: string;
+};
+
+interface Preferences {
+  photoroomApiKey: string;
+}
+
```
mil3na, line 42: Is this file duplicated or is just my impression?

mil3na (changes requested):
Hi!

extensions/photoroom-image-editing/README.md
ubuntudroid: Makes sense, done! ✅

extensions/photoroom-image-editing/index.tsx
ubuntudroid: The difference between "Width x Height" and "Max Height/Width" is described when hovering over the info icons to the right of the text input fields. It essentially depends on the selected "Output Size".

That's also why you were able to put in text into "Max Height" while having the "Custom" output size selected. When "Custom" is selected, "Max Height" and "Max Width" are ignored.

ubuntudroid: Pre-defined options for sizes are a good idea, but non-essential for a first release, I'd say.

extensions/photoroom-image-editing/src/index.tsx
ubuntudroid: Yes, you are right, not sure how that happened. I've removed it. 🙏

mil3na (approved):
Just added a bit more info on the README. LGTM, but I do recommend taking a second look on sizes input for a future update.

PR #13336, opened 2024-07-05
Brew: Update dependencies, upgrade deprecated APIs, fix lint

vpukhanov:
## Description

In this pull request, I have:

- Upgraded all dependencies to their latest versions
- Updated the project template files (e.g., .gitignore, package.json, tsconfig) to the versions used by the current new extension template
- Removed the eslintignore file as it is no longer necessary
- Fixed the lint errors that arose after the dependencies were upgraded
- Replaced the use of the deprecated `accessoryTitle` prop of `List.Item` with the recommended `accessories` prop

## Checklist

### DIFF
--- a/extensions/brew/CHANGELOG.md
+++ b/extensions/brew/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Brew Changelog
 
+## [Improvements] - 2024-06-14
+
+- Updated dependencies
+
 ## [Bug Fix] - 2024-07-03
 
 - Fix a crash that could occur when showing installed formula or cask


nhojb (approved):
Looks good, thanks for the update.

The only thing I noticed is that there are now deprecation warnings when running `npm install`:

Screenshot 2024-07-07 at 18 10 59

vpukhanov:
> The only thing I noticed is that there are now deprecation warnings when running `npm install`

Yes, I noticed those too. These are all transitive dependencies of ESLint 8. ESLint 9 is now available, but `@raycast/eslint-config` does not yet work with it, so we can't get rid of these deprecation warnings just yet.

PR #13420, opened 2024-07-11
Add Japanese Number Converter extension

mrsideshowjack:
## Description

This extension lets you convert japanese kanji symbols for numbers into Number format.

examples:
```
三十 --> 30
35万 --> 350,000
```

Under the hood the value is passed into the package [@geolonia/japanese-numeral
](https://github.com/geolonia/japanese-numeral) and converted with `kanji2number()`

**Why use this over google translate?**

I find myself using this a bit as when you put '三十万' into google translate it will output 'Three hundred thousand' which I almost never need

However if you think its too similar to (the many) translation extensions, then its fine to decline 😄 

## Screencast

image
image

## Checklist

image


mil3na:
Hi!

We have one for Chinese numerals, so I think that having one for Japanese makes sense. https://www.raycast.com/raffeyang/chinese-numbers/commands

One thing that came to mind was to either have more outputs for each number and the user could chose which one they would like OR make this a one line command, where the input would be inline and "enter" would paste the output directly to the frontmost app or add to the clipboard. What do you think?

mil3na:
Ping @mrsideshowjack :)

PR #13432, opened 2024-07-11
Update genius-lyrics extension

johnstclair:
## Description

I added a command, Search Songs by Lyrics, to the extension. This command allows the user to find songs from the Genius Lyric API using remembered lyrics.

## Screencast

This is the result of querying "strangers to love" using the new, Search Lyrics by Title, command:

genius-lyrics 2024-07-11 at 09 54 34

This is the result of querying "strangers to love" using the existing, Search Song by Lyrics, command:

genius-lyrics 2024-07-11 at 09 56 11

## Checklist

### DIFF
--- a/extensions/genius-lyrics/CHANGELOG.md
+++ b/extensions/genius-lyrics/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Genius Lyrics Changelog
 
-## [New Command] - {PR_MERGE_DATE}
+## [New Command] - 2024-07-12
 
 - Added the `Search with Lyrics` Command
 


TKDKid1000:
Overall, I think this looks really good, and I like the idea of searching by lyrics. However, I think the command names are a bit confusing. Do you have any better name suggestions, or do you think the current names (with your changes) will be straightforward?

Also, would it be possible to update this with the native `useFetch` from `@raycast/utils`? I can also do it myself, but if you have the time, it would be easier. Removing an external dependency is always nice, especially when it's a large and somewhat unnecessary library like axios. Since your code doesn't appear to use any axios-specific features, a change to `useFetch` doesn't look too difficult.

Lastly, addressing the naming bit, do you think interweaving the results from both the lyric and title search would be a bad idea? I can likely write this, I'm just curious if it would be better to have two separate commands, or to use one unified command that searches both things.

Excellent work here! If you can make that slight change away from axios, that would be great, but otherwise I love this.

johnstclair:
I took a look at `useFetch`, I should be able to rewrite the extension to use that instead. I'll get to work on that now.

As for the naming, I agree that "Search for Lyrics" and "Search with Lyrics" are a bit confusing, maybe "Find Lyrics by Title" and "Find Song by Lyrics" could be better? Let me know what you think.

About interweaving results, I find this idea interesting. After all, that's what the Genius site does;

image

Personally I like the two separate commands, but if you think combining them would be better, go ahead.

I also noticed some inconstancies between the two commands that missed before, so I'll clean those up as well.

TKDKid1000:
For the purposes of efficiency, it might be smarter to keep them separate, so people can search for exactly what they want.

Yes, those names would certainly work much better. People can more easily differentiate those, and they can add aliases if it's needed. Personally, I think "search" should be used instead of "find," but otherwise your suggested names are perfect. I say that because none of the core Raycast commands used "find," but instead "search."

Thank you for migrating to `useFetch`, by the way. It's just nice to remove a dependency when you can.

TKDKid1000 (approved):

mil3na (approved):
Hi!

Great job, you guys! LGTM!

PR #13447, opened 2024-07-12
[GitHub Gist] Reduce GitHub authentication

koinzhang:
## Description

- Reduce GitHub authentication permissions to access only GitHub Gist content
- Fix the issue of **Open in Ray.so** command displaying garbled code
- Add new Action: Save as Quicklink

## Checklist

### DIFF
--- a/extensions/github-gist/CHANGELOG.md
+++ b/extensions/github-gist/CHANGELOG.md
@@ -1,6 +1,6 @@
 # GitHub Gist Changelog
 
-## [Reduce GitHub Authentication] - {PR_MERGE_DATE}
+## [Reduce GitHub Authentication] - 2024-07-16
 
 - Reduce GitHub authentication permissions to access only GitHub Gist content
 - Fix the issue of **Open in Ray.so** command displaying garbled code


extensions/github-gist/src/api/oauth.ts
```diff
@@ -6,7 +6,7 @@ import { GithubClient } from "./github-client";
 let client: GithubClient | undefined = undefined;
 
 export const githubOAuthService = OAuthService.github({
-  scope: "repo gist read:user",
+  scope: "gist",
```
jfkisafk, line 9: Tyty, I just copied the scope from README but it looked overkill.

extensions/github-gist/src/util/utils.ts
```diff
@@ -2,13 +2,14 @@ import { parse } from "path";
...
 import { GistItem } from "./gist-utils";
+import { encodeURI } from "js-base64";
 
 export const isEmpty = (string: string | null | undefined) => {
   return !(string != null && String(string).length > 0);
 };
 
 export const raySo = (title: string, content: string) =>
-  `https://ray.so/?colors=cnady&background=true&darkMode=false&padding=32&title=${title}&code=${content}&language=auto`;
+  `https://ray.so/#colors=cnady&background=true&darkMode=false&padding=32&title=${title}&code=${encodeURI(content)}&language=auto`;
```
jfkisafk, line 12: Thought (non-blocking): It would be great if we can provide a preference for theme to use. The current themes are: https://ray.so/api/config

jfkisafk (approved):

extensions/github-gist/src/util/utils.ts
wilmoore: I suspect that reviewers will review much faster if the PR is singly-focused.

It is a small PR with only two changes; however, currently one change is security/attack-surface focused; while the other is an unrelated utility change.

If the PR is focused to a single thing, I believe it will get reviewed much faster.

wilmoore:
Also, for the PR, here's a screenshot of what the OAuth user prompt looks like with the superfluous scopes.

image

extensions/github-gist/src/util/utils.ts
jfkisafk: +1 that's why I suggested it is non-blocking. Just a suggestion for future iteration.

wilmoore: @jfkisafk I see what you're saying; I think it's non-ideal to increase the cognitive load of this (or any) PR with anything (even tiny) outside of the original scope (no pun intended) of the PR.

It might seem nitpicky; however, it isn't an ideal practice and is a bit of an anti-pattern.

mil3na (approved):
LGTM!

PR #13449, opened 2024-07-12
Refactor to `dub` TypeScript SDK

steven-tey:
## Description

Simplify Dub Raycast Extension by using `dub` TypeScript SDK.

Also added support for shortening links with custom domains:

CleanShot 2024-07-15 at 08 20 36

## Checklist

### DIFF
--- a/extensions/dub-link-shortener/CHANGELOG.md
+++ b/extensions/dub-link-shortener/CHANGELOG.md
@@ -1,7 +1,7 @@
 # Dub Link Shortener Changelog
 
 
-## [Workspace API Keys] - {PR_MERGE_DATE}
+## [Workspace API Keys] - 2024-07-19
 
 - Simplify Dub Raycast Extension by using [`dub` TypeScript SDK](https://dub.co/solutions/typescript)
 - ⚠️This is a breaking change if you are using _User API Keys_. Please switch to using [Workspace API keys](https://dub.co/blog/workspace-api-keys) to continue using the extension.


jfkisafk:
Hey @steven-tey , I tried to use the sdk too, but got the same error before! So I stuck with `axios` for now.

steven-tey:
@peduarte @thomaslombart any ideas here? 👀

extensions/dub-link-shortener/src/search-links.tsx
```diff
@@ -226,7 +180,7 @@ const tryDeleteLink = async (linkId: string, mutate: MutatePromise<LinkSchema[]>
         primaryAction: {
           title: "Retry",
           onAction: async (toast) => {
-            await tryDeleteLink(linkId, mutate, workspaceId);
+            await tryDeleteLink(linkId, mutate);
             await toast.hide();
           },
```
jfkisafk, line 185: `await toast.hide()` might be a mistake on my part. I noticed this hides the toast after the first retry making second retry impossible. Can you please change this to:

```typescript
onAction: () => tryDeleteLink(linkId, mutate);
```

extensions/dub-link-shortener/src/search-links.tsx
```diff
@@ -1,74 +1,28 @@
...
-        ) : undefined
-      }
     >
-      {(workspacesError || linksError) && (
+      {linksError && (
         <List.EmptyView
-          title={workspacesError ? "Failed to fetch workspaces" : "Failed to fetch short links"}
-          description={workspacesError?.message ?? linksError?.message}
+          title="Failed to fetch short links"
+          description={linksError?.message}
```
jfkisafk, line 21:
```suggestion
          description={linksError.message}
```

extensions/dub-link-shortener/src/shorten-link.tsx
```diff
@@ -69,67 +69,50 @@ const ShortenLinkForm = ({ retryValues, args }: { retryValues?: ShortLinkFormVal
...
+          />
+        ))}
       </Form.TagPicker>
       <Form.Dropdown {...itemProps.domain} title="Domain">
         <Form.Dropdown.Item
           key="dub.sh"
           value="dub.sh"
           title="dub.sh"
-          icon={{ source: Icon.Link, tintColor: Color.Blue }}
+          icon={{ source: Icon.Globe, tintColor: Color.Blue }}
```
jfkisafk, line 106: Question (non-blocking): Should this color also be Purple like other domains?

extensions/dub-link-shortener/package.json
```diff
@@ -45,7 +45,8 @@
   "dependencies": {
     "@raycast/api": "^1.61.2",
     "@raycast/utils": "^1.16.0",
-    "axios": "^1.6.2"
+    "axios": "^1.6.2",
```
jfkisafk, line 48: Question (non-blocking): Can't we just remove `axios` now?

extensions/dub-link-shortener/src/api/index.ts
```diff
@@ -1,66 +1,47 @@
...
-    data: { domain, url, key, tagIds, comments },
+}: {
+  originalUrl: string;
+  key?: string;
+  domain?: string;
+  tagIds?: string[];
+  comments?: string;
+}) => {
+  return dub.links.create({
+    url: originalUrl,
```
jfkisafk, line 22: Question (non-blocking): Was the `hasHttps` check not required?

extensions/dub-link-shortener/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Dub Link Shortener Changelog
 
+## [Workspace API Keys] - {PR_MERGE_DATE}
+
+- Simplify Dub Raycast Extension by using [`dub` TypeScript SDK](https://dub.co/solutions/typescript)
+
```
jfkisafk, line 6:
```suggestion
## [Workspace API Keys] - {PR_MERGE_DATE}

- Simplify Dub Raycast Extension by using [`dub` TypeScript SDK](https://dub.co/solutions/typescript)
- ⚠️This is a breaking change if you are using _User API Keys_. Please switch to using [Workspace API keys](https://dub.co/blog/workspace-api-keys) to continue using the extension.

```

jfkisafk (commented):
We can address these while we try to resolve that Header issue. After that, we might need to see if the DUB error(s) are being properly propagated to the user.

extensions/dub-link-shortener/src/api/index.ts
steven-tey: yup, we handle this on the backend automatically via Zod + the URL API: https://github.com/dubinc/dub/blob/main/apps/web/lib/zod/schemas/utils.ts#L4-L7

extensions/dub-link-shortener/src/shorten-link.tsx
steven-tey: Trying to distinguish between [default Dub domains](https://dub.co/help/article/default-dub-domains) and [custom domains](https://dub.co/help/article/how-to-add-custom-domain), hence the color difference 😄

steven-tey:
@jfkisafk thank you so much! Addressed all your feedback 🙏

jfkisafk:
Hey @steven-tey this is not a raycast issue, b/c the error is happening during the `getAllShortLinks` call:

```
ReferenceError: Headers is not defined
    at Links.list (/Users/stelo/.config/raycast/extensions/dub-link-shortener/search-links.js:22338:26)
    at getAllShortLinks (/Users/stelo/.config/raycast/extensions/dub-link-shortener/search-links.js:23970:20)
    at /Users/stelo/.config/raycast/extensions/dub-link-shortener/search-links.js:216:77
    at /Users/stelo/.config/raycast/extensions/dub-link-shortener/search-links.js:339:11
    at Cr (/Applications/Raycast.app/Contents/Resources/RaycastNodeExtensions_RaycastNodeExtensions.bundle/Contents/Resources/api/node_modules/react-reconciler/react-reconciler.production.min.js:7:20793)
    at Bn (/Applications/Raycast.app/Contents/Resources/RaycastNodeExtensions_RaycastNodeExtensions.bundle/Contents/Resources/api/node_modules/react-reconciler/react-reconciler.production.min.js:7:39556)
...
```

In the generated file, the source code is copied from the dub TS SDK, and the `Header` is not found here:

https://github.com/dubinc/dub-ts/blob/07f40571e93bc17f0fe0000dc14f625cf18c34cc/src/sdk/links.ts#L81-L83

steven-tey:
@jfkisafk this should be the [`Headers` Web API](https://developer.mozilla.org/en-US/docs/Web/API/Headers) which doesn't need to be imported/defined. Is this not supported in Raycast extensions?

jfkisafk:
It is strange, I can see that the import is not necessary (my IDEs are also indicating that). Not sure why in the generated .js file this is causing an issue @steven-tey . Let me investigate more. 

Or easier debugging option would be to just use `new Header()` in the src `.ts` code and see if that throws an error.

jfkisafk:
Hey @steven-tey the problem is exactly this: https://stackoverflow.com/a/65355673

I tested it by just placing:

```typescript
new Headers({
  "Content-Type": "application/json",
})
```

in the codepath without the `node-fetch` import and it failed with `Headers not found` error. But when I added `import {Headers} from "node-fetch", it succeeded. So I think an amendment in the SDK is required to gracefully handle this.

steven-tey:
@jfkisafk just added a polyfill patch thanks to @disintegrator that fully fixes this! 🤩

jfkisafk (approved):

extensions/dub-link-shortener/fetch-polyfill.patch
```diff
@@ -0,0 +1,81 @@
+diff --git a/extensions/dub-link-shortener/package-lock.json b/extensions/dub-link-shortener/package-lock.json
```
mil3na, line 1: Is this file necessary?

mil3na (commented):
Other than this `.patch` file, things look okay on my side. Ready to merge after I get some answers :)

extensions/dub-link-shortener/fetch-polyfill.patch
jfkisafk: @steven-tey I had the same question, it looked like a git patch that was already applied. Shall we clean this up?

jfkisafk:
Also @steven-tey , I'm not sure if we want to rename this preference (and change the `title` and `description` to reflect workspace key while we are at it):

https://github.com/raycast/extensions/blob/bcce8f1242b54fc0f719657868bdead497a496d4/extensions/dub-link-shortener/package.json#L53

This will prompt everyone to input the workspace API key. It might be redundant for the folks already using workspace API key. On the other hand, folks using User API key atm won't be treated with error straight away.

extensions/dub-link-shortener/fetch-polyfill.patch
thomaslombart: Same question here 😄

extensions/dub-link-shortener/src/hooks/use-domains.ts
thomaslombart, line 1: As an improvement, the error toast could use a `message` property to let the user know what has exactly failed. I guess that those who were logged in with a user API key may be confused if facing an error like that. Three solutions:

1. Cover an authentication error and ask them in the failure toast to switch to workspace API keys with an action guiding them to switch their API keys.
2. Change the preference's API key name, though it will impact those who already have a workspace API key. This could be a good idea if the latter is not used a lot.
3. I don't know if it's possible on your end to differentiate a workspace API key from a user API key. But it could also be possible to display some kind of alert for users who use User API keys when starting up a command (either with an empty view or `confirmAlert`).

extensions/dub-link-shortener/fetch-polyfill.patch
steven-tey: @disintegrator can we remove this patch file since it's already been applied? or do we still need to keep it?

steven-tey: OK removing the patch!

extensions/dub-link-shortener/src/hooks/use-domains.ts
steven-tey: Ended up going with #2, which is also what @jfkisafk suggested above:

CleanShot 2024-07-18 at 22 35 14

This PR should be ready for merge now!

thomaslombart (approved):

PR #13465, opened 2024-07-13
Add exercises extension

ajk68:
## Checklist


mil3na (changes requested):
Hi!

- On the `Find Exercice` command, the top image/gif always blink when first loading it. Can you please fix that?
- Why do the `Create Workout` command directs to a list and not the form itself? Shouldn't the list have its own command?
- Why there's an empty parenthesis here?
CleanShot 2024-07-16 at 14 38 30

- What is the goal of `Manage Workouts` command? I was expecting being able to add exercises to my workout or start it.
- What does "Do Workout" mean? Shouldn't it be something like "Start Workout?"
- Selecting an workout with 0 exercises gives me the following error:
CleanShot 2024-07-16 at 14 40 11

- On exercises, if the exercise is not on your favorites, the "Remove From Favorites" action should not be shown
- This screen should be similar to the one you get after "starting" the workout. 
CleanShot 2024-07-16 at 14 42 59

- What should be the workflow of a workout? I didn't find a way to mark each exercise as done for instance.
- This page can be better formatted.
CleanShot 2024-07-16 at 14 44 02

- Have you considered using our icons instead of custom ones?  https://developers.raycast.com/api-reference/user-interface/icons-and-images

Please, take a look at my comments and tell me what you think.

ajk68:
Hi Milena,

Thanks for your insightful feedback. Here's my reaction below:


   -

   On the Find Exercice command, the top image/gif always blink when first
   loading it. Can you please fix that?

   i noticed as well and struggled to fix. It is due to unknown size of
   image when loaded. When i wait until image is download, the page will load
   slow. I could not find another way to predefine width and height in
   raycast. Is there a way to do this in raycast markdown?


   -

   Why do the Create Workout command directs to a list and not the form
   itself? Shouldn't the list have its own command?

   I had this first, but i thought it would be useful to avoid recreating a
   workout again if it already exists. But you are right, this is not a good
   solution. I will try to rework this to make it more elegant. Is there a
   good design pattern for this? Any extension that you can recommend to look
   at?


   -

   Why there's an empty parenthesis here?
   CleanShot.2024-07-16.at.14.38.30.png (view on web)
   <https://github.com/user-attachments/assets/43d13348-4ded-4882-a6d9-90cd4fe8fca0>

   That's a bug. Will fix it.
   -

   What is the goal of Manage Workouts command? I was expecting being able
   to add exercises to my workout or start it.

   This is to change the order of exercises in a workout (see the actions)
   and to delete exercises from a workout. Maybe i should combine it with
   create workout?
   -

   What does "Do Workout" mean? Shouldn't it be something like "Start
   Workout?"

   You're totally right! I'm non native english. Wasn't happy with Do, but
   didn't come up with Start. Will change.
   -

   Selecting an workout with 0 exercises gives me the following error:
   CleanShot.2024-07-16.at.14.40.11.png (view on web)
   <https://github.com/user-attachments/assets/76b8bb51-05f3-4a0a-a635-b1b213263d9f>

   I will fix this. Starting a workout with 0 exercises doesn't make sense.
   -

   On exercises, if the exercise is not on your favorites, the "Remove From
   Favorites" action should not be shown

   Yes i will fix.
   -

   This screen should be similar to the one you get after "starting" the
   workout.
   CleanShot.2024-07-16.at.14.42.59.png (view on web)
   <https://github.com/user-attachments/assets/d32208b8-5b9a-48fa-847b-a660a3347e8a>

   i'm not sure what you mean. can you please explain?
   -

   What should be the workflow of a workout? I didn't find a way to mark
   each exercise as done for instance.

   The the workflow is basically progressing through each exercise until
   the end. There's not a timer or a number of reps (at least not for now, not
   to overcomplicate). But i could change the label on the primary action to
   "mark as done" instead of "next".

   This page can be better formatted.
   -

   CleanShot.2024-07-16.at.14.44.02.png (view on web)
   <https://github.com/user-attachments/assets/0afcc194-1615-457a-80bb-a365862239e1>

   Normally, a workout consists of many exercises (usually between 6 to 12
   exercises). Then the page would make more sense. Do you have any specific
   suggestions?
   -

   Have you considered using our icons instead of custom ones?
   https://developers.raycast.com/api-reference/user-interface/icons-and-images

   I have tried, but i couldn't find any exercises related icons. I did use
   the progres icon in the workout action (see screenshot). Also, for the
   exercise description there's standard emoji's in the text as generated by
   LLMs, so i used the standard emojis for the icons as well to make it more
   consistent with the exercise text. Is that OK?


   Looking forward to your feedback. Please note, i'm on holiday until
   August 5, so can only start working on the extension in two weeks.


   Bye, Arthur

PR #13478, opened 2024-07-14
Add lexicon-search extension

futurGH:
## Description

This extension makes it easy to look up [ATproto lexicon](https://github.com/bluesky-social/atproto) documentation.

## Screencast

lexicon-search-3

## Checklist


mil3na (changes requested):
Hi!

- Lexicon is a very broad term, can you change your extension name to point that this is a documentation search for ATproto?
- I was not able to use your extension. This screen was loading for more than 30 seconds. Can you please check what is going on so I can proceed with the review?

CleanShot 2024-07-16 at 15 32 06

futurGH:
Hey, thanks for your time! I've changed the extension name to "ATproto Lexicon Search". Loading the documentation can take a minute or two the first time the extension is run — I've added a description to the EmptyView making that clearer. Let me know if anything else is needed!

mil3na:
Hi! Thanks for addressing my comments. However, I cannot approve an extension that takes more than a minute to load. Isn't there a way to make it faster? Pagination, etc?

futurGH:
Got it, made some tweaks to hopefully speed things up a bit!

mil3na:
This is still too slow. Even the items loading one by one feels slow. Also, the empty state still says that it takes 1-2 minutes to load. 
Why loading text is so slow? If you check other documentation extensions, they don't have this delay.

futurGH:
I'm not sure, honestly. It's about 300 KB of JSON documentation — it may just come down to the user's network speed. Here's what it looks like on my end, for reference.

https://github.com/user-attachments/assets/1901d5bc-07f3-4575-b21b-eff16c7d81b1

futurGH:
The latest commit should reduce the time it takes for the first list item to show up — not sure there's more I can do.

futurGH:
Yes, those items have an "Open" action to view children. As for a loading indicator, my understanding was that that was the purpose of the white bar displayed when isLoading=true — should I be indicating it in some other way?

mil3na (approved):
Yes, you are correct. Maybe it is something we should fix on our side, on the component. Anyway, LGTM! Thanks for your patience and improvements!

PR #13493, opened 2024-07-15
Replace readme images with WebP

LitoMore:
I used the [`webp`](https://developers.google.com/speed/webp) for converting the PNG image to WebP. The has the much smaller file size. So your readme can have a better experience on first load.

Here is my command:

```shell
webp store-logo.png -o store-logo.webp
webp header.png -o header.webp
```

|File|Before|After|
|---|---|---|
|store-logo.png -> store.logo.webp|2.71 MB|120 KB|
|header.png -> header.webp|230 KB|28.9 KB|


LitoMore:
### Additional

I think we can replace all PNG files with WebP to get a better performance on http://developers.raycast.com.

Please let me know your thoughts. I can convert those files into WebP in another PR.

mil3na (approved):
Converting the images is indeed a good idea. I will keep that in mind for when we update the images on both our manuals :)

LGTM!

Issue #13488, opened 2024-07-15
[I Don't Have Spotify] Spotify link not being validated properly


brunocbreis:
### Extension

https://www.raycast.com/sjdonado/idonthavespotify

### Raycast Version

1.80.0

### macOS Version

14.5

### Description

I pasted a Spotify link (specifically this one: https://open.spotify.com/album/3kjHLu1pL7tdY88GFwEkl6?si=oVIZ-ZbfQM2S1M7IM1f6RQ) and got an error that printed some json output saying it isn't valid. 
![idonthavespotify 2024-07-15 at 06 49 48](https://github.com/user-attachments/assets/ebd6c78b-78ed-4fb5-849a-df6934f63684)

### Steps To Reproduce

1. Have Spotify link in clipboard
2. Launch the extension

### Current Behaviour

throws error

### Expected Behaviour

validates link and shows other sources

sjdonado:
Fixed already https://github.com/raycast/extensions/pull/13494
@raycastbot close this issue

PR #13494, opened 2024-07-15
[idonthavespotify] new commands and bug fixes

sjdonado:
## Description

This PR adds new commands and restructures the extension to make link validation on the client side more flexible.
Fixes issue [#13488](https://github.com/raycast/extensions/issues/13488).
Updates the screenshots to match the latest visual updates.

## Screencast

https://github.com/user-attachments/assets/d6559f46-09a6-4807-8506-9a4616e9580f

## Checklist


mil3na (changes requested):
Hi!

Testing your extension, I tried to put a apple music link on this command and gave me this error. I would suggest to make this error more user friendly, it doesn't say much to a non-tech person. Also, if you only accept Spotify links, please state on the command "from Spotify to X", so it is not confusing.
CleanShot 2024-07-17 at 10 35 32

Also, with the same song, I get different results here. Why is that?

CleanShot 2024-07-17 at 10 37 24

CleanShot 2024-07-17 at 10 37 34

sjdonado:
> 
> Testing your extension, I tried to put a apple music link on this command and gave me this error. I would suggest to make this error more user friendly, it doesn't say much to a non-tech person. Also, if you only accept Spotify links, please state on the command "from Spotify to X", so it is not confusing. CleanShot 2024-07-17 at 10 35 32
> 
> Also, with the same song, I get different results here. Why is that?
> 
> CleanShot 2024-07-17 at 10 37 24
> 
> CleanShot 2024-07-17 at 10 37 34

Hi Milena, thank you for the review! Regarding the first screenshot, the Reverse Search is only available for YouTube Music links, but I plan to include Apple Music and SoundCloud later this week. I will fix the error to make it more user-friendly, but renaming the commands seems a bit misleading.

I will check the misbehavior in the second and third screenshots. It is 100% related to the server response, so it is not a blocker for the changes introduced to the extension

Let me know your thoughts :)

sjdonado:
@mil3na comments addressed :)

LitoMore:
I'd prefer to implement these commands into one command. And use an action list to select the target platform when converting. You can make a preference item to let users save their most frequent platform or always move the last selection to the top.

I usually try to keep the number of commands as short as possible, because too many commands will make searching cumbersome and you will need to type more characters to hit the command you want.

What do you think?

sjdonado:
> I'd prefer to implement these commands into one command. And use an action list to select the target platform when converting. You can make a preference item to let users save their most frequent platform or always move the last selection to the top.
> 
> I usually try to keep the number of commands as short as possible, because too many commands will make searching cumbersome and you will need to type more characters to hit the command you want.
> 
> Especially the word "convert". There are too many tools with the convert function. When you type "convert", you may see a lot of irrelevant commands, and it is difficult to locate the result you want.
> 
> On the other hand, too many commands will also increase the file size of the plugin exponentially. Maybe this is not a big burden for us, but we still need to pay attention to it.
> 
> What do you think?

Good point. The idea of having multiple commands is to only activate the ones you frequently use (and to attach to them a hotkey). I can disable all of them by default and add instructions in the README to make this clear. What do you think?

On the other hand, the main command `search` already performs a general search (including all links), so an action list is not necessary.

LitoMore (approved):

sjdonado:
@mil3na waiting for your approval :)

mil3na:
I tried your extension again. Got a Spotify link converted to Deezer. Then, tried to convert back to Spotify and I got an "invalid link" response. 
It is very confusing which conversions I can and cannot do. Maybe you should make it clear on your README? Also, the only command that was disabled by default was the "convert to. Tidal".

sjdonado:
> I tried your extension again. Got a Spotify link converted to Deezer. Then, tried to convert back to Spotify and I got an "invalid link" response. It is very confusing which conversions I can and cannot do. Maybe you should make it clear on your README? Also, the only command that was disabled by default was the "convert to. Tidal".

Command descriptions have been updated to clearly indicate which links can be converted 264d71d.

> Also, the only command that was disabled by default was the "convert to. Tidal".
To see these changes you have to uninstall and reinstall the extension.

Please let me know if everything works fine on your local. :)

mil3na (approved):
LGTM!

PR #13515, opened 2024-07-17
[Just Delete Me] unified view + chore + send email

xmok:
## Description

- Running `Show Details` for each list item to view details seemed like unnecessary overhead to me so I added a new Unified view that shows details using the List.Item.Detail. However, in case there is disagreement, I have added a new Preference checkbox that one can use to toggle the new view. If the new View seems better, I can remove the Preference and replace otherwise this is OK.
- updated markdown to show `Send email` for supported items
- updated deps
- updated metadata images to show both views

## Screencast

BEFORE

https://github.com/user-attachments/assets/46f72fd3-4bec-4b9c-864b-c7477098675d

AFTER

https://github.com/user-attachments/assets/96ac9ea4-ae56-4f54-a6ea-db9d0a1bb5d8

## Checklist


mil3na:
Hi!

I believe that the split view should be the default, not a preference. It looks way better and saves a click/press. What do you think, @andreaselia?

andreaselia:
Hey, this looks great, thanks for contributing!

I think we should make the split view the only view, I see no reason why anyone would prefer to have to make 2 actions instead of 1, and it also simplifies the extension codebase.

xmok:
Glad to hear it!
I've updated the extension accordingly:
- remove Preference
- remove old view metadata images

andreaselia:
Looks good to me @mil3na 👍

mil3na (approved):
Awesome, LGTM!
Thanks, guys!

PR #13518, opened 2024-07-17
[Session] Add new session form command

benkates:
## Description

Add a new form-based command to create a new session. @jamesjlyons

## Checklist


mil3na:
Hi!

I have two minor things about this change:
- What is the unit of time for the duration? Maybe it would be best to put a info by the input or, better yet, some pre-selected times.
- Are this categories synced with Session? If so, shouldn't this be setup and retrieved from the app?

Other than that, it looks good to me. @jamesjlyons what do you think?

benkates:
@mil3na Thanks Milena!

> * What is the unit of time for the duration? Maybe it would be best to put a info by the input or, better yet, some pre-selected times.

Minutes! I think the beauty of this productivity app is that you can deeply customize, so I'm thinking it might be best to leave it as freetext, happy to hear other suggestions though!

> * Are this categories synced with Session? If so, shouldn't this be setup and retrieved from the app?

Unfortunately there is no way to retrieve data from Session via an API or something like that at this time :/ so this solution, with the semicolon-separated list in preferences, seems to be the best right now. Let me know if there are other solutions that folks have implemented for something like this!

mil3na:
- I don't mind the free text, my concern is the user not knowing which unit of time they should use. So, some kind of indication on what to input would be appreciated.

- If the categories are not synced with the app, what is the goal of having them?

benkates:
> * I don't mind the free text, my concern is the user not knowing which unit of time they should use. So, some kind of indication on what to input would be appreciated.

Yep I'll add that after we gather all feedback!

> * If the categories are not synced with the app, what is the goal of having them?

In the Session app UI, the user selects a category for the intention like a particular project, area, etc. In the app's analytics view it summarizes activities/time spent by category. This information is not exposed via an API but is accepted in the URL scheme. So when creating the session in Raycast, I'd like to be able to tell Session what category the intention belongs to.

You can [how this works on this documentation page](https://www.stayinsession.com/learn/session-url-scheme).

jamesjlyons:
@benkates Thanks for contributing to this! Much appreciated 🙌

One of the reasons I didn't include the categories was because we can't pull them via an API, but it is a big downside to not have them accessible. I like how you're giving us the option to bring them in. Only concern from a UX perspective is that it could be easy for the list to fall out of sync with the main app. But since we have no other option, I think it might be the only good way to handle it!

I'm good with this addition. I'm curious though, did you explore adding these as optional arguments to the main "Start New Session" command? Not sure if it would work, but it could reduce confusion around having multiple new session commands.

benkates:
Thanks @jamesjlyons - great to hear from you. Agreed on bad UX but hoping us powerusers can keep them in sync ourselves!

I love the idea of removing the form alltogether and using the [command arguments interface](https://developers.raycast.com/information/lifecycle/arguments)!! What do you think about the three arguments being: intent (text), category (dropdown), and duration (text)?

jamesjlyons:
@benkates Ah I see. Yeah in that case the form will don!

How do you feel about changing the title of the command to something like "Start Custom Session"? Not sure if "with form" will be clear to users what is offers before trying the command.

benkates:
@mil3na added the unit of time
@jamesjlyons changed the command name

Let me know if ya'll have any other feedback!

jamesjlyons:
@benkates nice! I've been thinking, and curious what you think, could this new command possibly just replace the "Start new session" command? It really just adds one extra cmd + enter if a user wants to use the defaults.

I haven't been using this extension much lately, so wondering what you think as someone who's been using it more.

mil3na (approved):
LGTM!  @jamesjlyons are you okay if the changes?

benkates:
@jamesjlyons I hear that - I'm always one for more options/customization, the user can always disable one or the other if they have a preference (I do this for 75% of the raycast extensions I use) - so I think keep as-is in the PR!

jamesjlyons:
@benkates Having the option makes sense! Thanks again for contributing to this 🙌

@mil3na Good to go for me!

PR #13527, opened 2024-07-18
Update claude extension

alejandrotercero:
## Description
Added support for the beta 8192 output for sonnet. 3.5 (added as an option on the extension preferences) 
## Checklist

### DIFF
--- a/extensions/claude/CHANGELOG.md
+++ b/extensions/claude/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Claude Changelog
 
-## [Update] - {PR_MERGE_DATE}
+## [Update] - 2024-08-08
 
 - Feature: Enabled Beta [8192 tokens output limit](https://x.com/alexalbert__/status/1812921642143900036) for Claude 3.5 Sonnet
 


alejandrot96:
About the constants, yeah we could do it that way but I think you can only use one beta header at a time with the api, in the end it depends on how Anthrophic implements the longer outputs when it's not in beta anymore or if they add other beta features.

florisdobber:
I think the `useBetaFeatures` preference flag is overkill for this functionality. Removing the flag should simplify both the implementation and usage. Can you update the PR to reflect those changes? So the beta header is always used when Sonnet 3.5 is selected and keep the code for allowing users to select up to 8192 tokens for Sonnet 3.5. I'm indifferent about the constants, that's your call.

I'm assuming this will become part of the standard API soon, so I want to minimize the changes made. 

Besides that, thanks for the quick spot and implementation! 🚀 (I also follow @alexalbert__ now)

alejandrot96:
Works for me! I'll make the changes later today

florisdobber:
Hey, it seems like not all references to the `preference` flag have been removed. Until those are cleaned up I don't think we should merge this PR.

alejandrot96:
Is this good now? 😅

mil3na (approved):
Sure! You should have pinged me earlier :)

PR #13545, opened 2024-07-18
Replace all docs PNG image with WebP

LitoMore:
Related to #13493

I converted all `.gitbook/assets/*.png` files to WebP.

BTW, the `extension fix / improvement` and `extension: doppler-share-secrets` labels can be removed from this PR.


LitoMore:
Sure. I've checked all markdown files in the GitBook. But I'm not sure if you are using these files anywhere else.

PR #13566, opened 2024-07-19
Add auto-source-switcher extension

MishaIvasyutyn:
## Checklist


pernielsentikaer:
Hi 👋

Sorry for the long response time, I lost track of it 🫠

We already have a [similar extension](https://www.raycast.com/danulqua/input-source-switcher) in the store, could you check that one out?

MishaIvasyutyn:
Hi there, thank you for your response! 

Let me clarify the difference! The Input Source Switcher and the auto-source-switcher extension are different. The auto-source-switcher is an upgraded version of the Input Source Switcher. It automatically recognizes the language layout and switches both the text and keyboard layout to the correct language.

In contrast, the Input Source Switcher only enables you to manually select one language to switch from (e.g., from English to Ukrainian). With the auto-source-switcher, you don’t need to manually select the language each time you make a mistake; it works automatically.

But you are right on one thing. I actually made two different extensions in terms of the value to user, but they have a little different code base, so that's another pull request  https://github.com/raycast/extensions/pull/13542. And to predict your next question—why did I create two different extensions if they both do the same thing? The answer is that while the other extension is more optimized for CPU efficiency, it might have some bugs. I hope this clears up any confusion

pernielsentikaer:
What do you think @danulqua here, for me it more sounds like an enhancement, there is also https://www.raycast.com/lucaschultz/input-switcher which is in the same alley

danulqua:
@MishaIvasyutyn Hi! Should the "Switch Input Source" command transform the selected text? It does not on my end, only the input source switches.


https://github.com/user-attachments/assets/7f3a8faf-2af0-417b-b561-b1775d675f2f

MishaIvasyutyn:
> @MishaIvasyutyn Hi! Should the "Switch Input Source" command transform the selected text? It does not on my end, only the input source switches.
> 
>  CleanShot.2024-08-13.at.13.26.23.mp4

Oh, yeah, strangely I quaff since the last Raycast update my auto-source-switcher extension just stopped working. However, my other extension, whose pull request was recently closed, still functions good. 
https://github.com/raycast/extensions/pull/13542

https://github.com/user-attachments/assets/e1abd7a0-aedb-44e9-8be5-253777249cbd

pernielsentikaer:
I'm still a bit confused about what the difference is and why it can't be included in one of the other extensions

danulqua:
@pernielsentikaer Agree. Purpose and goal are the same.

PR #13578, opened 2024-07-20
Update passphrase-generator extension

iwex:
## Description

Enhance the dictionary words by adding capitalization and numbers to make it universally applicable for password requirements.

## Checklist

### DIFF
--- a/extensions/passphrase-generator/CHANGELOG.md
+++ b/extensions/passphrase-generator/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Passphrase Generator
 
-## [Improve] - {PR_MERGE_DATE}
+## [Improve] - 2024-08-09
 
 - Enhance the dictionary words by adding capitalization and numbers to make it universally applicable for password requirements.
 


omBratteng (commented):
I think it is better to make it an option to have the words capitalised.

And for numbers, either an option to add the number at the end, or convert words to l33t (lite) speak. I.e. `somaplasm-savagize-vaticinant` -> `s0m4pl4sm-s4v4g1ze-v4t1c1n4nt`. As this is based on the [XKCD 936](https://xkcd.com/936/) Correct Horse Battery Staple generator idea.

pernielsentikaer:
Thanks for the review, @iwex can you have a look 🙂

iwex:
I don't think it's necessary to make many different generators with options and sub-options. The solution should be simple for end users. Most sites or applications ask for such password rules: 1 uppercase letter, 1 lowercase letter, a number, and a special character. I believe that the case when we capitalise letters, add dashes and numbers is ideal for any usecase.


If we want a complex set of letters, we already have such a generator - ASCII characters with punctuation

pernielsentikaer:
I'm up for sticking with how it is now, what do you think @omBratteng

iwex:
> I'm up for sticking with how it is now, what do you think @omBratteng

Why?

pernielsentikaer:
I agree with you @iwex 🙂

iwex:
But it may be worth adding only one digit as Bitwarden.
https://bitwarden.com/password-generator/
0r1HbwFm@2x

pernielsentikaer:
Fine with me, at least

omBratteng:
> But it may be worth adding only one digit as Bitwarden or Keeper. [bitwarden.com/password-generator](https://bitwarden.com/password-generator/) 0r1HbwFm@2x
> 
> [keepersecurity.com/features/passphrase-generator](https://www.keepersecurity.com/features/passphrase-generator/) dtWXSpMw@2x

Doing it like that is fine. Can do it to the pronouncable generator too, not just dictionary.

iwex:
@omBratteng 
Pronouncable extends Dictionary, so it will work the same way
xFhFiOFb

omBratteng:
Ah, good.

iwex:
@pernielsentikaer

pernielsentikaer:
Let's get it merged, thanks for flagging 🔥

PR #13661, opened 2024-07-25
Add app-icon-generator extension

hunter-ji:
## Checklist


pernielsentikaer (changes requested):
Hi 👋

Thanks for an awesome extension 🔥

Would it make sense to send the user back to root of Raycast after processing the files instead of staying in the view?

hunter-ji:
Thank you for your valuable feedback regarding the navigation flow. I appreciate your suggestion about returning the user to the root of Raycast after processing the files.

After careful consideration, I agree that improving the user experience is crucial. However, I've made some adjustments that I believe will better serve our users:

1. Progress Visibility: Instead of closing the extension during processing, I've implemented a system where the entire icon generation process is visible within the extension page. This allows users to see real-time progress and provides immediate feedback on the operation.

2. Detailed Status Updates: The extension now shows step-by-step updates during the icon generation process, keeping users informed about what's happening at each stage.

3. Clear Completion Indication: Upon successful generation, the extension displays a completion message along with the location of the generated icons, all within the same view.

4. Action-Based Navigation: I've added an action button that allows users to open the output folder directly from the extension. When this action is triggered, the extension automatically returns to the Raycast root, aligning with your initial suggestion.

5. User Control: This approach gives users more control and visibility over the process while still providing an easy way to return to the Raycast root when they're done.

I believe these changes strike a balance between keeping users informed and maintaining efficient navigation within Raycast. The automatic return to the root after opening the output folder ensures that users can easily start new tasks in Raycast once they've accessed their generated icons.

Thank you again for your insights. I'm open to any further suggestions or feedback you might have regarding these changes.

new page

pernielsentikaer:
The only thing that's annoying at the moment is that it pushes to the new view, so hitting escape goes back to the form. Maybe we can return to the other view instead somehow?

hunter-ji:
So now the progress is all displayed on one page. When the event processing is completed, the generated folder is automatically opened, and the raycast home page is returned and closed. It's all very smooth.

During event processing, users are prevented from submitting events repeatedly.

pernielsentikaer:
I added another cool idea where you can set your default platforms in preferences, could you check that out?

hunter-ji:
Yes, that's cool!

hunter-ji:
Thanks for the idea, this is really cool. I ran it and it works fine.

I will add this capability to all my future raycast extensions.

image

hunter-ji:
Thank you for your cool ideas, it was really an interesting experience.

PR #13673, opened 2024-07-25
Add OAuth flow to Dub extension

devkiran:
## Description

Add OAuth flow to Dub extension

## Screencast
CleanShot 2024-07-29 at 22 55 14@2x

CleanShot 2024-07-29 at 22 55 42@2x

CleanShot 2024-07-29 at 22 56 24@2x

CleanShot 2024-07-29 at 22 55 53@2x

## Checklist

### DIFF
--- a/extensions/dub-link-shortener/CHANGELOG.md
+++ b/extensions/dub-link-shortener/CHANGELOG.md
@@ -1,7 +1,7 @@
 # Dub Link Shortener Changelog
 
 
-## [OAuth flow] - {PR_MERGE_DATE}
+## [OAuth flow] - 2024-07-30
 
 - Added OAuth flow to connect to Dub account
 


jfkisafk:
Hey @devkiran looks like you're already on top of this, but you can use this PR for reference: https://github.com/raycast/extensions/pull/13403

extensions/dub-link-shortener/src/api/oauth.ts
```diff
@@ -0,0 +1,38 @@
...
+
+export const dubOAuth = new OAuthService({
+  client,
+  clientId: "dub_app_cf2e6f232bd9b1ca5918bc130b61ca127c2f1557ffcc95c4",
+  scope: ["workspaces.read", "links.write", "tags.read", "domains.read"].join(" "),
+  authorizeUrl: "https://app.dub.co/oauth/authorize",
+  tokenUrl: "https://api.dub.co/oauth/token",
+  refreshTokenUrl: "https://api.dub.co/oauth/token",
+  bodyEncoding: "url-encoded",
+  personalAccessToken: workspaceApiKey,
```
jfkisafk, line 24: We can totally just remove the workspaceKey from preferences if that is not required anymore and stick to OAuth. What do you think?

```suggestion
```

extensions/dub-link-shortener/CHANGELOG.md
```diff
@@ -1,6 +1,10 @@
 # Dub Link Shortener Changelog
 
 
+## [OAuth flow] - 2024-07-25
```
jfkisafk, line 4:
```suggestion
## [OAuth flow] - {PR_MERGE_DATE}
```

extensions/dub-link-shortener/src/api/oauth.ts
devkiran: We discussed this and decided we should support existing users with `workspaceKey` preferences. That way, users logged through a token don't have to authenticate again.

steven-tey: I'm okay with removing it completely and prompt users to login with Dub, but maybe we want to keep it for backwards compatibility?

jfkisafk: I mean both approaches make sense to me. What I was thinking was if we remove it, then users will be prompted to logIn anyways (one-time) and that way we can manage the permissions in the code-base, rather than rely on users to define granular permissions and run into user issues. Especially when the downloads are low atm - that's what I did with Github Gist.

pernielsentikaer:
@jfkisafk is this ready for a merge? 🙂

jfkisafk:
Hey @pernielsentikaer, it isn't ready. I was getting a `redirectUrl is invalid` error when I tried these changes. @steven-tey @devkiran can you please take a look? If dub.co OAuth provider does not support queryParams, we can make these changes:

> If the provider does not accept query parameters in redirect URLs, you can alternatively use https://raycast.com/redirect/extension and then customize the [AuthorizationRequest](https://developers.raycast.com/api-reference/oauth#oauth.authorizationrequest) via its extraParameters property. For example add: extraParameters: { "redirect_uri": "https://raycast.com/redirect/extension" }

devkiran:
@jfkisafk Thanks for testing/reviewing. The OAuth app was using the wrong URL; I updated it with `https://raycast.com/redirect?packageName=Extension`, and the flow is working now.

extensions/dub-link-shortener/src/api/oauth.ts
jfkisafk: So are we keeping both versions going forward @devkiran @steven-tey ?

jfkisafk (approved):
Just need minor clarity on the approach going forward: are we keeping both `workspaceKey` and OAuth or just sticking to OAuth @devkiran @steven-tey . Otherwise LGTM.

extensions/dub-link-shortener/src/api/oauth.ts
steven-tey: okay yeah sounds good to me – let's remove the workspaceId flow completely cc: @devkiran

steven-tey:
> Just need minor clarity on the approach going forward: are we keeping both `workspaceKey` and OAuth or just sticking to OAuth @devkiran @steven-tey . Otherwise LGTM.

We will remove the workspaceKey and stick to just OAuth since the download count is relatively low atm 👍

steven-tey:
Okay this PR should be ready for review now! Just went through the whole flow and everything was super smooth 😁 

CleanShot 2024-07-29 at 22 55 14@2x

CleanShot 2024-07-29 at 22 55 42@2x

CleanShot 2024-07-29 at 22 56 24@2x

steven-tey:
Okay this PR should be ready for review now @pernielsentikaer! Just went through the whole flow and everything was super smooth 😁 

CleanShot 2024-07-29 at 22 55 14@2x

CleanShot 2024-07-29 at 22 55 42@2x

CleanShot 2024-07-29 at 22 56 24@2x

pernielsentikaer:
While we're at it, would it make sense to add a action panel action to show [tags](https://dub.co/docs/api-reference/endpoint/retrieve-a-list-of-tags) (list, delete, create)

At least in the first version add that OR add info="Add tags in Dub.co"

steven-tey:
@pernielsentikaer Great idea! Would it be okay if we keep that to a separate follow-up PR so we can get the OAuth flow out there before more folks download the extension? 👀

jfkisafk:
@steven-tey I think what @pernielsentikaer means is that we add `info="Add tags in Dub.co"` here:

https://github.com/raycast/extensions/blob/0e6c25180227b5ff5d4f4c61c980067083ef6580/extensions/dub-link-shortener/src/shorten-link.tsx#L92

that way it will show the info on the side when user hovers on top. I think action to add, delete, list tags should be handled as part of a separate PR.

cc: @devkiran

pernielsentikaer:
Do we have all the right scopes? Would be best to have them all added now instead of later since people then need to re-auth

extensions/dub-link-shortener/src/api/oauth.ts
```diff
@@ -0,0 +1,36 @@
...
+  providerName: "Dub",
+  providerIcon: "command-icon.png",
+  providerId: "dub",
+  description: "Connect your Dub account",
+});
+
+export const dubOAuth = new OAuthService({
+  client,
+  clientId: "dub_app_d70ed8f23a7eb32a0532776e5de0be08796cc82e12a3c806",
+  scope: ["links.write", "tags.write", "workspaces.read", "domains.read"].join(" "),
```
steven-tey, line 18: Oh gotcha, sorry I misunderstood, thanks for clarifying! 🙏 

@pernielsentikaer yes, we've added all the necessary scopes and also included `tags.write` for now if we do want to add the ability to create/update tags in the future:

steven-tey:
This PR should be ready to merge! 🙏

jfkisafk (approved):

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

steven-tey:
Thank you so much @pernielsentikaer! 🙏

PR #13679, opened 2024-07-25
[Bitwarden] Ensure search vault section order

### LINKED ISSUE #13487: [Bitwarden Vault] When searching the vault, Other Items displayed before Favorites
### Extension

### Raycast Version

1.78.1

### macOS Version

14.5

### Description

First of all, I'll mention that I'm putting this as an issue and not a feature request because I've noticed some inconsistencies with this behavior, so I'm not exactly sure if it is a bug or this is how this fuzzy search implementation is supposed to work.

Up until lately, when I typed into the search bar filtered items would be always shown with Favorites first and Other Items after that. However, now in some cases, Other Items are shown before the Favorites, which is very inconvenient, since in 99% of cases the item I'm looking for is in my favorites, so now I have to scroll all the way down to find it, while before it was always the first or just a couple of rows down.

It still works 'correctly' when the searched term is in the beginning of items' names:
Screenshot 2024-07-15 at 09 49 30

But differently when items just contain the search term (typo in a search term to limit the amount of items in the list):
Screenshot 2024-07-15 at 09 54 34

Not sure when this happened exactly, nor do I seem to find any commits related to this that would explain this change.

Tried re-installing both the extension and bitwarden-cli.

### Steps To Reproduce

See description above for screenshots

### Current Behaviour

Favorites are sometimes listed below Other Items

### Expected Behaviour

Favorites are always shown first

jomifepe:
## Description

Resolves #13487

## Checklist

### DIFF
--- a/extensions/bitwarden/CHANGELOG.md
+++ b/extensions/bitwarden/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Bitwarden Changelog
 
+## [Fix] - 2024-07-25
+
+- Prevent search vault sections from changing order
+
 ## [New Commands] - 2024-04-10
 
 - Added 3 Send commands (Search, Create and Receive) with delete, edit and remove password actions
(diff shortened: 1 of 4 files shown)


pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

Issue #13701, opened 2024-07-26
[Brew] ... please add more terminal options


edheltzel:
### Extension

https://www.raycast.com/nhojb/brew

### Description

The current options are limited to 3: item, terminal, and warp. It would be nice to have other options or to be able to select whatever app I want. 

### Who will benefit from this feature?

everyone who is not using iterm, terminal, or warp – Alacritty, Kitty, and WezTerm are very popular options for macOS

PR #13744, opened 2024-07-29
[Brew] Adds more terminal options and minor refactoring

jfkisafk:
## Description

- Added new terminal options for Alacritty, kitty, WezTerm and Hyper.
- Used app icons for actions to run commands in terminals.
- Apply Destructive style to Uninstall actions.

## Checklist

* closes https://github.com/raycast/extensions/issues/13701

### DIFF
--- a/extensions/brew/CHANGELOG.md
+++ b/extensions/brew/CHANGELOG.md
@@ -5,6 +5,7 @@
 - Added new terminal options for Alacritty, kitty, WezTerm and Hyper.
 - Used app icons for actions to run commands in terminals.
 - Apply Destructive style to Uninstall actions.
+- Fallback to Terminal.App in case the selected terminal app is not installed.
 
 ## [Improvements] - 2024-06-14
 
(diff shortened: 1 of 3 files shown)


Aayush9029:
LGTM

jfkisafk:
> The only thing that worries me is hardcoded path

Thanks @Aayush9029! Yeah, I think most of these apps will be in Applications folder. And in the worst/rare case scenario, if the app icon is not found the image will not be rendered (it will **not** cause the command to crash).

nhojb (approved):
Looks good. Thanks for the contribution 👏 

I don't think the hardcoded app paths are a big problem since we don't check whether the terminal is installed anyway 😞

A future improvement (not necessary for this PR), would be to check at runtime if the selected terminal app is installed, and if not fallback to Terminal.app for the actions.

jfkisafk:
> A future improvement (not necessary for this PR), would be to check at runtime if the selected terminal app is installed, and if not fallback to Terminal.app for the actions.

Good suggestion @nhojb! I added it to this PR itself. Please take a look when you have some time.

PR #13768, opened 2024-07-30
Update base64 extension

VitorLostada:
## Checklist


sammarks:
Perhaps this would be better inside the **image** encode / decode extension, rather than this one? And just modify that one to support image formats other than SVG? The whole point of the UX on this text-based extension is a quick clipboard to clipboard encode or decode, in one step.

Then there are 2 Base64 extensions in the store, each tailored to their individual purpose. One for text, and one for images and other rich media.

It would make sense to have the clipboard reader in the existing extension support images / SVG as well, instead of just reading text from the clipboard.

A lot of times (especially when working with GraphQL) I am doing quick encode / decode operations and to add more steps to this really hurts the extension's usability to me.

It's also very helpful to have the live preview of the encoded / decoded content there without having to copy it to my clipboard first because oftentimes I'm just trying to get the internal database ID from a GraphQL ID, and this replacement doesn't have that functionality (it's annoying to have to find somewhere to paste the decoded result just to be able to read it).

mil3na:
Hi! Your extension is currently not passing all the checks. I will mark it as a draft. Please, mark as "Ready for review" as soon as you fix it :)
Cheers!

vitorLostadaC:
@sa

> Perhaps this would be better inside the **image** encode / decode extension, rather than this one? And just modify that one to support image formats other than SVG? The whole point of the UX on this text-based extension is a quick clipboard to clipboard encode or decode, in one step.
> 
> Then there are 2 Base64 extensions in the store, each tailored to their individual purpose. One for text, and one for images and other rich media.
> 
> It would make sense to have the clipboard reader in the existing extension support images / SVG as well, instead of just reading text from the clipboard.
> 
> A lot of times (especially when working with GraphQL) I am doing quick encode / decode operations and to add more steps to this really hurts the extension's usability to me.
> 
> It's also very helpful to have the live preview of the encoded / decoded content there without having to copy it to my clipboard first because oftentimes I'm just trying to get the internal database ID from a GraphQL ID, and this replacement doesn't have that functionality (it's annoying to have to find somewhere to paste the decoded result just to be able to read it).

Hey @sammarks sorry for the delay, it was a chaotic week hahaha😅

You’re right; I’ve updated the UI so that all tools now have a consistent interface. However, this might not provide the best developer experience. I’m not keen on creating a separate extension because the current one has a very generic name, “base64,” and I’d like to add more functionalities to it.

So, to ensure everyone is satisfied, here’s my suggestion:

- I’ll modify all tools to retrieve text from the clipboard.
- I’ll rename some options to be more descriptive:
encode => text to base64
decode => base64 to text
index => base64 realtime or base64 realtime text
- I’ll implement new actions:
image to base64
base64 to image

Is this okay? If so, I’ll begin development. 😁

sammarks:
May want to have @jfkisafk take a look at this as well since they just created a PR improving the codebase of the existing extension.

As far as I'm concerned, as long as you're preserving the initial "quick" functionality of the existing extension exactly as it is today (or extending it to support more types of data, but within the same interface), feel free to make modifications.

The original PR had completely replaced the existing extension with something else, and I'd prefer you **enhance** the existing extension rather than replace it.

jfkisafk (changes requested):
Hello @vitorLostadaC , I'd like to echo what @sammarks said regarding not changing the existing extension functionalities and only adding functionalities for image conversion. That said, I'd like to highlight a few things with the latest proposition:

> I’ll modify all tools to retrieve text from the clipboard.

Let's hold on this PR until https://github.com/raycast/extensions/pull/13844 has been merged to avoid any duplicate work.

> I’ll rename some options to be more descriptive

Renaming the command(s) will erase all aliases and shortcuts users might have already setup, and I'd not recommend it for an extension downloaded 16k+ times. Let's leave the commands as they are and add the image support as part of feature flags through preferences and new List items (see next comment for proposal).

> I’ll implement new actions: image to base64, base64 to image

I like the idea of new commands, but it might set wrong precedence (unnecessarily overpopulate the extension) if for specific purposes, we add separate commands. I'd argue why we even need a new form based interface when we can just handle images through the Base64 command? When the user enters an image base64 in the search field (or it's already in their clipboard), we can just parse it and dynamically display a new list item for Image (in addition to Encode + Decode items) to show the image as list detail metadata or a new detail view.

PR #13770, opened 2024-07-30
[MapleStory.gg] Use WebP icon

LitoMore:
## Description

This pull request is similar to #13545. I changed my extension icon from PNG to WebP. This can save some disk spaces for users.

I also want to try if this can work properly. Which means we can convert all extension icons from PNG to WebP to get a better performance.

This CI fails. The linting part from Ray CLI seems not open-source. So I can't help to update it.

## Checklist

### DIFF
--- a/extensions/maplestory-gg/CHANGELOG.md
+++ b/extensions/maplestory-gg/CHANGELOG.md
@@ -1,5 +1,9 @@
 # MapleStory.gg
 
+## [Optimization] - {PR_MERGE_DATE}
+
+- Use WebP icon
+
 ## [Initial Version] - 2024-05-29
 
 - Add lookup characters feature


mil3na:
Hi! Your extension is currently not passing all the checks. I will mark it as a draft. Please, mark as "Ready for review" as soon as you fix it :)
Cheers!

LitoMore:
@mil3na As I mentioned in decription, I can't fix this since the error is from Ray CLI. I just want to let your team notice the WebP stuff so I created this PR.

I am looking forward to the WebP support for icons and metadata screenshots.

mil3na:
To be fair , I don't think this change would be necessary. Icons are already super small and you wouldn't see much change on only one icon. @pernielsentikaer what do you think?

mil3na:
Are you having performance issues on loading icon images?

LitoMore:
I noticed that your web store display page has processed the images into WebP before displaying them. If we use WebP directly in the source files, it can also save you the work in this regard.

mil3na:
It will not make the repo smaller tho, because Git stores history. So it would be even worse, having those 2 image formats.
This concerns more the app than the website. I understand when optimising for the website, but for the extension which will be bundled and so on, I don't think that a few hundred KB will make any difference.

LitoMore:
Now you have the opportunity to slow down its growth, isn't it good?

Whether to use WebP depends on you. For me, hundreds of KB is not a small number, and it has a lot of space for optimization.

Thank you for your reply. Let's see what other members think about this topic.

pernielsentikaer:
I agree that it might be something we'll look into in the future, but for now, we're not ready for it since I'm pretty sure that AppKit on `macOS 12 Mojave` doesn't support `.webp` out of the box, which would be the preferred way without the need for custom code. As soon as it's fully supported, it makes sense to allow that.

As stated earlier, it doesn't make the repository smaller if you clone it directly since it's in the history, but shallow cloning does, of course.

Does it make sense, @LitoMore? So, it's more of a "not right now" than a decline

LitoMore:
@pernielsentikaer Agree!

pernielsentikaer:
Thanks for your understanding, let's circle back to this as soon as possible 🙂

PR #13777, opened 2024-07-30
Add modrinth-search extension

MagnusHJensen:
## Description
Adds [Modrinth Search](https://modrinth.com) for quick search access, to discover new projects and get relevant project links to communities, wiki, source code and more..

## Screencast

https://github.com/user-attachments/assets/d857f3bc-6dce-4699-a13c-e06857def9a1

## Checklist


extensions/modrinth-search/README.md
LitoMore, line 1: Worth making a introduce to your documentation, such as features list and usages.

Users will usually read your extension read before installing it, and a good profile can make them more likely to choose your extension.

extensions/modrinth-search/package.json
```diff
@@ -0,0 +1,41 @@
...
+    {
+      "name": "search-mods",
+      "title": "Search Mods",
+      "description": "Searches mods on Modrinth.com",
+      "mode": "view"
+    }
+  ],
+  "dependencies": {
+    "@raycast/api": "^1.69.0",
+    "got": "^11.8.2"
```
LitoMore, line 24: Got 11 is a bit outdated. I understand that you choose because of the CommonJS. But you can consider updating the project to ESM.

You can refer to my project configuration to see how to use ESM:

- tsconfig.json: https://github.com/raycast/extensions/blob/main/extensions/simple-icons/tsconfig.json
- package.json: https://github.com/raycast/extensions/blob/main/extensions/simple-icons/package.json

MagnusHJensen: I've opted for the `useFetch` call inside the utils package instead, since it came with fetch and cache already built in, and functions a lot like react-query :D

LitoMore: Then the Got can be removed.

MagnusHJensen: That is my bad, I thought I ran npm uninstall, but apparently not.

It has now been removed, thank you :D

mil3na (changes requested):
Hi!

The extension in general looks good! But it seems like you are capping the results. In other words, shouldn't I be able to see all the plugins if I scrolled without a query param?

MagnusHJensen:
> 
> The extension in general looks good! But it seems like you are capping the results. In other words, shouldn't I be able to see all the plugins if I scrolled without a query param?

That is not my intention, as seen I took out the pagination and passed it to the list, but it does not work.
There is also a slack message for getting pagination to work: https://raycastcommunity.slack.com/archives/C02HEMAF2SJ/p1722365320376059 but he couldn't either.

mil3na:
Do you know if the API you are using supports pagination?

MagnusHJensen:
> Do you know if the API you are using supports pagination?

It does indeed, but the loadMore method passed into the list is never called (verified with a console.log)

mil3na:
I didn't have much time to dig into this, but I believe that the key is that there is nothing watching for when there's a "page change". Like in this line: https://developers.raycast.com/api-reference/user-interface/list#pagination:~:text=)%3B-,useEffect(()%20%3D%3E%20%7B,-cancelRef.current

Have you tried a `useEffect` hook?

MagnusHJensen:
> I didn't have much time to dig into this, but I believe that the key is that there is nothing watching for when there's a "page change". Like in this line: https://developers.raycast.com/api-reference/user-interface/list#pagination:~:text=)%3B-,useEffect(()%20%3D%3E%20%7B,-cancelRef.current
> 
> Have you tried a `useEffect` hook?

It feels odd for me listening for the `id` updating to see if it's at the end of the list.

When the docs sounds like:
```
onLoadMore - will be called by Raycast when the user reaches the end of the list, either using the keyboard or the mouse. When it gets called, the extension is expected to perform an async operation which eventually can result in items being appended to the end of the list.
```

mil3na:
Hi again! Sorry for the delay answering this. I discussed with our API engineers and turns out that we have a limitation that we can't really fix on our side. However, there is a workaround! Changing the limit of your request should make pagination work. 

My suggestion is changing the file `use-mondrith-search`, lines 43 - 55 to:

```
(options) => {
      const urlParams = new URLSearchParams();

      if (searchQuery) {
        urlParams.set("query", searchQuery);
      }

      urlParams.set("offset", String(options.page * 20));

      urlParams.set("limit", String(20));

      return `${BASE_URL}/search?${urlParams.toString()}`;
    },
```

Setting the limit to 15 (remember to update the offset) should work as well, but 20 seems fine. What do you think?

MagnusHJensen:
> Hi again! Sorry for the delay answering this. I discussed with our API engineers and turns out that we have a limitation that we can't really fix on our side. However, there is a workaround! Changing the limit of your request should make pagination work.
> 
> My suggestion is changing the file `use-mondrith-search`, lines 43 - 55 to:
> 
> ```
> (options) => {
>       const urlParams = new URLSearchParams();
> 
>       if (searchQuery) {
>         urlParams.set("query", searchQuery);
>       }
> 
>       urlParams.set("offset", String(options.page * 20));
> 
>       urlParams.set("limit", String(20));
> 
>       return `${BASE_URL}/search?${urlParams.toString()}`;
>     },
> ```
> 
> Setting the limit to 15 (remember to update the offset) should work as well, but 20 seems fine. What do you think?

If that fixes it, I'll update it to that!

Do you mind elaborating a bit more on the internal limitation, as to why a limit of 10 does not work/trigger this method?

mil3na:
Sure! We added a little note to our docs:
image

MagnusHJensen:
> Sure! We added a little note to our docs: image

Great stuff, thank you very much! I'll make the update later or tomorrow :D

MagnusHJensen:
> Hi again! Sorry for the delay answering this. I discussed with our API engineers and turns out that we have a limitation that we can't really fix on our side. However, there is a workaround! Changing the limit of your request should make pagination work.
> 
> My suggestion is changing the file `use-mondrith-search`, lines 43 - 55 to:
> 
> ```
> (options) => {
>       const urlParams = new URLSearchParams();
> 
>       if (searchQuery) {
>         urlParams.set("query", searchQuery);
>       }
> 
>       urlParams.set("offset", String(options.page * 20));
> 
>       urlParams.set("limit", String(20));
> 
>       return `${BASE_URL}/search?${urlParams.toString()}`;
>     },
> ```
> 
> Setting the limit to 15 (remember to update the offset) should work as well, but 20 seems fine. What do you think?

Pagination should now be working.

I find it odd, as my first approach used the `options.cursor` which I assumed would work, since I'm setting that in the response object, however that made it do duplicate fetches, so no.
The `options.page` did indeed work.

mil3na (approved):
LGTM! Thanks for your effort!

PR #13831, opened 2024-08-01
Update openvpn extension

julienR2:
## Description

`Openvpn` extension is very practical to automatize connectivity with vpn, however it wasn't supporting multiple profiles.

This PR brings support for multiple profiles, along with a couple of improvements and bug fix:
### ✨ New

- The extension can now handle multiple profiles!
- A couple of new command make it easy to conenct/disconnect from one of you profile, or connect/disconnect from the last used profile

### 💎 Improvements

- The apple scripts have been revamped, to avoir explicit clicking and navigation slowing down the process, and instead target directly the correct entry and click it (which then happen in the background without visible action)

### 🐞 Fixes

- `isRunning` function has been updated to rely on the existance of the menu bar (which we leverage later to connect/disconnect anyway) instead of command line checks which are less reliable

## Screencast
|New commands|Profiles list|Connected profile|
|-|-|-|
|CleanShot 2024-08-01 at 18 04 49@2x|CleanShot 2024-08-01 at 18 09 04@2x|CleanShot 2024-08-01 at 18 13 38@2x|

## Checklist

### DIFF
--- a/extensions/openvpn/src/utils.ts
+++ b/extensions/openvpn/src/utils.ts
@@ -31,7 +31,9 @@ export const isRunning = async () => {
 
 export const startOpenVPN = async () => {
   try {
+    // Start the app
     execSync('"/Applications/OpenVPN Connect/OpenVPN Connect.app/contents/MacOS/OpenVPN Connect"');
+    // Minimize it right away (running this command without the app running might fail)
     execSync('"/Applications/OpenVPN Connect/OpenVPN Connect.app/contents/MacOS/OpenVPN Connect" --minimize');
   } catch (e) {
     console.error(e);


julienR2:
Hey @benekuehn! Thanks for the work you initiated with this plugin 🙂 

It's been a while I wanted to automate connecting my vpn through OpenVPN client, but the limited API provided by their CLI put me off.. Until I saw your extension leveraging AppleScript 🙌 

Since I use multiple profile at work, I immediately jumped on it and add support for it. Of course along the way I did some changes here and there (although I tried to keep it simple and closed to your implementation). That's what pushed me to add the "connect to last profile", to keep the instant toggle available especially if you only have on profile.

Let me know what you think!

pernielsentikaer:
That's a really awesome update! I just tested it with my current setup and it works as expected 🔥

julienR2:
@pernielsentikaer Thanks for having a look 🙏 Glad it worked as expected!

Regarding the menu bar, I also though about it, but then this extension already pilots the OpenVPN Connect menu bar 😁 So it might be a bit redundant to introduce another menu bar just next to it, with the same functionality..



What do you think? Did you have a specific usage in mind for the menu bar?

extensions/openvpn/src/utils.ts
```diff
@@ -1,32 +1,123 @@
...
   } catch (e) {
     return false;
   }
 };
 
 export const startOpenVPN = async () => {
   try {
-    await execAsync('"/Applications/OpenVPN Connect/OpenVPN Connect.app/contents/MacOS/OpenVPN Connect"');
+    execSync('"/Applications/OpenVPN Connect/OpenVPN Connect.app/contents/MacOS/OpenVPN Connect"');
+    execSync('"/Applications/OpenVPN Connect/OpenVPN Connect.app/contents/MacOS/OpenVPN Connect" --minimize');
```
benekuehn, line 35: why do we need it twice?

benekuehn (approved):
thanks a lot for all these improvements ❤️

extensions/openvpn/src/utils.ts
julienR2: Ah this could have been worth a little comment there..

In a [previous update](https://github.com/raycast/extensions/commit/3685d21bae36df2b411ce3f1fb0222c1b8d16644#diff-db0f5bd5fef7e5e97a25225fc79b4d00c0478382600c3ef8ff0edf415b639519R3) you remove the flag, because it would prevent the app from starting. In deed the minimize works when the app is already open. So we start and then minimize it.. Weird code but this way we do achieve the "start minimised" 🙂 

I'll add a comment though 👍

benekuehn:
@julienR2 do you now also get extension bug reports for OpenVPN? In case you dont, there seems to be an issue, see screenshot below

Screenshot 2024-08-13 at 07 26 26@2x

julienR2:
@benekuehn ah I don't! (I'll ping the team to see if I can subscribe to those bug too, to better jump on them)

Thanks for the heads up, I'll investigate and push a fix soon 👍

julienR2:
@pernielsentikaer anyway to be subscribed to extensions errors report even though we are not the author?
Would be practical to be able to monitor them, as the new error will most probably be introduced by me.

PR #13877, opened 2024-08-05
Add phone extension

Alexk2309:
## Checklist


koinzhang:
Hi @Alexk2309 As mentioned in the README, [Quick Phone Call](https://www.raycast.com/karbassi/quick-call) and [FaceTime](https://www.raycast.com/ron-myers/facetime) extensions are already available in the shop.
Perhaps it wouldn't be more appropriate to submit PRs for these two extensions to improve them

Alexk2309:
> Hi @Alexk2309 As mentioned in the README, [Quick Phone Call](https://www.raycast.com/karbassi/quick-call) and [FaceTime](https://www.raycast.com/ron-myers/facetime) extensions are already available in the shop. Perhaps it wouldn't be more appropriate to submit PRs for these two extensions to improve them

Hi, these projects are different. I chose not to submit this PR because I would be modifying the underlying principle of their extensions. Their extensions are supposed to be able to call/facetime quickly. My design is not built on being able to do it quickly, but rather its elegance.

Additionally, these two extensions, have a lot in common it would be better to integrate phone calls/Facetime, into one project as there are a lot of similarities between these two hence reducing code repetition, which is another reason why I chose to create a new extension. Furthermore those extensions have been inactive for quite some while, so a PR would be quite hard to accomplish.

pernielsentikaer (commented):
Hi 👋

Thanks for the contribution. As you mentioned, there are already a few extensions doing similar things, but they might not be as sophisticated as yours.

Everything of this is actually already built-in inside Raycast, did you see `Search Contacts` (which might be disabled by default), that should cover all needs

image

Alexk2309:
> Hi 👋
> 
> 
> 
> Thanks for the contribution. As you mentioned, there are already a few extensions doing similar things, but they might not be as sophisticated as yours.
> 
> 
> 
> Everything of this is actually already built-in inside Raycast, did you see `Search Contacts` (which might be disabled by default), that should cover all needs
> 
> 
> 
> image
> 
> Hi do you think it would be better to update the existing extensions ? Would it then allow my feature branch to be accepted ?

pernielsentikaer:
I'm just curious, what are you missing from the built-in version?

Alexk2309:
> I'm just curious, what are you missing from the built-in version?

I think the main thing for the built in version is that you're not able call someone using a number, eg being only able to call someone who is in your contacts.

Furthermore with the FaceTime call there isn't an option to remove the confirmation before calling , calling straight.

pernielsentikaer:
It could maybe work in the FaceTime extension by @ron-myers - the context feels wrong if it's in the quick dial 🙂

PR #13880, opened 2024-08-05
Update 1password extension

bxb100:
## Description

image

## Screencast

origianl
image

## Checklist

### DIFF
--- a/extensions/1password/src/v8/utils.ts
+++ b/extensions/1password/src/v8/utils.ts
@@ -269,8 +269,9 @@ export function getFavicon(urls?: Url[]) {
     if (hostname.endsWith("apple.com")) return "https://www.apple.com/favicon.ico";
     if (hostname.includes("battle")) return "https://www.google.com/s2/favicons?domain=battle.net&sz=32";
 
-    // get hostname domain
-    const domain = hostname.split(".").slice(-2).join(".");
+    // get main domain
+    const datum = hostname.split(".");
+    const domain = datum.slice(Math.min(-datum.length + 1, -2)).join(".");
     // https://dev.to/derlin/get-favicons-from-any-website-using-a-hidden-google-api-3p1e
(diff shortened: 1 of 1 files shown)


extensions/1password/src/v8/utils.ts
```diff
@@ -21,8 +22,11 @@ export class ExtensionError extends Error {
 }
 
 export class NotFoundError extends ExtensionError {}
+
 export class CommandLineMissingError extends ExtensionError {}
+
 export class ZshMissingError extends ExtensionError {}
+
 export class ConnectionError extends ExtensionError {}
```
LitoMore, line 30: Can you revert these unrelated changes? Let's keep the original author's code style and only modify the code relevant to the pull request.

If every contributor formats the code once according to their preferences, then git blame will become very bad.

LitoMore (changes requested):

extensions/1password/src/v8/utils.ts
bxb100: 👌 Thanks review

LitoMore:
I don't encourage loading web resources in this security-sensitive extension, and I'm conservative about it to make sure it feels safe.

Of course, it would be a better choice if we could find a logic to let it only get icons from https://simpleicons.org.

bxb100:
@LitoMore agree with you, so I choose to use Google server with domain info `const domain = hostname.split(".").slice(-2).join(".");`

and the community had using `https://icons.bitwarden.net/${domain}/icon.png` 
https://github.com/raycast/extensions/blob/cbd0ba64a05aac10c4687e7ee8ea8c7b7135cc26/extensions/bitwarden/src/utils/search.ts#L42

bxb100:
@LitoMore find this https://support.1password.com/rich-icons-privacy/ , now I think it's not a good idea to introduce this feature, thanks review

PR #13905, opened 2024-08-07
Update mozilla-firefox extension

Laptop765:
## Checklist

### DIFF
--- a/extensions/mozilla-firefox/src/util/index.ts
+++ b/extensions/mozilla-firefox/src/util/index.ts
@@ -2,10 +2,6 @@ import fs from "fs";
 import path from "path";
 import { getPreferenceValues } from "@raycast/api";
 
-interface Preferences {
-  profileDirectorySuffix: string;
-}
-
 const userDataDirectoryPath = () => {
   if (!process.env.HOME) {
     throw new Error("$HOME environment variable is not set.");


extensions/mozilla-firefox/src/util/index.ts
```diff
@@ -1,5 +1,10 @@
 import fs from "fs";
 import path from "path";
+import { getPreferenceValues } from "@raycast/api";
+
+interface Preferences {
+  profileDirectorySuffix: string;
+}
```
pernielsentikaer, line 7: You don't need to manually set preference types as this is autogenerated in `raycast-env.d.ts`x

pernielsentikaer (changes requested):
This way could work @Laptop765, I'm not sure if there is a better way, but I guess it's hard to do anything better than this. @crisboarna, what do you think?

extensions/mozilla-firefox/src/util/index.ts
Laptop765: Ah thanks!

Laptop765:
> This way could work @Laptop765, I'm not sure if there is a better way, but I guess it's hard to do anything better than this. @crisboarna, what do you think?

We could parse `profiles.ini` as a separate fix but I'm not sure if that's better or not 😅

pernielsentikaer (approved):

PR #13917, opened 2024-08-07
Update messages extension

thomaspaulmann:
## Description

Just addeda bit of logic to close the Messages app after sending a message again if the app was closed before. This way the dock stays clean.

I noticed that the Adress field isn't restored. A bit of a bummer because I always send it to the same number. Is there something off with the `store` property on form items? I couldn't figure out what's off.

## Checklist


thomaslombart:
There's a bug indeed with the address field. I'll fix it in the upcoming extension's update.

PR #13921, opened 2024-08-08
[Whois] Added IP2Location.io as failover for WHOIS data

ip2location-com:
## Checklist

### DIFF
--- a/extensions/whois/CHANGELOG.md
+++ b/extensions/whois/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Whois Changelog
 
-## [Added IP2Location.io API] - 2024-08-08
+## [Added IP2Location.io API] - {PR_MERGE_DATE}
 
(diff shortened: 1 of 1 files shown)


j3lte:
This code is not going to work. `useFetch` is a hook, you can't return another hook conditionally. Furthermore, the `try... catch` won't be caught, because useFetch uses the `onError` method to catch for errors.

This would only work if you would place the `useFetch` after the first one and execute it when the first one errors which would set some outside state.

Is this code tested?

pernielsentikaer:
I agree with @j3lte. This is not the right approach to take as it doesn’t work

j3lte:
@pernielsentikaer I think we can just close this PR. I checked the activity of this Github account, it's just spamming its own service everywhere. 

Screenshot 2024-08-21 at 19 00 33

pernielsentikaer:
Let's do that @j3lte, that's a really good point

PR #13933, opened 2024-08-08
Add nordic-energy-prices extension

mSarheed:
## Description

Shows today's electricity prices for a selected region. Works for Nordic countries (Denmark, Sweden, Finland, Norway)

## Screencast

nordic-energy-prices-1

## Checklist


pernielsentikaer:
I really like that! A very long time ago I was doing something like that with some calculation of cheapest/most expensive hours - do you want to have the source code to implement some of my ideas?


https://github.com/user-attachments/assets/aff4eb4e-7b99-4011-acfb-b7ee8480471c

Mine was not done so it looks a bit weird but you'll get the sense of it 🙂

mSarheed:
@pernielsentikaer yes please. I think that would add value to the extension. Please share with me and I will see if I can incorporate some of your ideas

pernielsentikaer:
Here you go, it's a bit old and wacky made but there was some good ideas (Like showing the cheapest and most expensive hours)

You'll find it here: [elpriser.zip](https://github.com/user-attachments/files/16636246/elpriser.zip)

mSarheed:
Hi again @pernielsentikaer,
I have updated the extension with some of your ideas (thank you for that). I have also added you to the contributors list.

pernielsentikaer:
I'm so glad that my code can be used. There might have been some bugs since I couldn't find the latest version I was working on, but it sounds like you got the most out of it and some awesome ideas

Will check it again 🔥

pernielsentikaer:
One thing I'm wondering is if we forgot to add the VAT here? It feels like when I add 25% on top of the Danish prices, it gives me a more accurate answer 🙂

extensions/nordic-energy-prices/src/utils/averageCalculator.tsx
```diff
@@ -0,0 +1,9 @@
+import { PriceCurrency, PriceType } from "../types/energyData";
+
+export const averageCalculator = (data: PriceType[]) => {
+  let sum = 0;
+  for (let i = 0; i < data.length; i++) {
+    sum += data[i][PriceCurrency];
+  }
+  return sum / data.length;
+};
```
pernielsentikaer, line 9:
```suggestion
export const averageCalculator = (data: PriceType[]) => {
  if (data.length === 0) return 0;

  const sum = data.reduce((acc, item) => acc += item[PriceCurrency], 0);
  return sum / data.length;
};
```

mSarheed:
> One thing I'm wondering is if we forgot to add the VAT here? It feels like when I add 25% on top of the Danish prices, it gives me a more accurate result 🙂

@pernielsentikaer , you make a valid point. The listed prices do not include any additional "fees." However, I don't believe we should add the "fees"/VAT to the Danish prices alone. If we do that, we should also apply the same approach to the prices for other countries, such as SE, NO, and FI. However, I'm not familiar with how the "additional fees" work in those other countries. To maintain consistency in the way the extension displays prices and information, we should either add the fees to all countries or none of them.

pernielsentikaer:
It would be most accurate if it shows the real prices, I can help you with vat rates if you'd like 🙂

mSarheed:
@pernielsentikaer  I have added the option for the user to choose if the prices should be displayed including or excluding VAT. Please review again.

pernielsentikaer (changes requested):
There is some weird flickering in the latest version. Do you think it's possible to solve this?

https://github.com/user-attachments/assets/4fb83676-66c3-4213-b318-b36e830ee074

mSarheed:
> There is some weird flickering in the latest version. Do you think it's possible to solve this?
> 
>  Raycast05092024-S95UFiKq.png.mp4

That's weird. Do you have any suggestions where the issue might lie?

pernielsentikaer:
It looks like the first renders of an element doesn't have average price correct, see above debug data where I only looks at the first row

Raycast10092024-fUkJ3n32 png@2x

pernielsentikaer:
Do you still want to work on this? 🙂

mSarheed:
@pernielsentikaer The flickering issue should have been fixed now.

pernielsentikaer:
Much better! Will it ever show other dates? Otherwise, we might want to remove the subtitle

mSarheed:
@pernielsentikaer I think for a future update it could be nice to show historic electricity prices or the price for tomorrow. However, for now, I have just removed the subtitle that shows the date. We can always add it afterwards if needed.

pernielsentikaer:
I think we're ready to get the first release out 🔥

pernielsentikaer (approved):

PR #13934, opened 2024-08-08
[minecraft-color-codes] allow customising color code prefixes

lafkpages:
## Checklist

### DIFF
--- a/extensions/minecraft-color-codes/src/components/actionpanel.tsx
+++ b/extensions/minecraft-color-codes/src/components/actionpanel.tsx
@@ -17,6 +17,35 @@ export function generateActionPanel({
   return (
     <ActionPanel>
       <ActionPanel.Section>
+        <Action.CopyToClipboard
+          content={chatCode}
+          title="Copy Chat Code"
+          shortcut={{
+            modifiers: ["cmd"],
+            key: "c",
+          }}
+        />
+        <Action.CopyToClipboard
+          content={chatCodeEscaped}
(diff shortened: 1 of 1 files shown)


pernielsentikaer (changes requested):
I don't know much about this extension, but would you ever really need to switch between the two of them that often? It could be a dropdown in preferences where you set the default action instead. Does that make sense?

lafkpages:
> would you ever really need to switch between the two of them that often?

Well, I guess if you always play Minecraft on the same server you wouldn't really need to switch between prefixes that often. However, I think that being allowed to toggle them easily could come in handy for several users, so unless you strongly think it should be removed, I'd leave it.

Another question I did have was how I should store the currently used prefix. I thought of using `LocalStorage`, but it seemed quite excessive to have to create an encrypted storage area just to store a single character. I also thought of having a preference that I could modify every time the user toggled between primary and secondary prefix, however I don't think it's possible to modify preference values programmatically. So I ended up using a 1-byte cache, but that also seems excessive. So what would be the ideal way to store the toggled prefix?

pernielsentikaer:
You can maybe use `useCachedState` from `@raycast/utils`, I guess that would also work? In the end it's not really changing that much

pernielsentikaer (changes requested):
I would move copy actions to the top so it's the default actions for enter and cmd+enter 🙂

image

Can we do that?

lafkpages:
Done

pernielsentikaer (approved):

Issue #13900, opened 2024-08-06
[MultiForce] - Login URL Redirects user to the Lightning Pages setup page vs setup or home


onabavian:
### Extension

https://www.raycast.com/brian_j_decoder22/multi-force

### Raycast Version

1.80

### macOS Version

14.5

### Description

 Login URL Redirects user to the Lightning Pages setup page vs setup or home

### Steps To Reproduce

1. Click a Login
2. Salesforce opens into the Lightning App Builder Tab

### Expected Behaviour

Opens Setup home or Home

Decoder22:
@raycastbot assign me

Decoder22:
@raycastbot good first issue

Decoder22:
@raycastbot keep this issue open

Decoder22:
@raycastbot close this issue Closes the issue.

A fix for this issue was accepted into the Raycast store.

Decoder22:
@raycastbot close this issue

PR #13950, opened 2024-08-09
Update multi-force extension

Decoder22:
## Description

This update resolves an [issue](https://github.com/raycast/extensions/issues/13900) opened regarding what page the extension opens the user to in Salesforce. As part of this change, we introduce the ability for users to select from a dropdown where they are redirected to when opening in Salesforce. Options include 
1. Lightning Home
2. Setup Home
3. Custom (With conditionally visible text field that allows entering custom URI)

Note: Because this feature is new, existing logins will default to the Lightning Home option.

Also, there are new actions for each org that users can now access:

| Shortcut    | Action |
| -------- | ------- |
| ⌘ H | Open the selected org in your browser and navigate to the Lightning Home.  |
| ⌘ S | Open the selected org in your browser and navigate to the Salesforce Setup Home.  |

If a user selects either Home or Setup, the Open (ENTER) action will double map onto the open Home or Setup options so that duplicate actions are not present.

### Setting the default path
multi-force 2024-08-09 at 10 02 53
When users create/update an Org, they are given the option to choose to open to Lightning home, Setup Home, or a Custom URL.

### Choosing Custom URL
multi-force 2024-08-09 at 10 03 27
If users choose a custom URL, they will see a text field where they can enter the starting path.

### Default Open Action
multi-force 2024-08-09 at 10 02 36
multi-force 2024-08-09 at 10 03 12
The default open action will be replaced by Open Setup or Open Home depending on their preference.

### Open Custom Path
multi-force 2024-08-09 at 10 03 35
If users set 'Custom' as the Open To Path and provide a path, they will see 'Open' as the main option with both Setup and Home as additional options. This is also how users will first see the actions when they open the update as they will not have set a starting path yet. In this case, the default will be the Lightning Home path.

## Checklist


pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

Issue #13830, opened 2024-08-01
[Notion] Pass Default Values to Quick Capture command via LaunchContext


zsidnam:
### Extension

https://www.raycast.com/notion/notion

### Description

I would like to pass default values for `captureAs` and `notionPage` to the "Quick Capture" command via LaunchContext when creating a QuickLink. Similar functionality has been added to the "Create Database Page" command, but it does not look like this is supported in "Quick Capture".

### Who will benefit from this feature?

Anyone who frequently uses Quick Capture to save content to the same Page/Database in Notion and likes to automate Raycast workflows with QuickLinks. Capturing is supposed to be as frictionless as possible, so this functionality would save users one or two steps each time that use the command! 🙏

zsidnam:
I'm happy to work on this if that would be helpful!

thomaslombart:
Thanks for the PR! Here's our guide about contributing to extensions: [Contribute to an extension](https://developers.raycast.com/basics/contribute-to-an-extension)

PR #13953, opened 2024-08-09
[Notion] Add LaunchContext support for Quick Capture

zsidnam:
## Description

Addresses Issue https://github.com/raycast/extensions/issues/13830

Adds `LaunchContext` support to the "Quick Capture" command. Similar to the "Create Database Page" command, users now have access to a "Create Quicklink as Configured" action that can be used to create a quicklink pre-loaded with the currently selected page and capture type.

## Screencast

Raycast 2024-08-09 at 11 15 23

## Checklist

### DIFF
--- a/extensions/notion/src/quick-capture.tsx
+++ b/extensions/notion/src/quick-capture.tsx
@@ -27,6 +27,7 @@ import {
   Page,
 } from "./utils/notion";
 import { notionService } from "./utils/notion/oauth";
+import { Quicklink } from "./utils/types";
 
 type QuickCaptureFormValues = {
   url: string;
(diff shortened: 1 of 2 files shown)


extensions/notion/src/quick-capture.tsx
```diff
@@ -135,11 +177,33 @@ ${result?.content}
     getText();
   }, []);
 
+  function getQuicklink(): Quicklink {
+    const url = "raycast://extensions/notion/notion/quick-capture";
```
zsidnam, line 181: question: Is there a better way to get this deep link value? Also, it looks like most of the Notion deeplinks use this value, but the deeplinks created from the "Create Database Page" command use the value "raycast://extensions/HenriChabrand/notion/create-database-page". Which value is correct?

erics118: I believe both work. Both the `author` (HenriChabrand) and the `owner` (notion) field can be used to create the quicklink. I think the author version came from before `owner` was a valid field, and should be switched to use the `owner` version of the quicklink

thomaslombart (approved):
🚀

PR #13963, opened 2024-08-10
[Hashnode] FIX extension not working (Migrate API) + Cache + Modernize

xmok:
## Description

The extension would fail loading due to the API having been deprecated. I updated the queries and endpoints to use the latest api. Though some fields have changed, the actual functionality of the extension is still the same.

QUESTION for reviewers: since this issue was not caught previously, does it make sense for us to also rename the commands in package.json in addition to the title? Users will lose out on ranking but it might be OK to rename as no one noticed.

## Screencast

Screencast showing both old extension and new one:

https://github.com/user-attachments/assets/0d8c224b-d9c5-48ab-8977-f026beedb277

## Checklist


pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

Issue #12646, opened 2024-05-29
[Jira] It does not allow searching for Epics


puneet-ekline:
### Extension

https://www.raycast.com/raycast/jira

### Raycast Version

1.75.1

### Description

When creating a jira issue using the extension, it does not allow searching for Epics. It always shows No epic

### Steps To Reproduce

1. Create a jira issue using the extension. 
2. Fill our the details
3. Search for a valid epic name

### Current Behaviour

It show no epic found

### Expected Behaviour

It should show the epic names

thomaslombart:
Could you share a recording? It'll make it easier to see what's happening.

puneet-ekline:
https://github.com/raycast/extensions/assets/135100789/a8fa42a2-fe53-4b48-af48-4fffa6727d22

Here is quick recording.

thomaslombart:
Does it happen for all of your Jira projects?

luarmr:
> Does it happen for all of your Jira projects?
^^^ This

It shows only the epics created in the current project where I am creating the issue, as the project ID is used to retrieve the list of epics. Even when possible, I used it for cross-team projects, which is not usual.

https://github.com/raycast/extensions/blob/63ed284e297dbbf9012e384d7829d96b47c9eb24/extensions/jira/src/components/FormParentDropdown.tsx#L20

dipakparmar:
same, not seeing any epics, previously I was able to.

ishassan:
The issue is happening consistently to me. I don't see Epics at all in search, whether within the correct intended project or not.

It used to work before, now not anymore.

ishassan:
Oh, I think I know what's going on. The Epic autocomplete uses the api `/rest/api/3/issue/picker`. The API specs, according to the [docs](https://developer.atlassian.com/cloud/jira/platform/rest/v3/api-group-issue-search/#api-group-issue-search), are defined as:
> Returns lists of issues matching a query string. Use this resource to provide auto-completion suggestions when the user is looking for an issue using a word or string.

> This operation returns two lists:

>    History Search which includes issues from the user's history of created, edited, or viewed issues that contain the string in the query parameter.
>    Current Search which includes issues that match the JQL expression in currentJQL and contain the string in the query parameter.

The extension seems to search for Epics within the "History Search" not the "Current Search", so it's only returning Epics that I interacted with. This matches with what I reported above that I'm only seeing Epics I created. It's actually Epics that I viewed or interacted with as well.

[...]

Issue #10127, opened 2024-01-12
[Todoist] Fails with JS heap out of memory error


gh0stdotexe:
### Extension

https://www.raycast.com/doist/todoist

#### Error:
```
Error: Worker terminated due to reaching memory limit: JS heap out of memory

NodeError:errors:406:5
[kOnExit]:worker:313:26
Worker.<computed>.onexit:worker:229:20
```

### Steps To Reproduce

trying to run the todoist menu bar extension in raycast, but it refuses to work. 

- activated menu bar in todoist settings (disabled desktop dock option)
- open raycast
- refresh / attempt to run todoist extension "menu bar activate"
- fails / even after multiple refreshes, cache clear, raycast app quit + restart, todoist app quit + restart, reinstalled todoist even.

lucymhdavies:
Looks like my issue is a duplicate: https://github.com/raycast/extensions/issues/10789

So copying over my latest comment from that issue:

The issue has been happening every few days since I initially reported it. Same error logs.

On days where the issue occurs (such as today), reloading the extension will work for a few seconds at best, and then crash again.

Given the error is related to an out-of-memory issue, my theory is that it occurs on days where there are many tasks in my todo list, and doesn't happen on days when there are fewer. But I've not yet been able to validate this theory.

Is there any way to get more detailed logs to see what's going on?

gh0stdotexe:
> Looks like my issue is a duplicate: #10789
> 
> So copying over my latest comment from that issue:
> 
> The issue has been happening every few days since I initially reported it. Same error logs.
> 
> On days where the issue occurs (such as today), reloading the extension will work for a few seconds at best, and then crash again.
> 
> Given the error is related to an out-of-memory issue, my theory is that it occurs on days where there are many tasks in my todo list, and doesn't happen on days when there are fewer. But I've not yet been able to validate this theory.
> 
> Is there any way to get more detailed logs to see what's going on?

I believe you might be onto something. I have experimented with the task text display limit, which currently is set to 35, but I have also tried settings ranging from 20 to over 40. Additionally, I've tested filtering techniques to display fewer tasks from my typically full to-do list.

Just like you, I haven't been able to fully validate this theory. However, I suspect that the solution to the issue lies in this direction.

lucymhdavies:
I've added a bunch of "test" tasks to my Today list, to see if I can force a crash.
(I was able to clear the crash in my case by reducing the number of tasks in my Today, and clearing the extension's cache)

[...]

PR #13965, opened 2024-08-10
[Jira] Fix epic autocomplete to include all epics of a selected project.

ishassan:
## Description

This PR addresses the issue #12646 

**Current behaviour:**
The Jira extension allows the user to create a Jira issue and link it to a parent epic under a specific Jira project. However, not all epics are available under the epic dropdown menu. A user might not be able to find the suitable epic they want to link the issue to.

**Root cause:**
The Epic autocomplete uses the api `/rest/api/3/issue/picker`. The API specs, according to the [docs](https://developer.atlassian.com/cloud/jira/platform/rest/v3/api-group-issue-search/#api-group-issue-search), are defined as:

    Returns lists of issues matching a query string. Use this resource to provide auto-completion suggestions when the user is looking for an issue using a word or string.

    This operation returns two lists:

    History Search which includes issues from the user's history of created, edited, or viewed issues that contain the string in the query parameter.
    Current Search which includes issues that match the JQL expression in currentJQL and contain the string in the query parameter.
    
The extension seems to search for Epics within the "History Search" not the "Current Search", so it's only returning Epics that the user interacted with. If a user wants to link an issue to a new epic that a colleague created or one that they haven't interacted with before, they might not find it.

**Expected behaviour after the fix:**
The fix prioritizes using Current Search over History Search, such that all the relevant epics are retrieved when the user searches for them.

## Screencast

**Current behaviour (before the fix)**

https://github.com/user-attachments/assets/a9e8795c-4d08-4e7a-8e71-2dff6d1c291c

New behaviour (after the fix)

https://github.com/user-attachments/assets/99310287-c17a-4fc1-994f-0a386a5ec880

You'll notice that a new epic (KAN-1) shows up that wasn't retrieved before

## Checklist

### DIFF
--- a/extensions/jira/CHANGELOG.md
+++ b/extensions/jira/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Jira Changelog
 
-## [Fix epic autocomplete to include all epics of a selected project.] - {PR_MERGE_DATE}
+## [Fix epic autocomplete to include all epics of a selected project.] - 2024-08-14
 
 - When a user searches for an epic, the extension prioritizes retrieving search results from Current Search over History Search
 


extensions/jira/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Jira Changelog
 
+## [Fix epic autocomplete to include all epics of a selected project.] - 2024-08-10
```
jfkisafk, line 3:
```suggestion
## [Fix epic autocomplete to include all epics of a selected project.] - {PR_MERGE_DATE}
```

extensions/jira/src/components/FormParentDropdown.tsx
```diff
@@ -23,7 +23,11 @@ const FormParentDropdown = forwardRef(
           query,
         });
 
-        if (result && result.sections.length > 0) {
+        if (result && result.sections.length > 1) {
+          // Retrieve from Current Search
+          return result.sections[1].issues;
```
jfkisafk, line 28: (non-blocking) I am wondering about the question here: https://github.com/raycast/extensions/issues/12646#issuecomment-2275694286 and why can't we include both current and historic searches in the typeahead dropdown? Jira's TTL for historic issues should be anyways out-of-scope for Raycast. If including historic might make the list too convoluted (or too many items) then feel free to disregard this suggestion.

```suggestion
          // Retrieve unique issues from Current and History Search
          return [...new Map([...result.sections[1].issues, ...result.sections[1].issues].map(issue => [issue.id, issue])).values()];
```

jfkisafk (approved):
LGTM with minor suggestions.

puneet-ekline (approved):
LGTM

extensions/jira/src/components/FormParentDropdown.tsx
ishassan: Thanks for the review!

History search is actually a subset of the Current Search

Current Search lists all the valid epics, while History Search lists only the valid epics that the user interacted with before.

Merging both Current and History and selecting the unique out of them should yeild exactly the same results as just the Current Search.

Let me know if that makes sense

jfkisafk: Ah ok, thanks for the clarification. Then it just makes sense to use Current. 

But are we assuming jira api won't always populate current issues? In that case it will aid future debugging to somehow differentiate issues (with a different icon color/section) when they are being populated from history. That way it would be easier to tell if this issue arises again rather than asking the user to run in development mode or trigger the api.

Or if jira api contractually should always have current items, then we can keep this dropdown contract simple by just populating current issues or an empty list otherwise (ignore history)

thomaslombart (approved):
LGTM. Just waiting for [this comment](https://github.com/raycast/extensions/pull/13965/files#r1713462042) to be resolved 🙂

extensions/jira/src/components/FormParentDropdown.tsx
puneet-ekline: My 2 cents: We could de-duplicate and combine both the lists. It allows people to see the epics they interact with at the top, and the remaining search below.

ishassan: > Or if jira api contractually should always have current items, then we can keep this dropdown contract simple by just populating current issues or an empty list otherwise (ignore history).

Yep this is the case, Jira APIs should always have current items (if the search query should indeed match some items). 

My initial implementation was to just keep current search only and ignore history altogether indeed, but I noticed one case where the extension makes use of the history results; this is at the very beginning when there's no search query to begin with.

This is a screenrecording that shows this behaviour better (also updated in the PR description)


https://github.com/user-attachments/assets/c6e0587b-b8f6-4b64-83cf-596427574b06

You'll notice that when we first open the dropdown, before we write or search for anything, the dropdown shows the list of history items (which is KAN-2 in this case). Once I start typing, the list is populated with current items (which is both KAN-1 and KAN-2).

If I remove history search altogether, I'll be changing the existing behaviour such that when the user opens the drop down, it'll show an empty list, and then populate the list with the user searches for something.

I'd rather keep my changes to the minimum, and only change the behaviour necessary to fix the bug. This is why I think it's better to:
- Use history search when the user initially opens the drop down (same as before)
- Use current search when the user searches for something (to fix the issue of missing epics).

ishassan: > My 2 cents: We could de-duplicate and combine both the lists. It allows people to see the epics they interact with at the top, and the remaining search below.

It's an interesting proposal, but it's leaning to be a new feature or an improvement. We might think of adding a UI change as well to separate the epics in history vs others.

I want though to keep this PR strictly to just fixing the bug of missing epics. Hence, if there are no major issues with this fix, I think I'll just ship it and followups can be addressed in future PRs.

jfkisafk: Ok so conventionally typeahead APIs should list first 'n' entries if no query is specified. And based on the screenshot it seems JIRA api made it complicated and listing the entries in history instead. I was under the impression it would list 'n' random entries in current. Thanks for the explanation and I agree we should not break existing behavior in that case and keep history intact. You can resolve this issue.

puneet-ekline: 👍  to shipping it as it is.

jfkisafk:
> [!IMPORTANT]
> [/rest/api/3/issue/picker API](https://developer.atlassian.com/cloud/jira/platform/rest/v3/api-group-issue-search/#api-rest-api-3-issue-picker-get) is not paginated like [/rest/api/3/search API](https://developer.atlassian.com/cloud/jira/platform/rest/v3/api-group-issue-search/#api-rest-api-3-search-get) - which might lead to [heap memory issues](https://github.com/raycast/extensions/issues/10127) crashes for large JIRA customers while creating new issues. We should try to move to a paginated typeahead API if we can in followup PRs. cc: @thomaslombart

ishassan:
Hello, sorry to reopen this

I have linked my Github account since then, but I don't seem to have received the mentioned credits. Where shall I raise this?

PR #13990, opened 2024-08-12
[Migadu] cache most results + minor api refactor

xmok:
## Description

Replaced some of the `useState` with `useCachedState` so the data is cached.

In terms of refactoring, I moved the type casting of api responses into the api function itself so there is fewer explicit casting.

## Screencast

No screencast needed as functionality has not changed besides data is now cached on future loads.

## Checklist

### DIFF
--- a/extensions/migadu/CHANGELOG.md
+++ b/extensions/migadu/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Migadu Changelog
 
-## [Cache Results + Change Delete Shortcut] - {PR_MERGE_DATE}
+## [Cache Results + Change Delete Shortcut] - 2024-08-21
 
 - `Delete` keyboard shortcut changed from `Cmd+d` to the common `Ctrl+x`
 - Cache Aliases, Mailboxes, Rewrites and Identities


extensions/migadu/package.json
```diff
@@ -59,17 +59,17 @@
...
   },
   "devDependencies": {
-    "@raycast/eslint-config": "1.0.5",
-    "@types/node": "18.8.3",
-    "@types/react": "18.0.9",
+    "@raycast/eslint-config": "^1.0.8",
+    "@types/node": "^20.8.10",
+    "@types/react": "^18.3.3",
     "eslint": "^7.32.0",
     "prettier": "^2.5.1",
```
jfkisafk, line 71: Btw there is a moderate severity due to these dependencies, can you please execute `npm i eslint@8 prettier@3 -D`?

extensions/migadu/src/aliases.tsx
```diff
@@ -24,7 +24,7 @@ export default function Aliases() {
 
 function AliasesIndex({ domain }: { domain: string }) {
   const { push } = useNavigation();
-  const [aliases, setAliases] = useState<Alias[]>([]);
```
jfkisafk, line 27: Suggestion (non-blocking): In an effort of pure refactoring, we can use `useCachedPromise` hook (instead of `useCachedState`) to replace the whole `getDomainAliasesFromApi()`, `isLoading`, `error` etc. and we won't need `useCachedState`. So something like:

```typescript
const {isLoading, error, mutate, data: aliases} = useCachedPromise(getDomainAliases, [domain], {failureToastOptions: {title: "Error while getting domain aliases"}});
```

This will require some refactoring in callApi: removing the failure toast there and bubbling up the error msg as part of `Error` instead of `ErrorResponse`.

This way we can also optimistically mutate based on deletion etc.

This refactoring can be applied to most commands where we are using `useCachedState` in this PR. If this is intrusive for this PR, it can be handled as part of a separate PR too. What do you think?

extensions/migadu/src/aliases.tsx
```diff
@@ -24,7 +24,7 @@ export default function Aliases() {
...
+  const [aliases, setAliases] = useCachedState<Alias[]>("aliases", []);
```
jfkisafk, line 27: Shouldn't we use `domain` as namespace to avoid cache collisions:

```suggestion
  const [aliases, setAliases] = useCachedState<Alias[]>("aliases", [], { cacheNamespace: domain });
```

extensions/migadu/src/components/identities.tsx
```diff
@@ -13,11 +13,11 @@ import {
   showToast,
   useNavigation,
 } from "@raycast/api";
-import { useForm } from "@raycast/utils";
+import { useCachedState, useForm } from "@raycast/utils";
 
 export function IdentitiesIndex({ mailbox }: { mailbox: Mailbox }) {
   const { push } = useNavigation();
-  const [identities, setIdentities] = useState<Identity[]>(mailbox.identities);
+  const [identities, setIdentities] = useCachedState<Identity[]>("identities", mailbox.identities);
```
jfkisafk, line 20:
```suggestion
  const [identities, setIdentities] = useCachedState<Identity[]>("identities", mailbox.identities, {
    cacheNamespace: `${mailbox.domain_name}-${mailbox.local_part}`,
  });
```

extensions/migadu/src/mailboxes-and-identities.tsx
```diff
@@ -29,7 +29,7 @@ type MailboxesIndexProps = {
 };
 function MailboxesIndex({ domain }: MailboxesIndexProps) {
   const { push } = useNavigation();
-  const [mailboxes, setMailboxes] = useState<Mailbox[]>([]);
+  const [mailboxes, setMailboxes] = useCachedState<Mailbox[]>("mailboxes", []);
```
jfkisafk, line 32:
```suggestion
  const [mailboxes, setMailboxes] = useCachedState<Mailbox[]>("mailboxes", [], { cacheNamespace: domain });
```

extensions/migadu/src/rewrites.tsx
```diff
@@ -24,7 +24,7 @@ export default function Rewrites() {
 
 function RewritesIndex({ domain }: { domain: string }) {
   const { push } = useNavigation();
-  const [rewrites, setRewrites] = useState<Rewrite[]>([]);
+  const [rewrites, setRewrites] = useCachedState<Rewrite[]>("rewrites", []);
```
jfkisafk, line 27:
```suggestion
  const [rewrites, setRewrites] = useCachedState<Rewrite[]>("rewrites", [], { cacheNamespace: domain });
```

jfkisafk (commented):
I really like this refactor @xmok. Thanks for raising the bar 🔥 

Another thing I noticed was the delete resource shortcut. Most Raycast commands/extensions will use `Ctrl + x` as a standard practice, for easy memorization. For example:

https://github.com/raycast/extensions/blob/03ffbfd70f212ad17236ca9adb7ed0541d0698a6/extensions/linear/src/components/docs/DocumentActions.tsx#L175

But `Cmd + d` also makes sense.

xmok:
Great suggestions. I really appreciate the review 👀.

- We could add the `cacheNamespace` to cache for sure.
- As for utilizing `useCachedPromise`, I have refactored a lot of extensions to use the native Raycast hooks like `useFetch` and `useCachedPromise` but refactoring for the sake of refactoring is something I try to avoid. A future PR could replace the current api with the hook but doing it right now seems redundant as the code is working for now. I'll submit a future PR for it and eventually a lot of my older extensions like [Purelymail](https://www.raycast.com/xmok/purelymail)
- The Keyboard shortcut is also a welcome suggestion.
- Also updated the deps you mentioned.

Thanks!

extensions/migadu/src/aliases.tsx
```diff
@@ -24,7 +24,9 @@ export default function Aliases() {
 
 function AliasesIndex({ domain }: { domain: string }) {
   const { push } = useNavigation();
-  const [aliases, setAliases] = useCachedState<Alias[]>("aliases", []);
+  const [aliases, setAliases] = useCachedState<Alias[]>("aliases", [], {
+    cacheNamespace: "domain",
```
jfkisafk, line 28:
```suggestion
    cacheNamespace: domain,
```

extensions/migadu/src/mailboxes-and-identities.tsx
```diff
@@ -29,7 +30,9 @@ type MailboxesIndexProps = {
 };
 function MailboxesIndex({ domain }: MailboxesIndexProps) {
   const { push } = useNavigation();
-  const [mailboxes, setMailboxes] = useCachedState<Mailbox[]>("mailboxes", []);
+  const [mailboxes, setMailboxes] = useCachedState<Mailbox[]>("mailboxes", [], {
+    cacheNamespace: "domain",
```
jfkisafk, line 34:
```suggestion
    cacheNamespace: domain,
```

extensions/migadu/src/rewrites.tsx
```diff
@@ -24,7 +25,9 @@ export default function Rewrites() {
 
 function RewritesIndex({ domain }: { domain: string }) {
   const { push } = useNavigation();
-  const [rewrites, setRewrites] = useCachedState<Rewrite[]>("rewrites", []);
+  const [rewrites, setRewrites] = useCachedState<Rewrite[]>("rewrites", [], {
+    cacheNamespace: "domain",
```
jfkisafk, line 29:
```suggestion
    cacheNamespace: domain,
```

jfkisafk (commented):
Can you also please indicate in the CHANGELOG that the delete shortcut has changed?

xmok:
@jfkisafk, thank you for catching the mistake 🤦‍♂️.
I updated the CHANGELOG.

I'll be sure to eventually migrate the extension to use the native hooks in the future which would supersede this PR 🔥.

jfkisafk (approved):

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #13993, opened 2024-08-13
add hide-application

that70schris:
## Description

New extension for a global `Hide Application` command

macOS uses a dynamic menu command title for Hide, ex: `Hide Slack`.  this makes it impossible to remap the assigned keyboard shortcut `command + h`.  This extension just provides a single `Hide Application` command which can be assigned any shortcut in Raycast.

I ultimately think this should live in the built-in "System" commands.  But creating this to start with so I can move on from Alfred ;)

## Checklist


that70schris:
just looking for an icon

koinzhang:
It is possible to hide the current application using the shortcut CMD+H, is it worth creating another extension to do this?

that70schris:
@koinzhang I really benefit from being able to remap this shortcut in my setup.  I used to do this with an Alfred workflow.  Creating an extension in Raycast is the only reliable and snappy way I've found to do this.  Ideally this would be a built-in System command rather than a separate extension

koinzhang:
Perhaps creating a Script Command would be a better option?
The following code creates a Script Command in Raycast to hide the frontmost application.

```
#!/usr/bin/osascript

# Required parameters:
# @raycast.schemaVersion 1
# @raycast.title Hide App
# @raycast.mode compact

# Optional parameters:
# @raycast.icon 🙈
# @raycast.packageName System

# Documentation:
# @raycast.author koinzhang
# @raycast.authorURL https://raycast.com/koinzhang

	tell (first process whose frontmost is true)
		set appName to displayed name
	end tell
end tell

	set visible of application process appName to false
end tell
```

that70schris:
ah nice! thanks @koinzhang I had tried that before, but couldn't get the code right, this works perfectly and I agree is a better place than a complete extension

that70schris:
thanks again @koinzhang, provided the script for the community here: https://github.com/raycast/script-commands/pull/956

PR #14001, opened 2024-08-13
Update 1password extension

### LINKED ISSUE #13513: [1Password] Item does not contain the field 'one-time password'.
### Extension

### Raycast Version

1.78.1

### macOS Version

_No response_

### Description

I use raycast to load the 1password plugin

After configuring 1password-CLI, it works well and can read the account and password I specified normally.

But there is a problem when reading the one-time password.

I clearly set a one-time password in 1password, but I can't access it through raycast, and it keeps prompting that the one-time password cannot be found

> Item does not contain the field 'one-time password'.

Of course, one-time passwords definitely exist in 1password

After further investigation, I found that my 1password is a Chinese version, so the key corresponding to the one-time password in the cli is [一次性密码], not [one-time password]

So of course it cannot be accessed.

So for Chinese users, can we optimize it and choose the language for extracting the one-time password according to the version of 1password instead of always using [one-time password]

### Steps To Reproduce

As above

### Current Behaviour

_No response_

### Expected Behaviour

_No response_

bxb100:
## Description

- close #13513
- add new view

## Screencast

image

## Checklist

### DIFF
--- a/extensions/1password/src/v8/components/RandomPassword.tsx
+++ b/extensions/1password/src/v8/components/RandomPassword.tsx
@@ -75,8 +75,8 @@ export function RandomPassword() {
         if (value && isNaN(parseInt(value))) {
           return "Must be a number";
         }
-        if (value && (parseInt(value) < 1 || parseInt(value) > 64)) {
-          return "Password length must be between 1 and 64 characters";
+        if (value && (parseInt(value) < 4 || parseInt(value) > 64)) {
+          return "Password length must be between 5 and 64 characters";
         } else if (!value) {
           return "The item is required";
         }


extensions/1password/src/generate-password.tsx
```diff
@@ -0,0 +1,11 @@
+import { getPreferenceValues, PreferenceValues } from "@raycast/api";
+import { AuthProvider } from "./v8/components/AuthContext";
+import { RandomPassword } from "./v8/components/RandomPassword";
+
+export default function Command() {
+  return getPreferenceValues<PreferenceValues>().version == "v8" ? (
```
LitoMore, line 6:
```suggestion
  return getPreferenceValues<PreferenceValues>().version === "v8" ? (
```

Let's use strict equal.

extensions/1password/src/v8/components/RandomPassword.tsx
```diff
@@ -0,0 +1,112 @@
...
+              concealed={true}
+              content={generatedPassword || ""}
+              icon={Icon.Clipboard}
+              shortcut={Shortcut.Common.Copy}
+            />
+          )}
+        </ActionPanel>
+      }
+    >
+      <Form.Description title={"🔑Password"} text={`${generatedPassword || "Generating..."}\n\n\n`} />
```
LitoMore, line 104:
```suggestion
      <Form.Description title={"🔑 Password"} text={`${generatedPassword || "Generating..."}\n\n\n`} />
```

Let's insert a space between them for a better visual.

LitoMore (changes requested):

bxb100:
@LitoMore Thanks review, all settle

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #14006, opened 2024-08-13
Add Luna Search extension

relevantsam:
## Description

Adds a new extension for searching [Amazon Luna](https://luna.amazon.com), quick copying of the URLs for games, and direct link to play. Supports exploring the trending games, seeing the various metadata about games.

## Screencast

Pretty straight forward so adding the metadata screenshots for reference from the PR:
luna-search-1
luna-search-2
luna-search-3
luna-search-4
luna-search-5

## Checklist


extensions/luna-search/src/components/GameActions.tsx
```diff
@@ -0,0 +1,91 @@
...
+            push(<GameDetail game={game} searchCallback={(input) => handleSearch(input, true)} />);
+          }}
+        />
+      ) : (
+        // Otherwise, defer into Chrome.
+        <Action
+          title={DISPLAY_VALUES.openTitle}
+          icon={Icon.Globe}
+          onAction={async () => {
+            await open(game.openUrl, CHROME_KEY);
```
pernielsentikaer, line 63:
```suggestion
            await open(game.openUrl);
```

How about just opening it in your default browser? Doesn't that make more sense?

extensions/luna-search/src/components/GameActions.tsx
```diff
@@ -0,0 +1,91 @@
...
+          }}
+        />
+      )}
+      {game.playUrl ? (
+        <Action
+          title={DISPLAY_VALUES.launchGame}
+          icon={Icon.GameController}
+          onAction={async () => {
+            await open(game.playUrl ?? "", CHROME_KEY);
```
pernielsentikaer, line 72:
```suggestion
            await open(game.playUrl ?? "");
```
Do we need to add `?? ""` since we already checking in line 67
How about just opening it in your default browser? Doesn't that make more sense?

pernielsentikaer (changes requested):
Hi 👋 

### Review

- It’s pretty hard to see the logo in light appereance as you can see in above image - you can provide a [dark alternative](https://developers.raycast.com/api-reference/user-interface/icons-and-images#image)
- You should be able to use [useFetch](https://developers.raycast.com/utilities/react-hooks/usefetch) which also caches the data, you can use `execute` to let it know it shouldn’t search with empty search string - but the current way also works 🙂
- Does the API support pagination? If yes, then it’s possible to [paginate results](https://developers.raycast.com/api-reference/user-interface/grid#pagination)
- Update metadata images after corrections 🙂

extensions/luna-search/src/components/GameActions.tsx
relevantsam: Amazon Luna supports gameplay in Chrome and Edge which is why I force Chrome here ☺️

relevantsam: I think what I’ll do is have a tiered list of applications to open with:

1. chrome
2. Edge
3. Undefined (default handler)


And use https://developers.raycast.com/api-reference/utilities to identify which one to use. I’ll remove the reference to Chrome in the action label. 

For me, I use Firefox for everything except gameplay so it’s useful to force a Luna compatible browser here where possible.

extensions/luna-search/src/components/GameActions.tsx
relevantsam: Overhauling organization of actions to prevent needing this

extensions/luna-search/src/components/GameActions.tsx
relevantsam: Implemented the above!

extensions/luna-search/src/components/GameActions.tsx
relevantsam: Overhauled, no longer present.

relevantsam:
Thanks for the review!

>     * It’s pretty hard to see the logo in light appereance as you can see in above image - you can provide a [dark alternative](https://developers.raycast.com/api-reference/user-interface/icons-and-images#image)

Tracked down a light-mode appropriate logo:
luna-search-1



>     * You should be able to use [useFetch](https://developers.raycast.com/utilities/react-hooks/usefetch) which also caches the data, you can use `execute` to let it know it shouldn’t search with empty search string - but the current way also works 🙂

I looked at the useFetch API. Pretty neat! I could imagine a way I could hook into my expiring cache to get the control over the freshness of the data and rate of requests, but I'm pretty happy with the service I have at the moment.

>     * Does the API support pagination? If yes, then it’s possible to [paginate results](https://developers.raycast.com/api-reference/user-interface/grid#pagination)

No, but to protect for memory usage based on image display etc, I went ahead and implemented a basic pagination approach on the grid.

>     * Update metadata images after corrections 🙂

Done! Will update the PR's images.

pernielsentikaer:
That looks so much better, let's get the first version live 🔥

PR #14051, opened 2024-08-16
Add split-video-scenes extension

remorses:


extensions/split-video-scenes/package.json
```diff
@@ -0,0 +1,40 @@
...
+  "icon": "command-icon.png",
+  "author": "xmorse",
+  "categories": [
+    "Productivity"
+  ],
+  "license": "MIT",
+  "commands": [
+    {
+      "name": "index",
+      "title": "Split Video Into Scenes",
```
pernielsentikaer, line 15:
```suggestion
      "title": "Split Video into Scenes",
```

pernielsentikaer (changes requested):
Hi 👋 

- Can we somehow improve this flow? I didn't have pip so I kinda stranded - is there a homepage we can land on instead for guiding the user?

I also tried with pip3 since it was installed but it didn't go

Raycast26082024-V4neIoKj png@2x

- I would also add a bit more to the `README` to explain what it does and which modes do what

remorses:
I updated the extension to install the command with pipx, so there is no longer the problem you faced. pipx creates a sandbox for each installed lib, so it does not create problem for the user python environment.

I also updated the code to detach the scenedetect process so an user can use other extensions while scenedetect is running (which can take minutes), if the user runs the extension while the process is in progress it will show the current logs. here is a demo

https://github.com/user-attachments/assets/60f3fe21-2634-4609-ad3d-a600456bd956

pernielsentikaer:
I think we might need to let the user handle this installation process. I ended up in a loop because something was wrong with my PATHs, and I was unable to see the message since the window closed. I had to record the screen in order to read it

Raycast27082024-xDpga2sN png@2x

remorses:
I added the .local/bin folder in path, this should fix it permanently because pipx always installs packages there

extensions/split-video-scenes/src/index.tsx
```diff
@@ -0,0 +1,289 @@
...
+          <Action.SubmitForm onSubmit={handleSubmit} title="Split Video into Scenes" />
+        </ActionPanel>
+      }
+    >
+      <Form.FilePicker
+        id="file"
+        canChooseDirectories={false}
+        canChooseFiles={true}
+        autoFocus
+        title="Select video files"
```
pernielsentikaer, line 255:
```suggestion
        title="Select Video Files"
```

extensions/split-video-scenes/src/index.tsx
```diff
@@ -0,0 +1,289 @@
...
+        shell: true,
+        stdio: ["ignore", outputLogFile, outputLogFile],
+        detached: true,
+      });
+      childProcess.unref();
+      attachToLogsFile();
+      exitCode = await new Promise((resolve) => childProcess!.on("close", (code) => resolve(code || 0)));
+    }
+    if (exitCode === 0) {
+      closeMainWindow();
```
pernielsentikaer, line 168: Should we popToRoot and show a toast instead(with actions to open folder) - it felt weird that the window just closed without anything

pernielsentikaer (changes requested):
That worked as expected 🔥

I gave it another spin and found a few things 🔥

extensions/split-video-scenes/src/index.tsx
remorses: Now it shows an HUD and the folder with the scenes in finder

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #14186, opened 2024-08-26
Update color-picker extension

xilopaint:
## Description

- Improved the `Organize Colors` command to dynamically fetch the frontmost application and display its icon in the paste action.

## Screenshot

color-picker-1

## Checklist


extensions/color-picker/swift/color-picker/.gitignore
LitoMore, line 1: This is unnecessary. Let's remove this file.

extensions/color-picker/src/utils.ts
```diff
@@ -1,6 +1,6 @@
 import { getPreferenceValues, Icon, Image, Keyboard, List } from "@raycast/api";
-import { HistoryColor, HistoryItem } from "./types";
 import ColorJS from "colorjs.io";
+import { HistoryColor, HistoryItem } from "./types";
```
LitoMore, line 3: I saw so many unrelated changes on the PR. You should modify relevant code only.

Notes unrelated changes are not good for `git blame`.

Please revert all these unrelated changes. Thanks.

LitoMore (changes requested):

extensions/color-picker/swift/color-picker/.gitignore
xilopaint: I need some `.gitignore` rule. Without this file I see hundreds of files added in the source control. Also, the Raycast VS Code extension creates a new `.gitignore` file for the Swift directory after running the `Add Swift Support` command. So, it's not a strange approach for extensions with Swift support.

extensions/color-picker/src/utils.ts
xilopaint: This is result of VS Code `Organize Imports` command. It sorts imports and removes unused ones. I have seen a lot of extensions with unused imports which are basically dead code. I even think that this should be a formatting routine included in the CI.

pernielsentikaer:
@samuelkraft, do you have an idea why it suddenly fails? It looks like you added some of the code that's now producing errors

extensions/color-picker/src/utils.ts
LitoMore: > By the way, you made the same change in your PR: [#13896 (files)](https://github.com/raycast/extensions/pull/13896/files)

I only changed those files in which I worked.

xilopaint:
> @pernielsentikaer It's related to the `@raycast/eslint-config` dependency.

I reverted back to 1.0.8 in the last commit and the check still doesn't pass.

extensions/color-picker/src/utils.ts
xilopaint: Yes, but you didn't need to resort the imports to implement the feature your PR was intended for. It's the same here.

xilopaint:
Thank you, @samuelkraft. It worked!

@pernielsentikaer the PR is ready.

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #14259, opened 2024-08-30
NextDNS extension

pernielsentikaer:
## Description

New updated extension for maintaining NextDNS

## Checklist


pernielsentikaer:
@xmok I opened this PR, you can continue to work in my fork in the branch `nextdns`

pernielsentikaer:
I also started on a new `Logs` command, very simple for now. I'm not sure if I broke some of your work in my latest updates - could you give it a new spin @xmok :)

xmok:
@pernielsentikaer it looks good 💪.
Toggling, adding and removing is working OK. I cleared my Logs recently so I'm unable to test those but the API doc makes it seem fine. All good from my side.

pernielsentikaer:
That's awesome. Thanks for giving it another go and for all your contributions. Next would be to add enable/disable for the logs directly and a lot of other features 🙂

pernielsentikaer:
@jfkisafk do you have a few spare minutes to check this? 🙂

extensions/nextdns/src/components/add-domain.tsx
```diff
@@ -0,0 +1,63 @@
...
+                result.unshift({ id: values.domain, type, active: true });
+              }
+
+              return { result, profileName };
+            },
+          });
+          toast.style = Toast.Style.Success;
+          toast.title = "Added Domain";
+          toast.message = values.domain;
+          pop();
```
jfkisafk, line 36: (non-blocking) Not sure if we should use success toast in `.then()` clause of mutate Promise and pop in `.finally()` clause. This also works but I am thinking about avoiding any race-conditions. Same goes for other mutates. Something like this: 

https://github.com/raycast/extensions/blob/b8778e0dd272b7dba9767de6d6f0037e3577a36e/extensions/linear/src/components/docs/DocumentActions.tsx#L151-L156

extensions/nextdns/src/components/add-domain.tsx
```diff
@@ -0,0 +1,63 @@
...
+export default function AddDomain(props: { type: string; mutate?: Mutate | undefined }) {
+  const { pop } = useNavigation();
+  const { type, mutate } = props;
+  const description =
+    type === "allow"
+      ? "Allowing a domain will automatically allow all its subdomains. Allowing takes precedence over everything else, including security features."
+      : "Denying a domain will automatically deny all its subdomains.";
+  const { itemProps, handleSubmit } = useForm<DomainSubmitValues>({
+    async onSubmit(values) {
+      if (mutate) {
```
jfkisafk, line 16: Is there a use-case where `undefined` is being sent for `mutate`? In this diff, at least mutate is populated always. I think we should simplify the logic here for easier maintenance down the line.

extensions/nextdns/src/components/error-view.tsx
```diff
@@ -0,0 +1,10 @@
+import { Icon, List } from "@raycast/api";
+import { EmptyView } from "./empty-view";
+
+export function ErrorView() {
+  return (
+    <List>
+      <EmptyView title="Failed to load domains" icon={Icon.Warning}></EmptyView>
```
jfkisafk, line 7: non-blocking: can use the color red with the icon for failure.

extensions/nextdns/src/components/list-item.tsx
```diff
@@ -0,0 +1,106 @@
...
+      <ActionPanel.Section title={`Manage ${item.id}`}>
+        <Action
+          title={`${item.active ? "Deactivate" : "Activate"} Domain`}
+          icon={item.active ? Icon.XMarkCircle : Icon.CheckCircle}
+          onAction={() => toggle(item)}
+        />
+        <Action
+          title={`Delete Domain`}
+          icon={Icon.Trash}
+          style={Action.Style.Destructive}
```
jfkisafk, line 93: Should we not prompt for confirmation here?

extensions/nextdns/src/components/list-item.tsx
```diff
@@ -0,0 +1,106 @@
...
+    }
+  }
+
+  return (
+    <ActionPanel>
+      <ActionPanel.Section title={`Manage ${item.id}`}>
+        <Action
+          title={`${item.active ? "Deactivate" : "Activate"} Domain`}
+          icon={item.active ? Icon.XMarkCircle : Icon.CheckCircle}
+          onAction={() => toggle(item)}
```
jfkisafk, line 88: Should we not prompt for confirmation in case someone is deactivating a domain? Seems like a destructive action (we can use destructive style too in that case).

extensions/nextdns/src/components/error-view.tsx
pernielsentikaer: Updated with `tintColor` and added Action for opening preferences

extensions/nextdns/src/components/domain-list.tsx
```diff
@@ -0,0 +1,30 @@
...
+      isLoading={isLoading}
+      searchBarPlaceholder={`Search ${type}list of ${data.profileName} (${PREFERENCES.nextdns_profile_id}`}
+    >
+      {data.result?.map((element: DomainListItem) => (
+        <ListItem key={element.id} domainItem={element} mutate={mutate} />
+      ))}
+
+      {Object.keys(data).length === 0 && (
+        <EmptyView title={`No domains in ${type}list`} icon={{ source: "no_view.png" }} mutate={mutate} type={type} />
+      )}
```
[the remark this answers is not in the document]
pernielsentikaer: Should be better now, hopefully 🙂

extensions/nextdns/src/components/add-domain.tsx
```diff
@@ -0,0 +1,63 @@
...
+        const toast = await showToast({ style: Toast.Style.Animated, title: "Adding Domain", message: values.domain });
+        try {
+          await mutate(addDomain({ domain: values.domain, type }), {
+            optimisticUpdate(data) {
+              const { result, profileName } = data;
+
+              if (result.some((item) => item.id === values.domain)) {
+                toast.style = Toast.Style.Failure;
+                toast.title = "Domain already added";
+              } else {
```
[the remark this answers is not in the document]
pernielsentikaer: This should now be implemented 🙂

pernielsentikaer:
Seems like I broke the remove optimistic update. Can you check if that happens for you too, @xmok 😊

extensions/nextdns/src/components/list-item.tsx
pernielsentikaer: Should be there now 🙂

extensions/nextdns/src/components/add-domain.tsx
pernielsentikaer: Yes, from `empty-view.tsx`

```ts
{mutate && (
   <Action.Push
     icon={Icon.Plus}
     title="Add New Domain"
     shortcut={Keyboard.Shortcut.Common.New}
     target={<AddDomain type={type ?? ""} mutate={mutate ?? undefined} />}
   />
)}
```

xmok:
I'm seeing it in EmptyView as in below video. I haven't messed with any default Raycast shortcuts so it might have something do with defaults? 

`Is ⌘⇧+; better?` 
Yes it's fine.

P.S. for me `⌘,` opens Raycast Settings which is the conflict.

https://github.com/user-attachments/assets/4afbd31d-08e9-42cc-80ad-eee51d558584

extensions/nextdns/src/components/add-domain.tsx
pernielsentikaer: This should be fine, leaving for now 🙂

extensions/nextdns/src/components/list-item.tsx
```diff
@@ -0,0 +1,121 @@
...
+      toast.style = Toast.Style.Failure;
+      toast.title = `Could not ${newStatus ? "activate" : "deactivate"} domain`;
+    }
+  }
+
+  async function removeItem(element: DomainListItem) {
+    await confirmAlert({
+      title: "Are you sure?",
+      icon: Icon.Trash,
+      message: `This will remove it from the ${element.type}list`,
```
jfkisafk, line 57:
```suggestion
      message: `This will remove it from the ${element.type} list`,
```

extensions/nextdns/src/components/add-domain.tsx
jfkisafk: Ah ok, but anyways what I was pointing to was maybe we don't render an `Action.Push` in cases where `mutate` is `undefined`, rather than rendering a form and having an edge-case where submitting can be a no-op as the current implementation suggests. That way we can always expect a `mutate` value for Form.

extensions/nextdns/src/components/add-domain.tsx
jfkisafk: Nvm this what you have right now works fine. You can resolve this.

extensions/nextdns/src/components/empty-view.tsx
```diff
@@ -0,0 +1,41 @@
...
+      description={description}
+      icon={icon}
+      actions={
+        <ActionPanel>
+          {mutate && (
+            <Action.Push
+              icon={Icon.Plus}
+              title="Add New Domain"
+              shortcut={Keyboard.Shortcut.Common.New}
+              target={<AddDomain type={type ?? ""} mutate={mutate ?? undefined} />}
```
jfkisafk, line 26:
```suggestion
              target={<AddDomain type={type ?? ""} mutate={mutate} />}
```

extensions/nextdns/src/libs/api.ts
```diff
@@ -0,0 +1,103 @@
...
+  return useFetch(`${NEXTDNS_API_BASE_URL}${endpoint}`, {
+    headers: headers,
+    async parseResponse(response) {
+      const json = (await response.json()) as NextDNSErrorResult | NextDNSSuccessResult<DomainListItem[]>;
+      if ("errors" in json) throw new Error(json.errors[0].code);
+      const results = json.data.map((item: DomainListItem) => ({ ...item, type: props.type }));
+
+      return { result: results, profileName: await getProfileName() };
+    },
+    initialData: { result: [], profileName: "" },
```
jfkisafk, line 67:
```suggestion
    failureToastOptions: { title: "Failed to get domains." },
    initialData: { result: [], profileName: "" },
```

jfkisafk (approved):
LGTM now, just some minor comments, but approving.

extensions/nextdns/src/components/list-item.tsx
pernielsentikaer: It's actually the name they're using 🙂

image

extensions/nextdns/src/components/add-domain.tsx
pernielsentikaer: Makes sense 🙂

pernielsentikaer:
Do you want to be added as contributor @jfkisafk

jfkisafk:
> Do you want to be added as contributor @jfkisafk 

For sure @pernielsentikaer , that'd be awesome

PR #14300, opened 2024-09-02
fix: handle preserve casing logic and more

### LINKED ISSUE #13954: [Change Case] ...
### Extension

https://www.raycast.com/erics118/change-case

### Raycast Version

1.80.0

### macOS Version

13.3.1 (a) (22E772610a)

### Description

When I add text with special characters, like "-" "+" etc, Change Case is removing those characters in the final string

This means I constantly have to add the special characters back...which basically negates the timesave by using Change Case in the first place 😅

Would love to see an update of this!

### Steps To Reproduce

Copy text with special characters
Select "Capital Case" within Change Case
Paste the string back (it won't have the special characters)

### Current Behaviour

_No response_

### Expected Behaviour

_No response_

erics118:
## Description

Properly handle preserve casing. Before, the preference wasn't being used in the code in all places.

Fix various issues with cases.

Fix #13954

## Screencast

See https://docs.google.com/spreadsheets/d/1rns_3Um_mLyOX3GxAg4MgKVrThspFdlGPoFRSx5TfDs/edit?gid=0#gid=0 for the changes.

## Checklist

### DIFF
--- a/extensions/change-case/CHANGELOG.md
+++ b/extensions/change-case/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Change Case Changelog
 
-## [Fix logic regarding preserve casing] - {PR_MERGE_DATE}
+## [Fix logic regarding preserve casing] - 2024-09-02
 - Properly handle the logic regarding the `Preserve Casing` preference
 - Rewrote some of the documentation to make it clearer
 - Handle `KEBAB-UPPER-CASE` transform correctly


erics118:
There is a possibility of making the prefixCharacters and suffixCharacters, as well as the [other options](https://github.com/blakeembrey/change-case/tree/main/packages/change-case) configurable per-case rather than as a whole, but that would be pretty messy.

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #14326, opened 2024-09-04
fix(mute-microphone): temporary fix to icon duplication

iamyeizi:
## Description

First,
`await launchCommand({ name: "mute-menu-bar", type: LaunchType.Background });` is causing the icon duplication to occur, due to the launch type that raycast is not handling well enough, temporarily with `UserInitiated`, it works as expected showing only for a few milliseconds the raycast bar.

Second,
the menu item toggle does the same duplication, it is all related to `return null` to not show the icon when it is disabled.
I think it should be fine to temporarily remove the toggle button from the menu items to preserve the 'hide when unmuted' flag. Like this:

CleanShot 2024-09-03 at 11  31 13@2x

## Screencast

CleanShot 2024-09-03 at 10  02 21

## Checklist

### DIFF
--- a/extensions/mute-microphone/CHANGELOG.md
+++ b/extensions/mute-microphone/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Mute microphone Changelog
 
+## [Bugfix] - 2024-09-03
+
+- Temporary fix for menubar icon duplication
+
 ## [Improvements] - 2024-06-23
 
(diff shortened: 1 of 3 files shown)


pernielsentikaer:
There will be a release (probably today) that will fix the double menu icons, so this PR might not be that relevant 🙂

iamyeizi:
Great!

PR #14360, opened 2024-09-06
Add lunar-client-server-mappings extension

Tuchan:
## Description

This extension allow you to better visualise and search through [Lunar Client's Server Mappings](https://servermappings.lunarclientcdn.com/servers.json) directly in Raycast! It also allows you to join the server directly or copy the specific part of a json response, to only include the specified server data.

## Screencast
lunar-client-server-mappings-1
lunar-client-server-mappings-2
lunar-client-server-mappings-3

## Checklist


Tuchan:
considering soundboard extension already has a [.gitignore with .DS_STORE](https://github.com/raycast/extensions/blob/main/extensions/soundboard/.gitignore) inside of it, can the changelog check be ignored?

pernielsentikaer (changes requested):
Hi 👋

I don't know anything about Lunar, but could we handle if I don't have anything installed? (Like showing an List.EmptyView if it's not running?

Tuchan:
You don't need anything installed since you access the data from lunar client's cdn.
https://github.com/Tuchan/raycast-extensions/blob/65cb1be14191d3e60c6ed4e714af8d32c7572c5f/extensions/lunar-client-server-mappings/src/view.tsx#L80-L88
After which it is getting cached for a day. You can obviously force reload the cache if you need to.

If for some reason you can't access the link, you will get an error.
image

The only thing that accesses Lunar Client in some way, is this:
https://github.com/Tuchan/raycast-extensions/blob/65cb1be14191d3e60c6ed4e714af8d32c7572c5f/extensions/lunar-client-server-mappings/src/view.tsx#L202-L206

Lunar automatically handles server connections, so the only thing you need to do is specify a deeplink. 
If you don't have it installed, it will just not work. I will add an error message if you try to select the `Join the Server` option but you don't have Lunar Client installed.

pernielsentikaer:
Yesterday I just got some weird state where nothing was shown, today it works as expected 🙂

Can we move the preference about ToS to the README file instead

PR #14374, opened 2024-09-07
Update tidal-controller extension

0xDreamWeaver:
## Description

This change fixes the extension for non-English users of Tidal. Users can now select the correct language to use for the  extension. It also improves the README.md slightly, as it seems many users were unaware that they would need to enable Accessibility for Raycast for this extension to work (it uses AppleScript).

## Screencast

https://github.com/user-attachments/assets/0205785a-74e5-492d-94d8-7fac1c1d4432

## Checklist

### DIFF
--- a/extensions/tidal-controller/package.json
+++ b/extensions/tidal-controller/package.json
@@ -2,7 +2,7 @@
   "$schema": "https://www.raycast.com/schemas/extension.json",
   "name": "tidal-controller",
   "title": "Tidal Controller",
-  "description": "Control your local Tidal client with your keyboard.",
+  "description": "Control your local Tidal client with your keyboard.\nNote: You **must** enable Accessibility permissions for Raycast in order for this extension to work.",
   "icon": "extension-icon.png",
   "author": "broc-dev",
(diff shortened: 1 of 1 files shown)


extensions/tidal-controller/package.json
```diff
@@ -56,6 +56,80 @@
       "title": "Show Messages",
       "description": "Show pop up messages when an action is performed, such as toggling play/pause. Will also show a message if the action fails for some reason. No messages will be shown (even errors) if this is disabled.",
       "default": true
+    },
+    {
+      "name": "language",
+      "label": "Choose Language",
+      "type": "dropdown",
+      "required": true,
```
pernielsentikaer, line 64: Can we set this `false` and set English as default? This way the user will see the error message when trying to use the command and we won't change the flow for many users using it today

pernielsentikaer (changes requested):
Hi 👋

Thanks for the update 🔥 

It seems like you placed an image in the root of this extension (/tidal-controller/tidal-controller-1.png
), can you remove that?

PR #14391, opened 2024-09-09
[Converter] Fix #14095, better UX and big refactor

### LINKED ISSUE #14095: [Converter] Pasting 0 prefixed values
### Extension

https://www.raycast.com/koinzhang/converter

### Raycast Version

1.81.2

### macOS Version

14.4

### Description

In base converter when pasting 0-prefixed numbers to e.g. hex input field, it is treated as decimal number and gets immediately converted into corresponding hexadecimal number. This means that when trying to paste number `0100` into 16-base input field, it results in `64` (`0x64` is `100` in decimal) being inputed instead.
Also, when trying to paste 0-prefixed hex value that includes characters, it gets changed to `0`. I suspect that is because this value can't be interpreted as decimal (which it shouldn't be attempting to do at all) and falls back to returning 0.

### Steps To Reproduce

Case 1:
1. Copy `0100` to the clipboard
2. Launch "Base Converter"
3. Focus on 16-base input
4. Erase whole input
5. Paste the number from clipboard
6. Notice that `64` is inputed instead

Case 2:
The same as in case 1 but copy `0a00` to the clipboard instead and notice that it gets converted to `0`.

### Current Behaviour

0-prefixed numbers are interpreted as decimal regardless of where they are pasted.

### Expected Behaviour

Value after pasting should always match value in the clipboard.
Also for hexadecimal values `0x` prefixed values should be accepted.

fmkra:
## Description

I found many bugs and bad user experience cases, i.a.:

1. In base converter pasting value in input fields caused it to be always pasted in 10-base field instead of focused one (more details in #14095).
2. Input is forced to be validated in `onChange` which causes multiple issues and makes code conversion from any to native pretty much unusable. E.g. when trying to convert ascii `97` to native, user would try to input value `97` one character at the time, but after writing `9`, input gets erased since characters `<13` are not supported.
3. Ascii -> native converter assumes that input is single character without mentioning it while native -> ascii supports multiple characters.

Changes:

1. The biggest refactor is to remove overuse of `useEffect`s for handling input data (which is a bad practice and cause of some of bugs described above) in favor of custom `useXConverter` hooks. They provide clear and much more readable interface as well as allow to input arbitrary text without manipulating user's input. (After inputting wrong data other fields may become empty).
2. Add hex code conversion.
3. Replace `getInputItem` and `priorityDetection` with easier input handling and `autoPaste` setting.
4. Add automatic base detection when opening Base Converter command, e.g. opening Base Converter with `0x3af` in clipboard automatically pastes it in base 16 field and focuses on it.
5. Add support for multiple ascii characters either separated by comma when `asciiCommaSeparated` is enabled, or automatically detected e.g. 11297 -> 112,97.
6. Add automatic unit detection when opening Byte Converter.

### DIFF
--- a/extensions/converter/CHANGELOG.md
+++ b/extensions/converter/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Converter Changelog
 
+## [Refactor and Bug Fixes] - 2024-09-09
+
+- Fix bug with pasting values in base converter.
+- Refactor code by creating custom hooks for input handling.
+- Add hex code conversion.
+- Remove `priorityDetection` and `autoDetect` settings.
+- Add `autoPaste` and `asciiCommaSeparated` settings.
+
 ## [Refactor Code] - 2024-08-06
 
 - Refactor code to improve performance and readability.


koinzhang:
@fmkra Thanks for your contribution, LGTM.

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #14397, opened 2024-09-10
Add ai-code-namer extension

hunter-ji:
## Checklist


pernielsentikaer:
I just tested the extension and it somehow searches by itself if you just wait a few seconds (didn't click enter) - leaving it open just keeps refreshing and making some weird flickers - can youl ook in to that?

hunter-ji:
I have tested it several times, but I cannot trigger this problem locally. 
I can only cancel an automatic trigger condition, and it will not automatically trigger the generation when the drop-down box is selected. 
Could you please try again?

pernielsentikaer:
It doesn't do it automatically anymore, awesome - but if I hit enter and wait a few seconds, it clears the result. Can you look into that?

https://github.com/user-attachments/assets/ee8f200a-4a9b-49a3-b799-77f50b763729

hunter-ji:
Thank you for your feedback! I also tried this problem several times but failed to trigger it, which is very strange... 😣
So I still modified the automatic trigger event. Only after clearing the input can the result be automatically triggered. Otherwise, the enter key will only trigger the copy, which makes it more accurate.

hunter-ji:
Sorry for so many questions, but I'm curious, do you know why there are so many differences?
 I understand that this is all running in a Raycast environment, how can I avoid such problems when I develop new plugins in the future?

pernielsentikaer:
It seems like your logic still fails, is all of this not happening when you're testing it?

https://github.com/user-attachments/assets/208c1593-70f8-45f8-a0c2-8529dc061329

hunter-ji:
Yes, the two problems you told me were not triggered at all when I tested it. I tried each situation several times. I also felt very confused. 

https://github.com/user-attachments/assets/7728a724-118d-45b1-a09f-495566ff34d7

pernielsentikaer:
I'm not sure either, but outcommenting this line `searchBarAccessory={<StyleDropdown onStyleChange={onStyleChange} />}` and it stops going wrong, can you debug that?

hunter-ji:
I tried to solve it, and currently it only updates the value of the drop-down box automatically, or clears the result when the input box is cleared so that the user can generate it quickly.
I have tested it and there is no problem, I hope everything goes well...

https://github.com/user-attachments/assets/d4eabaa3-4575-47a4-805b-549ff383e6a2

hunter-ji:
Originally, in order to make it more convenient for users, the results were automatically cleared and regenerated when the drop-down box was updated. However, it is now found that the drop-down box will be triggered automatically unexpectedly, which is very strange, although I have not reproduced it.
So now the drop-down box only updates the value of style without doing any operation.

pernielsentikaer:
Could we make it preferences? I don't think many users would change this very often, do you?

hunter-ji:
Thanks for your suggestion, but I often have to change it myself.
I have tried to use preferences, but I found it not very good.

First of all, the file names, function names, and constants of the same language can be in different styles, especially the Go language I often use, so using the drop-down box, the drop-down box in different scenarios will remember the last user selection and keep it different, which I think is good.

Secondly, I often need to develop in multiple languages, so I often use function names, and I need to switch styles. For example, I often need to switch between Go and Python, and I have to switch styles constantly, so I originally thought that I could automatically regenerate after switching styles.

Therefore, I did not use the preferences in the end.

Maybe I don’t know enough about preferences, and there will be a better way to solve it. If it is convenient, I hope you can tell me, thank you.

hunter-ji:
Ok, I'll redo it as you wish...

pernielsentikaer:
I like your idea and it's fine as is, maybe we should get it to work with dropdown changes to redraw, can you get back to that stage and I can help you get it work correctly?

hunter-ji:
Thank you for your understanding. I submitted a new version. When the drop-down box is updated, new results will be generated. And I added some details to solve the previous problem. 
I did many tests locally and could not reproduce the previous problem, which made me very distressed. 😣

https://github.com/user-attachments/assets/139bfe32-e17d-4aac-8a9b-526529273689

hunter-ji:
According to your previous description, the change function of the drop-down box should be triggered. I added some case handling, hoping to solve the problem. But I don't know why it is triggered when you use it.

image

pernielsentikaer:
It works as expected now 🔥

The only things I have right now i that first time enter does the search, second time it copies to clipboard - maybe enter should always be search and then secondary could be copy/paste?

I also think that one image should be enough, maybe something like this, no need to show the root of Raycast

ai-code-namer-1

hunter-ji:
Really glad to hear that! 🥳

I'm not quite sure which feature you're referring to. Is it the function where input a description and generate a name, or the dropdown menu you opened?

If it's about inputting a description to generate a name, I understand what you mean is that using the Enter key to search each time, and then using another shortcut key to copy? I have tried to modify the Enter key control of this function, but when the Enter trigger is generated, the focus is transferred, and it seems that there is no way to solve this problem.

pernielsentikaer:
You're right, it doesn't make sense to add the generate as primary action when you searched, but adding it as a secondary item by adding `<Action title="Generate Names" onAction={generateNames} />` so List.Item would make sense

hunter-ji:
This is such a great idea! 👍

image

hunter-ji:
I think it's OK. 😊

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #14484, opened 2024-09-16
[Prettier] optional preference to use Tabs + remove Preferences type as it is auto-generated

xmok:
## Description

In some sites like certain forums, pasting using `Tabs` can have unintended consequences due to different rendering engines. This is why I added a checkbox in Preferences to allow us to choose `Tabs` or `Spaces`.

- removed the `Preferences` type as it is now auto-generated
- updated deps
- tweak the `switch` code

## Screencast

The sceencast is a bit iffy but I first paste using default setting i.e. `Tabs` then I toggle checkbox and use `Spaces`

https://github.com/user-attachments/assets/089e5fcc-25e3-4e5a-b0b6-b8d0efcc77f2

## Checklist

### DIFF
--- a/extensions/prettier/CHANGELOG.md
+++ b/extensions/prettier/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Prettier Changelog
 
-## [Maintenance Release] - {PR_MERGE_DATE}
+## [Maintenance Release] - 2024-10-02
 
 ## [New Icon] - 2023-04-25
 


pernielsentikaer:
Looks like a nice update @xmok 

@aeither can you check this?

jasonwaters:
The extension will load a prettier configuration if one exists in the user folder, for example in `~/.prettierrc`.  I recommend using that config file for prettier formatting specifics like this rather than add an option to the extension preferences.  That way users can customize any number of prettier options as they see fit.

see https://prettier.io/docs/en/configuration.html

xmok:
Yup I'm aware of that and it's easy enough to do that especially when running the extension locally but when installed from the Store, I'd prefer to not have to override `~/.prettierrc` or similar. However, maybe instead we could add a Preference allowing user to specify their prettier config of choice? Does that seem worthwhile?

pernielsentikaer:
@jasonwaters what do you think?

xmok:
It's not a deal breaker for me as I've  started using a local copy of the extension with my preferences pre-filled in. I'll convert this PR into a maintenance release and remove the changes I implemented so the deps are up-to-date otherwise we can re-visit in the future 😄.

pernielsentikaer:
Let's do that. It seems like the dependencies were pretty old. @jasonwaters, let me know your thoughts 😊

PR #14542, opened 2024-09-19
Add Noun Project extension

RobErskine:
## Description

Adds in a new extension for [the Noun Project](https://thenounproject.com). Allows users to search for Icons and then immediately copy them to their clipboard, as an inline image, svg string, or even a React component.

## Screencast

https://github.com/user-attachments/assets/21fb4fa0-d9cb-4b92-86f0-2c3c3a3844ec

noun-project-1

noun-project 2024-09-18 at 21 36 30

## Checklist

xtensionsnoun-project-2024-09-18-21 00 56@2x


pernielsentikaer:
Hi 👋

I think the error shown when you have incorrect credentials could be better. Could we show an error toast?

noun-project 2024-09-22 at 22 28 31

pernielsentikaer:
That's much better 👏

Do you think it's possible to have icons in white when using Raycast in dark mode?

noun-project 2024-09-25 at 08 31 46

RobErskine:
@pernielsentikaer I don't think so... at least within a reasonable amount of effort. When using the Noun Project API(https://api.thenounproject.com/documentation.html#get--v2-icon) to search for icons, the icon itself is returned as a PNG 
```

"thumbnail_url": "https://static.thenounproject.com/png/148533-84.png"
```
image

Noun Project also recommends using PNGs instead of SVGs:
> By default, PNG urls are included. SVG sizes vary from a few hundred bytes to several megabytes. We recommend using PNGs for displaying large numbers of icons at once, and retrieving the SVG for a specific icon when a vector format is needed.

To make this a little bit better on Dark Mode, I can think of the following
1. Use CSS Filters or anything to negate or invert the color of the icon (I don't know if this is possible?)
2. Add a background color to a `Grid.Item` (I also don't think this is possible)
3. Download all icons to the local fs and use something like [Node Sharp](https://sharp.pixelplumbing.com/) to negate them and then serve those locally? This doesn't feel great though, but maybe it is fine?
4. Attempt to return SVGs instead of PNGs, against Noun Projects recommendation, and see if that works or not. 

Or if you have any other ideas for the best way to make this work for Dark Mode, I'd appreciate your insights there.

Thanks - Rob

pernielsentikaer:
You should be able to do something like this `icon={{ source: "https://static.thenounproject.com/png/148533-84.png", tintColor: Color.PrimaryText }}`

RobErskine:
@pernielsentikaer Woah, didn't know i could do that 🤯. That worked like a charm. Let me know if you have any other feedback!

Dark mode:
index tsx — noun-project-2024-10-01-12 44 09@2x

Light mode:
Theme Studio-2024-10-01-12 48 28@2x

pernielsentikaer:
Much better 🔥

PR #14559, opened 2024-09-19
Add spiceblow-database extension

remorses:
## Description

A Raycast database extension to manage Postgres or Mysql databases inside Raycast, you can think of it like TablePlus but built into Raycast.

## About security

The extension can generate where clauses based on natural text, this feature depends on a backend service, this server does not have access to the database connection string, only the database schema is sent to the service to be able to generate the SQL query. 

All the data returned by the server cannot update or delete any data in the database, all the queries returned by the server are run in a read only session to prevent any destructive update to the database.

Any update, insertion and deletion to the database is presented to the user before running, showing the full SQL query, the user has to explicitly agree to run the query.

## About Google login

This extension uses a google token to identify the user, this token is used to connect a license to the user after the payment via Stripe via webhooks.

## Demo

Here is a quick demo of what it can do:

https://github.com/user-attachments/assets/02e7683a-5acb-4906-bd71-5199a01e126d


remorses:
Do you need any help with the review?

pernielsentikaer:
No, we just came back from a offsite last week, let me have a look now 🙂

pernielsentikaer (commented):
I'm a bit confused. What is https://spiceblow.com since it just redirects to Raycast, and why do we need to log in? I might be missing something very simple here 🙂

remorses:
It’s the API backend, It is used to generate the sql query and sql filter based on user natural text using an LLM, I did not create a landing page yet so it just redirects to Raycast now.

Login is required to later connect the user payment, it also prevents possible abuse in the backend, which uses OpenAI API to generate the sql. This is a paid extension, after a free requests limit the user is prompted to buy a one time purchase license which will be connected to the user Google email in the Spiceblow database.

extensions/spiceblow-database/CHANGELOG.md
```diff
@@ -0,0 +1,3 @@
+# Search Database Changelog
+
+## [Initial Version] - 2024-08-17
```
pernielsentikaer, line 3:
```suggestion
## [Initial Version] - {PR_MERGE_DATE}
```

extensions/spiceblow-database/README.md
```diff
@@ -0,0 +1,13 @@
+# Spiceblow
+
+Search and manage your SQL Database, currently supports PostgreSQL and MySQL
```
pernielsentikaer, line 3: I think we should tell a bit more about what it is to avoid confusion

pernielsentikaer (changes requested):
Hi 👋 

- Drafted values are not inserted when opening a draft

https://github.com/user-attachments/assets/2c1cf74d-837a-46e5-ae0b-10e9030aedab

- Maybe we should show a [toast](https://developers.raycast.com/api-reference/feedback/toast#showtoast) and [poptoRoot](https://developers.raycast.com/api-reference/window-and-search-bar)mysql://falkecenter_dk:CXZ1SA1oSOPSXNxvVbLkWgfxMc@mysql97.unoeuro.com/falkecenter_dk_db?statusColor=6D0000&env=production&name=falkecenter.dk&tLSMode=0&usePrivateKey=false&safeModeLevel=0&advancedSafeModeLevel=0&driverVersion=0&lazyload=true

- Use the term SQL instead of Sql so it's consistent

- Can a lot go wrong with this extension, do you think? It feels a bit insecure to just have it like that (since you have access without any passwords)

remorses:
Hi, 
> Drafted values are not inserted when opening a draft

I fixed the draft values, thank you

> Use the term SQL instead of Sql so it's consistent

Done

> Can a lot go wrong with this extension, do you think? It feels a bit insecure to just have it like that (since you have access without any passwords)

Some popular apps like TablePlus already exist where there is no password to access the app, Raycast already stores LocalStorage data in encrypted storage so the connection string is safe. 

> Maybe we should show a [toast](https://developers.raycast.com/api-reference/feedback/toast#showtoast) and [poptoRoot](https://developers.raycast.com/api-reference/window-and-search-bar)

What do you mean? In what part of the extension?

pernielsentikaer:
> What do you mean? In what part of the extension?

Sounds like some of my text was not included 🫠 When deleting a database, otherwise you might end up in a view where a database is deleted

- It's not possible to delete databases when you reaching the limit, maybe it's possible to delete from this view too?

spiceblow-database 2024-10-01 at 11 44 34

remorses:
> Sounds like some of my text was not included 🫠 When deleting a database, otherwise you might end up in a view where a database is deleted


Ok, I will add a pop after you delete a database, thank you for telling me

You can use code LC51UYGZ to buy the license for free and see everything works if you want to test database deletion again.

> It's not possible to delete databases when you reaching the limit, maybe it's possible to delete from this view too?

If the use case is to delete the data without buying a license you can use the "clear extension data" action in Raycast.

pernielsentikaer:
> OK, I will add a pop-up after you delete a database. Thank you for telling me

Awesome 🔥

> If the use case is to delete the data without buying a license, you can use the "clear extension data" action in Raycast.

Maybe we could add an action to push to the overview, where it's possible to delete, or is that too much? It should be intuitive

remorses:
> Maybe we could add an action to push to the overview, where it's possible to delete, or is that too much? It should be intuitive

I don't think it is useful, to use the extension you have to buy the license anyway, it would just add noise in my opinion.

remorses:
The first thing an user would do to delete the data is to uninstall the extension, which will clear the extension storage, right?

pernielsentikaer:
It would yes, let's stick with the current flow - I'll re-test it

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

remorses:
Thank you!

PR #14607, opened 2024-09-21
[Amazon AWS] feat: glue command

b0lle:
This reopens https://github.com/raycast/extensions/pull/13316

I implemented the changes required in the mentioned PR from @jfkisafk. This improves the performance of the feature very well!

### DIFF
--- a/extensions/amazon-aws/CHANGELOG.md
+++ b/extensions/amazon-aws/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Amazon AWS Changelog
 
-## [Glue command] - {PR_MERGE_DATE}
+## [Glue command] - 2024-10-21
 
 - Add Glue command to list Glue jobs, job runs, job definitions and to trigger a job run [#13316](https://github.com/raycast/extensions/pull/13316)
 ## [Pipelines, DDB, CFN commands] - 2024-07-29


b0lle:
I am facing issues when changing the account, sorry.

pernielsentikaer:
Do you need any help @b0lle

b0lle:
Hey @pernielsentikaer, when I run the extension with `npm run dev --verbose`, it runs as expected.
But when I build it, using `npm run build`, I am facing this screen
image
May you have a look?

pernielsentikaer:
I'm not sure how to test this since I don't use AWS - do you @jfkisafk?

b0lle:
May you have any idea @pernielsentikaer ? I think this is not an issue with the aws resources itself, but with the way of how I query those. Unfortunately I am a beginner in Typescript. For my understanding I tried to make sure that no process is loading and that the functions attribute is already loaded, but correct me, if I am wrong.

pernielsentikaer:
What happens if you debug it, adding console.log() right before calling `GlueRun` - at some point it must be null?

b0lle:
it logs me the correct object. Its not null - otherwise debugging this would be less complicated ;)

pernielsentikaer:
I'm not sure how I can test this, I'd like to give it a go but I don't have anything related to AWS. Could we add a simple check for it, just in case?

b0lle:
No with this issue I wouldnt give it a go ;)

b0lle:
@pernielsentikaer fixed it, ready for review

pernielsentikaer (approved):
Let's merge it and see how it goes 🙂

PR #14666, opened 2024-09-25
[Facetime]: Add dropdown for contacts and possibility for audio calls

arthur-fontaine:
## Checklist

### DIFF
--- a/extensions/facetime/CHANGELOG.md
+++ b/extensions/facetime/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Facetime
 
-## [Improve UX] — {PR_MERGE_DATE}
+## [Improve UX] — 2024-11-13
 
 - Add a dropdown to select a user's contact
 - Add a checkbox to use audio only


pernielsentikaer (changes requested):
It looks like a nice update, but you removed the possibility insert a number manually - could we still keep that?

arthur-fontaine:
> It looks like a nice update, but you removed the possibility insert a number manually - could we still keep that?

Thank you for your feedback! Sorry for the delay, I was very busy. I agree that it's important to keep the ability to manually input a number. What I wanted to do initially was to replace the `No results` message with a `Call [typed number]` suggestion if the entered number is not registered. However, I couldn't find a way to do this using the Raycast components. Do you have any suggestions?

Capture d’écran 2024-10-17 à 14 35 30

arthur-fontaine:
> It looks like a nice update, but you removed the possibility insert a number manually - could we still keep that?

Thanks for your feedback! Sorry for the delay, I've been very busy. This was originally what I wanted to do, but I couldn't find a way to do it the first time I worked on it. With the latest commit, if a number is not in the user's contacts, it will show `Call [what you typed]`.

Capture d’écran 2024-10-17 à 15 24 13
Capture d’écran 2024-10-17 à 15 24 22

pernielsentikaer:
Can we support phone numbers with just 8 digits too?

arthur-fontaine:
> Can we support phone numbers with just 8 digits too?

In fact it does. It's just for the icon. But can you give me an example? (in my country there is no 8 digit phone number, I don't see what it looks like) I will edit the regex. I took it on ihateregex.io, I thought it would be adapted for all the world.

pernielsentikaer:
A number here is just like this: 30123454

pernielsentikaer:
Are you still around @arthur-fontaine

arthur-fontaine:
oops forgot this one... just pushed a commit (it should be there in few minutes)

arthur-fontaine:
instead of checking if the input is a phone number, i just show the email icon if the input contains a character other than a number, a `+`, a `-` or a space

pernielsentikaer:
Looks good, let's do that

PR #14680, opened 2024-09-26
Add pilot-ynwr extension

romubuntuv2:
## Checklist

### DIFF
--- a/extensions/pilot-ynwr/package.json
+++ b/extensions/pilot-ynwr/package.json
@@ -5,6 +5,7 @@
   "description": "Pilot your Calendar, TaskManager and Journals quickly and effectively through a custom Notion template.",
   "icon": "logo_pilot.png",
   "author": "romubuntu",
+  "owner":"romubuncorp",
   "categories": [
     "Productivity",
     "Design Tools",


romubuntuv2:
hi ! thanks for your review ! :) 

i updated the extension to use a similar way to link databases from Notion as Hypersonic.
i understand that my extension have similarities with Hypersonic such as the taskmanager.
maybe my extension is less versatile but has more functionalities such as timers or keystones.

may you take a second look at it ? and thanks for your time :)

pernielsentikaer:
When I make the connection to Notion I end up in this view

pilot-ynwr 2024-10-14 at 08 37 48

If I go inside the command again I end with this failure toast

pilot-ynwr 2024-10-14 at 08 38 28

Could you look into this?

romubuntuv2:
hi ! thank you for diving in my extension :)
yes, I had this problem bc my oauth authorization was been triggered twice sometimes.. I made sure that it will happen once now. 
does it work correctly now ?

pernielsentikaer:
Works better now, some more feedback 🙂

- Instead of the custom loader (List.EmptyView and toast) it might be better just to stick with `isLoading={}` property on the `<List />`

- I would stay in `isShowingDetails` all the time since it flickers when data is being loaded

- Only use `style={Action.Style.Destructive}` for destructive actions

- Remember to add icons for all actions 🙂

- Is there a way to either automatically insert the right elements or `popToRoot()` after authentication?

pilot-ynwr 2024-10-15 at 10 30 23

romubuntuv2:
thanks again for this review ! :) 
i think i correct all your changes requested
i also go for the useCachePromise instead of my custom hook

pernielsentikaer:
Awesome, let me know when you're ready 🙂

romubuntuv2:
hi
i think it's all good and ready
if you have times to review my extension

pernielsentikaer:
The setup still feels veery clunky and it's very confusing, can the setup happening without setting a lot of fields up?

After setting it up I visited Task Manager but it feels like I'm stuck there?

romubuntuv2:
i can hide the form with all the dbs and fill it automatically so there is no setup
i think you should at first add a new project from the Home view. and then you will be able to add keystones or tasks in taskmanager.

pernielsentikaer:
There is something with the flow that doesn't work well, I'm still lost after going to the home command - how can I create a new project?

romubuntuv2:
hi ! 
i changed the configuration view with all the dbs to be 100% automatic, with no fill from the user. 
to add a project, there is a green button in Home to add a project.

pernielsentikaer:
I'm still concerned that it's too complex for users to figure out how to use it. Even with all your guidance, I'm still not entirely sure how to use it and where to start. It feels pretty niche for your needs and I'm not sure its suited for the store

romubuntuv2:
hi, thanks for all your reviews ! 
i understand that you have some guidelines and my extensions really suits my need haha ! 
thanks for your time :)

pernielsentikaer:
Will close for now 🙂

PR #14682, opened 2024-09-26
Update unicode-symbols extension

j3lte:
## Description

- Add full dataset with option in settings to enable/disable. Full Dataset contains +40.000 characters. This is only supported in Grid view.
  - Full dataset = 2.8Mb, Minimal dataset (default) = 404 kB
- Add option to copy the optional `HTML` entity of the selected character (e.g. `&amp;`)
- Added a browser view for the character in [compart](https://www.compart.com/en/unicode/)
- Make the JSON files smaller by renaming the fields. This does break the cache, so the cache key is renamed to `recently-used-v2`

## Checklist

### DIFF
--- a/extensions/unicode-symbols/CHANGELOG.md
+++ b/extensions/unicode-symbols/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Unicode Symbols
 
-## [More improvements] - {PR_MERGE_DATE}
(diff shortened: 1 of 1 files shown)


pernielsentikaer:
Hi 👋 

Looks good to me, approved 🔥

PR #14724, opened 2024-09-29
Add SnippetsLab Extension

renfeisong:
## Description

Hello! This is the official extension for [SnippetsLab](https://renfei.org/snippets-lab/). This extension allows users to search and view their library as well as perform common snippet actions. More details are in README.

## Screencast

https://github.com/user-attachments/assets/85117254-67e2-4a19-a826-bbf2226bbd4b

## Checklist


pernielsentikaer (changes requested):
Both on SetApp and App Store, I only see version 2.4.2. It seems like this extension relies on version 2.5?

renfeisong:
@pernielsentikaer Yes, 2.5 is pending release on both stores, and my plan is to release them at the same time as this extension, so users can check it out immediately following the "What's New" experience. I can add you to a Test Flight group so you can try it out, though, if that's helpful!

pernielsentikaer:
That would indeed be helpful for testing it. You can find my email in my GitHub bio 🙂

pernielsentikaer:
First thing we might look into is to both support acces from Applications folder and setApp, I do something similar in https://www.raycast.com/pernielsentikaer/tableplus

renfeisong:
From my testing, Setapp version is already supported out of box (it uses `getApplications` from `@raycast/api`, which was able to locate both apps).

pernielsentikaer:
I'm getting this error when I try to use the extension (which mentions setapp)

snippetslab 2024-10-09 at 08 38 38

renfeisong:
Ah, looks like you have the current Setapp version installed? Since the latest on Setapp is 2.4.2 and the `lab` binary is only bundled with SnippetsLab 2.5 and above, this error is expected (it is actually a good sign because the extension is able to locate the app bundle 🙂). Like I mentioned above, I will release the 2.5 update right after this extension is live. Users will see a "What's New" screen where the extension will be a highlighted item, and the app will also provide a link to Raycast store so the user can install it right away.

As for testing, if you have both the Test Flight and Setapp versions installed, it's possible that the extension happens to choose the Setapp version. You could try uninstall it temporarily so that you can test with the Test Flight build, which contains the required `lab` binary.

pernielsentikaer:
You're right! Makes sense 🙂

Should we consider moving accessories a bit around and use `tag` for tags like this?

Raycast09102024-uV0ee91C png@2x

renfeisong:
The new order definitely works better! I have adjusted it.

I am not sure about the `tag` look, however, as it's a bit too prominent to me and also takes up more space, which could push other elements including the title out:

Screenshot 2024-10-09 at 5 27 48 PM

The plain text style works better in this case:

Screenshot 2024-10-09 at 5 27 19 PM

pernielsentikaer:
You're right about the tags; let's stick with the current view 🙂

pernielsentikaer:
I'm not sure if it makes sense; in the tableplus extension I'm checking for [version](https://github.com/raycast/extensions/blob/9cf6a66f16f53e069a5032193b9b930bf52f1237/extensions/tableplus/src/utils.tsx#L22) to add some actions, maybe we could show something if they versions is outdated, would that be possible?

renfeisong:
Makes sense; done.

pernielsentikaer:
I think we're ready to launch first iteration when you're ready to release 2.5.0 🙂

renfeisong:
Awesome! Thanks for all your input 🙏

2.5 is ready, let's merge this PR when you can and I will push the button on my end.

pernielsentikaer:
I pushed the button 🙂

PR #14727, opened 2024-09-30
Update youtube extension

### LINKED ISSUE #14691: YouTube pinned channels unpin unexpectedly
### Extension

https://github.com/raycast/extensions/tree/f425d5cd51d8710c9d09c121624a284ba13a49a6/extensions/youtube

### Raycast Version

1.82.5

### macOS Version

14.6.1

### Description

while searching channels, they can be pinned.  proceeding to view the channel details and then channels recent videos, via the default action on each view (enter key), before escaping back through the views (escape key) to the channels list page, shows the channels as unpinned (but in recent).  They can be pinned again.

But they should not be unpinned at this stage.

image

image

image

### Steps To Reproduce

configure youtube extension with api key.

use search channnels feature to find some channels, pin them.

view channel details (enter).

go back to channels list (esc).

### Current Behaviour

Viewed channel becomes unpinned and is displayed in the recent channels list.

### Expected Behaviour

Viewed channel should remain pinned, and be displayed in both pinned channels and the recent channels list.  (or, if it must only appear in one location, appear in the pinned channels list.  Or at least appear there again whenever the channel no longer appears in the recent channels list).

ridemountainpig:
## Description

- Update the cache in the `search-videos` and `search-channels` commands to use local storage, ensuring that the data does not disappear after quitting the extension.
- Update the user’s pin details will not remove it from the recent list. Checking recent details will not remove it from the pin list.

closes #14691

## Screencast

https://github.com/user-attachments/assets/398de3b4-91a1-480c-beae-f8c2b2a04013

## Checklist

### DIFF
--- a/extensions/youtube/CHANGELOG.md
+++ b/extensions/youtube/CHANGELOG.md
@@ -1,6 +1,6 @@
 # YouTube Changelog
 
-## [Enhancement] - {PR_MERGE_DATE}
+## [Enhancement] - 2024-10-22
 
 - Update the cache in the `search-videos` and `search-channels` commands to use local storage
 


xilopaint:
> Update the cache in the search-videos and search-channels commands to use local storage, ensuring that the data does not disappear after quitting the extension.

Can you explain better what are you trying to fix here?

ridemountainpig:
> > Update the cache in the search-videos and search-channels commands to use local storage, ensuring that the data does not disappear after quitting the extension.
> 
> Can you explain better what are you trying to fix here?

Now, with the YouTube extension, users can pin channels and videos, as well as record their recently viewed channels and videos. However, the pinned and recent data is stored in cache, which means it disappears after the user quits the extension. I believe it would be better to use local storage to save this data, preventing it from disappearing when the extension is closed.

xilopaint:
> However, the pinned and recent data is stored in cache, which means it disappears after the user quits the extension. I believe it would be better to use local storage to save this data, preventing it from disappearing when the extension is closed.

We need to be cautious about this. In https://github.com/raycast/extensions/pull/5673, someone replaced the Storage API with Cache, and now you're changing it back. From what I can see, the pinned videos are lost only when we enter the `Show Details` action. This shouldn't happen. I use the Cache API to pin items in a bunch of extensions, and this has not been an issue.

xilopaint:
In my tests I can't reproduce the issue with the `Search Channels` command, but only with the `Search Videos` command.

pernielsentikaer:
I tried to test this, but it seems that my test key isn't functioning (yet) - hopefully it will work soon again

ridemountainpig:
@pernielsentikaer If your key still isn't functioning, I can send you my key to test. Thanks.

pernielsentikaer:
Let's merge it and see if it's fixed

PR #14754, opened 2024-10-02
Update json-format extension

m4l0n:
## Description

Add support for escaped JSON strings. The validation would not allow this previously, even though the `beautify()` method can format it.

## Screencast

https://github.com/user-attachments/assets/17809cd7-8131-4500-a8b2-39440a33339d

## Checklist

### DIFF
--- a/extensions/json-format/src/utils.ts
+++ b/extensions/json-format/src/utils.ts
@@ -34,7 +34,7 @@ export async function formatJS(text: string) {
 }
 
 function unescapeJsonString(str: string): string {
-  return str.replace(/\\(["\\/bfnrt]|u[0-9a-fA-F]{4})/g, function(match, p1) {
+  return str.replace(/\\(["\\/bfnrt]|u[0-9a-fA-F]{4})/g, function (match, p1) {
     const unescapeMap: { [key: string]: string } = {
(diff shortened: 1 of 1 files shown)


pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

m4l0n:
Does anyone mind explaining what are those credits for? I've just linked my Github and don't seem to find any page related to it. Thanks

andyfase:
@m4l0n @pernielsentikaer I believe this PR breaks the ability of this extension to decode JSON that has nested JSON within it (i.e. a escaped JSON object within a string value)

This valid JSON fails to decode:

```
{
   "foo": "{\"bar\": \"foo\"}"
}
```

Because the `unescapeJsonString` function strips out the needed escaping within the string value itself.

I believe the object of this change is manage a complete payload that is escaped, personally I think this should be possible if we add a condition to check if the first JSON key in the payload is escaped - i.e. look for `\"` at the start of the first key

Alternatively (and perhaps better) is to only perform a unescape if the call to JSON.parse originally fails, then unescape and try again.

andyfase:
PR raised https://github.com/raycast/extensions/pull/16251

PR #14791, opened 2024-10-04
Add veganify-application extension

philipbrembeck:
## Description

- Incorporated changes of https://github.com/raycast/extensions/pull/14513, requested by @pernielsentikaer

- This application checks a list of ingredients for their vegan status

## Screencast

https://github.com/user-attachments/assets/b0891c64-6f8b-4477-b921-d538a3ecc0ae

## Checklist


pernielsentikaer (changes requested):
Hi 👋

Thanks for the corrections 🔥

It looks like the API has an error 504 at the moment so I'm not able to test it 🫠

image

veganify-application 2024-10-07 at 06 56 57

When it's up again I can give it a spin, should we catch a http error and show a failuretoast

philipbrembeck:
@pernielsentikaer The API should be back up again! Sorry for that 😅

pernielsentikaer:
I'm so sorry, your custom debounce performed better here because it now queries your server too frequently. I think it's because it matches on anything, which confuses me. Just writing "x" shows a list item—should it?

philipbrembeck:
@pernielsentikaer 
Thank you for your feedback and for catching that issue. You're right, the current implementation was querying the server too frequently (this did not cause it to be down however today morning, the controller of the relatively small Kubernetes cluster just decided to update itself midday without having a second controller in place).

I've reintroduced the custom debounce function while keeping the `throttle` prop on the `<List />` component.
This combination should reduce the frequency of server queries, which will provide a better balance between responsiveness and server load.

Regarding "x" showing a list item: This is actually intended behavior, since the extension is designed to check even single ingredients, which is why it responds to any input.
This should allow users to quickly check individual ingredients without needing to enter a comma (for example when you're not sure about one specific ingredient and maybe have a shortcut for the extension at hand).
The debounce ensures we don't make API calls for every keystroke, while still allowing these single ingredient checks.

Please review the updated implementation and let me know if you have any further suggestions or if you think we should adjust this single-ingredient behavior. 😊

pernielsentikaer:
I tried to update the logic a bit to make it seem more like it's loading while the debounce is running, giving it a more natural feeling 🙂 I think we're good to go 🔥

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

philipbrembeck:
@pernielsentikaer Thank you very much for your help & contributions as well as your reviews 😊

pernielsentikaer:
Thank you for your cooperation. It was a pleasure working with you 🫶

Issue #14356, opened 2024-09-06
[ChatGPT] ... Lost all conversations and custom models, also new custom models cannot be created


lawnzapper:
### Extension

https://www.raycast.com/abielzulio/chatgpt

### Raycast Version

1.82.3

### macOS Version

14.5

### Description

Did nothing in between one use and another, but I suddenIy lost all my conversations and custom models, also new custom models cannot be created anymore, so I'm stuck with a default gpt-3.5-turbo model

### Steps To Reproduce

Just using raycast as usual, triggered by keyboard shortcut

lawnzapper:
probably related to #14125

Issue #14125, opened 2024-08-21
[ChatGPT] extension corrupted and all my custom models got deleted - help


turnipsforme:
### Extension

https://www.raycast.com/abielzulio/chatgpt

### Raycast Version

1.81.2

### Description

Is there a way to recover my custom models? I tried looking through the extension config using FindAnyFile and couldn't find where the models are stored. I had lots of models that I spent a while making, would be amazing to be able to find the text files for them!

### Steps To Reproduce

not sure what caused it to corrupt, it just stopped letting me interact with the extension (all options came up blank)

mormubis:
I have the same issue. I cannot create new models.

pernielsentikaer:
That does sound weird, and you're not getting any errors at all?

mormubis:
Nope, for me, suddenly, it was using GPT 3.5 when I had two models with 4o and two different prompts. I needed to reinstall the extension to make it work.

I saw no models in the models view, and I couldn't create any, even though it says Cmd + T to create a model.

jwhazy:
Finding this happens once I reach a certain amount of conversations. Has happened a couple times now.

Issue #8977, opened 2023-10-26
[ChatGPT] Can't create new model with ⌘ + T.


nandastone:
### Extension

https://www.raycast.com/abielzulio/chatgpt

### Description

Listing models shows "No custom models, create a new model with ⌘ + T shortcut."

The shortcut doesn't work.

### Steps To Reproduce

Open the ChatGTP Models command.
Press ⌘ + T.

### Current Behaviour

Pressing ⌘ + T plays the Mac invalid input sound, and nothing happens.

### Expected Behaviour

Pressing ⌘ + T allows adding a new model.

nandastone:
Suggestion - all the "Create Model" action to be triggered with regular Raycast actions like pressing enter.

jchuder:
@nandastone I also had the same issue, but I found that it was resolved after reinstalling the extension.

jblogs:
it appears again in version 1.73.2.
I do customized models than the default one, but the model page now shows no model
the cmd T keystroke does not work
i know if i reinstall the extension might resolve the problem, but it would lead to lose all my conversation data

PR #14804, opened 2024-10-05
Update chatgpt extension

SmileMachine:
## Description

- Feature: Change all occurence of `gpt-3.5` and the like to `gpt-4o-mini`, as the latter is cheaper and more efficient.
- Feature: Update `setOption` to include all model IDs, removing the filter for 'gpt' prefix. (To support `o1` models).
- Chore: Remove unnecessary union type literals in `src/type.ts:Model.option`, since 'option' was already effectively a 'string'.

## Checklist

### DIFF
--- a/extensions/chatgpt/src/utils/import-export.ts
+++ b/extensions/chatgpt/src/utils/import-export.ts
@@ -4,58 +4,59 @@ import moment from "moment";
 import { homedir } from "os";
 import { join } from "path";
 
-
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 export function ExportData(data: any, moduleName: string) {
-	const timestamp = moment().format("YYMMDD-HHmmss");
-	const filename = `Downloads/raycast-chatgpt-${moduleName}-${timestamp}.json`;
-	writeFile(join(homedir(), filename), JSON.stringify(data, null, 2))
-		.then(() => {
-			showToast({
-				title: `${moduleName} exported`,
-				message: `to ~/${filename}`,
-				style: Toast.Style.Success,
-			});
-		})
-		.catch((error) => {
-			showToast({
-				title: `Failed to export ${moduleName}`,
-				message: error.message,
-				style: Toast.Style.Failure,
-			});
-		});
-};
+  const timestamp = moment().format("YYMMDD-HHmmss");
+  const filename = `Downloads/raycast-chatgpt-${moduleName}-${timestamp}.json`;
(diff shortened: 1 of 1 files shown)


pernielsentikaer:
How should we test this most effectively? Will everything work as it did before? I assume the default model is only changed for new users?

SmileMachine:
> How should we test this most effectively? Will everything work as it did before? I assume the default model is only changed for new users?

Yes, you are correct. The default model is only changed for new users, as existing users may have already edited their models, so their data from `localStorage` will be used instead.

This is a temporary solution to address issues like #8977, #14356, and #14125. I have also encountered this issue recently, where all my model data and conversation data were lost, though the chat history remained intact, which is quite unusual.

Specifically, in the `Models` section, if no models exist, the action button becomes blank, preventing the creation of new models. In such cases, the hard-coded default model is used (see `src/hooks/useModel.tsx:83`).

However, I'm unable to consistently reproduce the bug or pinpoint the root cause. I'm actively investigating the issue. One possible workaround is to reinstall the extension, which suggests the issue might be caused by a runtime error that overwrote the `localStorage` data.

Perhaps implementing an option for users to import/export their data could serve as a more robust long-term solution to mitigate data loss.

Additionally, is there a way to migrate existing data into a development environment for debugging? Or is it possible to debug the existing extension directly?

pernielsentikaer:
No, that's not really possible 🫠 When the extension is built from the store, it uglifies everything and groups it into single files

SmileMachine:
I believe I’ve identified the root cause and have made some code changes that should resolve the issues. Unfortunately, it seems some data might have already been lost.

As a GitHub newcomer, this is my first PR and I'm still getting familiar with the workflow. Additionally, it looks like contribute for the Raycast extension differs from those in common repositories.

When submitting this PR, is it possible to open another PR separately, or should I include all changes within the same PR? Alternatively, is there a different approach you’d recommend?

Thank you for your guidance!

pernielsentikaer:
Could anyone of you help testing this?  @abielzulio @erodactyl @shichongrui @timolins @CruelMoney @k8scat @Nauxscript @aeither @leeonfield @husniadil @luckykong

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

SmileMachine:
> Hi 👋
> 
> Looks good to me, approved 🔥

Thanks！This is actually the first real PR I'm submitting outside of my personal practice projects, so I'm super excited about this! 😄 I'm looking forward to any feedback or suggestions you might have. Thanks a lot for the opportunity to contribute!

Issue #14773, opened 2024-10-03
[Airtable] find and update records


mspych:
### Extension

https://www.raycast.com/marks/raycast-airtable-extension

### Description

The current Raycast Airtable extension allows viewing and listing bases but lacks functionality for directly interacting with records. I propose adding features that enable users to search for specific records by name and update them directly within the extension. 
Not sure how this could work in details, but maybe it would be possible to configure some fields that are editable in the configuration.
Achieving functionality like [this one](https://youtu.be/m5MDv9qwhU8?t=281) would be beyond amazing. 
This enhancement will greatly improve efficiency by minimizing the need to switch to the base or interface.

marks:
Chiming real quick to say that I agree this would be a great feature set. I dont think I'll be able to take it on in the near future, so any other developer should definitely consider it!

xmok:
I took a very quick stab at this:

https://github.com/user-attachments/assets/86f19a0c-c4b8-47ca-89f8-2a5b26367e76

In the above screencast we are now able to View/List records in a base - there is no editing as of yet because will have to think of an efficient way to handle it. In the meantime, I can at least make a PR for the above screencast but we will need to modify the OAuth application in Airtable to ask for the additional scope of `data.records:read`. Presently I have created my own Dev OAuth Integration for the screencast.

mspych:
This looks great. Keeping my fingers crossed for the edit functionality!

marks:
Breadcrumb to comment I left on the PR about contacting Airtable support to add the additional scope needed for this record read/write functionality: https://github.com/raycast/extensions/pull/14864#issuecomment-2419875303

PR #14864, opened 2024-10-10
[Airtable] view records + edit singleLineText multilineText fields ONLY

xmok:
## Description

This PR is related to Issue #14773 - though it does not implement the functionality from the video shared in the Issue, this update does allow one to VIEW records of a base. EDIT is supported but only for singleLineText & multilineText fields at the moment.

⚠️ NOTE: For the update to work, the current OAuth Integration needs to be updated so it has the following scopes:

a. `schema.bases:read` - needed to view a Base
b. `data.records:read` - needed to view Records of a Base
c. `data.records:write` - needed to edit Records of a Base

W/o updated scopes, it is still possible to test - follow the steps in the README to create your own OAuth Integration with the required scopes!

## Screencast

https://github.com/user-attachments/assets/5c3c48d1-7d19-4ba7-8f07-5f22f8729d7d

## Checklist

### DIFF
--- a/extensions/airtable/CHANGELOG.md
+++ b/extensions/airtable/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Airtable Changelog
 
-## [Base Record Enhancements] - {PR_MERGE_DATE}
+## [Base Record Enhancements] - 2024-10-17
 
 - `View Records` of a Base
 - `Edit` "singleLineText" and "multilineText" fields of a record (more fields will follow in the future)


pernielsentikaer:
Thanks for your contribution here @xmok 

What happens if the user tries to use some of the new functionality without reauthing?

pernielsentikaer:
We did something similar for the GitHub extension. Here's [an example](https://github.com/raycast/extensions/blob/cd322444996ce98ff6d8bc028d283d65236373ce/extensions/github/src/components/ProjectActions.tsx#L37) with the GitHub extension, it's been added as part of [this PR](https://github.com/raycast/extensions/pull/11999).

xmok:
The GitHub example looks great - I can try to show a more verbose Failure Toast for users who are already logged in but for users who try to log in again or for the first time, the OAuth will fail and not allow them to log in - this is because as far as I can figure out the GitHub extension uses `OAuth flow` or `PAT` whereas Airtable is using `OAuth App`. In the latter the scopes are specified by the OAuth Integration so it's not possible to authorize due to the scope mismatch whereas the OAuth and PAT allow one to modify their scopes by re-authenticating.

To summarize:
can't show toast if user can't even authoize via OAuth in the first place 😅.

xmok:
Here's a screencast where I am logged in to the old scope and the Failure Toast returned by Airtable is already pretty verbose:

https://github.com/user-attachments/assets/c575f1cd-3778-4dba-9061-2f86baaec520

I'll update the toast with an action to open Preferences.

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

xmok:
> The GitHub example looks great - I can try to show a more verbose Failure Toast for users who are already logged in but for users who try to log in again or for the first time, the OAuth will fail and not allow them to log in - this is because as far as I can figure out the GitHub extension uses `OAuth flow` or `PAT` whereas Airtable is using `OAuth App`. In the latter the scopes are specified by the OAuth Integration so it's not possible to authorize due to the scope mismatch whereas the OAuth and PAT allow one to modify their scopes by re-authenticating.
> 
> To summarize: can't show toast if user can't even authoize via OAuth in the first place 😅.

@pernielsentikaer as I mentioned earlier, OAuth will fail since the integration has not been updated from settings. Since this PR was merged it will not work for those who try to login again due to scope mismatch; I've attached a video showing what I mean:

https://github.com/user-attachments/assets/df1e0185-c15a-45cf-b948-6f267c101b46

There are 2 quickfixes I can immediately make PR for to rectify but both have tradeoffs:
option 1: simply change LINE 6  of `oauth-client` from
`const scopes = ["schema.bases:read", "data.records:read", "data.records:write"];` to the initial
`const scopes = ["schema.bases:read"];`

CON: the new Record views will not work
PRO: ought not to break existing log ins

option 2: change the OAuth application itself to a new one I have created (see screenshots below)
CON: will probably? log out existing users
PRO: everything will work!

airtable-new-oauth-integration-1
airtable-new-oauth-integration-2
airtable-new-oauth-integration-3

Please let me know what seems like the best approach and I'l make a quick PR.

marks:
Hi @xmok - thanks for taking this on! I no longer work at Airtable but if you write into support@airtable.com and mention this, they may be willing to add the scope to the OAuth app whose client ID I included as the default for the extension.

PR #14888, opened 2024-10-12
Google Workspace: Add file path information

### LINKED ISSUE #14874: [Search Google Drive Files] When searching for folders show the folder path next to the folder result
### Extension

https://www.raycast.com/raycast/google-workspace

### Description

When the folder results come back, the result only displays the folder's name.
This confuses when folder with the same name come back in the results.

_No response_

### Anything else?

_No response_

ridemountainpig:
## Description
Adds the file path information to the list of accessories.
closes #14874

## Screencast
CleanShot 2024-10-11 at 15 40 02@2x

## Checklist

### DIFF
--- a/extensions/google-workspace/CHANGELOG.md
+++ b/extensions/google-workspace/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Google Workspace Changelog
 
-## [Add file path info to the file list] - {PR_MERGE_DATE}
+## [Add file path info to the file list] - 2024-10-21
 
 - Adds the file path information to the list of accessories.
 


thomaslombart (commented):
Thanks for the PR! I have some pieces of feedback:
- Isn't it possible to make that a preference, like `Display File Path`, and show it in a different accessory instead of in the tooltip?
- Is there a more performant way to find the path of each file? Right now, there can easily be dozens and dozens of calls fired (especially since it's implemented in a recursive way) to display the file path. Also, it'd be probably better to add all of that inside the [`useFetch` call](https://github.com/raycast/extensions/blob/5174e21ce15e1aef2f1a9650cd8659d9dc696f21/extensions/google-workspace/src/search-google-drive-files.tsx#L17) to make all of the information display at once.

ridemountainpig:
> Maybe you can turn this hook into a useCachedPromise to add the file path information from there since I'm not sure it's possible to do that from useFetch.

It seems that `useFetch` can't be used inside `useCachedPromise` because both are hooks that handle asynchronous data fetching independently. React hooks must be called at the top level, and combining them would violate that rule.

ridemountainpig:
@thomaslombart, do you mean to change it to this?

PR #14896, opened 2024-10-12
Add "Create anonymous identity" command to Migadu extension

### LINKED ISSUE #14890: [Migadu] Shortcut to create random identities, similar to Apple's "Hide My Email"
### Extension

https://www.raycast.com/xmok/migadu

### Description

Apple has a "Hide My Email" feature that will generate a random, unique email that you can use to sign up for random services. Using Migadu's API and Identities, a similar functionality could be implemented without the dependency on Apple.

To this end, it would be useful to have a command that basically just takes the name of the identity you want to create (which would probably end up being the name of the service you're generating the email for) and then creates the identity and copies the new email address to your clipboard. I would suggest a dropdown that allows you to change the mailbox for which the identity is created, but I think it should be pre-populated with a pre-selected default mailbox. It might also be cool to allow for customization of the created identity, i.e. having an input that's prefilled with a random and unique email address.

Any migadu user who would like to use "Hide My Email", but without depending on Apple.

### Anything else?

I'd be happy to implement this myself! All I need is the go-ahead that this feature would be appreciated. I've started looking into the code and saw that queries are currently handled via vanilla fetch. Would you mind if I migrate them to [TanStack Query](https://tanstack.com/query/latest)? This would remove the need for a manually managed `isLoading` state and enable some useful things in the future. [Other extensions seem to also use this library](https://github.com/search?q=repo:raycast/extensions%20@tanstack/react-query&type=code), so I'm optimistic it won't cause any significant issues.

LBBO:
## Description

Adds a new "Create Anonymous Identity" command (closes #14890). This command will prompt the user for a service name, auto-generate a local part and pre-select the most recently chosen mailbox (or the first one that exists by no particular order). 

The primary action will create an identity for the chosen mailbox with the chosen name and local part. The address will be copied to clipboard. The secondary action will create the same identity without copying it to clipboard.

There are two more actions: one to re-generate the local part. It can also be edited manually, but this action allows the user to quickly replace a string they don't like. The generated values will have the form `adjective-color-animal` from [`unique-names-generator`](https://www.npmjs.com/package/unique-names-generator). According to their docs, this results in 24.500.000 possible values. Currently, no measures are taken to guarantee uniqueness as the sample size is assumed to be large enough to avoid collisions in most cases. In the case of collision, the user will see an error message and can quickly generate a new local part.

The last action opens the GitHub issue page for this extension. This is because there's a sanity check in the code that should never fail (if it does, it's a programming error). If it does, I wanted to make it as easy as possible for users to file a bug report.

## Screencast

Omitted for privacy reasons. If it's necessary, I can create one using mock data.

## Checklist

### DIFF
--- a/extensions/migadu/CHANGELOG.md
+++ b/extensions/migadu/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Migadu Changelog
 
-## [Add Anonymous Identity Command] - {PR_MERGE_DATE}
+## [Add Anonymous Identity Command] - 2024-10-16
 - Create new command to create anonymous identities (made to imitate Apple's Hide My Email)
 
 ## [Cache Results + Change Delete Shortcut] - 2024-08-21


extensions/migadu/package.json
```diff
@@ -56,12 +59,19 @@
       "subtitle": "Migadu",
       "description": "View, Create, Update or Delete Rewrites in your Migadu Account",
       "mode": "view"
+    },
+    {
+      "name": "create-copy-anonymous-identity",
+      "title": "Create Anonymous Identity (Custom Hide My Email)",
```
LBBO, line 65: I chose this lengthy name so that users can search for "Hide My Email" and find this command. If it's too long, maybe we can find another way to enable that?

extensions/migadu/src/utils/api.ts
```diff
@@ -28,8 +28,17 @@ const headers = {
   "Content-Type": "application/json",
 };
 
-const callApi = async <T>(endpoint: string, method: APIMethod, body?: BodyRequest) => {
-  await showToast(Toast.Style.Animated, "Processing...");
+export type ApiToastVerbosity = "verbose" | "error";
```
LBBO, line 31: I added this because the fetch of all mailboxes was triggering a "success" toast the moment the user opens the command. IMO, this isn't useful information, hence this refactoring. Let me know if you disagree or want this implemented some other way.

extensions/migadu/package.json
xmok: We can use `keywords` for a command (https://developers.raycast.com/information/manifest#command-properties) but the title seems OK to me.

extensions/migadu/src/create-copy-anonymous-identity.tsx
```diff
@@ -0,0 +1,151 @@
...
+      if ("error" in response) {
+        // the API call already shows an error message. If all calls fail, no email can be selected, which is fine.
+        return [];
+      }
+      return response.mailboxes;
+    });
+    return {
+      data: mailboxes,
+      hasMore: false,
+    };
```
xmok, line 34:
```suggestion
    return mailboxes;
```
Since we're not using Pagination we could simplify the response to only be the data.

extensions/migadu/src/create-copy-anonymous-identity.tsx
```diff
@@ -0,0 +1,151 @@
...
+  }, []);
```
xmok, line 35:
```suggestion
  }, [], {
  initialData: []
  });
```
Optionally, we can set the `initialData` so we can use the data w/o having to always do a null check.

extensions/migadu/src/create-copy-anonymous-identity.tsx
```diff
@@ -0,0 +1,151 @@
...
+    onSubmit: async ({ mailbox, ...partialIdentity }: FormValue) => {
+      const fullNewIdentity: IdentityCreate = {
+        ...partialIdentity,
+        footer_active: false,
+        footer_html_body: "",
+        footer_plain_body: "",
+      };
+
+      // By definition of the drop down items, this must exist. Just in case it doesn't, show a nice error message.
+      const fullMailbox = mailboxesQuery.data?.data?.find((curr) => curr.address === mailbox);
```
xmok, line 52:
```suggestion
      const fullMailbox = mailboxesQuery.data.find((curr) => curr.address === mailbox);
```

xmok: If we change the response to just be the data, don't forget to tweak all the uses of `data` accordingly.

extensions/migadu/src/create-copy-anonymous-identity.tsx
```diff
@@ -0,0 +1,151 @@
...
+            title="Create Without Copying"
+          />
+          <Action
+            onAction={regenerateLocalPart}
+            title="Generate New Local Part"
+            shortcut={{
+              key: "r",
+              modifiers: ["ctrl"],
+            }}
+          />
```
xmok, line 112: Would be consistent if all of these have `Icons`.

extensions/migadu/src/utils/api.ts
xmok: That's fine. This was one of the earlier extensions I authored so it doesn't have a lot of best practices - we might re-write this function and phase out said Toasts eventually.

xmok (changes requested):
The function itself seems to work great!

Could you add an `Action` to `IdentitiesCreate` in components/Identities.tsx that uses `unique-names-generator` to fill in a random alias should the user choose so? i.e. add a new Action after `Submit Form` e.g. "Generate Random Address" and it uses `setValue` to set a new identity 😄 - that way the functionality is also available in existing command.

LBBO:
Thanks for the helpful review! I agree with all of your comments and really like the suggestion to add the generation to the regular creation form.

I've implemented all of the suggestions. I also saw that the creation 1) uses the term "Address" instead of "Local part" and 2) displays the resulting email address, so I decided to do both of those things in the new command as well for consistency and a nicer UX. Let me know if there's anything else for me to do!

extensions/migadu/CHANGELOG.md
```diff
@@ -1,9 +1,12 @@
 # Migadu Changelog
 
+## [Add Anonymous Identity Command] - 2024-10-13
```
xmok, line 3:
```suggestion
## [Add Anonymous Identity Command] - {PR_MERGE_DATE}
```

xmok: Just a couple of items left:
- update the CHANGELOG date
- update the README to mention the new command (you could add a note mentioning one can use the new command for quick identity generating)

After that LGTM 🚀

xmok (approved):
LGTM 🚀.
A member of the Raycast Team will review and do the final approval.

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

Issue #14920, opened 2024-10-14
[Word Search] Details view enabled by default


cynicaljoy:
### Extension

https://www.raycast.com/rishabswift/word-search

### Description

Add setting to enable Details View to be enabled by default.

![image](https://github.com/user-attachments/assets/fc56742b-2d2e-424d-82e0-b8244f820f9b)

![image](https://github.com/user-attachments/assets/9be24090-0bc5-4ce1-a2ee-d96fb74d2468)

### Who will benefit from this feature?

Save some keystrokes for those that want to always have the details view available when reviewing spell results.

bkeys818:
@raycastbot close

PR #14923, opened 2024-10-14
Update word-search extension

j3lte:
## Description

- Added a preference to open details by default (#14920) 

## Checklist

### DIFF
--- a/extensions/word-search/CHANGELOG.md
+++ b/extensions/word-search/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Word Search Changelog
 
-## [Default open details] - {PR_MERGE_DATE}
+## [Default open details] - 2024-10-17
 
 - Cache the 'View Details' state, keeping it open ([#14920](https://github.com/raycast/extensions/issues/14920))
 - Updated dependencies


RishabSwift:
Can we have it so that the details view is not enabled by default? I think the toggle is a good option in preferences but seeing all definitions at once might be better

j3lte:
@RishabSwift by default it is off

pernielsentikaer:
Another approach could be using `useCachedState` to remember if it's open or not. This way, it wouldn't require a new setting. Would that be a good idea?

j3lte:
@pernielsentikaer agreed, I'll update the PR

j3lte:
@RishabSwift @pernielsentikaer I updated the code. Removed the introduced setting and instead just cache the 'Show Details' state

pernielsentikaer:
Awesome 👏 That feels like a better solution - do you agree @RishabSwift

RishabSwift:
That’s awesome, I like this solution!

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #14938, opened 2024-10-15
[Mattermost] fix `Search Channel` not loading + better error handing + move some `useEffect` to `usePromise` + add Preference to launch app + MORE

### LINKED ISSUE #14876: [Mattermost] ...Geting 401 error using lattermost search extension
### Extension

https://www.raycast.com/ernest0n/mattermost

### Raycast Version

newest version just downloaded it on 10-10-24

### macOS Version

15.1

### Description

Geting 401 error using Mattermost search extension. It does show that I logged in successfully.

### Steps To Reproduce

using the search feature, instantly responds with 401 error

### Current Behaviour

na

### Expected Behaviour

na

xmok:
## Description

- Fix: `Search Channel` not loading when ZERO channels (and related)
- `signIn` function is now `async`
- Show better error message when Authorization fails (should close #14876)
- `usePromise` in `withAuthorization` component for better control
- In `search-channels` you can specify in Preferences whether to launch installed Mattermost application or not
- Fix: various typos
- Add an `Axios` intercept for robust error handling
- Update dependencies

## NOTES

Extension would show `Success Toast` with "Authorized" even when not done properly which might be the reason for the mentioned issue. During testing, suddenly all my attempts to `signIn` began to fail which is why I added an intercept for error handling which is how I found the following error:

mattermost-account-locked

`usePromise` has replaced `useEffect` in some places for this very reason i.e. the hook will be able to display `Failure Toast` by itself. However, to avoid previous caches being invalidated, the `LocalStorage` calls have not been removed.

## Screencast

Screencast for a user who has been added to a fresh server but is part of no channels:

https://github.com/user-attachments/assets/63550a0f-7a45-487e-a4c8-da23699c495f

Screencast for a user who has been locked out and needs to reset password:

https://github.com/user-attachments/assets/5edc0223-db32-4513-9cb9-a67b445362fe

Screencast for a user who has successfully authorized:

https://github.com/user-attachments/assets/94b95d3f-f8ff-4bae-9401-e788d5cf4832

## Checklist

### DIFF
--- a/extensions/mattermost/CHANGELOG.md
+++ b/extensions/mattermost/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Mattermost Changelog
 
-## [Enhancements] - {PR_MERGE_DATE}
+## [Enhancements] - 2024-10-22
 
 - Fix: `Search Channel` not loading when ZERO channels (and related)
 - `signIn` function is now `async`


extensions/mattermost/src/channels.tsx
```diff
@@ -265,30 +255,32 @@ function ChannelList(props: { profile: UserProfile; team: TeamUI }) {
...
 
   async function openChannel(channel: ChannelUI) {
     console.log("select channel", channel);
     const baseDeeplink = preference.baseUrl.replace("https://", "mattermost://");
     const fullDeeplink = baseDeeplink + "/" + team.name + channel.path;
     console.log("open deeplink", fullDeeplink);
-    open(fullDeeplink);
-    closeMainWindow();
+    try {
+      await open(fullDeeplink);
```
Ernest0-Production, line 274: Are you sure this is working correctly? After the application has opened, it takes some time for the launch screen to disappear. If you try to open a deep link earlier, this deep link will most likely be ignored. (I remember encountering a similar problem. I will try to check this behavior again)

Ernest0-Production (approved):

extensions/mattermost/src/channels.tsx
xmok: I'm not using the Mattermost application so I don't really know 🙁.
I simply enclosed the code in a try...catch otherwise it is the same as you wrote it originally. One thing we could do is check if the application is running first e.g. through `run-applescript` and if app is running, THEN open the deeplink otherwise show a failure toast asking user to launch Mattermost application. Does that make sense?

xmok: OK check out the latest commit: 
we have an AppleScript function that checks if Mattermost is running first.
Hopefully that helps.

Ernest0-Production: Oops, sorry. I didn't notice that the logic for opening the app under the launchMattermostAppOnLoad flag was still there.
In that case, the current behavior shouldn't break.

----

However, instead of the preference flag launchMattermostAppOnLoad, it is worthwhile to automatically determine whether the application is selected as a method for opening deep links.
Examle of new preference in package.json
```json
 {
      "name": "application",
      "type": "dropdown",
      "required": false,
      "title": "Open In",
      "data": [
        {
          "title": "Mattermost Application",
          "value": "mattermost"
        },
        {
          "title": "Browser",
          "value": "browser"
        }
      ],
      "description": "How to open a deep link",
      "default": "mattermost"
    }
```

And then add expression:
```
if (preference.application === "mattermost") {
   runAppleScriptSync('tell application "Mattermost" to launch');
}
```

And compute url  to open:
```ts
const baseDeeplink = (() => {
      switch (preference.application) {
        case "browser": return preference.baseUrl
        case "mattermost": return preference.baseUrl.replace("https://", "mattermost://");
      }
    })()
```

xmok: Love this idea.
Yeah let's go with this it makes a lot of sense!

xmok: I used slightly different wording to what you suggested:
`deepLinkType` for the "name" as it's more obvious than "application"
`application` instead of "mattermost" for application option as that is also more precise.

Ernest0-Production (approved):

Ernest0-Production:
Thank you very much for contribution

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #14967, opened 2024-10-18
Add raycast-google-gemini-chatbot extension

blindFS:
### Motivation of creating another extension for google gemini

There's already a fantastic google gemini extension thanks to @EvanZhouDev 

However it lacks some new features like code execution on-the-fly, document based QA, online RAG.
Besides, I personally is not a fan of how it manages its chat history where each entry only contains single query, and switching threads need mouse interactions on the top-right. 

image

### How is this new extension different?

1. Focuses mainly on multi-round conversations. 
  - Action defaults to continue the conversation instead of copy the answer
  - Manages chat history in threads, one entry for all messages within the same topic, with attached contents.

raycast-google-gemini-chatbot-6

2. Image/math latex preview in the view

raycast-google-gemini-chatbot-1

3. New features of official gemini-api like code execution, doc QA, function call based online RAG

### Why creating new instead of contributing to the existing one?

1. Current gemini extension uses a third-party gemini api under the hood, which is not as feature rich as the official one. Replacing it will lead to a large overhaul of the code base.
2. Chat history management/default enter behaviour is  more like a personal choice instead of improvement. However it is not possible for a single extension to provide both convenience, so it might be a reason for providing another choice from the extension store.

## Screencast

raycast-google-gemini-chatbot-2

raycast-google-gemini-chatbot-4

## Checklist

### DIFF
--- a/extensions/raycast-google-gemini-chatbot/src/api/utils.jsx
+++ b/extensions/raycast-google-gemini-chatbot/src/api/utils.jsx
@@ -140,7 +140,7 @@ async function retrieveByUrl(fileManager, urlText = "") {
   // default behavior for URL
   var rawHTMLObject = await rawHTMLByURL(fileUrl);
   rawHTMLObject.content = bufferToGenerativePart(rawHTMLObject.content, "text/html");
-  return rawHTMLByURL;
+  return rawHTMLObject;
 }
 
 export const retrievalTypes = {


pernielsentikaer:
It would still be better to enhance the current so we don't end up with two almost identical extensions in the store, that will confuse a lot. Maybe @EvanZhouDev is up for an enhancement of the current one in the store 🙂

EvanZhouDev:
@pernielsentikaer Sure, I'm up for it. @blindFS feel free to send a PR. I've seen your extension, and the features look great. I realize it may take a bit of rewriting, but feel free to contribute to mine.

blindFS:
Cool, I'm happy that you like my idea. This even saves me some maintenance effort.

EvanZhouDev:
Feel free to DM me on the Raycast Slack (Evan Zhou). I realize there are many discrepancies especially in the API used, so perhaps we can get the `gemini-ai` library up-to-speed with those features and use it in that way.

PR #15050, opened 2024-10-23
[Dropbox][The Blue Cloud] Extension Renamed + new OAuth ID

xmok:
## Description

To get approved for OAuth we need to comply with [Dropbox Branding Guidelines](https://www.dropbox.com/developers/reference/branding) so we are renaming the extension and giving it a new logo - no functionality has changed.

## Screencast

N/A

## Checklist


xmok:
@pernielsentikaer I went ahead and changed the OAuth ID in the extension to the new one since I'm guessing users might get logged out due to the changes so might as well let them log into the new one. If you disagree lmk and I'll swap back the ID to the old one so as to not log out those who are already logged in.

pernielsentikaer:
Do you think you can stash the two file from the Dropbox folder, our CI don't link the that there is two folders 🙂

xmok:
@pernielsentikaer I had to try a couple of times as git kept on considering the metadata image as a new one. Finally solved by parsing the files manually and force pushing 😭.

QUESTION: I have left the Command title as "Dropbox Files Search" - as that seems to add context. But should we retitle it to be safe?

 It's running locally so all good:

https://github.com/user-attachments/assets/b3182754-dd36-493a-b189-a00bfd65a21a

pernielsentikaer (changes requested):
- I'm getting this error if I just type space, are you able to check that?

the-blue-cloud 2024-10-31 at 10 55 38

- First time I connect to it I'm getting this error, do you have any clue about that one?

the-blue-cloud 2024-10-31 at 10 58 30

xmok:
@pernielsentikaer I was able to reproduce the issues.

Honestly it took me way longer fixing the CI issues 😭.
- Moved to Raycast's OAuth (https://developers.raycast.com/utilities/oauth) solves the first `Bad Request` issue.
- The "error on space" is an API issue because Dropbox doesn't understand the argument and, in fact, this also happens on their own site observing the Network. I have added an EDGE CASE check until the issue is fixed upstream.

FInally, I gave up fighting the CI so there are 2 files which we can remove later i.e. 
- the metadata image dropbox-1
- the (now unused) `auth.ts` file

pernielsentikaer:
Works better for me now, let's get it in so we can continue

PR #15054, opened 2024-10-23
Update `Slugify` extension - support 'strict' especially useful for apostrophes

xmok:
## Description

Quick little update because I was using the extension today to slugify some blog post titles and the _apostrophes_ were not removed which is not ideal for post URLs. So the extension also shows `Strict` slugs.

## Screencast

https://github.com/user-attachments/assets/42d90620-5764-4b56-a016-60c481c9e33a

## Checklist


unreadablename:
Thanks @xmok for such great finding and work. But let's keep old number of menu items but make them strict (without strict badge)
What do you think?

xmok:
I'm fine with that but just to give the option, we could do what I did in a previous iteration of the extension i.e. we can move the Strict options into a `List.Dropdown` defaulting to **Strict** - this way on the off-chance someone _does_ want the non-strict option, they can select it.

xmok:
Updated screencast 😄.

https://github.com/user-attachments/assets/fa78f179-7a52-49e3-b7a9-49e97719b5d7

unreadablename:
@xmok thanks, let's go this way

PR #15102, opened 2024-10-27
Add to-streamshare extension

0PandaDEV:
## Checklist

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -855,7 +855,7 @@
 /extensions/macupdater                                              @karlsander
 /extensions/mail                                                    @yug2005 @ikupenov @pernielsentikaer @fengtianze @tleo19 @Luk4h @jfkisafk @StevenRCE0 @ridemountainpig
 /extensions/mailboxlayer                                            @xmok
-/extensions/mailerlite-stats                                        @nklmilojevic
+/extensions/mailerlite-stats                                        @nklmilojevic @xmok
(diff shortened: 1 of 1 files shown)


0PandaDEV:
I would like to replace the current version of the extension called Streamshare Uploader with this extension here.

pernielsentikaer (changes requested):
Then you would need to "move" the files from the old folder into the new (and change the content, does that make any sense)

0PandaDEV:
So I need to also delete the old extension?

pernielsentikaer:
Fork the original extension, move files to the new folder 🙂

0PandaDEV:
> Fork the original extension, move files to the new folder 🙂

No, that does not work because I changed the identifier name so it gets detected as a new extension by the ray CLI.

pernielsentikaer:
Can we start with a PR where we only move all files to the new directory instead? Our CI doesn't like multiple folders 🙂

0PandaDEV:
> Can we start with a PR where we only move all files to the new directory instead? Our CI doesn't like multiple folders 🙂

Yeah, sure, but I don't get what you mean by moving files.

pernielsentikaer:
If you delete your local copy and fork the updated one from store, then it should be easier to make the modifications

0PandaDEV:
Yes, I tried that, but now I get this when I try to publish again.
image

Issue #15058, opened 2024-10-23
[AirPods Noise Control] ...


JuanCaarPapers:
### Extension

https://www.raycast.com/chrahe/airpods-noise-control

### Raycast Version

1.84.3

### macOS Version

15.0.1

#### Error:
```
Error: Command failed with exit code 1: osascript
3358:3365: execution error: La variable menuBar no está definida. (-2753)

d0:run-applescript.js:38:21

---
35:         timedOut,
36:         command: "osascript",
37:         options,
38:         parentError: new Error(),
39:     })

### Steps To Reproduce

1. Configure as normal the extension
2. Try Switch Noise Control Mode

### Current Behaviour

<img width="746" alt="Captura de pantalla 2024-10-23 a las 23 19 25" src="https://github.com/user-attachments/assets/4a31608e-2a86-4bdc-a9c7-c407fde5d7ed">


### Expected Behaviour

_No response_

JuanCaarPapers:
The error in english is "The menuBar variable is not defined"

PR #15120, opened 2024-10-27
[AirPods Noise Control] Fix index of AirPods expand toggle

dnlfrst:
## Description

On macOS 15 (Sequoia), the extension no longer works, while previous macOS versions had no issues. The internal order of user interface elements in the sound menu changed with macOS 15. While earlier versions of macOS had the expansion toggle placed before the AirPods menu item, the toggle is now placed after the corresponding menu item.

After changing the corresponding index in the extension's AppleScript from `i - 1` to `i + 1`, the extension works as expected. Unfortunately, this fix probably breaks the extensions for other versions of macOS. Is there a different way to expand the detailed menu of the AirPods menu item? Anything through accessibility, maybe? This would also make the extension more future-proof.

This fix may be related to the problem described in #15058.

## Screencast

_N/A_

## Checklist

### DIFF
--- a/extensions/airpods-noise-control/CHANGELOG.md
+++ b/extensions/airpods-noise-control/CHANGELOG.md
@@ -1,6 +1,6 @@
 # AirPods Noise Control Changelog
 
-## [Bug Fix] - {PR_MERGE_DATE}
+## [Bug Fix] - 2024-11-14
 
 - Fixed incorrect sound menu expand toggle index computation on macOS Sequoia
 - Gracefully handle AppleScript runtime errors


pernielsentikaer:
There are other extensions looking up version to do something different, maybe we can do the same here? 

https://github.com/raycast/extensions/blob/3ecb72fc6119efab3178943c440364680739ce72/extensions/toggle-desktop-visibility/src/utils/index.ts#L15

Could you add an entry in the `CHANGELOG.md` file? Here's our guidelines about that: [Version History](https://developers.raycast.com/basics/prepare-an-extension-for-store#version-history).

pernielsentikaer:
That's much better 👏 It seems like the extension is able to crash, can you handle that gracefully?

airpods-noise-control 2024-11-13 at 12 22 13

dnlfrst:
I did not encounter this issue, yet. Neither was I able to reproduce the error using any of the extension's settings or commands. Do you know how to reproduce the error?

Regardless, I have added graceful handling of AppleScript runtime errors. Hence, the extension should no longer crash.

PR #15160, opened 2024-10-29
[Airtable] update the OAuth ID to fix login not working (scope mismatch)

### LINKED ISSUE #15159: [Airtable] OAuth errors connecting to Raycast
### Extension

https://www.raycast.com/marks/raycast-airtable-extension

### Raycast Version

1.84.7

### macOS Version

14.5 (23F79)

### Description

When I try to connect Airtable to Raycast, I get the following error:
> Your OAuth application requested the scope "schema.bases:read data.records:read data.records:write" which your application does not have access to. You can edit this in your OAuth Application's settings 

Note that this error happens even though the Raycast web page says that I'm "Successfully connected to Airtable."

Also, the "Open Raycast" button on the Raycast website doesn't seem to work.

### Steps To Reproduce

1. Install the Airtable extensions
2. Run the "List Bases" command
3. Try to perform OAuth.

### Current Behaviour

An OAuth error is received:
> Your OAuth application requested the scope "schema.bases:read data.records:read data.records:write" which your application does not have access to. You can edit this in your OAuth Application's settings 

### Expected Behaviour

It should connect to OAuth correctly.

xmok:
## Description

This will close #15159 and fix the scope issue by replacing the ID with a new one.
The Airtable Team hasn't update the scopes yet so we switch to another one that has proper scopes. I have mentioned it in README so if future changes are needed, one knows who to reach out to.

I have added comments for the logic of hard coding OAuth IDs and checking them.

## Screencast

N/A - logs in fine. 

## Checklist

### DIFF
--- a/extensions/airtable/src/BaseList.tsx
+++ b/extensions/airtable/src/BaseList.tsx
@@ -53,8 +53,10 @@ function AirtableBaseListItem(props: { baseMetadata: AirtableBaseMetadata }) {
             target={<AirtableBaseDetails baseMetadata={baseMetadata} />}
             shortcut={{ modifiers: ["cmd"], key: "d" }}
           />
-          <Action.OpenInBrowser title="Open Api Docs in Browser" url={baseMetadata.apiDocsUrl} />
+          {/* eslint-disable-next-line @raycast/prefer-title-case */}
+          <Action.OpenInBrowser title="Open API Docs in Browser" url={baseMetadata.apiDocsUrl} />
           <Action.CopyToClipboard
+            // eslint-disable-next-line @raycast/prefer-title-case
             title={`Copy Base ID (${baseMetadata.id})`}
             content={baseMetadata.id}
             shortcut={{ modifiers: ["cmd"], key: "i" }}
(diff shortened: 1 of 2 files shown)


pernielsentikaer:
@marks, could you take a look here? I don't use Airtable myself

marks:
I do not use Airtable anymore either. At quick glance, this change looks fine, though I think it would have been better to have Airtable update the existing OAuth app :\ (but it seems @xmok tried and hence the need for this PR)

xmok:
I'll send another email as maybe the first one got drowned in other threads?

It makes sense to use the OG OAuth App as it doesn't break UX. The only reason I can think of for a new App is so it's easier/quicker if we need to update scopes again in the future but keeping the old one makes more sense to me still as we can always ask Support to update accordingly and it won't log out existing users.

I suppose it's your call @marks as the author.
We can wait and I'll email them again otherwise yeah we can proceed with this PR.

marks:
Hey @xmok - I agree having Airtable fix is the purest solution, but it sounds like you're unfortunately not getting good responsiveness from them (sorry).

Since #14864 is already merged and people are running into this issue, I guess it makes sense for you to change the client ID.

@pernielsentikaer - do you all usually expect individuals to own default client IDs like this?

pernielsentikaer:
No I don't think we do that often, let's merge and see how it goes 🙂

PR #15166, opened 2024-10-30
Update restart-system-processes extension

### LINKED ISSUE #15148: [Restart System Processes]: Improve UX
### Extension

https://www.raycast.com/muzhen_gaming/restart-system-processes

### Description

The extension is working fine and I really like it, however, there are some UX details that could be improved.

This is the current UI:
grafik

I suggest the following changes (open for discussion):
- Replace the select with a [list](https://developers.raycast.com/api-reference/user-interface/list)
- in that list, display the most recent command on top (I only need it for restarting audio, however, finder is always the first)
- The sudo option for each command can only be toggled by pressing Tab and Space. Also, that is not remembered. We could move that to [preferences](https://developers.raycast.com/api-reference/preferences)
- The advanced mode option has nothing to do per process like the sudo option, but rather switches the available options in the select dropdown. This should also be moved to the preferences.

All users of that extension for a faster, more intuitive and raycast-like experience.

### Anything else?

I'd be happy to implement all those changes, but I wanted to discuss them first and get feedback 🙌
Thanks for reading and your input!

hrueger:
## Description
fixes #15148

## Checklist

## ToDo
- [x] Move Sudo option to preferences
- [x] Use nested lists instead of dropdown form
- [x] Order list by most recent used
- [x] Add Changelog
- [x] Add me as contributor (was removed while publishing somehow)
- [x] Update Screenshots


XInTheDark:
@hrueger Amazing update - just tested it and it works well! I really like the new design :)

A small detail: the list doesn't re-render immediately after an item is restarted; it's only reordered the next time the command is launched. The list also reorders when an item is selected then canceled, which is probably unnecessary.

Regarding the feature suggestion, it's probably unnecessary for now - advanced mode is mainly just a shortcut for users who would otherwise type the command manually in terminal.

I'll generate the screenshots now, and link it here - I don't have write access to this PR I think :)

XInTheDark:
here are the screenshots uploaded to github

extensions/restart-system-processes/src/index.tsx
```diff
@@ -0,0 +1,266 @@
...
+  const sudo = sudoOption ? "sudo" : "";
+  const processName = "process" in values ? values.process.label : values.advancedMode;
+  let cmd = "";
+
+  if (!process) {
+    await showToast(Toast.Style.Failure, "No process selected");
+    return;
+  }
+
+  await saveLastUsedItems("process" in values ? values.process.process : values.advancedMode);
```
XInTheDark, line 200: this should be moved after the user confirmation, so it only updates the order if the action actually runs

extensions/restart-system-processes/src/index.tsx
```diff
@@ -0,0 +1,266 @@
...
+
+  clearSearchBar();
+  closeMainWindow();
```
XInTheDark, line 203: Missing await calls

hrueger:
Thanks for the review, should all be fixed now 👍

hrueger:
> A small detail: the list doesn't re-render immediately after an item is restarted; it's only reordered the next time the command is launched

I tried to implement this, however, I was not be able to make it work. I moved the fetch and state update logic to a small helper:
```ts
  const [orderedProcesses, setOrderedProcesses] = useState<Process[]>([]);
  const fetchProcesses = async () => {
    const processes = await orderItems(commonProcesses);
    setOrderedProcesses(processes);
  };

  useEffect(() => {
    fetchProcesses();
  }, []);
```
and I tried calling this after `performAction` in the `onAction` prop value or in the `performAction` itself. However, it seems like that must happen before the main window is closed, but we only change the order after closing the main window. So there's kind of a deadlock there 😅
I think we could only close the main window later, but I prefer the current user experience and would rather skip updating the list immediately. What do you think?

XInTheDark:
Ah sure, in that case I think we can just leave it as it is - not a big deal either way :)

XInTheDark (approved):
LGTM, great work!

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #15169, opened 2024-10-30
[Zeplin Project Search] update deps + make users 🪝 more robust + view details of proj + add CHANGELOG,images,cats

xmok:
## Description

- fix: command would be stuck on `isLoading` and show 2 toasts if invalid token - now we check for valid user then load everything else
- See Project Status
- Toggle Details of a project
- Add CHANGELOG
- Add metadata images
- Add categories

The `useGetCurrentUser` no longer fetches from `LocalStorage` since the hook is used to determine if the API token is valid or not so always needs to be fresh. The other hooks are as they were simply tweaked to throw errors we can handle through `usePromise`.

### BEFORE
https://github.com/user-attachments/assets/0ab5519b-0548-4170-818e-c8bbe38a1f49

### AFTER
https://github.com/user-attachments/assets/c01a158e-ec24-4277-97db-d2242cf5cc87

## Checklist


pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

dorukgunes:
Hi @xmok, sorry for the late reply. Thank you for diving in and updating the extension, it looks great 💯

Issue #15184, opened 2024-10-31
[Pomodoro] Set 'Focus' mode when starting a Focus timer


alexandrelissowski:
### Extension

https://www.raycast.com/asubbotin/pomodoro

### Description

Would it be possible to trigger the 'Focus' or 'Do Not Disturb' mode on the mac when the Focus timer is started. 

### Who will benefit from this feature?

All user who don't want to be annoyed with notifications during a focus time.

mikikiv:
@raycastbot assign me

LitoMore:
Resolved in #15193.

PR #15193, opened 2024-11-01
#15184 Enable DND when focus is enabled

mikikiv:
## Description

#15184 Adds the settings option to enable Do Not Disturb when focus mode is enabled.
Borrows some code and the shotcuts file from the DoNotDisturb Raycast extension by @Yakitrak 

## Checklist


mikikiv:
There is a bug I cant quite figure out where the DND is not able to be enabled always by the toolbar thing. Considering how many complaints we get about that thing not working right in general (hard to get it to disappear from the menu bar) maybe something else is causing this problem too.

Yakitrak:
Hi this seems like a good idea to add to Pomodoro. But if its the identical usage as the DND extension and shortcut, would it worth just using a raycast link and triggering DND when on focus mode. `e.g. raycast://extensions/yakitrak/do-not-disturb/on` Of course you'd need both extensions installed but could be easier.

pernielsentikaer:
I think @LitoMore made something that could ease this (link between them) - Maybe it's not that smart here but I'll let him answer :)

LitoMore:
Hi. I think @pernielsentikaer mentions the [Raycast Cross-Extension Conventions](https://github.com/LitoMore/raycast-cross-extension-conventions) I created for this situation.  

Here is the code for you:

```ts
import { LaunchType } from '@raycast/api';
import { crossLaunchCommand } from 'raycast-cross-extension';

const dndLaunchOptions = {
	type: type: LaunchType.Background,
	extensionName: 'do-not-disturb',
	ownerOrAuthorName: 'yakitrak',
};

export async function setDND = (enabled: boolean) => {
	return crossLaunchCommand({
		...dndLaunchOptions,
		name: enabled ? 'on' : 'off'
	}, false).catch(() => {
		// Do nothing here because we're going to check when mounting the extension
	})
}
```

This is a feature related to the `enableFocusWhileFocused` preference settings. So I would suggest checking if the `yakitrak/do-not-disturb` is installed when rendering your commands.

```ts
import { confirmAlert, open } from '@raycast/api';

useEffect(() => {
	// Here we use the `status` command to check if the extension is installed
	if (!enableFocusWhileFocused) return;
	crosssLaunchCommand({
		...dndLaunchOptions,
		name: 'status',
		context: {
			// The `status` command will show a HUD when running.
			// We can choose to suppress it for better user experience.
			// But you must edit some code in the `yakitrak/do-not-disturb` extension to receive this parameter.
			suppressHUD: true,
		}
	}).catch(async () => {
		const yes = await confirmAlert({
			title: "Need to Install Additional Extension"
			message: 'The "Enable Do Not Disturb mode while focused" feature requires the "Do Not Distrub" extension, do you want to move to the install page now?',
		})
		if (yes) {
			await open('raycast://extensions/yakitrak/do-not-disturb');
		}
	});
}, [])
```

mikikiv:
Thats neat, didnt know that was an option too! I went a different route after my initial try with the other extension led to some weird user experiences at the end of a focus session.
The pomodoro timer would loop when it ended rather than show any of the completion things enabled by Pomodoro (quote, picture, confetti)

mikikiv:
the last 2 commits was meant for another issue

mikikiv:
This MR is ready for review

pernielsentikaer:
@KELiON can you check this

LitoMore:
I prefer to use the cross-extension solution rather than the shortcut binary file.

Let’s ignore the security of binary files. If the Apple Shortcut API changes, the author will need to update each file individually.

The cross-extension solution is much easier to maintain and update.

mikikiv:
Also created https://github.com/raycast/extensions/pull/15433 for compatibility with that extension. 
@LitoMore is there any known issues with triggering crossLaunchCommands from menu-bar items? The function `checkDNDExtensionInstall` is not triggering opening the store page when its triggered by starting an interval from the menu bar. Any idea how to remedy that?

extensions/pomodoro/package.json
```diff
@@ -132,6 +135,10 @@
         {
           "value": "90",
           "title": "90:00"
+        },
+        {
+          "value": "0.1",
+          "title": "6 seconds"
```
KELiON, line 141: I think it was added just for testing purposes? I'd remove it

KELiON (approved):
I like the idea (and I'm sure I'll use it myself), let's merge it 👍

LitoMore:
> The function checkDNDExtensionInstall is not triggering opening the store page when its triggered by starting an interval from the menu bar.

Let's ignore this for now. I will check this after your PR.

mikikiv:
This is all ready for review again. We should aim to have #15433 updated either before or in sync with this.
Thanks for all the guidance and suggestions! @LitoMore @KELiON

pernielsentikaer:
Let's get it live to see how it goes 🙂

haydencbarnes:
I will take a look at this later this evening but the UX with this change is not the prettiest 🫠 I was basically forced to download the new DnD extension and use it or else it breaks the extension. Plus, the extension does not appear adhere to whatever the set preferences are when I turn the DnD feature off.

LitoMore:
I will file a PR to solve this issue.

PR #15265, opened 2024-11-07
Update stripe extension

bradgarropy:
## Description

Hey @pradeepb28 and @farisaziz12,
I'm @bradgarropy and I work at Stripe. I ported over our internal `Fill Checkout` command to this public Raycast extension. We thought it would be good to spread this to other developers!

## Screencast

https://github.com/user-attachments/assets/53e779bc-e9a1-46dd-974c-85e6a156a1e3

fill-checkout
cards
card-search
card-actions

## Checklist

### DIFF
--- a/extensions/stripe/CHANGELOG.md
+++ b/extensions/stripe/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Stripe Changelog
 
-## [Fill Checkout] - {PR_MERGE_DATE}
+## [Fill Checkout] - 2024-12-04
 
 Added the `Fill Checkout` command that uses Stripe's test cards to automatically fill in a checkout form.
 


bradgarropy:
The checks are failing due to some type errors that were present in the application before my changes.

Would y'all like me to fix those as well?

extensions/stripe/package.json
```diff
@@ -118,7 +126,6 @@
     "eslint": "^7.32.0",
     "eslint-config-prettier": "^8.3.0",
     "prettier": "^2.5.1",
-    "react-devtools": "^4.19.2",
```
farisaziz12, line 121: **praise:** thanks for removing this! Definitely not needed 😄

extensions/stripe/src/fill-checkout.tsx
```diff
@@ -0,0 +1,108 @@
...
+    data: favoriteCards,
+    visitItem,
+    resetRanking,
+  } = useFrecencySorting<CardType>(cards, {
+    key: (card) => createCardKey(card),
+  });
+
+  return (
+    <List>
+      <List.Section key={"favorites"} title={"Favorites"}>
```
farisaziz12, line 31: **nitpick**

```suggestion
      <List.Section key="favorites" title="Favorites">
```

extensions/stripe/src/utils/cards.ts
```diff
@@ -0,0 +1,42 @@
...
+  const now = new Date();
+
+  const month = now.getMonth();
+  const year = String(now.getFullYear() + 1).slice(-2);
+
+  const expiryDate = `${month}/${year}`;
+  return expiryDate;
+};
+
+const createAppleScript = (card: Card) => {
```
farisaziz12, line 18: **thought:**


This works great while Raycast is consistently used in the MacOS ecosystem, this might be something we need to revisit down the line when [Raycast comes to Windows](https://www.raycast.com/windows). We can go with this approach for now.

farisaziz12 (approved):
Thanks for this awesome addition @bradgarropy! Really cool to see someone from the Stripe team introduce new functionality to the Raycast extension. 

I haven't touched the extension in a while, so I'm not sure where the type errors have originated from but I'm entirely happy for you to fix them or I can make a commit to your branch with the required changes to get the CI passing. Whatever is easiest 😄 

Other than that all looks good to me, I just left a few non-blocking comments as food for thought.

farisaziz12:
Thoughts on a UX improvement if you would like to go the extra mile 😉

Currently, with each command of the extension, you can individually set the environment to `test` or `live`. If I set `test` as the env for the first command that I use, upon opening a second command to use the environment I previously selected does not carry over (e.g. first command I set to `test`, second command opens with env as `live`).

From what I remember, at the time of this extension's creation, we didn't have APIs to work around this as commands are their own individual views rather than being routed from a parent component, which means they could not be wrapped in a React context globally for example. We could use the newer [`Cache`](https://developers.raycast.com/api-reference/cache) API to introduce saved environments across all commands since cache is shared across all commands. Happy to do this myself in a future PR but noting down the thought in this PR as it came up during my testing of the new functionality

extensions/stripe/src/utils/cards.ts
bradgarropy: We should absolutely consider this. I would assume that in the future Raycast will implement a similar utility function like [runAppleScript](https://developers.raycast.com/utilities/functions/runapplescript) on Windows.

bradgarropy:
I'm happy to implement this! But in a different pull request if you're cool with it. This one is getting a bit chunky.

> Thoughts on a UX improvement if you would like to go the extra mile 😉
> 
> Currently, with each command of the extension, you can individually set the environment to `test` or `live`. If I set `test` as the env for the first command that I use, upon opening a second command to use the environment I previously selected does not carry over (e.g. first command I set to `test`, second command opens with env as `live`).
> 
> From what I remember, at the time of this extension's creation, we didn't have APIs to work around this as commands are their own individual views rather than being routed from a parent component, which means they could not be wrapped in a React context globally for example. We could use the newer [`Cache`](https://developers.raycast.com/api-reference/cache) API to introduce saved environments across all commands since cache is shared across all commands. Happy to do this myself in a future PR but noting down the thought in this PR as it came up during my testing of the new functionality

bradgarropy:
Thanks for doing this!

Before merging in these changes I'd like to do a bit of investigation. I feel like these types should be included in one of our `npm` packages. Let me take a look...

> @bradgarropy I went ahead and adjusted the code to fix the TS issues and ran the build locally on my end to validate all is good. I don't believe I have write access to your fork so feel free to copy the below code for the respective files to fix the issue:

bradgarropy:
After rethinking it, the Raycast team will probably take a bit to review this PR, so let's get as much in as we can.

I merged your changes to fix the types. Still investigating moving over to the `stripe` package rather than defining custom ones.

farisaziz12:
> After rethinking it, the Raycast team will probably take a bit to review this PR, so let's get as much in as we can.
> 
> I merged your changes to fix the types. Still investigating moving over to the `stripe` package rather than defining custom ones.

I definitely agree with the move to the stripe built in types! This would be a great simplification, let me know how I can support. We shouldn't have done the custom types to begin with, i'm unsure the reasoning for it back when this was done

farisaziz12:
> I'm happy to implement this! But in a different pull request if you're cool with it. This one is getting a bit chunky.
> 
> > Thoughts on a UX improvement if you would like to go the extra mile 😉
> > Currently, with each command of the extension, you can individually set the environment to `test` or `live`. If I set `test` as the env for the first command that I use, upon opening a second command to use the environment I previously selected does not carry over (e.g. first command I set to `test`, second command opens with env as `live`).
> > From what I remember, at the time of this extension's creation, we didn't have APIs to work around this as commands are their own individual views rather than being routed from a parent component, which means they could not be wrapped in a React context globally for example. We could use the newer [`Cache`](https://developers.raycast.com/api-reference/cache) API to introduce saved environments across all commands since cache is shared across all commands. Happy to do this myself in a future PR but noting down the thought in this PR as it came up during my testing of the new functionality

This is definitely out of scope! Thought i'd note it down just in case. Thanks!

bradgarropy:
Made good progress moving to `stripe` types. There are only a few problems remaining in the `connected-accounts.tsx` file. The `Stripe.Account` shape doesn't match exactly what was present in the custom type.

@farisaziz12 mind taking a look?

pernielsentikaer:
Do you need help with anything?

bradgarropy:
Don't need help @pernielsentikaer, just need the original authors (@pradeepb28 / @farisaziz12) to reconcile the type differences between what Stripe actually returns from its API, and what the application had expected.

pradeepb28:
LGTM

pernielsentikaer:
Makes sense, I'll await 🙂

bradgarropy:
Calling in @farisaziz12 again to take a look at integrating the `Stripe.Account` type in the `connected-accounts.tsx` file. I don't want this PR to go stale!

farisaziz12:
> Calling in @farisaziz12 again to take a look at integrating the `Stripe.Account` type in the `connected-accounts.tsx` file. I don't want this PR to go stale!

Thank you for the reminder! Apologies for the delayed response as I have been travelling for conferences. I had a look at this code, i'll clean it up in the future but to quickly adjust the types and get it working you can update the resolver and return type to this:

For context, the issue here is that some of the fields have changed since this was first integrated so the mapper/resolver simply had to be aligned to the new connected account structure and once I tested it all was good again. @bradgarropy let me know if you need support with anything else!

```ts
type ConnectedAccount = {
  id: string;
  created: string;
  first_name: string;
  last_name: string;
  email: string;
  capabilities: string;
  default_currency: string;
  company_address: string;
  dob: string;
};

const resolveConnectedAccount = (connectedAccount: Stripe.Account): ConnectedAccount => {
  const { city, country, line1, postal_code, state } = connectedAccount.company?.address ?? {};
  const companyAddress = [line1, city, state, postal_code, country].filter(Boolean).join(", ");

  const resolvedConnectedAccount: ConnectedAccount = {
    ...connectedAccount,
    default_currency: connectedAccount.default_currency?.toUpperCase() ?? "",
    created: connectedAccount.created ? convertTimestampToDate(connectedAccount.created) : "",
    dob: createDateOfBirth(connectedAccount),
    company_address: companyAddress,
    capabilities: Object.keys(connectedAccount.capabilities ?? {}).join(", "),
    first_name: titleCase(connectedAccount.individual?.first_name ?? ""),
    last_name: titleCase(connectedAccount.individual?.last_name ?? ""),
    email: connectedAccount.email ?? "",
  };

  return resolvedConnectedAccount;
};

```

bradgarropy:
Okay @farisaziz12 / @pradeepb28, all the types line up now!

@pernielsentikaer we're ready to roll!

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

bradgarropy:
@raycastbot Got everything linked up!

bradgarropy:
@raycastbot @pernielsentikaer Got my GitHub account all linked up. Would love to get that swag code! 🙏🏼

pernielsentikaer:
It happens automatically; credit is added, but we don't have a way to use it yet (but we will)

PR #15292, opened 2024-11-09
Add miti-raycast extension

Parajulibkrm:
## Description
It adds two commands
- Menu Bar to Show Current Bikram Sambat Calendar (Nepal uses a different calendar system 57 years ahead of AD calendar) 
- Converter command to convert between AD to BS and vice versa

## Screencast

image
image
image

## Checklist

### DIFF
--- a/extensions/miti-raycast/src/convert-date.tsx
+++ b/extensions/miti-raycast/src/convert-date.tsx
@@ -151,7 +151,7 @@ function getResults(query: string): LabeledDate[] {
     }
   }
 
-  let results: LabeledDate[] = [];
+  const results: LabeledDate[] = [];
 
   // Try parsing as Nepali Date
   const nepaliDate = parseNepaliDate(query, now);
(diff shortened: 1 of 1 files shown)


pernielsentikaer:
I don't know anything about the subject, so sorry if my answer here is all wrong 🙂

Is this related to this: https://www.raycast.com/amin_saedi/jalali-date-convertor

Parajulibkrm:
@pernielsentikaer Not at all, it's an entirely different calendar system.

Parajulibkrm:
https://www.raycast.com/beedhan/nepali-date-converter
This is the closest one available in Raycast but this one takes a completely different approach like Menu bar

Parajulibkrm:
@pernielsentikaer  any update on this?

pernielsentikaer:
What about enhancing the current setup? We all want the best possible extensions. We can ask @Beedhan what he thinks :)

Parajulibkrm:
Still think it's best published as a separate one, the features are entirely different and the approach is different. Also this one consumes some endpoints from https://github.com/PoskOfficial/Miti project and contains some Miti Branding? Not sure if replacing the entire code, functionality and adding Miti branding would do justice to @Beedhan's project?

Beedhan:
I developed the Nepali Date Converter because I couldn’t find any existing extension. If an instant (no-view mode) date converter, like mine, I no longer see the need for my version. In that case, I think it’s better to remove my extension now, as Miti offers much more.

Parajulibkrm:
@pernielsentikaer can we have this merged?

pernielsentikaer:
What about instead of removing your extension we just override with this extension so all current users will get the new version?

So basicly you need to override the files in the current extension, does that make sense @Parajulibkrm

Parajulibkrm:
@pernielsentikaer I still don't see the point in that, I'm not sure why that would be a good idea. Is there no other way to this can go through?

PR #15302, opened 2024-11-10
Add window-layouts extension

teemusuvinen:
## Description

An extension to arrange open windows on a desktop into predefined layouts.
This extensions uses Raycast's WindowManagement API without external dependencies.
It is not an automatic window tiling manager, but provides a quick way to organize your currently open windows on a desktop.

In extension preferences the user can choose a gap that's used around and between the windows, hide a toast "Windows arranged" or to keep the Raycast window open after layout has been applied.

Why not just use the window layout command or window management commands that come out of the box with Raycast?
With this extension:
1. There is no need to define positions for each app separately
2. You can choose which window gets the biggest area (when applicable) just by focusing that window and running the layout command
3. There is no need to define positions for each desktop / workspace separately
4. You can set gaps between and around windows

## Screencast

https://github.com/user-attachments/assets/e8bcd93d-ac22-4188-a292-9c3588894a55

## Checklist


pernielsentikaer:
Approved ✅

PR #15322, opened 2024-11-11
Add zoxide-git-projects extension

theherk:
## Description

Allows the user to select git repositories found by filtering a zoxide query, then open them in an application of choice.

I use and love [zoxide](https://github.com/ajeetdsouza/zoxide) ❤️. With this extension, all the directories returned by `zoxide query -l` (i.e. the directories you use) are filtered to just those that are git repositories and ordered by most frequently used. From there, this extension allows you to fuzzy find a specific project, then open that project with the tool of your choice. For me, [zed](https://zed.dev/).

## Screencast

https://github.com/user-attachments/assets/664b15a6-ff19-4528-9405-1a5e45fc1ac6

## Checklist


theherk:
I'm a little confused about the "I checked that assets used by the README are placed outside of the metadata folder" task. I'd like to use this same image in the README.md and it works fine. If you want me to store a second copy of it, I can, but I'm just unclear.

pernielsentikaer (changes requested):
What about using [useCachedPromise](https://developers.raycast.com/utilities/react-hooks/usecachedpromise) instead and skip the custom implementation

theherk:
@pernielsentikaer: The simple answer is probably ineptitude :). My goal was to have control over the life of the cache, and set explicitly how long it will live since the call can have some overhead. Can that be done with this util method. Maybe be having some timer to call revalidate? I'm unsure. But the point is to call the query only when needed, but let the user change it to force an update if repositories change.

theherk:
One option would be to always make the query but immediately return and work from the cached list until the query returned. Then update the cache if the query response differs from the cached dataset.

theherk:
Also, your update to "Git Application" doesn't make sense to me. You can use many applications to open git repositories. Imagine VS Code or Finder. Those aren't exactly "Git" applications.

theherk:
Actually, turns out the method I describe above is precisely how `useCachedPromise` works. See... ineptitude. :D

So, this is a much more simple implementation. I'm still not sure about the change to Git Application, but I'll await feedback.

In the meantime, I also modified the filter to be a bit more friendly. Frankly, I'd be open to using the build in sort, if you prefer the simplicity, but not being able to find "my-repo" with "my repo" or "repo my" is not great.

pernielsentikaer:
I think it's fine as is. Let's get the first version live so we can iterate on it - thanks for looking into useCachedPromise 🙂

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #15324, opened 2024-11-12
Add raycast-jq extension

vladimir-kotikov:
## Description

This adds the extension to query JSON files or clipboard content with [jq](https://jqlang.github.io/jq/).

## Screencast

<details><summary>Hidden due to large size</summary>

jq_cast_1
jq_cast_2
jq_cast_3

</details> 

## Checklist


pernielsentikaer:
There is a crash when you open the extension and leave the file picker open for about 10 seconds. Could you look into that before I dive deeper?

raycast-jq 2024-11-18 at 13 27 06

pernielsentikaer:
The issue with timeout should be fixed in the latest version of @raycast/utils (1.18.1) - could you try that?

vladimir-kotikov:
> The issue with timeout should be fixed in the latest version of @raycast/utils (1.18.1) - could you try that?

Yep, it works now. Updated the code accordingly

pernielsentikaer (changes requested):
Would it be better to let the user install jq with brew (or link to the installation page) instead of just installing it, I don't think this way will work for intel machines, right?

vladimir-kotikov:
I've actually initially implemented the logic to determine system arch and download the appropriate binary, but it seems it has been lost during my experiments, so I think I can reimplement it back so it works for any processor architectures.

As for the option to use system jq - that would be nice (and I was considering adding an option to choose between system and downloaded jq) but it appears that the extension not getting the system PATH variable so spawning system jq would require asking user to provide full path to it, which I feel is not very friendly (or guessing it, which is error-prone). 

Also, the current [documentation suggests](https://developers.raycast.com/basics/prepare-an-extension-for-store#binary-dependencies-and-additional-configuration) to

> avoid asking users to perform additional downloads and try to automate as much as possible from the extension

so I was trying to follow that.

What do you think?

vladimir-kotikov:
Done!

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #15393, opened 2024-11-15
Update clipboard.md

luarmr:
## Description

On the clipboard documentation for copy mention that we can use content the type content as parameter, this type do not have html as a possibility in the doc. 

I was able to do:
```
 Clipboard.copy({ html: message })
```

So I went to the definition of this type:
```
    export type Content = {
        /**
         * The text representation of the content.
         */
        text: string;
    } | {
        /**
         * The file representation of the content.
         */
        file: PathLike;
    } | {
        /**
         * The HTML representation of the content.
         */
        html: string;
        /**
         * The alternative text representation of the content.
         */
        text?: string;
    };
```

I assume is good idea to add it here

https://developers.raycast.com/api-reference/clipboard#clipboard.content

## Screencast

[](https://developers.raycast.com/api-reference/clipboard#clipboard.content)

## Checklist

### DIFF
--- a/docs/api-reference/clipboard.md
+++ b/docs/api-reference/clipboard.md
@@ -177,6 +177,10 @@ type Content =
     }
   | {
       file: PathLike;
+    }
+  | {
+      html: string;
+      text?: string; // The alternative text representation of the content.
     };
 ```
 


pernielsentikaer:
Thanks for the fix 🙂

PR #15438, opened 2024-11-19
Add `Doppler` extension - View Projects with their Configs, Environments and Secrets.

xmok:
## Description

❗[Doppler Share Secrets](https://www.raycast.com/thomas/doppler-share-secrets) exists in the Store but it deals only with sharing secrets so I think it makes sense for this one to be a separate extension.

Our use case is [Doppler](https://doppler.com/) syncing secrets with [Laravel Forge](https://forge.laravel.com/) and [DigitalOcean](https://digitalocean.com/) for an e-commerce platform past ~4 years.

This first pass of the extension makes use of Doppler's [Node SDK](https://github.com/DopplerHQ/node-sdk) and a PAT to:

- View Projects, Environments, Configs, Secrets AND
- Copy Secrets and Update Secret Note

## Screencast

Unfortunately, I can not share the real Doppler account so I am using a temp account to record the screencast:

https://github.com/user-attachments/assets/9dd005a1-4867-4e73-b97a-74b7c85d3ed4

## Checklist


pernielsentikaer:
Do you think it would be a bad idea to merge into one?

xmok:
Not at all! I didn't want to complicate the existing extension but I would 100% be down if the extensions are combined as it's one less extension to install 😅. Would it be easier to edit this PR or will it be easier for the bot if I make a new PR modifying old extension directly?

pernielsentikaer:
What do you think, @thomaspaulmann? Don't you think we should enhance your extension, or would you prefer to keep it as close to the example as possible?

pernielsentikaer:
I think we should do it 🙂

PR #15443, opened 2024-11-19
Remove duplicate bookmarks from search results.

### LINKED ISSUE #15442: [Zen Browser] Bookmarks Duplicated for tags
### Extension

https://www.raycast.com/Keyruu/zen-browser

### Raycast Version

1.85.2

### macOS Version

15.1

### Description

Sorry @Keyruu, but something is still awry with the plugin, albeit minor. In the initial list for Search Bookmarks, each bookmark shows once with title and link (correct) but then an additional time with just link for each tag a bookmark has. Once filtering it is just fine, but it makes the initial list a bit haphazard for those of us that tag bookmarks.

### Steps To Reproduce

1. Edit a bookmark.
2. Add the tags `this, is, just, a, test`.
3. Witness all the additional entries in the Search Bookmarks list.

### Current Behaviour

As described in the description.

### Expected Behaviour

One entry per bookmark no matter how many tags.

theherk:
The bookmark search was showing duplicate entries when a URL was bookmarked
multiple times. This was caused by the query returning all bookmark entries
for a URL, including tag-related entries.

The fix uses a window function (ROW_NUMBER) to partition bookmarks by their
foreign key (URL reference) and selects only the entry with the lowest ID
for each URL. This ensures each bookmarked URL appears only once in the results while preserving all bookmark metadata including optional titles.

fixes #15442

## Checklist

### DIFF
--- a/extensions/zen-browser/CHANGELOG.md
+++ b/extensions/zen-browser/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Zen Changelog
 
-## [Fix Bookmark Duplicates Issue] - {PR_MERGE_DATE}
+## [Fix Bookmark Duplicates Issue] - 2024-11-21
 
 - Fix duplicate entries in bookmark search results. Previously, bookmarks would show up multiple times if they had tags. Now each bookmark appears only once regardless of how many tags it has.
 


theherk:
I was also getting some db locked and read issues periodically, so I added some retry logic.

Keyruu:
Awesome stuff man! Thanks for the contribution.

extensions/zen-browser/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Zen Changelog
 
+## [Fix Bookmark Duplicates Issue] - 2024-11-19
```
Keyruu, line 3: This shouldn't include an actual date. This should have this placeholder `{PR_MERGE_DATE}`. This will be replaced later when the PR gets merged. https://developers.raycast.com/basics/prepare-an-extension-for-store#version-history

theherk: My bad. I forgot about that.

PR #15462, opened 2024-11-21
fix: remove process.env from pass execution

camiloaromero23:
## Description

- Fix: Remove process.env from pass execution

This change is done due to errors in the extension using nix-darwing messing with the locales

## Screencast

Screenshot 2024-11-21 at 00 31 42

## Checklist

### DIFF
--- a/extensions/pass/CHANGELOG.md
+++ b/extensions/pass/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## [1.2.0] - 2024-11-21
 
-- Fix: Remove process.env from pass execution.
+- Fix: Delete LC_ALL env variable messing with nix-darwin.
 
 ## [1.1.0] - 2024-06-16
 
--- a/extensions/pass/src/pass.ts
+++ b/extensions/pass/src/pass.ts
@@ -15,11 +15,14 @@ async function envPath(): Promise<string> {
 
 export async function pass(cmd: string, storeDir: string = ''): Promise<string> {
   const passCmd = `pass ${cmd}`;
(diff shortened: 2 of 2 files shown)


pernielsentikaer:
Could you check this @capipo

capipo:
I'm passing `process.env` because I need to support the `PASSWORD_STORE_DIR` environment variable, and potentially other variables supported by the `pass` command.

Perhaps if you filter out variables that start with `PASSWORD_STORE` we could continue to support those variables and your problem would be solved.

capipo:
Nevermind, I see that that variable is passed later

capipo (approved):

camiloaromero23:
Hi @capipo,

I'm not really sure about that since the extension is overwriting the `PASSWORD_STORE_DIR` env variable. So either way, it would not use the env variable from `process.env`

```typescript
export async function pass(cmd: string, storeDir: string = ''): Promise<string> {
  const passCmd = `pass ${cmd}`;
  const { stdout, stderr } = await execAsync(passCmd, {
    timeout: 10000,
    env: {
     ...process.env
      PATH: await envPath(),
      PASSWORD_STORE_DIR: storeDir,
    },
  });

  if (stderr) {
    throw new Error(stderr);
  }
  return stdout;
}
```

camiloaromero23:
@capipo pls check the above solution 👆🏽

camiloaromero23:
@pernielsentikaer I think this can be merged

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #15469, opened 2024-11-21
Update jira extension

Silv-1:
## Checklist

### DIFF
--- a/extensions/jira/CHANGELOG.md
+++ b/extensions/jira/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Jira Changelog
 
-## [Added Parent Issue support] - 2024-11-29
+## [Added Parent Issue support] - 2024-12-02
 
 - Implemented the ability to open the parent issue from a child issue.
 


extensions/jira/src/components/IssueActions.tsx
```diff
@@ -169,6 +169,15 @@ export default function IssueActions({
       </ActionPanel.Section>
 
       <ActionPanel.Section>
+        {issue.fields.parent && (
+          <Action.Push
+            target={<IssueDetail initialIssue={issue.fields.parent} issueKey={issue.fields.parent?.key} />}
+            title="Open Parent Issue"
+            icon={Icon.ChevronUp}
+            shortcut={{ modifiers: ["cmd", "shift"], key: "n" }}
```
thomaslombart, line 177: This shortcut is already used to add a new comment. Could you use another one? 🙂

Silv-1: Of course. Which one do you think would be suitable?

Silv-1: What about using the same shortcut as for `Open Child Issues`? I assume the actions should not be displayed at the same time as each issue either is a subtask and has a parent or is an issue and has subtasks.

Silv-1: @thomaslombart The shortcut of the new command `Open Parent Issue` is now set to `cmd` + `shift` + `o` . It's the same shortcut as for the action `Open Child Issues`.

Silv-1: It seems issues can have children and a parent at the same time. I changed the shortcut to `cmd` + `shift` + `u`.

Issue #15130, opened 2024-10-28
[Browser Bookmarks] Support for Zen Browser


theherk:
### Extension

https://www.raycast.com/raycast/browser-bookmarks

### Description

Should probably support [Zen Browser](https://zen-browser.app/). I don't think it will be that difficult to mimic the behavior of the forked Firefox implementation.

### Who will benefit from this feature?

Users of the Zen Browser.

thomaslombart:
Thanks for the feature request. Would you like to work on it? If yes, here's our [guide on extension contributions](https://developers.raycast.com/basics/contribute-to-an-extension) so that you can open up a pull request. Feel free to contact me here or in the [Slack community](https://raycast.com/community) if you need further help.

theherk:
I may. The zen extension is currently working well, but for posterity I might work on this too.

nicoandmee:
I would also like this.

PR #15495, opened 2024-11-23
Add support for Zen Browser Desktop bookmarks.

theherk:
## Description

This is an amalgam of the Firefox implementation in the same extension and the seach method used in the Zen Browser extension.

Fix: #15130

## Checklist


pernielsentikaer (changes requested):
I gave this a try, but I can't see any bookmarks from Zen after enabling it? I also tried restarting Zen and Raycast without any changes 🫠

image

theherk:
Interesting. Assuming your have Zen selected in the Select Browser interface, have you checked your profile folder in about:profiles. If it isn't the alpha suffix, we may have to work on the select profiles logic.

theherk:
Screenshot 2024-11-25 at 13 16 31

I see. For me, just the one. So, I'm not sure the best way to handle that. But when I'm no longer at work, I can try some things. Maybe we need a select other profile option. So even if it finds the alpha profile, one can select another if the other is housing the bookmarks. Open to suggestions.

theherk:
@pernielsentikaer: This should be fixed now. Give it another whirl. The path was being constructed badly. I even added some test multi-profile setups and the profile switcher works fine now.

pernielsentikaer:
That was better 🙂

pernielsentikaer (approved):

PR #15510, opened 2024-11-24
Update `Wireguard` extension - `showToast` then `showHud` for better UX

xmok:
## Description

Simple but useful addition because it is not evident something has happened:

- `showToast` then `showHud` on **toggle**
- `showToast` then `showHud` on **disconnectAll**

Previously, extension would exit w/o feedback but now the `Hud` lets us know what has happened.

## Screencast

https://github.com/user-attachments/assets/dfdac62d-b16b-4424-b726-dabdb6fc316f

## Checklist

### DIFF
--- a/extensions/wireguard/CHANGELOG.md
+++ b/extensions/wireguard/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Wireguard Changelog
 
-## [Add Toasts and Huds] - {PR_MERGE_DATE}
+## [Add Toasts and Huds] - 2024-12-03
 
 - `showToast` then `showHud` on **toggle**
 - `showToast` then `showHud` on **disconnectAll**


sbugzu:
Hi @xmok  Thanks for the suggestion.
Now the idea of the extension is to show the status by menubar, it will make the operation faster. 

wireguard-status
you can consider the following two ways:
1. Using wireguard client
2. Using VPN status, `Setting -> Control Center -> VPN` set to `Show in Menu Bar`

Hope it helps.

xmok:
The menu bar items are valid but it might not work for everyone e.g. in my own case:
1. I don't have VPN Status enabled in **Menu Bar**
2. When I use the extension to toggle VPN, Wireguard isn't running for me by default (I don't like to run the GUI/MenuBar unless needed) so it doesn't show up in the Menu Bar either.

If you prefer, we could hide the `Hud` and `Toast` behind a `Preference` so users can choose what they want but at least in my own case, they improve UX.

pernielsentikaer:
Could you check the response @sbugzu

sbugzu (approved):

sbugzu (approved):

PR #15567, opened 2024-11-27
Add Rebound extension

gohadar:
## Description

The `Rebound` extension is a keyboard-first HTTP request manager, just like a glorified cURL client, at your fingertips.

In this version there is a new command, `Rebound`, which opens the rebound manager to let the user start creating and working with their requests.

## Screencast

All Rebounds
Create Rebound
Rebound View
Response View

## Checklist

### DIFF
--- a/extensions/rebound/.gitignore
+++ b/extensions/rebound/.gitignore
@@ -34,6 +34,14 @@ yarn-error.log*
 *.pem
 .pnpm-debug.log*
 
+.DS_Store
+
+# Raycast specific files
+raycast-env.d.ts
+.raycast-swift-build
+.swiftpm
+compiled_raycast_swift
+
 # nextjs
 /.next/
 /out/


pernielsentikaer:
Hi 👋

Did you see [this extension](https://www.raycast.com/jonathangiardino/curl) before creating this? It feels like it's doing the same job - it would be better to enhance the current one instead of creating a similar extension.

gohadar:
> Hi 👋
> 
> 
> Did you see [this extension](https://www.raycast.com/jonathangiardino/curl) before creating this? It feels like it's doing the same job - it would be better to enhance the current one instead of creating a similar extension.

Hi @pernielsentikaer, thanks for your input.
I did see this extension but I feel like I tackled the problem from a different direction, while the cURL extension provides a way to send HTTP requests directly from Raycast, my extension has the intention of being a full-fledged HTTP request manager much like [postman](https://www.postman.com) with capabilities like defining custom request groups, variables and managing the history of a requests. While I see the similarities between the two extensions, I do feel like this is a departure from the concept of the cURL command and the existing extension that justifies it being a whole new extension. That being said, I'd love to hear your take on the issue as I have many ideas and plans for lots of features for this extension to make it much more than the current cURL extension.

pernielsentikaer:
@jonathangiardino what do you think?

jonathangiardino:
> @jonathangiardino what do you think?

Isn't it exactly the same as cURL? @pernielsentikaer? I didn't look at all the commands but looks like it does the same thing right?

gohadar:
> > @jonathangiardino what do you think?
> 
> Isn't it exactly the same as cURL? @pernielsentikaer? I didn't look at all the commands but looks like it does the same thing right?

It achieves the same goal - allowing you send HTTP requests - but it approaches this task differently, while `cURL` is about letting you send the requests and get the response and that's it, my extension is more about "replacing" common tools like [postman](https://www.postman.com/) by treating each request as a separate entity that has multiple results in order to show history. This behavior is more then just acting as a `cURL` command with GUI and is meant to provide the same capabilities as the `cURL` extension but with more capabilities to allow for easier flow while working and testing an API. 

The reason that I created a different extension is because the whole approach to managing the requests is different so if I'll be changing the `cURL` extension I'd be replacing most if not all of the existing code which seemed to me like it was too much change to an existing extension. If you think that it's better I can "migrate" all this code to the existing `cURL` extension and "override" most of it.

jonathangiardino:
The cURL extension also has a command for listing your requests history, in addition to doing the requests and getting the responses @gohadar 

But anyway, for me it's not a problem if you want to have your own extension with your own approach. I am not a reviewer for Raycast, I was tagged by @pernielsentikaer probably because I am the cURL author and first contributor I guess and looking at Rebound, it seems to do the same as my extension.

It's not up to me anyway to review and merge 👍🏻

gohadar:
@jonathangiardino I saw the history command but while trying to use it it felt cumbersome to have two different commands for working with my requests, and not having the ability to view the history of responses for a specific URL was a problem for my use.

Anyways, @pernielsentikaer please let me know what you feel like is the best course of action, should I continue with this extension or should I move all the code to the `cURL` extension?

jonathangiardino:
@gohadar ok then it's different.

If you find it cumbersome is not the same as it's a different extension isn't it? It's been used by thousands of users so far and having two different commands did not seem to bother anyone.

So I guess the options are:

- Simply publishing your own extension and for that @pernielsentikaer will have to approve and merge. That's totally fine with me, is not a for profit competition for me :)
- Open a PR for cURL with improvements you'd like to see

Have a great week!

gohadar:
That sounds good to me, thanks for your input @jonathangiardino and have a great week too 😄.

I'll wait and see what @pernielsentikaer says and take it from there.

pernielsentikaer:
I guess we could also add an action in the current version that launches the history; this way, it could work from both places. It would be nice to enhance rather than adding something similar 🙂

You can also see our [extension guidelines](https://manual.raycast.com/extensions) section about `Reasons to be rejected`

PR #15680, opened 2024-12-03
Add unix-time-converter extension

rpandey1234:
## Description

The goal of this extension is to make it easier to translate between UNIX time and what that actually means in terms of a human date. Given an input, this command tries to smartly detect a UNIX (epoch) time or human time and convert it to the other. 

As a JavaScript/TypeScript developer, we often work with time in milliseconds, not seconds, so this extension also tries to detect the appropriate input unit.

## Screencast

https://github.com/user-attachments/assets/aadb23fd-fedb-4954-9ce2-d54ed8884904

## Checklist

### DIFF
--- a/extensions/unix-time-converter/package.json
+++ b/extensions/unix-time-converter/package.json
@@ -12,7 +12,7 @@
   "commands": [
     {
       "name": "convert",
-      "title": "Auto-convert Unix <> Human Time",
+      "title": "Auto-Convert Unix <> Human Time",
       "description": "Making it easier for humans and computers to understand each other. Given an input, this command tries to smartly detect a unix (epoch) time or human time and convert it to the other.",
       "mode": "view"
     }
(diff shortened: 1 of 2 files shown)


pernielsentikaer:
Hi 👋

This should also be working directly in Raycast, could you try that?

Raycast 2024-12-12 at 14 07 16

rpandey1234:
Nice @pernielsentikaer, I had no idea this was built into Raycast! (I saw another extension `Unix Timestamp` which I was trying to improve upon.)

The benefit of this extension is: 
- supporting milliseconds instead of only seconds
- the additional info shown with the screen real estate: relative time

Happy to close this out if this doesn't offer enough functionality. As a JS dev who works in milliseconds, I'd personally use this.

pernielsentikaer:
It should also work the other way `1737802800 to date` - I guess another extension doing this might be a bit too much

rpandey1234:
Thanks for the review, will close this out then.

PR #15701, opened 2024-12-05
feat: add new extension jue-jin

AkaShark:
## Description

Add a new extension that can read recommend article from [juejin](https://juejin.cn/)

## Screencast

CleanShot 2024-12-05 at 10 40 11@2x

CleanShot 2024-12-05 at 10 40 19@2x

## Checklist


pernielsentikaer (changes requested):

AkaShark:
Thank you for your help. I modified the localization string used in the plugin. Still, some strings in the screenshot above are in Chinese, mainly because the source of the [juejin](https://juejin.cn/) website is a Chinese technical community.

extensions/juejin/package.json
```diff
@@ -0,0 +1,41 @@
+{
+  "$schema": "https://www.raycast.com/schemas/extension.json",
+  "name": "jue-jin",
```
LitoMore, line 3: I think we can use its Chinese name `掘金 juejin.cn` since your extension is Chinese only.

```suggestion
  "name": "掘金 juenjin.cn",
```

We can append its website to the name so users can search by its Pinyin name.

pernielsentikaer: We can't have Chinese in the name according to the manual; however, we can allow it in `keywords: []` instead

extensions/juejin/README.md
```diff
@@ -0,0 +1,3 @@
+# 掘金
+
+展示掘金各个分类下的推荐文章
```
pernielsentikaer, line 3: Same here, if you really wants it, then you can make a README-cn.md or something like that

extensions/juejin/package.json
LitoMore: Ok, then let's use `Juejin` as its name here. Dash is not needed.

pernielsentikaer: This is only the `name` field, which is not something the user will see anyway. They will only see the title 🙂

LitoMore: Ah, then the lowercase `juejin` would be better.

extensions/juejin/package.json
```diff
@@ -0,0 +1,41 @@
...
+  "title": "juejin",
```
LitoMore, line 4:
```suggestion
  "title": "Juejin",
```

Let's use the title case `Juejin` for its display name.

extensions/juejin/package.json
AkaShark: Okay, I've modified the `title` field to `Juejin` and added the keywords `Juejin - juejin.cn` to the `keywords` field. @pernielsentikaer Can I add `掘金` the Chinese name for this extension to `keywords`?

extensions/juejin/package.json
AkaShark: okay, I also think this is better.

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

AkaShark:
Thanks for all the help! This is my first time publishing a Raycast extension, and I'm glad I was able to create a simple one.

PR #15705, opened 2024-12-05
Update claude extension

### LINKED ISSUE #15695: [Claude] Feature Request: Copy Question
### Extension

https://www.raycast.com/florisdobber/claude

### Description

Hey, sometimes it happens that I get no response from Claude (Ask Question feature). Lets say because of the internet connection or I am running out of credits ...
In those cases I would love to copy the question I asked Claude to paste it to another service.
Can you please add a feature which copies the selected question?

The user :) This would also be nice to see the complete question in history, which become truncated.

### Anything else?
Here I ask something without internet connection (see the error):
image

Now I have no chance to copy the question:
image

Nice extension!

ridemountainpig:
## Description
Added copy question actions.
Closes #15695

## Screencast
claude 2024-12-05 at 14 43 31

## Checklist

### DIFF
--- a/extensions/claude/CHANGELOG.md
+++ b/extensions/claude/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Claude Changelog
 
-## [Update] - {PR_MERGE_DATE}
+## [Update] - 2024-12-09
 
 - Feature: Added copy question actions
 


alexander-neumann-webdesign:
The recently introduced “Copy Question” functionality would benefit from additional customization options. Currently, its default binding to <kbd>command</kbd>+<kbd>C</kbd> conflicts with the system-wide copy function, making it difficult to selectively copy portions of the output text.

I think a better approach would be to implement an alternative default shortcut, such as <kbd>Shift</kbd>+<kbd>Command</kbd>+<kbd>C</kbd>, while also allowing users to define their own custom keyboard combinations or disable the “Copy Question” action entirely in the extension settings panel.

ridjis:
As Alexander wrote, this PR introduced regression where we can't CMD+C selection from Claude's answer.

This doesn't looks like proper solution to the problem as this extension already had "Copy Question" action.

image

PR #15772, opened 2024-12-09
Add water-reminder extension

alejandrofalcar:
## Checklist

### DIFF
--- a/extensions/water-reminder/package-lock.json
+++ b/extensions/water-reminder/package-lock.json
@@ -19,74 +19,8 @@
         "typescript": "^5.4.5"
       }
     },
-    "node_modules/@esbuild/aix-ppc64": {
-      "version": "0.24.0",
-      "resolved": "https://inditex.jfrog.io/artifactory/api/npm/node-public/@esbuild/aix-ppc64/-/aix-ppc64-0.24.0.tgz",
-      "integrity": "sha512-WtKdFM7ls47zkKHFVzMz8opM7LkcsIp9amDUBIAWirg70RM71WRSjdILPsY5Uv1D42ZpUfaPILDlfactHgsRkw==",
-      "cpu": [
-        "ppc64"
-      ],
-      "license": "MIT",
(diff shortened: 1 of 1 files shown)


LitoMore:
There is already a water reminder extension in the store: https://www.raycast.com/FezVrasta/acqua.

pernielsentikaer:
As @LitoMore mentions, we already have that. Did you search for it before starting yours? If yes, and it didn't appear, what did you search for so we can enhance the keywords for better visibility? 😊

alejandrofalcar:
> As @LitoMore mentions, we already have that. Did you search for it before starting yours? If yes, and it didn't appear, what did you search for so we can enhance the keywords for better visibility? 😊
> There is already a water reminder extension in the store: https://www.raycast.com/FezVrasta/acqua.

Of course I know Acqua, I even used it, but it did not satisfy me for what I consider normal use cases. I have looked into contributing, but I think their approach is very different from mine.

In Water Reminder, you can set how much you will drink per day, as well as how much you will drink at each sip, including a macOS notification to inform you to drink when the time is up, which you can turn off if you wish.

The menu will show you the remaining amount you have left, as well as any configuration information the user has added.

In the future I am looking to add mobile settings and other features to improve the extension.

pernielsentikaer:
I'm sure @FezVrasta is open to improvements too. Could that be a possibility? Otherwise, we'll end up with two almost identical extensions in the store 😊

FezVrasta:
I'm okay to bring down my extension if this one is better.

pernielsentikaer:
I don't think it's about being the best, but having the best possible extension in the store. Enhancements should be the way rather than just replacing if possible. I guess it depends on who you ask about what's best 🙂

LitoMore:
> I don't think it's about being the best, but having the best possible extension in the store. Enhancements should be the way rather than just replacing if possible. I guess it depends on who you ask about what's best 🙂

Agree.

PR #15782, opened 2024-12-10
Add badger extension

bobooon:
## Checklist


pernielsentikaer (changes requested):
Is this language aware? I switched my machine into `Danish` and now I don't see any numbers - do you know if that's possible?

pernielsentikaer:
The problem is if it's not is, then it will fail for a lot of users which makes it a bad experience, so I would say yes - you can maybe use index (click element 2) instead of names

Issue #15571, opened 2024-11-27
[Apple Mail] Extension command "Refresh mail" regularly displays error message 


BSoDium:
### Extension

https://www.raycast.com/yug2005/mail

### Raycast Version

1.86.1

### macOS Version

15.1.1

### Error message
As it appears inside Raycast
```
Error: Toast API is not available when command is launched in background
```

### Screenshots
![CleanShot 2024-11-27 at 20 04 38@2x](https://github.com/user-attachments/assets/ae87abdb-359d-48d0-91d2-1f3f1d912aa8)
![CleanShot 2024-11-27 at 20 10 12@2x](https://github.com/user-attachments/assets/020f1c26-8dca-4612-a499-f90d77f7ec54)

### Steps To Reproduce

- open Raycast
- scroll down in the suggestions until "Refresh mail"
- hit ⌥⌘E to display the error

If you need any other informations regarding my system configuration, I'll gladly at them to this issue.

### Current Behaviour

An error message/notice initially appears next to the command after a system restart, then disappears once the command has been executed successfully. It will occasionally reappear later, and disappear again after executing the command.

### Expected Behaviour

No error message displayed, or a more user-friendly explanation of the failure.

yug2005:
@raycastbot assign me

I'll take a look

yug2005:
@BSoDium Yes you can close this. It should be resolved

PR #15795, opened 2024-12-10
[Mail] Fix timeout error messages with refresh mail

yug2005:
## Description

The refresh mail command was running some applescripts that were going over the timeout, which was producing an error. This addresses the issue #15571.

This PR has the following changes: 
* Applescript is ran within a try and catch block so that the timeout error is properly handled
* Increased the timeout duration for the applescript from 30s to 60s
* Decreased the message limit options (max limit is now 20). This was the limit for me when I was testing with 60s timeout
* Added some more toasts for when fetching accounts and mailboxes fail
* Only showing toasts when the launchType is user initiated

Can someone else verify this change is working on their end? I noticed that the timeout parameter for the function `runAppleScript` is very buggy and sometimes just refuses to work at all.

## Checklist

### DIFF
--- a/extensions/mail/CHANGELOG.md
+++ b/extensions/mail/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Mail Changelog
 
+## [Fix] - 2024-12-10
+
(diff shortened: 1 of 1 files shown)


pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #15814, opened 2024-12-11
Update pomodoro extension

mikikiv:
## Description

Moves the setting for menu bar time display to the `pomodoro-menu-bar` command.
Adds the time left as a menu bar dropdown item when time display is disabled
Adds a time left countdown to the control timer command in the Raycast window 
Refactors a hook to get the amount time left easier
Reorder preferences to make it a bit easier to look through
Removes the custom type for preferences in favor of the generated type from Raycast

## Screencast

CleanShot_2024-12-11_10 02 27AM

## Checklist

### DIFF
--- a/extensions/pomodoro/CHANGELOG.md
+++ b/extensions/pomodoro/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Pomodoro Changelog
 
+## [Enhancement] - {PR_MERGE_DATE}
+
+- Show time left in menubar submenu when time left is hidden in menubar
+- Fixes focus not turning off when manually ending session
+- Reorders preferences
+
 ## [Enhancement] - 2024-12-11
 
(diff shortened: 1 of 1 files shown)


LitoMore:
I just changed my PR to move the timer to the dropdown list if it's hidden.

I think your implementation makes the process complicated.

What do you think?

mikikiv:
The main reason for the complexity in this is to get the timer to update every second in the menu dropdown while its open
You're solution would work but would only update when manually rerendered, or every 10 seconds

mikikiv:
Attached a screencast of the proposed change here so you can see the benefit

But maybe I've made too many changes for one PR, I can work on that. If you want to continue with your change first I don't have a problem with it

pernielsentikaer:
So what's the status regarding this PR after the other one was merged?

mikikiv:
Going to break this up into a couple of different requests

PR #15842, opened 2024-12-13
Add synonyms extension

remorses:


extensions/synonyms/README.md
```diff
@@ -0,0 +1,3 @@
+# Synonyms
+
+Find synonyms for words and translate if necessary
```
pernielsentikaer, line 3: I think we need to explain more about what it does and which services it uses :)

We have other alternatives in the store. Why use this one?

remorses: I added a more in depth description, thank you

I already use this extension every day and though I should publish it, too useful

pernielsentikaer (changes requested):
Could we support Raycast AI too?

remorses:
I added support for Raycast AI

pernielsentikaer:
I just tried the new version and I'm getting this error, could you look into that?

Raycast02012025-r4DCelp9 png@2x

remorses:
Should be fixed now, not sure why It was not happening for me too

pernielsentikaer:
It seems like the subtitle sometimes returns `null` as a string, which results in "null" being printed in the subtitle. 
Could you look into that?

remorses:
bump

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #15868, opened 2024-12-15
Update Pomodoro preference organization and use generated types

mikikiv:
## Description

-  Update organization of preferences to move preferences that are often overlooked to the top
- Remove type definition on favor of using the generated type for preferences

## Checklist

### DIFF
--- a/extensions/pomodoro/CHANGELOG.md
+++ b/extensions/pomodoro/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Pomodoro Changelog
 
-## [Chore] - {PR_MERGE_DATE}
+## [Chore] - 2025-01-03
 
 - Update preferences organization
 


LitoMore:
We have two menu bar commands. I think move the preference to the extension level is better.

pernielsentikaer:
I guess it's a tough one when moving preferences because if users have already set this preference, then we're resetting it for them, right?

pernielsentikaer:

mikikiv:
@pernielsentikaer Yep, sure is
If all the settings remain at the same level, reordering them shouldn't impact previously saved preferences right?

pernielsentikaer:
Thats should be correct 🙂

mikikiv:
This can be reviewed. It only reorganizes preferences to move some more popular ones to the top and removes an unused defined type

image

mikikiv:
This can be rereviewed. It updates the organization of preferences and removes an unused type

Settings_2024-12-23_01 00 27PM@2x

pernielsentikaer:
I guess it's nice to have them a bit more organized. What do you think, @LitoMore?

LitoMore:
For this extension, I think we move all configurations to the root level. The common and slack menu bars are two different commands but have almost the same functionaries.

It is better to put it at the root level to allow users to configure it once and apply it to both commands.

pernielsentikaer:
That makes sense, also, you don't use both I guess - let's do that change and then users will have to enable/re-enable what's missing @mikikiv

mikikiv:
We are all set then, all the timer preferences are at the root level

pernielsentikaer (approved):

PR #15928, opened 2024-12-18
Add translate-send-webpage-to-reader extension

ajason:
## Checklist


pernielsentikaer (changes requested):
Hi 👋

It looks like that's already possible https://www.raycast.com/zach/readwise-reader - could you check that one out?

ajason:
Hi,

The functionality added by this pull request cannot be supported in the Readwise Reader plugin for the following reasons:

The Readwise Reader plugin can only save a single URL to Reader, and the parsing of the corresponding web content is handled by the reader itself. 
The plugin I submitted retrieves the webpage's markdown content through the Raycast browser plugin and sends that markdown content to the reader via an API for saving.
The reason for retrieving the webpage content is that I can translate the entire text using Raycast AI before saving it, as the Reader does not offer a full-text translation feature.

pernielsentikaer (changes requested):
- You're missing an extension icon, you can generate [one here](https://icon.ray.so/) 

PR #15983, opened 2024-12-21
Update IPCheck to V1.1

jason5ng32:
## Description

Update my extension IPCheck to V1.1

## Checklist

### DIFF
--- a/extensions/ipcheck-ing/CHANGELOG.md
+++ b/extensions/ipcheck-ing/CHANGELOG.md
@@ -1,6 +1,6 @@
 # IPCheck Changelog
 
-## [Version 1.1] - {PR_MERGE_DATE}
+## [Version 1.1] - 2025-01-02
 
 - Add options to choose IP sources to fetch from
 - Improve code quality


extensions/ipcheck-ing/package.json
```diff
@@ -9,6 +9,66 @@
...
   "license": "MIT",
+  "preferences": [
+    {
+      "type": "checkbox",
+      "name": "getLocalIPs",
+      "title": "Local IPs",
+      "description": "Get Local IPs",
+      "label": "Get Local IPs",
+      "default": true,
+      "required": true
```
pernielsentikaer, line 20: Let's set them all to `false` so that users who already have it installed will not be interrupted here

jason5ng32: I've test locally, I think it might be better to set all `default` to true and all `required` to false, so that new users and old ones both can fetch IP details acquiescently

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #16028, opened 2024-12-24
Update battery-health extension

menahishayan:
## Description

feat: ✨ added the ability to view your battery's health relative to when it was new

## Screencast

Screenshot 2024-12-24 at 2 02 43 PM

## Checklist

### DIFF
--- a/extensions/battery-health/src/index.tsx
+++ b/extensions/battery-health/src/index.tsx
@@ -46,9 +46,10 @@ export default function Command() {
   const { health, percentage } = useMemo(() => {
     const values: TValues = DEFAULT_VALUES;
     if (state?.batteryRegistry && !state.isLoading) {
-      const { AppleRawMaxCapacity, DesignCapacity, CurrentCapacity } = state.batteryRegistry;
-      if (!isNaN(AppleRawMaxCapacity / DesignCapacity)) {
-        values["health"] = ((AppleRawMaxCapacity / DesignCapacity) * 100).toFixed(1) + "%";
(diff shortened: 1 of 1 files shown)


o1y:
Danke für deine Contribution. Die Health Angabe scheint nicht übereinzustimmen mit der Angabe der maximalen Kapazität im Systembericht. Meines Wissens nach lässt sich dieser Wert nicht wie 


```

**Comment by o1y** [comment]
Thanks for your contribution @menahishayan! As far as I understand, the Health value cannot be accurately represented using the following calculation:

```javascript
values["health"] = ((AppleRawMaxCapacity / DesignCapacity) * 100).toFixed(1) + "%";
```

The result doesn't align with the health information displayed in the system report. When I developed the extension, I couldn't find a reliable way to extract that value directly. In my case, the calculated value is significantly lower than the one provided by Apple, which could be confusing for users.

menahishayan:
Hey @o1y,

Thanks for the info. I've compared this value with other popular apps like [CoconutBattery](https://coconut-flavour.com/coconutbattery/) and the value is consistent with what they display. 

While it may not be exactly the same as the System Report, it's close enough to a point where I believe it's more useful to **_show_** the value than to omit it entirely.

menahishayan:
Update: I've changed the calculation to be slightly more accurate to the System Report value. It's not 100% but the margin of error is now smaller.

o1y:
I've added the Maximum Capacity value from the system profiler in #16032.

0unsalalcin0 (approved):

menahishayan:
Closing in favor of https://github.com/raycast/extensions/pull/16032

menahishayan:
Closing in favor of https://github.com/raycast/extensions/pull/16032

PR #16046, opened 2024-12-26
Add multi-prompt extension

edgarasben:
## Checklist


pernielsentikaer (changes requested):
What if the prompt should not be in the end of the URL, does it support that?

edgarasben:
> What if the prompt should not be in the end of the URL, does it support that?

Currently, it only supports chatbots that use a URL data structure, such as query parameters.

PR #16051, opened 2024-12-27
Add vainjs-raycast extension

youngluo:
## Checklist


pernielsentikaer:
Hi 👋

Sounds a bit like [this extension](https://www.raycast.com/ajaypremshankar/projects) to be, could you check that one out?

youngluo:
> Hi 👋
> 
> Sounds a bit like [this extension](https://www.raycast.com/ajaypremshankar/projects) to be, could you check that one out?

This extension is simpler to use and supports configuring multiple workspaces and editors, as well as quick switching of editors, which is necessary

pernielsentikaer:
I'm sure @ajaypremshankar is up for enhancements since having multiple versions of the same extension is confusing for many

youngluo:
Uh, at present, it seems that it cannot satisfy my scene. If you take PR contribution code, it is estimated that the thinking between us is also very different.For the time being, I only use it myself, haha～
I list my functional points, maybe others have similar needs：
1. Support configuring multiple workspace, which is convenient for targeted search for personal or job projects
2. Support configuring multiple editors, allowing the frontend to freely switch between the default editor that opens the project
3. The list display project path and current git branch
4. Support configuration to exclude unnecessary folders

pernielsentikaer:
It might be more in line with https://www.raycast.com/zerowidth/quick-open-project which missing some of the features, yes, but looks like it's doing the same, right?

youngluo:
In my personal opinion, as an expansion of the market, similar expansions should not be prevented from being released.

pernielsentikaer:
It's actually mentioned in our [Extensions Guideline](https://manual.raycast.com/extensions) under `Reasons to be rejected` 

We're trying hard to enhance everything possible to not confus users with too many extension of the same thing and similarities

youngluo:
Maybe, but if this rule, when a expansion function cannot meet the user and the update is slow, it is likely to cause users to lose. There is no progress without comparison.

pernielsentikaer:
I'm not sure I understand what you mean by slow update. What are you referring to?

PR #16065, opened 2024-12-28
Add advanced-replace extension

Gitsack:
## Description

This extension allows for a quick but advanced way to search and replace a text (selection or clipboard).

It works by letting users define one or multiple Regex items which extract strings from the given input and saves them under specific keys. These keys can be used to construct some output afterwards using the curly braces syntax.

Using a hotkey to trigger the extension, the replacement can be done easily and quickly.

### Example: Github Image Replacements

Replace the markdown version of an image with the HTML `` version.

#### Item 1 - key: src

```
\!\[.*\]\((.*)\)
```

#### Item 2 - key: alt

```
\!\[(.*)\]
```

#### Output

```
<img src="{src}" alt="{alt}" width="400" height="300">
```

## Screencast

https://github.com/user-attachments/assets/78bb4b5e-601e-4de9-a2c5-a6709d8b7e74

## Checklist


pernielsentikaer:
Should we consider moving the output field last, it's a bit confusing it's before the matches?

I can't insert more than one char is the key field, could you look into that?

Raycast09012025-pSrKFDY0 png@2x

Gitsack:
@pernielsentikaer thanks for the feedback. I switched the order which really makes more sense, and I fixed the issue regarding the key field. 👍

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #16078, opened 2024-12-29
Updated `fastmail-masked-email` extension

mike182uk:
## Description

Misc updates to `fastmail-masked-email` extension

- When a masked email is created, it will now be in the `pending` state. This allows Fastmail to automatically delete the masked email after 24 hours if it is not used
- Updated create command to show the masked email after creation and subsequently allow creation of a new masked email if you do not like the one previously created (similar functionality to 1Password's Fastmail integration)
- Added email state filter to the list command and removed now redundant configurable preference for displaying `pending` & `deleted` masked emails
- List command now sorts masked emails by creation date and displays the relative creation date in the list

## Screencast

**Filterable list with creation date**:

**Show masked email after creation**:

## Checklist

### DIFF
--- a/extensions/fastmail-masked-email/src/create.tsx
+++ b/extensions/fastmail-masked-email/src/create.tsx
@@ -33,9 +33,11 @@ export default function Command() {
     try {
       const email = await createMaskedEmail(prefix, description);
 
-      setMaskedEmail(email);
+      Clipboard.copy(email);
 
-      toast.hide();
+      await showHUD(`🎉 Masked email copied to clipboard`);
+
+      await closeMainWindow({ clearRootSearch: true, popToRootType: PopToRootType.Immediate });
     } catch (error) {
       toast.style = Toast.Style.Failure;
       toast.title = "Failed to create masked email";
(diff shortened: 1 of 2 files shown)


extensions/fastmail-masked-email/src/create.tsx
```diff
@@ -44,11 +46,29 @@ export default function Command() {
     }
   };
 
+  const handleCopy = async () => {
+    if (!maskedEmail) {
```
klaaspieter, line 50: I'm not a fan of this behavior (nor the code to implement it). I think this state is undesirable and unnecessary:

CleanShot 2025-01-07 at 16 09 54@2x

What is the reason we can no longer copy the created email immediately? I for one rarely create more than one masked email at a time so I'd strongly prefer the old behavior to this.

Maybe your desired use case can be a separate action with it's own shortcut?

mike182uk: With this change I wanted to try and emulate behaviour from 1Password's Fastmail integration where you can preview the generated email before using it

https://github.com/user-attachments/assets/14f98775-8c06-485b-93e1-06ab111f9849

This is nice for when the randomly generated email is a little _too_ random for you and you would prefer something else. With the current workflow, you don't know what the email is going to without pasting it from the clipboard, and you would then have to trigger the action multiple times to get an email more to your liking, each time providing the prefix + description which is tedious

We have the `Quick Create Masked Email` action already that bypasses the UI altogether, generates an email and adds it to the clipboard which provides which I thought would mostly covers the previous behaviour of this action

klaaspieter: I see. Thanks for explaining. I did consider quick create as well but that doesn't allow adding a label. The way I use the extension is generally because 1Password fails to recognize I'm signing up somewhere and I like to label the websites I use the email with like 1Password does.

I see a few possible options that could satisfy us both:

1. Copy the email address _and_ show it like you're already doing. Kind of like the 1Password extension does when you “Generate Password”
    https://github.com/user-attachments/assets/fd5e3aa9-3f32-4b3a-870e-d7f316b128d4

2. Make your use case a separate “Preview Masked Email” action
3. Add a way to configure the preference

What I don't like about the first one is that it leaves Raycast in that state until you dismiss it but there is precedence for it working that way. I think 2/3 are hard to explain to new users but I wanted to at least acknowledge the possiblities.

mike182uk: Thanks for the feedback @klaaspieter 

I've reverted the `cmd + enter` functionality to what is was, and added the new interactive functionality to `shift + cmd + enter` which I believe should suit both of us, biasing to retaining the existing workflow

interactive-create

I've made it so the new "interactive" create functionality also copies the generated masked email to the clipboard instead of having a dedicated action to copy the generated masked email

Let me know what you think

klaaspieter (approved):
Looks good. Thank you!

PR #16113, opened 2025-01-02
[Google Chrome] fix Search Tabs command

### LINKED ISSUE #15047: [Google Chrome] ...search all command not working on bookmark it return opentabs instead of bookmark
### Extension

### Raycast Version

1.84.3

### macOS Version

15.1

### Description

[Google Chrome] ...search all command not working on bookmark it return opentabs instead of bookmark

### Steps To Reproduce

[Google Chrome] ...search all command not working on bookmark it return opentabs instead of bookmark

### Current Behaviour

[Google Chrome] ...search all command not working on bookmark it return opentabs instead of bookmark

### Expected Behaviour

[Google Chrome] ...search all command not working on bookmark it return opentabs instead of bookmark

### LINKED ISSUE #15681: [Google Chrome] Search Tabs Command Empty
### Extension

### Raycast Version

1.86.1

### macOS Version

15.1.1

### Description

In the Google Chrome extension, the "Search Tabs" command appears empty. I only use one Chrome profile (which I've read having more can cause issues).

### Steps To Reproduce

Open "Search Tabs" command. Result will be empty.

### Current Behaviour

_No response_

### Expected Behaviour

_No response_

Tarocch1:
## Description

This PR fixes the following two issues:

1. When the "Extract the favicon from every open tab" option is enabled, the Search Tabs command cannot be used.
2. Closes #15047
3. Closes #15681

## Screencast

For issue 1:

|Before|After|
|--------|--------|
|CleanShot 2025-01-02 at 23 06 35@2x|CleanShot 2025-01-02 at 23 06 01@2x|

For issue 2:

|Before|After|
|--------|--------|
|CleanShot 2025-01-02 at 23 08 22@2x|CleanShot 2025-01-02 at 23 08 51@2x| 

## Checklist

### DIFF
--- a/extensions/google-chrome/CHANGELOG.md
+++ b/extensions/google-chrome/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Google Chrome Changelog
 
-## [Fix] - {PR_MERGE_DATE}
+## [Fix] - 2025-01-09
 
 - Fix Search Tabs command when "Extract the favicon from every open tab" is enabled.
 - Fix bookmark list in the Search All command.


a-laughlin:
Nice PR! Thanks for submitting it.

On Issue 2, the before/after images show `vite/packages` moving from history to bookmarks. That's intentional?

extensions/google-chrome/src/hooks/useHistorySearch.tsx
```diff
@@ -8,6 +8,11 @@ const whereClauses = (tableTitle: string, terms: string[]) => {
   return terms.map((t) => `(${tableTitle}.title LIKE '%${t}%' OR ${tableTitle}.url LIKE '%${t}%')`).join(" AND ");
 };
 
+/**
+ * Generates a query to search the history for a list of terms.
+ *
+ * Using `last_visit_time > 0` to filter out bookmarks.
+ */
```
a-laughlin, line 15: Appreciate the comment.

I'm unfamiliar with this part of the code. Chrome stores the bookmarks in history? I thought it had a separate bookmarks file.

extensions/google-chrome/src/components/ChromeListItems.tsx
```diff
@@ -9,10 +9,18 @@ export class ChromeListItems {
...
 
-function HistoryItem({ profile, entry: { url, title, id } }: { entry: HistoryEntry; profile: string }): ReactElement {
+function HistoryItem({
+  profile,
+  entry: { url, title, id },
+  type,
+}: {
+  entry: HistoryEntry;
+  profile: string;
+  type: "History" | "Bookmark";
```
a-laughlin, line 19: I'm unfamiliar with the implementation difference between bookmarks and history. It seems strange that bookmarks would use a history entry here. Are they so similar that bookmarks and history will always use the same logic?

Tarocch1:
@a-laughlin We extract the browsing history from the file `~/Library/Application Support/Google/Chrome/Default/History`, which is an SQLite database file. The `urls` table inside contains bookmarks as well as history, and it includes the following fields:

1. `url`: The url of bookmark or history.
2. `title`: Website title.
3. `visit_count`: If a url is only a bookmark and not in the history, this field is 0; otherwise, it will record the number of times it has been accessed.
4. `last_visit_time`: If a url is only a bookmark and not in the browsing history, this field is 0; otherwise, it will record the timestamp of the most recent visit.
5. `id`: Index.

Therefore, you can see that this table contains both history and bookmarks.

| Type | `last_visit_time` |
|--------|--------|
| History | >0 |
| Bookmark | =0 and part of >0 (For urls that are both history and bookmarks.) | 

Therefore, we can only precisely identify the history from this file, but we cannot accurately identify the bookmarks. As a result, in the original code, bookmarks are obtained from another file `~/Library/Application Support/Google/Chrome/Default/Bookmarks`

This explains the two screenshots for issue 2, as per the previous code:

1. Since the last_visit_time filter was not used, it retrieves bookmark data that has never been accessed (specifically the ones you mentioned `vite/packages`). The timestamps for these URLs are all 0, which is why you can also see in the screenshots that their access time is all set to the year 1601.
2. The previously known issue is that the incorrect data source was used to display bookmarks, which is why the first screenshot does not show the bookmark list at all.

The second screenshot can display correctly.

Tarocch1:
@a-laughlin In the original code, both history and bookmarks share the same data structure, which is the HistoryEntry type. This is not an issue, because for both, we actually only need the information of the `url` and `title` fields. Precisely for this reason, they can also use the same component to render the list.

However, if a url is both a bookmark and history, then this url will have the same ID in both the `History` and `Bookmarks` files. In the `Search All` command, this url will be rendered in the list twice: once as a bookmark and once as history.

Two identical IDs will cause errors in the list, so I added a Type parameter to the list component to differentiate the IDs.

a-laughlin:
Thanks for taking the time to explain things @Tarocch1!

It sounds like -all- the plumbing for retrieving bookmarks is unnecessary since both bookmarks and history can be retrieved by just partitioning history into two lists.  Is that accurate?

If so, I'm fine with this PR.  I'm also traveling and my availability/reception are insufficient to fully follow the logic and test the various functions. At a minimum more testers would be helpful. I believe Raycast staff will test before merging.

If there's a chance this PR will fix https://github.com/raycast/extensions/issues/15681, maybe @andrewcincotta or @tomh4 can test to confirm it.

tomh4:
I can test tomorrow!

Tarocch1:
@a-laughlin 

> It sounds like -all- the plumbing for retrieving bookmarks is unnecessary since both bookmarks and history can be retrieved by just partitioning history into two lists. Is that accurate?

It is not the case. We can confirm that in the `History` file, any last_visit_time greater than 0 is definitely a history entry, while those equal to 0 are definitely bookmarks. However, some entries with a last_visit_time greater than 0 may also be bookmarks, and we do not know which ones.

Therefore, we still need to obtain an accurate list of bookmarks from the `Bookmarks` file, just as the previous code did.

tomh4:
Gave it a test run today and search tabs looks like it is all working as expected @Tarocch1 @a-laughlin 
but history e.g. only worked on the second attempt, the first try threw an error but I cant get the message back.

Tarocch1:
> in prepare, database disk image is malformed (11)

This error is due to an issue with the SQLite database file, possibly caused by Chrome performing some operation on the file. It can be confirmed that this is not an issue introduced by our code. Perhaps we can simply catch and ignore this exception?

a-laughlin:
> Gave it a test run today and search tabs looks like it is all working as expected @Tarocch1 @a-laughlin
but history e.g. only worked on the second attempt, the first try threw an error:

Thanks for testing @tomh4 !

a-laughlin:
> This error is due to an issue with the SQLite database file, possibly caused by Chrome performing some operation on the file. It can be confirmed that this is not an issue introduced by our code. Perhaps we can simply catch and ignore this exception?

If we can still return correct results when the error occurs, ignoring it makes sense.  If not, we'll still get bug reports, so we should have something more robust.

A couple options:
1. Check if chrome is operating on the file (if possible) and wait for it to finish. Throw if it still fails after waiting. This would be the most precise and assure we're handling cases correctly, but potentially requires more maintenance of the different cases on chrome changes, and more bug reports until we get all the possible error cases.
2. Catch the error, then delay and retry without knowing what chrome is doing. On failed retry then throw an error.  This works even if chrome throws a different error, and we don't have to know all the cases or maintain them.

In this volunteer extension context, # 2 seems better to me since we only contribute intermittently and less maintenance time is more important than error handling precision.

Tarocch1:
> 2. Catch the error, then delay and retry without knowing what chrome is doing. On failed retry then throw an error. This works even if chrome throws a different error, and we don't have to know all the cases or maintain them.

I agree. I will add the retry logic later.

andrewcincotta:
> Thanks for taking the time to explain things @Tarocch1!
> 
> It sounds like -all- the plumbing for retrieving bookmarks is unnecessary since both bookmarks and history can be retrieved by just partitioning history into two lists. Is that accurate?
> 
> If so, I'm fine with this PR. I'm also traveling and my availability/reception are insufficient to fully follow the logic and test the various functions. At a minimum more testers would be helpful. I believe Raycast staff will test before merging.
> 
> If there's a chance this PR will fix #15681, maybe @andrewcincotta or @tomh4 can test to confirm it.

The PR works on my end!

Tarocch1:
@a-laughlin I have added the retry logic. After the SQLite error, wait one second before retrying. If the retry fails again, it will throw an error just like the original logic. I think this should resolve the issue.

a-laughlin:
Awesome. Thanks for adding a retry @Tarocch1 . PR seems good to go once Raycast staff review it.

cc @pernielsentikaer

a-laughlin (approved):

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

Issue #11107, opened 2024-03-05
[KeePassXC] General security note


imkebe:
### Extension

https://www.raycast.com/zkytech/keepassxc

### Description

First of all, it's worth saying thank you for your contribution!
However I would raise an important issue in terms of security. The architecture of this extension is in my opinion, it negates the assumptions that Keepass* should have.

My main objective is that it's not using the HTTP server that is exposed by the open Keepass* app, which is should instead. That should works similar to the Browser extension that query the Keepass* to obtain the credentiials when the database is unlocked in Keepass*.

Instead the extension stores the:
1. Location of the database
2. Password

Keepass* database password shoudn't be saved anywhere other than the user's memory. If the Raycast has an vulnerability there will be an easy way of gaining access to the database. In fact i can image targeted attack on Raycast.

### Steps To Reproduce

1. Install the extension
2. Configure the extension

zkytech:
1. Is there any http api to get account list ? or search by keyword ?
2. `If the Raycast has an vulnerability there will be an easy way of gaining access to the database. In fact i can image targeted attack on Raycast.` . Why http api can avoid this ?
3. In theory, there are indeed some methods that can give you a stronger sense of security, but I don't have time to implement them. Anyway, Pull requests welcome.

Sulfyderz:
Hello @imkebe,

>> if the Raycast has an vulnerability there will be an easy way of gaining access to the database. In fact i can image targeted attack on Raycast. 

>Why http api can avoid this ?

I agree with @zkytech. However @imkebe, I understand your concern. Whenever I export my Raycast config file, I always erase the password saved in the extension before. If you can propose an alternative, feel free to do a PR. Unfortunately, I don't have time as well.

imkebe:
The main difference is that the database is only accesible then user unlock the database while with the extension it's always accesible. I know that http api has also needs to be authenticated by exchanging some credentials - however this is not the master password.

I just wanted to point those security issues to the community.

Issue #10709, opened 2024-02-11
[KeePassXC] Allow more control over the presented information when searching for records


yaronuliel:
### Extension

https://www.raycast.com/zkytech/keepassxc

### Description

Currently, when searching for a password - the results show the title of the password, and the path to the records inside the directories as tags, as shown in the following image

<img width="752" alt="image" src="https://github.com/raycast/extensions/assets/5229872/486bef90-cfe0-4e9b-a170-9b75b09a1ac2">

Inside KeePassXC - a lot more information is shown on the search results, which help a lot when trying to understand what record to pick

I think it would be very helpfull if the extension allowes some configration of "what information to show" - including 

* Username
* URL
* Has TOTP?

As a  good start, showing the Username is crucial for picking the correct result.

Also, it might save some space if the path to the records is shown in a "breadcrumbs" format rather then individual tags (i.e. `root dir > dir1 > dir2 > dir 3`

### Who will benefit from this feature?

Every user of the extension, especially those that have big KeePass files, with multiple records with the same title, differing in other property, such as username

Sulfyderz:
Hello,

Thank you for your feedback. I am a contributor to the extension, not the original creator (i.e. @zkytech).

> I think it would be very helpfull if the extension allows some configration of "what information to show" - including
> 
> Username
> URL
> Has TOTP?
> As a good start, showing the Username is crucial for picking the correct result.

Well, I do agree with you for the Username. Currently, I am using a workaround. My titles contain the username when there are homonyms (e.g. 'Google (Sulfyderz@mail.com)' and 'Google (Boba@mail.com)'). I think it would be a great feature to add that information.

Concerning the URL, I don't think it's that useful. Give me an example, maybe I'm wrong.

About the TOTP, it can be more or less useful. Why more or less? When you are on a website, you'll be prompted anyway to insert a TOTP code (you can then apply the code with `opt + enter` on the entry from the extension). Before that moment, you don't really need to know if an input does have a TOTP set. Still, give me an example in which case you find it insightful.

> Also, it might save some space if the path to the records is shown in a "breadcrumbs" format rather then individual tags (i.e. root dir > dir1 > dir2 > dir 3

I had repositories inside my keepassXC, before I use that extension. Not anymore. But yes, I do agree with you that your idea is a best UI.

If I have some time, I would gladly do all of this. But currently, I don't. May be in the end of 2024. @zkytech seems kind of busy as well. What I suggest is to do a PR.

Best,
Sulfyderz.

[...]

Issue #10634, opened 2024-02-07
[KeePassXC] Master Password


vitusli:
### Extension

https://www.raycast.com/zkytech/keepassxc

### Description

Dear Developer, I am happy about your great extension. It is straight forward and almost exactly what I was looking for. I wanted to ask you something about safety concerns: Would it be possible to add a lock/unlock database feature, so I can protect my database when I want to? I dont feel comfortable with having a always unlocked password manager on my mobile laptop. One can simply copy all my passwords out of raycast.

### Who will benefit from this feature?

Probably everyone

vitusli:
Thank you for your very fast answer. I am really new to coding so it is something I am very interested in but I have no clue yet. I got a friend who is very good with that and will come back later. Thank you very much! 

Best

Issue #9600, opened 2023-12-08
[KeePassXC] Add PIN to use extension


boggleboggle:
### Extension

https://www.raycast.com/zkytech/keepassxc

### Current behavior:
Once you have entered the KeePass password and key file, anyone with access to the computer can access all the passwords in the database. This is potentially insecure - particularly on laptops, which could be used in public areas.

### Suggested feature:
Add an option to add a PIN, which would be required to access the database.

### Who will benefit from this feature?

Users who are concerned about unauthorized access to the database.

Sulfyderz:
Hello,

It sounds a nice idea. @zkytech is the owner of the extension. I don’t know what he would think about it.

PR #16143, opened 2025-01-03
Update keepassxc extension

pabroux:
## Description

Hello @zkytech,

I hope you're fine.

Here's what've done:

- Add a timer to lock the database after a user specified time of inactivity ([#9600](https://github.com/raycast/extensions/issues/9600), [#10634](https://github.com/raycast/extensions/issues/10634), [#11107](https://github.com/raycast/extensions/issues/11107));
- Remove the password and the key file fields from the extension preference;
- Add indicators (TOTP, password and URL) and usernames while searching entries ([#10709](https://github.com/raycast/extensions/issues/10709));
- Add a folder filter on the search bar when folders exist in the database;
- Use the official KeePassXC.app icon as the extension icon;
- Use `keepassxc-cli export` to retrieve entries and most other data from the database instead of relying heavily on `keepassxc-cli show`;
- Update the README.md;
- Update the screenshots;
- Rewrite the code entirely;
- Comment the code.

Give me your thoughts.

_p.s.: If you no longer want to maintain this extension (not enough time?), tell me. If you agree, I'll set myself as the author in order to keep it up-to-date more often. I've just learned that some want to integrate the YubiKey ([#3668](https://github.com/raycast/extensions/issues/3668))._

## Checklist


zkytech:
Awesome!

You can set yourself as author

zkytech (changes requested):
1. "-q" option is required on my mac, otherwise it will throw error before input password.
2. filter out "Trash" folder. 
3. show folder path.

Sulfyderz:
Hello @zkytech,

> 1. "-q" option is required on my mac, otherwise it will throw error before input password.

I don't have any bug on my side concerning the omission of the quiet (_-q_) option. It doesn't bother me to add it but could you tell me what's the error? What's your macOS version and KeePassXC version? May be I can find a fix.

> 2. filter out "Trash" folder.

When I delete an entry in KeePassXC 2.7.9, it goes to a new group/directory called "_Recycle Bin_". Here is a simple example of a CSV line with a deleted entry generated by `keepassxc-cli export -f csv`:
`"MyDumb-KeePass/Recycle Bin", "KFC", "ElonMusk@live.fr", "myPassword", "https://www.kfc.fr", "", "", "0", "2025-01-05T08:47:38Z", "2023-07-31T17:36:28Z"`.

I know you had this treatment before my changes, especially with the chinese translation `"回收站"` (thanks _Deepl_ !), but I wasn't able to reproduce this. Is it specific to you?

In addition, I understand you don't want to see the trashed entries but I've found sometimes to be useful to have them. I have two cases in mind:

- On a website, even if I did delete an account, it wasn't removed (due to weird things). I was still then able to log in with by trashed entry;
- Another example is for those who delete an entry even if they have just temporary deleted an account on a website (e.g. instagram). It may be useful to keep them at hand.
 
What do you think? 

By the way, in a future version, I've plan to add a [filter to the search bar](https://developers.raycast.com/api-reference/user-interface/list) in order to choose the folder (e.g. _All_, _Trashed_, etc.)

> 3. show folder path.

Well, it bothers me for several reasons:

1. Too much data rendered, that's why I've planned to add a [filter to the search bar](https://developers.raycast.com/api-reference/user-interface/list), much like the [Brew extension](https://www.raycast.com/nhojb/brew) (check the first screenshot of the extension);
2. If someone has a KeePass database with nested groups/directories, the rendering is awful. Worst, the username disappears.

Sulfyderz:
I'm working on making the filter on the search bar. I'll get back to you once done.

zkytech:
1. Error message without -q 

- keepassXC version `2.7.9`
- macOS `15.1.1 (24B91)`  

It is a password input hint: "Input Password to unlock"



2. Consider using the KeepassXC application directly in extreme scenarios, rather than increasing the cost of information filtering in this extension that focuses on convenience. 
> The default behavior of KeepassXC's search function regarding the recycle bin is not to display it. 

3. What you described is still an extreme scenario; please prioritize convenience for the majority of situations. For me personally, the display of usernames is not that important because, in most cases, the usernames are exactly the same. On the other hand, directory paths are crucial information because if I intend to use directories, it means that the names of entries under each directory are highly likely to be duplicated.

For example, database accounts for the same business in different environments have the same username but different IP addresses. I need to use directories to distinguish between environments. I have created many three-level nested directories for managing service/server/database accounts under different environments, the new ui cannot distinguish them. 

image

- For user who use folder to manage passwords, the directory name matters. 
- For user who do not use folder to manage passwords, the directory name is short and will not cause the username to disappear

zkytech:
`src/index.tsx` and `src/utils/keepassLoader.ts` are not used,  please delete them.

Sulfyderz:
> 1. Error message without -q
> 
> * keepassXC version `2.7.9`
> * macOS `15.1.1 (24B91)`
> 
> It is a password input hint: "Input Password to unlock"
> 
> 

Thank you for your details.
Ok, weird behavior. I'll add the `-q` option.

> 2. Consider using the KeepassXC application directly in extreme scenarios, rather than increasing the cost of information filtering in this extension that focuses on convenience.
> 
> > The default behavior of KeepassXC's search function regarding the recycle bin is not to display it.

Good advice. I think you're right about using KeePassXC in extreme scenarios. Thank you.
Moreover, I didn't know the default behavior of the search function. I'll add your work. I've found a little improvement I can make.

> 3. What you described is still an extreme scenario; please prioritize convenience for the majority of situations. For me personally, the display of usernames is not that important because, in most cases, the usernames are exactly the same. On the other hand, directory paths are crucial information because if I intend to use directories, it means that the names of entries under each directory are highly likely to be duplicated.
> 
> For example, database accounts for the same business in different environments have the same username but different IP addresses. I need to use directories to distinguish between environments. I have created many three-level nested directories for managing service/server/database accounts under different environments, the new ui cannot distinguish them.
> 

Great advice. You're right again. I lost myself in details.
In addition, I didn't think about different IP addresses.
I'll add your work.

> * For user who use folder to manage passwords, the directory name matters.
> * For user who do not use folder to manage passwords, the directory name is short and will not cause the username to disappear

I saw a little improvement to do in your code: hidding the icon if no folder is used. I'll do a commit.

zkytech:
> @zkytech, everything you wanted has been committed.

-q is still missing

Sulfyderz:
> > @zkytech, everything you wanted has been committed.
> 
> -q is still missing

My bad.

Sulfyderz:
I've added a [folder filter to search bar when folders exist.](https://github.com/raycast/extensions/pull/16143/commits/02ff349b9b94079df6cb815de525c1030d129a16) as a new feature.

Feel free to give your thoughts as usual.

pernielsentikaer:
Is this ready @zkytech 😊

Sulfyderz:
I've committed a more cleaned rendering while searching. It is much inpired by the [Bitwarden Vault](https://www.raycast.com/jomifepe/bitwarden) extension. I've also added two new indicators: password and URL as wanted in a certain way in [#10709](https://github.com/raycast/extensions/issues/10709).

Here are the screenshots:

pernielsentikaer:
Can you approve @zkytech

zkytech:
> Can you approve @zkytech

approved

Sulfyderz:
@pernielsentikaer, is it normal that the update didn’t replace the older one in the store ? It created a new extension instead.

I think it’s due to the « author » swap.

PR #16172, opened 2025-01-05
Update brew extension

ViGeng:
## Description

Minor change on checkmark icons. Easier to recognize.

## Screencast

| Old                                                                                    | New                                                                                      |
|----------------------------------------------------------------------------------------|-------------------------------------------------------------------------------------------|
| image | image |
| image | image |

## Checklist


extensions/brew/src/components/list.tsx
```diff
@@ -59,14 +58,16 @@ export function FormulaListItem(props: {
   let version = formula.versions.stable;
   let tintColor = Color.SecondaryText;
```
nhojb, line 59: I wonder if we could use a lighter tint color here. For example this works quite nicely - it's closer to the tertiary text color use for the item's subtitle:

```
const tertiaryText: Color.Dynamic = {
  light: "#00000066",
  dark: "#FFFFFF66"
}
```

extensions/brew/src/components/list.tsx
```diff
@@ -59,14 +58,16 @@ export function FormulaListItem(props: {
...
 
   if (brewIsInstalled(formula)) {
     version = brewFormatVersion(formula);
     tintColor = formula.outdated ? Color.Red : Color.Green;
+    iconMark = Icon.CheckCircle;
     tooltip = formula.outdated ? "Outdated" : "Up to date";
   }
 
-  const icon = { source: Icon.Checkmark, tintColor };
+  const icon = { source: iconMark, tintColor };
```
nhojb, line 70: Old bug, but turns out we are missing the `tintColor:` label here.

extensions/brew/src/components/list.tsx
```diff
@@ -95,14 +96,16 @@ export function CaskListItem(props: {
   let version = cask.version;
   let tintColor = Color.SecondaryText;
```
nhojb, line 97: Use dynamic `tertiaryText` color here also.

extensions/brew/src/components/list.tsx
```diff
@@ -95,14 +96,16 @@ export function CaskListItem(props: {
...
 
   if (brewIsInstalled(cask)) {
     version = brewFormatVersion(cask);
     tintColor = cask.outdated ? Color.Red : Color.Green;
+    iconMark = Icon.CheckCircle;
     tooltip = cask.outdated ? "Outdated" : "Up to date";
   }
 
-  const icon = { source: Icon.Checkmark, tintColor };
+  const icon = { source: iconMark, tintColor };
```
nhojb, line 108: And missing the tintColor: label here.

nhojb (changes requested):
Thanks for the contribution 👏 

Generally the `Circle` and `CheckCircle` icons are clearer - especially for users with color vision deficiency.

Few comments:

1. I think we could use a lighter tint color for the default circle icon (see inline comment)

2. Could you update the screenshots to match the originals. Please use dark mode and the same wallpaper as previously used (red_distortion_4 IIRC). You can download the Raycast wallpapers [here](https://www.raycast.com/wallpapers).

Note you can use Raycast's Window Capture feature to set a custom wallpaper without changing your Desktop wallpaper.

Screenshot 2025-01-05 at 15 24 11

extensions/brew/src/components/list.tsx
```diff
@@ -57,16 +61,21 @@ export function FormulaListItem(props: {
...
+  let tintColor = null;
   let tooltip: string | undefined = undefined;
+  let iconMark: Icon = Icon.Circle;
 
   if (brewIsInstalled(formula)) {
     version = brewFormatVersion(formula);
     tintColor = formula.outdated ? Color.Red : Color.Green;
+    iconMark = Icon.CheckCircle;
     tooltip = formula.outdated ? "Outdated" : "Up to date";
+  } else {
```
ViGeng, line 73: They seem to be different types, so I use a branch (not elegant, though) to assign dynamic types. Let me know if it could be written better. (layman for JS here...)

ViGeng:
secondary
image

tertiary
image


https://github.com/user-attachments/assets/a55793a5-a4cf-43fd-ba8b-fe61d440b1ee


It shows that the tertiary color is slightly darker than the secondText.

ViGeng:
> Thanks for the contribution 👏
> 
> Generally the `Circle` and `CheckCircle` icons are clearer - especially for users with color vision deficiency.
> 
> Few comments:
> 
> 1. I think we could use a lighter tint color for the default circle icon (see inline comment)
> 2. Could you update the screenshots to match the originals. Please use dark mode and the same wallpaper as previously used (red_distortion_4 IIRC). You can download the Raycast wallpapers [here](https://www.raycast.com/wallpapers).
> 
> Note you can use Raycast's Window Capture feature to set a custom wallpaper without changing your Desktop wallpaper.
> 
> Screenshot 2025-01-05 at 15 24 11

Thanks for the screenshot (metadata) tips; they work pretty great!

extensions/brew/src/components/list.tsx
nhojb: You can initialize `tintColor` like this:

```
let tintColor: ColorLike = tertiaryText;
```

nhojb (changes requested):
Thanks for the changes 😄 

The screenshots still don't match the original ones. The first screenshot now shows the brew commands in development mode, which we don't want for the public store. Better to try and match the original screenshots, with the new icons of course.

extensions/brew/metadata/brew-6.png
ViGeng, line 1: It's wired that I can remove brew-6 and brew-7 though my local git has already removed them. Anyway, then I added 2 extra screenshots to override them. that seems reasonable. (I remember the Raycast doc says the max number is 7)

nhojb (approved):
Great, thanks for fixing up the screenshots 🙏

PR #16187, opened 2025-01-06
Add marcus extension

frilo-eth:
## Checklist


pernielsentikaer (changes requested):
Thanks for your PR 🔥

This extension feels a bit like a mix of other extensions in the store, like [SnapJot](https://www.raycast.com/k41531/snap-jot), [Year in Progress](https://www.raycast.com/thomas/year-in-progress) and one of the [Quote extensions](https://www.raycast.com/rrajath/stoicquotes)

What do you think?

pernielsentikaer:
I'm not sold on the idea since it just combines a lot of other tools and I'm having a hard time to see the value for users of Raycast and they might get confused.

We also states here that [extensions that is similar to what's already in the store might](https://manual.raycast.com/extensions) get rejected (since it's better to enhance)

Does that make sense?

PR #16214, opened 2025-01-08
Update mail extension

ViGeng:
## Description

- Add preference for `See Recent Mail`: `unreadonly` to show whether only unread messages.

The motivation is that:
When I used the `See Recent Mail` action, I thought it had bugs since I could not see any emails from my inbox.
When I checked the extension code, I finally realized this command (action) by default only shows unread emails while there is no convenient alternative command to check inbox. (`see mail accounts needs further steps, command -> choose account -> email list).

- [x] The most simple way is to add a preference which lets users decide what they want to show.
- [ ] There is another way of providing extra command (action) for checking inbox emails, which is to say, `Check inboxes.`

## Screencast

image

## Checklist

### DIFF
--- a/extensions/mail/CHANGELOG.md
+++ b/extensions/mail/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Mail Changelog
 
-## [Improvement] - {PR_MERGE_DATE}
+## [Improvement] - 2025-01-15
 
 - Add preference for `See Recent Mail`: `unreadonly` to show whether only unread messages.
 


ViGeng:
I also would ask for opinions regarding  the action names. `See` seems wired; `Check` would be better, I guess?
If this works for you, please let me know, and I will fix this together in this PR.
image

pernielsentikaer:
What do you think @yug2005

neo773:
I had the exact same thought when I was trying out this extension so I uninstalled it, so I agree with OP's perspective here.

pernielsentikaer:
Makes sense for me to use `Search`, `View`, or `Check`😊

yug2005 (approved):
I don't see the issue with "See" but I guess "View" would work better. Other than that, the change looks good to me.

ViGeng:
Okay, let's use `view` then. :)

extensions/mail/package.json
```diff
@@ -103,15 +103,15 @@
...
-      "name": "see-important-mail",
-      "title": "See Important Mail",
-      "description": "See messages in the important mailboxes.",
+      "name": "view-important-mail",
+      "title": "View Important Mail",
+      "description": "View messages in the important mailboxes.",
       "mode": "view"
     },
     {
-      "name": "see-mail-accounts",
```
pernielsentikaer, line 100: Can we keep the old `name`? It is the unique ID of the command where ranking, aliases, and hotkeys are saved. You can, of course, change the title 🙂

Same for all commands

ViGeng: Yes, it would be better if we kept it unchanged from the perspective of a user.
However, there should be some inconsistency then from the perspective of a developer.
This can be a disaster for the maintenance of the code.
In this case, I would prefer keeping the `title` and `name` unchanged, using `see.`

pernielsentikaer: That's up to you, we can't change the name just like that without ruin the flow for many users

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #16254, opened 2025-01-09
Update raycast-gemini extension

ViGeng:
## Description

- 🔄 Translate Selected Text: now you can set first and second target language.
  - All languages are by default translated to first target language.
  - First target language is translated to second target language.
  - Force Target Language overrides first target language.

## Checklist

### DIFF
--- a/extensions/raycast-gemini/CHANGELOG.md
+++ b/extensions/raycast-gemini/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Google Gemini Changelog
 
-## [Improved Translate Command] - {PR_MERGE_DATE}
+## [Improved Translate Command] - 2025-01-17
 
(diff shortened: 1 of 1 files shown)


pernielsentikaer (changes requested):
I think it's a good idea but I hate that users will lose their selected values because we're switching `name` values - we can't avoid that, right?

ViGeng:
> I think it's a good idea but I hate that users will lose their selected values because we're switching `name` values - we can't avoid that, right?

Technically we can avoid changing `name` though we will lose some readability of the code.
I can change back to the original `name`.

d0zingcat:
LGTM! 
Thanks for your contribution! @ViGeng

pernielsentikaer:
Hi 👋 

Looks good to me, approved 🔥

PR #16262, opened 2025-01-10
Update raycast-sink extension

ARui-tw:
## Description

Add quick create short link

## Checklist

### DIFF
--- a/extensions/raycast-sink/CHANGELOG.md
+++ b/extensions/raycast-sink/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Raycast Sink Changelog
 
+## [New Command] - {PR_MERGE_DATE}
+
+- Added a new command quick create short link with AI generated slug
+
 ## [Fix and New Features] - 2024-11-15
 
 - Added statistics to the detail view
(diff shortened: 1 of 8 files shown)


pernielsentikaer (changes requested):
Hi 👋

I gave it a spin and it copied https://sink.cool//raycast-ext as the link, but it's actually https://sink.cool/raycast-ext (with only one /) - can you look into a fix for that?

foru17:
Hi @ARui-tw,

Thank you for the submission. I noticed that you used a separate getPreferenceValues method to obtain the configuration. It would be better to expand the existing useConfig hooks for configuration reading to maintain consistency with the plugin's structure.

Additionally, could you please add the corresponding i18n fields for internationalization? This can also be automatically generated using AI.

ARui-tw:
>  I noticed that you used a separate getPreferenceValues method to obtain the configuration.

I did try to use the existing hook to read the config, but it seems that command in `no-view` mode can't use hook. Would you prefer if I slightly refactor the existing `useConfig`?

ARui-tw:
> Hi 👋
> 
> I gave it a spin and it copied https://sink.cool//raycast-ext as the link, but it's actually https://sink.cool/raycast-ext (with only one /) - can you look into a fix for that?

@pernielsentikaer Thanks for the review! I believe this is because the `host` value in the config is set to `https://sink.cool/`, but it should be `https://sink.cool`.

pernielsentikaer:
@foru17 can you check comments here

PR #16275, opened 2025-01-10
Add open-in-shopify-admin extension

Rafi993:
## Description

It will open your current store URL in Shopify admin for you to edit

## Screencast

https://github.com/user-attachments/assets/637304a5-a3f4-4106-8ae2-3bcdf3403777

## Checklist


extensions/open-in-shopify-admin/src/open-in-shopify-admin.ts
```diff
@@ -0,0 +1,33 @@
...
+    const url = await getBrowserURL(frontmostApp.name);
+
+    if (!url) {
+      await showToast({ title: "You don't have any website open", style: Toast.Style.Failure });
+      return;
+    }
+
+    const urlParts = url.split("/");
+
+    if (urlParts.length < 2) {
```
pernielsentikaer, line 20: Can shopify URLs only have a specific length?

extensions/open-in-shopify-admin/src/open-in-shopify-admin.ts
```diff
@@ -0,0 +1,33 @@
+import { closeMainWindow, getFrontmostApplication, showToast, Toast } from "@raycast/api";
+import { getBrowserURL, openURL } from "./utils";
+
+const command = async () => {
+  await closeMainWindow();
+  await showToast({ title: "Opening in Shopify admin...", style: Toast.Style.Animated });
+
+  const frontmostApp = await getFrontmostApplication();
+
+  if (frontmostApp.bundleId !== "com.apple.finder") {
```
pernielsentikaer, line 10: Is this the best implementation? Looks a bit funny

extensions/open-in-shopify-admin/src/open-in-shopify-admin.ts
```diff
@@ -0,0 +1,33 @@
...
+    const url = await getBrowserURL(frontmostApp.name);
+
+    if (!url) {
+      await showToast({ title: "You don't have any website open", style: Toast.Style.Failure });
+      return;
+    }
+
+    const urlParts = url.split("/");
```
pernielsentikaer, line 18: Can't we use `new URL` for better and smarter parsing?

Rafi993: Fixed it now

extensions/open-in-shopify-admin/src/open-in-shopify-admin.ts
Rafi993: You are right. This check is not reliable. I have removed it.

extensions/open-in-shopify-admin/src/open-in-shopify-admin.ts
Rafi993: Actually this check was redundant and not needed since we do check if browser URL is there and show a error message if the user tries to run the extension on a normal window

Rafi993: https://github.com/user-attachments/assets/2b2acdb5-a6e0-4516-a549-33a1fad239a0

pernielsentikaer:
You can maybe also do something like this for the url: https://github.com/raycast/extensions/blob/a5923b9e258398c002381c308a7d374e312b7c9e/extensions/whois/src/utils/index.ts#L40

Rafi993:
> You can maybe also do something like this for the url:
> 
> https://github.com/raycast/extensions/blob/a5923b9e258398c002381c308a7d374e312b7c9e/extensions/whois/src/utils/index.ts#L40

I actually tried that method but it doesn't work with Brave browser so will need to add other browsers like Zen, Arc or other browsers that comes up explicitly to the list.

pernielsentikaer:
Brave(Firefox) is just working different than all other browsers which is super anoying - it doesn't feels like a good solution to copy the URL to clipboard if it's possible not to. Maybe we could do someyhing like https://gist.github.com/prashanthrajagopal/08ab39d62725c8a8716b but add Brave, Zen and other branches of Firefox?

Rafi993:
> Brave(Firefox) is just working different than all other browsers which is super anoying - it doesn't feels like a good solution to copy the URL to clipboard if it's possible not to. Maybe we could do someyhing like https://gist.github.com/prashanthrajagopal/08ab39d62725c8a8716b but add Brave, Zen and other branches of Firefox?

Awesome let me check

Rafi993:
> Brave(Firefox) is just working different than all other browsers which is super anoying - it doesn't feels like a good solution to copy the URL to clipboard if it's possible not to. Maybe we could do someyhing like https://gist.github.com/prashanthrajagopal/08ab39d62725c8a8716b but add Brave, Zen and other branches of Firefox?

I have done the change. Thanks for the quick feedback.

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #16287, opened 2025-01-11
Update google-chrome extension

rsperezn:
## Checklist

### DIFF
--- a/extensions/google-chrome/CHANGELOG.md
+++ b/extensions/google-chrome/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Google Chrome Changelog
 
-## [Allow to configure profile path] - {PR_MERGE_DATE}
+## [Allow to configure profile path] - 2025-01-18
 - The path for your profile can be configured in the settings. This allows you to use a different profile than the default one.
 
 ## [Fix] - 2025-01-09


rsperezn:
This allows us to configure the Chrome profile path rather than hardcoding it

rsperezn:
image

a-laughlin:
Thanks for submitting a new PR. Taking a look now.

extensions/google-chrome/package.json
```diff
@@ -63,6 +64,16 @@
...
+    {
+      "description": "Find the full path to the Google Chrome Profile path via chrome://version/. If empty, the default profile will be used.",
+      "label": "Profile",
+      "default": "",
+      "name": "profilePath",
+      "required": false,
+      "title": "Profile Path",
+      "type": "textfield",
+      "placeholder": "Application Support/Google/Chrome"
+    },
```
a-laughlin, line 76: The description says the full file path. The placeholder has a partial path. I feel a little confused about that and guess that others will too. Would it make sense for them to match?

extensions/google-chrome/src/util/index.ts
```diff
@@ -8,11 +10,21 @@ const userLibraryDirectoryPath = () => {
...
+  if (profilePath) {
+    return path.join(profilePath, fileName);
+  }
+  return path.join(
+    userLibraryDirectoryPath(),
+    ...defaultChromeProfilePath,
+    profile ?? DEFAULT_CHROME_PROFILE_ID,
+    fileName
+  );
+};
```
a-laughlin, line 24: Many bug reports come from an incorrect profile location. Checking whether the file exists here, and explaining what to do if it doesn't, would help prevent those reports.

extensions/google-chrome/package.json
rsperezn: Good catch, will fix it soon

extensions/google-chrome/src/util/index.ts
```diff
@@ -1,26 +1,36 @@
...
+    resolvedProfilePath= path.join(
     userLibraryDirectoryPath(),
     ...defaultChromeProfilePath,
     profile ?? DEFAULT_CHROME_PROFILE_ID,
     fileName
-  );
+  );}
+
+  if (!fs.existsSync(resolvedProfilePath)) {
+    throw new Error(`The path ${resolvedProfilePath} does not exist. Please check your Chrome profile location by visiting chrome://version -> Profile Path.`);
```
a-laughlin, line 30: Looks good.  Just needs prettier run in order to pass the tests.

One small addition to the message. Those experiencing the error won't always know that profile customization exists now (yay!), so a hint pointing them to the setting will likely help them resolve the problem efficiently.


```suggestion
    throw new Error(`The path ${resolvedProfilePath} does not exist. Please check your Chrome profile location by visiting chrome://version -> Profile Path. Then update it in Extension Settings -> Profile Path.`);
```

a-laughlin:
Great! Looks good to merge.

FYI I got curious and looked at what happens to that error once it's thrown.  It appears to be ignored (useBookmarksSearch) and unhandled (useHistorySearch).

To avoid blocking this PR, I created a [separate issue](#16334) for updating error handling in useHistorySearch and useBookmarksSearch.

In the unhandled `useHistorySearch` case, the error screen should show the text from the thrown error, which is more useful and accurate than the `Google Chrome not installed` that currently appears.

rsperezn:
yeah @a-laughlin I guess we can have a new erroView for it, great call and thanks for the patience reviewing the PR

extensions/google-chrome/package.json
```diff
@@ -63,6 +64,16 @@
...
   ],
   "preferences": [
+    {
+      "description": "Find the full path to the Google Chrome Profile path via chrome://version/. If empty, the default profile will be used.",
+      "label": "Profile",
+      "default": "",
+      "name": "profilePath",
+      "required": false,
+      "title": "Profile Path",
+      "type": "textfield",
```
pernielsentikaer, line 74: Why not use `directory` as type?

rsperezn: TIL :)

pernielsentikaer:
What do you think @a-laughlin - use the type `directory` or use what's there today, a textfield?

I'm unsure what's best and easiest in this case

a-laughlin:
What are the effects of setting the type to `directory` instead of `textfield`? The documentation doesn't specify how they differ.

a-laughlin:
I have a bunch going on this week. Limited time to test. Assuming `directory` does something like a file chooser, I'd go with the text field. The chrome profile screen gives a path to copy. That's easier to insert in a text field than going through a file chooser.

rsperezn:
yeah it's a file chooser @a-laughlin . We should just go with something to start with. The latest commit change it to `directory` type

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #16294, opened 2025-01-11
[Apple Reminders] Include overdue list view and default due date for new reminders

maxnyby:
## Description

I added two things:
- A new optional list view that only shows overdue todos
- An option to have new reminders default to "today".

The idea is that I use reminders to manage both work and personal life, and it helps me stay on task if the reminders for stuff I need to do after work don't show up until after work.
And when I default to only show overdue reminders, it's helpful to have newly created reminders default to have a due date, so I don't forget them.

I figured if that works for me, maybe others would like to try that workflow as well. But I made the features optional, since I guess everybody don't work the same way.
I hope you will accept these features, because they are really helpful for me.

And I'm sorry if I made a mess of this pull request. It seems some changes made it in between me forking the extension and publishing it again, and I'm not very well versed in git.

I would also have liked to use `Icon.CheckList` in `src/my-reminders.tsx:48` (as shown in https://developers.raycast.com/api-reference/user-interface/icons-and-images#:~:text=CheckCircle-,CheckList,-CheckRosette), and it worked fine during tests, but the ray-linter told me it wasn't defined and the type definition also seems to be missing. I don't see a way to add the icon to the type definition to satisfy the linter, but if you know how, please feel free to swap out `Icon.List` for `Icon.CheckList` :-)

## Checklist

### DIFF
--- a/extensions/apple-reminders/CHANGELOG.md
+++ b/extensions/apple-reminders/CHANGELOG.md
@@ -1,7 +1,6 @@
 # Apple Reminders Changelog
 
-## [Add "Overdue" view and default date setting] - {PR_MERGE_DATE}
-
+## [Add "Overdue" view and default date setting] - 2025-02-10
 - A new "Overdue" view has been added, displaying only reminders that are overdue.
 - There's now an option to automatically set the default date of new reminders to the current day.
 


extensions/apple-reminders/swift/AppleReminders/Sources/Reminders.swift
michalzuch, line 1: This change breaks weekdays and weekends interval introduced in previous extension updates.

extensions/apple-reminders/src/menu-bar.tsx
michalzuch, line 1: Why was sorting removed? It doesn't seem related to the feature you are introducing.

extensions/apple-reminders/src/helpers.ts
```diff
@@ -15,7 +15,13 @@ export function getDateString(date: string) {
 export function getTodayInLocalTime() {
   return formatISO(new Date(), { representation: "date" });
 }
+export function getTimeInLocalTime() {
```
michalzuch, line 18: `getTodayInLocalTime()` implemented above does the same thing, thing function is obsolete

extensions/apple-reminders/src/helpers.ts
```diff
@@ -15,7 +15,13 @@ export function getDateString(date: string) {
...
+  return formatISO(new Date());
+}
 
+export function isPast(date: string) {
```
michalzuch, line 22: `isOverdue()` implements the same thing, this function is obsolete

michalzuch (changes requested):

michalzuch:
Hey 👋  
Thanks for the PR! Overall, the new changes seem fine to me. However, there are a few things that need to be fixed. Some of them break current functionalities, while others will help maintain cleaner code. I left all the details in the review notes.

michalzuch (approved):
Looks good to me 🙌

thomaslombart (changes requested):
Setting today as the default date looks good to me. 

However, I'm uncertain about the "Todo" view. I would rename it to "Overdue" and avoid displaying tasks that are due today to reduce confusion for users.

We aim to keep extensions flexible and aligned with the core features of their native apps. That’s why we focus on achieving feature parity rather than adding specific custom workflows. We totally get that certain features might work better for your unique setup, but adding them could make the extension more complicated, harder to maintain, and less user-friendly for everyone ☺️

thomaslombart (commented):
Thanks for the changes. I've noticed that there's a new preference to display full-day reminders due for today in the `Overdue` view, which looks more like a "hack" to fit the workflow (unless I didn't understand the use case).

As I mentioned in my previous message, we're not really keen on adding custom workflows since this can turn into a mess in the long run. Could you remove the optional setting so that the overdue list actually only displays overdue reminders? 🙂

maxnyby:
You're right in a way. My personal understanding of "overdue" reminders include fullday reminders that are due today, and that understanding also fits my workflow well.
But I get that other people don't see fullday reminders as "overdue" until that entire day is in the past. I was hoping that by adding a preference I could accommodate both understandings, and thereby also more workflows. But I also understand the desire for lean software that doesn't turn into spaghetti code.

Would it be better if I remove the "Overdue" view completely and set out to see if I can find a solution to display smart lists instead?

thomaslombart:
Yeah, unfortunately, the Apple Reminders API doesn't support many features, including smart lists 😔

Let's remove the "overdue" preference to keep the extension clean from a DX and UX point of view. Thanks a lot for understanding!

maxnyby:
No problem. Thank you for making it easy to run a modified version locally ;-)

I've removed the preference to show fullDay reminders as overdue now. If you want me to remove the "Overdue" list altogether, just say so.

Completely unrelated, I stumbled on the error `Property 'CheckList' does not exist on type 'typeof Icon'.`
It seems that the CheckList icon is defined (and shown here https://developers.raycast.com/api-reference/user-interface/icons-and-images), but it's missing in the typescript definition in `raycast/api/types/index.d.ts`, giving a linter error when using it, even though it works as intended. Just thought I'd mention it.

thomaslombart (approved):
Thanks for the changes. I made minor updates, including the `CheckList` icon due to the extension using an outdated `@raycast/api` version. All set!

PR #16332, opened 2025-01-13
Add validatorjs-functions extension

djankies:
## Checklist


pernielsentikaer (changes requested):
Is there no way we can skip the GitHub integration for this? I'm not sure what it complete does

djankies:
Hi @pernielsentikaer! That would be possible by cloning the functions and manually updating the extension as ValidatoJS is updated. The GitHub integration allowed the extension to pull validator functions directly from the repo to avoid that. Would the manual route be preferable to you?

pernielsentikaer:
@xmok, have you worked with this in the past? Adding a GitHub key to the search functionality is, in my opinion, a poor user experience 😊

xmok:
Yup I agree this doesn't seem like the best approach!
@djankies I've contributed to a number of docs extensions and there are a couple of ways you can do it:

A
as you suggested: manually clone all the files and show to the user.
Pro: also available offline
Con: need to update manually

B
do something similar to the [Next.js Documentation](https://www.raycast.com/PrasadChinwal/nextjs-docs) extension:
step 1: we use OctoKit to fetch pages from the repo
step 2: when user opens a page, it is downloaded and saved locally to user's system
Pro: also available offline (and no need for PAT)
Con: need to first download a file/page

C (not recommended)
do something similar to the [Axios Docs](https://www.raycast.com/denizyoldas/axios-docs) extension:
show all the links/pages, but open them directly in browser
Pro: takes up less space
Con: not really a good experience

D (not recommended)
keep the current method of entering a GitHub PAT (this is also done by [1 LOC - JavaScript Utilities in Single Line of Code](https://www.raycast.com/kud/1loc) but not the best for user

E (probably not relevant)
if docs are available in Algolia, fetch using Algolia e.g. as in [Alpine.js](https://www.raycast.com/aurawindsurfing/alpinejs)
if docs are available via a package, fetch using that e.g. as in [NSIS Reference](https://www.raycast.com/idleberg/nsis-reference)

You could go with option A, option B, or a combo e.g. you can choose to fetch docs on-the-fly but implement caching to ensure new data is only fetched when some time has passed which will make sure you don't hit rate limits.

Hope this helps!

PR #16360, opened 2025-01-14
Add memorkeys extension

3HertzLFO:
## Description

This is a quick reference for raycast commands that have hotkeys assigned to them. Forgot how to snap the window to the left half just open Memorkeys and search for left half. If you are using an application and need to see your hotkeys for that app, open Memorkeys and press cmd + return to only show the shortcuts for that app. This is accomplished by exporting only the settings check box from the advanced page with no password and load it into Memorkeys to format. There is also an inspect page that gives you insight into how ergonomic your shortcut is.

## Screencast

memorkeys-1
memorkeys-2
memorkeys-3
memorkeys-4
memorkeys-5
memorkeys-6

## Checklist

### DIFF
--- a/extensions/memorkeys/src/components/ShortcutListItem.tsx
+++ b/extensions/memorkeys/src/components/ShortcutListItem.tsx
@@ -1,4 +1,13 @@
-import { List, Icon, ActionPanel, Action, closeMainWindow, popToRoot, getPreferenceValues } from "@raycast/api";
+import {
+  List,
+  Icon,
+  ActionPanel,
+  Action,
+  closeMainWindow,
+  popToRoot,
+  getPreferenceValues,
+  environment,
+} from "@raycast/api";
 import { exec } from "child_process";
 import { promisify } from "util";
 import path from "path";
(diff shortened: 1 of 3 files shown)


pernielsentikaer:
Hi 👋

Is it possible to export `rayconfig` without a password, if so, how do you do that?

3HertzLFO:
Yes! You can export without a password if you export only the settings and skip the password page (see pics). Save the `rayconfig` anywhere and load it into this extension which will parse it into a json and use the json file to display the shortcuts. You can also edit that parsed json file if there are any errors. The only downside to the current method is you must export and upload a new `rayconfig` anytime you change or update hotkeys.

Raycast 2025-01-22 at 00 29 22
Raycast 2025-01-22 at 00 29 45

pernielsentikaer:
I just see this when I try to work with the extension. Do you have any clue why? I tried to uninstall and delete caches, but it seems like I can't get it to stop loading


https://github.com/user-attachments/assets/f2cbd820-0215-4f84-a967-64b8649c7584

3HertzLFO:
Sorry about that. I think the issue is with the way i was handling file paths, i have changed to use the application support path provided by raycast. Other than that I'm at a bit of a lose for why it might be stuck loading. Let me know if its still stuck and if you have any ideas for things I could check.

pernielsentikaer:
I'll check this again 🙂

pernielsentikaer (changes requested):
Don't you think we should shorten the accessories to something like this?

CleanShot 2025-02-13 at 14 23 27@2x

- I have one element that maps `⌘$` - it just returns ⌘ UNKNOWN

PR #16364, opened 2025-01-14
Update imessage-2fa extension to add support for mail 

taschaub:
### Overview
This PR extends the iMessage 2FA code scanner to support email-based 2FA codes (Apple Mail or Gmail) while maintaining the option to use iMessage-only functionality. Users can now choose their preferred source for 2FA codes.

### New Features
- Flexible source selection:
  - iMessage only (original functionality, no Mail app interference)
  - Email only (Apple Mail or Gmail)
  - Both sources combined
- Email provider options:
  - Apple Mail integration
  - Gmail integration via OAuth (no need for Mail app to be running)
- Detailed Gmail OAuth setup guide in README
- Real-time code detection from all enabled sources

### User Experience
- Users can avoid Mail app background processes by selecting iMessage-only mode OR
- Gmail integration provides a Mail app alternative
- Automatic source filtering based on user preferences

### Testing Notes
- Test source selection (iMessage only, Email only, both)
- Verify Gmail OAuth setup process using README instructions
- Check initial Mail.app automation permission flow (if using Apple Mail)

## Checklist

### DIFF
--- a/extensions/imessage-2fa/CHANGELOG.md
+++ b/extensions/imessage-2fa/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Messages 2FA Changelog
 
-## [Update] - {PR_MERGE_DATE}
+## [Update] - 2025-02-13
 
 - Add support for Email 2FA codes alongside iMessage
 - Add ability to switch between iMessage, Email, or both sources


pernielsentikaer:
Hi 👋

- Is there a way to not have this, I don't use mail.app and would be annoyed about having it to pop up all the time

@yuercl do you want to test this?

taschaub:
@pernielsentikaer yes, in the settings you can select which source you want to enable as inputs. When you select "imessage only" the mail.app wont open and it will basically behave as it did before fetching only the codes from imessage. 

a72c742bfaf274778708a35cb982b89cc1d59d6d2241a847b54ebff7c67c8762

taschaub:
@pernielsentikaer @yuercl I actually added now the Gmail OAuth functionality following up on your comment @pernielsentikaer So that now you can also use this with the Gmail API and not have to have the mail.app running in the background all the time. I also copied that tutorial on how to setup the OAuth from the gmail extension into this Readme. You still have the option to use the Apple Mail, if you rather do that, or for example have another email client. And as I said, if you don't want any of the mail functionality, you can still set in the settings to use iMessage as the single source.

I think it really would be great to have this together in one single extension🙏

pernielsentikaer:
We shouldn't change the default value, iMessage should be the default 🙂

Awaiting what @yuercl thinks about it

taschaub:
@pernielsentikaer alright, I changed the default back to iMessage 👍

yuercl:
Big thanks to @taschaub , this is a great feature. I tested the newly submitted functions in this PR, including Apple Main and Gmail, and I can get the verification code. Currently, in the latest version submitted by @taschaub, the mode of using Mail has been made optional.

yuercl:
In the initially envisioned scenario, there was a dropdown filter that allowed users to select between viewing all messages (ALL) or only 2FA (code) messages. The idea was to account for the possibility that some non-standard format verification codes might not be extractable, and I wanted to ensure they could still be displayed. However, it seems that a past change may have caused the ALL filter to stop functioning. In practice, I rarely encountered the feature that allowed querying all messages. I'm very curious to hear your thoughts on this, especially since I've noticed that this plugin has users from various countries. @taschaub @pernielsentikaer

taschaub:
@yuercl Yes, I could never use this feature of seeing ALL messages. I never needed to really, but also the feature didn't work for me. So it didn't show me all my messages. I personally would prefer to keep it clean and leave this feature out, as I really also think that we detect most of the codes anyways.

yuercl (approved):
Nice commit

taschaub:
@yuercl thanks for approving! @pernielsentikaer @thomaspaulmann do I need to do anything else so that the PR will be merged?

pernielsentikaer:
It looks like you have a folder called `imessage-2fa` inside `imessage-2fa` - could you have a look? 🙂

taschaub:
@pernielsentikaer good catch! I removed it and it should be fine now! 🙂

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #16410, opened 2025-01-16
Add Capmo Extension

d0rN834:
## Description

Easily manage and interact with projects and tickets from Capmo directly in Raycast. List tickets, update their status, filter projects, and more – all with quick keystrokes.

## Screencast

Please see screenshots in Readme.

## Checklist


pernielsentikaer:
Could you make a screencast of the flow so I can see how it works 😊

d0rN834:
[Screencast](https://gifyu.com/image/SeinI

extensions/capmo/src/list-projects.tsx
```diff
@@ -0,0 +1,86 @@
...
+        "An unknown error occurred.";
+      showToast({
+        style: Toast.Style.Failure,
+        title: "Error Fetching Projects",
+        message: errorMessage,
+      });
+    } finally {
+      setIsLoading(false);
+    }
+  };
```
pernielsentikaer, line 65: In all the files, should we then use `useFetch` from `@raycast/utils` instead which also caches the data and give you isLoading too

pernielsentikaer (changes requested):
Thanks for the screencast. Could you look at this?

pernielsentikaer:
Then what about using `useCachedPromise` instead, then it's caching the data while getting new

d0rN834:
@pernielsentikaer : I don't know whats going on. I've adjusted the commands to use a cached approach. I don't now, why the checks keep failing after re-uploading the commands. Prettier seems to be fine on my end. Do you have an idea?

```
capmo_api_r1 % ray lint --fix
ready  - validate package.json file
ready  - validate extension icons
ready  - run ESLint
ready  - run Prettier 3.4.2
```

pernielsentikaer:
Should be better now @d0rN834 - there are some german actions, could you correct them 😊

d0rN834:
@pernielsentikaer: Thanks for the support. Capmo is a german software altogether. Question would be, if all commands could be in german and the readme is in 2 languages. If necessary, i will translate everything to english

pernielsentikaer:
Below is taken from the developer docs:

> At the moment, Raycast doesn't support localization and only supports US English. Therefore, please avoid introducing your custom way to localize your extension. If the locale might affect functionality (e.g., using the correct unit of measurement), please use the preferences API.

So we need to have it in English 😊

PR #16413, opened 2025-01-16
Update tmux-sessioner extension

janwvjaarsveld:
## Description

Add support for rename window in a Tmux session.
- This is done by modiying the current `RenameTmuxSession.tsx` (renamed to `RenameTmux.tsx`) to handle both session and window rename and share the similarities.

## Screencast

image
image
image

## Checklist

### DIFF
--- a/extensions/tmux-sessioner/CHANGELOG.md
+++ b/extensions/tmux-sessioner/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Tmux Sessioner Changelog
 
-## [Add support for renaming tmux windows] - {PR_MERGE_DATE}
+## [Add support for renaming tmux windows] - 2025-01-19
 
 - Created a new Component to handle both renaming windows and sessions
 - Added a new command to rename windows, with shortcut `⌥⌘R`


louishuyng (approved):
LGTM 🚀

extensions/tmux-sessioner/src/index.tsx
```diff
@@ -78,7 +78,14 @@ export default function Command() {
...
+                    push(
+                      <RenameTmux
+                        sessionName={session}
+                        windowName=""
+                        type="Session"
+                        callback={() => setupListSesssions()}
+                      />,
+                    );
                   }}
                   shortcut={{ modifiers: ["cmd"], key: "r" }}
```
janwvjaarsveld, line 90: @louishuyng I see that the shortcut used here is `cmd + r`, but for delete it is `ctrl + x`. I went with the same shortcut for the window renaming, but would it not make sense to use the same modifier for all actions?

louishuyng: Yes i agree with @jasonshanks

janwvjaarsveld: I agree with consistency as well, but for one of the shortcuts it will be a breaking change as Rename session already exists with `cmd`. 
Which modifier do you prefer between `cmd` and `ctrl`?

louishuyng: Hi @janwvjaarsveld 

i am little bit confused about grouping Rename session and Rename window in the same component. 

Currently, we have `Mange Tmux Windows` and `Manage Tmux Session` commands. So i belive we can have separte rename function for each command and can use the same shortcut `cmd + r`

To summary
`Mange Tmux Windows` command will have shortcut `cmd + r` to rename selected window
`Mange Tmux Session` command will have shortcut `cmd + r` to rename selected session

janwvjaarsveld: Hi @louishuyng the grouping of the rename component is only because of all the similarities window and session renaming share and it feels strange to have so much duplication of essentially the same thing. 

But that does not change anything for the modifier discussion. My question is more with regard to the different modifiers that the current rename session (`cmd`) and delete session (`ctrl`) uses and by extension the rename window (`cmd`) and delete window (`ctrl`). 
It would be nice to align these two modifiers to be the same, so you have the same experience across the extension.

louishuyng: I see sorry for my mislead on this @janwvjaarsveld. For me **cmd** modifier is more popular for mac user. So i prefer all shortcut should use cmd modifer for consistency

What do you think?

louishuyng: I see sorry for my mislead on this @janwvjaarsveld. For me **cmd** modifier is more popular for mac user. So i prefer all shortcut should use cmd modifer for consistency

What do you think?

janwvjaarsveld: I agree with cmd, it feels more natura. We can even introduce a preference setting for users to configure this. 
Do you want me to do this in a seperate PR so it does not get lost in this PR's changes, or is a commit here fine?

louishuyng: I think it fine to put commit here if you want to change cmd for all actions

But if you want to introduce new preference to change between cmd and ctrl it better to have separate PR

janwvjaarsveld: I did a commit to align the modifiers, then we can later introduce the new preference

louishuyng (approved):

jasonshanks:
Consistency with the general Raycast trend is preferable where possible I think. But the only reference I could find to another extension Renaming is:
- Raycast AI Chat: `⌥⌘R` is what they use
- `⌘R` is used in many extensions for 'Refresh' (Raycast Apple Notes, Raycast Reminders)

So `⌥⌘R` would actually be more in line. But certainly not `⌃R`. Can certainly do your own thing though, so if there is not possibility of there being a 'Refresh' command in future, the existing '⌘R' is fine.

extensions/tmux-sessioner/src/index.tsx
jasonshanks: Consistency with the general Raycast trend is preferable where possible I think. But the only reference I could find to another extension Renaming is:

Raycast AI Chat: ⌥⌘R is what they use
⌘R is used in many extensions for 'Refresh' (Raycast Apple Notes, Raycast Reminders)
So ⌥⌘R would actually be more in line. But certainly not ⌃R. Can certainly do your own thing though, so if there is not possibility of there being a 'Refresh' command in future, the existing '⌘R' is fine.

janwvjaarsveld: @jasonshanks I did see that during development ⌘R is mapped to debug reloading (you can see in first screenshot), so maybe going with ⌥⌘R to align with trends in other extensions makes more sense?

louishuyng: I just read the doc and feel like we have two common shortcut can be applied to rename

ctrl + r
cmd + shift + r

Or we can use different word E instead of R
cmd + e

extensions/tmux-sessioner/src/index.tsx
```diff
@@ -87,7 +94,7 @@ export default function Command() {
                   onAction={() =>
                     deleteSession(session, setIsLoading, () => setSessions(sessions.filter((s) => s !== session)))
                   }
-                  shortcut={{ modifiers: ["ctrl"], key: "x" }}
```
jasonshanks, line 90: We recently changed Delete to `ctrl x` to be inline with Raycast standards

extensions/tmux-sessioner/src/index.tsx
louishuyng: Ah yes i refer from the doc seems like you r correct they are missing rename

I will follow the popularity. So i am ok with ⌥⌘R

janwvjaarsveld: @louishuyng and @jasonshanks PR has been updated with ⌥⌘R for rename and ⌥⌘X for delete

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #16458, opened 2025-01-18
feat: split AIs into separate commands

iKasch:
### Split AIs into commands

For maintainability and simplicity reasons, I created three separate commands for each AI available. The previous command will break, but is not removed, it will prompt the user to use a different command. (Will delete the note command at some point).

### Add Clipboard & Browser Extension detection

The Extensions can now detect active youtube tabs and youtube links in the clipboard.

## Screencast

CleanShot 2025-01-18 at 22 07 01

## Checklist

### DIFF
--- a/extensions/summarize-youtube-video-with-ai/CHANGELOG.md
+++ b/extensions/summarize-youtube-video-with-ai/CHANGELOG.md
@@ -5,8 +5,6 @@
 - 🔎 Added support for clipboard and browser extension detection
 - 🪓 Split AIs into separate commands
 
-NOTE: This update may requires you to update your preferences for the extension. Consider using one of the new commands from now on.
-
 ## [New Features] - 2024-12-21
 
 - Added preferences to set a custom OpenAI API endpoint


pernielsentikaer (changes requested):
This will break the flow for so many users. Can't we keep the old way? I can see why splitting might be easier, but it should also be possible to split without the big change, or?

iKasch:
@pernielsentikaer I was thinking about it, but I'm not sure what the best way to introduce separate commands would be.
Is there a way to migrate the settings to the respective new commands? Or can I detect a specific setting and launch another command from it?

pernielsentikaer:
You could use getpreference and [launchCommand](https://developers.raycast.com/api-reference/command#launchcommand) maybe?

iKasch:
Perfect, that makes sense. I'll check it out and adjust the PR. Thank you! 🙇

iKasch:
@pernielsentikaer Hey! I added the old preferences back and handle now the new commands, accordingly to the preferences set by the user. This will still break for anything other then Raycast as AI provider since users need to add their API keys again. 

What do you think?

pernielsentikaer:
That's much better, let's do that 😊

extensions/summarize-youtube-video-with-ai/src/summarizeVideo.tsx
```diff
@@ -35,114 +23,57 @@ export type Preferences = {
...
-          style: Toast.Style.Failure,
-          title: ALERT.title,
-          message: "Error fetching video data: " + error.message,
-        });
+  switch (chosenAi) {
+    case "anthropic":
+      showToast({
+        style: Toast.Style.Failure,
+        title: "Command has changed",
+        message: "Please run `summarizeVideoWithAnthropic` or enter your anthropic API key above",
```
pernielsentikaer, line 39: Maybe we should show the commands title in all the cases instead of the name?

iKasch: Fair point, I'll adjust 😊

pernielsentikaer:
The more I think about it, the more I'm getting second doubts - I can see that it's easier to maintain, but can't we simply load the different parts from a utils or models file? I'm not sure about anything right now, just want to give the user the best experience

iKasch:
I agree, it's not the best situation, but I think the changes provide more benefits then just maintainability. For example, I could easily add new commands in the future if needed without breaking the experience and cluttering the code more.
To me the current solution is also temporary and at some point I'd like to remove the old default command. It was not ideal to  write the extension like this in the first place, but that's on me. 

I'm open for ideas and to discuss further though 😊

pernielsentikaer:
Let's just break the old command in this case—can we do that? Let's just go ahead and remove the old command and use the new ones if it helps you.

iKasch:
So you'd suggest to remove the old command completely (or like it was before, with the note) and go full on three new commands?

pernielsentikaer:
Yes, it feels too confusing to have the new "deprecated" command - sometimes we might need to break it for a few moments to get on the right course. Sorry for the confusion; I'm also a bit split myself 😊

iKasch:
Haha no worries, I was also not sure and I liked the discussion! I'll prep everything and lets you know when it's ready ✌️

iKasch:
@pernielsentikaer Done 😊  let's see how this goes ✌️

pernielsentikaer:
Let's do it!

PR #16480, opened 2025-01-19
Add generate-beautiful-qr extension

na-bal:
## Checklist


na-bal:
Hey!
Do I have to correct an error that occurred before request approval?

pernielsentikaer:
Hi 👋

What about enhancing the [one we have in the store](https://www.raycast.com/Melvynx/qrcode-generator) by @Melvynx

na-bal:
> Hi 👋
> 
> 
> 
> What about enhancing the [one we have in the store](https://www.raycast.com/Melvynx/qrcode-generator) by @Melvynx

Hi☺️!

Yes, it's a good extension. I used it at first and that's why I wanted to make my own version. 
There is no way to customize the cr-code, the output is only in png and I wanted to improve the user experience.

pernielsentikaer:
We're trying to keep similar extensions to a minimum. What about enhancing the one from @Melvynx? Are you okay with that, @Melvynx? 😊

na-bal:
> We're trying to keep similar extensions to a minimum. What about enhancing the one from @Melvynx? Are you okay with that, @Melvynx? 😊

@pernielsentikaer Hi again :) 

What do we do if we don't hear from @Melvynx? 

I'm worried that in this case we have different ways to generate qr code. And in case of merging the author will have to reflash the whole extension. 

In any case, I would be happy and ready to add png output. I output only svg in my extension

pernielsentikaer:
I think it could be nice the enhance the current with new functionality, are you up for that?

CC @Melvynx

na-bal:
@pernielsentikaer Yes, let’s do it this way, but I’ll have to completely replace the code with mine. Am I understanding correctly?

Also, I’m not quite sure (I have no experience with this), can we implement it without the author?

Melvynx:
Hey, sorry. Yes, it will be so nice to improve the extensions. You can merge and update the current extensions. Just add a setting to avoid making it mandatory so we can keep it minimalistic.

pernielsentikaer:
Thanks for checking in @Melvynx 🔥

na-bal:
> Hey, sorry. Yes, it will be so nice to improve the extensions. You can merge and update the current extensions. Just add a setting to avoid making it mandatory so we can keep it minimalistic.

@Melvynx Hi:)
To not make it mandatory to do what, not quite sure?

Melvynx:
Not mandary to have this fancy stuff. Like default, minimalistic styles.

PR #16485, opened 2025-01-19
[Messages] Filter out phone numbers from OTP codes

haydencbarnes:
## Description

Added a number filter for OTP Code messages that include phone/cellular numbers that should not be used as OTP Codes for this command.

The following example of a Sofi OTP would pull in the last four digits of the phone number, now it pulls in the correct OTP code:

CleanShot 2025-01-19 at 14 53 51@2x

## Checklist

### DIFF
--- a/extensions/messages/CHANGELOG.md
+++ b/extensions/messages/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Messages Changelog
 
-## [Fix `Paste Latest OTP Code` Command for Phone #s] - 2025-01-19
+## [Fix `Paste Latest OTP Code` Command for Phone numbers] - 2025-02-10
 
 Added a number filter for OTP Code messages that include phone/cellular numbers that should not be used as OTP Codes for this command.
 


thomaslombart (commented):
Thanks for the PR! I've stumbled upon some results like that as well. But instead of filtering out phone numbers, I think we should have a more precise way of identifying OTP codes and modify that regex instead.

haydencbarnes:
Hi @thomaslombart,

Thanks for your feedback/review!

While I would agree in-general when it comes to a regex issue, I thought about this a little before submitting this PR and I believe filtering out phone numbers is the more effective approach because OTPs lack standardization, making it hard to craft a regex that accurately captures all without false positives. While refining the regex could help, it may introduce unnecessary complexity and maintenance overhead. Filtering phone numbers which have some standards, while not totally fool proof either, provides a practical and scalable solution for now.

thomaslombart:
But wouldn't it be better in that case to modify the base regex to limit it to numeric sequences of length 4–8? That would filter phone numbers without too much complexity.

haydencbarnes:
@thomaslombart My concern with that, would be we would miss out on codes like:
CleanShot 2025-01-31 at 19 44 29@2x

&

that does not filter out the phone number because there is a numeric sequence of length 4 at the end of the phone number.

thomaslombart:
I've investigated that further, and looking at one of our extensions in the store, they seem to have opted in for a [more precise search of an OTP code](https://github.com/raycast/extensions/blob/0894e9e6bf60897313a7eea0ba42aa7106e57c73/extensions/imessage-2fa/src/utils.ts). I still think that's the way to go, as filtering out only phone numbers won't eliminate false positives. Another example where this may fail is if the message includes an opt-out number, such as: `{COMPANY_NAME}: Your verification code is XXX XXX. STOP XX XXX`. In this case, the latter number will be wrongly considered as the code.

thomaslombart (approved):
All right, let's go with your approach then. I don't have much bandwidth for that either way, so that's already an improvement compared to before! 🙌

PR #16516, opened 2025-01-21
Update file-manager extension

### LINKED ISSUE #16268: [File Manager] ... respect hidden files
### Extension

https://www.raycast.com/erics118/file-manager

### Description

There's lots of trash in my main directory and i hide it with chflagshidden so it doesn't bother me, i would like to not see these through the extension either.

People who don't want to see hidden files

### Anything else?

_No response_

ridemountainpig:
## Description
Added the preference to show files hidden by the `chflags hidden` command.
Close #16268

## Checklist

### DIFF
--- a/extensions/file-manager/CHANGELOG.md
+++ b/extensions/file-manager/CHANGELOG.md
@@ -1,6 +1,6 @@
 # File Manager Changelog
 
-## [Features] - {PR_MERGE_DATE}
+## [Features] - 2025-01-28
 
 - Added the preference to show hidden files.
 


extensions/file-manager/src/components/directory.tsx
```diff
@@ -45,6 +46,13 @@ export function Directory(props: { path: string; ignores: GitIgnoreHelper[]; ini
           return false;
         }
       }
+      if (
+        !preferences.showChflagsHiddenFiles &&
+        f.type == "file" &&
+        execSync(`ls -lO "${props.path}/${f.name}"`, { encoding: "utf8" }).toLowerCase().includes("hidden")
```
erics118, line 52: Using `.includes("hidden")` causes issues if the username, or file name contains `hidden`. 

Instead, maybe use `stat -f%f {file}` and check for the bit 32768 (2^15).

See: https://discussions.apple.com/thread/2392066?sortBy=rank

I think that splitting by whitespace could also work, but I'm not sure if would get messy

extensions/file-manager/package.json
```diff
@@ -54,6 +55,14 @@
       "required": true,
       "type": "checkbox"
     },
+    {
+      "default": false,
+      "description": "Show the file hidden by the chflags hidden command.",
```
erics118, line 60: I would just say "hidden files", no need to elaborate that it is via chflags, as I believe there's only one to show/hide files on macos, aside from dotfiles

extensions/file-manager/src/components/directory.tsx
```diff
@@ -45,6 +46,13 @@ export function Directory(props: { path: string; ignores: GitIgnoreHelper[]; ini
           return false;
         }
       }
+      if (
+        !preferences.showChflagsHiddenFiles &&
+        f.type == "file" &&
```
erics118, line 51: Is checking type to be file necessary? What about directories?

erics118 (changes requested):
Looks pretty good, I like the feature. Just a few changes on the implementation

ridemountainpig:
@erics118 I have made the updates based on your comment. Could you please check it again? Thank you.

extensions/file-manager/src/components/directory.tsx
```diff
@@ -46,12 +46,15 @@ export function Directory(props: { path: string; ignores: GitIgnoreHelper[]; ini
...
-        execSync(`ls -lO "${props.path}/${f.name}"`, { encoding: "utf8" }).toLowerCase().includes("hidden")
-      ) {
-        return false;
+      if (!preferences.showHiddenFiles) {
+        const statOutput = execSync(`stat -f%f "${props.path}/${f.name}"`, {
+          encoding: "utf8",
+        }).trim();
+        const fileFlags = parseInt(statOutput, 10);
+
+        if (fileFlags == 32768) {
```
erics118, line 55: This should be `fileFlags & (1 << 15)` because there could be other bits present

ridemountainpig: Updated.

erics118 (approved):

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #16525, opened 2025-01-21
Add intermittent-fasting extension

pernielsentikaer:
## Checklist


pernielsentikaer:
I drafted something quick here @xmok - not sure if it's something for you to look into too - haven't tested fully yet but there is a good base 🙂

pernielsentikaer:
Also, you @LitoMore, if you have a minute, I think there are a lot of improvements that could be made 😂

LitoMore:
Thanks for inviting me to review. I have a flight tonight, and I will look at it tomorrow.

pernielsentikaer:
No worries at all @LitoMore, it will not be merged this week. It started as a silly idea, but I actually think it could be cool to have. Thanks in advance 💯

xmok:
Love the idea.
I tried it out for a couple of hours and logged a few "test" fasts (used smallest hour in `Preferences`). Some thoughts (maybe for now or for future):

1. Some form of calendar integration so fasts can be logged and seen in daily life (can be in future version)
2. For those fasting for weight reasons, add some form of weight tracking (maybe we ask for a start and target weight then prompt them to enter their current weight every X number of fasts)
3. For those fasting for religious reasons, add some links e.g. prayer times
4. The note in edit is great but for extra control, I suggest we let user set a custom emoji/icon
5. A Dropdown filter in History 👀

In terms of actual functionality, when I delete an item, there is a delay apart from that it's working great - MenuBar as well as the View.

I'll wait for LiteMore to chime in with their feedback 💯.

extensions/intermittent-fasting/src/components/actions/importHistory.tsx
```diff
@@ -0,0 +1,85 @@
...
+        title: "History Imported",
+        message: `Added ${newItemsCount} new fasting records`,
+      });
+    } catch (error) {
+      console.error(error);
+      await showToast({
+        style: Toast.Style.Failure,
+        title: "Error",
+        message: error instanceof Error ? error.message : "Failed to import fasting history",
+      });
```
LitoMore, line 80: Does the `showToast()` have a fallback to `showHUD()` when the Raycast window is closed? The Raycast window will lose focus and close when selecting files. I can't see a toast even though I reopen the Raycast window immediately. And it shows a HUD instead of the Toast.

extensions/intermittent-fasting/package.json
```diff
@@ -0,0 +1,105 @@
...
+            },
+            {
+              "title": "Icon with Text",
+              "value": "full"
+            },
+            {
+              "title": "Show Eating Window",
+              "value": "eating"
+            }
+          ],
```
LitoMore, line 57: How about adding a checkbox option to allow users to switch `Show Eating Window on Idle Display` for all options? Then we have:

- Icon Only (show eating window)
- Icon Only (no eating window)
- Icon with Text (show eating window)
- Icon with Text (no eating window)

extensions/intermittent-fasting/src/components/actions/importHistory.tsx
pernielsentikaer: Yeah, I think it needs to be a new `Form` view with a picker instead, since if you have your Pop to Root search setting set to `Immediately` it will terminate the second you touches the save dialog

Well spotted! 👏

pernielsentikaer: But yes, if Raycast closes and you're showing a toast, then it shows HUD 

image

LitoMore: > But yes, if Raycast closes and you're showing a toast, then it shows HUD.

I think we can document this behavior in the developer doc.

pernielsentikaer: Makes sense 🙂

extensions/intermittent-fasting/package.json
pernielsentikaer: That's a good idea!

pernielsentikaer:
Added better import/export @LitoMore :)

pernielsentikaer:
Could you approve this @grzegorzkrukowski

grzegorzkrukowski (approved):

PR #16564, opened 2025-01-22
[Todoist] Support task deadlines alongside due dates

kcole93:
## Description

Added support for [task deadlines](https://todoist.com/help/articles/introduction-to-deadlines-uMqbSLM6U) in addition to due dates:
- Tasks can now have both a due date (when you plan to do it) and a deadline (when it must be done by)
- Deadlines are displayed with a bullseye icon in task lists and detail views
- In task list views, the bullseye icon is rendered as `BullsEyeMissed` and in red when the deadline is overdue
- Added new helper function `displayDeadlineDate()` to format deadline dates, following the pattern in `displayDueDate()`:
  - "Today" for today
  - "Tomorrow" for tomorrow
  - Day name (e.g., "Wednesday") for dates within the next week
  - "dd MMMM" for dates in the current year
  - "dd MMMM yyyy" for dates in other years

## Screencast

[Screencast Demo](https://share.cleanshot.com/GwG9tMn4)

## Checklist


extensions/todoist/src/create-task.tsx
```diff
@@ -57,6 +58,14 @@ function CreateTask({ fromProjectId, fromLabel, fromTodayEmptyView, draftValues
         };
       }
 
+      if (values.deadline) {
+        body.deadline = {
+          date: Form.DatePicker.isFullDay(values.deadline)
+            ? getAPIDate(values.deadline)
+            : values.deadline.toISOString(),
```
thomaslombart, line 65: No need for that condition since deadlines should always be full day dates.

thomaslombart (commented):
Thanks a lot for the PR! Here's some feedback:

- Would it be possible to rename `Due Date` to `Date` as done in Todoist? And also rename `Deadline Date` to just `Deadline`?
- Could you make `displayDueDate` a more generic helper and use it for deadlines since `displayDeadlineDate` has the exact same implementation? You could go with `displayDate` for example.

extensions/todoist/src/create-task.tsx
kcole93: Hi Thomas,

Thanks for your review. I renamed `Due Date` to `Date` in the UI as well as in the extension's code where possible to reflect the shift away from the mental concept of `Due Dates`; however, the Todoist REST API [still uses the `due` object](https://developer.todoist.com/rest/v2/#tasks) to handle Task dates.

I also made `displayDueDate` a more generic helper function `displayDate` as you suggested, which now handles formatting both Dates and Deadlines for display in the UI.

Let me know how this looks to you -- I was nervous to make such broad changes as my first contribution to this extension.

Thank you!

thomaslombart (approved):
I've just fixed a small issue where the deadline would be set to today when creating a task from today's empty view. It doesn't really make sense in that context. Otherwise, everything's good to go. Thanks for going the extra mile!

jesminewhee:
@thomaslombart @kcole93 just wanted to say a huge thank you, as this is really important for my work. Appreciate you both taking the time!

RobErskine:
@kcole93 sorry to drop this comment on a merged PR; not sure where else to bring it up.

I currently don't have Todoist premium, so I can't access this Deadline feature. I mostly avoid it, but I still fat finger it from time to time when tabbing through creating a new todo item. 
CleanShot 2025-02-19 at 11 51 57@2x

My question is, would it be useful to hide/show this Deadline field (and other premium only features) depending on a user's access?

When syncing content, we do get [that data back as part of the `User` object](https://developer.todoist.com/sync/v9/#user) with a `premium_status` attribute, which will give us back `not_premium`, `current_personal_plan`, `active_business_account` or `teams_business_member`.

For example, add a check in the `create-task.tsx` file:
```
if (data?.user?.premium_status !== "not_premium") {
  <Form.DatePicker {...itemProps.deadline} title="Deadline" type={Form.DatePicker.Type.Date} />
}
```
So only premium users who can add deadlines actually get that UI?

thomaslombart:
Yeah, that makes sense! Feel free to come up with a PR 🙂

PR #16566, opened 2025-01-22
Add hipster-ipsum extension

plevik:
## Description

Extension to seamlessly connect the Hipster Ipsum API with Raycast. The extension offers a couple of options that serve as API defaults, along with two commands: Generate Paragraphs and Generate Sentences. Each command currently has one parameter for the number of sentences or paragraphs to generate. This parameter defaults to the extension's option, or to the value of 3 if not specified. The output from the command can be stored in the clipboard, pasted into the current app, or both simultaneously. There is only one required extension option: the Type of output, which also has a default value. This requirement is enforced by the API.

Note/Observation/Idea for improvement: There is no mention of where screenshot files should be located, neither in the preparation for publishing article nor in the publishing extensions article. I have only assumed—based on tests that seem to confirm this—that they should be in the metadata folder in the extension root.

## Screencast

hipster-ipsum-generate-sentences

hipster-ipsum-generate-paragraphs

## Checklist


plevik:
Extension to seamlessly connect the Hipster Ipsum API with Raycast. The extension offers a couple of options that serve as API defaults, along with two commands: `Generate Paragraphs` and `Generate Sentences`. Each command currently has one parameter for the number of sentences or paragraphs to generate. This parameter defaults to the extension's option, or to the value of 3 if not specified. The output from the command can be stored in the clipboard, pasted into the current app, or both simultaneously. There is only one required extension option: the `Type` of output, which also has a default value. This requirement is enforced by the API.

**Note/Observation/Idea for improvement**: There is no mention of where screenshot files should be located, neither in the preparation for publishing article nor in the publishing extensions article. I have only assumed—based on tests that seem to confirm this—that they should be in the `metadata` folder in the extension root.

extensions/hipster-ipsum/package.json
```diff
@@ -0,0 +1,225 @@
...
+        }
+      ],
+      "default": "3",
+      "required": false
+    },
+    {
+      "name": "startWithLorem",
+      "type": "checkbox",
+      "title": "Other Options",
+      "description": "Select to start the first paragraph with the totally fucking ridiculous phrase “I’m baby”",
```
pernielsentikaer, line 198: Could we make a better wording here 😂

plevik: Sure we can, any suggestions? I have just re-used what API author used in [the documentation](https://hipsum.co/the-api/). Thanks.

pernielsentikaer: Shouldn't it reflect that it starts with "I’m baby"?

plevik: Gotcha, see updated in the last commit. Let me know if anything else needs to be changed/updated. Thank you!

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #16615, opened 2025-01-26
Update google-search extension

jlokos:
## Description

- Add a fall back to clipboard if no text is selected

## Checklist

### DIFF
--- a/extensions/google-search/CHANGELOG.md
+++ b/extensions/google-search/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Google Search Changelog
 
-## [Update] - {PR_MERGE_DATE}
+## [Update] - 2025-02-22
 
 - Add clipboard fallback preference for selected text searches
 


pernielsentikaer:
Thanks for your update @jlokos 

@mblode do you want to check this?

pernielsentikaer (changes requested):
Are you still around @mblode - we could consider making this a preference

extensions/google-search/package.json
```diff
@@ -18,7 +19,8 @@
       "name": "index",
       "title": "Google Search",
       "description": "Google search with autosuggestions",
-      "mode": "view"
+      "mode": "view",
+      "permissions": ["clipboard"]
```
LitoMore, line 23: The `permissions` field looks useless since they don't support defining permissions from the manifest.

https://developers.raycast.com/information/manifest

jlokos: this is correct!

pernielsentikaer:
Would it make sense to make a preference here so it's possible to do as before, or doesn't that make any sense?

jlokos:
> Would it make sense to make a preference here so it's possible to do as before, or doesn't that make any sense?

This simply provides a fallback for when text is not selected.

LitoMore:
It's better to launch this feature with a new preference option. And default it to disabled to avoid affecting existing users.

jlokos:
@pernielsentikaer i've mad the clipboard fallback a preference as recommended.

extensions/google-search/package.json
```diff
@@ -35,6 +36,14 @@
       "default": true,
       "label": "Remember Search History",
       "description": "Enable this checkbox to locally store Google search history."
+    },
+    {
+      "name": "useClipboardFallback",
+      "type": "checkbox",
+      "required": false,
+      "default": true,
```
LitoMore, line 44:
```suggestion
      "default": false,
```

LitoMore (changes requested):

extensions/google-search/package.json
jlokos: oops - i've resolved this

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #16619, opened 2025-01-26
Add canvas extension

jakmaz:
## Description

This is an extension that connects to the public API of the Canvas LMS, a widely used Learning Management System.  

There is a [similar extension](https://www.raycast.com/Yoda/canvascast), but it has been unmaintained for three years and lacks many features that this extension provides and plans to implement in the future.  

## Screencast

All screenshots are available in the `metadata` directory.

## Checklist


pernielsentikaer:
We should always try to enhance current extensions before creating a similar, we could reach out to @yodalightsabr to see if he is still around 😊

jakmaz:
> We should always try to enhance current extensions before creating a similar, we could reach out to @YodaLightsabr to see if he is still around 😊

Totally understand that, however as mentioned, the other extension hasn't had any updates for several years and follows a completely different approach. I am sure users will like it and I really hope you look into it ;)

Thank you!

YodaLightsabr:
Hi there! It's awesome to see more people involved in the Canvas API community. I've been maintaining the CanvasCast extension for the last 3 years, making bug fixes and small changes as needed. I'll give this PR a thorough look after class today, but I wanted to ask: what specifically is different between our two extensions? Many people, including myself, still use CanvasCast daily, and I'm happy to add new features to improve the extension.

jakmaz:
Hey @YodaLightsabr,

I assumed it was no longer maintained based on the last commit, which was over three years ago. Before developing my own extension, I tried CanvaCast but wasn’t entirely satisfied with it. Below are a few issues I encountered—some more significant than others—but I aimed to improve (and hopefully did) upon all of them:

- The feed only displays the newest activities, with no way to check older ones.
- Courses only show favorited ones, rather than all enrolled courses.
- Past assignments are not visible.
- There’s no option to view assignments and announcements for a specific course.
- Date formatting often causes the time to be cut off when activity titles are long (e.g., 1:22 ...).
- No README clearly explaining how to set up the token.
- No screenshots

That said, there are also features I liked—such as the ability to open a module or assignment directly in Raycast—which I haven't implemented yet in my version.

Ultimately, I decided to create my own extension because the required changes were too extensive. Additionally, I wanted to use a different command structure to mirror Canvas’s structure by including all courses and dedicated pages for them, as shown below:

https://github.com/user-attachments/assets/6271ed0d-8b10-46e1-8304-29a948a5b9d8

I wanted this extension to give the most insights, with the ability to search through all of their contents, rather than giving the small overview of what is currently happening.

I look forward to hearing from @pernielsentikaer and @YodaLightsabr

YodaLightsabr:
I can understand wanting some more flexibility with regards to the different views. However, many of the things you listed could be small PRs to the existing extension, such as adding screenshots or explaining the token generation process in the README. The favorite courses, for example, could be a simple preference toggle in extension settings.

I want to make sure that Canvas works well on Raycast, and I don't think that having two extensions gets us there. That said, I'm happy to work with you on expanding the current extension. Every institution / teacher / class does things differently, and as a result, I want it to support as many workflows as possible. We've had a lot of [community engagement](https://github.com/raycast/extensions/issues?q=is:pr%20canvascast) in the past, and I think that's what makes the extension so great.

jakmaz:
I’ll admit that part of my motivation was to create something new, rather than extensively modifying an existing extension that already has users who appreciate it. The single-extension-per-service rule feels a bit restrictive, as it limits new users from having multiple options and choosing the one that best fits their needs.

Would there be a way to introduce enough differentiation in this new extension to justify its inclusion in the store? Otherwise, my only option would be to use it locally and share it privately with friends.

YodaLightsabr:
I would love to hear your thoughts on this @pernielsentikaer. I love building new things, but I think having another extension that serves a very similar purpose would clutter the store and make things confusing. However, I would definitely be down to work with you on including any new functionality in the existing extension, and I want to make sure it suits everyone's needs.

pernielsentikaer:
It doesn't make any sense to have two so similar extensions in the store, this is also a part of our [Extensions Guidelines](https://manual.raycast.com/extensions) - there is only one loser and it's the users being very confused about which to take.

It should be possible to enhance the current so it matches what's possible

pernielsentikaer:
I understand your perspective on the limitations that arise when developers are encouraged to enhance existing extensions instead of creating new ones. This approach can restrict creativity and innovation, as developers may feel compelled to work within the confines of existing solutions rather than exploring new possibilities.

From a user's standpoint, having multiple options can be overwhelming and lead to confusion about which solution best fits their needs. While variety can be beneficial, it also requires users to spend more time evaluating and selecting the right option, which can be frustrating.

It's unfortunate that the constraints around extension development weren't clear from the beginning, and I can see how that could be disappointing. I hope that by sharing your insights, we can inspire updates to the current extension, whether by you, me, or anyone else.

Ultimately, we all want the best possible extension for a service

PR #16624, opened 2025-01-26
Add github-search extension

webpro:
## Description

Raycast extension to open the GitHub Search web UI with pre-defined filters

- Includes options from both [search][1] and [advanced search][2]
- Save searches
- Reusable filters
- Exclude apps/bots from code, issue and pull request results
- Reuse existing tab with GitHub Search open in browser
- Dedicated command to run saved searches directly

[1]: https://github.com/search
[2]: https://github.com/search/advanced

## Checklist


LitoMore:
How about moving this feature to the existing [GitHub extension](https://www.raycast.com/raycast/github)?

webpro:
That I've considered, but this extension does not require auth. There's something to say for keeping things focused and not bloated. It's basically just a url builder.

extensions/github-search/src/run-saved-search.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+  return (
+    <List>
+      {sortedSearches.length === 0 ? (
+        <List.EmptyView
+          title="No Saved Searches"
+          description="Create your first saved search in the main GitHub Search view"
+          actions={
+            <ActionPanel>
+              <Action title="Open GitHub Search" onAction={() => launchCommand({ name: 'search', type: LaunchType.UserInitiated })} />
```
pernielsentikaer, line 21: Should probably be `async` with `await` - this can potential crash the extension so we also need it in a try catch

pernielsentikaer (changes requested):
Hi 👋 

- We could probably use [useForm](https://developers.raycast.com/utilities/react-hooks/useform) here which handles states, validation and errors - right now it's showing a success toast on failure

github-search 2025-02-04 at 09 34 23

- Should it be possible to edit a saved search?

github-search 2025-02-04 at 09 38 04

- It's possible to save a search without a name, should we validate for that, otherwise it looks line an empty line in this dropdown

github-search 2025-02-04 at 09 38 57

- Should we consider moving `Saved Seaches` in the top with `Filters`

- There might be more, but let's stick with this for now 😊

webpro:
> Hi 👋
> 
> Thanks for your contribution 💪
> 
> I have now tested your extension, and I have some feedback ready for you:

Great, thanks for the review!

> * We could probably use [useForm](https://developers.raycast.com/utilities/react-hooks/useform) here which handles states, validation and errors - right now it's showing a success toast on failure

Thanks, I didn't really consider that before. But now that I'm looking at it I have two questions:

1) Is it possible to set values without using the fields' `onChange` handler? This is necessary e.g. to update the `name` and `filter` fields in `ReusableFilterForm` when another saved filter is selected. Same for the main form when a saved search is selected.
2) If positive, should `handleSubmit(values)` be called in the `onSubmit` handler, and can some logic be added after that?

> * Should it be possible to edit a saved search?

Yes. The action menu has options to Update and Rename saved searches. Or are there better patterns available for this?

> 
> * It's possible to save a search without a name, should we validate for that, otherwise it looks line an empty line in this dropdown

Good catch, fixed.

> * Should we consider moving `Saved Seaches` in the top with `Filters`

Do you mean to move the Saved Searches + Reusable Filters dropdowns to same block as Query i.e. basically remove the horizontal rule? Not having a strong opinion here, imho it's a bit of a hotchpotch either way. Also, felt like two separate forms/commands won't cut it either (but open to try).

extensions/github-search/src/run-saved-search.tsx
webpro: Okay :)

webpro:
1. The idea is to save first (`⌘-S`), then select (`⌘-S` again).
2. On save, it should return to the previous screen with the filter selected.
3. The amount of text cq the number of lines here is unknown upfront, hence a text area. Managing input fields is not impossible, but sounds a bit cumbersome. Got a good example for this maybe? Personally I don't mind a textarea, users can choose to make a single long line or split up over multiple lines as they please.
4. Done (2x). Also looked for more opportunities to use `Shortcuts` but didn't find more.
5. Exactly this is in the filter view, where it's more relevant (for extra/custom filters). Not in the main view as I felt it could be distracting, and it shouldn't be necessary (the form is meant to abstract away that complexity).

pernielsentikaer:
1. That doesn't feel intuitive and I bet users would feel the same, should we maybe just auto select somehow, is that possible? I was pretty confused
2. Might be the same as (1) since I was confused about the flow - maybe an action to update/save and select and one for just saving? Not sure if that makes sense?

webpro:
Happy to change it, but not sure what's the best way to go about it.

Selection is usually done using the default `⌘-Enter`. So I'd say then it makes sense to use `⌘-S` for Save, and that works, but also the first/default action is always assigned `⌘-Enter`. Wouldn't have issues with this solution, though. One can use `⌘-S` for save and `⌘-Enter` to select. Just pushed this change.

Screenshot 2025-02-13 at 6 41 38 AM

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

webpro:
> Hi 👋
> 
> Let's go with that, let's get it in the store so we can see feedback :)

Thanks! Yeah let's see if/how anything violates the "don't make me think" rule :)

PR #16625, opened 2025-01-26
Add ukrop extension

yarul0:
## Checklist


pernielsentikaer:
Thanks for your PR @yarul0 🔥

I'm not sure it makes sense to branch out here. Wouldn't it be better to have one unified extension that offers multiple languages as the only difference?

@red-avtovo what do you think?

red-avtovo:
Thanks @pernielsentikaer for moderation.  I do agree, but the latest keyboard switch should somehow know the exact system name of the layout to switch to. I only recently switched to the new implementation and still discover what would be the best options here.

@yarul0 sorry that my changes caused a disruption in your extension experience. I added all ru layouts only because I was thinking that the use cases are limited to these layouts only. Obviously, I was wrong, so I would graldly extend the Cyrillic layout list to all Ukranian keyboards and promote in the future to extend that list instead of forking the extension

Additionally, please let me know @yarul0 if you would like to adjust the layout conversion map if that doesn't work in all cases

yarul0:
Hello!
All good, no  worries. I'm not going to do something else.

Thanks, for you time!

red-avtovo:
@yarul0 please take a look at my PR and let me know if anything is missing

yarul0:
Looks good to me.

Issue #16579, opened 2025-01-23
[Todoist] Change hotkey for task completion


ironcrafter54:
### Extension

https://www.raycast.com/doist/todoist

### Description

It would be great if the hotkey to complete a task was changed to something more intuitive as it is something done quite frequently, command shift e is quite difficult to remember. I feel like command + enter would be much better suited

### Who will benefit from this feature?

new users of the integration, as well as any other

RobErskine:
This seems pretty easy, I could probably take it on. I don't think changing it would be intuitive for existing users (I'm a todoist user personally). However maybe this could be a command preference or something?

jfkisafk:
@RobErskine afaik modifier keys for shortcuts can't be injected dynamically (I can be wrong) - so that rules out having a textfield preference. The best we can do is have a dropdown of common shortcuts in the command preference (defaulting to current one) and go from there. Is that what you were thinking?

RobErskine:
oh hey @jfkisafk. I'll have to check to see if the shortcut key itself can be dynamic... but i think you're right, that it has to be a dropdown. I've done this before in [my storyblok extension](https://github.com/raycast/extensions/blob/3e251f581ebf5c321e52d2fa1bf6c27d6217767e/extensions/storyblok/src/index.tsx#L12).

![storyblok keyboard modifier preference](https://github.com/user-attachments/assets/97e5d5ba-a2cc-4095-92e9-709893dade17)

---

That doesn't necessarily satisfy OP's ask of simplifying the shortcut from (`cmd` + `shift` + `r`) to just (`cmd` + `enter`). You can set the shortcut to be (`cmd` + `enter`) using `...shortcut={{ modifiers: ["cmd"], key: "enter" }}` so long as it _isn't_ the first shortcut (which will default to just `enter`).

Maybe an extension preference for a "Complete Task Shortcut" which by default is set to (`cmd`+ `shift` + `e`) but can be changed to (`cmd` + `enter`)? I guess if something like that were to exist does it make sense to make preferences for other things?

jfkisafk:
@RobErskine in my relative experience, Raycast uses "Cmd + Enter" for the second action in the list (if it is not wrapped under a section). So we can try this:
1. Make "Complete Task" second action in the list of actions (if not already).
2. Don't assign the shortcut parameter to this action if the user has selected "Cmd + enter" from preference dropdown. For all other options, we need to specify the shortcut parameter to that action (to change it from default "Cmd + Enter".

[...]

PR #16633, opened 2025-01-27
Todoist: Adding Complete Task shortcut preference

RobErskine:
## Description

Closes https://github.com/raycast/extensions/issues/16579

The requester was asking for a simpler keyboard shortcut for completing tasks. This PR adds a global preference to leave the shortcut as it is today (`shift` + `cmd` + `e`) as the default for existing users, as well as a new simplified preference to make the shortcut (`cmd` + `enter`) if users prefer to change it.

## Screencast

Here is the default dropdown value (to keep the experience the same to what it is today for existing users):
Settings-2025-01-26-19 51 03@2x

Here is the shortcut for when the dropdown preference is set to the simplified (`cmd` + `enter`) shortcut:
Settings-2025-01-26-19 50 23@2x

## Checklist

onsextensionstodoist-2025-01-26-20 18 50@2x

---

ccing @jfkisafk since we were discussing the best way to tackle this [in the original issue thread, here](https://github.com/raycast/extensions/issues/16579).

### DIFF
--- a/extensions/todoist/README.md
+++ b/extensions/todoist/README.md
@@ -1,4 +1,4 @@
-# Todoist
+My understanding is that scheduled means planned for today or tomorrow or even further into the future ?# Todoist
 
 **This extension is not created by, affiliated with, or supported by Doist.**
 
@@ -8,10 +8,10 @@ Check your Todoist tasks from within Raycast and quickly create new ones.
 
 You can quickly access your `Home` command favorites views using deep-links and quick-links. Here's how:
 
-1.  Open the `Home` command and select your view in the dropdown (e.g `All Tasks` or a specific project)
-2.  Search for `Create View Quicklink` in the actions (`⌘` + `K`).
-3.  Give the quicklink a custom name (optional) and create it (`⌘` + `⏎`).
-4.  Use Raycast root search to find your new quicklink and quickly access your tasks view.
+1. Open the `Home` command and select your view in the dropdown (e.g `All Tasks` or a specific project)
+2. Search for `Create View Quicklink` in the actions (`⌘` + `K`).
+3. Give the quicklink a custom name (optional) and create it (`⌘` + `⏎`).
+4. Use Raycast root search to find your new quicklink and quickly access your tasks view.
 
 This makes it easy to access any of your views, including projects and labels!
 


extensions/todoist/src/components/TaskActions.tsx
```diff
@@ -40,6 +50,12 @@ import TaskCommentForm from "./TaskCommentForm";
 import TaskComments from "./TaskComments";
 import TaskEdit from "./TaskEdit";
 
+interface Preferences {
+  completeTaskShortcut: string;
+}
+
```
jfkisafk, line 56:
```suggestion
```

We don't need this as `Preferences` is already present in _raycast-env.d.ts_

extensions/todoist/src/components/TaskActions.tsx
```diff
@@ -200,6 +216,15 @@ export default function TaskActions({
...
+        icon={Icon.Checkmark}
+        shortcut={
+          preferences.completeTaskShortcut === "shiftCmdE" ? { modifiers: ["cmd", "shift"], key: "e" } : undefined
+        }
+        onAction={() => completeTask(task)}
+      />
+
       {isTodoistInstalled ? (
         <Action.Open
           title="Open Task in Todoist"
```
jfkisafk, line 230:
```suggestion
          title="Open Task in Todoist"
          shortcut={Keyboard.Shortcut.Common.Open}
```

Since this was the second option before, we lose the backwards compatibility. After this change the option to open task in Todoist will be left without a shortcut. I think it would be better/simpler to just use the same shortcut "Cmd + o" for opening task in the app or browser - rather than different shortcuts that existed before. Please make sure to indicate this breaking change in Changelog.

extensions/todoist/src/components/TaskActions.tsx
```diff
@@ -1,4 +1,14 @@
-import { Action, ActionPanel, Color, Icon, Toast, confirmAlert, showToast, useNavigation } from "@raycast/api";
+import {
+  Action,
+  ActionPanel,
+  Color,
+  Icon,
```
jfkisafk, line 5:
```suggestion
  Icon,
  Keyboard,
```

Just to fix the import issue in case you apply suggestions directly from PR.

extensions/todoist/src/components/TaskActions.tsx
RobErskine: I added `shortcut={Keyboard.Shortcut.Common.Open}` to both the Open in Todoist Browser and Open in Todoist app. (`cmd` + `o`) should still work just like before, only difference is it is no longer the second option to select. `Complete task` is sandwiched in-between "show details" and "Open in Todoist" now.

jfkisafk (commented):
Approving however we have changed the shortcut to open in Todoist from Cmd + Enter to Cmd + O without making a note in Changelog.

jfkisafk (approved):

thomaslombart (commented):
I'm not sure the preference makes sense here. Since `Complete Task` has been moved to the second action, ⌘ + ⏎ will always be the shortcut preference.

This change also affects how Todoist tasks open. They no longer open with ⌘ + ⏎ but now require ⌘ + O. I don't know if this will disrupt many users' muscle memory, as we don't have analytics on that, though.

thomaslombart:
Yeah, that would make sense, although I'm not really fond of making these changes. Changing shortcuts to suit others' workflows and preferences may seem fine at first, but it can lead to a mess in the long term.

I'm leaning towards just changing the shortcuts without any preferences, as completing a task is a popular action and ⌘ + O makes more sense to open a task.

kcole93:
Hi @thomaslombart -- just a heads up that it looks like some unintended text got added to the readme on the latest commit here.

thomaslombart:
> Hi @thomaslombart -- just a heads up that it looks like some unintended text got added to the readme on the latest commit here.

Thanks for the heads-up, I just fixed it 🙂

PR #16644, opened 2025-01-27
Add discord-for-raycast extension

nicollaspetrelli:
## Description

This PR introduces a new Raycast extension that integrates Discord via OAuth2, allowing users to seamlessly navigate their Discord servers directly from Raycast. The extension provides a single, highly efficient command that displays the user's servers as a list, enabling quick access without the need to open the Discord app or manually search for a server.

Key Features:
OAuth2 Integration: Securely authenticates users with Discord, ensuring a safe and seamless login experience.

Server List Navigation: Displays the user's Discord servers in a clean, searchable list within Raycast.

Quick Access: Users can directly jump to a specific server without opening the Discord interface, saving time and streamlining their workflow.

Despite being a single command, this extension is incredibly useful for Discord power users who frequently switch between servers. It eliminates the need to open the Discord app, search for a server, and then navigate to it, making the process much faster and more efficient.

## Checklist


julianapetrelli (approved):

pernielsentikaer:
Would it be possible to unify this into the current [Discord extension](https://www.raycast.com/Aayush9029/discord) by @Aayush9029  - would the other features be possible with oAuth too?

nicollaspetrelli:
Of course we can unify it in a single extension, about OAuth unfortunately discord is not very permissive when it comes to listing your contacts to send a direct message to a friend (that was my initial idea), but due to the high number of bots I believe they do not allow this for security reasons.

Aayush9029:
We could do whatever SwiftCord is doing (reverse engineering their internal APIs but that will turn this extension into a complex app) 

I think it's better to merge the extensions for now, we can always add more stuff later.

pernielsentikaer:
Both extensions have the same name, so we need to change this before it’s possible. However, wouldn’t it make sense to add it to your @Aayush9029? I guess it’s fine that one of the commands needs OAuth 🙂

Aayush9029:
> I think it's better to merge the extensions for now, we can always add more stuff later. - A

> wouldn’t it make sense to add it to your @Aayush9029? I guess it’s fine that one of the commands needs OAuth 🙂 - P

yeah that's what I meant 😅, add the oauth as another command to the existing ext.

pernielsentikaer:
Agree, could you look into that @nicollaspetrelli

PR #16675, opened 2025-01-28
Add strlen extension

WeeJeWel:
## Description

This extension shows the length of a string.

## Screencast

Screenshot 2025-01-28 at 19 19 28

## Checklist

### DIFF
--- a/extensions/strlen/src/index.tsx
+++ b/extensions/strlen/src/index.tsx
@@ -1,6 +1,9 @@
 import { List, Icon, Color, Clipboard } from "@raycast/api";
 import { useEffect, useState } from "react";
 
+import stringLength from "string-length";
+import stringWidth from "string-width";
+
 export default function Command() {
   const [input, setInput] = useState<string>("");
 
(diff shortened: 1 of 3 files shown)


LitoMore:
I think you should use https://github.com/sindresorhus/string-length to calculate the correct length.

Also, it's better to add https://github.com/sindresorhus/string-width for calculating width.

WeeJeWel:
> I think you should use https://github.com/sindresorhus/string-length to calculate the correct length.
> 
> Also, it's better to add https://github.com/sindresorhus/string-width for calculating width.

Hm, might be nice for a second and third field.

extensions/strlen/src/index.tsx
```diff
@@ -0,0 +1,55 @@
...
+            key={"width"}
+            icon={{ source: Icon.Hashtag, tintColor: Color.Magenta }}
+            title={"Width"}
+            subtitle={{
+              value: stringWidth(input).toString(),
+              tooltip:
+                "The width of this string differs from its length due to special characters.",
+            }}
+          />
+        )}
```
LitoMore, line 51: You can always show the result of `string-length` and `string-width` since they are under different calculation logics.

You can also provide the library link to users to let know why we have three results.

pernielsentikaer:
This is also already possible in Clipboard History 🙂

image

WeeJeWel:
Yea but not all strings I want to check are on the clipboard.

Plus, `strlen` is just easier to remember imho.

pernielsentikaer:
What about [Toolbox](https://www.raycast.com/Kang/toolbox)? We might be able to make some support for making quicklinks for a specific view 😊

WeeJeWel:
Why the urge to bundle this into another extension?

I think the Linux philosophy of ‘one tool one job’ is still the best.

pernielsentikaer:
It's because our store is massive and we're trying not to add too many with the same purpose

I also found [this extension](https://www.raycast.com/itsmingjie/word-count) in the store which seems to be what you're looking for with extra options too 🙂

Issue #12509, opened 2024-05-22
Jira Enable to use self hosted instances


giolf:
### Extension

https://www.raycast.com/raycast/jira

### Description

Currently it seems not to be possibile to use this extension on self hosted jira instances.
It would be super handy to use it also on own jira instances.

### Who will benefit from this feature?

Each user which has to deal with self hosted jira instances

thomaslombart:
It is possible by using an [API token](https://www.raycast.com/raycast/jira).

giolf:
However if I use the extension called jira self hosted with the same api tokens it works. so I guess the issue is in ur extension.

aghilas-xb:
Hello @thomaslombart and thank you for your work !
Could you please reopen this issue as it seems many of us can't make it work using api tokens.
As stated by someone else on another issue, the token/auth proved to be working using the [Raycast Jira Search (Self-Hosted Edition)](https://www.raycast.com/emanguy/jira-search-self-hosted) extension.

giolf:
This is the main reason to stop me to be a paid user of raycsast: discordant core team

thomaslombart:
Jira is included for free in Raycast fyi. I've reopened the issue, feel free to take a look at it.

PR #16734, opened 2025-01-31
Improve self-hosted Jira support

shlikhota:
## Description

I've encountered the same issue as https://github.com/raycast/extensions/issues/12509.
The main problem is that self-hosted Jira instances don't support API version 3 and require authentication using PAT (Personal Access Token) instead of an API token.

Closes https://github.com/raycast/extensions/issues/17424

## Checklist


thomaslombart (commented):
I didn't look too much into that but are all API endpoints retrocompatible with v2?

shlikhota:
I did some research about backward-compatibility of API v3 and found that the key difference is added ADF (Atlassian Document Format) in fields like descriptions and comments. Fortunately, this doesn't seem to be a big deal, as the extension renders issues correctly. 

Additionally, the extension already uses new API v3 endpoints for creating issues and retrieving project list.
I overlooked this when I checked functionality — my mistake.

Here is how we could address the problem:
1. Add a disclaimer that the old API version is deprecated and not fully supported by the extension.
2. Add backward-compatibility for retrieving project list by using an old endpoint, as it's essential for search and active sprints.
3. Handle creating issues by opening the issue creation link in the browser when using the old API.

@thomaslombart could you confirm if this approach makes sense before I proceed with any changes?

shlikhota:
Another idea is to simply fork the extension as was done with the Jira search extension and maintain it separately.

thomaslombart:
I wouldn't go with solution 2, as this would directly introduce technical debt in the codebase for little added value. Solution 1 isn't ideal, as it would lead people to think the extension can do all that the extension offers when it cannot.

Have you looked into supporting personal access tokens instead? Is this technically feasible?

shlikhota:
The latest supported API for self-hosted Jira is version 2, which imposes constraints on the choice of the solutions because the extension relies on API endpoints from version 3 that don't exist in API v2.

I see two main options for solving this issue:
1. Add backward compatibility for the functionality that already uses the new API endpoints. However, this will introduce challenges when adding new features in the future, as backward compatibility would need to take into account.
I've also come to believe that this approach is not worth it in the long run.
2. Fork the extension for users of self-hosted Jira who only require basic functionality.
I believe this option is more maintainable and minimizes complications when making further changes to the original extension.

thomaslombart:
You said in the issue's description:

> The main problem is that self-hosted Jira instances don't support API version 3 and require authentication using PAT (Personal Access Token) instead of an API token.

Is it possible at all to support personal access tokens instead to make use of v3's endpoints?

shlikhota:
> Is it possible at all to support personal access tokens instead to make use of v3's endpoints?
Yes, I've added an option in this PR to use a Personal Access Token (PAT).
However it still relies on API v2, as I previously explained.

thomaslombart:
Ah, I thought PATs were supported in v3, all right. But this doesn't solve the backward compatibility problems for some endpoints, right? So what's not working in v2 is:

- ADF
- Creating issues
- Retrieving projects

Anything more? If it's just that, that's probably fine and we should add a disclaimer somewhere if a user uses v2. But that'd be more of a problem if there are more breaking changes since that would make the codebase more complex.

shlikhota:
Apologies for my slow response here.

Yes, I checked functionalities and it seems like only those don't have backward compatibility.

> But that'd be more of a problem if there are more breaking changes since that would make the codebase more complex.

I completely agree with that point and I don't think that this workaround should be an obstacle for further extension changes and improvements.

I propose the following disclaimer: **"Experimental. Future updates may break compatibility."**

Also, I think it would be better to have just one checkbox, **"Use token as PAT for self-hosted Jira"** instead of a dropdown for API versions and a separate checkbox. What do you think?

thomaslombart (approved):

ahmed-rezk-dev:
I tried to run it. 
open issue working. But not Active sprint im getting the error blow

```
Error: {"errorMessages":["No project could be found with key 'search'."],"errors":{}}
    at ht (/.config/raycast/extensions/jira/active-sprints.js:94:2544)
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
    at async L2 (/.config/raycast/extensions/jira/active-sprints.js:94:3231)
```

thomaslombart:
This is probably happening because some APIs don't have backwards-compatibility. Not sure we can't do much more without making the extension way more complex.

PR #16745, opened 2025-01-31
[Obsidian] Add prepend option to append to daily note command

KevinBatdorf:
This adds a new option to allow users to prepend to the daily note rather than append.

It's false by default so wont impact existing users.

Requested by a user here: https://github.com/KevinBatdorf/obsidian-raycast/issues/90


pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #16763, opened 2025-02-01
Apple Reminders: Add option to group by "Upcoming"

vapourismo:
## Description

Adds a new grouping mode to the "My Reminders" command of the "Apple Reminders" extension.

Its groups are:
  - Overdue
  - Today
  - Upcoming
  - Later

"Upcoming" groups all reminders for the next 7 days.

"Later" contains reminders whose due date is more than 7 days into the future and those reminders without a due date.

## Screencast

apple-reminders 2025-02-01 at 18 24 48

## Checklist

### DIFF
--- a/extensions/apple-reminders/CHANGELOG.md
+++ b/extensions/apple-reminders/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Apple Reminders Changelog
 
-## [Add new "Upcoming" grouping option] - {PR_MERGE_DATE}
+## [Add new "Upcoming" grouping option] - 2025-02-10
 
 - Add a new mode which allows group reminders by when they are upcoming in the `My Reminders` view.
 


thomaslombart (commented):
Thank you for the PR! The current form is somewhat confusing, as "Upcoming" could be mistaken for "Due Date." 

The use case makes sense though, can you think of a clearer name for it? Additionally, I wonder if this should be a global preference in the extension's settings for those who prefer this grouping for the whole extension (e.g. for the `Scheduled` dropdown).

vapourismo:
> The current form is somewhat confusing, as "Upcoming" could be mistaken for "Due Date."

I considered this but struggled to come up with a better name. In the end I stuck with "Upcoming" as it seemed a strictly coarser grouping than "Due Date" and different enough, especially since both are available in the same menu.

I explored some other options:

- Temporal Status or Time Status (both are from "task classification" domains)
- Timeline (implies only ordering to me)
- Timeframe (overlaps with intra-day groups like morning and so on)
- Schedule (implies more precise grouping, not consistent with Raycast's inline "Schedule")
- Horizon (not precise enough?)
- Roadmap (implies larger time scale, e.g. months)
- Pipeline (implies linear/1 at a time)
- Forecast
- Outlook
- Agenda (suggests intra-day grouping)

@thomaslombart See any you like among them?

PR #16844, opened 2025-02-05
Update toggl-track extension

AlexIsMaking:
## Description

Adds form fields to enable users to edit the start / stop dates of time entries.

## Screencast

toggl-track 2025-02-05 at 13 27 24

Also adds an action to enable users to update the start time of the running time entry. I keep forgetting to start timers so this lets me start one and then update the start time retrospectively...

toggl-track 2025-02-15 at 13 12 12

## Checklist

### DIFF
--- a/extensions/toggl-track/src/index.tsx
+++ b/extensions/toggl-track/src/index.tsx
@@ -74,9 +74,9 @@ function ListView() {
               title={timeEntry.description || "No description"}
               subtitle={(timeEntry.client_name ? timeEntry.client_name + " | " : "") + (timeEntry.project_name ?? "")}
               accessories={[
-                { text: formatSeconds(timeEntry.duration) },
                 ...timeEntry.tags.map((tag) => ({ tag })),
                 timeEntry.billable ? { tag: { value: "$" } } : {},
+                { text: formatSeconds(timeEntry.duration) },
               ]}
               icon={{ source: Icon.Circle, tintColor: timeEntry.project_color }}
               actions={


lukebars:
Nice idea! Looks good to me :shipit:

extensions/toggl-track/src/components/UpdateTimeEntryForm.tsx
```diff
@@ -33,6 +33,8 @@ function UpdateTimeEntryForm({ timeEntry, revalidateTimeEntries }: UpdateTimeEnt
   const [billable, setBillable] = useState(timeEntry.billable);
 
   const [taskSearch, setTaskSearch] = useState("");
+  const [startDate, setStartDate] = useState<Date | null>(timeEntry.start ? new Date(timeEntry.start) : null);
```
franzwilhelm, line 36: You could store `const defaultStartDate = timeEntry.start ? new Date(timeEntry.start) : null` and use this both in the useState, and in the defaultValues in form. Minor change, but a bit cleaner

franzwilhelm (changes requested):

extensions/toggl-track/src/components/UpdateTimeEntryForm.tsx
franzwilhelm: Same goes for end

franzwilhelm (approved):

AlexIsMaking:
On second thoughts, I think "Change Start Time" is a better label for the action so I've updated that now.

AlexIsMaking:
@franzwilhelm while we're looking at this, I've made one more change to my fork of the extension, to rearrange the accessories in the Start/Stop Time Entry command, to display the time entry's time after the entry's tags.
```
accessories={[
  ...timeEntry.tags.map((tag) => ({ tag })),
  timeEntry.billable ? { tag: { value: "$" } } : {},
  { text: formatSeconds(timeEntry.duration) },
]}
```
That lets me quickly check the duration of each recent entry, which is more important for me than their tags. I'm guessing you won't want to change this for everyone but just thought I'd double check?

CleanShot 2025-02-19 at 11 11 35@2x

franzwilhelm:
@AlexIsMaking I think that actually looks way better! I suggest adding that change here too

pernielsentikaer (approved):

PR #16883, opened 2025-02-07
Update jira extension to fix Git branch name format

JokeyChen:
## Description

This PR fixes an issue introduced in #16222 , where `issueSummary` is not slugified, which means when the user tries to generate a branch name, the result will contain spaces in them, making it unusable. 

For example, before this PR, if the branch name format is set to `{issueKey}-{issueSummary}`, then the result will be something like:

```
PROJ-1234-Wrong origin agent for air shipment from India
```

After this PR, it will correctly generate to something like:

```
PROJ-1234-wrong-origin-agent-for-air-shipment-from-india
```

## Checklist

### DIFF
--- a/extensions/jira/CHANGELOG.md
+++ b/extensions/jira/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Jira Changelog
 
-## [Fix Git branch name format] - {PR_MERGE_DATE}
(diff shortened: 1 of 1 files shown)


thomaslombart (approved):
Thanks for the fix! 🙌

PR #17105, opened 2025-02-16
Add favicon-generator extension

ralfboltshauser:
## Checklist

### DIFF
--- a/extensions/favicon-generator/src/favicon-generator.ts
+++ b/extensions/favicon-generator/src/favicon-generator.ts
@@ -161,8 +161,7 @@ export async function generateFavicon(
     await Promise.all([
       fs.promises.unlink(configPath).catch(() => {}),
       fs.promises.unlink(outputIndexPath).catch(() => {}),
-      ...fs
-        .readdirSync(outputDir)
+      ...(await fs.promises.readdir(outputDir))
         .filter(
           (file) =>
             file.startsWith("index-") && file.endsWith(".html") && path.join(outputDir, file) !== outputIndexPath,


LitoMore:
I just glanced at your code. Your extension would not work since you're reading environment variables from the OS.

I cloned your code to my local, and it failed with some exceptions as expected.

pernielsentikaer:
- I'm confused about this command, as a user I would not know what it did at all?

CleanShot 2025-02-28 at 11 28 25@2x

@LitoMore what didn't work for you?

@greptileai could you also check?

extensions/favicon-generator/src/favicon-generator.ts
```diff
@@ -0,0 +1,189 @@
...
+    fs.mkdirSync(tempInstallDir, { recursive: true });
+  }
+
+  try {
+    await execCommand(npxPath, ["realfavicon", "--version"], tempInstallDir);
+  } catch {
+    const npmPath = env.npm_execpath || "npm";
+    await execCommand(npmPath, ["install", "--save", "cli-real-favicon"], tempInstallDir);
+  }
+}
```
LitoMore, line 57: The `env.NPX_PATH` and `env.npm_execpath` are both `undefined` on my machine.

I don't think Raycast could read environment variables out of its box. Also, Node.js doesn't have such environment variables exported.

extensions/favicon-generator/src/check-favicon.tsx
```diff
@@ -0,0 +1,64 @@
...
+        title: "Error checking favicons",
+        message:
+          error instanceof Error && error.message.includes("undici")
+            ? "No web app found on that port"
+            : error instanceof Error
+              ? error.message
+              : "Unknown error occurred",
+      });
+    }
+  };
```
LitoMore, line 34: You might need some validation before running this. See https://developers.raycast.com/api-reference/user-interface/form#validation.

I can execute the `handleSubmit` even when some fields are empty.

extensions/favicon-generator/src/check-favicon.tsx
```diff
@@ -0,0 +1,64 @@
...
+      )}
+      <Form.TextField
+        id="port"
+        title="Port"
+        placeholder="Enter port number (e.g., 3000)"
+        defaultValue={port}
+        info="The port where your website is running locally"
+        storeValue={true}
+      />
+    </Form>
```
LitoMore, line 62: The use case is too specific. I think you can make it support any URL to cover more scenes.

LitoMore (changes requested):

extensions/favicon-generator/src/favicon-generator.ts
pernielsentikaer: You're indeed correct

extensions/favicon-generator/src/check-favicon.tsx
pernielsentikaer: I was also very confused in general actually

extensions/favicon-generator/src/favicon-generator.ts
ralf-boltshauser: this is for custom npx paths. That's exactly the purpose of getShellEnv: https://www.npmjs.com/package/shell-env?

Or am I misunderstanding smth here?

LitoMore: Raycast runs its own Node.js runtime rather than your local one. As I know, Raycast's Node.js runtime wasn't bundled with the npm.

If you want to use the user's local `npm` and `npx`, you might need to add some extension preference options to let the user manually specify the path.

You can refer to the [Brew](https://www.raycast.com/nhojb/brew) and [PM2](https://www.raycast.com/litomore/pm2) to learn how to handle user's local environment variables.

ralf-boltshauser: I mean it works, it just allows for people who have custom paths to read it. But it also works with your npx setup! If I am missing something here, just reopen!

extensions/favicon-generator/src/check-favicon.tsx
pernielsentikaer: I went through the flow but I was a but confused about that step

pernielsentikaer:
Is this ready for another review or are you still working on it

ralf-boltshauser:
ready for another review

extensions/favicon-generator/package.json
```diff
@@ -0,0 +1,66 @@
...
+      "subtitle": "Website Icons",
+      "description": "Generate favicons in multiple sizes and formats for your website",
+      "mode": "view"
+    },
+    {
+      "name": "check-favicon",
+      "title": "Check Favicon",
+      "subtitle": "Website Icons",
+      "description": "Validate favicon configuration and presence on your website",
+      "mode": "view",
```
pernielsentikaer, line 26: This should probably be a `no-view` command instead

pernielsentikaer (changes requested):
The problem here is if you have set "Pop to Root Search" to immediately then it just unloads the command, not sure if it makes sense? Maybe add a message to tell the user to check it after adding it could be better?

favicon-generator 2025-03-02 at 08 13 57

pernielsentikaer:
image

We could show a descriptive text above the textarea that tells the user what to do with it. The problem with the "Submit Form" is that many have `immediately` selected. Maybe it's better to pop to the root and tell the user that they can test it after inserting it?

ralf-boltshauser:
> image
> 
> We could show a descriptive text above the textarea that tells the user what to do with it. The problem with the "Submit Form" is that many have `immediately` selected. Maybe it's better to pop to the root and tell the user that they can test it after inserting it?

What do you mean with "they have immediately" selected? are they auto-submitting the form? If I understand that correctly, than I totally see, this would be an issue!

pernielsentikaer:
This preference will unload the extension as soon as the window loses focus. This way, when the user copies the code and wants to paste it into the code editor, it will already be unloaded.

CleanShot 2025-03-04 at 10 46 54@2x

ralf-boltshauser:
> This preference will unload the extension as soon as the window loses focus. This way, when the user copies the code and wants to paste it into the code editor, it will already be unloaded.
> 
> CleanShot 2025-03-04 at 10 46 54@2x

Ahhh I understand ... Okay yeah in this case I need to structure it differently I am thinking about moving the html tags directly to the generate step, so after generation you see the tags. And then it just tells you to go check afterwards. Thanks for the insights!!

pernielsentikaer:
You're welcome, that setting is a bit cruel for extensions, so we need to work around it 😊

PR #17127, opened 2025-02-17
Add yamli extension

menisy:
## Description

This extension allows users to convert English text of Arabic transliteration to Arabic text using the [Yamli API](https://www.yamli.com/api/docs/). It is a very useful productivity boost for users who need to write Arabic text without having an Arabic keyboard.

## Screencast

https://github.com/user-attachments/assets/11aea4d9-54b9-4657-a24a-fc32f14e6a08

## Checklist


LitoMore:
Here is a tip for you if you want to make those Arabic lines to RTL:

https://github.com/raycast/extensions/blob/c6cf3f1b47c1535de3edbec357a90ff6ddac8afd/extensions/united-nations/src/un-news.tsx#L64-L66

Here is a preview from my [United Nations](https://raycast.com/litomore/united-nations) extension:

CleanShot 2025-02-17 at 20 34 55@2x

menisy:
> Here is a tip for you if you want to make those Arabic lines to RTL:
> 
> https://github.com/raycast/extensions/blob/c6cf3f1b47c1535de3edbec357a90ff6ddac8afd/extensions/united-nations/src/un-news.tsx#L64-L66
> 
> Here is a preview from my [United Nations](https://raycast.com/litomore/united-nations) extension:
> 
> CleanShot 2025-02-17 at 20 34 55@2x

Thank you very much, that's actually quite a smart way to address this. I've considered doing this but since the input text in the search bar will have to be LTR, I think it will be a bit visually exhausting for the user to keep switching their gaze from the search bar text on the left to the list results on the right.
Is there a way to make the search bar RTL as well? Reading through the docs I couldn't find any but please correct me if I'm wrong.

LitoMore:
> Is there a way to make the search bar RTL as well?

There is no good solution yet. Raycast does not support RTL.

pernielsentikaer:
The list flickers on every keypress; is there a way to get around that @menisy?

@greptileai can you also check here

menisy:
@pernielsentikaer it's because I clear the list items and do a new API request with every new input to get the relevant translitration for the current word being input. I suppose the default selection of the second item on the list (first Arabic translitration) makes it look jumpy and causes flickering. To avoid this I can maybe keep empty placeholder list items and just change their values, or maybe change the order of the default selected item. Let me check if this improves it without affecting the intended UX.
I'll also address the AI review comments, thanks a lot 🙌🏼

LitoMore:
@pernielsentikaer This extension can convert English-Arabic transliteration to Arabic text using the Yamli API. This is useful for those people who don't have an Arabic keyboard to type Arabic words.

You can imagine that you want to write in Japanese, but your computer keyboard is in English layout. At this point, you only need to enter "kawaii" to view a list of candidate words similar in pronunciation to "kawaii" in the extension.

Back to the extension, it can help you automatically convert your words to Arabic word by word. When you finish typing, you can press <kdb>Copy Full Transliteration</kbd> full to copy the entire sentence to the clipboard.

pernielsentikaer:
So it's a bit like https://www.raycast.com/red.avtovo/punto or not quite

LitoMore:
That's quite different. The Punto Switcher is just an input method switcher. However, some users didn't have the Arabic input method installed, and they didn't have Arabic characters on the keyboard buttons.

The Arabic keyboard would be like this:



I don't know much about the real use cases; maybe @menisy could tell some.

pernielsentikaer:
Thank you for taking the time to explain, much appreciated 🙏

menisy:
Thanks a lot @LitoMore for your thorough explanation. In retrospect, I should've maybe clarified the context better in my PR description.
@pernielsentikaer The main usecase I personally go through is needing to write something in Arabic for a Google search or Youtube video search with my English only keyboard (given that I haven't learnt to touch-type in Arabic so I always get the keys mixed up). Yamli allows you to spell the Arabic words phonatically using English characters and it automatically converts it to Arabic text (i.e. transliteration).
I'll address the AI bot review comments as soon as I can and ping you for another round if that's okay 🙌🏼

menisy:
> * Missing metadata folder with screenshots for the view command as required by the store guideline

@pernielsentikaer I addressed the review bot comments. I noticed that it says I'm missing the `metadata` folder with the screenshots but I can see that it's part of the diffs so I'm wondering if that's a glitch on the bot's side.

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #17138, opened 2025-02-17
Add current-date-and-time extension

MikeshCZ:
## Description

This Raycast extension allows you to quickly insert the current date and/or time into your documents, emails or to any field. 

## Screencast

https://github.com/user-attachments/assets/3b72ed9e-b6c2-462b-b368-fdec050a007b

## Checklist

### DIFF
--- a/extensions/current-date-and-time/src/insert-date-and-time.ts
+++ b/extensions/current-date-and-time/src/insert-date-and-time.ts
@@ -29,7 +29,7 @@ function formatTime(date: Date, use24HourFormat: boolean): string {
 export default async function main() {
   const preferences = getPreferenceValues();
   const use24HourFormat = Boolean(preferences.use24HourFormat);
-  const dateFormat = ['EU', 'ISO', 'US'].includes(preferences.dateFormat) ? preferences.dateFormat : 'EU';
+  const dateFormat = ["EU", "ISO", "US"].includes(preferences.dateFormat) ? preferences.dateFormat : "EU";
 
   const now = new Date();
   const formattedDate = formatDate(now, dateFormat);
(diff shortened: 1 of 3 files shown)


MikeshCZ:
> Everything in this PR is built-in inside Raycast with snippets and dynamic placeholders, did you [play with that](https://manual.raycast.com/dynamic-placeholders#Custom%20Date%20Formats) before creating this pr?

@pernielsentikaer This is definitely a good alternative. However, it’s missing possibility to add snippet directly as menu item and the key binding cannot be set.

pernielsentikaer:
Instead of hotkeys, then you can use `keywords` to make it auto-expand snippets

MikeshCZ:
Yes, there are many ways to do this. However, in my experience, keywords don’t work in all applications.

MikeshCZ:
For example in Microsoft Excel, it can be quite frustrating when it’s not working consistently. Here is example where i type keyword and hit space. In some cases it is work, in some not. I tried also press F2 for edit column first, sometimes it is works, sometimes not.

https://github.com/user-attachments/assets/74a1eb96-ff75-410d-ad79-44b2ffaee207

pernielsentikaer:
That sounds more like a bug, do you mind reporting that to support@raycast.com with an image of your settings here
 
CleanShot 2025-02-27 at 10 06 52@2x

Would also be nice with a screencast where you have a terminal running with log stream --predicate "subsystem == 'com.raycast.macos'"

So we can see if any errors occurs

pernielsentikaer:
I think we should focus on fixing the bug rather than creating something that's already possible (Which is also mentioned in the [extension guidelines](https://manual.raycast.com/extensions))

Issue #16965, opened 2025-02-10
[BitWarden] "Search Vault" also in username


grzywek:
### Extension

https://www.raycast.com/jomifepe/bitwarden

### Description

Hello,
would that be possible to allow "Search Vault" to search also in username and not just in title? The functionality could be optional as a checkbox in the settings.

Thanks!

### Who will benefit from this feature?

It will make it easier to use the search engine, which currently does not always find what I am looking for.

jomifepe:
Hey! I'm not sure I got the issue. Isn't this already implemented? The username and URL are added to the search keywords, is it not working for you?

grzywek:
Not really. It looks like the search works only for the full username and not for a part of it.

Not found when typed part of username:
![Image](https://github.com/user-attachments/assets/3cad14a3-9bc8-45d0-a719-56c8d1f0a7cf)

Found when typed part of title:
![Image](https://github.com/user-attachments/assets/bd3f944e-0384-4d56-8039-14fc59cce0f1)

jomifepe:
Got it, that's clear, will look into it 🙂

jomifepe:
Implemented in the latest version, give it a try!

@raycastbot close this issue

Issue #16419, opened 2025-01-17
[Bitwarden Vault] When searching for keywords, increase the matching of URLs.


choby:
### Extension

https://www.raycast.com/jomifepe/bitwarden

### Description

When searching for keywords, increase the matching of URLs.

jomifepe:
Hey! I'm not sure I got the issue. Isn't this already implemented? The username and URL are added to the search keywords, is it not working for you?

jomifepe:
Implemented in the latest version, give it a try!

@raycastbot close this issue

PR #17153, opened 2025-02-18
Update bitwarden extension

marinsokol:
## Description

This MR introduces options to search vault using username or url.

By default, search using username or url is disabled. To enable, options need to be activated in `Search Vault` command preferences.

Inspired by issues [16419](https://github.com/raycast/extensions/issues/16419) and [16965](https://github.com/raycast/extensions/issues/16965)

## Checklist


jomifepe:
Hey! Thanks for contributing, but isn't this already implemented? The username and URL are added to the search keywords, is it not working for you?

marinsokol:
sorry, i did not notice that username is working, but url is not working for me.

jomifepe:
> sorry, i did not notice that username is working, but url is not working for me.

I just saw that the built-in List search is limited, and is not capable of fully fuzzy searching whole keywords, so this contribution is indeed needed.

extensions/bitwarden/src/context/search.tsx
```diff
@@ -0,0 +1,26 @@
...
+          const urls = shouldSearchUrl && item.login?.uris ? item.login.uris.map((uri) => uri.uri)?.join(" ") : null;
+          const matchesUrl = shouldSearchUrl && urls?.length ? urls.includes(searchTxt) : false;
+
+          return matchesName || matchesUsername || matchesUrl;
+        })
+      : items;
+  }, [items, searchText, shouldSearchUsername, shouldSearchUrl]);
+
+  return { items: searchedItems, folders, isLoading, isEmpty, setSearchText };
+};
```
jomifepe, line 26: 💡 Personally, I would not move the vault context here, and only use it for the search/filtering logic. This also replaces the `extractKeywords` function, so it can live in that file, and reuse part of its logic. 

Something like this:

```ts
// utils/search.ts

function hasValue(value: string | null | undefined): value is string {
  return !!(value && value !== SENSITIVE_VALUE_PLACEHOLDER);
}

function searchItems(items: Item[], searchText: string): Item[] {
  if (!searchText) return items;

  return items.filter((item) => {
    const searchableProps = new Set([item.name, ITEM_TYPE_TO_LABEL[item.type]]);
    switch (item.type) {
      case ItemType.LOGIN: {
        const { username, uris } = item.login ?? {};
        if (hasValue(username)) searchableProps.add(username);
        if (uris?.length) uris.forEach(({ uri }) => hasValue(uri) && searchableProps.add(uri));
        break;
      }
      case ItemType.CARD: {
        const { brand, number } = item.card ?? {};
        if (hasValue(brand)) searchableProps.add(brand);
        if (hasValue(number)) searchableProps.add(number);
        break;
      }
    }

    const searchTextLower = searchText.toLowerCase();
    return Array.from(searchableProps).some((p) => p.toLowerCase().includes(searchTextLower));
  });
}

export function useVaultSearch(items: Item[]) {
  const [searchText, setSearchText] = useState("");
  const filteredItems = useMemo(() => searchItems(items, searchText), [items, searchText]);
  return { searchText, setSearchText, filteredItems };
}
```

extensions/bitwarden/package.json
```diff
@@ -78,6 +79,24 @@
...
+        },
+        {
+          "name": "shouldSearchUrl",
+          "type": "checkbox",
+          "required": false,
+          "default": false,
+          "title": "Include url when searching",
+          "label": "Search by URL",
+          "description": "By default search uses only title, this will also include url."
+        },
```
jomifepe, line 99: 💡 I saw that this was mentioned in one of the issue, but personally I don't think we need to have an option for this, let's leave it as the default behavior. WDYT?

jomifepe (commented):
Thanks again for contributing, this is a good addition 🚀 

I left some suggestions, tell me what you think!

pernielsentikaer:
Testing a AI review bot too, let's see what it says

extensions/bitwarden/src/context/search.tsx
marinsokol: yep, this makes sense. Will update MR

extensions/bitwarden/package.json
marinsokol: Yeah, I was thinking about this. I added this as a safety barrier, so users do not get more results unexpectedly. I am more for removing it. If you agree, lets simplify this.

jomifepe: Yeah let's keep it simple for now.

LitoMore:
You can consider using https://npmjs.com/fast-fuzzy for a better search experience.

marinsokol:
> You can consider using https://npmjs.com/fast-fuzzy for a better search experience.

good idea, thanks. I updated MR with all suggestions and it is ready for next review.

pernielsentikaer:
@LitoMore does this work as you proprosed now?

Is this ready for a final review @marinsokol

marinsokol:
> Hi @marinsokol, here is a simpler code of `fast-fuzzy` for you:
> 
> https://github.com/LitoMore/simple-icons-figma/blob/f710a5f1d1ecf30fefde51c20187dd47c3a0053f/source/components/icons.tsx#L26-L39
> 
> You only need the `Searcher` to create a search instance. You can also use a `useMemo` to improve its performance:
> 
> ```
> import useMemo from 'react';
> import {Searcher} from 'fast-fuzzy';
> 
> const searcher = useMemo(new Searcher(icons, {
> 	keySelector(icon) {
> 		return [
> 			icon.title,
> 			icon.description,
> 			icon.fooIsArrayOfString,
> 			icon.moreContentsForSearching,
> 		]
> 			.flat()
> 			.filter(Boolean) as string[];
> 	},
> }), [icons]);
> 
> const searchResult = searchString ? searcher.search(searchString) : icons;
> ```
> 
> But it's fine to keep your current implementation. That's up to you.

good idea, code looks much cleaner now that i updated MR

pernielsentikaer:
Do you want to test this a last time @jomifepe

jomifepe:
> Do you want to test this a last time @jomifepe

Yes, will take a final look soon 🙂

extensions/bitwarden/src/utils/hooks/useVaultSearch.ts
```diff
@@ -0,0 +1,38 @@
...
+        ? searcher.search(searchText, {
+            ignoreCase: true,
+            threshold: 0.5,
+            sortBy: sortKind.bestMatch,
+          })
+        : items,
+    [searcher, searchText]
+  );
+  return { setSearchText, filteredItems };
+}
```
jomifepe, line 38: Some requests/suggestions:
1. Let's add `ignoreSymbols`, I want to be able to write '@' and see emails, for example, without it was showing nothing.
2. Increase the `threshold`. This was trial and error, below 0.7 I was getting way to much single character results, and since (I think) there's no way to add weights to the keys, 0.7 was the minimum where I started to see it giving more importance to the title.
3. We can move the searcher options to the constructor. This is not just preference, I found that `ignoreSymbols` was being ignored (pun not intended) when passing it to the `search` function, rather than the class.
4. Removed the cast and used an `is string` type predicate.
5. General formatting preferences (braces and returns).

```suggestion
export function useVaultSearch(items: Item[]) {
  const [searchText, setSearchText] = useState("");

  const searcher = useMemo(() => {
    return new Searcher(items, {
      ignoreSymbols: false,
      threshold: 0.7,
      sortBy: sortKind.bestMatch,
      keySelector: (item) => {
        const { login, card } = item;
        return [
          item.name,
          ITEM_TYPE_TO_LABEL[item.type],
          login?.username,
          login?.uris?.map(({ uri }) => uri),
          card?.brand,
          card?.number,
        ]
          .flat()
          .filter((value): value is string => !!value);
      },
    });
  }, [items]);

  const filteredItems = useMemo(() => {
    if (!searchText) return items;
    return searcher.search(searchText);
  }, [searcher, searchText]);

  return { setSearchText, filteredItems };
}
```

extensions/bitwarden/src/utils/hooks/useVaultSearch.ts
```diff
@@ -0,0 +1,38 @@
+import { useMemo, useState } from "react";
+import { Searcher, sortKind } from "fast-fuzzy";
+import { ITEM_TYPE_TO_LABEL } from "~/constants/labels";
+import { Item } from "~/types/vault";
+
+export function useVaultSearch(items: Item[]) {
```
jomifepe, line 6: Let's move this to `utils/search.ts` and delete `extractKeywords` and its usages, we don't need the keywords anymore.

jomifepe (commented):
Great job, left some final suggestions! After those it should be ready to ship 🚀 

Thank you @LitoMore for the `fast-fuzzy` suggestion, looks clean this way!

extensions/bitwarden/src/utils/hooks/useVaultSearch.ts
marinsokol: fixed

extensions/bitwarden/src/utils/hooks/useVaultSearch.ts
marinsokol: fixed

jomifepe (approved):
Looking great, thanks for implementing this :shipit:

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #17169, opened 2025-02-19
create new extension `new`

ViGeng:
## Description

As in [.New](https://whats.new/shortcuts/), `New` extension allows you to new common stuff with the ease of several strokes.

## Screencast

new-3

## Checklist


xmok:
@ViGeng, we already have [dot-new](https://www.raycast.com/natterstefan/dot-new) extension that does something very similar to what you've come up with. It might be better if you updated the existing extension rather than submit a new one that does the same thing. For example, adding the `Dropdown` to the existing extension could be a cool update!

ViGeng:
> @ViGeng, we already have [dot-new](https://www.raycast.com/natterstefan/dot-new) extension that does something very similar to what you've come up with. It might be better if you updated the existing extension rather than submit a new one that does the same thing. For example, adding the `Dropdown` to the existing extension could be a cool update!

Hi xmok, thanks for the information. I did not notice that. Will rethink about it. :)

PR #17297, opened 2025-02-24
Add unducklink extension

jafupy:
## Description

This extension allows for the use of unduck.link, which is a "search engine" that is essentially google with !bangs, from raycast. The extension creates a single command: search. It takes in 1 arg, which is the search query and it opens unduck.link/search?q={query} in your browser. Thats it!

## Screencast
Raycast 2025-02-24 at 20 38 59

## Checklist

### DIFF
--- a/extensions/unducklink/package.json
+++ b/extensions/unducklink/package.json
@@ -2,7 +2,7 @@
   "$schema": "https://www.raycast.com/schemas/extension.json",
   "name": "unducklink",
   "title": "Unduck.link",
-  "description": "Access Theo's (t3.gg) unduck.link search engine through Raycast",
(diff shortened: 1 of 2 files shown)


extensions/unducklink/package.json
```diff
@@ -0,0 +1,51 @@
+{
+  "$schema": "https://www.raycast.com/schemas/extension.json",
+  "name": "unducklink",
+  "title": "Unduck.link",
+  "description": "Access Theo's (t3.gg) unduck.link search engine through Raycast",
```
katafrakt, line 5: This is certainly not a search engine. Would be great to have something less misleading in the description.

pernielsentikaer:
This seems to be the same as #17418. Could we work together on creating one by taking the best of both?

PR #17383, opened 2025-02-27
Add image-flow extension

godruoyi:
## Checklist


pernielsentikaer (commented):
This looks like a really niche extension, it might be a bit difficult for users to set it up as is? I'm not that much into what it does

@greptileai can you also check this

pernielsentikaer:
Did you see my question above?

godruoyi:
> Did you see my question above?

@pernielsentikaer  Do you mean "it might be a bit difficult for users to set it up as is"?  

Yes, the end user must set up a workflow file, but they can easily copy it from [here](https://github.com/godruoyi/imageflow/blob/master/workflow.yaml). I think this extension isn't very user-friendly for non-developers, so yes, it could be a little more difficult if the user has no development experience.

```
curl https://raw.githubusercontent.com/godruoyi/imageflow/master/workflow.yaml | tee ~/workflow.yaml

```

godruoyi:
If it is valuable, then it should also be acceptable, regardless of whether it's niche or not, right? And my friends around me quite appreciate this extension.

extensions/image-flow/README.md
```diff
@@ -0,0 +1,121 @@
+## Description
```
pernielsentikaer, line 1:
```suggestion
# Image Flow
```

pernielsentikaer (changes requested):
Could you look into all the suggestions you received from me and Greptile? We always try to create a good UI/UX for users, and this is indeed hard to set up.

extensions/image-flow/README.md
godruoyi: updated, PTAL @pernielsentikaer thank you,

pernielsentikaer:
Would it make sense to add it as a private extension, so all of your friends can sign up for that org? Just wondering 🤔

Do you think anyone other than your closest friends would use this?

godruoyi:
@pernielsentikaer  I'm really curious why you think this extension is unacceptable. Is it because its configuration is too complex for users? But the Raycast ecosystem has many extensions that require users to do a lot of additional setup, like installing extra apps or configuring API keys. I still stand by my point: for the Raycast community, if an extension is valuable and helps people improve efficiency, then it should be acceptable—regardless of how niche it is. Let me know your thoughts, and perhaps you could also have others on your team take a look at this.

extensions/image-flow/README.md
pernielsentikaer: I don't see that this is changed?

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

Let's see how it goes

extensions/image-flow/README.md
godruoyi: thank you @pernielsentikaer , update and add some screenshots metadata support now, can you PTAL, thanks.

PR #17386, opened 2025-02-27
add native features to exa extension

abizer:
## Description

Raycast extension for the Exa AI API, using native raycast UI semantics (vs the other exa extension, which just redirects to the browser)

## Screencast

exa-1

## Checklist


thomaspaulmann:
Hey @abizer, thanks for the contribution. How about we combine both extensions? Your commands would be a great addition for the extising extension which focuses on it's integration with Raycast AI. Would you be open to contribute to it? This way, users see a single extension and don't need to decide which one they want to install.

thomaspaulmann (changes requested):
Waiting for response from @abizer about merging both extensions to guarantee a smooth user experience.

abizer:
@thomaspaulmann how would you propose merging them? feels like it would mostly make sense to just merge some of the tool calling stuff from that one into this one rather than the other way around given that this implementation is a lot more fleshed out

thomaspaulmann:
> @thomaspaulmann how would you propose merging them? feels like it would mostly make sense to just merge some of the tool calling stuff from that one into this one rather than the other way around given that this implementation is a lot more fleshed out

I would propose it the other way around for the following reasons:
- The other extension is the official one built in collaboration with the Exa team
- The other extension uses our proxy, allowing users to use the extension without requiring an API key
- The other extension is already installed for some users, meaning when you add your changes it will give them a nice update

Let me know what you think.

abizer:
i mean practically speaking: there's fewer features in that extension vs this one, so realistically a merger would basically involve moving ai.yaml and the tool call references from package.json into this one. we can configure this to use the proxy if necessary (although I'm loathe to do something like that without making it extremely obvious to the user that their requests are being intercepted and proxied rather than going directly to exa's servers, for privacy and security reasons). it doesn't really matter to me the name of folder the extension code ultimately ends up

abizer:
@thomaspaulmann @pernielsentikaer I've updated the PR to merge the new extension into the old extension.

thomaspaulmann (changes requested):
> i mean practically speaking: there's fewer features in that extension vs this one, so realistically a merger would basically involve moving ai.yaml and the tool call references from package.json into this one. we can configure this to use the proxy if necessary (although I'm loathe to do that without making it extremely obvious to the user that their requests are being intercepted and proxied rather than going directly to exa's servers, for privacy and security reasons). it doesn't really matter to me the name of folder the extension code ultimately ends up in 

Hey, this is mainly to keep the functionality in one place. Given the other extension is already in the store and was build in collaboration with Exa, I think it would be best ot merge it over there.

Can we keep the name the same? Otherwise it won't replace the extension and existing users get the update.

abizer:
yep i did as you suggested, the exa-ai extension has been removed and the changes merged into exa

abizer:
@thomaspaulmann @pernielsentikaer just bumping this for your attention

pernielsentikaer:
Did you remove this on purpose @abizer

abizer:
> Did you remove this on purpose @abizer

replaced with #18673

PR #17394, opened 2025-02-27
Update doccheck extension

h16nning:
## Checklist

### DIFF
--- a/extensions/doccheck/CHANGELOG.md
+++ b/extensions/doccheck/CHANGELOG.md
@@ -1,6 +1,6 @@
 # DocCheck Changelog
 
-# [Version 2.1.2] - {PR_MERGE_DATE}
+# [Version 2.1.2] - 2025-03-11
 
 - Fixed: articles with neither author nor aliases weren't shown when searching for them
 


h16nning:
@justspacedog thank you for this great extension, It is very helpful! I have, however, continued to stumble upon one  bug 🐛 If a docccheck page has an empty `title_alias` array and `author` is an empty string, it won't be shown when searching for it using `Search Flexikon`. I have no idea why the `author` field is empty on the so many and surprisingly prominent pages like _Insulin_, _Kardiologie_ or  _Gentamicin_ (although that is not our problem). Nevertheless, finding these pages in DocCheck is impossible right now (or am I missing something 🤔)

I just added one more else case at the end to catch all entries that don't fullfill the requirements? Are you ok with this change?

h16nning:
As a side note: I personally suspect that these entries were all part of doccheck when it originally launched / some legacy pages and nobody cared to update metadata. They all lack`date_publish` as well (at least those I checked against) 😅

Most of them are still shown as they usually have aliases, but those that don't arent't shown

justspacedog:
Hi @h16nning,

Thank you for pointing that out! Could you maybe just add `date_publish` at the end of the list item? Or is this field empty as well?

h16nning:
I have checked a few dozen pages and they all lack `date_publish`. Still, I could add it conditionally, just to make sure. :)

justspacedog:
Looks good to me, but I think someone else needs to approve it.

h16nning:
As I was fiddling around, I was noticing another issue with the extension 😅 (not only in this pr but also in the live version)

On the "start page" all history items will have an `Invalid Date` tag. Did some digging around and found out that you determine `date_publish` in `HistoryItem` differently than in the search menu, were it just uses the data provided by the query. Instead it is grabbed from the page like this, right?

```ts
let date_publish = "";
$(".has-gap-1").each((i, el) => {
    date_publish = $(el).find(".is-black ").text().trim();
});
```

However, `date_publish` ends up getting the value `Wichtiger Hinweis zu diesem Artikel`, so not quite what is intended. Appearantly DocCheck changed there layout here a bit from this

Bildschirmfoto 2025-03-03 um 22 13 12

to this

Bildschirmfoto 2025-03-03 um 22 05 03

Now it is hidden in the little (i) menu…

Though I haven't figured out yet, where in the dom this popup hides. Can look at it again, as soon as I find more time.

pernielsentikaer:
Let me know when this is ready 😊

h16nning:
@pernielsentikaer should be all ready :)

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #17412, opened 2025-02-27
Add dev-toolkit extension

mblode:
## Description

Dev Toolkit brings essential developer utilities to your fingertips. Quickly generate UUIDs, format JSON, convert between encodings, transform text case, and handle timestamps all from Raycast. A comprehensive toolbox for developers that eliminates the need for multiple browser tabs and standalone applications.

## Checklist

### DIFF
--- a/extensions/dev-toolkit/src/utils.ts
+++ b/extensions/dev-toolkit/src/utils.ts
@@ -68,7 +68,7 @@ export const safeNumberArg = async (
 
 export const showError = async (msg: string) => {
   await closeMainWindow();
-  await showToast(Toast.Style.Failure, msg);
+  await showFailureToast(msg);
 };
 
 export const produceOutput = async (content: string) => {


pernielsentikaer (commented):
Hi 👋

It looks a bit like [Toolbox](https://www.raycast.com/Kang/toolbox), did you play with that one? We're trying to limit similar extensions a bit and it looks like this is pretty close to it

mblode:
I looked through a lot of extensions but I must have missed that one! I can see overlap but the UI has some core differences in the following ways:

- This uses top level commands instead of 1 top level "Toolbox" command
- The input comes from the clipboard instead of having to type it out manually
- The output goes straight to the clipboard instead of having to run an extra click

So my extension is a more straightforward all-in-one utility with as minimal steps as possible

pernielsentikaer:
It feels like an extension where most of the commands should be disabledByDefault, what do you think about that?

@greptileai could you also check here

pernielsentikaer:
Do you need help fixing the CI errors?

PR #17436, opened 2025-02-28
Add Unduck extension

lacherogwu:
## Description

A fast, local-first \"search engine\" for !bang users

As of now, two pull requests have been created to add this extension. However, with all due respect to the developers, their implementation is not optimal. They are not using it locally, and as you know, Raycast is a state-of-the-art software that deserves state-of-the-art extensions. 🌟

Here are references to those pull requests:
- #17297
- #17406 

[EDIT]
I just found another pull request that also implements local functionality. It takes a different approach and no longer resembles the source. Additionally, it uses a database, which might not be necessary, and doesn't fully utilize Raycast's APIs, which is unfortunate.

- #17418

# Unduck

DuckDuckGo's bang redirects are too slow. Add the following URL as a custom search engine to your browser. Enables all of DuckDuckGo's bangs to work, but much faster.

```
https://unduck.link?q=%s
```

## How is it that much faster?

DuckDuckGo does their redirects server side. Their DNS is...not always great. Result is that it often takes ages.

I solved this by doing all of the work client side. Once you've went to https://unduck.link once, the JS is all cache'd and will never need to be downloaded again. Your device does the redirects, not me.

## Credits

[Unduck](https://github.com/t3dotgg/unduck) by [Theo](https://github.com/t3dotgg)

## Screencast
You can use it regularly, and I highly recommend it as a fallback command.

https://github.com/user-attachments/assets/4577090b-e101-4299-914e-2c1c7c95bfec

CleanShot 2025-02-28 at 15 31 03@2x

## Checklist


franzwilhelm:
@lacherogwu. I found that using `Array.prototype.find` with Raycast sporadically caused the search to cause memory issues, especially after performing multiple queries in a short time period. This is the reason I went for the SQLite approach. 

Also my PR adresses searching through bangs. I know that Racast discourages use of custom settings. That being said, they also allow for connecting with external APIs where the external service stores the config values. In my opinion my extension provides better UX by letting the user browse through the bangs and selecting their default easily, instead of having to remember and manually enter the bang in preferences. That being said, if the Raycast team wants this to change in my PR I'm happy to, and will change it in 5mins.

Lastly, I submitted my PR way earlier and with more functionality. I suggest that we instead collaborate from my branch and adress the review from RC together. Clearly we had the same idea, and I'm happy to make the best possible extension together

franzwilhelm:
@lacherogwu I've now done some changes and referenced you in my PR. Please have a look and let me know what you think!

pernielsentikaer:
Closing this one since #17518 is in the store

PR #17459, opened 2025-03-01
Update brew.ts

Rendusaz:
Custom Brew Executable Path, Prefix Path mix

## Brew Path ##

I had issues with the "Custom Brew Executable Path" , upon inspection it appears it should ask for brew prefix, not executable path, so  I created this fix I applied here. Which simply specifies the exact location of the brew executable, or defaults back if incorrect. This works, confirmed I can switch back and forth between /usr/*/*/brew and /opt/*/*/brew

### DIFF
--- a/extensions/brew/src/brew.ts
+++ b/extensions/brew/src/brew.ts
@@ -106,8 +106,10 @@ export interface InstalledMap {
 /// Paths
 
 export const brewPrefix = (() => {
-  if (preferences.customBrewPath && preferences.customBrewPath.length > 0)
-    return path_join(preferences.customBrewPath, "..", "..");
+  if (preferences.customBrewPath && preferences.customBrewPath.length > 0) {
+    // Use the directory containing the provided executable path.
+    return path_join(preferences.customBrewPath, "..");
+  }
   try {
     return execSync("brew --prefix", { encoding: "utf8" }).trim();
   } catch {
@@ -116,7 +118,7 @@ export const brewPrefix = (() => {
 })();
 
 export const brewPath = (suffix: string) => path_join(brewPrefix, suffix);
-export const brewExecutable = () => brewPath("bin/brew");
+export const brewExecutable = () => preferences.customBrewPath && preferences.customBrewPath.length > 0 ? preferences.customBrewPath : brewPath("bin/brew");
 
 /// Fetching
 


extensions/brew/src/brew.ts
```diff
@@ -116,7 +118,7 @@ export const brewPrefix = (() => {
 })();
 
 export const brewPath = (suffix: string) => path_join(brewPrefix, suffix);
-export const brewExecutable = () => brewPath("bin/brew");
+export const brewExecutable = () => preferences.customBrewPath && preferences.customBrewPath.length > 0 ? preferences.customBrewPath : brewPath("bin/brew");
```
nhojb, line 121: Thanks for spotting the omission here 😄

extensions/brew/src/brew.ts
```diff
@@ -106,8 +106,10 @@ export interface InstalledMap {
 /// Paths
 
 export const brewPrefix = (() => {
-  if (preferences.customBrewPath && preferences.customBrewPath.length > 0)
-    return path_join(preferences.customBrewPath, "..", "..");
+  if (preferences.customBrewPath && preferences.customBrewPath.length > 0) {
+    // Use the directory containing the provided executable path.
+    return path_join(preferences.customBrewPath, "..");
```
nhojb, line 111: I'm not sure that this is correct. I wonder if there is just confusion about the custom brew path setting?

For example the output of `brew --prefix` would normally be `/opt/homebrew` (on Apple Silicon)

If a user specifies a custom brew executable path, it might look like: `/custom/prefix/bin/brew`. Using the original logic this would result in a brew prefix of `/custom/prefix`. Using the new logic we'd end up with `/custom/prefix/bin` - which isn't correct.

Or if a user set `/opt/homebrew/bin/brew` as their custom executable path, we'd end up with a prefix of `/opt/homebrew/bin` as the brew prefix - which clearly isn't correct.

Screenshot 2025-03-01 at 14 31 10

nhojb (changes requested):
Rendusaz thanks for looking into this! I've added comments, but I think only part of your fix is correct - the missing check for `customBrewPath` when setting the `brewExecutable`.

extensions/brew/src/brew.ts
nhojb: Actually, now I think about it, this change is required. Or at least it should not be if the custom brew path is set correctly since `brewPath("bin/brew") == customBrewPath`.

PR #17472, opened 2025-03-01
[extensions/get-favicon] new feature: specify icon size

ViGeng:
## Description

- Added support for specifying icon size, you can set the size of the icon in the settings.
- Minor refactoring: extracted common interfaces and preferences to separate files.

image

### DIFF
--- a/extensions/get-favicon/CHANGELOG.md
+++ b/extensions/get-favicon/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Get Favicon Changelog
 
-## [Add support for specifying icon size] - {PR_MERGE_DATE}
+## [Add support for specifying icon size] - 2025-03-03
 
 - Added support for specifying icon size, you can set the size of the icon in the settings.
 - Minor refactoring: extracted common interfaces and preferences to separate files.


extensions/get-favicon/src/common.tsx
```diff
@@ -0,0 +1,8 @@
+export interface Arguments {
+  url: string;
+}
+
+export interface Preferences {
+  default_icon_size: number;
+  downloadDirectory: string;
+}
```
pernielsentikaer, line 8: Can be skipped:

https://developers.raycast.com/api-reference/preferences
https://developers.raycast.com/information/lifecycle/arguments

ViGeng: Hi, `Preference` can be removed.
I don't understand the second one. How should I use the argument? Thanks.

pernielsentikaer (changes requested):
You're right, it uses an very old version of @raycast/api

Can you try to update with `npx @raycast/migration@latest .` and fix if any errors occurs?

That should give you `raycast-env.d.ts` which has the type for Preference and Arguments

vimtor (approved):

pernielsentikaer:
Do you need help fixing the errors?

ViGeng:
> You're right, it uses an very old version of @raycast/api
> 
> Can you try to update with `npx @raycast/migration@latest .` and fix if any errors occurs?
> 
> That should give you `raycast-env.d.ts` which has the type for Preference and Arguments
At very beginning I was trying to use the type from the generated Arguments types in "raycast-env.d.ts".
But it's reporting it can not find the type. That's why I ask how I can use the arguments.

Finally I found the generated "raycast-env.d.ts" is not included in the tsconfig.json.
It's a bit confusing for a frontend layman 😅 But thanks to ChatGPT.

image

ViGeng:
> Do you need help fixing the errors?

Hi, @pernielsentikaer , I need some help regarding the dependencies. I have no idea what is going on there. thx!

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #17473, opened 2025-03-01
Update visual-studio-code extension

lacherogwu:
## Problem

The extension was experiencing performance issues due to multiple calls to `getApplications()` API. This function is computationally expensive and was being called repeatedly for each item in the list, causing UI blocking and triggering Raycast's rendering loop warning:

```
Looks like the command is rendering a lot without any changes. This might indicate a rendering loop caused by a `setState` not being wrapped in a `useEffect`.
```

The root cause was that each `LocalItem` component was independently calling `getEditorApplication()`, which in turn called `getApplications()`. With many items in the list, this created significant performance overhead and UI responsiveness issues.

## Solution

Implemented a simple caching mechanism for the `getApplications()` function using the lightweight `cache-func` utility. This ensures that regardless of how many components request application data, the expensive API call is only made once, with subsequent calls returning the cached result.

```ts
import { Application, getApplications } from "@raycast/api";
import { cacheFunc } from "cache-func";

const cachedGetApplications = cacheFunc(getApplications);

export async function getEditorApplication(buildName: string): Promise<Application | undefined> {
  const apps = await cachedGetApplications();
  // Rest of the function...
}
```

This change significantly improves performance and responsiveness of the extension, especially for users with many recent projects.

## Screencast

https://github.com/user-attachments/assets/de7e14e8-ba1d-4106-b361-6aacf571ad73

## Checklist

### DIFF
--- a/extensions/visual-studio-code-recent-projects/CHANGELOG.md
+++ b/extensions/visual-studio-code-recent-projects/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Visual Studio Code Changelog
 
-## [Performance] - {PR_MERGE_DATE}
+## [Performance] - 2025-03-02
 
 - Fixed performance issue in ⁠Search Recent Projects command by implementing caching for application data retrieval. This significantly improves responsiveness when displaying many recent projects and eliminates rendering loop warnings.
 


VatsalSy:
Looks good to me. I tested it out locally and it is working fine.

VatsalSy (approved):

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #17475, opened 2025-03-01
Update supernotes extension to enable AI tools

mattdelliott:
## Description

- Enables AI Extension and allows for searching for cards, append to an existing card, or creating a new card.

## Screencast
https://cln.sh/FRWgv49yw1XjNh0PX0KN

## Checklist

### DIFF
--- a/extensions/supernotes/package.json
+++ b/extensions/supernotes/package.json
@@ -4,7 +4,7 @@
   "title": "Supernotes",
   "description": "Interact with the Supernotes platform in Raycast.",
   "icon": "supernotes-icon.png",
-  "author": "acnebs",
+  "author": "connebs",
   "owner": "supernotes",
   "access": "public",
   "license": "MIT",
--- a/extensions/supernotes/src/tools/append-card.ts
+++ b/extensions/supernotes/src/tools/append-card.ts
@@ -8,7 +8,7 @@ type Input = {
   id: string;
 
   /**
-   * The content markup to append to the card. If not provided then ask for it.
+   * The content markup to append to the card. Required.
    */
   content: string;
 };
@@ -29,7 +29,7 @@ export default async function (input: Input) {
   });
 
   if (!fetched.ok) {
-    throw new Error("There was a problem appending to the card.");
+    throw new Error(`There was a problem appending to the card: ${fetched.body.detail}`);
   }
 
   return fetched.body;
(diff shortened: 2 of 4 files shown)


connebs:
Hey @mattdelliott, thanks for this. We were just discussing integrating the new Raycast AI features with the Supernotes extension, looks like you've beat us to it!

The only issue is that the date formatting feature unfortunately cannot work as written, as only a certain date format (ISO 8601 without a time) is accepted as the `local_date` property in the daily append endpoint. The good news is that we are actually already in the process of adding a standardized date format option to Supernotes itself, which will be coming in our next release. 

However this would be a good time to better standardize that call so that it matches other integrations, which use two digits for month and day (this was actually an oversight when we first added `local_date` to the Raycast extension). So if you could remove the date formatting option feature, but leave in the formatting call with a strict date like below, that would be great.

```ts
const localDate = format(new Date(), "yyyy-MM-dd")
```

mattdelliott:
Hey @connebs! That is great to hear. I removed that option and hardcoded the `yyyy-MM-dd` format. I could remove the `date-fns` dependency and just use `padStart` on the original way it was written if you want. Let me know if you would prefer for me to do that.

As far as the AI Extension goes, I just got some basics to get started, but I am sure there will be a lot more awesome use cases to implement.

connebs:
Good point, I think for now `padStart` would be better since we are otherwise not using other features of `date-fns`

mattdelliott:
Sounds good. I've made those changes.

pernielsentikaer:
Why did this extension change `author`, is that on purpose?

mattdelliott:
It was giving an error saying the user no longer existed. So I changed it to what I thought was perhaps his new handle. I can change it back though.

pernielsentikaer:
Just checked and you're right, it's `connebs`🙂

connebs:
I think this is all good to go if you are happy with this @peduarte

thomaspaulmann (approved):
All looks good and sorry for the delay here. We've been swamped with reviews lately and making our way through it. I'll approve the changes but want to hold-off until we know if we need to do anything on our side to make sure that existing users get update of the extension due to the change of the author...

thomaspaulmann:
im-ready-lets-go-gif-by-leroy-patterson

Let's ship!

mattdelliott:
@thomaspaulmann it looks like it failed to deploy due to the author You may already be aware, but thought I would ping you just in case.

mattdelliott:
Looks good @pernielsentikaer, thanks! 🎉

PR #17493, opened 2025-03-02
[coffee] show caffeinate duration in menu bar item

kargeor:
## Screencast
image

## Checklist

### DIFF
--- a/extensions/coffee/src/index.tsx
+++ b/extensions/coffee/src/index.tsx
@@ -1,20 +1,49 @@
 import { Color, LaunchProps, MenuBarExtra, getPreferenceValues, showHUD } from "@raycast/api";
 import { useExec } from "@raycast/utils";
 import { useEffect, useState } from "react";
-import { startCaffeinate, stopCaffeinate } from "./utils";
+import { formatDuration, startCaffeinate, stopCaffeinate } from "./utils";
 
-function useExtraInfoStr(): string {
-  const { data } = useExec("ps -o args= -p $(pgrep caffeinate)", [], {
+function parseEtime(etime: string): number {
+  const parts = etime.split(":").reverse();
+  const seconds = parseInt(parts[0]) || 0;
+  const minutes = parseInt(parts[1]) || 0;
+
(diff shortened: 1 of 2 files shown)


pernielsentikaer:
Could we make the time prettier instead of showing 3600s it should be nice to show 1h and string like 5h 30m instead

kargeor:
> Could we make the time prettier instead of showing 3600s it should be nice to show 1h and string like 5h 30m instead

Is there a pre-existing function somewhere that does this that I can import?

pernielsentikaer:
You can maybe do something like
```
function formatDuration(seconds: number): string {
  const units = [
    { label: 'd', value: 86400 },
    { label: 'h', value: 3600 },
    { label: 'm', value: 60 },
    { label: 's', value: 1 }
  ];

  const result: string[] = [];

  for (const unit of units) {
    const amount = Math.floor(seconds / unit.value);
    seconds %= unit.value;
    if (amount > 0) {
      result.push(`${amount}${unit.label}`);
    }
  }

  return result.join(' ');
}

// Example usage:
console.log(formatDuration(1337000)); // Output: "15d 11h 23m 20s"
console.log(formatDuration(3600));    // Output: "1h"
console.log(formatDuration(65));      // Output: "1m 5s"
console.log(formatDuration(86400));   // Output: "1d"
```

zakj:
Won't this be a bit confusing after some time progresses? For example, if I caffeinate for 2 hours, when there's one minute left to go it will still display "For 7200s".

kargeor:
> Won't this be a bit confusing after some time progresses? For example, if I caffeinate for 2 hours, when there's one minute left to go it will still display "For 7200s".

I updated it to show the remaining time. Any advice on wording/format would be appreciated. (see updated screenshot)

zakj:
Thank you, that's so much better! I think "remain" is fine. Other alternatives would be "left", or nothing at all—the time on its own is pretty clear in this context.

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

Issue #17188, opened 2025-02-20
[Safari] ...Search bookmarks not working


ksandvik:
### Extension

https://www.raycast.com/loris/safari

### Raycast Version

1.92.0

### macOS Version

15.3.1

#### Error:
```
Error: maxObjectCount exceeded

y1.parseBuffer:bplistParser.js:92:10

---
89:   }
90: 
91:   if (numObjects > exports.maxObjectCount) {
92:     throw new Error("maxObjectCount exceeded");
93:   }
94: 
95:   // Handle offset table
---

iv:search-bookmarks.js:157710:24580
search-bookmarks.js:157710:25123
FSReqCallback.readFileAfterClose:context:68:3
```

### Steps To Reproduce

Open search bookmarks

### Current Behaviour

Nothing happens

### Expected Behaviour

Show initial browser before starting to look at entries, but the errror above shows the issue with max object count exceeeded, some of us have a lot of bookmarks... And this works inside Safari.

axsuul:
@raycastbot remove from list

ChangeHow:
Can you provide more info about this crash? More detailed logs or screen recording is better

ksandvik:
I would think that the Javascript log gives enough clues that a certain limit is exceeded:
```
  if (numObjects > exports.maxObjectCount) {
92:     throw new Error("maxObjectCount exceeded");
93:   }
```
but hey here are the screenshots. 

![Image](https://github.com/user-attachments/assets/94a6b6b3-2c7a-4119-8e20-525c5826daaf)
![Image](https://github.com/user-attachments/assets/51d15190-c867-40da-8dee-3af692d088ba)

ksandvik:
Or if nothing else provide us with a setting to increase the maxObject count (in case it's a static global value) to go around this bug.

ChangeHow:
@raycastbot assign me

ChangeHow:
> [@ChangeHow](https://github.com/ChangeHow) duplicates [#14384](https://github.com/raycast/extensions/issues/14384) [#13833](https://github.com/raycast/extensions/issues/13833)
> 
> it's limitation of https://github.com/joeferner/node-bplist-parser

thank you! I just found the source code in bplist-parser.

ksandvik:
Yes, send me the fixed version and I install it here for testing.

ChangeHow:
> Yes, send me the fixed version and I install it here for testing.

[...]

Issue #14384, opened 2024-09-08
[Safari] Search Bookmarks command doesn't work


alexschcom:
### Extension

https://www.raycast.com/loris/safari

### Raycast Version

1.82.3

### macOS Version

14.6.1

### Description

Whenever I run the command, I get an error message saying "Something went wrong". Error log below:

#### Error:
```
Error: maxObjectCount exceeded

He.parseBuffer:bplistParser.js:92:10

---
89:   }
90: 
91:   if (numObjects > exports.maxObjectCount) {
92:     throw new Error("maxObjectCount exceeded")

### Steps To Reproduce

1. Open Raycast
2. Run command "Search Bookmarks"

### Current Behaviour

_No response_

### Expected Behaviour

_No response_

iwex:
@alexschcom hello! you have more bookmarks than https://github.com/joeferner/node-bplist-parser can parse.

alexschcom:
I see -- `maxObjectCount` is over 32K but I'm not familiar with plist files and didn't realize there's multiple objects/bookmark, decreasing the total number of "allowed" bookmarks significantly.

FYI for anyone else running into this issue in the future: I deleted some old bookmarks, getting my total down to ~2920 and it finally worked. It seems like the number of objects/bookmark can vary, but this should give you some idea...

Issue #13833, opened 2024-08-01
[Safari] fails to run


nikitavoloboev:
### Extension

https://www.raycast.com/loris/safari

### Raycast Version

1.79.1

### macOS Version

15.0

#### Error:
```
Error: maxObjectCount exceeded

He.parseBuffer:bplistParser.js:92:10

---
89:   }
90: 
91:   if (numObjects > exports.maxObjectCount) {
92:     throw new Error("maxObjectCount exceeded")

### Steps To Reproduce

1. try run extension, break

iwex:
@nikitavoloboev I see that you know JS, could you please debug by yourself
Problem is somewhere here https://github.com/joeferner/node-bplist-parser/blob/a76affe9caec05a3a47c5e02494de661b7204816/bplistParser.js#L90

nikitavoloboev:
how can I add debug statements to this code, i.e. run this extension locally

I am new to raycast extension dev

iwex:
It's easy as 1,2,3 :)

![HROdYItz](https://github.com/user-attachments/assets/fe11613d-6a79-442e-b8fe-e2d5f3b20468)
Choose dir and proceed.

1) Run `npm install` and `npm run dev`
2) Open extension with your code editor and feel free to change everything you want.

nikitavoloboev:
Thank you.

After debugging, issue was that my `~/Library/Safari/Bookmarks.plist` was too large so it was breaking. I deleted some bookmarks and it worked again.

PR #17494, opened 2025-03-02
Update safari extension

ChangeHow:
## Description

Improvements:
- Speed up extension by loading plugins asynchronously.
- Support `parseBookmarksWithGo` option to unlimit the size of bookmarks
  - https://github.com/raycast/extensions/issues/17188
  - https://github.com/raycast/extensions/issues/14384
  - https://github.com/raycast/extensions/issues/13833

## Checklist

### DIFF
--- a/extensions/safari/src/lang-adaptor/pinyin.ts
+++ b/extensions/safari/src/lang-adaptor/pinyin.ts
@@ -17,12 +17,12 @@ export class PinyinHandler implements LanguageHandler {
   serialize(text: string): string {
     const chineseRegex = new RegExp(PinyinHandler.chineseChar, "g");
     const chineseChars = Array.from(text.matchAll(chineseRegex));
+    // improve the performance by importing pinyin in runtime
+    // eslint-disable-next-line @typescript-eslint/no-var-requires
+    const pinyin = require("pinyin").default; // import pinyin from 'pinyin'
     if (chineseChars.length > 0) {
       return chineseChars.reduce((formatted, matchItem) => {
         const [char] = matchItem;
-        // improve the performance by importing pinyin in runtime
-        // eslint-disable-next-line @typescript-eslint/no-var-requires
-        const pinyin = require("pinyin").default; // import pinyin from 'pinyin'
         const pinyinCollection = pinyin(char, { style: pinyin.STYLE_NORMAL });
         return formatted.replace(char, pinyinCollection.join("") + " ");
       }, text);
--- a/extensions/safari/src/reading-list.tsx
+++ b/extensions/safari/src/reading-list.tsx
@@ -38,6 +38,7 @@ export default function Command() {
           bookmarks,
           [
             { name: "title", weight: 3 },
+            { name: "title_formatted", weight: 2 },
             { name: "url", weight: 1 },
             { name: "description", weight: 0.5 },
           ],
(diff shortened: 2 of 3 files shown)


ChangeHow:
@pernielsentikaer Hi, for asking help if you have time.

I just met a issue that I can't run my go exe file in extension even if I run `build:go_parser` script in `build` script. The extension keep warning that can't find `tools/bookmarks-parser`.
image
image

I have no idea how to solve this issue.

ChangeHow:
> @pernielsentikaer Hi, for asking help if you have time.
> 
> 
> 
> I just met a issue that I can't run my go exe file in extension even if I run `build:go_parser` script in `build` script. The extension keep warning that can't find `tools/bookmarks-parser`.
> 
> image
> 
> image
> 
> 
> 
> I have no idea how to solve this issue.

I just checked the development doc. It seems that I need to move my exe file to the asset folder. I'll try it  later.

ksandvik:
Is there a need to use Go, why not Swift (or Objective-C) with a bridge as this is a Safari MacOSX only limit?

ChangeHow:
> Is there a need to use Go, why not Swift (or Objective-C) with a bridge as this is a Safari MacOSX only limit?

The main reason is that  I don't know other high performance language except Go😄.

The other reason is that I try to handle data  with WASM in the earlier version, and Go has good support for that.

ChangeHow:
> Is there a need to use Go, why not Swift (or Objective-C) with a bridge as this is a Safari MacOSX only limit?

I have met a performance issue that the Go parser has a wrose performance than node. Especially in the first time running
image
image
It seems that go paresr need a time to load or warm up 🥲.

pernielsentikaer:
Let's not make it slower 😊

ChangeHow:
> Let's not make it slower 😊

I do think so, I'll try some optimize methods to keep it instancy.

I will try some methods to reduce startup time. Let confirm this PR would fixed the issue from @ksandvik 😄 first.

pernielsentikaer:
Sounds like a plan; maybe @loris also has some input

ksandvik:
I still think using the same APIs as Spotlight (where this works just fine) would be a better overall approach. (So I'm using Spotlight for bookmarks for the time being).

ksandvik:
I think there are Javascript to Objective-C bridges in case this is a Javascript plugin rather than something written in for example Swift.

ChangeHow:
> I still think using the same APIs as Spotlight (where this works just fine) would be a better overall approach. (So I'm using Spotlight for bookmarks for the time being).

I think the API for searching bookmarks works fine. However, formatting in JavaScript causes that issue. We need to find other ways to format them.

I'm not sure how to use the original Apple API in the JavaScript runtime. I have tried to execute the bash directly, but it's still slow.

If you have some experience in building a bridge between the JavaScript and the Apple API, it would be great if you could update this PR or create a new one.

ChangeHow:
> I think there are Javascript to Objective-C bridges in case this is a Javascript plugin rather than something written in for example Swift.

Another question, have you ever tried this PR and made sure that it wouldn't fix your problem, so that you need a Swift/OC solution?

ksandvik:
It's fine, I'm using Spotlight instead for Safari bookmarks that work like a charm.

Issue #17556, opened 2025-03-05
[Apple Notes] Crash immediately


theboringmonkey:
### Extension

https://www.raycast.com/raycast/apple-notes

### Raycast Version

1.93.1

### Description

Hi
I have just installed the Apple Notes extension from the store, and it crashes immediately when I use any commands for Apple Notes. 

It shows this error : 

![Image](https://github.com/user-attachments/assets/d32b559f-d6cf-4695-aa5c-b208e97989e1)

Here is the issue log : 

```
Error: Minified React error #300; visit https://reactjs.org/docs/error-decoder.html?invariant=300 for the full message or use the non-minified dev environment for full errors and additional helpful warnings.

x8:index.tsx:14:54

---
11: };
12: 
13: export default function Command() {
14:   const { data, isLoading, permissionView, mutate } = useNotes();
15:   const [searchText, setSearchText] = useState<string>("");
16: 
17:   if (permissionView) {
---

_r:index.js:8:2593
    at ray-navigation-stack
No:index.js:8:2094
```

Thanks for the help.
Kind regards.

### Steps To Reproduce

Just install Apple extension and launch any commands related to this extension

theboringmonkey:
My bad, the issue is coming from disk access authorization for Raycast. 

Didn't ask this before, had to try all commands in Notes extension to get the Pop-up for disk access.

PR #17569, opened 2025-03-05
fix for app crash when user has not granted disk access

saldanaj97:
#17556 
## Description

Fix for the "Rendered fewer hooks than expected. This may be caused by an accidental early return statement." error. This was happening because the useMemo() hook was being called **after** the permission view return statement. Now, if a user has not granted disk access to Raycast, they will be shown the proper window notifying them to grant access. 

### Before
image
image

### After
image
image

## Checklist


thomaslombart (approved):

thomaslombart:
Thanks for the fix!

PR #17573, opened 2025-03-05
Add office-converter extension

XInTheDark:
## Description

Add office-converter extension: Convert office files to various formats using LibreOffice.

LibreOffice must be installed for the commands to function.

## Screencast

no screenshots available; all commands are `no-view`

## Checklist

### DIFF
--- a/extensions/office-converter/src/convert-file.tsx
+++ b/extensions/office-converter/src/convert-file.tsx
@@ -38,7 +38,7 @@ export default function ConvertFiles(props: { arguments: { format?: string; inpu
       }
     }
 
-    const message = `Converted ${successful} file(s) successfully`;
+    const message = `Converted ${successful} file(s) to ${params.format}`;
     if (failed > 0) {
       await showToast(Toast.Style.Failure, `${message}, ${failed} failed`);
     } else {


pernielsentikaer (commented):
Before I dive into the review, you might want to make them `view` commands instead and show a `<Form>` with a file/dir picker instead of the current text argument.

XInTheDark:
Hi, i feel like it's kinda redundant to render an entire form for file selection, especially since I think almost all users will be more comfortable with the "select files in Finder -> run command -> finish" workflow. how about showing the form only if there are no selected files? again seems like adding unnecessary complexity, but let me know your thoughts on this approach! :)

pernielsentikaer:
How would an average user of Raycast know how to write paths? If we can make both happy I'd like to see what you have in mind 😊

XInTheDark:
@pernielsentikaer Hi, I've rewritten the commands:
- Both commands are now `view` type, although only partially. If the input is considered complete then it will run in the background as far as possible. This hopefully keeps the workflow clean for most users.
- If the input is incomplete (e.g. no files selected, or format not specified), then we show a form to pick files and specify the format.

Please let me know your thoughts! Hopefully this should be suitable :)

XInTheDark:
@raycastbot don't close

PR #17575, opened 2025-03-05
Add gmail-accounts extension

lacherogwu:
## Description

This was a fun proof-of-concept project aimed at enhancing productivity by allowing users to open specific Gmail accounts directly from Raycast. It improves productivity by eliminating the need to manually switch accounts after opening the Gmail interface.

## Why This Extension

I love the Gmail inbox interface, but I hate that I have to open the interface and only after that switch to the desired account. This extension solves this problem by letting users directly open specific Gmail accounts from Raycast.

The reason I didn't just add this functionality to the current [Gmail Raycast extension](https://www.raycast.com/tonka3000/gmail) is because they work completely differently - the existing Gmail extension requires you to add a token, while this extension doesn't require any setup beyond having Chrome installed.

## Features

- Lists all Gmail accounts connected to your Chrome browser
- Ability to pin favorite accounts
- Organize pinned accounts by moving them up/down
- Open Gmail accounts directly in Chrome with a single click

## Security Note

This extension accesses Chrome cookies to retrieve your Google account information. It:

- Uses the [chrome-cookie-decrypt](https://github.com/lacherogwu/chrome-cookie-decrypt) package to read cookies from Chrome
- Only accesses cookies from Google domains
- Does not send your data to any external servers except Google's own services
- Stores a cache of your accounts and pinned preferences locally

The extension needs these permissions to show you which accounts you're logged into. If you have security concerns, you can review the source code to see exactly how your data is handled.

## How It Works

The extension reads your Chrome cookies to identify Google accounts you're logged into, then displays them in a convenient list within Raycast. You can then quickly access your Gmail inbox for any account with a single click.

## Question About Implementation

I have one concern that I'm not sure if I implemented correctly:

If Chrome isn't installed on the system, I show a `List` with `List.EmptyView` to inform the user. This seems a bit off, but I couldn't find a dedicated error view. Is there a better way to handle this, or is my implementation correct? Simply throwing an error doesn't look nice and gives less control over actions.

## Screencast

gmail-accounts-1

## Checklist


lacherogwu:
I'm unable to push changes to repo, help will be appreciated 🙏

CleanShot 2025-03-05 at 19 02 43@2x

lacherogwu:
@greptileai check again

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #17626, opened 2025-03-07
Update readwise-reader extension: add category filter to list documents

civ2boss:
## Description

Add the ability to filter by categories in the List Documents command

## Screencast

Article List Interface-zeaebi_Tn7

## Checklist

### DIFF
--- a/extensions/readwise-reader/CHANGELOG.md
+++ b/extensions/readwise-reader/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Readwise Reader Changelog
 
-## [Add category filtering for documents in list view] - {PR_MERGE_DATE}
+## [Add category filtering for documents in list view] - 2025-03-12
 
 - Add category filtering for documents in list view
 


nicolas-marien:
That's super cool!
I would add an action at the `List` level to show when there are no results. Otherwise the user will be "trapped" and would have to go back to the root to get the article again.
CleanShot 2025-03-07 at 12 46 53

Something like this (be could be conditioned to the `category` having a value to prevent displaying an irrelevant action when the list is empty w/o any filtering).
CleanShot 2025-03-07 at 12 41 43@2x

extensions/readwise-reader/src/list-documents.tsx
```diff
@@ -28,10 +29,11 @@ export default function ListDocumentsCommand() {
   const [documentLocation, setDocumentLocation] = useState<Document["location"]>(
     getPreferenceValues<Preference>().defaultListLocation,
   );
+  const [category, setCategory] = useState<Category | undefined>(undefined);
```
nicolas-marien, line 32: nitpick:
```suggestion
  const [category, setCategory] = useState<Category | undefined>();
```

extensions/readwise-reader/src/list-documents.tsx
```diff
@@ -63,6 +65,7 @@ export default function ListDocumentsCommand() {
         </List.Dropdown>
       }
       pagination={pagination}
+      navigationTitle={`Documents in ${documentLocation.charAt(0).toUpperCase() + documentLocation.slice(1)}${category ? ` (${category.charAt(0).toUpperCase() + category.slice(1)})` : ""}`}
```
nicolas-marien, line 68: Maybe this could be extracted to ease readability, what do you think?

civ2boss: Yep I agree, I'm going to actually just make a utility function for titlecase as I think that will be the cleanest way to extract it.

civ2boss:
Ok I added a check if the list is empty and category was set to something, show the empty list view with a reset category filter option:

No Documents Found Message

nicolas-marien (approved):

nicolas-marien:
lgtm ✌️

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #17640, opened 2025-03-07
Slack: Retry to auth if some scopes are missing

mathieudutour:
## Checklist


thomaspaulmann (approved):
Looks good and hopefully resolves the issue. I wonder if we should move that logic down into the utils. So other extensions would benefit of it as well.

mathieudutour:
`res.error?.includes("missing_scope")` is pretty specific to the Slack API - each API returns different stuff so hard to generalize

PR #17678, opened 2025-03-08
Update pocket extension

hrishabhn:
## Checklist

### DIFF
--- a/extensions/pocket/CHANGELOG.md
+++ b/extensions/pocket/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Pocket Changelog
 
-## Handle Deleted Items - {PR_MERGE_DATE}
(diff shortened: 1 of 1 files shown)


extensions/pocket/src/lib/api.ts
```diff
@@ -211,3 +222,100 @@ interface CreateBookmarkRequest {
   tags?: string[];
   url: string;
 }
+
```
vimtor, line 225: We should probably remove this comment right?

vimtor (approved):
Thanks for the contribution @hrishabhn!

extensions/pocket/src/lib/api.ts
hrishabhn: Yes, thanks for catching that!

PR #17693, opened 2025-03-09
Update password-store extension

TimShilov:
## Description

This PR adds an ability to specify a additional custom `PATH` values.

For me the extension wasn't working because I use Nix Home Manager to install packages and in my case the `pass` executable is located in `/etc/profiles/per-user/<username>/bin/pass` and extension was unable to find it.

## Checklist

### DIFF
--- a/extensions/password-store/CHANGELOG.md
+++ b/extensions/password-store/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Password Store Changelog
 
-## [Improvements] - {PR_MERGE_DATE}
+## [Improvements] - 2025-03-17
 
 - Added ability to specify custom PATH to allow custom executable locations
 - Fixed some typos in the README


extensions/password-store/src/utils/cmd.util.ts
```diff
@@ -1,5 +1,7 @@
+import { getPreferenceValues } from "@raycast/api";
 import { exec } from "child_process";
 import { promisify } from "node:util";
+import { Preferences } from "../interfaces";
```
pernielsentikaer, line 4: You don't need to manually set preference types as this is autogenerated in `raycast-env.d.ts`

Can we remove and rely on the auto generated?

extensions/password-store/src/interfaces/preferences.interface.ts
```diff
@@ -1,4 +1,5 @@
 export interface Preferences {
```
pernielsentikaer, line 1: Remove this file

pernielsentikaer:
Could you check my comments @TimShilov 😊

TimShilov:
@pernielsentikaer Thanks for review. I have removed the custom `Preferences` interface.
Also, while at it, I have also fixed a few typos in `README`... which caused `prettier` to reformat the whole file causing bigger diff than it should be. 😬

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

TimShilov:
@pernielsentikaer Sorry for jumping in after the approval, but I have a general question.
I just noticed that the `NPM Check` mentions that there are a ton of deprecated and outdated depedencies. This is just the second time I contribute to Raycast extensions, so I wonder what's the Raycast's policy on the dependency updates?
Is it okay if I just update then as part of this feature PR? Should the new PR be created just for the updates? Or maybe the dependencies are managed centralized somehow? 🤔 
I'm happy to update them, just not sure if it's okay to include that in the same PR. I usually prefer PRs to be atomic and focus on just one thing.

PR #17699, opened 2025-03-09
Add alarms extension

ocodista:
## Description
Add a new "Alarms" extension to create scheduled musical alarms.

## Screencast
A video showcasing the main functionalities of the extension: Create, View and Trigger an alarm
https://github.com/user-attachments/assets/d843a2a2-2a0f-40ee-9e37-775bc9baf7be

## Checklist

### DIFF
--- a/extensions/alarms/src/create-alarm.tsx
+++ b/extensions/alarms/src/create-alarm.tsx
@@ -2,7 +2,6 @@ import React from "react";
 import { Action, ActionPanel, Form, Icon, showToast, Toast, popToRoot, environment } from "@raycast/api";
 import { useState, useEffect } from "react";
 import { spawn } from "child_process";
-import os from "os";
 import { initializeExtension } from "./utils/initialize";
 import path from "path";
 import fs from "fs";
--- a/extensions/alarms/src/manage-alarms.tsx
+++ b/extensions/alarms/src/manage-alarms.tsx
@@ -1,5 +1,16 @@
 import React from "react";
-import { List, ActionPanel, Action, Icon, showToast, Toast, launchCommand, LaunchType, Keyboard, environment } from "@raycast/api";
+import {
+  List,
+  ActionPanel,
+  Action,
+  Icon,
+  showToast,
+  Toast,
+  launchCommand,
+  LaunchType,
+  Keyboard,
+  environment,
+} from "@raycast/api";
 import { useCallback, useState, useEffect } from "react";
 import { spawn } from "child_process";
 import fs from "fs";


ocodista:
> Hi @ocodista👋
> 
> 
> It looks pretty similar to what the [Timers extension](https://www.raycast.com/ThatNerd/timers) offers by @ThatNerdSquared - could you check it out? If something is missing it could be nice to enhance 😊

Hey @pernielsentikaer, thank you! It is similar, but different. The Timers extension doesn't allow you to add a `fixed hour` alarm, only relative timers, which doesn't really fit my usecase.

pernielsentikaer:
Sounds like something that could be a nice enhancement. What do you think, @ThatNerdSquared?

ThatNerdSquared:
Yes, this is something i've been meaning to implement in Timers since the beginning! I think it would be pretty straightforward to add in since all the infra for timers, ringtones, alerts, etc is already present in the extension as well.

@ocodista i would totally welcome this as a contribution to [Timers](https://github.com/ThatNerdSquared/timers-for-raycast), it's something I would be implementing in the near future anyway :)

If you need to make changes to how the Timers timing system works to accommodate this better, I'm also open to that, but I think it should be pretty straightforward to add (an easy way is you could calculate timers from set times, could also be expanded out from there)

ocodista:
@pernielsentikaer @ThatNerdSquared Personally, I just like the single-feature Alarms extension, I've been using it locally for the past week and it's a simple way of creating/managing my time-scheduled alarms.

Do you think it's a bad idea to have a single-feature extension on the store?

pernielsentikaer:
We do actually have a things about this subject in our [extensions guidelines](https://manual.raycast.com/extensions) in the section `Reasons to be rejected`

We'd always wants to enhance before creating a similar extension to not cause confusion for users and this extension is pretty similar to the other one mentioned where this feature could be nice

ThatNerdSquared:
normally, i would push for a feature within the scope of Timers to be integrated into the extension, but i do think alarms may be a major enough feature to warrant its own extension. my only concern is that the user experience between Timers/Alarms wouldn't be very unified, even though they seem like similar features.

@ocodista i'm not sure how you want to proceed from here, i'm totally open to this becoming a part of Timers, but also understand if you think it makes more sense to keep separate (assuming @pernielsentikaer and the raycast team is okay with it).

if you do want to keep it as a separate extension, one way of solving the un-unified issue would be starting a shared lib that backs both extensions — i can share with you some of the code that already exists for features users request, like menu bar commands, ringtone volume, etc, and i think some of your ringtone code is more robust than what i have in place right now. i'd be open to this if you think it would be a good idea :)

ocodista:
Hey @ThatNerdSquared, thank for your comment.

That would be nice!

I personally don't use timers so much and the name of the extension by itself would be a little misleading to me. I'm used to the native "Clock" alarms from IOS, which unfortunately is not scriptable, even though I'm using the ringtone files it provides.

I would like to keep this as a standalone extension, if it's ok with Raycast's team (cc @pernielsentikaer).

I liked your idea of a shared lib to empower both extensions and would love to work together to achieve it (will send you an email).

I'm also down to just integrating the scheduled alarms functionality to the Timers Extension if that's the will of Raycast, even though I do believe that there may be other users who would prefer to have alarms-only and not timers+alarms.

What do you guys think?

pernielsentikaer:
I think we can keep it as a seperate one, it seems like they are different enough for that.

- Can you look into the [CI errors](https://github.com/raycast/extensions/actions/runs/13751620486/job/38453284059?pr=17699)

- There is a lot of `sh` scripts, does it work as expected this way? Anything made with AI?

ocodista:
> I think we can keep it as a seperate one, it seems like they are different enough for that.
> 
> * Can you look into the [CI errors](https://github.com/raycast/extensions/actions/runs/13751620486/job/38453284059?pr=17699)
> * There is a lot of `sh` scripts, does it work as expected this way? Anything made with AI?


Thank you!

Yes, will fix the CI errors.

There are some .sh scripts because I need a way to:

- Add
- List
- Remove

Alarms from Raycast, but also from background jobs. As I'm using crontab, we need to remove the job once the alarm is triggered to prevent it from re-triggering on the future.

pernielsentikaer:

ocodista:
> Is this ready for another review?

Yes, now it is!

ocodista:
@pernielsentikaer Any comments/suggestions?

pernielsentikaer (changes requested):
I gave it a new spin and changed a bit, here are a few more comments:

- What happens if the user set an alarm and the machine is sleeping or shutdown at that time?
- You should use `environment.supportPath` for files related to your extension instead - you can read more about [[it here](https://developers.raycast.com/api-reference/environment#properties)](https://developers.raycast.com/api-reference/environment#properties)

ocodista:
> * What happens if the user set an alarm and the machine is sleeping or shutdown at that time?
I tested with the locked screen and it still rings. You have to unlock you mac to see the popup, but the song will be there. If the computer is sleeping though, it won't work.

ocodista:
> * You should use `environment.supportPath` for files related to your extension instead - you can read more about [[it here](https://developers.raycast.com/api-reference/environment#properties)](https://developers.raycast.com/api-reference/environment#properties)

Just updated, @pernielsentikaer could you please review this PR again when you have some time?

pernielsentikaer:
I'm not sure how it works, but somehow I got this strand that is not a part of anything. Can we do something to handle that since the extension doesn't find it?

CleanShot 2025-03-24 at 10 56 11@2x

ocodista:
> I'm not sure how it works, but somehow I got this strand that is not a part of anything. Can we do something to handle that since the extension doesn't find it?
> 
> CleanShot 2025-03-24 at 10 56 11@2x


Yeah, not sure why it didn't work for you... Any chance we could get the error log of the running terminal? I'll also add a toast notification if the read fails.

pernielsentikaer:
Do you want anything from me? I removed the timer manually since it went off daily for me 😊

PR #17705, opened 2025-03-09
Add Finder File Actions extension

pa1ar:
## Description

Finder File Actions is an extension which is allowing moving and copying files selected in folder into a folder found with Raycast. For now only Copy and Move actions are supported, but if this gets approved i consider adding a couple of more actions which some Alfred users might want to see, basically turning the extension into a clone of Universal Action from Alfred for Finder files.

## Screencast

https://github.com/user-attachments/assets/f7c8b086-2a22-4df7-ab43-774a2eb4ae1a

## Checklist


pernielsentikaer:
Hi @pa1ar👋 

We already have something like this [in the store](https://www.raycast.com/Mehdi_Hp/copymoveto), could you check that one out?

pa1ar:
Hi @pernielsentikaer thank you for the answer!

i haven't seen this extension you mentioned (and after asking around on reddit i think other Raycast users too), so now i have checked it out and i see a couple of differences in comparison to mine:

- CopyMoveTo extension requires you to add destination folders upfront
image
- you can't search your file system on the fly to move to a folder which you haven't added before
image
- the process of adding a folder into the extension requires navigating your system with system folder picker (it opens up when you click on "select folder" button on the screenshot below
image

in my extension the process is much more streamlined and keyboard/raycast oriented: 

1. select files in finder
2. type to search any folder on your computer (no folder's definition upfront required)
3. paste with cmd+return

i am using my own extension for several days now and can say it works actually pretty well.

additionally, in my extension the folders you have used as destinations are automatically added to the recent folders and are available without any actions required from the user whatsoever.

if you want, you can test both extensions in action and tell for yourself which approach is more user friendly.

pernielsentikaer:
I think it's a very personal preference. If I should use it, I would add the directories myself, actually. I also tried to search for `Desktop`, and it didn't find it since I use another system language.

We can ask @Mehdi-Hp what he thinks. Since it's pretty new, it will be possible to add some of the functionality from this one into the other one, as we won't add two so identical extensions.

Mehdi-Hp:
Hi everyone. I agree that this is very much a personal preference and depends on how you use your system. However, I don't think it makes much sense to add search functionality to CopyMoveTo, since the main idea was to have quick access to the destinations you need on a daily basis.

pernielsentikaer:
As I understand it, you'll also get it this way since it shows you recent directories (we could even add pinning directories to mimic the same functionality)

pernielsentikaer:
Would you be open for some features from this in yours, like the way this does find the directories (and pinning) so it's like yours

pa1ar:
in my version you can not only search but also browse file system. so it's not just searching it is also navigating. 

i planned to also allow writing path to folder destination and allow folder selection with path (similar to Alfred's behavior).

Mehdi-Hp:
@pernielsentikaer I know the final job is essentially the same, but maybe they're not that identical after all, and we can have them both?
This one is trying to be more low-level with navigation features, and mine is as simple and straightforward as possible.

pernielsentikaer:
The issue is that this is so alike, since if you pin directories, then you'll have the exact same extension 🤔

ultgames:
What harm is there in letting users choose for themselves even if there is some overlap in extensions.

pa1ar:
i see that the critical tests are passing, meaning the extension seem to be adhering to the guidelines. but please let me know if there is anything i can do to make it better - this is my first raycast extension and despite reading the docs i might have been overlooking something.

I am using the extension locally for couple of weeks by now and it seems to work fine.

tldr: i guess it's ready for the review. ;D

pernielsentikaer:
- Some actions doesn't have icon, could you make sure that all of them is having?

finder-file-actions 2025-03-28 at 09 49 15

Did you consider the pin feature?

pa1ar:
first off, i've added action icons. 

regarding the pins - i have thought about it and someone on Reddit also suggested it, yet i am not sure if that won't clutter the list. i already have recent folders, so when you are in navigation mode, you already have a lot going on. in Alfred, which serves as the reference for my extension, there are no pinned folder afaik, yet everyone seem to be pretty happy with the "Move to..." functionality.

pernielsentikaer:
I'm not sure we can do anything about it, most of the times I have macOS in my native language to catch AppleScript errors, can we improve something here?

finder-file-actions 2025-03-31 at 11 19 43

If I search for `Desktop` it's not finding my desktop folder at all

pernielsentikaer:
Do you think it could work, just tested a fresh build and it's still not showing up

pa1ar:
i though it's fixed but apparently it's not. just switched to german as an OS language (i do not recommend) to be able to test it properly. sorry about the several pushes.

pa1ar:
@pernielsentikaer can you please test it? i've added bunch of stuff:

- the folder 
- pinning folders
- caching pinned folder for instant loading 
- async file processing
- further refactoring, tests coverage, performance improvements etc.

the localized folders should also be searchable in both localized and english version see screenshots ↓

SCR-20250401-pug
SCR-20250401-ptq

pernielsentikaer:
Awesome, let's go with that 😊

pernielsentikaer (approved):

pa1ar:
amazing, thank you! hope raycast users will like it ^___^ 

thank you for your help and thorough review!

PR #17766, opened 2025-03-11
Turn Apple Mail into an AI Extension

### LINKED ISSUE #17615: [Apple Mail] ...
### Extension

https://www.raycast.com/yug2005/mail

### Description

You should add the ability to ask apple mail @mail. 

People That would benefit from this feature would be anyone anywhere even if you are a developer or work for your local UK Council.

### Anything else?

_No response_

danteissaias:
## Description

Closes #17615

This PR turns the Apple Mail extension into an AI extension.

Supported tools:
- Search emails
- List addresses
- Send email

Also added support for setting the default mail account, which is now used as the default in `Compose New Message` and by the AI when sending an email.

### How does email searching work?

I have opted for a combination of searching the local .emlx files on the filesystem using ripgrep (which this extension downloads a copy of) and using the SQLite database for filtering by date and sender. This approach is necessary because Apple does not store the complete content of all emails within the SQLite database.

## Screencast

Screenshot 2025-03-11 at 22 07 57

## Checklist

### DIFF
--- a/extensions/mail/src/utils/validation.ts
+++ b/extensions/mail/src/utils/validation.ts
@@ -1,6 +1,7 @@
 import { formatFileSize, getSizeSync, maximumFileSize } from "./finder";
 
-const emailRegex = /[^\\.\\s@:](?:[^\\s@:]*[^\\s@:\\.])?@[^\\.\\s@]+(?:\\.[^\\.\\s@]+)*/;
+// https://colinhacks.com/essays/reasonable-email-regex
+const emailRegex = /^([A-Z0-9_+-]+\.?)*[A-Z0-9_+-]@([A-Z0-9][A-Z0-9-]*\.)+[A-Z]{2,}$/i;
 
 const required = (value: unknown) => {
   let isValid = true;


thomaspaulmann (changes requested):
Great to see this extension coming to the Store! I had a few issues while testing that would be great to fix:

I got "Error: No accounts found" when I wanted to send an email. Can we make this somehow work? Anyway I can help debug it?
I got sometimes that the context window was too long. Mainly when searching for emails. Anyway we could improve this? We could create another tool for reading the content of an email and only return the subject and a summary in the "search-emails" tool, wdyt?
Otherwise I like how you handle the download of ripgrep 👌

danteissaias:
Thanks for your review!

> I got "Error: No accounts found" when I wanted to send an email. Can we make this somehow work? Anyway I can help debug it?

Looks like the extension loads your accounts lazily. So, if the send-email tool was on and you hadn’t used the extension before, the cached accounts wouldn’t show up. Should be a quick fix.

> I got sometimes that the context window was too long. Mainly when searching for emails. Anyway we could improve this? We could create another tool for reading the content of an email and only return the subject and a summary in the "search-emails" tool, wdyt?

What idea did you have in mind for generating the summaries? Another thought is to set a strict upper limit on the total number of emails we can return. If the LLM doesn't find what it's looking for in the first search, it can try more targeted searches.

danteissaias:
Pushed a few fixes:
- Your email accounts are now loaded properly when you use the send email tool.
- I've added a new tool to read the full email content, and the search emails tool instead just return a summary of the email (the first 100 characters).
- I've also made the tool descriptions better so the AI tries the search emails tool several times before giving up, and I've added evals for this too.

thomaspaulmann (changes requested):
This is great and the split into getting emails and then reading specific ones works fine. I ran into one more issue, where I got "Invalid email for recipient" when I try to send an email, e.g. for something like "send email to joe@example.com saying hi". I see a toast popping up and the email isn't sent. Is the validation off?

danteissaias:
I've switched to a different email regex (zod's https://colinhacks.com/essays/reasonable-email-regex). Let me know if you continue to run into that issue.

thomaspaulmann (approved):
Yeah, that seems to do the trick! I'll approve this request as it seems to work really nicely now.

I had one more issue that the account returns multiple email addresses formatted with " | " as separator. It doesn't seem to cause any issue but might be something to look into. This might be related to my personal setup where I have multiple email addresses setup in my Apple Mail.

PR #17774, opened 2025-03-12
Add csv-to-json-array extension

Gialicus:
## Checklist

### DIFF
--- a/extensions/csv-to-json-array/package.json
+++ b/extensions/csv-to-json-array/package.json
@@ -13,7 +13,6 @@
     {
       "name": "csv2json",
       "title": "CSV to JSON Array",
-      "subtitle": "Select File",
       "description": "Convert selected csv file into json file",
       "mode": "no-view"
     }


andreaselia:
Hey @Gialicus 👋 

This looks similar to existing extensions such as [ToolBox](https://www.raycast.com/Kang/toolbox) and [Transform](https://www.raycast.com/Idered/transform).

Would it be worth contributing to one of those with this command functionality instead?

Gialicus:
Hey there! 👋

Thanks for the feedback! You're absolutely right about the similarity to ToolBox and Transform extensions. I hadn't considered that approach.

I'd be more than happy to contribute this CSV-to-JSON functionality to one of those existing extensions instead. It makes perfect sense to enhance what's already available rather than creating something redundant.

Could you point me to which one might be the better fit for this functionality?

andreaselia:
@Gialicus I'd probably recommend the [Transform](https://www.raycast.com/Idered/transform) extension in this case.

Gialicus:
@andreaselia 
Thanks for the suggestion! I took a look at the "Transform" extension, but it uses a form-based approach that has some limitations when it comes to the size of the input and output text. In my case, since the extension is generating a .json file as output, I need a more flexible solution that can handle larger data and write the result to a file directly. This is why I'm focusing on using streams for both parsing the CSV and outputting the JSON file, which offers more control over memory usage and large data handling.

Given the difference in how "Transform" works (based on forms for input/output), do you think it would be worth adding a non-form-based functionality to the extension to handle scenarios like this, or would it be better to keep this as a separate command? Since the basic input/output structure differs, I'm wondering if merging the approaches would fit well or if it might introduce unnecessary complexity.

andreaselia:
@Gialicus good points, I think in this case it will be fine to have this as its own extension.

andreaselia:
@Gialicus nice work on the extension 🙌 

Could you add at least one [screenshot](https://developers.raycast.com/basics/prepare-an-extension-for-store#screenshots) of the extension and add them to a folder called `metadata` please?

extensions/csv-to-json-array/package.json
```diff
@@ -0,0 +1,41 @@
...
+    "Data"
+  ],
+  "icon": "extension-icon.png",
+  "author": "giali_coletta",
+  "license": "MIT",
+  "commands": [
+    {
+      "name": "csv2json",
+      "title": "CSV to JSON Array",
+      "subtitle": "Select File",
```
andreaselia, line 16: You can remove the subtitle here.

andreaselia (changes requested):

andreaselia:
@Gialicus any issues with generating the screenshots? Happy to provide some support if needed.

Gialicus:
I added the metadata folder with the screenshots in the last commit. Let me know if it's okay.

andreaselia:
@Gialicus you added screenshots which are ok for things like the README, but there's a standardised way of generating the metadata screenshots solely for the Raycast Store.

You can read more here: https://developers.raycast.com/basics/prepare-an-extension-for-store#adding-screenshots

PR #17791, opened 2025-03-12
Add awork extension

Malte-Hypercode:
## Description

This is a Raycast extension with commands for the work management platform [awork](https://www.awork.com/).

## Screencast

https://github.com/user-attachments/assets/f43dbf7d-662a-4d5b-9bc6-5d247bad9f7f

## Checklist


extensions/awork/README.md
```diff
@@ -0,0 +1,64 @@
...
+
+## Installation & Authorization
+
+To use this extension you need to add an API client to awork. You can add this under "Settings / Integrations / API clients" and you may need admin rights. However, the users of the extension will only be able to see/access the projects that they are individually allowed to see in your workspace.
+
+Please enter something like:
+
+- Name: `awork Extension for Raycast`
+- Client ID: `raycast-awork-extension`
+- Redirect URI: `https://raycast.com/redirect?packageName=awork` (hit the "+")
```
xmok, line 14:
```suggestion
- Client ID: `raycast-awork-extension`
- Redirect URI: `https://raycast.com/redirect?packageName=Extension` (hit the "+")
```

extensions/awork/README.md
```diff
@@ -0,0 +1,64 @@
+# awork for Raycast
+
+This is a Raycast extension with commands for the work management platform [awork](https://www.awork.com/).
+
+
+## Installation & Authorization
+
+To use this extension you need to add an API client to awork. You can add this under "Settings / Integrations / API clients" and you may need admin rights. However, the users of the extension will only be able to see/access the projects that they are individually allowed to see in your workspace.
```
xmok, line 8: I couldn't find this as I was looking in "Account Settings" so we can mention it's Workspace over here.

```suggestion
To use this extension you need to add an API client to awork. You can add this under "(Wokspace) Settings / Integrations / API clients" and you may need admin rights. However, the users of the extension will only be able to see/access the projects that they are individually allowed to see in your workspace.
```

extensions/awork/src/composables/WebClient.ts
```diff
@@ -0,0 +1,168 @@
+import { Alert, confirmAlert, getPreferenceValues, LocalStorage, OAuth, PreferenceValues } from "@raycast/api";
```
xmok, line 1:
```suggestion
import { Alert, confirmAlert, getPreferenceValues, LocalStorage, OAuth } from "@raycast/api";
```

extensions/awork/src/composables/WebClient.ts
```diff
@@ -0,0 +1,168 @@
...
+interface User {
+  id: string;
+  workspace: workspace;
+}
+
+export const baseURI = "https://api.awork.com/api/v1";
+export let authorizationInProgress = false;
+export let revalidating = false;
+
+const preferences = getPreferenceValues<PreferenceValues>();
```
xmok, line 20: `Preferences` is auto-generated and is type-safe.
```suggestion
const preferences = getPreferenceValues<Preferences>();
```

extensions/awork/package.json
```diff
@@ -0,0 +1,90 @@
...
+      "required": true,
+      "title": "Client ID"
+    },
+    {
+      "name": "clientSecret",
+      "description": "Enter the Client Secret of your workspace",
+      "type": "password",
+      "required": true,
+      "title": "Client Secret"
+    }
```
xmok, line 56: Let's add `placeholder` to both of these.

extensions/awork/src/projects.tsx
```diff
@@ -0,0 +1,95 @@
...
+  );
+};
+
+const ProjectItem = (props: { project: project }) => {
+  return (
+    <List.Item
+      title={props.project.name}
+      subtitle={props.project.company?.name}
+      actions={<Actions projectID={props.project.id} isBillable={props.project.isBillableByDefault} />}
+    />
```
xmok, line 67: An `icon` would look nice here.
Maybe `Icon.Folder`?

extensions/awork/src/projects.tsx
```diff
@@ -0,0 +1,95 @@
...
+import { getProjects, project } from "./composables/FetchData";
+import { authorizationInProgress } from "./composables/WebClient";
+
+const Actions = (props: { projectID: string; isBillable: boolean }) => {
+  const { data: BaseUrl } = useCachedPromise(() => LocalStorage.getItem<string>("URL"));
+
+  return (
+    <ActionPanel>
+      <Action.OpenInBrowser url={`${BaseUrl}/projects/${props.projectID}`} />
+      <Action.CopyToClipboard content={`${BaseUrl}/projects/${props.projectID}`} />
```
xmok, line 13: I though I was copying title but it was URL so let's clarify.
```suggestion
      <Action.CopyToClipboard title="Copy URL to Clipboard" content={`${BaseUrl}/projects/${props.projectID}`} />
```

extensions/awork/src/projects.tsx
```diff
@@ -0,0 +1,95 @@
...
+    pagination,
+    revalidate: updateSearch,
+  } = useCachedPromise(getProjects, [searchText, 100], {
+    onData: (data) => {
+      if ((!data || (data.length === 0 && !searchText)) && !authorizationInProgress) {
+        setTimeout(() => {
+          updateSearch();
+        }, 500);
+      }
+    },
```
xmok, line 85: Instead of `setTimeout`, could we use `execute`?
e.g.
```
execute: !authorizationInProgress && !!searchText
```

extensions/awork/src/projects.tsx
```diff
@@ -0,0 +1,95 @@
...
+  });
+
+  return (
+    <List isLoading={isLoading} pagination={pagination} throttle onSearchTextChange={setSearchText}>
+      {projects &&
+        Array.isArray(projects) &&
+        projects.map((project) => <ProjectItem key={project.id} project={project} />)}
```
xmok, line 92: Doesn't seem like your data is ever not an Array - could we remove the Array check if it makes sense?
We can do `initialData: []` in the above ☝️ `useCachedPromise` for a guaranteed result.

extensions/awork/src/tasks.tsx
```diff
@@ -0,0 +1,117 @@
...
+        setTimeout(() => {
+          console.log("Reloading projects");
+          updateProjects();
+        }, 500);
+      }
+      if (props.launchContext?.projectId) {
+        setProjectId(props.launchContext.projectId);
+      }
+    },
+  });
```
xmok, line 90: See my suggestions in "projects.tsx" - we should avoid setTimeout if possible opting for `Arguments` and `execute` if viable so our controllers don't mess up.

extensions/awork/src/tasks.tsx
```diff
@@ -0,0 +1,117 @@
...
+          <List.Dropdown.Item title="All" value="" key="all" />
+          {projects &&
+            Array.isArray(projects) &&
+            projects.map((project) => <List.Dropdown.Item title={project.name} value={project.id} key={project.id} />)}
+        </List.Dropdown>
+      }
+    >
+      {tasks &&
+        Array.isArray(tasks) &&
+        tasks.filter((value) => value.projectId === projectId).map((task) => <TaskItem key={task.id} task={task} />)}
```
xmok, line 114: When I choose "All" in Dropdown I don't see any tasks - let's add a check so it doesn't filter when there is no projectId e.g.

```suggestion
      {tasks &&
        Array.isArray(tasks) &&
        tasks.filter((value) => !projectId || (value.projectId === projectId)).map((task) => <TaskItem key={task.id} task={task} />)}
```

extensions/awork/src/logTime.tsx
```diff
@@ -0,0 +1,262 @@
...
+      }
+    },
+    initialValues: {
+      date: new Date(),
+      isBillable: true,
+      ...props.draftValues,
+    },
+    validation: {
+      projectId: (value) => {
+        setValidationError("projectId", undefined);
```
xmok, line 163: This seems redundant?
```suggestion
```

xmok (changes requested):
Hi @Malte-Hypercode:

Cool extension! Hadn't heard of awork but I signed up for a trial to review your extension 😄.

Have left a bunch of comments - let's try to remove the `setTimeout` - I haven't marked them in `logTime` but they are there as well. Since we already have access to `execute` and arguments, it makes sense to use them rather than rely on siteTimeout.

If there is a relevant use case, please let me know and we can disregard it.

extensions/awork/src/projects.tsx
Malte-Hypercode: Thank you for your feedback. 
The timeout is used to wait for the re-validation of an access token before calling the updateSearch and trying to re-validate again, this is done to avoid calling the function too often in a short intervall.
The execute approach wouldn't work in this context as it would block the initial call if there is no valid and we would just have to call the updateSearch again.

xmok: Hmm 🤔 .
We could use the token as an `Argument` so it will trigger a `revalidation`? There are plenty of extensions using OAuth in the Store and `setTimeout` hasn't been needed for it to work. This would also account for the initial call not running.

pernielsentikaer:
Could you check again @xmok

xmok (commented):
My token expired so until it refreshed, I got 3 or 4 back-to-back Failure Toasts before the token was refreshed.
@Malte-Hypercode this is the type of situation I'm trying to avoid in regards to `setTimeout`. The refresh should not throw so many errors. Is removing setTimeout really not an option?

If so, please lmk again and we can get this merged 💯.

xmok:
@Malte-Hypercode my free trial has expired so I can't test the extension.
I'm pretty sure the underlying issues are still there but if the extension is working for you, LGTM. If other users experience issues they can be addressed in future PRs.

stefangrund:
> @Malte-Hypercode my free trial has expired so I can't test the extension. I'm pretty sure the underlying issues are still there but if the extension is working for you, LGTM. If other users experience issues they can be addressed in future PRs.

Hi @xmok, thanks for your feedback, @Malte-Hypercode's colleague Stefan here. We've been using the extension successfully in our internal team of 5 since December. We've also already contacted the team at awork, who have also successfully tested it internally. So I would follow your suggestion to optimise it later if necessary. :)

xmok (approved):
@stefangrund thank you for being so patient 💯.
Yup then LGTM.

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

cmantsch:
@Malte-Hypercode Are there any plans to include a command for quickly creating a new task? That's something I really miss in awork to have systemwide creation of tasks, would be awesome if this could be solved in Raycast!

stefangrund:
@cmantsch Hi Christoph, thanks for your feature request! @Malte-Hypercode and I talked with the team at awork about this, who also thought that was a great idea. So we're currently working on a "Create Task" feature (which also includes Raycast AI support). It should be ready in the next few days.

PR #17825, opened 2025-03-13
Notion ai token fix

thestumonkey:
## Description

When in dev mode, the "ask AI" part was not able to use the oauth token, even though the extension itself was fine.
This fixes the issue, although this was with heavy use of AI so needs review.

NOTE: AI seems to require the internal token and doesn't currently work with the standard OAuth.

To repro issue:
- Checkout extension
- Call  npm install && npm run dev
- Use non-AI command to prompt oauth notion login, and/or create a new managed app and use token
- Try to ask notionAI to get databases
- Observe it cannot connect

Changes:
- Move token initialization to top of file to ensure AI tools have access
- Add 10s timeout to prevent 504 Gateway Timeout errors
- Maintain consistent client initialization across all tools

## Checklist

### DIFF
--- a/extensions/notion/src/utils/notion/oauth.ts
+++ b/extensions/notion/src/utils/notion/oauth.ts
@@ -7,9 +7,9 @@ const { notion_token } = getPreferenceValues();
 // Initialize the client with the personal access token if available
 let notion: Client | null = null;
 if (notion_token) {
-  notion = new Client({ 
+  notion = new Client({
     auth: notion_token,
-    timeoutMs: 10000 // 10 second timeout
+    timeoutMs: 10000, // 10 second timeout
   });
 }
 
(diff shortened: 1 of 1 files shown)


extensions/notion/src/utils/notion/oauth.ts
```diff
@@ -23,14 +30,34 @@ export const notionService = new OAuthService({
...
+// Export a function to get the token for AI tools
+export function getNotionToken() {
+  return notion_token;
+}
+
+// Wrapper function to handle token and client initialization for all tools
+export async function withNotionClient<T>(fn: (client: Client) => Promise<T>): Promise<T> {
+  const client = getNotionClient();
+  return fn(client);
+}
```
thomaslombart, line 63: I'm unsure of what the purposes of these functions are, as they aren't used anywhere in the code.

thomaslombart (commented):
Could you send a recording? I wasn't able to reproduce the issue you're facing.

extensions/notion/src/utils/notion/oauth.ts
thestumonkey: That's weird.  The issue that caused me to make this patch no longer occurs.  Will abandon this PR unless I can repro it again.

PR #17853, opened 2025-03-14
Add pinia-docs extension

jeffnawroth:
## Checklist


extensions/pinia-docs/src/open-pinia-docs.ts
LitoMore, line 1: Ditto: https://github.com/raycast/extensions/pull/17855#pullrequestreview-2686097793.

jeffnawroth: Done!

LitoMore: It looks like the file is still there. But this should be fine. As far as I remember, it's a bug in the ray publish. I can help to remove it later.

LitoMore (approved):
@pernielsentikaer LGTM.

@jeffnawroth BTW, I think there is still room for improvement since those URLs are hardcoded in the source and the documentation sections are subject to change.

PR #17871, opened 2025-03-15
Add note to ESLint Customization section

LitoMore:
Add a note for the developer to avoid adding too many personal ESLint preferences to an extension.

And the Prettier reformatted the markdown file.


docs/information/developer-tools/eslint.md
mormubis, line 1: Should we update the config to the new ESLint flat config?

pernielsentikaer:
@LitoMore can you check the comments from @mormubis

LitoMore:
Updated to ESLint 9 flag config.

pernielsentikaer:
@mathieudutour could you take this since it's related to config files

mathieudutour (approved):

Issue #17904, opened 2025-03-16
[GitHub Profile] Heatmap data not correct


ViGeng:
### Extension

https://www.raycast.com/ViGeng/github-profile

### Raycast Version

1.93.2

### Description

Contribution data is not correct.

### Steps To Reproduce

as title

ridemountainpig:
Oh, I see you are the author of this extension. Are you documenting this bug? 🤔

By the way, I think the token needs to be added so the extension can display the actual contribution data instead of mock data. This information can be included in the README or show in the extension to help users understand it easily.

ViGeng:
> Oh, I see you are the author of this extension. Are you documenting this bug? 🤔
> 
> By the way, I think the token needs to be added so the extension can display the actual contribution data instead of mock data. This information can be included in the README or show in the extension to help users understand it easily.

Haha, yes, I just found this issue but did not have a clue for this.
And fortunately you told me the answer. You are very helpful!

Yes, you are right. Maybe we could change the token to a `required` field!

ridemountainpig:
> Maybe I need to remove the mock, which was not originally intended.

I think we can just add a small word to inform the user, instead of removing it.

ViGeng:
> > Maybe I need to remove the mock, which was not originally intended.
> 
> I think we can just add a small word to inform the user, instead of removing it.

Using a REST can also work with a rate limit though. REST API can be planned as a future feature. 
Okay, then, let's inform the user of mock data if the token is not provided. :)

PR #17906, opened 2025-03-16
update extension GitHub Profile

ViGeng:
## Description

fix issue #17904 

~TLDR: we don't force users to provide a token but show the alerts when mock data is returned, as a fix.
Since REST API can be used to fetch even without a token, which has not been implemented yet.~

Now we change GitHub token to required field, as a safe choice.

## Screencast

image

## Checklist


ViGeng:
Hi @ridemountainpig , what do you think? :)

extensions/github-profile/src/api/github.ts
```diff
@@ -104,6 +104,7 @@ export const fetchContributionData = async (username: string): Promise<Contribut
 
   if (!preferences.githubToken) {
     // If no token provided, return mock data
+    showFailureToast("⚠️ Showing MOCK data without token");
```
ridemountainpig, line 107: I think this will be better.

```suggestion
    showToast({
      title: "⚠️ Showing MOCK data without token",
      style: Toast.Style.Failure,
      primaryAction: {
        title: "Add GitHub Token",
        onAction: () => {
          openExtensionPreferences();
        },
      },
    })
```

ViGeng: In this case, if we want to let users fill in their tokens, we can just set the token field as required in package.json file.

ridemountainpig: > In this case, if we want to let users fill in their tokens, we can just set the token field as required in package.json file.

I think it will be great.

ViGeng:
As in the description, now let's require users to input a token as a safe choice.

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

Issue #16033, opened 2024-12-25
[Change Case] Preserve the punctuations such as period when changing to upper case


Garfield-Finch:
### Extension

https://www.raycast.com/erics118/change-case

### Description

The current implementation of changing text to uppercase removes symbols such as `.` and `'`. Would it be a better approach to retain the symbols as they are and only adjust the case of the textual content?

### Who will benefit from this feature?

All users of the Change Case Extention who change cases for sentences or paragraphs.

PWrzesinski:
Personally the current behavior is what I need and what I would expect, however I understand you have a different use case. It's true that the current way is more suited to software development than writing plain sentences.

I'm thinking perhaps there could be an "alternative" behavior that would work like you describe? This could be set in the extension settings or even have a different shortcut, like alt+enter. Would that cover your use case? It might require a bit of an effort, I guess all the individual transformations would have to be updated separately, because each one will probably need slightly different behavior.

@erics118 @pernielsentikaer what do you think?

erics118:
This can easily be done.

This is the current behavior for upper case:
1. convert to "no case", ie `snake_case` -> `snake case`, or `KEBAB-UPPER-CASE`-> `kebab upper case`
2. uppercase everything (with `.toUpperCase()`)

I can change it so that the new behavior simply just skips step 1, and uppercases everything directly. I can also add this change for lower case too. But, It would be more confusing in certain scenarios for sentence/title case.

christianmagill:
@erics118, Title case is where I run into the most issues with this.

PR #17931, opened 2025-03-17
Update change-case extension

gianpispi:
## Description

This PR fixes the lowercase behavior of the extension. Previously, if I wanted to lowercase a uuid for example it would lose all the dashes and encounter weird spaces, which are important for a uuid. Screenshots comparing the behavior in the next session.

Also if everyone agrees on this I can also do it for the uppercased one, because it has the same weird behavior for uuids as the lowercased had

## Changes

For context the UUID I had in my clipboard was: `4AC89DA5-3897-455A-B218-86BA385646B6`

| Before | After |
| --- | --- |
| Screenshot 2025-03-17 at 15 22 14 | Screenshot 2025-03-17 at 15 22 01 |

## Checklist

### DIFF
--- a/extensions/change-case/CHANGELOG.md
+++ b/extensions/change-case/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Change Case Changelog
 
+## [Fix lowercase implementation] - 2025-03-17
+- Fixes the implementation of the lowercased change case
+
 ## [Focus last used case] - 2024-11-24
 - Focus the last used case when opening the extension
 


erics118:
Could this instead be added as a setting? Some people use the current behavior for other use cases

gianpispi:
@erics118 sure. or maybe as a new case in the list?

I dont really understand what the behavior is to be honest.
can you please help me understand?

erics118:
Sorry, disregard that comment. In this case, I believe the proper way to do this is to uncheck the `Preserve Casing` preference and use the `kebab case`. Read the README for more details

PWrzesinski:
> Could this instead be added as a setting? Some people use the current behavior for other use cases

+1 to this, I definitely use and prefer the current behavior. This is actually a common pattern I see with Raycast's extensions - someone needs a slightly different behavior to already existing one and modifies a behavior for everyone. Maybe it should be a policy that outside bug fixes every change should be additive?

In this particular case I think it should either be a setting, or a new case, like "to lower case keeping special characters".

gianpispi:
it's fine for me to add it as a preference, but can somebody help me understand why you want this behavior? I don't understand the logic there. It doesn't only remove special characters, it also adds spaces where there are none...

PWrzesinski:
> it's fine for me to add it as a preference, but can somebody help me understand why you want this behavior? I don't understand the logic there. It doesn't only remove special characters, it also adds spaces where there are none...

kebab-case separates words by hyphens, lowercase separates words by spaces - when you convert one to the other you should use the separator appropriate to the new format.

One example I have used this for is for rest endpoint names like user-ticket-extension, where I converted it to normal sentence ("user ticket extension endpoint") for documentation. This is admitedly not a great example, I know I have used it in other places, but that's the first thing that came to my mind.

And I really want to add, we absolutely appreciate your addition, just please make it optional 🙂

erics118:
> I don't understand the logic there.

This behavior is mostly explained in the README.

For some more clarification:

Suppose the `Preserve Casing` preference is enabled (default behavior). It allows for transformations like so (to lowercase):

1. `HelloWorld` -> `hello world` splits by the lowercase letter followed by uppercase
2. `Thing2Object` -> `thing2 object` also splits by number followed by uppercase
3. `HELLOWorld` -> `hello world` splits in a useful manner, keeping the `W` with the rest of `orld`
4. `hello-world` -> `hello world` splits by dash (all non-alphanumeric chars I believe)
5. `HELLO-WORLD` -> `hello world` lowercases everything.

Now, to explain the behavior for the UUID, "word" by "word":

`4AC89DA5-3897-455A-B218-86BA385646B6` -> `4 ac89 da5 3897 455 a b218 86 ba385646 b6`
- `4` it sees the uppercase `A`, and the rule showcased in example `1.`, it creates a new word.
- `ac89` it splits because it sees the `-`. See example `4.`
- `3897` see example 4
- `455` see 1
- `a` see 4
- `b218` see 4
- `86` see 1
- `ba385646` see 1
- `b6` end of string

gianpispi:
Maybe we can implement a new setting named "preserve special characters"?

erics118:
This PR also kinda related to #16033, so I guess we can keep change the behavior.

Using "Preserve special characters" is a misnomer because this would actually change the entire behavior of the transformation, not just preserve special characters.

IMO, it can be named "Use Simple Upper/Lower Case Transformations", and have it disabled by default.

If you do this, make sure to also add to the README that clearly explains what it does and how it differs from the default behavior

extensions/change-case/CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Change Case Changelog
 
+## [Fix lowercase implementation] - 2025-03-17
```
pernielsentikaer, line 3:
```suggestion
## [Fix lowercase implementation] - {PR_MERGE_DATE}
```

pernielsentikaer (changes requested):
Sounds like a plan

PR #17933, opened 2025-03-17
Update copymoveto extension

snok3r-krk:
## Checklist

### DIFF
--- a/extensions/copymoveto/src/destination-form.tsx
+++ b/extensions/copymoveto/src/destination-form.tsx
@@ -1,9 +1,22 @@
-import { Action, ActionPanel, Form, LocalStorage, Toast, popToRoot, showToast } from "@raycast/api";
+import {
+  Action,
+  ActionPanel,
+  Form,
+  LocalStorage,
+  Toast,
+  popToRoot,
+  showToast,
+  getPreferenceValues,
+} from "@raycast/api";
 
 import { FormValidation, useForm } from "@raycast/utils";
 import { useEffect, useState } from "react";
 import { type Destination, destinationRepo } from "./repo/destination";
 import { checkDirExists } from "./utils/filesystem";
+
+interface Preferences {
+  showHiddenFolders: boolean;
+}
 
 interface EditDestinationProps {
   destination?: Destination;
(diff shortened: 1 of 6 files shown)


mehdi-hp-spdm:
This seems unnecessary to me as you can just hit `cmd+shift+.` to show hidden files

Mehdi-Hp:
This seems unnecessary to me as you can just hit cmd+shift+. to show hidden files

extensions/copymoveto/src/destination-form.tsx
```diff
@@ -100,6 +114,7 @@ export default function DestinationForm(props: EditDestinationProps) {
         allowMultipleSelection={false}
         canChooseDirectories={true}
         canChooseFiles={false}
+        showHiddenFiles={preferences.showHiddenFolders}
```
ridemountainpig, line 117: Here, use the `showHiddenFolders` preference. However, this preference is not available in package.json.

pernielsentikaer:
Could be a nice preference to set - I think it's pretty helpful actually @Mehdi-Hp

extensions/copymoveto/src/destination-form.tsx
```diff
@@ -1,10 +1,23 @@
...
 
 import { FormValidation, useForm } from "@raycast/utils";
 import { useEffect, useState } from "react";
 import { type Destination, destinationRepo } from "./repo/destination";
 import { checkDirExists } from "./utils/filesystem";
 
+interface Preferences {
+  showHiddenFolders: boolean;
+}
+
```
pernielsentikaer, line 20:
```suggestion
```

You don't need to manually set preference types as this is autogenerated in `raycast-env.d.ts`

extensions/copymoveto/src/destination-form.tsx
pernielsentikaer: Yes, the most important part is msising

Mehdi-Hp:
> Could be a nice preference to set - I think it's pretty helpful actually @Mehdi-Hp

It is helpful, but that's more like a Finder thing. Otherwise, every extension that is using `Form.FilePicker` should implement this setting.  
If you all really think this is useful, I personally prefer a checkbox in the same form, not in the extension's preferences. Or we can even always show the hidden files. But I feel like every Raycast user already knows how to toggle the hidden files' visibility.

pernielsentikaer:
The new preference is also being set to required: false, so everything will be working for everyone as it did before this update

PR #17969, opened 2025-03-19
Fix default `text` property value in `List.Item.Detail.Metadata.Label`

xilopaint:
## Greptile Summary

This PR modifies the default color value for metadata labels in list item details, changing from `Color.SecondaryText` to `Color.PrimaryText` in the API documentation.

- The change in `api-reference/user-interface/list.md` affects the default appearance of `List.Item.Detail.Metadata.Label` text color
- This could impact visual consistency in existing extensions that rely on the default secondary text color
- Consider keeping the current default of `Color.SecondaryText` since it's an established pattern used across many extensions
- The change could affect readability and hierarchy in metadata sections where primary/secondary text distinction is important


pernielsentikaer:
Apparently, we need to change the API docs, not the docs here. Sorry about that 🫠

PR #18026, opened 2025-03-21
Add todoisttoggl extension

hayuse:
## Description
This extension integrates the time tracking feature of [Toggl](https://toggl.com/) with the task management tool [Todoist](https://app.todoist.com).  
It is designed for individual task management, so some features are limited.

``Tracking Task``
- Create/Edit/Complete Todoist tasks (Project, Title, Description, Labels, Priority)
- Start/Stop time tracking for tasks (Tracking is done via Toggl.)
- Display total time spent on a task (Only for tasks managed through this extension)

## Screencast
You can link Todoist Projects with Toggl Projects.
To do this, simply add @{Toggl Project ID} at the end of the Todoist Project name.
Even without linking them, time tracking with Toggl still works, but it will automatically start under the Default workspace.

**Home - default**
スクリーンショット 2025-03-21 10 43 44
- Select "Create new task" and ``enter`` Transition to Create task form.
- Select task and ``enter`` Start tracking timer in toggl.
- Select task and ``cmd + enter`` Completed task in Todoist.
- Select task and ``cmd + e`` Transition to Update task form.
- Select task and ``cmd + ->`` Calculate time in task.

**Create / Edit Task**
スクリーンショット 2025-03-21 10 47 27
- Labels (Tags) show the Todoist Labels and Toggl Tags associated with the linked Project (from Todoist and Toggl).
You can’t add new Labels (Tags).
- ``cmd + enter`` Create task only.
- ``cmd + shift + enter`` Create task and start tracking time.

**Tracking Task**
スクリーンショット 2025-03-21 10 47 42
- Select Running Timer and ``enter`` Stop tracking timer in toggl.

## Checklist

### DIFF
--- a/extensions/todoisttoggl/package-lock.json
+++ b/extensions/todoisttoggl/package-lock.json
@@ -8,11 +8,11 @@
       "license": "MIT",
       "dependencies": {
         "@doist/todoist-api-typescript": "^4.0.0-alpha.3",
-        "@electron/get": "^3.1.0",
+        "@electron/get": "^4.0.0",
         "@raycast/api": "^1.94.0",
         "@raycast/utils": "^1.19.1",
         "dayjs": "^1.11.13",
-        "electron": "^34.2.0",
+        "electron": "^35.1.0",
         "node-fetch": "^3.3.2",
         "rimraf": "^6.0.1",
         "uuid": "^11.1.0"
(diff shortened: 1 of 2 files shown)


pernielsentikaer:
What do you think, @thomaslombart? I'm not sure the right approach is to "merge" two extensions into one; I don't think this is the best approach

thomaslombart:
@pernielsentikaer That indeed doesn't make sense to add the Toggl features in the Todoist extension, as this would lead to a mess further down the road if we do that for every third-party tool.

hayuse:
@pernielsentikaer @thomaslombart 
Thank you for the feedback!

I totally understand the concern about depending heavily on two third-party APIs in one extension.

That said, the main motivation behind building this extension was to make task-based time tracking as frictionless as possible — by managing tasks via Todoist, and tracking working hours or billable time with Toggl.

With that in mind, here's a potential improvement I'd love your thoughts on:

Instead of directly integrating with Toggl, I could develop a custom time tracking feature inside the Todoist-focused extension and allow exporting the time logs in a CSV format compatible with Toggl.

This would preserve the unified experience I’m aiming for, while also keeping the extension self-contained and in line with Store policies.

Let me know what you think!

pernielsentikaer:
This idea?

> Instead of directly integrating with Toggl, I could develop a custom time tracking feature inside the Todoist-focused extension and allow exporting the time logs in a CSV format compatible with Toggl.

I'm not sure what the best approach is, we have multiple time tracking extensions in the store - I'm just affraid that it will give a lot more work if you make yet another extension doing so @thomaslombart

hayuse:
@pernielsentikaer 
@thomaslombart 

Thanks again for your thoughts — I really appreciate the open discussion!

Just to clarify: I completely agree with your concerns about reimplementing an entire time tracking system — that’s definitely not the goal.

The reason I suggested the idea of building a custom timer and exporting CSV was simply to respond to the earlier feedback about not integrating two third-party services in a single extension.

But given that reimplementing time tracking is actually a bigger concern, wouldn't it be more appropriate to leverage existing tools like Todoist and Toggl directly?

Integrating the two (without reinventing the wheel) seems like the most direct and lightweight way to support the common user need: tracking time for Todoist tasks.

Also, I may have misunderstood the initial concern about merging two third-party services — I interpreted it as an issue of tight dependency or coupling between distinct services,  
but maybe the real issue was that they’re seen as functionally similar (i.e. both productivity tools).  
If that's the case, I’m happy to clarify that the intent is to leverage their complementary strengths, not to merge overlapping functionality.

pernielsentikaer:
Can you have a look at the comment @thomaslombart

thomaslombart:
> I'm not sure what the best approach is, we have multiple time tracking extensions in the store - I'm just affraid that it will give a lot more work if you make yet another extension doing so @thomaslombart

I think there's a misunderstanding here. Why would I have to make another extension? 😄

I was suggesting that this extension could be published in the store as-is and live on its own, so @hayuse would maintain it. But if you think it doesn't make sense because there are already a lot of time tracking extensions in the store, here are other alternatives:

- Like @hayuse said, implement the CSV export action in Todoist in a generic way. I could see it providing value for users who want to have their Todoist data in another format (integrating with Toggl, feeding it to an LLM, making a spreadsheet out of this, etc.)
- Implement AI extensions in Toggl, which will make it possible to use both Toggl and Todoist inside an AI Command, but this requires having Raycast Pro, so it's not a suitable solution for free users.
- Keep it private if the demand behind this use case is low.

All in all, chaining extensions and combining commands is something users have been asking for, but we don't have a solution for that as of now. However, we're aware of it, and it'll probably come later down the road. 🙂

hayuse:
@pernielsentikaer 
@thomaslombart 

I realize now that I misunderstood parts of the earlier feedback... 🙇‍♂️  
This is my first time contributing to a project like this, so I might have gotten a bit too eager along the way 😅

That said, my hope is to publish this extension as it is.  
I'm planning to be its main user myself, so I’ll take full responsibility for maintaining it.  
Some of my coworkers are also looking forward to using it!😆

Thank you so much for your patience and thoughtful responses throughout — I really appreciate it.  
And if any part of this discussion came off the wrong way, I truly apologize.

I hope I can continue contributing to the community in the future. Thanks again!

pernielsentikaer (commented):
I'm a bit unsure if it's too niche - it sounds like it's very specific to your workplace. You could also create a private organization where you can publish it. Did you consider that?

hayuse:
@pernielsentikaer 
I actually didn’t know about the option to publish within a private organization — thanks for pointing that out!
I should have read the documentation more carefully, and I apologize for missing that.  
If the use case is considered a bit too niche for the Store, that might indeed be a better fit...

pernielsentikaer:
Let me know if you need to setup a private org and publishing it 😊

hayuse:
@pernielsentikaer 
@thomaslombart 
Thanks again for all the kind and thoughtful support!

PR #18036, opened 2025-03-21
Update aave-search extension

smbdy:
## Description
* Store previously fetched values in localStorage to allow offline usage
* There was an minor issue that crashed the extension when you typed "ad". Added a null check to fix that.

## Checklist

### DIFF
--- a/extensions/aave-search/CHANGELOG.md
+++ b/extensions/aave-search/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Aave Search Changelog
 
-## [1.1.0] - {PR_MERGE_DATE}
+## [1.1.0] - 2025-03-25
 
 ### Offline fallback & Fixes
 


extensions/aave-search/src/utils/useAddresses.ts
```diff
@@ -16,8 +16,16 @@ const TAG_MAP: Record<string, string[]> = {
   STATA_TOKEN: ["stata", "static"],
 };
 
+const STORAGE_KEY = "aave-addresses-cache";
+
 export const useAddresses = () => {
-  return useFetch<string, undefined, ListItem[]>(DATA_CSV_URL, {
+  const { value: cachedData, setValue: setCachedData } = useLocalStorage<ListItem[]>(STORAGE_KEY, []);
```
xmok, line 22: `useFetch` returns a cached version of the data by default since it is using `useCachedPromise` - we don't really need to have `LocalStorage` here as it is redundant. You can add `keepPreviousData: true` to the `useFetch` to make it more evident otherwise I suggest we remove this.

xmok (changes requested):
Yup can confirm that searching "ad" would crash but no longer is.

xmok (approved):
Disregard previous review.
If you did it for offline reasons, then it's OK.
LGTM!! 🚀

extensions/aave-search/CHANGELOG.md
```diff
@@ -1,8 +1,16 @@
 # Aave Search Changelog
 
-## [Initial Version] - 2025-03-21
+## [1.1.0] - {PR_MERGE_DATE}
+
+### Offline fallback & Fixes
+
+- Store previously fetched values for offline use
+- Fixed sorting error that was causing extension crashes. Kudos to @StErMi.
```
LitoMore, line 8:
```suggestion
- Fixed sorting error that was causing extension crashes. Kudos to [@StErMi](https://github.com/StErMi).
```

How about this? The readme cannot convert the mention to a link. But we could manually make it to a link.

pernielsentikaer (approved):

PR #18065, opened 2025-03-23
Add surge-outbound-switcher extension

the3ash:
## Checklist


pernielsentikaer:
Hi @the3ash👋 

It seems to some very similar to what https://www.raycast.com/ysj151215/raycast-surge - did you try that one out?

the3ash:
Hi @pernielsentikaer 

Thank you for your reply.

Yes, I have tried the Surge plugin from the store. However, it only supports controlling Surge via the HTTP API. For users like me who use a proxy configuration hosted on a remote server, configuring the HTTP API is not possible (as shown in the screenshot).

That’s why I created this plugin, which uses AppleScript to simulate click events to switch the Surge Outbound Mode. I believe there are other users who would find this plugin useful as well.


image
image

pernielsentikaer:
Sounds more like a really good enhancement to the current extension by @ysj151215, actually

ysj151215:
The main reason for the problem @the3ash mentions is that he is using a managed profile, and most configurations in a managed profile are not freely adjustable, including the HTTP API switch. In other words, if the profile does not have the HTTP API enabled by default, the best approach is to create a local copy and then enable it to use the plugin.

A less recommended alternative is to do this via a self-hosted (or trusted third-party) Surge configuration hosting platform.

the3ash:
Thank you for your response.

As @ysj151215  mentioned, it’s possible to achieve HTTP API control by copying the profile, but this would prevent me from receiving updates to the managed profile. That’s why I created this plugin.

pernielsentikaer:
Is the name enough to differentiate them?

the3ash:
@pernielsentikaer  I think that's enough. My extension is just about switching outbound mode, which is a more specific feature compared to the other one.

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #18068, opened 2025-03-23
Update zen-browser extension

elliothco:
## Description

Adds the ability to use the default search engine in the Zen Browser.

## Checklist

### DIFF
--- a/extensions/zen-browser/CHANGELOG.md
+++ b/extensions/zen-browser/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Zen Changelog
 
-## [Add Browser Default Search Engine] - {PR_MERGE_DATE}
+## [Add Browser Default Search Engine] - 2025-04-22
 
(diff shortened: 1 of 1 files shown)


extensions/zen-browser/package.json
```diff
@@ -30,9 +31,13 @@
           "type": "dropdown",
           "required": false,
           "title": "Search Engine",
-          "description": "Search Engine for query (default: Google).",
-          "default": "Google",
```
pernielsentikaer, line 34: Maybe it would be better to keep the old value and then users can switch to default? This way it's possible to keep the old behavoir

theherk: I suspect only non-Google users would even notice this change, and it is probably a welcome update for those of us. I think in this case the new behavior is desirable.

elliothco: Also an unintended but positive effect is that you can now also open links via the mod, whereas before if you typed in a link it would just search that link on your desired search engine

PR #18104, opened 2025-03-25
[Zed] Fix removing remote recent projects

### LINKED ISSUE #17837: [Zed] ...how to clear recent project list
### Extension

https://www.raycast.com/ewgenius/zed-recent-projects

### Description

Hello, I want to clear recent project list, how should I to do ?

_No response_

### Anything else?

_No response_

tmwrnr:
## Description

Fix removing remote recent projects.

Relates to #18027
Fixes #17837

## Screencast

n/a

## Checklist

### DIFF
--- a/extensions/zed-recent-projects/CHANGELOG.md
+++ b/extensions/zed-recent-projects/CHANGELOG.md
@@ -1,6 +1,6 @@
 #  Zed Recent Projects Changelog
 
-## [Fixes] - {PR_MERGE_DATE}
+## [Fixes] - 2025-03-25
 
 - Fix remove recent remote projects
 


tmwrnr:
@true-real-michael, @sunliang711

Can one of you check if remote projects can be removed?

pernielsentikaer (approved):

sunliang711:
> @true-real-michael, @sunliang711
> 
> Can one of you check if remote projects can be removed?

After pressing ctrl+x on a remote project, all remote projects disappeared. Then I open a new window to connect to a remote project, all disappeared remote projects come back.

tmwrnr:
@sunliang711 That's unfortunate. I really don't know what Zed does internally. Either someone with a remote setup needs to take a look, or I will only allow the removal of local projects.

@pernielsentikaer What is your suggestion?

true-real-michael:
@tmwrnr 
Turns out that Zed stores info about remote projects not only in its db, but in the `~/.config/zed/settings.json` as well. This explains the behavior experienced by @sunliang711. When I manually removed these records, the issue was gone. Curiously, it is necessary to remove **both** the record in the db and in the config. 

There is another thing however. It is sometimes necessary for the user to delete items twice. This is because remote projects are launched via `zed ssh://[user@](server)[:port]/(path)`. If the path looks like `~/path/to/project`, upon launch Zed automatically adds another entry `/~/path/to/project` which starts with `/`. 

This duplicate is not shown because the leading slashes are removed during processing, however because there are two entries, it appears to the user that they have to delete a project twice.

This seems like a problem on Zed's side, but I'm not sure. A possible solution on our side is to find and remove these entries too when user removes a project.

PR #18108, opened 2025-03-25
feat: Add country-code flag for mulvad exit-nodes

ajnart:
## Description

Added a simple feature that detects if an exit-node contains a country-code and displays the emoji for it. 
This feature is done in order to help people that are using the [Mullvad VPN](https://tailscale.com/mullvad) ad-on  

## Screencast
image

## Checklist

### DIFF
--- a/extensions/tailscale/src/exit.tsx
+++ b/extensions/tailscale/src/exit.tsx
@@ -1,15 +1,23 @@
-import { ActionPanel, List, Action, popToRoot, closeMainWindow, Image, Icon } from "@raycast/api";
+import { Action, ActionPanel, closeMainWindow, Icon, Image, List, popToRoot } from "@raycast/api";
 import { useEffect, useState } from "react";
 import {
-  StatusResponse,
-  getStatus,
-  getDevices,
-  tailscale,
-  sortDevices,
-  ErrorDetails,
-  getErrorDetails,
-  Device,
+    Device,
+    ErrorDetails,
+    getDevices,
+    getErrorDetails,
+    getStatus,
+    sortDevices,
+    StatusResponse,
+    tailscale,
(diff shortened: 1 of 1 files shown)


extensions/tailscale/src/exit.tsx
```diff
@@ -1,16 +1,24 @@
...
+    getStatus,
+    sortDevices,
+    StatusResponse,
+    tailscale,
 } from "./shared";
 
+function getCountryFlag(name: string): string {
+  const countryCode = name.substring(0, 2).toUpperCase();
+  if (countryCode.match(/^[A-Z]{2}$/)) {
+    return String.fromCodePoint(...[...countryCode].map(c => 0x1F1A5 + c.charCodeAt(0)));
```
itsmatteomanf, line 17: What happens if the device name starts with any country code combination, even if it's not meant to be a country code?

And second, this makes the list fully misaligned, because not all rows will have the initial emoji. For this a solution would be to add it to the subtitle or the other side of the row.

extensions/tailscale/src/exit.tsx
```diff
@@ -47,7 +55,7 @@ export default function ExitNodeList() {
           setIsActive(true);
         }
       } catch (error) {
-        setError(getErrorDetails(error, "Couldn’t load exit nodes."));
+        setError(getErrorDetails(error, "Couldn't load exit nodes."));
```
itsmatteomanf, line 58: Revert this change, the correct apostrophe is the one that existed before.

itsmatteomanf (changes requested):
I don't really agree with the changes, the flag isn't a specific node attribute, it's inferred based on a name which is subject to change and can be confusing for users, some changes are also required...

PR #18221, opened 2025-03-29
Auto language detection and some small fixes

Rafo1994:
## Description
Added automatic language detection option with native NPM package franc. Displayed  language which is used for OCR upon successful image OCR. Added fallback language so OCR doesn't fail if user types in wrong language code.

## Screencast
N/A

## Checklist

### DIFF
--- a/extensions/easy-ocr/CHANGELOG.md
+++ b/extensions/easy-ocr/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Tesseract OCR Changelog
 
-## [Enhancements] - {PR_MERGE_DATE}
+## [Enhancements] - 2025-04-15
 
 - Added automatic language detection option with either Raycast AI, Franc, LanguageDetect or Tinyld
 - Language used for OCR is now being displayed on successful OCR toast


extensions/easy-ocr/src/index.tsx
```diff
@@ -15,27 +16,46 @@ export default async function main() {
...
-        message: "No text found on image!",
+        title: `No text found on image! Language: ${languageUsed}`,
+        message: `No text found on image! Language: ${languageUsed}`,
       });
       return;
     }
 
+    if (getPreferenceValues<Preferences>().autodetect_lang) {
+      // Detect language
+      const detectedLangCode = franc(text);
```
LitoMore, line 45: Do you want to try the [raycast-language-detector](https://github.com/LitoMore/raycast-language-detector) instead?

It supports Raycast AI and also has some other fallback detectors.

extensions/easy-ocr/src/utils.ts
```diff
@@ -19,13 +19,13 @@ const isTesseractInstalled = async () => {
   return fs.existsSync(getPreferenceValues<Preferences>().tesseract_path);
 };
 
-async function checkGoogleCredentials() {
-  return !(!private_key || !client_email);
-}
+const isValidLanguage = (lang: string) => {
+  return lang in languages;
+};
```
LitoMore, line 24: The [raycast-language-detector](https://github.com/LitoMore/raycast-language-detector) has exposed a utility function for converting language code to the language name:

```typescript
import {languageCodeToName} from 'raycast-langauge-detector/utils';

const languageName = languageCodeToName(langCode);
```

So that you can drop the `languages.ts` and simplify your code.

extensions/easy-ocr/src/index.tsx
Rafo1994: Thanks for suggestion, just wanted to check if there are any chances of detected function imoprted from  "raycast-language-detector" will always return a language or is there a chance of returning undefined or similar data type?

extensions/easy-ocr/src/utils.ts
Rafo1994: Language.ts will be required because user can disable automatic language detection and use specific language which he specified in options. In case he typed in non existing language code, tesseract will throw an error, so we need a way to determinate if user has selected exisitng language.

Rafo1994: Also Tesseract uses 3-letter language codes, while your library returns only 2-letter code. Do you have a plan of adding option to be passed to detect function to accept different iso standards? If you need contributors, I'd like to help :D

LitoMore: I just checked the code of the three libraries I've used. All of them support the 3-letter code mode, so this should be very easy to implement.

I should be able to add this support to the package tonight. But PR is still welcome if you want to contribute it. I can leave the work to you.

extensions/easy-ocr/src/index.tsx
```diff
@@ -45,3 +68,14 @@ export default async function main() {
...
+
+async function autoDetectedLanguage(text: string) {
+  if (!getPreferenceValues<Preferences>().autodetect_lang) {
+    return;
+  }
+
+  // Detect language
+  return await detect(text, {
+    languageCodeFormat: LanguageCodeFormat.ISO_639_3,
+  });
```
LitoMore, line 80:
```suggestion
  return detect(text, {
    languageCodeFormat: LanguageCodeFormat.ISO_639_3,
  });
```

FYI: https://stackoverflow.com/questions/38708550/difference-between-return-await-promise-and-return-promise

LitoMore (changes requested):

extensions/easy-ocr/src/index.tsx
Rafo1994: Thanks for this!

pernielsentikaer:
Is this ready?

LitoMore (approved):
@pernielsentikaer LGTM

pernielsentikaer (approved):
;

PR #18227, opened 2025-03-29
Update browser-history extension

### LINKED ISSUE #18154: [Browser History] Fallback to web search
### Extension

https://www.raycast.com/crisboarna/browser-history

### Description

If the search term doesn't return any results in browser history, pressing enter should initiate a web search for the term.

Users that want to use Browser History search as their primary interface to opening new tabs. The current flow looks like this: 
1. search browser history for an item
2. no results
3. exit out of command
4. enter into search command
5. run search

it could look like:
1. search browser history for new or existing item
2. select / press enter
3. open new tab

### Anything else?

_No response_

ridemountainpig:
## Description
- Add a `search in browser` action when no search history is available. Close #18154.
- Update dependencies.

## Screencast
browser-history-5

## Checklist

### DIFF
--- a/extensions/browser-history/CHANGELOG.md
+++ b/extensions/browser-history/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Browser History Changelog
 
-## [Enhancement] - {PR_MERGE_DATE}
+## [Enhancement] - 2025-03-31
 
 - Add a `search in browser` action when no search history is available.
 - Update dependencies.


extensions/browser-history/src/search-history.tsx
```diff
@@ -1,4 +1,5 @@
-import { ActionPanel, getPreferenceValues, List } from "@raycast/api";
+import React from "react";
```
LitoMore, line 1: Duplicate `react` usage. See line 4.

extensions/browser-history/src/search-history.tsx
```diff
@@ -1,4 +1,5 @@
...
+import { ActionPanel, Action, getPreferenceValues, List } from "@raycast/api";
 import { useHistorySearch } from "./hooks/useHistorySearch";
 import { ReactElement, useState } from "react";
```
LitoMore, line 4:
```suggestion
import { ReactElement, isValidElement, useState } from "react";
```

extensions/browser-history/src/search-history.tsx
```diff
@@ -10,22 +11,27 @@ export default function Command(): ReactElement {
...
+  let noHistory = true;
 
   let entries = Object.entries(preferences)
     .filter(([key, val]) => key.startsWith("enable") && val)
-    .map(([key], idx) => useHistorySearch(key.replace("enable", "") as SupportedBrowsers, searchText))
+    .map(([key]) => useHistorySearch(key.replace("enable", "") as SupportedBrowsers, searchText))
     .map((entry) => {
       if (entry.permissionView) {
-        permissionView.push(entry.permissionView);
+        if (entry.permissionView && React.isValidElement(entry.permissionView)) {
```
LitoMore, line 22:
```suggestion
        if (entry.permissionView && isValidElement(entry.permissionView)) {
```

LitoMore (changes requested):

ridemountainpig:
@LitoMore, thank you for your review. I have already updated it. Could you please check again? Thank you.

extensions/browser-history/src/search-history.tsx
```diff
@@ -60,6 +59,18 @@ export default function Command(): ReactElement {
...
           }
         />
+      ) : noHistory ? (
+        <List.EmptyView
+          title={searchText ? `No ${searchText} history found` : "No history found"}
+          actions={
+            <ActionPanel>
+              <Action.OpenInBrowser
+                title="Search in Browser"
+                url={`https://www.google.com/search?q=${searchText ?? ""}`}
```
LitoMore, line 69: It looks like the code is hardcoded to `www.google.com`.

How about adding a support to customize the search engine?

ridemountainpig: I think support customize search engine is good. Do you think it's better to add common search engines to the preferences with a dropdown list, or to use `{{query}}` as a placeholder, like `https://www.google.com/search?q={{query}}` or `https://www.bing.com/search?q={{query}}`?

TKDKid1000: A query placeholder system would be more effective and customizable for different search engines. I would encourage the latter.

LitoMore: Yes. Using a `{{query}}` would be good. You may also need to encode the query:

```js
const query = encodeURIComponent(searchText);
const url = engine.replace(/your_pattern/, query);
```

ridemountainpig: I have added it.

LitoMore (approved):
@pernielsentikaer LGTM

pernielsentikaer (approved):

PR #18240, opened 2025-03-29
Update sportssync extension

daniyalmaster693:
## Checklist


daniyalmaster693:
I have question: I refactored and rewrote a lot of code since the last update, so the setup of commands is a lot more modular. However, the issue I've had is I'm not sure how to handle the isLoading attribute. Right now its a mess, and most commands don't have it.

daniyalmaster693:
Sorry it's been a while, had some exams recently, so haven't had time to work on the extension. But I've gone through and applied the changes greptile suggested. Could you take a look?

pernielsentikaer (changes requested):
How should the menubar work? I didn't select a sport and then I'll see this

CleanShot 2025-04-11 at 08 44 00@2x

Which calls this https://site.api.espn.com/apis/site/v2/sports/default/default/scoreboard

daniyalmaster693:
https://github.com/user-attachments/assets/1a2fb9c0-a28a-4639-8f71-40832c366db3

daniyalmaster693:
Yes it only displays one league, which is the one you select from the preferences

pernielsentikaer:
Is this used anywhere else? It should not make any network calls if you select none

daniyalmaster693:
The preferences are also used for the favorite team dashboard command. The reason it makes network calls is that it's just inputting "none" into the api call

daniyalmaster693:
I removed the none option from the preferences settings, and I replaced it with the first option for each dropdown (so the default is now: hockey, nhl, anaheim ducks). That way both the live scores and favorite team dashboard commands will still work

pernielsentikaer:
It still feels a bit cluttered, why does it show Select a game in the menu bar, there is also only one possibility to select - I'm a bit confused actually 😊

CleanShot 2025-04-14 at 08 08 16@2x

daniyalmaster693:
The games are within the submenu (hover over it) amd you'll see them. From there you can select a game

pernielsentikaer:
When you haven't selected a game, then we shouldn't make any fetch calls, should we? I can see that it calls `https://site.api.espn.com/apis/site/v2/sports/hockey/nhl/scoreboard` pretty often

daniyalmaster693:
I agree, but I'm not sure how to achieve that, since it's fetching the api call each time the menubar refreshes, which is updating the game item's scores.

extensions/sportssync/src/live-scores-menubar.tsx
```diff
@@ -0,0 +1,321 @@
...
+import { useFetch } from "@raycast/utils";
+import { useState, useEffect } from "react";
+import { getPreferenceValues } from "@raycast/api";
+
+// Get User Preferences
+
+interface preferences {
+  name: string;
+  id?: string;
+}
```
pernielsentikaer, line 11: You don't need to manually set preference types as this is autogenerated in `raycast-env.d.ts`

```suggestion
```

extensions/sportssync/src/live-scores-menubar.tsx
```diff
@@ -0,0 +1,321 @@
...
+        setSelectedGameIndex(Number(savedIndex));
+      }
+    })();
+  }, []);
+
+  const {
+    isLoading: scheduleLoading,
+    data: scheduleData,
+    revalidate: scheduleRevalidate,
+  } = useFetch<Response>(`https://site.api.espn.com/apis/site/v2/sports/${favoriteSport}/${favoriteLeague}/scoreboard`);
```
pernielsentikaer, line 126: there is an execute param you can use to check if it should run or not

daniyalmaster693:
But this isn't needed since I removed the option to select none all together. A default league, team and sport is already assigned (hockey, nhl, Anaheim ducks), therefore when it fetches game information, its returning actual info.

pernielsentikaer (approved):

PR #18276, opened 2025-03-31
Add ghostty-cheatsheet extension

conlonj25:
## Checklist

### DIFF
--- a/extensions/ghostty-cheatsheet/README.md
+++ b/extensions/ghostty-cheatsheet/README.md
@@ -4,7 +4,7 @@
 
 # Ghostty Keyboard Shortcuts
 
-Default keyboard shortcuts for Ghostty terminal emulator.
+Default keyboard shortcuts for Ghostty terminal emulator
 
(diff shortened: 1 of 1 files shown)


extensions/ghostty-cheatsheet/src/ghostty-cheatsheet.tsx
ridemountainpig, line 1: I think we can add the `List.Dropdown` feature to allow users to select their platform. For example, someone using a Mac to remotely access Linux may want to search for Linux shortcuts but cannot change the platform to Linux to view them.

conlonj25: I think I will just remove the OS selection and leave both sets of commands. If I can format it nicely.

You're right, it's better to have all the information available.

conlonj25:
> Would it make sense to built this into the [Ghostty extension](https://www.raycast.com/jarry_chung/ghostty) by @JarryChung

I guess it could fit in. I'll make a new PR for that instead.

pernielsentikaer:
That would be nice! 🔥

PR #18281, opened 2025-04-01
Update Tailscale extension

### LINKED ISSUE #18258: [Tailscale] Cannot switch to exit nodes that are shared by others
### Extension

https://www.raycast.com/tailscale/tailscale

### Raycast Version

1.94.2

### macOS Version

_No response_

### Description

There are exit nodes in my tailnet that my friend has kindly shared for me to use. When I try to switch to them using rayscast, nothing happens.

I suspect it's because the Mac Tailscale CLI cannot set exit nodes when you pass --exit-node argument with just a hostname for exit nodes that are shared.

```
$ tailscale set --exit-node=shared-exit-node # this fails with the following error message
2025/03/30 12:35:02 invalid value "shared-exit-node" for --exit-node; must be IP or unique node name
```

If you pass FQDN or IP address instead of hostname for this case, it works.

```
$ tailscale set --exit-node=shared-exit-node.tailhogefuga.ts.net # this works fine
```

I'm thinking that modifying code such as [this](https://github.com/raycast/extensions/blob/58a23ca299be438d1fa1f35b12cae87059f2b142/extensions/tailscale/src/exit.tsx#L102) would probably fix the issue, but I'm too new to actually try it myself.

### Steps To Reproduce

1. have a shared exit node in your tailnet
2. try to switch to that exit node using raycast
3. nothing happens

### Current Behaviour

Nothing happens.

### Expected Behaviour

Switch to exit node that is shared by others.

harupong:
fix #18258

## Description

Use IPv4 address to set exit node instead of hostname, so that shared exit node can be selected.

## Checklist

### DIFF
--- a/extensions/tailscale/CHANGELOG.md
+++ b/extensions/tailscale/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Tailscale Changelog
 
-## [Fixes] - {PR_MERGE_DATE}
+## [Fixes] - 2025-04-09
 
 - Better handling of shared exit node ([#18258](https://github.com/raycast/extensions/issues/18258))
 


ridemountainpig (changes requested):
Please add an entry to the CHANGELOG.md file. Here are the guidelines: [Version History](https://developers.raycast.com/basics/prepare-an-extension-for-store#version-history).

harupong:
@ridemountainpig Thanks for the headsup! CHANGELOG.md is updated.

itsmatteomanf (changes requested):
If we are going the IP route vs. the hostname route (makes sense given the issue presented), we do need to consider that this ACL option exists.

https://tailscale.com/kb/1337/acl-syntax#disableipv4

I'd probably use the IPv6 address, instead of the IPv4 one.
Can't really check right now, but do validate the command works...

harupong:
@itsmatteomanf Thanks for taking a look. I agree that IPv6 should be used, not IPv4.

I've modified the code to use IPv6, and validated that the command works as expected.

itsmatteomanf:
@harupong sorry for doing this to you (and sorry for the delay, didn't see the notification)... but I just checked, and using the full dns name (with the full address) works. What would you say to do a final update and use that? I'm not sure I can do the fix myself... will approve after.

https://github.com/raycast/extensions/blob/747e981af42cc7bf99bd6a95cdd23c5ee127d9f9/extensions/tailscale/src/shared.tsx#L163

This would result in using, for example, `no-osl-wg-008.mullvad.ts.net.` for one of the Mullvad nodes.

harupong:
@itsmatteomanf No problem!! I haven’t done these kind of things for quite some time, so I’m really having fun interacting with the community :)

I modified the related code, and squashed the related commits into one commit. Hope it’s not an issue…

itsmatteomanf (approved):
Looks good to me!

itsmatteomanf:
I think this is ready to be approved and merged, I think we need just @ridemountainpig to remove the requested changes and approve.

Well done @harupong!

ridemountainpig (approved):
Looks good to me, thanks!

pernielsentikaer (approved):

PR #18287, opened 2025-04-01
Add movie-quicklook extension

corymhayes:
## Checklist


corymhayes:
My idea for this, compared to the others, was to create a more streamlined, quick in and out experience, to see if a movie is worth taking the time to watch. One of the ways, and one of the things the other two omit, are user reviews. These can not only help gauge if a movie is something you want to watch, but can also help in gaining a better understanding of a film.

pernielsentikaer:
Sounds like what both of thoose extensions do already? @Aayush9029 and @ryan-exe what do you think?

Aayush9029:
you should add it as another command on existing tmdb ext, you are already using the tmdb api too

this way people don't have to download two separate extensions :D

pernielsentikaer:
Agree

corymhayes:
I will try that out, thank you for considering it

PR #18320, opened 2025-04-02
docs: asset folder clearing hint

wesleycoder:
## Description

Recently I got confused on the behavior of the extension when trying to check for `fs.exists` on the assets folder during development.
@LitoMore clarified for me that the assets folder gets cleared during the start of te development task, which makes much sense, but I thought this could be placed in the docs to help set the mental model for the devs.

I'm not sure this is the best place to put such hint, any advice is welcome.

Please, feel free to update the text to fit any writing guidelines that I'm not aware, or make it better in any way.

## Checklist

### DIFF
--- a/docs/information/file-structure.md
+++ b/docs/information/file-structure.md
@@ -37,7 +37,7 @@ Each command has a property `name` that maps to its main entry point file in the
 The optional `assets` folder can contain icons that will be packaged into the extension archive. All bundled assets can be referenced at runtime. Additionally, icons can be used in the `package.json` as extension or command icons.
 
 {% hint style="info" %}
-During development the `assets` folder for [`environment.assetsPath`](../api-reference/environment.md#properties) will be cleared when starting the extension via `npm run dev` to ensure you get the same behavior the user would get when installing the extension.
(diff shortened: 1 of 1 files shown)


docs/information/file-structure.md
```diff
@@ -36,6 +36,10 @@ Each command has a property `name` that maps to its main entry point file in the
 
 The optional `assets` folder can contain icons that will be packaged into the extension archive. All bundled assets can be referenced at runtime. Additionally, icons can be used in the `package.json` as extension or command icons.
 
+{% hint style="info" %}
+During development the `assets` folder of the extension in raycast will be cleared when starting the extension via `npm run dev` to ensure you get the same behavior the user would get when installing the extension.
```
LitoMore, line 40: Let's clarify which `assets` folder would be cleared. Because there are two `assets` folders. One is our project source `assets` folder, and the another one is `~/.config/raycast/extensions/YOUR_EXTENSION/assets`.

I think we can use `environment.assetsPath` instead in the documentation.

wesleycoder: Great suggestion, what about this phrasing then?
```suggestion
During development the `assets` folder for `environment.assetsPath` will be cleared when starting the extension via `npm run dev` to ensure you get the same behavior the user would get when installing the extension.
```

pernielsentikaer (approved):

PR #18359, opened 2025-04-03
Add mtg-card-search extension

JuanluR8:
## Description

Add new extension to consume Scryfall API and find MTG cards easily.
- Adds possibility to follow links to market webs in USA and Europe, as to follow links to popular MTG cards webpages.
- UI fits to work keyboard-only

## Screencast

mtg-card-search-1
mtg-card-search-2
mtg-card-search-4

## Checklist

### DIFF
--- a/extensions/mtg-card-search/src/types/server.ts
+++ b/extensions/mtg-card-search/src/types/server.ts
@@ -15,7 +15,7 @@ export interface ScryfallCard {
   cardmarket_id?: number;
   name: string;
   lang: string;
-  released_at: Date;
+  released_at: string;
   uri: string;
   scryfall_uri: string;
   layout: string;


pernielsentikaer:
Hi @JuanluR8👋 

What's the difference between this and [this one](https://www.raycast.com/madmaxmckinney/lotus-mtg-companion) in the store? They looks pretty similar to me

JuanluR8:
Hi @pernielsentikaer, yes, you're right, it's pretty similar but with a couple differences

I was the first one in doubt of opening a new extension or not because of that, but main difference is that one you mention is not quite adjust to UI and depends on mouse interaction to scroll, and some card information is broken due to MtG cards can be double face, but that isn't covering that, so it breaks

I wanted to fix that, same as UI to be more keyboard focus

In addition, that extension is quite focus to US, only showing card links to US webpages, showing avg price in USD and so on, I wanted to also add support to Euro prices and european stores

pernielsentikaer:
Sounds like an enhancement could be better actually, what do you think @MadMaxMcKinney

Our extension guidelines states
> There is an active extension available on our Store providing very similar value
    — We encourage the Author to iterate on existing extensions, add new functionality, fix bugs, or even re-write it.

mmckinney-figma:
Hey everyone! Yeah reading through this I think we could easily get @JuanluR8's first improvement in the existing extension
> Adds possibility to follow links to market webs in USA and Europe, as to follow links to popular MTG cards webpages.

I'm not totally sure what the feedback is on the keyboard navigation, everything in the current extension already is keyboard first since it's using just native Raycast components/behavior 🤷  

Re:currencies/locales — that could be a nice extension user-setting to provide and we could expand even further the supported currencies/locales without overwhelming the UI

MadMaxMcKinney:
Hey everyone! Yeah reading through this I think we could easily get @JuanluR8's first improvement in the existing extension
> Adds possibility to follow links to market webs in USA and Europe, as to follow links to popular MTG cards webpages.

I'm not totally sure what the feedback is on the keyboard navigation, everything in the current extension already is keyboard first since it's using just native Raycast components/behavior 🤷  

Re:currencies/locales — that could be a nice extension user-setting to provide and we could expand even further the supported currencies/locales without overwhelming the UI

pernielsentikaer:
Would it make sense to just have one? It seems like two is overkill. If it's possible to keep to one, otherwise users will be confused, or what?

MadMaxMcKinney:
@pernielsentikaer I agree, we should just have one. My feedback was about implementing this into the existing mtg extension.

JuanluR8:
I agree on the plan

But, as a hint from my point of view @MadMaxMcKinney I don´t know what do you think, but I'd reduce the sidebar information to avoid scrolling, and only leave scroll (where can be done with keyboard) only to main pane

I also use existing extension, and would like to get rid of the ruling text, and set common game modes legalities, what do you think? I set that in the sidebar, but at least I'd like to get that into the main pane

If you want to implement these new features, you can always re-use my code/deisng to help :)

PR #18371, opened 2025-04-04
Add solana-blockchain-explorer extension

dharminnagar:
## Checklist

### DIFF
--- a/extensions/solana-blockchain-explorer/src/utils/solana.ts
+++ b/extensions/solana-blockchain-explorer/src/utils/solana.ts
@@ -195,7 +195,7 @@ async function getNFTMetadata(nftAddress: string, network: Network) {
  * Detects the type of search query based on its format and content.
  * This function helps determine whether a search query is a domain, transaction,
  * block number, token/NFT, or regular address.
- * 
+ *
  * @param query - The search string to analyze
  * @param network - The Solana network to use for validation
  * @returns A promise that resolves to the detected search type


pernielsentikaer:
Hi @dharminnagar👋 

How is this compared to the the many other extensions in the store already doing this?

dharminnagar:
@pernielsentikaer I checked out the other extensions on the store before building mine. The main issue was that none of them had proper NFT support, and one kept crashing every couple of minutes.

Screenshot 2025-04-13 at 9 48 08 PM

Also, the command for opening on the explorer directly was also very complicated making the whole experience of using it quite frustrating.

I'll make the changes suggested and update the PR :)

pernielsentikaer:
If something doesn't work as expected, then it's still better to enhance it than just create another one and having multiple not really working - don't you think that's a good idea?

dharminnagar:
Well I do agree, but I'd made this and was using it for myself for a while so I thought why not add to the store so that everyone can use it and I created a PR. It's upon you to add it or not, but I'll say it'll only benefit the ecosystem :)

pernielsentikaer:
We have an [extension guideline](https://manual.raycast.com/extensions) about this so we don't end up with too many of the same extension in the store

*Reasons to be rejected*
> There is an active extension available on our Store providing very similar value
    — We encourage the Author to iterate on existing extensions, add new functionality, fix bugs, or even re-write it.

PR #18395, opened 2025-04-05
Update linkding extension

jose-elias-alvarez:
## Description

This PR more or less rewrites the Linkding extension, including a notable breaking change: it no longer handles multiple Linkding accounts, which is in line with the behavior of the [official Linkding browser exension](https://github.com/sissbruecker/linkding-extension). Removing the feature greatly simplifies the extension's code, which enabled me to add the following:

- Bookmark caching
- Offline search + fuzzy filtering
- Standardized keyboard shortcuts
- Better error / state transition handling, including a much smoother UI with faster load times and less flickering
- Many fixes, including a fairly critical one where UI state wasn't updated on bookmark deletion

I don't know what the Raycast policy is on breaking changes to an extension (I can see that removing a command locally somehow is *not* reflected in the commit created by the `publish` script, which is breaking CI) but I imagine most users should not be affected by this. Given that @InteractiveNinja created the extension in the first place and included this feature, it's possible that they don't want it removed, in which case I'd be happy to close this and just use my version locally.

## Screencast

raycast-linkding-1

raycast-linkding-2

## Checklist


InteractiveNinja:
Enhancements to performance are always great to see. Since I recently lack the time to really maintain this extension. I am completely fine with the removal of the account feature. I built it for my use case, but I no longer use it.

I'll try to check the changes this weekend, but maybe in the meanwhile @xmok is capable of approving it for me since I lack the time. He has already made some notable changes to the extension :)

xmok (changes requested):
Haven't gone through all of the code yet but if we're rewriting, I'd suggest we use Raycast best practices and with that in mind:
1. There's no need for `axios` - we can use native `fetch` and/or `useFetch`, `useCachedPromise` which should streamline code
2. No need for `LocalStorage` if we use caching through hooks unless we explicitly want offline support

React hooks are great but if there is another option like Raycast hooks, best to go for it!

jose-elias-alvarez:
Thanks for the feedback! I will update the extension to use modern Raycast best practices and also figure out what's going on with the stale command.

jose-elias-alvarez:
Moving to `useFetch` is a very nice improvement - thanks for the suggestion @xmok. The extension feels a lot smoother, and I'm happy with the diff stats.

One issue: I played around with a bunch of different options, but I wasn't able to wire a custom `Agent` into Node `fetch` such that the `ignoreSSL` option continues to work. If this is a dealbreaker, let me know, since I think it means that we'll have to move back to Axios / another library that does support it.

Separately, it looks like there's an issue with the Raycast `publish` script in that it doesn't seem to include *deleted* files in the commits it creates. I had to manually pull my branch and update it, which was pretty painful given the size of the repo. Is there something I'm missing, or is this a known issue?

xmok:
I haven't seen the updated code yet but feel free to use `axios`, `node-fetch` or whatever you want for the **ignoreSSL** feature as it makes sense to keep it for the users who might need it - it's ideal to minimize deps but it's not a requirement.  By combining with `useCachedPromise` we will have our own `useFetch` equivalent hook 💯.

re: the maual pulling, yeah the publish is a bit of a chore.

Please request another review after restoring the SSL feature 😅.

jose-elias-alvarez:
Thanks! Before I go ahead and put more work into this, do you have an opinion on `ignoreSSL` @InteractiveNinja? I imagine it was meant to support self-signed HTTPS, but it's pretty easy to get a legitimate certificate for self-hosting these days using Caddy / Certbot. 

If we're already including a breaking change, then I'd argue it makes sense to rip off the Band-aid here, too, but let me know.

InteractiveNinja:
You are right. I was actually a little bit lazy to trust the self-signed certs on all my systems, which is why I added this feature. I don't believe there are too many people who benefit from that. If there are any, they will make them noticeable :)
I don't mind the removal.

pernielsentikaer:
Sounds like a plan 😊

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #18427, opened 2025-04-07
Arc extension: (Re-)Add support to open blank incognito window

josephlou5:
## Description

A previous update removed the special handling for blank URLs when opening an incognito window ([link](https://github.com/raycast/extensions/commit/7160d0ab27090b4eaf44197a102d6637ca25bd85#diff-7ef5f0f191ae7fcbf2af0de7c017aa5519348fb4c95e4a3b42d0870b7b9b3a3b)). The `validateURL` function fails for blank URLs, which means opening blank incognito windows was no longer supported. That change broke my workflow, so I'm making a contribution to add support for that feature back.

## Checklist

### DIFF
--- a/extensions/arc/CHANGELOG.md
+++ b/extensions/arc/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Arc Changelog
 
-## [Improvements] - {PR_MERGE_DATE}
+## [Improvements] - 2025-04-16
 
 - When opening a new incognito window, open a blank tab if no URL is given.
   - A previous update caused a regression where this behavior was no longer


josephlou5:
Thanks @pernielsentikaer! I think this suggestion is not fully accurate. `validateURL` actually does show a proper toast on every path when returning `false`:

https://github.com/raycast/extensions/blob/a0899ae39e0d32f0d0983a92791fe09d0b2dd8ac/extensions/arc/src/utils.ts#L126-L147

I think this makes sense, as the function knows what specific error message to return, and you also don't want to repeat showing the error toast everywhere this function is called.

pernielsentikaer:
Awesome - so how can I test this easiest (want to check your changes)

josephlou5:
This is what I did, following instructions from the Raycast documentation:

```
# Change into Arc extension
$ cd extensions/arc
# Install and run
$ npm install && npm run dev
```

Now Raycast should show a development version of the Arc extension commands. I found the dev version of "Open New Incognito Window" and ran the command without providing a URL. It opened a new incognito window to the Google homepage, as expected, since that's the default in the settings. I then went to the Raycast Extension settings (Raycast Settings > Extensions), found the dev version of Arc, and removed the "Default URL" value for the "Open New Incognito Window" command. I then ran the command again, and a blank incognito window successfully opened.

If you try to run the command with the current public version of the extension, you should see a "No URL found" error if you remove the default URL and also provide no URL. My normal flow is assigning a keyboard shortcut to the command, which will result in an error toast with the same message.

josephlou5:
Hi @pernielsentikaer, any updates on this PR?

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #18515, opened 2025-04-11
[Things] Display only incomplete todo in the menu bar

vmrjnvc:
## Description

If you choose to log your completed tasks as “Daily” or “Manually” in the general settings of Things 3, these tasks will remain visible in your menu bar even after they are marked as completed. I believe it would be better to display only the first incomplete item from today’s list instead.

## Screencast

/

## Checklist

### DIFF
--- a/extensions/things/CHANGELOG.md
+++ b/extensions/things/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Things Changelog
 
-## [✨ Menu Bar Todo] - {PR_MERGE_DATE}
+## [✨ Menu Bar Todo] - 2025-04-14
 
 - Update the menu bar to display only incomplete todos from today’s list
 


pernielsentikaer:
I don't use Things, but they could also be showed in the bottom, would that make sense? What do you think @loris

vmrjnvc:
@pernielsentikaer Yes, that would solve the problem, but perhaps someone prefers to maintain a specific order in today’s menu bar list. As I understand it, this order may be disrupted if we display all completed items at the bottom.

bendrucker (approved):
I would agree that this is the right behavior for the circumstance. The setting in question is this:
 
Screenshot 2025-04-11 at 11 18 16

If you select anything other than `Immediately`, you expect your tasks to remain in place even after completion.

But the `Display To-Do` functionality in the Raycast menu bar command doesn't have any equivalent in the Things UI. There's no concept of "next task" in Things. It's a user invention.

I have some Shortcuts that use similar logic. My "next task" is the first one scheduled for today with an open status.

Issue #18541, opened 2025-04-12
[Toggl Track] Increasing entries history


matt08:
### Extension

https://www.raycast.com/franzwilhelm/toggl-track

### Description

When I create a new time entry, in most cases I use a list and a search engine to choose from previously made entries. 

My suggestion is to significantly increase the number of entries that can be searched, similar to the official Toggl Track app, where the history is much larger. Alternatively, the ability to set the number of entries in the settings.

### Who will benefit from this feature?

All people using Toggl Track to track the time of repetitive tasks.

PR #18543, opened 2025-04-12
Update toggl-track extension

ARui-tw:
## Description

Fix https://github.com/raycast/extensions/issues/18541
Add preferences to select the history lookback window.

## Screencast

image

## Checklist

### DIFF
--- a/extensions/toggl-track/CHANGELOG.md
+++ b/extensions/toggl-track/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Toggl Track Changelog
 
-## [Enhancements] - {PR_MERGE_DATE}
+## [Enhancements] - 2025-04-23
 
 - Add preferences to select the history lookback window
 


extensions/toggl-track/package.json
```diff
@@ -148,6 +149,37 @@
...
+        },
+        {
+          "title": "Weeks",
+          "value": "week"
+        },
+        {
+          "title": "Months",
+          "value": "month"
+        }
+      ],
```
franzwilhelm, line 181: While the UI looks nice, I think all we need is lookbackDays. Let's default to 14, and set a maximum of a reasonable number? I think we might end up overfetching and making loading speeds slow if we allow to fetch all time entries in history?

franzwilhelm (changes requested):
🙌 Thanks for fixing this. See my comment and let me know what you think

extensions/toggl-track/package.json
ARui-tw: Good point! I’ll update it to only include `lookbackDays` with a default of 14. I tried going back 60 days on my end and it works well, so I was thinking of setting that as the max limit. Does 60 days sound okay to you? That should help keep things performant while still offering a decent range.

franzwilhelm: Sounds good!

ARui-tw: @franzwilhelm I've updated it, can you give it a look?

pernielsentikaer:
@franzwilhelm can you have a second look?

franzwilhelm (approved):

franzwilhelm:
Done @pernielsentikaer

Issue #16969, opened 2025-02-11
[JetBrains Toolbox Recent Projects] ...


mite404:
### Extension

https://www.raycast.com/gdsmith/jetbrains

### Raycast Version

1.91.1

### macOS Version

15.3

#### Error:
```
TypeError: Cannot read properties of undefined (reading '/Users/ea/Programming/web/film-review-js')

ol:util.ts:343:36

---
340:     return undefined;
341:   }
342:   const contents = (await getReadJsonFile(settingsFile.path)) as JetBrainsToolboxSettings;
343:   const project = contents.projects[path] ?? {};
344:   project.favorite = true;
345:   project.launchMethod = appId;
346:   await writeSettingsFile(settingsFile.path, {
---

async add:recent.js:54:5710
```

### Expecting favorited projects to populate at top of list with recent projects following

mite404:
@gdsmith no worries, I updated JetBrains Toolbox and on the latest version of Webstorm, PyCharm, Rust Rover and Fleet. Same issue. I can search fine and it shows me recent projects, but when I try and favorite a project so it stays at the top of the list I get this error:

TypeError: Cannot read properties of undefined (reading '/Users/ea/Programming/web/ev-exploit')

ol:util.ts:343:36

---
340:     return undefined;
341:   }
342:   const contents = (await getReadJsonFile(settingsFile.path)) as JetBrainsToolboxSettings;
343:   const project = contents.projects[path] ?? {};
344:   project.favorite = true;
345:   project.launchMethod = appId;
346:   await writeSettingsFile(settingsFile.path, {
---

async add:recent.js:54:5710

PR #18545, opened 2025-04-12
Update jetbrains extension

PeraSite:
## Description
- Fix for #16969 
    - It is caused by `projects` field is not present in Jetbrain Toolbox's configuration file, `.settings.json`
    - Previously, removing last favorite project also removes `projects` field.
    - I fix it by marking `projects` field optional in type, then use optional chaining to handle the case.
- Fix when recent project doesnt contains last opened value
    - This is because some project like LightEdit projects doesn't include `projectOpenTimestamp` field in the `recentProjects.xml`

## Checklist

### DIFF
--- a/extensions/jetbrains/CHANGELOG.md
+++ b/extensions/jetbrains/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Jetbrains Changelog
 
-## [Bugfix] - {PR_MERGE_DATE}
+## [Bugfix] - 2025-04-14
 - Fix sorting projects which doesn't have a opened value
 - Fix crashing when `projects` field in `settings.json` is not present
 


extensions/jetbrains/package.json
```diff
@@ -8,7 +8,7 @@
   "contributors": [
     "adrienbaron",
     "dima-m711",
-    "vette"
+    "perasite"
```
dgrzedzielski, line 11: I would appreciate it if you could add yourself without removing me 😄

PeraSite: I know it's weird commit but it says you are not existing users in raycast account. Can you double-check for it?

dgrzedzielski: Oh that makes sense, I've changed my raycast handle to `dgrzedzielski`, sorry for the trouble. If you could update it by the way I would be grateful.

PeraSite: That works 😄 I made a new commit for that.

dgrzedzielski (approved):

extensions/jetbrains/package.json
dgrzedzielski: Thank you!

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

mite404:
it works!! ty @PeraSite !! 🫶🏻

Issue #18595, opened 2025-04-15
[Raindrop.io] ...Only Limited number of bookmarks (upto 25) showing up in Raycast extension


nimishpastaria:
### Extension

https://www.raycast.com/lardissone/raindrop-io

### Raycast Version

1.95.0

### macOS Version

15.4

### Description

I use Raindrop.io daily to shortlist job applications and then apply using its Raycast extension. I've noticed that when I save more than about 25 bookmarks, the extension doesn't display all of them and cuts off the list at around 25. This is really frustrating because it disrupts my workflow. I have to manually check the Raindrop app to see which ones are missing, filter them, and then use them.

### Steps To Reproduce

To reproduce the issue:  
1. Create a new collection (folder) in Raindrop.io.  
2. Add about 30-35 bookmarks to that collection.  
3. Use the Raycast extension of Raindrop.io to open the list of bookmarks in the collection by selecting "Search Bookmarks."  
4. Manually compare the number of bookmarks shown in Raycast with the total saved in the original Raindrop.io collection.  
5. You should notice a difference of about 5-10 bookmarks, as the extension seems to list only up to 25 bookmarks.

### Current Behaviour

I use Raindrop.io daily to shortlist job applications and then apply using its Raycast extension. I've noticed that when I save more than about 25 bookmarks, the extension doesn't display all of them and cuts off the list at around 25. This is really frustrating because it disrupts my workflow. I have to manually check the Raindrop app to see which ones are missing, filter them, and then use them.

### Expected Behaviour

The extension should list and exactly match the contents of the original raindrop.io collection, whether it has a small or large number of bookmarks. I'm assuming at least 100 should be enough, depending on the use case.

pewpeo:
It seems that the raindrop API defaults to 25.

One could set the maximum value, which is 50 according to the docs: https://developer.raindrop.io/v1/raindrops/multiple#common-parameters

pewpeo:
This should be solved with the new setting introduced: https://github.com/raycast/extensions/pull/18668

PR #18668, opened 2025-04-18
Update raindrop-io extension

pewpeo:
## Description

Addresses https://github.com/raycast/extensions/issues/18595

- Set maximum items for raindrop requests: 50 according to docs (https://developer.raindrop.io/v1/raindrops/multiple#common-parameters)
- Add option to fetch all results, i.e. perform multiple API requests for each result page.

Minor fixes:
- Fix changelog layout
- Fix default value for sortBy option
- Use suggested showFailureToast instead of showToast
- Replace my work github handle (oops...)

## Screencast

image

## Checklist

### DIFF
--- a/extensions/raindrop-io/CHANGELOG.md
+++ b/extensions/raindrop-io/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Raindrop.io Extension Changelog
 
-## [Enhancements] - {PR_MERGE_DATE}
+## [Enhancements] - 2025-04-22
 
 - Add option to fetch all results, i.e. perform multiple API requests for each result page
 - Set maximum items for raindrop requests


pernielsentikaer:
Hi @pewpeo👋 

@lardissone do you want to check this?

lardissone:
Looks good!
@pewpeo have you tried it with many bookmarks (more than ~5000)? I tried it myself and works fine (I have  5k and works fine, takes some time, but works).

pewpeo:
@lardissone I only have 250 and I don't even notice loading time. But that is a good point! Should we maybe limit it to a max value? IMO even 500 should be more than enough, wich would be 10 pages. But we could also use 1000 or 5000, WDYT?

pewpeo:
@lardissone Ok, the reporter in the linked issue suggested at least 100. I would go with 200 and keep the setting disabled by default. Personally, I also do not need so many results.

lardissone:
@pewpeo perfect. Makes sense, cap it to 200. Thanks!

pernielsentikaer (approved):

PR #18676, opened 2025-04-19
Add secure-password-generator extension

chanwonlee:
## Checklist


andreaselia:
Hey @chanwonlee 👋 

How does this differ from existing password generator extensions on the store such as [Password Generator](https://www.raycast.com/joshuaiz/password-generator)?

chanwonlee:
Thanks for pointing that out! I wasn't aware that there was already a similar password generator extension in the store.

My extension adds a few additional features:
- **Entropy Validation**: It checks that the generated password meets a minimum entropy requirement to ensure strong security.
- **Custom Special Characters**: Users can define their own set of special characters when generating passwords.

I focused on providing extra flexibility and stronger password quality through these features.

Would it be difficult to publish if there is already a similar extension?

andreaselia:
Hey @chanwonlee 👋 

Not difficult to publish, but it's better to avoid multiple extensions doing the same thing.

These sound like they could be enhancements as a contribution to the existing extension, would you be open to contributing those ideas to the existing extension by submitting a PR for it instead of a new extension for those enhancements?

LitoMore:
Agree with @andreaselia.

@chanwonlee I appreciate your hard work on this PR. But let's improve our existing [Password Generator](https://www.raycast.com/joshuaiz/password-generator) instead of creating a new one.

chanwonlee:
Thank you for the feedback and for considering my contribution.

I understand your point, and I agree that it makes sense to enhance the existing extension rather than creating a separate one.
However, due to time constraints, I won’t be able to work on contributing to the existing extension at this time.
Given that, I will step back from publishing this as a new extension.

Thank you again for your time and guidance!

andreaselia:
@chanwonlee you're welcome mate.

If you do end up with time to contribute what you'd like to the existing extension, then I look forward to seeing it. You could always create an issue too with what you'd like to see within the existing extension, and perhaps someone else will pick it up.

PR #18712, opened 2025-04-22
Add animated-window-manager extension

MatheusChein:
## Checklist


pernielsentikaer:
I tried to install this with the oneliner command and I can't get the commands to work even after doing the symlink

```
08:28:22 Error: Command failed: hs -c "maximizeWindowAnimated()"
/bin/sh: hs: command not found

    at checkExecSyncError (node:child_process:882:11)
    at execSync (node:child_process:954:15)
    at main (/Users/pernielsentikaer/.config/raycast/extensions/animated-window-manager/maximize.js:7574:39) {
  status: 127,
  signal: null,
  output: [
    null,
    <Buffer >,
    <Buffer 2f 62 69 6e 2f 73 68 3a 20 68 73 3a 20 63 6f 6d 6d 61 6e 64 20 6e 6f 74 20 66 6f 75 6e 64 0a>
  ],
  pid: 18100,
  stdout: <Buffer >,
  stderr: <Buffer 2f 62 69 6e 2f 73 68 3a 20 68 73 3a 20 63 6f 6d 6d 61 6e 64 20 6e 6f 74 20 66 6f 75 6e 64 0a>
}
```

Do you have an idea? If I manually set it to `/opt/homebrew/bin/hs`, then it works

pernielsentikaer:
I added that step manually and it works from a console, but from Raycast it can't find `hs` as you can see here:

```
08:28:22 Error: Command failed: hs -c "maximizeWindowAnimated()"
/bin/sh: hs: command not found
```

pernielsentikaer:
That would be nice; many extensions do checks like that or let the user set a path in preferences, but performing the check automatically is the best solution for the users :)

MatheusChein:
> That would be nice; many extensions do checks like that or let the user set a path in preferences, but performing the check automatically is the best solution for the users :)

Done! Could you check again please? I tested locally and it worked nice with the new validation

pernielsentikaer (approved):
Seems to work for me; let's see how it goes

MatheusChein:
> Seems to work for me; let's see how it goes

I added a [new PR](https://github.com/raycast/extensions/pull/18865) with some improvements to the validation and configuration instructions, including checking if it's bash / zsh or Fish shell. Should be much better now!

PR #18771, opened 2025-04-25
Things - Set Reminders in Today and Upcoming

dinocosta:
## Description

This Pull Request updates the available list item actions for both the "Show Today List" and "Show Upcoming List" commands in order to allow users to change the to-do's reminder.

- For the "Show Today List", it'll only display times after the current time
- For the "Show Upcoming List", all times between `00:00` and `23:45`, with 15-minutes interval, will be shown

Unfortunately, as far as I can tell, Things does not provide information about whether the to-do has a reminder or not, and for when that reminder is set, so we can't really display that information in the list.

If this ends up being something useful, I'll go ahead and also update both the "Edit To-Do" action and the "Add New To-Do" commands to also allow setting the to-do's reminder.

## Screencast

<details>
<summary>Action</summary>

set_reminder_action
</details>

<details>
<summary>Today</summary>

today_set_reminder
</details>

<details>
<summary>Upcoming</summary>

upcoming_set_reminder
</details>

## Checklist

### DIFF
--- a/extensions/things/CHANGELOG.md
+++ b/extensions/things/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Things Changelog
 
-## [✨ Reminders] - {PR_MERGE_DATE}
+## [✨ Reminders] - 2025-05-06
 
 - Update the "Today" and "Upcoming" lists to allow updating todo's reminders.
 - Fixed issue with the Deadline action to correctly remove a deadline from a todo.


bendrucker:
> Things does not provide information about whether the to-do has a reminder or not

I found the same, you can verify this here:

https://support.apple.com/guide/script-editor/view-an-apps-scripting-dictionary-scpedt1126/mac

Screenshot 2025-04-25 at 09 43 19


The `dueDate` is the deadline and `activationDate` does not reflect reminder times.

dinocosta:
@bendrucker Thank you so much for the great feedback! 🙌 

> And second, I don't think the list of times on a fixed increment is a good UI. A simple text input would be better. But at that point it feels like you'd just be reaching for `Action.PickDate` rather than a form and so it's not really clear this is actually any different than the existing functionality.

100% agree here! Unfortunately Raycast's API does not have an `Action.PickTime` that we could leverage here, hence me adding the list of times, which I also don't like to much because it forces the user to pick a specific time, as in, the user can't choose `10:10`, for example.

> I would first suggest fixing this before adding a dedicated reminder action.

Makes sense! I'll revisit my approach to see if I can simply make it work with the existing `Schedule > Date` action 🙂

Thank you once again for your feedback, appreciate it!

dinocosta:
@bendrucker I pushed a new commit – https://github.com/raycast/extensions/pull/18771/commits/959a954dfd1873c78ef10c63871037c4ee1320b3 – which attempts to update the "Schedule > Date" action to also support setting reminders.

It does so by updating the `type` to `Action.PickDate.Type.DateTime` so that both date and time can be specified, as well as converting the selected date to the user's timezone on the `onChange` callback.

There's a small detail which I don't love with this approach and it has to do with the fact that we don't have a way to tell whether the user actually selected only a date (`Tomorrow`) or a date and a time (`Tomorrow 10:00`). I ended up finding out that, at least on my machine, the selected date would have it's milliseconds set to `1` if the user only specified the date, so I leveraged it and, if the milliseconds are set to `1` the implementation will simply update the to-do's date, while if the milliseconds are set to `0`, it'll set both the date and time. Although it works I'm not sure if this is something that will always remain true, and I'm not a big fan of relying on undocumented behavior, but let me know what you think 👍 

I can also re-add the "Set Reminder" action, leveraging `Action.PickDate` with `Action.PickDate.Type.DateTime` and properly setting both the `min` and `max` values, for both the "Today" and "Upcoming" lists. I had a commit with this implementation but when running `npm run publish` it squashed both commits 🫠

Thanks!

bendrucker:
Nice! I'll give this a try and let you know. Re: time vs date, did you try this?

https://developers.raycast.com/api-reference/user-interface/actions#action.pickdate.isfullday

dinocosta:
> Nice! I'll give this a try and let you know. Re: time vs date, did you try this?
> 
> https://developers.raycast.com/api-reference/user-interface/actions#action.pickdate.isfullday

Wondering how I completely missed this while reading the docs... 🫠

Thank you for pointing it out! I'll try updating the changes to leverage this and see if it works as expected, will keep you posted!

dinocosta:
@bendrucker I've updated the implementation with this commit - https://github.com/raycast/extensions/pull/18771/commits/f1cc0a13b7e061357ff4c0d5bae116345056a179 . Feel free to take a look and let me know what you think, thank you so much for your help! 🙇

pernielsentikaer:
Hi @dinocosta👋 

@loris do you want to check this?

extensions/things/src/components/TodoListItemActions.tsx
```diff
@@ -208,8 +208,26 @@ New title:
             title="Date…"
             icon={Icon.Calendar}
             min={new Date()}
-            onChange={(date) => schedule(date ? date.toISOString() : 'anytime')}
-            type={Action.PickDate.Type.Date}
+            onChange={(date) => {
+              if (date) {
+                // Before proceeding, we need to ensure we create a date time in the user's timezone.
+                const offset = new Date().getTimezoneOffset();
+                const scheduleDate = new Date(date.getTime() - offset * 60 * 1000).toISOString();
```
bendrucker, line 215: Things only uses ISO 8601 date time strings for `creation-date` and `completion-date`. The `when` attribute uses a special format:

https://culturedcode.com/things/support/articles/2803573/#json
https://culturedcode.com/things/support/articles/2803573/#overview-data-types-date-time-string

You shouldn't have to offset the `Date` object itself, which should be correct/local. `toISOString` is just outputting in the wrong time zone (Z). `toLocaleString()` is the local-time alternative but it doesn't adhere to Things expected format precisely and varies by locale.

This should do the job:

```typescript
function formatDateTime(date: Date): string {
  const year = date.getFullYear();
  const month = String(date.getMonth() + 1).padStart(2, '0');
  const day = String(date.getDate()).padStart(2, '0');
  const datePart = `${year}-${month}-${day}`;

  if (Action.PickDate.isFullDay(date)) {
    return datePart;
  }

  // Include a reminder time
  const hours = String(date.getHours()).padStart(2, '0');
  const minutes = String(date.getMinutes()).padStart(2, '0');

  return `${datePart}@${hours}:${minutes}`;
}
```

extensions/things/src/components/TodoListItemActions.tsx
```diff
@@ -208,8 +208,26 @@ New title:
             title="Date…"
             icon={Icon.Calendar}
             min={new Date()}
-            onChange={(date) => schedule(date ? date.toISOString() : 'anytime')}
-            type={Action.PickDate.Type.Date}
+            onChange={(date) => {
+              if (date) {
```
bendrucker, line 212: Minor: prefer returning edge cases early over `if`/`else`. E.g.:

```typescript
if (!date) {
  schedule('anytime')
  return
}

// handle date
```

bendrucker (commented):
Nice! Looks good to me in terms of functionality. Left some hints on how to handle date/time formatting consistently/directly.

extensions/things/src/components/TodoListItemActions.tsx
dinocosta: Makes sense to me! Updated with this commit – https://github.com/raycast/extensions/pull/18771/commits/213a7beb6c61fabe6f8fabb8089fc5b13aa3b1dc ✌️

dinocosta:
@bendrucker Thank you for the suggestion to also use the formatted date time logic for the "Set Deadline" action. I went ahead and pushed a new commit – https://github.com/raycast/extensions/pull/18771/commits/2c4453c1bcc64a15a28548388a8071ed94128bfb – that moves the logic to a function and re-uses it in the "Set Deadline" action.

While testing this I also noticed that the "No Date" option, when selecting a new deadline, didn't actually remove the to-do's deadline date, so I went ahead and updated the `setDeadline` function so that if `null` is provided it passes an empty string to Things, effectively removing the deadline date, let me know what you think! ✌️ 

Thank you so much for the help so far!

extensions/things/src/components/TodoListItemActions.tsx
```diff
@@ -208,8 +208,29 @@ New title:
...
+              const month = String(date.getMonth() + 1).padStart(2, '0');
+              const day = String(date.getDate()).padStart(2, '0');
+              const datePart = `${year}-${month}-${day}`;
+
+              // If the user picked a full day, we avoid providing the time, as
+              // Things leverages the time part to understand whether a reminder
+              // should be set.
+              if (Action.PickDate.isFullDay(date)) {
+                schedule(datePart);
+              }
```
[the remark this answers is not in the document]
dinocosta: Nice catch! 👀 

Marking this as resolved, as it was fixed in the meantime ✅

extensions/things/src/components/TodoListItemActions.tsx
```diff
@@ -161,6 +174,29 @@ New title:
     }
   }
 
+  function timeEntries(command: FilteredCommandListName) {
+    const now = new Date();
+    const times = [];
+    const startHour = command === 'today' ? now.getHours() : 0;
+    let startMinute = command === 'today' ? Math.ceil(now.getMinutes() / 15) * 15 : 0;
```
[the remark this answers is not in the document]
dinocosta: Marking this as resolved seeing as this diff is outdated and this logic has since been removed 🧹

extensions/things/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Things Changelog
 
+## [✨ Reminders] - {PR_MERGE_DATE}
+
+- Update the "Today" and "Upcoming" lists to allow updating todo's reminders
```
bendrucker, line 5: Add your bug fixes here too

bendrucker (approved):
Good catch on `null`-ing the deadline, LGTM!

extensions/things/CHANGELOG.md
dinocosta: @bendrucker Thank you for the reminder! Added here - https://github.com/raycast/extensions/pull/18771/commits/88500c551e90d2a0a4cffc13341820b0eed19dfd ✅

PR #18786, opened 2025-04-25
Add clipboard-line-concatenator extension

rogerwcpt:
## Description

Given the text to prepend or append from the command input,  each line on the current clipboard will be modified by concatenating the prefix and suffix text, and put a new entry back in the clipboard for pasting.

## Screencast

image

## Checklist

### DIFF
--- a/extensions/clipboard-line-concatenator/src/clipboard-concatenator.ts
+++ b/extensions/clipboard-line-concatenator/src/clipboard-concatenator.ts
@@ -15,7 +15,7 @@ export default async function Command(props: { arguments: Arguments }) {
   }
 
   let clipboardText: string | undefined;
-  
+
   try {
     clipboardText = await Clipboard.readText();
   } catch (error) {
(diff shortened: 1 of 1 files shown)


RogerDerekWeiss:
Hi guys, any progress on this to get it moving.?

pernielsentikaer:
This sounds a lot like [this extension](https://www.raycast.com/koinzhang/append-clipboard) - we should probably update the name a bit - could you look at that extension by @koinzhang

RogerDerekWeiss:
> This sounds a lot like [this extension](https://www.raycast.com/koinzhang/append-clipboard) - we should probably update the name a bit - could you look at that extension by @koinzhang

Thank you I will have a look at your comments.  I did review this extension but I found it didn't append for EACH line, rather just at the beginning and end.

For example

aa
bb
cc
dd

That extention would just append a character (for example a quote.  ") the beginning and end . so the result would be this

"aa
bb
cc
dd"


A common requirement in CSV file manipulation is to apped for each line, thats while I spedified LINE appender in the exention. So the result for my extension would be:

"aa"
"bb"
"cc"
"dd"

or 


aa,
bb,
cc,
dd,

Which now forms a very nicely formatted CSV line

pernielsentikaer:
Sounds like a new command or preference for the above extension, what do you think @koinzhang

rogerwcpt:
Thank you for the PR review @pernielsentikaer .

I've made the changes as suggested by your PR comments and tested the extension again.  I've pushed an update to resolve the comments.

pernielsentikaer:
Doesn't it make sense to add it to the existing extension instead of creating a new one?

rogerwcpt:
It probably does, to reduce fragmentation of extensions. What is the correct way of merging the two? Should i fork the current main branch and then copy my fucntionality to the other plugin, then push a PR?

pernielsentikaer:
Using the [Fork action](https://developers.raycast.com/basics/contribute-to-an-extension) on the extension should be best option 😊

PR #18806, opened 2025-04-26
Add fuzzy search for Google Chrome bookmarks

oscmarb:
## Description

Add configurable fuzzy search for Google Chrome bookmarks. Three options are available in the extension:
-  Exact (default): returns truly exact results.
-  Balanced: balances accuracy and flexibility, allowing for relevant matches with minor deviations.
-  Loose: offers the most flexibility, capturing a broader range of potential matches.

I would apply this change to tab and history search too, but I'm looking for feedback to proceed.

## User problem
Users usually make typos in their queries or don't remember the exact name of the bookmark, so if they have a lot of bookmarks, they end up entering a partial query and checking the result list.

## Checklist

### DIFF
--- a/extensions/google-chrome/src/util/index.ts
+++ b/extensions/google-chrome/src/util/index.ts
@@ -9,6 +9,7 @@ import {
 import { getPreferenceValues } from "@raycast/api";
 import { Preferences } from "../interfaces";
 import { BookmarkDirectory, HistoryEntry, RawBookmarks } from "../interfaces";
+import Fuse from "fuse.js";
 
 type ChromeFile = "History" | "Bookmarks";
 const userLibraryDirectoryPath = () => {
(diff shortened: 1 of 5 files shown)


a-laughlin:
Thanks for submitting this PR. How does it change the default behavior from what exists now?

oscmarb:
The default behavior of the bookmark search is "**_Exact_**", which looks for precise matches. For instance, if you have a bookmark titled "Whatsapp" and you type "Watsapp" you will not get any results. This is how bookmark searches are implemented today.

In addition to this, I have introduced two new search options: **_Balanced_** and **_Loose_**. These options enhance the search functionality by allowing approximate matches. Using the earlier example, both "Watsapp" and "Whatsapp" would yield results in your bookmarks search, as the algorithm now performs a fuzzy search rather than a strict literal search.

Here are some screenshots:

- Config menu:
Screenshot 2025-04-30 at 18 18 19

- Exact search:
Screenshot 2025-04-30 at 18 19 09

- Balanced search:
Screenshot 2025-04-30 at 18 18 47

- Loose search (only a few results for privacy 😁):
Screenshot 2025-04-30 at 18 19 41

a-laughlin:
I like the direction and am mostly good with it.

For anyone else reviewing, it does respect partial results even in "exact" mode, so searching for "zzz" does yield all results that include the substring "zzz" anywhere.

Agreed that the same searching should apply across all search types: bookmarks, tabs, history, and search-all. Especially since they're implemented differently right now. The most flexible of the implementations is the tabs search, so that's the best one to compare behavior against.

Things I see as blockers:
- The current default implementation matches substrings in strings of any length. Fuse only matches in strings up to 60 characters. Setting `ignoreLocation:true` fixes this.
- The current tab search treats space separated words as "and" conditions. For example, "github io" and "io github" both match "github.io" anywhere in the title or url. More cases [here](https://github.com/raycast/extensions/blob/0779d4f9fce2ce3e9baa1001a598bb6a1e64981b/extensions/google-chrome/src/hooks/useTabSearch.tsx#L8). Neither "exact" or "balanced" replicate this behavior. Loose yields so many results that what it's matching on is unclear.
- The help popup for the configuration requires a common understanding of words like "exact" and "variations". Those can mean many things. Exact confused me. Providing examples of how matches differ in each setting will help understand what it will mean to change the setting.

I have some minor type concerns and will comment on those in the file. Overall looks like it's going in a great direction though. Thanks for adding it!

extensions/google-chrome/src/util/index.ts
```diff
@@ -81,6 +82,42 @@ export const getBookmarks = async (profile?: string): Promise<HistoryEntry[]> =>
...
+      threshold: threshold,
+      keys: keys,
+    });
+
+    return fuse.search(query).map((result) => result.item);
+  }
+
+  return items.filter((item: T) =>
+    keys.some((key) => (item as unknown as Record<string, string>)[key].toLowerCase().includes(query.toLowerCase()))
+  );
```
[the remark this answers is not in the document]
a-laughlin: Agreed with what greptile said here.  Constrain the function with the actual type it will receive.
`filterElements(items: HistoryEntry[], query: string | undefined, keys: (keyof HistoryEntry)[], threshold: number)`.

Doing that highlights a potential error case on the "toLowerCase()" call.

extensions/google-chrome/src/util/index.ts
```diff
@@ -81,6 +82,42 @@ export const getBookmarks = async (profile?: string): Promise<HistoryEntry[]> =>
...
+  ["balanced", 0.4],
+  ["loose", 0.7],
+]);
+
+export function filterElements<T>(items: T[], query: string | undefined, keys: string[], threshold: number) {
+  if (!query) {
+    return items;
+  }
+
+  if (threshold) {
```
a-laughlin, line 104: Why this conditional? Wouldn't  searching with a threshold of 0 have the same effect?

extensions/google-chrome/src/util/index.ts
a-laughlin, line 115: What's the purpose of wrapping the search in an `if (threshold)` conditional? That creates two ways of searching to maintain. Curious if it's possible to simplify to one way.

pernielsentikaer:
Awaiting approval by @a-laughlin (and maybe @rgomezcasas)

Also, this branch has conflicts that must be resolved, can you look into that @oscmarb

PR #18896, opened 2025-05-01
Messages Paste OTP Regex Update

th3jamin:
The original logic to select the OTP from the "filtered phone" list was missing certain characters like ':' '.' '/'. This was causing issues with messages such as the following:

> BANK: Enter code 678456 (expires in 30 min) on [bank.com](http://bank.com/). We will not call to ask you for the code. Questions? Call 800.123.4567

**The above would paste "4567"**

> Verification Code: 287961 Expiration Date: 4/30/2025 10:02:26 AM Mountain Standard Time. BANK will never contact you for this code. Do not share this code with anyone.

**The above would paste "2025"** 

This PR fixes that by adding additional common characters the regex and adding a bit more logic to select the largest OTP  from the filtered list that is > 10 digits as most OTP's are 6-8 digits.

### DIFF
--- a/extensions/messages/CHANGELOG.md
+++ b/extensions/messages/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Messages Changelog
 
-## [Fix OTP Code Parsing for Messages with Special Characters] - {PR_MERGE_DATE}
+## [Fix OTP Code Parsing for Messages with Special Characters] - 2025-05-08
 
 Fixed issues with wrong number being parsed by adding additional common characters the regex and adding a bit more logic to select the largest OTP from the filtered list.
 


haydencbarnes:
... this has broken the changes I made on the previous fix to REGEX for this extension. Example:

> From: SoFi
> NEVER share this code with anyone including SoFi agents. If anyone asks for this code, HANG UP and call 1 (844) 908-7634.
> Login Code: 561603
> 
> Returns 7634.

haydencbarnes:
@th3jamin The bug is that the colon ⁠: shouldn't be treated as phone punctuation in this context. Why was this added to the Regex?

PR #18930, opened 2025-05-03
Update migadu extension

LBBO:
## Description

Currently, identites are only searchable by their address and not their name. Especially in combination with #14896, this can be annoying most addresses are meaningless to me and I care much more about the name. Therefore, the main goal of this PR is to make the names searchable.

I tried adding the name to the keywords instead, since those are also searchable (see [docs](https://developers.raycast.com/api-reference/user-interface/list#search-bar)). However, the keywords seem to be matched just from start to end, so e.g. the name "Some Test" would be matched for the query "so" but not "te". I tried doing some ad-hoc stuff to make the keywords more useful but couldn't get it working. And since the names are probably interesting anyway, I just decided to display them in the title and use the regular filtering.

## Screencast

image

## Checklist

(Didn't check assets since I didn't make any changes related to assets)

### DIFF
--- a/extensions/migadu/CHANGELOG.md
+++ b/extensions/migadu/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Migadu Changelog
 
-## [Make Identities searchable by their name] - {PR_MERGE_DATE}
+## [Make Identities searchable by their name] - 2025-05-07
 - Identities are now listed by their name and address, meaning that both can be used to search for them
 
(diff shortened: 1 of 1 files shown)


pernielsentikaer:
Thanks for your contribution @LBBO 

@xmok do you want to check this?

xmok (approved):
Pretty straightforward. 🚀

pernielsentikaer (approved):

PR #18944, opened 2025-05-04
Add command to Zed extension

mathieubellon:
## Description

The PR introduces a new command that allows users to define a custom folder containing all local projects. This command is unique because it lists all folders, not just those previously opened with Zed.

After creating a new project or attempting to open old ones that you haven't accessed in Zed before, you can either launch Zed and open the project from there or start your terminal and use the command `$ zed myproject` to open your project in Zed.

When you do this multiple times a day, it creates friction and prevents you from leveraging the full power of Raycast, which allows you to accomplish everything with just a few keystrokes, without using your mouse.

This command enables you to set a parent folder in your preferences and list all projects within it, including old or newly created ones and open them in Zed immediately.

## Screencast
CleanShot 2025-05-04 at 15 56 41

## Checklist

### DIFF
--- a/extensions/zed-recent-projects/src/open-from-projects-folder.tsx
+++ b/extensions/zed-recent-projects/src/open-from-projects-folder.tsx
@@ -11,17 +11,20 @@ import {
 } from "@raycast/api";
 import { usePromise } from "@raycast/utils";
 import { getZedBundleId } from "./lib/zed";
-import { readdirSync } from "fs";
-import { createContext, ComponentType, useContext } from "react";
+import { promises as fs } from "fs";
+import { join } from "path";
+import { createContext, ComponentType, useContext, useState, useEffect } from "react";
 import { zedBuild } from "./lib/preferences";
 
 const preferences = getPreferenceValues<Preferences.OpenFromProjectsFolder>();
 
 const dir = preferences.projectsdir;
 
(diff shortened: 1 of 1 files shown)


pernielsentikaer:
Hi @mathieubellon👋 

ewgenius:
@mathieubellon thanks for the contribution!

I have couple concerns about this new command:
- new command UI doesn't provide pinned projects and git branch labels, which is inconsistent with recent projects command and might cause confusion. I would expect to have consitent UX for similar commands under single extension.
- limiting to a single folder is somewhat opinionated UX, in my view. For example, I have multiple projects roots, but I wouldn't want to list all projects from all of them. I think it would be more flexible to allow setting multiple folders.

I think for this specific scenario it would make sence to create this as a separate extension and allow users to configure the target editor/app from all available apps, and not limit it to Zed only.

What do you think?

mathieubellon:
Hello @ewgenius. 

I agree with your concern. To be honest, after publishing this PR, I even wondered if this new command was not in fact completely outside the scope of this extension, which is named "recent project." 

So I am totally ok to move out of this extension and start a new one and target Zed / VScode for example

Thanks for your quick feedbacks, I will proceed and close the PR

mathieubellon:
As seen with @ewgenius this PR will move to a proper extension

PR #18964, opened 2025-05-05
Update flush-dns extension

c0b41:
## Checklist

### DIFF
--- a/extensions/flush-dns/CHANGELOG.md
+++ b/extensions/flush-dns/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Flush DNS Changelog
 
-## [Windows Support] - {PR_MERGE_DATE}
+## [Windows Support] - 2025-06-17
 
 - Added Windows support.
 


extensions/flush-dns/package.json
```diff
@@ -16,7 +16,7 @@
   "commands": [
     {
       "name": "index",
-      "title": "Flush DNS",
+      "title": "Flush Dns",
```
rasmusbe, line 19: 'Flush Dns' should be 'Flush DNS' since DNS is an acronym

extensions/flush-dns/src/index.ts
```diff
@@ -1,66 +1,134 @@
...
+    },
+    shell: "/bin/bash",
+  },
+  win32: {
+    commands: {
+      ipconfig: "ipconfig /flushdns",
+    },
+    shell: undefined, // Uses cmd.exe by default
+  },
+};
```
rasmusbe, line 27: Since this is not dynamic, I think we should keep it as const for better type hinting.

```suggestion
const OS_COMMANDS = {
  darwin: {
    commands: {
      dscacheutil: "/usr/bin/dscacheutil -flushcache",
      mDNSResponder: "/usr/bin/killall -HUP mDNSResponder",
      mdnsflushcache: "/usr/bin/discoveryutil mdnsflushcache",
    },
    shell: "/bin/bash",
  },
  win32: {
    commands: {
      ipconfig: "ipconfig /flushdns",
    },
    shell: undefined, // Uses cmd.exe by default
  },
} as const satisfies Record<string, CommandSet>;
```

extensions/flush-dns/src/index.ts
```diff
@@ -1,66 +1,134 @@
...
+      mDNSResponder: "/usr/bin/killall -HUP mDNSResponder",
+      mdnsflushcache: "/usr/bin/discoveryutil mdnsflushcache",
+    },
+    shell: "/bin/bash",
+  },
+  win32: {
+    commands: {
+      ipconfig: "ipconfig /flushdns",
+    },
+    shell: undefined, // Uses cmd.exe by default
```
rasmusbe, line 25: Should this be undefined or `"cmd.exe"` in that case. Feels like cmd.exe is making more sense, but I don't have a strong opinion regarding this.

extensions/flush-dns/src/index.ts
rasmusbe: The key in Record should probably be partial of the return value of platform() and not string, for good type hinting about what platforms are valid values.

rasmusbe (changes requested):
Great addition! I'm unable to try the Windows part it since I only own a mac, but it looks valid. 👏 

I haven't verified that the mac parts still work, I will be able to do that on Thursday since I'm not having access to my development computer until then but I've added some small feedback based on what I see in the code.

extensions/flush-dns/package.json
c0b41: Linter was complaining, that why i changed

extensions/flush-dns/src/index.ts
c0b41: raycast windows don't have showHUD api right so hard to know if throwing error, but Nodejs docs says right [here ](https://nodejs.org/api/child_process.html#child_processexecsynccommand-options) node uses process.env.ComSpec as default shell as cmd.exe

c0b41:
i dont see any lint error locally is ci bugged or something? @pernielsentikaer

extensions/flush-dns/package.json
pernielsentikaer: I will look into this, let's use DNS for sure :)

pernielsentikaer:
Should be better now @c0b41 - I'll leave this until we have support for everything that this extension uses

c0b41:
everything working on windows side, but someone should test macos stuff again, i hope i didn't break anything

pernielsentikaer:
@rasmusbe did you have a chance to test it yet?

c0b41:
can someone test this changes

pernielsentikaer:
This will be reviewed by our Windows team when they are ready

rasmusbe (approved):
Sorry for the delay, I have tested it now on Mac, and it still works as it should. Great job @cob41!

PR #19039, opened 2025-05-08
Update kill-node-modules extension

M-Subhan1:
## Description

The PR aims to fix the following issues in the extension:

- Slow searching
- Searching across the entire files system
- Unable to specify configuration options
- Code readability & modularity

The current implementation fixes the pain points by:
- Replacing the slower du npm implementation with native du + fast-glob for file search and file size rendering
- Improving initial load by using `useCachedPromise` hook from Raycast
- Allowing the user to specify `rootFolder` and `scanDepth` in extension settings to avoid scanning the entire system (inspiration from `Git Repos` extension)
- Rendering useful error messages
- Refactoring the code to be modular and easier to work with

## Checklist

### DIFF
--- a/extensions/kill-node-modules/CHANGELOG.md
+++ b/extensions/kill-node-modules/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Kill Node Modules Changelog
 
-## [Overhauled Extension for Performance and Configurability] - {PR_MERGE_DATE}
+## [Overhauled Extension for Performance and Configurability] - 2025-05-12
 
 This major update addresses key performance issues, introduces essential configuration options, and significantly refactors the codebase for better readability and modularity.
 


extensions/kill-node-modules/src/kill-node-modules.tsx
```diff
@@ -1,97 +1,187 @@
...
+import { useMemo, useState, useCallback, useRef } from "react";
 import { exec } from "child_process";
 import { promisify } from "util";
 import path from "path";
 import { formatSize } from "./utils/format";
+import { tryCatch } from "./utils/try-catch";
 
 const execPromise = promisify(exec);
 
-export type NodeModulesItem = {
```
NicolasMontone, line 12: the `NodeModulesItem` was removed and It's used in the file `useNodeModules.ts`

M-Subhan1: Weird. I had already removed the useNodeModules file. It's not being used anymore

M-Subhan1: It's not visible to me locally either. Any idea what the issue might be?

NicolasMontone: Wired I cloned the PR and I saw the import, also the cli is failing

extensions/kill-node-modules/src/kill-node-modules.tsx
```diff
@@ -1,97 +1,187 @@
...
+  const abortable = useRef<AbortController>();
 
-  const handleDelete = useCallback(async (id: string, path: string) => {
-    if (
-      await confirmAlert({
+  const {
+    data: allItems,
+    isLoading,
+    mutate,
+  } = useCachedPromise(
```
NicolasMontone, line 68: The only “bad” thing about useCachedPromise is that we lose the ability to stream the files and start showing the node_modules as soon as we have some data. But I think it really improves performance, so it’s a trade-off we can accept and revisit streaming later.

NicolasMontone (approved):
LGTM

extensions/kill-node-modules/src/kill-node-modules.tsx
M-Subhan1: Yeah. I wasn't able to find a way to "stream" the data whilst maintaining a good initial load experience. If we reduce the pageSize within the useCachedPromise, theoretically we should be able to get the "streaming" thingie working whilst having some initial data as well. 

However, as more data streams in the positioning of the elements would radiply change especially if we are sorting by size. The reason being to extract good performance we need to delay folder size calculation based off of when that data is needed. Since size isn't known initially as more items render the sort applies and the list would move a lot.

M-Subhan1:
> As a minor thing I would add a skeleton item in the list, I think we end up with a better UX image

I wasn't able to find any raycast components to allow us to add the skeleton loader. I think right now they only aim to use the "isLoading" state to show progress. Is there any extension you've seen that uses the loader? I'll have a look at their source code so see what they've done I guess.

pernielsentikaer:
Let's continue with this for now, `isLoading` is the thing to use right now

pernielsentikaer (approved):

PR #19079, opened 2025-05-10
Add CryptoTracker extension

qusaismael:
## Description
This PR introduces **CryptoTracker**, a new Raycast extension that leverages the CoinMarketCap API to deliver real-time cryptocurrency data right in Raycast. It provides:

- **List Top Coins**: View the top cryptocurrencies by market cap.  
- **Bitcoin Price**: One-click access to detailed BTC metrics.  
- **Search Coins**: Search any coin by name or symbol.  
- **Customizable Currency & Refresh**: Choose your preferred fiat (USD, EUR, GBP, JPY) and automatic refresh interval.
- **Rich Details**: Inline 7-day sparkline, market cap, volume, supply, and description.

## Screencast

image
image
image

## Checklist
- [x] I ran `npm run build` and tested the distribution build in Raycast  
- [x] I checked that files in the `assets` folder are used by the extension  
- [x] I verified that README assets are correctly referenced


qusaismael:
@pernielsentikaer Thanks for looking into it.

_Yes, both get crypto data from CoinMarketCap._

The main difference is **HOW**: mine uses the official CoinMarketCap API, while the other extension crawls the website as far as my knowledge <correct me if im wrong>.

API version is more stable and reliable, it won't break whenever CoinMarketCap changes its site layout.

also it gives users a really detailed look at each coin (with descriptions, supply data, and a visual price chart) and includes configurable auto-refresh for live updates.

I actually see the UX of it is better than the extension you mentioned, cuz the user can see all the coins prices at once with price changes and up/down arrows, with no offense to the guy who created it, good job that he did it at the first place.

pernielsentikaer:
@moojing are you still around? Are you up for a revamp? 

In our [Extensions Guidelines](https://manual.raycast.com/extensions) you'll see that you should try to enhance rather than creating a new extension

qusaismael:
@pernielsentikaer I am open to enhance the old extension, but I don't think he is here since it has been 2 weeks,  is there any other solution?

I want to add a status bar to my extension too.

pernielsentikaer:
Just send an e-mail to @moojing

pernielsentikaer:
If you're up for enhancing then I think we should do that 😊

moojing:
Sorry for the late reply, I've been busy job hunting these days. 

The reason why I used the scraping approach is because I didn't want users to figure out how to set up an API key before they could start searching.  I wanted the experience to be more straightforward and intuitive. **Mine extension only focus on letting user get the data they want the most (which is the price), and they can open the website to get any detail data they want with merely 1 click.**

So if we "enhance" my extension by switching to an Api approach, it would be another totally different extension from my personal perspective.

But according to the rule in guideline, Maybe we can consider offering two different user flows,
- one that fetches data via the API
- one that follows the current scraping-based flow 

(IMO, the extension I made and the one by @qusaismael are fundamentally different, given their entirely different approaches. It would be fairer for him to treat them as separate extensions.)

pernielsentikaer:
Then let's continue as is. Is everything ready for another review?

qusaismael:
Yes, the extension should be working fine as is.

PR #19197, opened 2025-05-15
Update paperless-ngx extension

csigritz:
## Description

Added 2 new fields to the extension configuration:

- showCorrespondentInSubtitle
- showDateInSubtitle

The user can select one or both of them. This will take control of what will be displayed as subtitle in the search result list. Before this update there was the correspondent only. I personally do not use them in Paperless, show the subtitle never showed anything. The date of the document ist something i needed. So now users can decide what they wanna see:

- Correspondent (Default)
- Date
- Both

## Checklist

### DIFF
--- a/extensions/paperless-ngx/CHANGELOG.md
+++ b/extensions/paperless-ngx/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Paperless-ngx Changelog
 
-## [Add subtitle options] - {PR_MERGE_DATE}
+## [Add subtitle options] - 2025-05-16
 
 - Added configuration options for subtitle in search result list
 - Show correspondent, date created or both


extensions/paperless-ngx/src/models/preferences.model.ts
```diff
@@ -3,4 +3,6 @@ export interface Preferences {
   apiToken: string;
   gridMode: boolean;
   dateFormat: string;
+  showCorrespondentInSubtitle: boolean;
+  showDateInSubtitle: boolean;
 }
```
pernielsentikaer, line 8: You don't need to manually set preference types as this is autogenerated in `raycast-env.d.ts`

csigritz: Just wanted to match it the existing code

pernielsentikaer: Would be better to remove it completely, do you mind doing that?

csigritz: 🫡 Done

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #3126, opened 2022-10-05
New Extension: MenuTime

Ek2100:
Jazz up your menubar with customization your way with MenuTime! Add custom text that you can chose from the menutime preferences!
mac-os-x-os-x-big-sur-hd-wallpaper-preview
SCR-20221005-mjg

PR #19303, opened 2025-05-20
Update Credits for One Thing Extension

Ek2100:
## Description

Added myself (Ek217) as a contributor in the package.json to reflect my role as the original creator and early developer of this extension.

I initially built and submitted the extension in PR #3126, including the concept, design, and implementation. Later, I communicated that I was no longer actively maintaining the project and explicitly said, “You can take over the extension,” meaning others could take over maintenance of the extension.

However, I did not expect that my name would be removed entirely from the project metadata and the Raycast Store listing, not only as a contributor but also as the author.

While I’m glad to see the extension evolve and improve, it’s important to me that my original contribution is recognised properly. This request is to restore the author credit where it’s due, however in saying this I don't want to undermine all the work of the contributors who built the extension further and am wondering what compromise can be made

### DIFF
--- a/extensions/one-thing/CHANGELOG.md
+++ b/extensions/one-thing/CHANGELOG.md
@@ -1,5 +1,8 @@
 # One Thing Changelog
 
+## [Update Credits] - 2025-05-21
+- Added Ek217 as contributor for the One Thing extension
+
 ## [✨ AI Enhancements] - 2025-02-21
 
 ## [Enhancement] - 2024-03-21


Ek2100:
I’m not really fussed about authorship or anything like that. But the author credit usually reflects the original creator, not the current maintainer, kind of like how the author of a book from 500 years ago doesn’t change just because a new publisher is printing it. It definitely feels rude looking at it now not just being removed as a creditor but even being removed as the author of my own extension.

pernielsentikaer:
Could you add an entry in the `CHANGELOG.md` file? Here's our guidelines about that: [Version History](https://developers.raycast.com/basics/prepare-an-extension-for-store).

Ek2100:
> Could you add an entry in the `CHANGELOG.md` file? Here's our guidelines about that: [Version History](https://developers.raycast.com/basics/prepare-an-extension-for-store).

I have updated the changelog 👍🏻

pernielsentikaer (approved):

Issue #19263, opened 2025-05-18
[Parcel] "Unknown date" For Package In My Deliveries


damian-duchnowski:
### Extension

https://www.raycast.com/marcmagn1/parcel

### Raycast Version

1.98.5

### macOS Version

15.4.1

### Description

In the My Deliveries section, one package (I don't have others atm) shows "Unknown date" for all entries under History.

```
Package: 16 May, UPS
Expected Delivery: Not available
Status: In Transit
Carrier: UPS
Tracking Number: 1ZA0A6019112360725
History
🔵 Unknown date The package is being held for a future delivery date. We’ll attempt to deliver on the date requested. (Northampton, United Kingdom)
⚪️ Unknown date Processing at UPS Facility (Northampton, United Kingdom)
⚪️ Unknown date Arrived at Facility (Northampton, United Kingdom)
⚪️ Unknown date Departed from Facility (Stanford Le Hope, United Kingdom)
⚪️ Unknown date Arrived at Facility (Stanford Le Hope, United Kingdom)
⚪️ Unknown date Departed from Facility (London, United Kingdom)
⚪️ Unknown date Arrived at Facility (London, United Kingdom)
⚪️ Unknown date Collection Scan (London, United Kingdom)
⚪️ Unknown date Shipper created a label, UPS has not received the package yet.
```

### Steps To Reproduce

Use "Add Delivery" and add 1ZA0A6019112360725, then go to the "My Deliveries" section. I already had it in that section as it's under my account.

### Current Behaviour

No dates shown under History

### Expected Behaviour

Show the dates available in Parcel
![Image](https://github.com/user-attachments/assets/1b234bdb-af6d-4ccf-9440-ca6d2d8f7d62)

damian-duchnowski:
The problem was with different carriers using different date formats (American or European), and the extension wasn't expecting that.
Went ahead and fixed it in the PR.

Example:
1ZA0A6019112360725 - UPS - European
FC920957912GB - Royal Mail - American

```
[...]

PR #19356, opened 2025-05-22
Update parcel extension

damian-duchnowski:
## Checklist

### DIFF
--- a/extensions/parcel/CHANGELOG.md
+++ b/extensions/parcel/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Parcel Changelog
 
-## [Improved Date Handling] - {PR_MERGE_DATE}
+## [Improved Date Handling] - 2025-05-28
 
 - **Enhanced date recognition:** Improved parsing of delivery dates, including better support for European formats (like DD.MM.YYYY HH:mm and DD.MM.YYYY HH:mm:ss).
 - **More reliable display for unknown dates:** Unspecified or unparseable dates will now more consistently display as "Not available", providing clearer information.


damian-duchnowski:
Created to fix #19263

extensions/parcel/src/my-deliveries.tsx
```diff
@@ -46,57 +46,75 @@ export default function Command() {
...
+      if (dateString.includes(",")) {
+        date = new Date(dateString);
       }
 
-      return date.toLocaleDateString(undefined, {
-        year: "numeric",
-        month: "short",
-        day: "numeric",
-      });
+      // If American format failed or wasn't used, try European format (DD.MM.YYYY HH:mm)
```
bendrucker, line 59: Better to pull in https://date-fns.org if the dates aren't strictly formatted. It's much more readable to use formatting strings versus imperatively extracting each field from multiple formats.

```ts
import { parse, isValid } from 'date-fns';

export const parseDate = (str: string): Date | null => {
  for (const fmt of ['MMMM dd, yyyy HH:mm', 'dd.MM.yyyy HH:mm']) {
    const date = parse(str, fmt, new Date());
    if (isValid(date)) return date;
  }
  return null;
};
```

extensions/parcel/src/my-deliveries.tsx
```diff
@@ -46,57 +46,75 @@ export default function Command() {
...
 
-      // Create 24-hour time format: "14:30"
-      const timeFormatted = date.toLocaleTimeString(undefined, {
-        hour: "2-digit",
-        minute: "2-digit",
-        hour12: false,
-      });
+  // Format tracking history dates in a compact format: "Feb 26, 14:30"
+  const formatCompactDate = (dateString: string | undefined | null): string => {
+    if (!dateString || dateString === "--//--") return "Not available";
```
bendrucker, line 96: Avoid hiding magic constants in code. Pull out `--//--` to a constant outside of this function and provide some proper documentation of when it's returned.

More importantly, this constant is a smell—is this really the way unknown dates are handled or just what you happened to see for the carrier you tried?

A more robust potential approach would be to look for a string without digits.

extensions/parcel/src/my-deliveries.tsx
bendrucker: Instead of `null` you may just want to return a string. But this is kind of a waste, if `parseDate` can't parse the date, it's fine to return an empty value. The decision to use the original string if the parsed version is empty is best left to the caller.

extensions/parcel/CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # Parcel Changelog
 
+## [Improved Date Handling] - {PR_MERGE_DATE}
+
+- Enhanced date parsing to support both American and European date formats
+- Added fallback to display raw date string if parsing fails
+- Improved error handling and logging for date-related issues
+- Refactored date parsing logic into a reusable utility function for better maintainability
```
bendrucker, line 8: The changelog is for users, don't include internal changes that have no user impact

extensions/parcel/CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # Parcel Changelog
 
+## [Improved Date Handling] - {PR_MERGE_DATE}
+
+- Enhanced date parsing to support both American and European date formats
```
bendrucker, line 5: Avoid assigning your own names to things—use and cite established standard. While MM/DD/YYYY can be reasonably called "American," nowhere else has such a simple 1:1 relationship. The common date format in Europe just swaps the month and date, i.e. DMY instead of MDY. 

https://en.wikipedia.org/wiki/List_of_date_formats_by_country

```suggestion
- Parse `DD.MM.YYYY HH:mm` formatted dates as used by European and other non-US carriers
```

extensions/parcel/CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
...
+- Added fallback to display raw date string if parsing fails
+- Improved error handling and logging for date-related issues
```
bendrucker, line 7:
```suggestion
- Improved error handling and logging for date parsing
```

extensions/parcel/src/my-deliveries.tsx
```diff
@@ -46,57 +46,75 @@ export default function Command() {
     return `(in ${daysUntil} day${daysUntil !== 1 ? "s" : ""})`;
   };
 
-  // Format date in a more readable way: "Feb 26, 2025"
-  const formatFriendlyDate = (dateString: string): string => {
-    if (!dateString) return "Unknown date";
-
+  // Utility function to parse dates in various formats
```
bendrucker, line 49:
```suggestion
```

The original comment gave a hint about the return string's format. Here, the type signature tells you everything you need, the comment is duplicative.

pernielsentikaer:
Hi @damian-duchnowski👋 

Could you look into the suggestion from @bendrucker (Thanks btw!)

bendrucker:
The Parcel API doesn't normalize parts of its carrier responses and so this will be a bit of a moving target. Making it easy for contributors to append a formatting string to an array of options should make it manageable though. And the fallback to the raw string is definitely the right move.

So these are mostly code quality comments, overall the functionality looks good. 👍🏻

pernielsentikaer:
Would be nice to get them in before merging 😊

damian-duchnowski (commented):
Feedback actioned, push incoming

pernielsentikaer (approved):

PR #19388, opened 2025-05-24
Add formula1-data extension

yashgupta-11:
## Checklist


pernielsentikaer:
Hi @yashgupta-11👋 

Did you see [this extension](https://www.raycast.com/csigritz/f1-standings) before starting on this one?

yashgupta-11:
Hey!

Yes, I did see that and was using it, but a problem was that it did not have an option to switch between 12 and 24-hour time. Also, since then, I have made the extension a lot more useful by adding the option to see driver and team information within the first two original commands. Furthermore, I also added another command to view the live updates happening in an ongoing race weekend, which was completely missing from the original extension. Overall, I would say that my extension is basically the extension by Christian but more enhanced and has a lot more features and QOL improvements.

pernielsentikaer:
In our [Extensions Guidelines](https://manual.raycast.com/extensions), you'll see that you should try to enhance rather than create a new extension  


Changing the time could be a preference, and the rest sounds like really good enhancements - what do you think @csigritz?

yashgupta-11:
Great! If this extension gets approved, do I submit the content that is shown on the store page, or is that something the team handles? If I need to submit a write-up and pictures, I will be happy to do so. Additionally, I have a lot more ideas for the extension that I might implement in the future, so in that case, do I need to go through this same review process?

pernielsentikaer:
Improving the existing extension would be great; we don't want to have two similar extensions in the store, as that would just confuse users

PR #19526, opened 2025-05-30
Add copee extension

gianpispi:
## Description
Copee provides three commands:

### Copy Text as File
This copies the current selected text as file.

### Make File from Clipboard
It generates and copies a text file from last entry in your clipboard.
      
### Paste File from Clipboard
It generates a text file from last entry in your clipboard and pastes it into the selected area.

## Preferences
- **File Name** is the name of the file that will be generated.
- **Show the file in Finder** if checked will open the finder when the file is created.
- **File Directory** the directory the file will be saved. By default and if empty or "\", the temp directory of the file system will be used.

## Screencast

https://github.com/user-attachments/assets/f68f8165-99fa-4d0d-aeb8-3c5d49dc373c

https://github.com/user-attachments/assets/2224fae1-7932-470c-8cef-380eacddff2e

## Checklist


extensions/copee/src/make-file-from-clipboard.ts
```diff
@@ -0,0 +1,6 @@
+import { Clipboard } from "@raycast/api";
+import { handleTextToFile } from "./api/helpers";
+
+export default async function main() {
+  await handleTextToFile(Clipboard.readText, Clipboard.copy);
```
ridemountainpig: There is a type error here.

```suggestion
  const text = await Clipboard.readText();
  await handleTextToFile(text, Clipboard.copy);
```

gianpispi: oh thanks @ridemountainpig! I deleted that file tho, not sure why it's still versioned. will push a fix now

pernielsentikaer (commented):
It feels like the output directory should be required? I didn't like the place it was saved and would like to be able to pick that right away. What do you think?

gianpispi:
Hey @pernielsentikaer! Thanks for the review.
In my mind I don't care where it gets saved. I just want a quick file with what I copied, to paste it around, or just to drag and drop. So the default folder is tmp, since it will get overridden next time you run the action.

What do you think?

gianpispi:
also I just had the idea to add a preference to generate the filename based on the content of the generated file. Can you please help me with preferences setup?

In particular:
- is there a way to disable a preference based on another preference?
- and is there a way to disable a preference based on the user account? (how should the extension behave if the user doesnt have access to AI? - so show or hide the preference based on the user account or should I show an error when generating the file name and save it with the default name?)

thanks you!

pernielsentikaer:
No that's not possible, might be in the future though 🫠

gianpispi:
Okay cool.
@pernielsentikaer what do you think about the folder setup? Would it help if i set the file name as non mandatory?

PR #19545, opened 2025-06-01
Update homeassistant extension

krsntn:
## Description

Added a custom layout to the Custom Entities command that categorizes entities by their states for easier viewing and management.

### Before
before

### After
after

## Checklist


xilopaint:
I like the idea of using `List.Section`, but in this case I would remove the `List.Item.Accessory` to avoid redundancy.

pernielsentikaer:
Hi @krsntn 

Did you see the feedback from @xilopaint

krsntn:
@xilopaint If we remove the list.item.accessory, users won’t be able to see the current status for some other entities like TVs or speakers. So removing it would make it harder for users to distinguish the status of these items at a glance.

homeassistant 2025-07-08 at 16 33 13

Issue #19674, opened 2025-06-08
[Country Lookup] Unable to load countries


enger05:
### Extension

https://raycast.com/pernielsentikaer/country-lookup

### Raycast Version

1.99.4

### macOS Version

15.5

### Description

When country lookup is opened, it is unable to fetch any country

#### Error:
```
Error: Bad Request
    at Object.Z0 [as current] (/Users/USER/.config/raycast/extensions/cd79e301-04c0-4a4d-bf41-149e80058965/index.js:8:315661)
    at /Users/USER/.config/raycast/extensions/cd79e301-04c0-4a4d-bf41-149e80058965/index.js:8:316567
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
```

### Steps To Reproduce

Opening the country lookup command

pernielsentikaer:
Looks like a recent change to the endpoint we're calling: https://gitlab.com/restcountries/restcountries/-/issues/265

I'll see if I can figure out the best way to handle this OR feel free to look into it yourself and make a PR

enger05:
Hello,

No worries, I've just assumed that it was an issue on my part for some reason. Thanks

pernielsentikaer:
No, I need to fetch data in a new way since the old method got deprecated 🫠

pernielsentikaer:
This is now fixed, you can now run `Check For Extension Updates` and try it out 😊

enger05:
Thank you for the quick fix and quick response I very much appreciate it

PR #19693, opened 2025-06-09
Update `Country Lookup` - Offline Support since API is now limited + Modernize

xmok:
## Description

Unfortunately, the API now only returns a maximum of 10 fields since costs are too high for the maintainer to cover on their own (ref: https://gitlab.com/restcountries/restcountries/-/issues/265). I initially tried to come up with a method that fetches 10 fields at a time then caches but realized it would be quicker, neater and lead to a better UX to have a copy of the file locally similar to what I did in PR #12930. The file is ~1.5MB so I `uglified` to bring down the size and prepended ISO date so future updates can know if the file needs to be updated or not.

This will close Issue #19674.

- The countries data is now stored locally allowing offline access (the API no longer returns all data)
- Modernize extension to use latest Raycast config

## Screencast

https://github.com/user-attachments/assets/af2d331e-bf20-44b5-a0f6-04ad2cbb074c

## Checklist

### DIFF
--- a/extensions/country-lookup/CHANGELOG.md
+++ b/extensions/country-lookup/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Country Changelog
 
-## [Offline Support] - {PR_MERGE_DATE}
+## [Offline Support] - 2025-06-10
 
 - The countries data is now stored locally allowing offline access (the API no longer returns all data)
 - Modernize extension to use latest Raycast config


pernielsentikaer (approved):

pernielsentikaer:
Thanks for updating, I was about to do it 😊

PR #19742, opened 2025-06-11
Update idonthavespotify extension

SeaOfTea:
## Description

Updates the idonthavespotify extension to use "sjdonado.com" instead of "donado.co". This mirrors how the frontend on https://idonthavespotify.sjdonado.com calls the API.
"donado.co" currently returns a 405 status resulting in the `.json()` call throwing an error. There's probably some changes to error handling that could be made here - but I've kept the changes as small as possible for now.

@sjdonado - thanks for the extension! Hopefully this isn't stepping on your toes at all! I'm not aware of the differences in your two websites so if this is the wrong change to make then let me know.

## Checklist

### DIFF
--- a/extensions/idonthavespotify/CHANGELOG.md
+++ b/extensions/idonthavespotify/CHANGELOG.md
@@ -1,5 +1,9 @@
 # I Don't Have Spotify Changelog
 
+## [Fix] - {PR_MERGE_DATE}
+
+- Changed the hostname from which the data was fetched from (donado.co to sjdonado.com)
+
 ## [New Commands] - 2024-07-31
 
 - Convert link from clipboard and copy it back for each platform.
--- a/extensions/idonthavespotify/package-lock.json
+++ b/extensions/idonthavespotify/package-lock.json
@@ -12,8 +12,7 @@
         "@raycast/api": "^1.78.1",
         "@raycast/utils": "^1.16.2",
         "node-fetch": "^3.3.1",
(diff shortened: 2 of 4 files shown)


SeaOfTea:
> * ❌ Invalid addition of `pastContributors` field in package.json - this field is not supported and violates Raycast's policy on author modifications

`pastContributors` is [documented here](https://developers.raycast.com/information/manifest#extension-properties). I wasn't sure which was more appropriate between `pastContributors` and `contributors` for a bugfix.

sjdonado:
@SeaOfTea contributors instead of pastContributors will be also fine IMO, thank you for taking the time :)

now that we have a PR already, could you help me with this quick fix https://github.com/raycast/extensions/issues/19225? we just need to remove all adapters regexes in favor of only checking if the search input is a valid url, that way we keep it flexible and delegate this responsibility to the api

open to suggestions though

pernielsentikaer:
I agree, you should be added to `contributors` if you want to 😊 Could you look into the suggestion from @sjdonado

SeaOfTea:
> @SeaOfTea contributors instead of pastContributors will be also fine IMO, thank you for taking the time :)
> 
> now that we have a PR already, could you help me with this quick fix #19225? we just need to remove all adapters regexes in favor of only checking if the search input is a valid url, that way we keep it flexible and delegate this responsibility to the api
> 
> open to suggestions though

hey! sorry for a delayed reply, but yeah sure, I can take a look

PR #19756, opened 2025-06-12
Update linkding extension

### LINKED ISSUE #19615: [Linkding] Don't find all expected entries
### Extension

https://www.raycast.com/InteractiveNinja/linkding

### Raycast Version

1.99.4

### macOS Version

15.5

### Description

I have successfully connected my own selfhosted Linkding instance with the Linkding extension of Raycast. If I try to search for some entries, I miss the expected results. So I search for espanso and don't find anything inside the extension, but the same search inside the Linkding instance finds the homepage of this mac software. Another example is search for Windsurf - an ai-based developer tool. No results in the extension, but the stored links to the homepage of Windsurf in the Linkding instance.
So - why the search inside the extension does not find this entries?

### Steps To Reproduce

1. install linkding extension and connect to a selfhosted linkding instance
2. Try some searches - get some results
3. Try the search for specific strings - no results inside the extension, but results for the same search string inside the linkding instance

### Current Behaviour

Get no results or miss the expected entries in the result list

### Expected Behaviour

Get the same amount of search results like in Linkding itself.

jose-elias-alvarez:
## Description

This PR bundles a few fixes to the Linkding extension

1. Fix a bug that prevented clearing bookmark metadata when automatically set.
2. Remove all code to get the URL of the active Firefox tab. AppleScript + Firefox is totally unreliable, and the only alternative is to simulate keystrokes, which is awful UX and wipes out the clipboard. I think it's better not to have this at all than to advertise it and have it not work.
3. Fix #19615 by reverting back to server-side filtering, which is slower but is the best way to handle large bookmark collections (and also means that search results in the extension will precisely match what Linkding itself returns). [The API](https://linkding.link/api/) supports pagination, but I don't think it's worth the effort of implementing it here, since I doubt many users will want to scroll through a large list of bookmarks vs. just filtering it by searching.

## Screencast

N/A

## Checklist

### DIFF
--- a/extensions/linkding/CHANGELOG.md
+++ b/extensions/linkding/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Linkding Changelog
 
-## [Fixes] - {PR_MERGE_DATE}
+## [Fixes] - 2025-06-26
 
 - fix!: use server-side filtering to handle large bookmark counts
 - fix!: remove Firefox-specific code (doesn't work)


pernielsentikaer:
Hi @jose-elias-alvarez👋 

@InteractiveNinja do you want to check this?

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #19803, opened 2025-06-14
[zed-recent-projects] Adds a preference checkbox to use the Zed development Sqlite database

remorses:
## Description

Adds a preference checkbox to use the Zed development Sqlite database, so that recent projects work for people that build their own Zed from source, like in my case

### DIFF
--- a/extensions/zed-recent-projects/CHANGELOG.md
+++ b/extensions/zed-recent-projects/CHANGELOG.md
@@ -1,6 +1,6 @@
 #  Zed Recent Projects Changelog
 
-## [Fixes] - {PR_MERGE_DATE}
+## [Fixes] - 2025-07-07
 
 - Add `Zed Dev` app build option
 


pernielsentikaer:
Hi @remorses👋 

ewgenius:
@remorses thanks for the contribution, that is a nice feature!

I'm a bit concerned about the UX. Could you please provide details about your local Zed build process? Which Zed build do you use in the extension settings (stable or preview)? I believe it would be more suitable to add a third build option, `Zed Local` or `Zed Dev`, which would map to the development database and preview bundle id.

remorses:
I just run cargo build release and copy paste the binary inside the Zed Preview app contents. I don’t know if there is a way to make the result binary use the preview sqlite database

ewgenius:
This looks like a very custom way to use Zed, and I doubt there are many other users who do similar.

> I don’t know if there is a way to make the result binary use the preview sqlite database

From zed sources, db name is inferred from current release channel:
https://github.com/ewgenius/zed/blob/69d7ea7b606f2321307623795c152ccffc1d01e5/crates/db/src/db.rs#L141-L151

I would suggest to try building zed using `bundle-mac` script first, or try `RELEASE_CHANNEL=stable cargo build --release` or `RELEASE_CHANNEL=preview cargo build --release` and see if that will allow to use stable/preview db.

https://github.com/ewgenius/zed/blob/69d7ea7b606f2321307623795c152ccffc1d01e5/script/bundle-mac#L68

remorses:
You are right, there was a script in the Zed repo `bundle_mac` that can create a `Zed Build` app, so I added a new build type `Zed Build` instead

extensions/zed-recent-projects/tsconfig.json
```diff
@@ -11,6 +11,7 @@
     "skipLibCheck": true,
     "forceConsistentCasingInFileNames": true,
     "jsx": "react-jsx",
+    "noEmit": true,
```
ewgenius: Why this change needed?

remorses: To make `tsc` not write the .js files inside the src directory, Raycast has its own build system

ewgenius: > Raycast has its own build system

None of the build or dev commands call `tsc`, so js files are not generated.

Let's keep it as is. Extension configs were generated by raycast, and should be updated only to match the default template or if something specific needed.

pernielsentikaer:
Is there any status here?

remorses:
Should be ready to merge, It now shows an additional option for the development bundle of Zed instead of using a checkbox.

ewgenius (approved):

pernielsentikaer (approved):
Hi 👋

Thank you for your contribution 🔥

This looks good to me 🚀

PR #19829, opened 2025-06-15
Update safari extension

ThanhDodeurOdoo:
Fixes for `Search Tabs` command: 
* allows the components that display the tab list to re-render when device data changes.
* allows the obtention of the local tabs even when the `areRemoteTabsUsed` ("iCloud devices" option) is set to `false`.

## Description

Before this commit, the `Search Tabs` command wouldn't display up-to-date information, but the result of the previous search. That was caused by the interface not re-rendering properly when the data changed.

## Screencast

before:
CleanShot 2025-06-15 at 16 51 22

after:
CleanShot 2025-06-15 at 16 48 08

## Checklist

### DIFF
--- a/extensions/safari/CHANGELOG.md
+++ b/extensions/safari/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Safari Changelog
 
-## [Update] - {PR_MERGE_DATE}
+## [Update] - 2025-06-19
 
 - Fixes for `Search Tabs` command:
   - allows the components that display the tab list to re-render when device data changes.


extensions/safari/src/hooks/useDevices.ts
```diff
@@ -84,10 +84,10 @@ export default function useDevices() {
         .reject(["name", deviceName])
         .value();
 
-      devices.current = [localDevice, ...remoteDevices];
+      setDevices([localDevice, ...remoteDevices]);
       permissionView.current = remoteTabs.permissionView || null;
     }
-  }, [localTabs, remoteTabs, deviceName]);
+  }, [localTabs.data, remoteTabs.data, deviceName]);
```
ThanhDodeurOdoo, line 92: this prevents a recursive dependency as `localTabs` and `remoteTabs` are always new objects even if their data does not actually change.

ChangeHow: It seems `useRef` to prevent too much updates but met react-hooks  closure trap?

Does `TAB.data` alway exist?

ChangeHow (approved):
LGTM

extensions/safari/src/hooks/useDevices.ts
```diff
@@ -84,10 +84,10 @@ export default function useDevices() {
         .reject(["name", deviceName])
         .value();
 
-      devices.current = [localDevice, ...remoteDevices];
+      setDevices([localDevice, ...remoteDevices]);
       permissionView.current = remoteTabs.permissionView || null;
```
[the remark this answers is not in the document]
ThanhDodeurOdoo: Maybe out of the fix's scope, but it makes sense, not sure if I should add that to this PR or if it should be a separate fix. The local device tabs are indeed not displayed when the "icloud" option is checked, and it probably should.

extensions/safari/src/hooks/useDevices.ts
ThanhDodeurOdoo: data seems indeed optional (there could be `TAB.error`instead) but I think that `localTabs` and `remoteTabs` seem to always defined at this point since they aren't guarded above,

I suspect that an eventual transition from `undefined` to `data` is fine (will just re-render properly).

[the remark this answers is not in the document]
ThanhDodeurOdoo: `localDevice` already depends on `localTabs`

pernielsentikaer (approved):

PR #19849, opened 2025-06-17
Add regex-batch-renamer extension

tarikkavaz:
## Checklist


pernielsentikaer:
Hi 👋

Thanks for your contributon 🔥

Instead of making a new extension, then it could make sense to make a new command in [this extension](https://www.raycast.com/theo_daguier/renaming) by @theodaguier instead, would you be up or that?

tarikkavaz:
> Hi 👋
> 
> Thanks for your contributon 🔥
> 
> Instead of making a new extension, then it could make sense to make a new command in [this extension](https://www.raycast.com/theo_daguier/renaming) by @theodaguier instead, would you be up or that?

Thank you for the suggestion. I haven’t inspected @theodaguier’s extension, so I’m not sure if they overlap. If you’re convinced they serve the same purpose, I’m happy to close this one and not publish it to the Raycast Store. Let me know what you think.

pernielsentikaer:
I think a new command would be nice, I'm not sure the regex part can fit into the current two commands 😊

tarikkavaz:
> I think a new command would be nice, I'm not sure the regex part can fit into the current two commands 😊

After giving it some thought, I believe this extension could be a valuable asset and tool. I’d be happy to see it in the store if it’s not inconvenient. I’ll leave the decision up to you, Per :)

PR #19936, opened 2025-06-20
[Bitwarden] Remove sensitive `captureException` usages

LitoMore:
## Description

This extension abuses a lot of `captureException`s to collect errors (even some errors can be ignored gracefully). And most of them are command executions. Users' sensitive data may be compromised in this way. I've removed all `captureException` usages from this extension because it is a password-related app.

I'm unable to view the collected errors in the developer hub. But @pernielsentikaer, you may have access to check if this extension has already collected any sensitive data from users.

For some ESLint and TypeScript errors, I just ignored them to keep the original code.

### Changes

- Remove all sensitive `captureExceptions` usages
- Bump extension utilities to the latest

## Checklist


pernielsentikaer:
Thanks for the PR @LitoMore - If I'm not mistaken, then we added this due to a lot of users failed to use the extension and we needed to figure out why

@jomifepe could you check this?

LitoMore:
> If I'm not mistaken, then we added this due to a lot of users failed to use the extension and we needed to figure out why

We can use the `showFailureToast` to show the raw error message to the user. So that users can copy and report the problem with a censored snapshot.

jomifepe:
Hey, I can see why this could seem like a concern @LitoMore, but you can see that the `captureException` function is under the development utils, and has a `captureToRaycast` flag set to `false` by default. So this function was mostly used to log the exceptions to the console and accumulate them so they can be copied by the `CopyRuntimeErrorLog` action. That said, there was never an instance of a password being explicitly captured to the issues dashboard.

The only instance of a catched exception being sent to raycast is this one, which you can see has the `captureException` flag set to `true`:

Screenshot 2025-06-20 at 10 55 05

If there were instances of sensitive values being sent to the issues dashboard because of crashes, not because of explicit logging, Raycast usually does a good job ensuring those values are redacted:

Screenshot 2025-06-20 at 11 36 24

---

> we added this due to a lot of users failed to use the extension and we needed to figure out why

The function itself has been there for a long time, but yes @pernielsentikaer  the `captureToRaycast` flag was added [when the extension transitioned to downloading the CLI binary](https://github.com/raycast/extensions/commit/a1aa4a32dc5fe4fdd334031303759294e0f3e595), which cause a surge of errors the first time.

I would say that, the only security change that could be made, is ensuring that nothing that could log a sensitive value causes the extension to crash, because crashes are captured automatically to the issues dashboard. Although, I don't think there currently any, as I go to the issues dashboard from time to time and I don't see any.

LitoMore:
@jomifepe Thanks for your explanation. I just checked the source code of [`execa@7.0.0`'s error instance](https://github.com/sindresorhus/execa/blob/6bb4a6d3b00e464b8ea36382e94329ecf4e21ee2/lib/error.js). It looks like the running environment variable will not be compromised within the error instance.

I'm closing this PR since the current implementation appears to be satisfactory.

> If there were instances of sensitive values being sent to the issues dashboard because of crashes, not because of explicit logging, Raycast usually does a good job ensuring those values are redacted.

I know this, but I don't think they will censor the password/token/credential from a random string.

jomifepe:
> @jomifepe Thanks for your explanation. I just checked the source code of [`execa@7.0.0`'s error instance](https://github.com/sindresorhus/execa/blob/6bb4a6d3b00e464b8ea36382e94329ecf4e21ee2/lib/error.js). It looks like the running environment variable will not be compromised within the error instance.
> 
> I'm closing this PR since the current implementation appears to be satisfactory.
> 
> > If there were instances of sensitive values being sent to the issues dashboard because of crashes, not because of explicit logging, Raycast usually does a good job ensuring those values are redacted.
> 
> I know this, but I don't think they will censor the password/token/credential from a random string.

@LitoMore Thank for the research, I appreciate you attention to the security of the extension 🙌

Don't hesitate to open a PR for something else!

PR #19943, opened 2025-06-20
[Google Translate] Fix incorrect language mapping logic

### LINKED ISSUE #19942: [Google Translate] Chinese not working
### Extension

https://www.raycast.com/gebeto/translate

### Raycast Version

1.100.2

### macOS Version

_No response_

### Description

Chinese stopped working recently:
Image

Image

### Steps To Reproduce

1. Try to translate from or to Chinese

### Current Behaviour

_No response_

### Expected Behaviour

_No response_

LitoMore:
## Description

Fixes #19942

- Fix incorrect language code mapping logic
- Move vendor folder to the project src directory
	- The Raycast's CLI cannot watch those source files outside the `src` directory. I've moved it to the right place so that we don't have to terminate and restart the dev process while development.

## Checklist


channingkuo (approved):

gebeto:
What is the fix? There was an issue related to code mapping, and keys was lowercasing which was causing check like
`ch-ZH === ch-zh is false`

But I don't see the fix anywhere in the PR's code

I've also missed this PR and made my fix for it: https://github.com/raycast/extensions/pull/19957

LitoMore:
Here is the error on my device:



I've checked the code, and the `getISOCode` uses the wrong value for comparison.

extensions/google-translate/src/vendor/@iamtraction-translate/src/languages.ts
```diff
@@ -0,0 +1,292 @@
...
+  if (language in languages) return language;
+
+  const keys = (Object.keys(languages) as (keyof typeof languages)[]).filter((key) => {
+    if (typeof languages[key] !== "string") return false;
+
+    return key.toLowerCase() === language;
+  });
+
+  return keys[0] || undefined;
+}
```
gebeto, line 283: Exactly, beause it is calling `toLowerCase` for `language`, and then checking if `language in languages`, so it is checking "if `ch-zh` in `{"ch-ZH": "Chinese"}`"

gebeto: so all we need to do is to lowercase all the keys there

LitoMore: Here is your code:

https://github.com/raycast/extensions/blob/f89de8d1eb6930d3c091efd7fa10696a2920e78f/extensions/google-translate/vendor/%40iamtraction-translate/src/languages.ts#L275-L278

The input parameter `language` of the `getISOCode` is a language code. The `key` is the language code, and the `languages[key]` is the language name. But you used it to compare with the `language` (code).

gebeto: all I did is made all the keys lowercase to fix the issue. In our extension case, this method is doing early return here, as we are using language iso codes:

https://github.com/raycast/extensions/blob/f89de8d1eb6930d3c091efd7fa10696a2920e78f/extensions/google-translate/vendor/%40iamtraction-translate/src/languages.ts#L273

I've also added types for that code, but overall it is a code of that library author

yes, this place is using language name, but there are early return above this part of code which passing most of the time. As I see, this method can receive language `name`, or language `key`, you can put for example `english` as param to this method, and it will return `"en"`, and if you will put `en`, it will return `"en"`

LitoMore:
Closing in favor of #19991.

I will create another PR for those improvements in this PR.

PR #20100, opened 2025-06-27
Add brightness-pilot extension

giovacalle:
## Checklist


LitoMore:
Can you check the existing extension https://www.raycast.com/cali/brightness-control?

It looks like we already have an extension for adjusting brightness.

giovacalle:
I checked the extension you mentioned, but its implementation differs. The one in the Store uses an AppleScript to “press” the brightness keys, whereas my proposed version uses a Swift-based CLI to set brightness levels directly—supporting both built-in (MacBook) and external displays.

I wasn’t sure whether we should merge these into a single extension—especially given the guidance under “When you should consider creating a new extension instead of contributing to an existing one” in the Spotify Controls example [(prepare-an-extension-for-store).](https://developers.raycast.com/basics/prepare-an-extension-for-store#contributing-to-existing-extensions-vs-creating-a-new-one)

This is my first Raycast extension, so I’m happy to hear any advice or best practices you might have :)

pernielsentikaer:
It could indeed make sense to update the current extension in the store by @CaliCastle if this is a faster and better way - they do have the same purpose so it doesn't make sense to have two of the same in the store

andreaselia:
Hey @giovacalle 👋 

This is a great implementation for brightness control!

Since there's already an existing extension that targets this functionality, I agree with @pernielsentikaer and @LitoMore in it being worth contributing to the existing one but improving the extension with your CLI setup instead.

With this in mind, I'm going to close this PR, but please do reach out if you have any questions or would like to discuss this.

giovacalle:
Hey @andreaselia, thanks for the feedback on the brightness control — really appreciate it! 🙏

I totally understand the decision, even if I’m a bit sorry to see the PR closed. If the author of the existing extension ever shows interest, we’d be happy to explore merging the two efforts and bringing in the CLI setup as an enhancement.

Thanks again everyone, have a great day! ✌️

PR #20126, opened 2025-06-29
Update `Pocket` extension - add an Alert in Show informing users to Export + required url in create

xmok:
## Description

I got an email that **Pocket** is shutting down :(.
I have added an Alert (which can be dismissed) so users can know about it and directly go to the Export link.

pocket-shutting-down

- EmptyView when no bookmarks
- URL in create is now Required
- Add an Alert to let users know Pocket is shutting down

## Screenshot

pocket-alert

## Checklist

### DIFF
--- a/extensions/pocket/CHANGELOG.md
+++ b/extensions/pocket/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Pocket Changelog
 
-## [Add Alert to Go to Export] - {PR_MERGE_DATE}
+## [Add Alert to Go to Export] - 2025-06-30
 - EmptyView when no bookmarks
 - URL in create is now Required
 - Add an Alert to let users know Pocket is shutting down


vimtor:
This is so nice of you to do. I'm not sure, though, if the right move would be to remove/archive the extension. I wouldn't want to lose those juicy contributor statistics, but such is life.

pernielsentikaer:
I'm not sure there is much else to do since if we don't remove it it has no value 🫠

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

xmok:
The best we could do is after July 8, replace the extension with an `EmptyVIew` that takes you to the "Export" page since Export will be available till October 8 (based on the email). After that, removal/deprecation it is.

PR #20203, opened 2025-07-03
Add pihole-admin extension

kennysamuerto:
## Description

# Pi-hole Admin - Raycast Extension

A complete Raycast extension to manage **Pi-hole v6** directly from Raycast. This extension has been completely rewritten to be compatible with Pi-hole v6's new REST API.

## ✨ Features

- 📊 **Status Dashboard**: View real-time Pi-hole statistics
- ⚡ **Blocking Control**: Enable/disable Pi-hole with duration options
- 📝 **Query Log**: Browse DNS queries with advanced filters
- 📈 **Top Domains**: See the most queried and blocked domains
- ➕ **Domain Management**: Add domains to the allow and block lists
- 🗑️ **Clear Logs**: Delete all query logs
- 🔒 **Secure Authentication**: Compatible with Pi-hole v6's new session system

## Screencast

Raycast 2025-07-04 at 17 53 03

Raycast 2025-07-04 at 17 52 38

Raycast 2025-07-04 at 17 53 23

## Checklist


pernielsentikaer:
Hi @kennysamuerto👋 

We already have a [pi-hole](https://www.raycast.com/GLaDO8/pie-for-pi-hole) extension by @GLaDO8  in the store, did you try that out? It might be better to enhance than creating a new extension

kennysamuerto:
Hello @pernielsentikaer,

I know, but there are many people who haven't updated to Pihole version 6 and may need that extension, and many others who have updated to version 6 and cannot use the extension from the store.

I think it could be interesting for people to be able to choose based on the version of Pihole they have, since the API changes completely from one version to another.

I hope you find this okay!

Thank you

pernielsentikaer:
Could it be something like how it's done in 1Password, where you can select between version 7 and 8? Could that be an option?

kennysamuerto:
Hi @pernielsentikaer 

Indeed, it is like a new application when talking about the API, as it has completely changed the way of interacting with the API. For this reason, many people prefer to stay on the version prior to 6, while many others are updating to this new version, leaving many people stuck with version 6 unable to use the Raycast extension. Conversely, if we update the existing version, many people who do not use version 6 will find the extension stops working for them.

Thanks!

pernielsentikaer:
Would it be too much work to make it possible to switch in the app like in 1password?

CleanShot 2025-07-09 at 08 48 35@2x

kennysamuerto:
Hi @pernielsentikaer,

It's complicated due to how the code is structured. I am currently using the extension for V6 set up locally, but I wanted to contribute because I see that more people are requesting this update, and I really find it quite difficult to understand the previous code. For my extension, whose code I do know, it might be possible (though I wouldn't have a way to test the previous version of the API anymore), but for the extension that is already in the store, as I mentioned, I cannot understand some things.

pernielsentikaer (changes requested):
I think we need to make sure to mention that it's for 6+ and up both in the name/readme/other files so its easier to find the right version?

treyg:
@kennysamuerto Nice work on this! A few things what would be nice to have are a quick command to open the web dashboard, restart the DNS (if you can figure out a good approach), and a timer for the disable command. I started an extension not realizing you already had one in progress if you want to see [an example](https://github.com/raycast/extensions/pull/20365) of setting a timer for the disable command.

treyg:
@pernielsentikaer @kennysamuerto Any updates on this? does it need anything still?

pernielsentikaer:
@kennysamuerto are you still around?

PR #20232, opened 2025-07-04
Add rounding-number extension

HackettLai:
## Description

Quickly round any number to the nearest value you choose. Simply enter a number and the place you want to round to (such as 10, 100, or 1000), and get your rounded result instantly.

## Screencast

  
CleanShot 2025-07-04 at 12 11 37
CleanShot 2025-07-04 at 12 12 22
CleanShot 2025-07-04 at 12 12 27


pernielsentikaer:
Hi @HackettLai👋 

Feels a bit like [this one](https://www.raycast.com/pkliem/8-divide) - I guess it's a bit hard to enhance that one but I would maybe just show a list with the values instead of a form

HackettLai:
> Hi @HackettLai👋
> 
> 
> Feels a bit like [this one](https://www.raycast.com/pkliem/8-divide) - I guess it's a bit hard to enhance that one but I would maybe just show a list with the values instead of a form

Oh, I’ve come across “8-divide” too, but it only finds the nearest number divisible by 8. I wanted something more flexible, so I learned how to make my first extension. In my UI design process, everything needs to be divisible by 4, and for sometimes I also need to set the parent size based on what’s divisible by its child element. 

With this extension, you can freely change the divisor, and it’ll automatically copy the result to your clipboard so you can paste it right into whatever application you’re using.

HackettLai:
> Hi @HackettLai👋
> 
> 
> Feels a bit like [this one](https://www.raycast.com/pkliem/8-divide) - I guess it's a bit hard to enhance that one but I would maybe just show a list with the values instead of a form

Oh, and about the form, I chose it because I wanted users to have a space to adjust or manipulate the values after the first run. The initial result is already copied to the clipboard, so users can just hit Esc to close Raycast, or they can stay in the window and keep calculating if they need to.

pernielsentikaer:
I don't think the default option for entering the command should be to copy it to the clipboard; maybe you could make that a preference?

HackettLai:
> I don't think the default option for entering the command should be to copy it to the clipboard; maybe you could make that a preference?

Umm I may need sometimes to learn how to do that 😎

HackettLai:
> I don't think the default option for entering the command should be to copy it to the clipboard; maybe you could make that a preference?

Hi @pernielsentikaer I have added the auto-copy result to the clipboard as a preference. Please let me know if this is acceptable.😎 Thx! 
CleanShot 2025-07-09 at 15 33 32

HackettLai:
> * Could you add at least one [screenshot with our tool](https://developers.raycast.com/basics/prepare-an-extension-for-store#screenshots) of the extension so they get the right padding
> 
> Maybe even add a result field, something like those two ideas: rounding-number-3

Hi @pernielsentikaer I have updated the screenshots and revised the script by displaying the results under the form.

pernielsentikaer:
Would this be cool? I guess the automatic part should only copy the result and then it the action should be copy instead :) 

https://github.com/user-attachments/assets/825e7f8b-2353-4fba-b431-2793c3c317f6

HackettLai:
> Would this be cool? I guess the automatic part should only copy the result and then it the action should be copy instead :)
> 
>  CleanShot.2025-07-09.at.12.01.48.mp4

Nice suggestion, @pernielsentikaer  . I've already implemented it. Now, it will auto-recalculate after every form change. The clipboard behavior will be followed  according to the preference settings.

The current version is highly satisfactory for my initial design needs. For your information, here are some screen recordings.


https://github.com/user-attachments/assets/7dbffc43-fd63-46bb-b782-0f05f83646d3

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #20395, opened 2025-07-12
Add commit-issue-parser extension

nekeo:
## Description

This new extension parses your issue url (from Github, Gitlab or Jira) to generate selectable commit names with the parsed url, displaying only the issue id as a scope.

The extension is also compatible with the Git CLI, and Lazygit, to automatically provide a commit body with the full url entered.

I had already made a pr #20334 (only verified by the bot) but I had to close it due to a fork problem that was no longer synchronized with my repo but it won't happen again 👍 

## Screencast

https://github.com/user-attachments/assets/525205ed-d63c-4e77-bd93-97b981263695

## Checklist


extensions/commit-issue-parser/src/models/preferences.ts
```diff
@@ -0,0 +1,9 @@
+import { ActionType } from "./actionType";
+import { ContentFormat } from "./contentFormat";
+import { TypeMode } from "./typeMode";
+
+export type Preferences = {
```
andreaselia: Not needed as Raycast automatically generates this type for you.

andreaselia:
Hey @JulesJuul 👋 

Cool extension. Left a small comment.

nekeo:
Hey! Thanks for your feedback!
I replied to your comment thinking it might be relevant to leave this typing.

I also take this occasion to add this minor fix

extensions/commit-issue-parser/src/models/preferences.ts
nekeo: I implemented it because it was recommended in the example in the `getPreferenceValues` doc:
```typescript
/**
 * A function to access the preference values that have been passed to the command.
 *
 * @returns An object with the preference names as property key and the typed value as property value.
 *
 * @remarks
 * Each preference name is mapped to its value and the defined default values are used as fallback values.
 *
 * @example
 * ```typescript
 * import { getPreferenceValues } from "@raycast/api";
 *
 * interface Preferences {
 *   name: string;
 *   bodyWeight?: string;
 *   bodyHeight?: string;
 * }
 *
 * export default async () => {
 *   const preferences = getPreferenceValues<Preferences>();
 *   console.log(preferences);
 * };
 * ```
 */
```

This allows me to have the exact type for each of my options via my variable `const preferences = getPreferenceValues<Preferences>();`, such as for `preferences.contentFormat`, where I get my `ContentFormat` type instead of `any`.

I can remove it without any problem because `any` type won't be a problem, but I think it's a good idea to keep it in so that I can have a more precise type than the default one.

What do you think?

andreaselia: The type shouldn't be `any` because the `Preferences` are generated via the `raycast-env.d.ts` file 🤔 

Example from one of my own extensions:

CleanShot 2025-08-05 at 09 13 17@2x

andreaselia: Bit more info on this can be found here: https://developers.raycast.com/api-reference/preferences#preferences

nekeo: My bad, the problem occurred because I hadn't used the Preferences of the `raycast-env.d.ts` file in my use of `getPreferenceValues`, which consequently displayed `any` in my IDE

So I've modified the code to use this type, and removed dependent types such as `ContentFormat`, `ActionType` and `TypeMode`

Thanks!

andreaselia: All good, not super clear in the docs, so room for improvement there since it's text underneath the example.

Thanks for actioning the feedback so quickly!

PR #20405, opened 2025-07-12
Update `Oracle Cloud (OCI)` extension - implement provider w/ context + basic objectstorage command

xmok:
## Description

This is mostly aimed at moving to the Provider w/ Context so the OCI Config check can be shared. This is a crucial step in extending the extension to manage more services.

- add initial "Object Storage" command to view buckets
- `dev`: implement Provider w/ Context so checks can be shared

## Screencast

https://github.com/user-attachments/assets/204d34e9-4f7c-4529-a5f6-6c5b3c347298

## Checklist

### DIFF
--- a/extensions/oci/CHANGELOG.md
+++ b/extensions/oci/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Oracle Cloud Changelog
 
-## [View Object Storage Buckets + Implement Provider w/ Context] - {PR_MERGE_DATE}
+## [View Object Storage Buckets + Implement Provider w/ Context] - 2025-07-14
 
 - add initial "Object Storage" command to view buckets
 - `dev`: implement Provider w/ Context so checks can be shared


extensions/oci/eslint.config.js
```diff
@@ -3,4 +3,9 @@ const raycastConfig = require("@raycast/eslint-config");
 
 module.exports = defineConfig([
   ...raycastConfig,
+  {
+    rules: {
+      "@raycast/prefer-title-case": "off"
```
andreaselia, line 8: Hey @xmok, how come we need this? Am I missing a line change here that requires it? 😅

xmok: Hi Andreas:

Yup this was previously line 1 of `Core.tsx` file. We need this as there are many titles especially in Actions which are otherwise changed by the linter e.g.

- "Open in OCI" -> "Open in Oci"
- "VNIC" -> "Vnic"

I moved it to the global eslint otherwise would need to add it to every component page separately.

Hope this makes sense.

andreaselia: Yep, that makes sense, thank you!

andreaselia (approved):
LGTM, nice work as always @xmok 🔥

PR #20454, opened 2025-07-15
Update toggle-fn extension

CaramelFur:
## Description

This changes this extension to use a native objective c binary to change between the two states, because the applescript implementation was awfully slow.
The source for this binary can be found here: https://github.com/CaramelFur/fntoggle
It's compatible with MacOS 11 and up, both ARM64 and X86.
This extension has the SHA1 sum hardcoded to ensure the binary can only be updated with a new pull request.

## Screencast

Not required

## Checklist

### DIFF
--- a/extensions/toggle-fn/CHANGELOG.md
+++ b/extensions/toggle-fn/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 All notable changes to this project will be documented in this file.
 
-## [2.0.0] - 2025-01-27
+## [2.0.0] - {PR_MERGE_DATE}
 
 ### Changed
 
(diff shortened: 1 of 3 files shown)


CaramelFur:
Lmao that Greptile ai really doesn't like external binaries does it.
Sha1 should be plenty fine to ensure the binary can't be modified.

elonwoo:
This is like code poisoning, please be careful about your behavior, it is very unsafe.

CaramelFur:
@elonwoo Do you have any suggestions on how to do it better and make it less unsafe? I couldn't think of a better way. Personally I much prefer the instant functionality of the little Objective C program to having to wait for the apple script to finish. But I can see how it seems a little riskier.

If you would prefer I submit these changes as a new package instead, I can do that.

And if it ends up not being fit for the public extension repository thats fine too, and the request can be closed. I'm already satisfied with it running on my machine only.

pernielsentikaer:
What do you think, @elonwoo - since we have the binary source and it's a fixed release with a checksum, it might be okay

PR #20458, opened 2025-07-15
[zoxide] fix compatibility with Intel Macs, clean up $PATH

deviant:
On Intel Macs, Homebrew places commands in /usr/local/bin instead of the newer /opt/homebrew/bin. As well as adding this directory (where appropriate) to the $PATH, this also cleans it up and reorders it for efficiency. The following changes have been made:

- Remove /sbin and variants: these are for administrative commands only, which zoxide and fzf are not.
- Order by likelihood: zoxide and fzf are more likely to be installed via Homebrew than manually, so it makes sense to check in Homebrew's directories first. Additionally, Apple Silicon before Intel, and /usr/bin before /bin.
- Only look in /usr/bin for mdfind(1). This is provided by Apple, not installed by the user, so we can be sure it'll only be in that one location. It is in this location on both Apple Silicon and Intel Macs.

### DIFF
--- a/extensions/raycast-zoxide/CHANGELOG.md
+++ b/extensions/raycast-zoxide/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Raycast Zoxide Changelog
 
-## [Added support for Intel Macs] - {PR_MERGE_DATE}
+## [Added support for Intel Macs] - 2025-08-05
 
 - Fixed compatibility with `zoxide` and `fzf` installed via Homebrew on Intel Macs
 - Made some small optimizations to program search paths that shouldn't affect anything


mrpunkin:
These seem like smart changes to me, and seems to work just fine on my Apple Silicon Mac still. 

@deviant I believe you'll need to update the CHANGELOG and manifest per this article: [Contribute to an Extension](https://developers.raycast.com/basics/contribute-to-an-extension)

deviant:
@mrpunkin Thanks, and good catch! I've added an entry to the changelog now. Looking at the manifest documentation, it has the following to say about the `contributors` field:

> An array of Raycast store handles (usernames) of people who have meaningfully contributed and are maintaining to this extension.

Given how small this change is, I don't think I can say that I've meaningfully contributed yet. I also don't have a Raycast store handle at the moment. If I were to add or change something more substantially in the future, then I think that it would make more sense for me to do this, but I think I'll leave this as is for the time being. Hope this sounds reasonable!

Please let me know if you want me to squash these changes into a single commit.

mrpunkin:
@deviant Makes sense to me. Regardless, thanks for the improvements! Looks great to me.

andreaselia (approved):
LGTM, thanks both!

PR #20498, opened 2025-07-17
Add `.gitattributes` & get rid of `\r\n`

LitoMore:
Mixing `\n`\ and `\r\n` may pollute the git blame and bring some code conflicts. It's time to use `.gitattributes` to avoid mixing `\n` and `\r\n` since we started supporting Windows.

But we still need a solution for avoiding mixing `\n` and `\r\n` for Ray CLI cloned extensions. Maybe we can add a check process to the `ray lint`.

### DIFF
--- a/extensions/lega-serie-a/CHANGELOG.md
+++ b/extensions/lega-serie-a/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Lega Serie A Changelog
 
-## [Maintenance] - [PR_MERGE_DATE]
+## [Maintenance] - {PR_MERGE_DATE}
 
 - Cleanup unused gitignore records
 


extensions/lega-serie-a/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Lega Serie A Changelog
 
+## [Maintenance] - [PR_MERGE_DATE]
```
andreaselia:
```suggestion
## [Maintenance] - {PR_MERGE_DATE}
```

pernielsentikaer:
We'll have a look in the middle of next week when we have more folks back from vacation 😊

sxn (approved):
thanks for this! 🙏

PR #20527, opened 2025-07-18
Update youversion-suggest extension

caleb531:
## Description

- Updates the YouVersion to use a more modern core and remove dependencies
- Improves the store listing with fresh screenshots and updated copy

## Screencast

See #2951 

## Checklist


extensions/youversion-suggest/eslint.config.mjs
```diff
@@ -0,0 +1,41 @@
...
+      "package",
+      ".env",
+      ".env.*",
+      "!.env.example",
+      "pnpm-lock.yaml",
+      "package-lock.json",
+      "yarn.lock",
+    ],
+  },
+];
```
xmok: Unless there is a reason you require some of these specific rules, we should use the default config. You may need to make sure all your deps are ok.
```suggestion
import { defineConfig } from "eslint/config";
import raycastConfig from "@raycast/eslint-config";

export default defineConfig([...raycastConfig]);
```

caleb531: @xmok Ooh great suggestion! Fixed in 3694fbe4bd1cfaee94e9df4df5ee4f8ebadfa89a.

caleb531:
@xmok

> This is **not** required, but if you are so inclined you could simplify your API call with `useCachedPromise` like in "PlanningCenter" PR but it is _not_ a blocker.

If it's alright, I will come back to the `useCachedPromise` refactor at a later time. It's a great idea, but I need to do some extensive testing first, and in the meantime, would like to get my extension updates out there.

andreaselia:
LGTM @caleb531, thanks @xmok too for your help 😁

PR #20590, opened 2025-07-22
Add pastable-clipboard-history extension

devenv:
## Checklist

### DIFF
--- a/extensions/pastable-clipboard-history/CHANGELOG.md
+++ b/extensions/pastable-clipboard-history/CHANGELOG.md
@@ -1,0 +1,43 @@
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [1.0.0] - 2025-01-22
+
+### Added
+
+- 6 instant paste commands for direct clipboard access
+- `paste-current` - Paste current clipboard item (position 0)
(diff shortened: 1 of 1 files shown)


andreaselia:
Hey @dotenv 👋 

Thanks for taking the time to submit you extension! 🔥 

I'd probably recommend creating snippets for this use case, as you can use [clipboard offset](https://manual.raycast.com/dynamic-placeholders) to then be able to type your snippet and have the clipboard history you're after (e.g. type "ch1" and get the first offset).

Just curious, how would you know which clipboard history item you're pasting if you can't see it, and that it's the one you want to paste?

This PR is also full of AI generated files which make no sense in an extension, the eslint file isn't the recommended default Raycast one, the test suite is unnecessary here given how simple the extension is, and there markdown in a metadata directory which is typically used for screenshots.

With all that said, I'm going to close this PR. I'm happy to discuss the need for such an extension if you feel there's a valid use case for it that isn't super niche though.

devenv:
I mostly use it to paste the previous clipboard, so you can copy "one", copy "two", then press "previous clipboard" shortcut, then cmd+v to get "onetwo".
Same for longer lists, you can copy up to 8 items and then cmd+v, 1st-cb-key, 2nd-cb-key, etc to retrieve the list in one place.
Snippets solution is really slow as cb is by nature temporary, I don't want to save an item in snippets just to retrieve it once, or retrieve them one-by-one from the cb history.
The solution in Linux is two have two cbs, and Raycast doesn't provide an easy alternative.
IMO this ability would be useful as built-in clipboard history itself.

PR #20691, opened 2025-07-29
Add ipscan extension

lfazioli:
## Checklist

### DIFF
--- a/extensions/ipscan/src/IPScan.tsx
+++ b/extensions/ipscan/src/IPScan.tsx
@@ -35,7 +35,7 @@ export default function PortScanner() {
   async function handleSubmit(values: { ip: string }) {
     setLoading(true);
     setOpenPorts(null);
-    await showToast({ title: "Scansione in corso...", style: Toast.Style.Animated });
+    await showToast({ title: "Scanning in progress...", style: Toast.Style.Animated });
 
     const portsToScan = Array.from({ length: 1024 }, (_, i) => i + 1);
     const results = await Promise.all(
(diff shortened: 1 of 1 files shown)


extensions/ipscan/package.json
```diff
@@ -0,0 +1,39 @@
+{
+  "$schema": "https://www.raycast.com/schemas/extension.json",
+  "name": "ipscan",
+  "title": "IPScan",
+  "description": "Scan IP addresses and ports directly from Raycast",
+  "icon": "extension-icon.png",
```
LitoMore, line 6: You just referred to the wrong image. Here means reading image assets from the `assets` folder, which is the Raycast extension placeholder icon file. You may need to move your `extension-icon.png` into `asset` to fix it.

extensions/ipscan/extension-diagnostics.txt
LitoMore, line 1: Is this file needed?

extensions/ipscan/src/IPScan.tsx
```diff
@@ -0,0 +1,95 @@
...
+  const [loading, setLoading] = useState(false);
+  const [ip, setIp] = useState("");
+  const [error, setError] = useState<string | undefined>();
+
+  async function handleSubmit(values: { ip: string }) {
+    setLoading(true);
+    setOpenPorts(null);
+    await showToast({ title: "Scanning in progress...", style: Toast.Style.Animated });
+
+    const portsToScan = Array.from({ length: 1024 }, (_, i) => i + 1);
```
LitoMore, line 40: Can you explain the reason why you limit the max port to 1024?

LitoMore (changes requested):

extensions/ipscan/src/IPScan.tsx
lfazioli: limited the scan to the first 1024 ports because these are the well-known ports, which are most commonly used by network services. Scanning all 65535 ports would be much slower and is rarely necessary for most use cases.

LitoMore: There are so many popular ports larger than 1024, like 3306, 6379, 3389, etc.

I think you can provide an input form to allow the user to specify the port range to scan.

PR #20780, opened 2025-08-04
update brew: Close Raycast in case the brew cmd typed in raycast window

ViGeng:
## Description

Previously, if we choose the action "Run XXX in Terminal," it will launch a terminal session to run the brew command. Furthermore, when I have a long-running command in the terminal, the brew command will break into the running command session where they are not really context-related. Besides that, I have to wait for the first to finish if I want to run the second, third, etc. Using a brand new session can help with multiple runs.

This PR adds a new configuration option to determine whether to use new terminal sessions. Now users can perform multiple actions simultaneously without waiting for the previous one to finish.

It's implemented by adding a preference that serves as a constant changing AppleScript behavior. Details can be seen in the code.

locally tested on 'Terminal' app.

## Final summary:

Considering the proposed feature can not work consistently among all terminal apps, we will stop continuing this feature.
Instead, a windows mis-typing bug has been fixed in this PR.

## Checklist

### DIFF
--- a/extensions/brew/CHANGELOG.md
+++ b/extensions/brew/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Brew Changelog
 
+## [Bug Fix] - {PR_MERGE_DATE}
+
+- Fixed an issue where the the terminal command may be mis-typed in the Raycast window
+
 ## [Improvements] - 2025-02-17
 
 - Add a new action to open the Brew formula/cask page in the browser


extensions/brew/src/components/runInTerminal.ts
```diff
@@ -37,27 +39,41 @@ const appBundleIds: { [key in TerminalApp]: string } = {
...
+};
 
 const appleScripts: { [key in TerminalApp]: (c: string) => string } = {
   alacritty: (c: string) => runCommandInTermAppleScript(c, names.alacritty),
   ghostty: (c: string) => runCommandInTermAppleScript(c, names.ghostty),
   hyper: (c: string) => runCommandInTermAppleScript(c, names.hyper),
   iterm: (c: string) => `
     tell application "iTerm"
-      set newWindow to create window with default profile command "bash -c '${c}; read -n 1 -s -r -p \\"Press any key to exit - will not quit\\" ; echo' ; exit"
+      create window with default profile command "bash -c '${c}; read -n 1 -s -r -p \\"Press any key to exit - will not quit\\" ; echo' ; exit"
```
nhojb: iTerm has a `create tab with default profile` command which should probably be used when `useNewTerminalSession` is `false`.

extensions/brew/src/components/runInTerminal.ts
```diff
@@ -37,27 +39,41 @@ const appBundleIds: { [key in TerminalApp]: string } = {
...
 
-const runCommandInTermAppleScript = (c: string, terminalApp: string): string => `
-    tell application "${terminalApp}" to activate
+const runCommandInTermAppleScript = (c: string, terminalApp: string): string => {
+  const shortcut = useNewTerminalSession ? "n" : "t"; // cmd+n for new window, cmd+t for new tab
+  return `
+    tell application "${terminalApp}"
+        activate
+        delay 0.2
+    end tell
```
nhojb: This only works for me the first time I run a command in a terminal - regardless of the `useNewTerminalSession` setting.

The second time I try to run a command the system beeps and the command is pasted into the Raycast search field - the terminal is never created or activated.

extensions/brew/src/components/runInTerminal.ts
```diff
@@ -37,27 +39,41 @@ const appBundleIds: { [key in TerminalApp]: string } = {
...
-  terminal: (c: string) => `
+  terminal: (c: string) =>
+    useNewTerminalSession
+      ? `
+    tell application "Terminal"
+      do shell script "open -a 'Terminal'"
+      set newWindow to do script "echo ; ${c} ; bash -c 'read -n 1 -s -r -p \\"Press any key to exit - will not quit\\"' ; exit"
+    end tell
+  `
+      : `
```
nhojb: This works, but the existing script (L77-79) does not create a new tab. It looks like working with Terminal tabs via AppleScript is not well supported:

https://www.macscripter.net/t/how-to-create-a-new-tab-in-terminal-app-without-using-systemevents/72627

Note: My Terminal uses the default settings. I also have the system setting "Prefer tabs when opening new windows" set to the default of "In Full Screen". The latter affects whether Terminal creates a new tab or window when running `do script ...`

The other problem with L77-79 is that it doesn't work if Terminal uses the default setting for `When the shell exists`, which is "Don't close the window". If the frontmost window's shell has completed, but not closed, then the applescript does nothing. I think this is an existing issue, but wanted to mention it.

Image

nhojb (changes requested):
Thanks for the PR. In theory it's a good idea and a nice improvement.

Unfortunately I've found the implementation to be unreliable and doesn't work correctly for all terminal apps, including the built-in Terminal. If we are going to add this setting then it should work reliably in all supported terminals. I've added more detailed comments inline.

ViGeng:
Hello, thanks for your very detailed review.. Let me try to find the reason of the mentioned issues.😊

This also raises a question: why does a brew extension involve so many terminal apps? I do not see the necessity to configure special apps for "running xx in terminal." Their macOS has a native terminal anyway. If users really need to run the command in their favorite terminal app (the finished cmd does not allow following interactions, though), they still have the choice to copy the command. This makes the extension heavier and heavier (though it really depends on the definition of the boundary of this extension), making it difficult to maintain. It also prevents contributors from adding or fixing really relevant features (it requires installing all terminals and testing them all out ). This is not an excuse for my bad test :)

extensions/brew/src/components/runInTerminal.ts
ViGeng: the default behavior of iTerm is creating a new session. So I think this feature better not to touch iTerm.

extensions/brew/src/components/runInTerminal.ts
ViGeng: after a lot of tries, I found the reason, it's because though apple script activate the target terminal app, the Raycast search bar still there. Then the to-be-executed command has been keystroked into the search bar. 
The fix should be closing the Raycast window first.

extensions/brew/src/components/runInTerminal.ts
ViGeng: this is exactly the motivation for this feature.
Did not find an elegant solution.
Users may have different settings, so "be simple to create a new window/tab" can be straightforward.

ViGeng:
I am trying to keep minimum changes:
**Terminal**: sometimes the brew cmd can be input to Raycast search bar because the activated target terminal is not at the foremost. Fixed by close Raycast window first.
**other terminal apps**: other than Terminal app, other apps' default behavior is creating a new tab (also locally tested). so better not to touch them. 


Terminal app tested as shown below, I would be much happy to fix if there is any issues.

https://github.com/user-attachments/assets/5f97af2c-ce9e-4037-8b2f-b9921d350c26

ViGeng:
> Apologies for the delay reviewing..
> 
> Thanks for the fixes for 3rd party terminals. That works well.
> 
> Regarding the new tab setting I am not happy including this in the PR - it just doesn't work reliably in any of the terminals. For example, as already described, it doesn't work at all in Terminal using the default macOS window settings. And Hyper creates a new tab regardless of the setting. If we add the tab setting and it doesn't work then users are just going to report bugs, so it doesn't make sense to add it if it's unreliable.
> 
> I'll approve the current PR once the setting is removed.
> 
> If you can make the setting work reliably in all supported terminals then feel free to submit a new PR.
> 
> Thanks 🙏

Hi @nhojb , thanks for your feedbacks. Yes, you are right, it's really important to have a consistent experience among all terminal apps. I agree with your taste.
Then let's remove the new tab feature, if I understand correctly.

nhojb (approved):
Great, thanks for the contribution.

Issue #19930, opened 2025-06-19
[Notion] Quick Capture as "bookmark" pasting linked text


isoscelesxyz:
### Extension

https://www.raycast.com/notion/notion

### Raycast Version

1.100.2

### macOS Version

15.5 (24F74)

### Description

When using the Quick Capture feature, I have "Bookmark" selected but it only adds linked text to the selected page.

### Steps To Reproduce

1. Copy a URL
2. Open raycast and enter the "notion quick capture" screen
3. Attempt to add the link, captured as "Bookmark" to any page
4. Notice it enters linked text on that page, not a bookmark link.

### Current Behaviour

<img width="862" alt="Image" src="https://github.com/user-attachments/assets/24988b12-29cc-4665-9251-1b01072e83db" />
results in 

<img width="638" alt="Image" src="https://github.com/user-attachments/assets/88507909-624f-489b-8325-07305cb9dd65" />

### Expected Behaviour

I would expect "Bookmark" to result in the same Bookmark behavior when pasting a link into a page:

<img width="499" alt="Image" src="https://github.com/user-attachments/assets/9ab543dc-68d9-47c1-9a64-84641a71619c" />

<img width="895" alt="Image" src="https://github.com/user-attachments/assets/22901e1c-e4aa-414f-91aa-6263b21a91e4" />

zsidnam:
Hey @isoscelesxyz ! I looked into this and we can update the extension to create a bookmark object (which would totally make sense) BUT it looks like Notion only fetches the opengraph image and text when you manually create a bookmark in Notion instead of creating one via the API 😞. Here is what it looks like:

<img width="1100" height="510" alt="Image" src="https://github.com/user-attachments/assets/604cd233-43d4-4f18-965d-d1a614a75f96" />

Given this limitation, is it still worth making this change?

wfendler:
> Given this limitation, is it still worth making this change?

Bummer! I think the change would still be nice if it's no trouble. Then I can hit up slack to see if they can fix the other part of the issue on their end, fetching OG data for links submitted via api. 

Thanks! 🙏

zsidnam:
@isoscelesxyz we just merged an update that should address this. Can you give it a try and close out this issue once you verify it is working?

PR #20906, opened 2025-08-12
[Notion] Quick Capture - Use bookmark block instead of markdown link when Capture As is set to Bookmark

zsidnam:
## Description

Updates the Quick Capture command to insert an actual "Bookmark" block instead of a markdown link when the "Capture As" option is set to "Bookmark". Addresses #19930.

**Note:** As mentioned in the Issue, it looks like Notion only enriches the Bookmark with preview content if you manually create the bookmark in Notion. When we create it with the API, we just get a bare bones bookmark object with the URL only.

## Checklist

### DIFF
--- a/extensions/notion/CHANGELOG.md
+++ b/extensions/notion/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Notion Changelog
 
-## [Use Bookmark Block for Quick Capture As Bookmark] - {PR_MERGE_DATE}
+## [Use Bookmark Block for Quick Capture As Bookmark] - 2025-08-13
 
 - Refactor Quick Capture to create a bookmark block instead of a markdown link when "Capture As" is set to "bookmark"
 


extensions/notion/src/quick-capture.tsx
```diff
@@ -230,7 +230,7 @@ function QuickCapture({ launchContext }: QuickCaptureProps) {
       />
 
       <Form.Dropdown {...itemProps.captureAs} title="Capture As" storeValue>
-        <Form.Dropdown.Item title="Bookmark" value="bookmark" icon={Icon.Link} />
+        <Form.Dropdown.Item title="Bookmark" value="url" icon={Icon.Link} />
```
zsidnam: note: I originally changed this to "bookmark" to be more accurate, but then I realized I might break quick links that were referencing "url" so I reverted this change.

pernielsentikaer (approved):
Hi 👋 

Looks good to me, approved 🔥

PR #20940, opened 2025-08-13
[Visual Studio Code] Add Windows support for VSCode

tonka3000:
## Description

Make the VSCode Extension Windows ready.

Downsides: sqlite is pretty slow on Windows, so startup time is not the best. I assume that upgrading node and using the native integrate sqlite can fix that in the future.

## Checklist

### DIFF
--- a/extensions/visual-studio-code-recent-projects/src/actions.tsx
+++ b/extensions/visual-studio-code-recent-projects/src/actions.tsx
@@ -1,8 +1,8 @@
-import { Action, open } from "@raycast/api";
+import { Action, Icon, open } from "@raycast/api";
 import { isWin } from "./utils";
 
 export function OpenInWindowsExplorerAction({ uri }: { uri: string }) {
-  return <Action title="Reveal in Explorer" onAction={() => open(uri)} />;
+  return <Action title="Reveal in Explorer" icon={Icon.Folder} onAction={() => open(uri)} />;
 }
 
 export function OpenInShell({ path }: { path: string }) {


tonka3000:
@pernielsentikaer Can somebody check the behavior. Most stuff works, but it is quite slow to get the project results in comparison to make which should not be there.

There is also kind of a bug with the current windows version: You can not press enter until all the branches are load, which take a long time in my case.

mathieudutour:
> You can not press enter until all the branches are load, which take a long time in my case.

I'm not sure what you mean, what branches?

thomaspaulmann:
> > You can not press enter until all the branches are load, which take a long time in my case.
> 
> I'm not sure what you mean, what branches?

I think we fetch the git status/branch per repository which could take a while. I could imagine that the GIT operations are slower on Windows, making the whole experience slower. There's a checkbox setting for it, so could quickly test if that's it.

tonka3000:
> I think we fetch the git status/branch per repository which could take a while. I could imagine that the GIT operations are slower on Windows, making the whole experience slower. There's a checkbox setting for it, so could quickly test if that's it.

The `useSQL` statement is often quite slow it it is not "hot loaded", sometimes it take 3-10 seconds.

I already turned of the branch feature, but it still take multiple seconds until a `Open in ...` action is available.

EDIT: There was a missing check and so the branch would get queryed anyway. I fix that with the latest commit. The problem that `useSQL` is quite slow, if it is not hot-loaded already, still is a big issue.

tonka3000:
@pernielsentikaer @thomaspaulmann I disabled all forks for now which I don't know the location on Windows. It will fallback to VSCode like in the Mac version

mathieudutour (approved):
Looks good to me! 

I tried to had a look at the `useSQL` slowdown but I'm not sure what it's about because we just use the built-in sql module. I'll try to take a deeper look later

yusifaliyevpro:
Hello @tonka3000 , the "Search Recent Projects" command has an issue. When I use the "Open in Code" action, two bash warnings appear in Terminal. Can you take a look? This only happens when open vscode with this extension



image

PR #20961, opened 2025-08-14
Add gpt-rewriter extension

alshimo:
## Checklist


pernielsentikaer:
Hi 👋

I'm not sure I understand the purpose of this. This should already be possible within Raycast with BYOK, so I think it's better to use that instead, which gives a better native feeling. What do you think about that?

PR #20966, opened 2025-08-14
Update battery-health extension: add adaptor power watts

yz3440:
## Description

I added a charger wattage string (if the computer is connect to power adapter) in the `Charge` item section.

## Screencast

CleanShot 2025-08-14 at 17 06 14@2x

## Checklist

### DIFF
--- a/extensions/battery-health/CHANGELOG.md
+++ b/extensions/battery-health/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Battery Health Check Changelog
 
-## [Add Charger Wattage] - {PR_MERGE_DATE}
+## [Add Charger Wattage] - 2025-10-01
 
 - Attached charger wattage (when available) to the `<ChargeItem>` component.
 
--- a/extensions/battery-health/package.json
+++ b/extensions/battery-health/package.json
@@ -44,5 +44,8 @@
     "fix-lint": "ray lint --fix",
     "lint": "ray lint",
     "publish": "npx @raycast/api@latest publish"
-  }
+  },
+  "platforms": [
+    "macOS"
+  ]
 }


pernielsentikaer:
Hi @yz3440👋 

@o1y do you want to check this?

extensions/battery-health/src/components/ChargeItem.tsx
```diff
@@ -1,8 +1,13 @@
 import StatsListItem from "./StatsListItem";
 
-const ChargeItem = (props: { currentCapacity: number; maxCapacity: number }) => {
-  const formatted =
+const ChargeItem = (props: { currentCapacity: number; maxCapacity: number; chargerWattage: number | undefined }) => {
```
LitoMore: Here we can make this prop optional.

```suggestion
const ChargeItem = (props: { currentCapacity: number; maxCapacity: number; chargerWattage?: number }) => {
```

o1y:
> @o1y do you want to check this?

Looks great! Thanks @yz3440 for the contribution!

o1y:
After playing around with the extension I think it would be more logical and intuitive to display the _charger wattage_ within the PowerSourceItem component, right next to the power source's name. What do you think @yz3440?

Bildschirmfoto 2025-08-18 um 18 16 20

pernielsentikaer:
Could you look into this @yz3440

yz3440:
@o1y I like the suggestion! I think the wattage fits in better in the `PowerSourceItem`. My only complaint is that it happens to be at the bottom of my default raycast window height and hard to see. I think I'll make a new PR later to allow user to hide unimportant items. 
CleanShot 2025-08-31 at 23 05 37@2x
I have a bunch of chargers with different PD quirks so being able to see the wattage while charging is quite essential.

pernielsentikaer:
Could you check this again @o1y

o1y:
Looks good, thanks @pernielsentikaer!

@yz3440 I’m not sure why you’re not seeing all lines. I’ve tested this on two macOS systems with Raycast, and here’s what it looks like for me by default:

Bildschirmfoto 2025-09-01 um 18 06 16

Could it be that you’re using _the larger text size_ setting in Raycast?

Bildschirmfoto 2025-09-01 um 18 12 36

yz3440:
@o1y I have a 14 inch and that might be why. 

I'll think about what's the best way going forward. I don't know how often the `Copy to clipboard` action are useful. Was it an arbitrary UX decision, or the `Copy to clipboard` is  an essential feature? Maybe we can use `Detail` component for displaying this.

I think this PR is good as is. Solves my problem with minor annoying UI bugs. Thanks for the review.

o1y:
> @o1y I have a 14 inch and that might be why.

Yep, same here 🙋‍♂️ This should show all entries. Have you checked the font size setting in Raycast?

yz3440:
@o1y I think you are right. It's a font thing.
CleanShot 2025-09-01 at 21 24 04@2x

pernielsentikaer:
How are we with this PR 😊

yz3440:
@LitoMore Hey when you have time can you review this again? We see some minor layout problems but it's because I had a different font setting. Nothing major.

pernielsentikaer (approved):

PR #21180, opened 2025-08-26
Update zed-recent-projects extension to use latest sqlite Zed schema

remorses:
The latest Zed app sqlite schema changed. I updated the sql to use the new schema. Not sure if this version is sill already published as a wide release so I will keep this PR as draft until someone else notices too

PR written by Claude

### DIFF
--- a/extensions/zed-recent-projects/CHANGELOG.md
+++ b/extensions/zed-recent-projects/CHANGELOG.md
@@ -1,6 +1,6 @@
 #  Zed Recent Projects Changelog
 
-## [Fixes] - {PR_MERGE_DATE}
+## [Fixes] - 2025-09-08
 
 - Use latest Zed sqlite schema
 


extensions/zed-recent-projects/src/lib/zedEntries.ts
```diff
@@ -162,20 +182,20 @@ export function useZedRecentWorkspaces(): ZedRecentWorkspaces {
...
       timestamp,
       host,
       user,
       port
     FROM workspaces
-    LEFT JOIN ssh_projects ON ssh_project_id = workspaces.ssh_project_id
-    WHERE (local_paths IS NOT NULL AND paths IS NULL)
-       OR (local_paths IS NULL AND paths IS NOT NULL)
+    LEFT JOIN ssh_connections ON ssh_connection_id = ssh_connections.id
+    WHERE paths IS NOT NULL AND paths != ''
```
ewgenius: In my opinion, the extension must be backward compatible with older versions; we can't simply use the new schema as it is.
Ideally, we should check which version of Zed the schema has changed from and use different queries accordingly.

ewgenius (changes requested):

extensions/zed-recent-projects/src/lib/zedEntries.ts
remorses: agree, will update that

pernielsentikaer:
Thanks for chiming in @ewgenius 🔥

pernielsentikaer:
@ewgenius can you check again?
@remorses can you also check Greptile feedback?

remorses:
The vulnerability is not really present given it's the user that submits the search query, it would be like saying a local bash script has a vulnerability if you directly pass process.argv in the sql

pernielsentikaer:
Makes sense, so we're waiting for @ewgenius 😊

extensions/zed-recent-projects/src/lib/zedEntries.ts
```diff
@@ -143,47 +164,101 @@ function processRemoteWorkspace(workspace: RemoteWorkspace): ZedEntry | undefine
       host: workspace.host,
     };
   } catch (error) {
-    showFailureToast(error, { title: "Error processing workspace" });
+    console.error("Error processing remote workspace:", error);
```
ewgenius: I think it is better to show the error to the user

extensions/zed-recent-projects/src/lib/zedEntries.ts
```diff
@@ -143,47 +164,101 @@ function processRemoteWorkspace(workspace: RemoteWorkspace): ZedEntry | undefine
...
+    }
+    const version = getSchemaVersionSync(path);
+    return {
+      schemaVersion: version,
+      query: getQueryForSchema(version),
+    };
+  }, [path]);
+
+  // Always call the hook, but use a dummy query if not ready
+  const { data, isLoading, error, mutate } = useSQL<Workspace>(path, query || "SELECT * FROM workspaces LIMIT 0");
```
ewgenius, line 238: Schema version inference should be moved higher up the stack, inside the `withZed` HOC. It already handles zed bundle detection with proper loading state.

extensions/zed-recent-projects/src/lib/zedEntries.ts
pernielsentikaer: `showFailureToast`'s first parameter should already display the `error` to the user, it might be okay as is, don't you think @ewgenius - the user won't see and `console` logs/errors

remorses: I replaced the console.error

pernielsentikaer:
@ewgenius can you check again?

extensions/zed-recent-projects/src/lib/zedEntries.ts
ewgenius: Ok, let's get this merged to support new schema. I'm planning to do some cleanup and refactoring, need to find a time for that.

ewgenius (approved):

ewgenius:
@remorses Thank you for the contribution and for adding support for the new schema!

ewgenius:
@pernielsentikaer could you please take a look?

remorses:
I think it makes sense to merge this right away given the extension is currently broken for most users, more tweaks can be made in follow up PRs

mrevanzak:
i need this to be merge 😿

PR #21213, opened 2025-08-28
Update memos extension

LittleQuartZ:
## Checklist

### DIFF
--- a/extensions/memos/CHANGELOG.md
+++ b/extensions/memos/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Send To Memos Changelog
 
-## [Update & Breaking Change] - {PR_MERGE_DATE}
+## [Update & Breaking Change] - 2025-09-03
 
 - support memos@0.25.0. 
 


xiyiji-official:
I don't know if you have tested it locally, but it cannot be used after installation on Raycast. By reading the code, I found that on line 23 of memosList.tsx, there is an issue with the line:

```javascript
const user = userData.user || {};
```

The problem is that the `getMe()` function returns a result from a `useFetch` hook, where the `data` field may be undefined during the initial load. When the component first renders, `userData` is undefined, so accessing `userData.user` will cause an error. 

You need to check if `userData` exists before accessing `userData.user`, like this:

```javascript
useEffect(() => {
  if (!isLoadingUser && userData?.user) {
    const user = userData.user;
    if (user.name) {
      const userId = +user.name.split("/")[1];
      setCurrentUserId(userId);
    }
  }
}, [isLoadingUser, userData]);
```

pernielsentikaer:
Can you look into above @LittleQuartZ

PR #21319, opened 2025-09-02
Add clipboard-to-obsidian extension

thegalexc:
## Checklist


pernielsentikaer:
How does this compare with e.g. [this extension](https://www.raycast.com/millin_gabani/obsidian-smart-capture) by @trillhause? We already have over 8 Obsidian extensions in the store; maybe we can enhance one of them?

thegalexc:
Definitely don't want to add unnecessary clutter to the extensions store. This is distinct in that
- it is intended to be highly efficient at a specific task: creating a new note with title and body
- Does not have any Obsidian plugin dependencies (Smart Capture requires Advanced URI plugin)

Required updates to Smart Capture:
- [very doable] Ability to set the Vault and Storage Path to defaults (as is, you need to enter the storage path manually each time, which defeats the purpose)
- [unclear to me] Somehow have a different mode that lets you draw from your clipboard instead of manually setting a title and body
  - This point is something I don't know how to accomplish given Smart Capture already tries to apply logic in terms of where you have copied text from in the first place - e.g. if from a browser, then it adds a "Link", etc.

That said, Smart Capture is great. I was about to try my hand at a PR but unclear to me from where I should fork:
https://github.com/trillhause/obsidian-smart-capture hasn't been updated in 2 years but https://github.com/raycast/extensions/tree/d480d47a5c3271f36134614ecdc49b2d447bccf2/extensions/obsidian-smart-capture/ got some (very minor) updates ~6 months ago it seems. I'd prefer not to have to also bring the plugin up to date in terms of dependencies and such from 2 years ago.

So, high level: I definitely agree with consolidation as goal but the path to accomplishing what I need is unclear.

If you all don't think this is a good candidate for the extension store, I can just use it personally no problem (a friend of mine asked for it, which prompted me to submit to the store).

Thanks!

pernielsentikaer:
Thanks for the detailed response and for checking it out, @trillhause. What do you think since you're the author of `Smart Capture`, which might be a good candidate for this feature?

pernielsentikaer:
@trillhause are you still around?

thegalexc:
All good - if this will just bloat your extension store, I'm happy to just use it locally. Good practice for me to write an extension, I'll find another opportunity to get something useful into the store :)

PR #21365, opened 2025-09-04
Change default date format to 24-hour format in Date Converter extension

pvinis:
## Description

There was no AM/PM, so I decided we can show the hours in 24-hour format.

## Checklist

### DIFF
--- a/extensions/date-converter/CHANGELOG.md
+++ b/extensions/date-converter/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Date Converter Changelog
 
+## [Update] - 2025-09-03
+
+- Use 24h time format by default.
+
 ## [Update] - 2025-03-03
 
 - Support configuring primary and secondary actions


asportnoy (changes requested):
Not a fan of this as the default personally. This is already configurable from within your settings if you want it to be 24h for yourself.

pvinis:
the other option is to add AM/PM at the end.

other either 24h or AM/PM. I'm ok with both, but the way it is now is wrong, as it has an incomplete picture of the time.

asportnoy:
I would be ok with adding AM/PM.

Keep in mind that the default format also appears elsewhere (I think the package.json) so make sure to update that too if you do that.

pernielsentikaer:
That's correct, it's already possible to set this in `package.json`

```
        {
          "name": "displayFormat",
          "title": "Default Display Time Format",
          "description": "Default time format for displaying time. By default, 'dddd, MMMM D, YYYY hh:mm:ss' will be used. Available formats can be found in the day.js library documentation.",
          "type": "textfield",
          "default": "dddd, MMMM D, YYYY hh:mm:ss",
          "required": false
        },
        {
          "name": "humanFormat",
          "title": "Human-readable Time Format",
          "description": "Time format for human-readable time. By default, 'dddd, MMMM D, YYYY hh:mm:ss' will be used. Available formats can be found in the day.js library documentation.",
          "type": "textfield",
          "default": "dddd, MMMM D, YYYY hh:mm:ss",
          "required": false
        }
 ```

pernielsentikaer:
Should we close this PR?

asportnoy:
Think so

pvinis:
don't close it. I'll update it and add the am/pm. not having anything with 12hour is just terrible 😁.

PR #21414, opened 2025-09-07
Update zen-browser extension

Jimmy-b36:
## Checklist

### DIFF
--- a/extensions/zen-browser/CHANGELOG.md
+++ b/extensions/zen-browser/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Zen Changelog
 
-## [Simplify Actions and Remove Browser Default Search] - {PR_MERGE_DATE}
+## [Simplify Actions and Remove Browser Default Search] - 2025-09-16
 
 - Simplified new tab and history actions to use native Raycast Action.Open instead of AppleScript
 - Removed Browser Default search engine option and set Google as the new default


Jimmy-b36:
This is meant as a temporary fix to get the extension back up and running since AppleScripts seem to have stopped running. By using the native `Action.Open`, it might also work on Windows now; however, this is untested

pernielsentikaer:
Hi @Jimmy-b36👋 

@Keyruu seems to work for me, can you check this (or know why the current implementation fails)

extensions/zen-browser/src/actions/index.ts
```diff
@@ -28,50 +28,3 @@ export async function openNewTab(queryText: string | null | undefined): Promise<
 
   return await runAppleScript(script);
```
theherk, line 28: Is this still used? It doesn't seem to be called by the action any longer.

extensions/zen-browser/package.json
```diff
@@ -48,13 +48,9 @@
       "type": "dropdown",
       "required": false,
       "title": "Search Engine",
-      "description": "Search Engine for query (default: Browser Default).",
-      "default": "Default",
+      "description": "Search Engine for query (default: Google).",
+      "default": "Google",
```
theherk, line 52: I'm unsure about this change. Ideally this extension shouldn't know anything at all about search engines. We should just be able to send the search text directly to the browser and let the browser's default do the search. For instance, I've been using Kagi, but there are many search engines out there.

That said, it could be outside the scope of this change. Nevertheless, it should be a goal that we remove this setting altogether, in my view.

Jimmy-b36: I agree with the sentiment. I couldn't find a way to make this behaviour happen though as Open is expecting a url

extensions/zen-browser/src/actions/index.ts
Jimmy-b36: Looks like it has to be returned for the AI extension to work

extensions/zen-browser/src/components/ZenActions.tsx
```diff
@@ -1,52 +1,30 @@
...
-  public static TabListItem = TabListItemAction;
 }
 
 function NewTabAction({ query }: { query?: string }) {
   return (
     <ActionPanel title="New Tab">
-      <Action onAction={() => openNewTab(query)} title={query ? `Search "${query}"` : "Open Empty Tab"} />
+      <Action.Open
+        title="Open with Zen"
+        target={`${SEARCH_ENGINE[getPreferenceValues<Preferences>().searchEngine.toLowerCase()]}${query || ""}`}
```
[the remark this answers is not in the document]
Jimmy-b36: @pernielsentikaer Is this something that needs changing? not sure if the thumbs down is directed at the suggested changes?

pernielsentikaer:
Is this ready for another review @Jimmy-b36 @theherk

MeleMannen:
I think I actually found the underlying issue. In the openNewTab() function, it ran popsToRoot() and closeMainWindow before the script got to run. My guess would be that something recently changed, causing either the code, or just the apple scripts, to no longer. One can try this by simply replacing the old openNewTab() function on the old version with this:
```
export async function openNewTab(queryText: string | null | undefined): Promise<boolean | string> {
  await Clipboard.copy(`${SEARCH_ENGINE[getPreferenceValues<Preferences>().searchEngine.toLowerCase()]}${queryText}`);
  

  const script = `
    tell application "Zen"
      activate
    end tell
    repeat while not application "Zen" is frontmost
      delay 0.1
    end repeat
      ${getNewTabShortcut()}
      delay 0.1
      keystroke "a" using {command down}
      key code 51
      keystroke "v" using {command down}
      key code 36 
    end tell
  `;
  await runAppleScript(script)
  popToRoot();
  closeMainWindow({ clearRootSearch: true });

  return true;
}
```

This actually still closes the Raycast window before doing things in Zen, and it still pops to root. But if it is before the running of the script, nothing happens. @pernielsentikaer if this is a Raycast issue, rather than just this specific extension, it might be worth having a look into what might have happened, as other extensions might also be effected by this sudden change.

PR #21499, opened 2025-09-11
Add azure-devops-branch-helper extension

tortortortortortor:
## Checklist

### DIFF
--- a/extensions/azure-devops-branch-helper/src/BuildLogsView.tsx
+++ b/extensions/azure-devops-branch-helper/src/BuildLogsView.tsx
@@ -12,7 +12,8 @@ import { useState, useEffect } from "react";
 import { runAz } from "./az-cli";
 import PullRequestDetailsView from "./PullRequestDetailsView";
 import { formatRelativeDate, formatDuration } from "./utils/DateUtils";
-import { getBuildStatusEmoji } from "./utils/IconUtils";
+import { getBuildStatusIcon, getBuildResultColor } from "./utils/IconUtils";
+import { getTargetBranch } from "./azure-devops/branch-operations";
 
 // Using runAz utility for Azure CLI execution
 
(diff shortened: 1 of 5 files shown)


pernielsentikaer:
Hi 👋

Did you see the comments from Greptile here? I guess this is not the same as the other extension for Azure in the store?

tortortortortortor:
> Hi 👋
> 
> Did you see the comments from Greptile here? I guess this is not the same as the other extension for Azure in the store?

Ill have a look at comments from Greptile. 

And this is a more "fully fledged" extension, does quite a bit more.

image
image
image
image
image
image
image

pernielsentikaer:
Could you have a look here @crisboarna 

It looks nice indeed, let's see if we can figure out where it fits in 😊

pernielsentikaer:
@crisboarna, are you still around?

tortortortortortor:
Ill have a looksie here soonish, btw.

pernielsentikaer:
Pinged @crisboarna in Slack 🙂

crisboarna:
Hey @pernielsentikaer ,

I like the features that are being proposed, see the value in them, my only concern as you highlighted is that these are in essence features part of Azure, hence I feel that these would be good additional actions to be part of the Azure extension as that is also leveraging the same core, it is building on the same feature set to provide insight into your Azure tenant.

tortortortortortor:
> Hey @pernielsentikaer ,
> 
> I like the features that are being proposed, see the value in them, my only concern as you highlighted is that these are in essence features part of Azure, hence I feel that these would be good additional actions to be part of the Azure extension as that is also leveraging the same core, it is building on the same feature set to provide insight into your Azure tenant.

I would say its more akin to this one https://www.raycast.com/willem_van_lent/ado-search

but that does not allow anything to support our way of working (more or less searching)

And this is thematically about "resources in azure" 
https://www.raycast.com/crisboarna/microsoft-azure

not about development.

pernielsentikaer:
I'm not sure what we should do. I don't know much about Azure, actually. Would it make sense to enhance some of what we already have?

crisboarna:
I would lean more towards integrating in one of the two existing ones as it covers features that are logically under the *Azure* umbrella and this adds extra advantage that it reduces potential for feature clash, not having multiple extensions with the same "feature" / skill as it would clash with what is already present there.

tortortortortortor:
Not all devopsers use Azure though 
image

But, lets just close and kill this. ill just make it available on github for die hard devops people.

pernielsentikaer:
I'm okay with having this external if you don't think it fits in 🙂 I'd make it your call since I don't have enough knowledge about it

tortortortortortor:
> I'm okay with having this external if you don't think it fits in 🙂 I'd make it your call since I don't have enough knowledge about it

Lets send it to the eternal hunting fields. ⚰️ have a nice weekend!

pernielsentikaer:
Closing as requested

PR #21547, opened 2025-09-14
[Forked Extensions] Add "isMac" and "isWindows" helpers

yusifaliyevpro:
## Description

Added helpers to distinguish macOS and Windows devices for better clean code

## Checklist

### DIFF
--- a/extensions/forked-extensions/CHANGELOG.md
+++ b/extensions/forked-extensions/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Raycast Fork Extensions Changelog
 
-## [Improvements] - {PR_MERGE_DATE}
+## [Improvements] - 2025-09-15
 
 - Add `isMac` and `isWindows` helpers to utils
 


extensions/forked-extensions/src/components/sync-fork.tsx
```diff
@@ -46,8 +46,8 @@ export default function SyncFork({
     includeParentheses: true,
   };
 
-  const remoteDiff = getCommitDiffMessage(commitDiff?.github, diffMessageOptions);
-  const localDiff = getCommitDiffMessage(commitDiff?.local, diffMessageOptions);
+  const remoteDiff = useMemo(() => getCommitDiffMessage(commitDiff?.github, diffMessageOptions), [commitDiff]);
+  const localDiff = useMemo(() => getCommitDiffMessage(commitDiff?.local, diffMessageOptions), [commitDiff]);
```
LitoMore: I don't want to use `memo`s in Raycast since they have strict memory usage limits.

See https://developers.raycast.com/information/lifecycle#unloading.

extensions/forked-extensions/src/utils.ts
```diff
@@ -4,6 +4,9 @@ import { Cache, getPreferenceValues } from "@raycast/api";
 import { upstreamRepository } from "./constants.js";
 import { CommitDiff, ForkedExtension } from "./types.js";
 
+export const isMac = process.platform === "darwin";
```
LitoMore, line 7: Unused `isMac`. There is a condition `if (os.platform() !== "darwin") return null` in the codebase. Let's change that to `!isMac`.

extensions/forked-extensions/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Raycast Fork Extensions Changelog
 
+## [Improvements] - {PR_MERGE_DATE}
+
+- Optimize event handling in Sync Fork by using `useMemo`
```
LitoMore: All changes should be noted in the changelog.

LitoMore (changes requested):

extensions/forked-extensions/src/components/sync-fork.tsx
yusifaliyevpro: But how do we assign the new values after syncing the fork? We could also handle this using another useEffect

yusifaliyevpro: Why even we should use setCommitDiff? we can delete them

```tsx
  const { push } = useNavigation();
  const [remoteDiff, setRemoteDiff] = useState<string>();
  const [localDiff, setLocalDiff] = useState<string>();
  
 // Other codes
 
   useEffect(() => {
    catchError(async () => {
      if (!forkedRepository) return;
      // Run local Git command before requesting GitHub in case the network request is too slow to block the execution of other Git operations.
      // But this still has a chance to be conflicted if the user execute another Git operation at the same time.
      const localCommitDiff = await git.getAheadBehindCommits();
      const githubCommitDiff = await api.compareTwoCommits(forkedRepository);
      setRemoteDiff(getCommitDiffMessage(githubCommitDiff, diffMessageOptions));
      setLocalDiff(getCommitDiffMessage(localCommitDiff, diffMessageOptions));
    }, diagnosticsAction)();
  }, [forkedRepository, lastCommitHash]);
```

LitoMore: The `commitDiff` is already a [reactive value](https://react.dev/learn/separating-events-from-effects#reactive-values-and-reactive-logic). There is no need to use more new states to store them.

FYI:

- https://react.dev/learn/choosing-the-state-structure.
- https://react.dev/learn/you-might-not-need-an-effect

extensions/forked-extensions/src/components/sync-fork.tsx
```diff
@@ -18,7 +17,8 @@ export default function SyncFork({
   readonly onSyncFinished: () => void;
 }) {
   const { push } = useNavigation();
-  const [commitDiff, setCommitDiff] = useState<{ github: CommitDiff; local: CommitDiff }>();
+  const [remoteDiff, setRemoteDiff] = useState<string>();
+  const [localDiff, setLocalDiff] = useState<string>();
```
LitoMore: Unnecessary states, as I mentioned in the previous [comment](https://github.com/raycast/extensions/pull/21547#discussion_r2347242738). Please don't resolve the comment until you have actually resolved the comment.

LitoMore (changes requested):

extensions/forked-extensions/src/components/sync-fork.tsx
yusifaliyevpro: So, why isn't "Sync Fork (%s commits behind)" updating after syncing the fork? I don't think these states are reactive in previous implementation.

This same issue occurred in the video-downloader extension as well. I fixed it there using useMemo.

LitoMore: I cannot reproduce. Can you record your local behavior and the diagnostics result? Here is mine:

https://github.com/user-attachments/assets/767e6541-26d5-40fd-9faf-a3e988c15c5e

> This same issue occurred in the video-downloader extension as well. I fixed it there using useMemo.

I don't think you fixed it right. That's not a [memo](https://react.dev/reference/react/memo) used for.

yusifaliyevpro: There must have been an update in the upstream repository. Running the following commands does not reproduce the issue

```bash
git reset --hard <commit>
git push origin main --force
```
However, I can confirm that I did observe the issue after the last update as well.

Let’s put this conversation on hold until tomorrow. If a new PR gets merged into the upstream repo, I will record a video to capture the behavior.

LitoMore: I know the root cause. It has nothing to do with `commitDiff` and `memo`.

```shell
git reset <commit> --hard
git push origin main --force
git pull upstream main
```

You will be able to reproduce the issue.

In the meantime, feel free to remove these changes from the PR. I can fix it and refactor some implementations for the next version.

yusifaliyevpro: Yes good catch


https://github.com/user-attachments/assets/7010a36e-f325-4d58-a92e-6fc48c83a785

yusifaliyevpro: So what is the cause? I will revert some commits

LitoMore: I only used the `(local)lastCommitHash` to control its renders. There should be a `remoteLastCommitHash` to indicate when the component should update.

LitoMore:
@yusifaliyevpro Feel free to mark this PR as ready now.

LitoMore (approved):
@pernielsentikaer LGTM

@yusifaliyevpro Thanks for taking time to contribute.

LitoMore:
@yusifaliyevpro The PR title and description need to be updated.

pernielsentikaer:
Can I merge this now @LitoMore

LitoMore:
@pernielsentikaer Yes, please. Thanks!

PR #21575, opened 2025-09-15
Add folder-organizer extension

mdev64:
## Checklist


pernielsentikaer:
I would say that Python in this extension is overkill and doesn't provide anything that Node.js/TypeScript can't do

For simple file organization by extension, Node.js/TypeScript would be:

✅ Faster (no subprocess overhead)
✅ More reliable (no Python dependency issues)
✅ Easier to maintain (single language/runtime)
✅ Better integrated with Raycast

Do you know how to change this, or do you want me to give it a stab?

mdev64:
So, here’s the thing, this Python script wasn’t originally written for this specific use case. It was the original script I used before I got my Mac. Then, I stumbled upon Raycast and thought, “Why not turn this into an extension?” That’s why I chose to use a Python script here. Let me know if you’d like me to change this 😊

pernielsentikaer:
It would be better to port it actually, then we don't need to think about whether the user has Python 😊

Do you want me to port it?

mdev64:
I’ve updated the script and removed the Python dependency.

pernielsentikaer:
Is there anything else you would like to do in this PR before merging?

mdev64:
I see you updated a few things. I think we can merge now. I don't have anything to push.

mdev64:
can we proceed this please?

pernielsentikaer (approved):

mdev64:
Thank you @pernielsentikaer for merging this. Hope this will be useful to others 🎉 🎉

PR #21594, opened 2025-09-16
Add enhanced OTP support for gopass extension

ajchemist:
- Support otpauth:// URLs in multiple formats (standalone lines, key-value pairs)
- Display real-time OTP codes with countdown timer and progress indicator
- Mask sensitive otpauth URLs while allowing copy/paste functionality
- Add fallback to existing gopass otp command for compatibility
- Add otpauth and @raycast/utils dependencies for OTP generation

Fixes display of OTP codes for entries containing otpauth:// URLs.

### Changes

- **OTP Display**: Shows real-time OTP codes with countdown timer (e.g., "123456 25/30s")
- **URL Support**: Handles `otpauth://` URLs in three common storage formats:
  1. As password replacement in first line
  2. As standalone lines in entry details
  3. As key-value pairs (e.g., `totp: otpauth://...`)
- **Security**: Masks otpauth URLs as "******" in UI while maintaining copy/paste functionality
- **Compatibility**: Falls back to existing `gopass otp` command when direct parsing fails

## Screencast

Screenshot
Screenshot (1)

## Checklist

### DIFF
--- a/extensions/gopass/src/details.tsx
+++ b/extensions/gopass/src/details.tsx
@@ -1,7 +1,28 @@
-import { Action, ActionPanel, Clipboard, closeMainWindow, Icon, List, showHUD, showToast, Toast } from "@raycast/api";
+import {
+  Action,
+  ActionPanel,
+  Clipboard,
+  closeMainWindow,
+  Icon,
+  List,
+  showHUD,
+  showToast,
+  Toast,
+  Color,
+} from "@raycast/api";
 import { useEffect, useState } from "react";
+import { getProgressIcon } from "@raycast/utils";
 import gopass from "./gopass";
-import { humanize, isValidUrl } from "./utils";
+import {
+  humanize,
+  isValidUrl,
+  generateOTPFromUrl,
+  isOtpauthUrl,
+  extractOtpauthUrls,
+  getOTPRemainingTime,
+  getOTPProgress,
+  parseOtpauthUrl,
+} from "./utils";
 import { copyPassword, pastePassword, copyOTP, pasteOTP, removePassword } from "./index";
 import CreateEditPassword from "./create-edit";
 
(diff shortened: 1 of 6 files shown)


extensions/gopass/src/index.tsx
```diff
@@ -77,6 +77,26 @@ export async function pastePassword(entry: string): Promise<void> {
 export async function copyOTP(entry: string): Promise<void> {
   try {
     const toast = await showToast({ title: "Copying OTP", style: Toast.Style.Animated });
+
+    // First try to get OTP from otpauth:// URLs in the entry details
```
andyfase, line 81: Whats the advantage of using the new OTP library vs the inbuilt one from gopass here? I get it for the details page where you can extract the time left on the code but in this case its just a copy into the clipboard?

extensions/gopass/src/index.tsx
```diff
@@ -91,6 +111,26 @@ export async function copyOTP(entry: string): Promise<void> {
 export async function pasteOTP(entry: string): Promise<void> {
   try {
     const toast = await showToast({ title: "Pasting OTP", style: Toast.Style.Animated });
+
+    // First try to get OTP from otpauth:// URLs in the entry details
```
andyfase, line 115: Same as above copy comment - do we need the new library here vs relying on the existing gopass native totp support.

extensions/gopass/src/details.tsx
```diff
@@ -54,32 +137,109 @@ export default function ({ entry }: { entry: string }): JSX.Element {
...
           })}
 
+        {/* Regular key-value pairs (excluding otpauth URLs) */}
         {details
-          .filter((item) => !item.startsWith("otpauth:"))
+          .filter((item) => {
+            // Exclude standalone otpauth:// URLs
+            if (isOtpauthUrl(item)) return false;
+
+            // Exclude key-value pairs that contain otpauth:// URLs
```
andyfase, line 235: This code doesnt work for me. TOTP URLS can contain ":" characters - an example is

```
otpauth://totp/Example:alice@google.com?secret=JBSWY3DPEHPK3PXP&issuer=Example
```

Also its typical that there will be a space between the key and the value, something like this would imo work better:

```
            const itemNoSpace = item.replace(" ", "")
            // Exclude key-value pairs that contain otpauth:// URLs
            if (itemNoSpace.includes(":")) {
              const value = itemNoSpace.substring(itemNoSpace.indexOf(':') + 1)
              if (value && isOtpauthUrl(value)) return false;
            }
```

extensions/gopass/src/details.tsx
```diff
@@ -54,32 +137,109 @@ export default function ({ entry }: { entry: string }): JSX.Element {
...
+          const { period } = parseOtpauthUrl(otpauthUrl);
+          const otpCode = otpCodes[otpauthUrl] || "XXXXXX";
+          const remaining = getOTPRemainingTime(period);
+          const progress = getOTPProgress(period);
+
+          return (
+            <List.Item
+              key={index}
+              title="OTP code"
+              subtitle={otpCode}
```
andyfase, line 150: So we deliberately hide sensitive data (passwords and OTP Codes) here we are going against that and displaying it. Likely we shouldnt in case anyone is shoulder surfing? Or was this deliberate?

extensions/gopass/src/utils.ts
```diff
@@ -16,3 +17,58 @@ export const isValidUrl = (string: string) => {
...
+export const generateOTPFromUrl = (otpauthUrl: string): string => {
+  try {
+    const totp = OTPAuth.URI.parse(otpauthUrl);
+    return totp.generate();
+  } catch (error) {
+    throw new Error(`Failed to generate OTP from URL: ${error}`);
+  }
+};
+
+export const extractOtpauthUrls = (details: string[]): string[] => {
```
andyfase, line 32: This works but actually will split many more times based on the TOTP URL containing ":" characters - perhaps we should just tighten this up a little bit with similar code to above i.e

```
            const itemNoSpace = item.replace(" ", "")
            // Exclude key-value pairs that contain otpauth:// URLs
            if (itemNoSpace.includes(":")) {
              const value = itemNoSpace.substring(itemNoSpace.indexOf(':') + 1)
              if (value && isOtpauthUrl(value)) return false;
            }
```

extensions/gopass/src/details.tsx
```diff
@@ -54,32 +137,109 @@ export default function ({ entry }: { entry: string }): JSX.Element {
...
+                    source: getProgressIcon(progress),
+                    tintColor: getProgressColor(remaining, period),
+                  },
+                },
+              ]}
+              actions={
+                <ActionPanel>
+                  <Action title="Copy to Clipboard" icon={Icon.Clipboard} onAction={() => copyOTP(entry)} />
+                  <Action title="Paste to Active App" icon={Icon.Document} onAction={() => pasteOTP(entry)} />
+                  <Action.Push
```
andyfase, line 166: If we were to add one more action here 

```
<Action title="Copy OTP URL Clipboard" icon={Icon.Clipboard} onAction={() => copy("OTP URL", otpauthUrl)} />
```

Perhaps we can remove any other OTP Items in the code? i.e. Case 2 and Case 3 wouldnt be needed as we can now copy the OTP code itself or the URL from the same Item. Views?

PR #21718, opened 2025-09-21
Update vscode-project-manager extension to add new projects, in case the searched project does not exist

tren03:
## Checklist

### DIFF
--- a/extensions/visual-studio-code-project-manager/CHANGELOG.md
+++ b/extensions/visual-studio-code-project-manager/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Visual Studio Code - Project Manager Changelog
 
+## [Update] - 2025-09-21
+
+- Add "Create New Project" functionality: When no search results are found, users can now create a new project directory and automatically add it to VS Code Project Manager. The feature appears when a "New Project Directory" is configured in preferences and allows users to quickly scaffold new projects or open existing folders that aren't yet managed by Project Manager.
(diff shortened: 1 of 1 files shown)


tren03:
Hey folks
I used to rely on fzf + Neovim to jump into projects, and this PR is my attempt to bring that same speed to Cursor + Raycast. I’ve been using it for about a week and it’s been super useful.
Happy to make any changes if needed.

Cheers!

tren03:
Screenshot 2025-09-21 at 4 24 46 PM

This is how it would Look
Onclick, would open a new window with that project open

If this folder already exists, and is not tracked, it will open the existing project, and add that to the projects.json config

tren03:
It also supports paths 
So if we were to enter a path like the one in the image, it would create the directories, add the project to the projects.json, and open it
image


This would be appended to the projects.json
```json
{
    "name": "helloworld",
    "rootPath": "/Users/user/Projects/caddy-scripts/nonexsistent/helloworld",
    "tags": [],
    "enabled": true
  }
```

tren03:
image
This is how the new config option would look like.
Only if this is set, would we try to create the new project.
If this path is not set, it will fallback to current behavior

theherk:
@raycastbot: This should not be closed. It seems like worthwhile change that should be evaluated by users of the extension. I don't use the extension, so I can't review critically, but it looks like a decent change.

PR #21722, opened 2025-09-21
Add denshobato extension

Yuki-Sakaguchi:
## Description

**Denshobato** is a simple extension that lets you type text and paste it into other applications.

Since we interact with many applications daily, it serves as a universal text input solution that avoids confusion from slight variations in input form specifications across different applications.

The name Denshobato (carrier pigeon)  comes from Japanese practice of using pigeons as a means of sending messages.

### Problem

- **Inconsistent Submit Actions:** Is it Enter or Shift+Enter? You never know until you accidentally send an incomplete message
- **IME Compatibility Issues:** Japanese, Chinese, Korean, and other IME inputs often malfunction in CLI tools, AI chat interfaces, and terminal applications
- **Input Field Limitations:** Some applications have quirky text input behaviors that interfere with natural typing

### Usage

1. Focus on the input form where you want to enter text (e.g., the input form in Claude Code running in the terminal, the Slack chat field, etc.)

2. Launch Denshobato

3. Since there is only one input field, enter the content you want to send there and click `Send Message`

3. Raycast closes, and the content entered in step 3 is pasted into the input field focused in step 1 (the same content is also saved to the clipboard).

## Screencast

denshobato 2025-09-21 at 21 49 51

## Checklist

### DIFF
--- a/extensions/denshobato/package-lock.json
+++ b/extensions/denshobato/package-lock.json
@@ -7,7 +7,8 @@
       "name": "denshobato",
       "license": "MIT",
       "dependencies": {
-        "@raycast/api": "^1.103.2"
+        "@raycast/api": "^1.103.2",
+        "run-applescript": "^7.1.0"
       },
       "devDependencies": {
         "@raycast/eslint-config": "^2.0.4",
@@ -2766,6 +2767,18 @@
       "engines": {
         "iojs": ">=1.0.0",
         "node": ">=0.10.0"
+      }
+    },
+    "node_modules/run-applescript": {
+      "version": "7.1.0",
+      "resolved": "https://registry.npmjs.org/run-applescript/-/run-applescript-7.1.0.tgz",
+      "integrity": "sha512-DPe5pVFaAsinSaV6QjQ6gdiedWDcRCbUuiQfQa2wmWV7+xC9bGulGI8+TdRmoFkAPaBXk8CrAbnlY2ISniJ47Q==",
+      "license": "MIT",
+      "engines": {
+        "node": ">=18"
+      },
+      "funding": {
+        "url": "https://github.com/sponsors/sindresorhus"
       }
     },
     "node_modules/run-parallel": {
--- a/extensions/denshobato/package.json
+++ b/extensions/denshobato/package.json
@@ -21,7 +21,8 @@
     }
   ],
   "dependencies": {
-    "@raycast/api": "^1.103.2"
+    "@raycast/api": "^1.103.2",
+    "run-applescript": "^7.1.0"
   },
   "devDependencies": {
     "@raycast/eslint-config": "^2.0.4",
--- a/extensions/denshobato/src/utils/applescript.ts
+++ b/extensions/denshobato/src/utils/applescript.ts
@@ -1,17 +1,9 @@
-import { exec } from "child_process";
-import { promisify } from "util";
-
-const execAsync = promisify(exec);
+import { runAppleScript } from "run-applescript";
 
 export async function execAppleScript(script: string): Promise<string> {
   try {
-    const { stdout, stderr } = await execAsync(`osascript -e '${script.replace(/'/g, "'\"'\"'")}'`);
-
-    if (stderr) {
-      throw new Error(`AppleScript error: ${stderr}`);
-    }
-
-    return stdout.trim();
+    const result = await runAppleScript(script);
+    return result.trim();
   } catch (error) {
     console.error("AppleScript execution failed:", error);
     throw new Error(`Failed to execute AppleScript: ${error}`);


pernielsentikaer:
Hi 👋

I'm not sure I really understand what this is about, and it might be too niche for others. Is it so you always have the same cmd/cmd+enter instead of every tool having its own?

Yuki-Sakaguchi:
@pernielsentikaer 

Hi! Sorry for the late reply, and thank you for taking the time to review this PR.

The main issue I wanted to solve is that when typing full-width characters (like in Japanese), many editors or CLI tools have input bugs.

For example:
- When using **Claude Code** inside **Zed’s terminal**, Japanese text doesn’t appear while typing.  
  *(I’ve recorded a short video showing this issue — please check it out below. The white editor in the video is Zed running Claude Code. You’ll see that nothing appears while typing, and the text only shows up after confirming the conversion.)*
- In some environments, the composing text appears in a different place than the actual focus.
- Depending on the setup, **Codex CLI** works fine, but **Claude Code** has input issues.

<details>
<summary>🎥 Video demo (click to expand)</summary>

The video shows the same typing operation performed in **five different environments**:

1. Ghostty + Claude Code  
2. iTerm2 + Claude Code  
3. VSCode + Claude Code  
4. Cursor + Claude Code  
5. Zed + Claude Code  

You can see that Japanese text input behaves differently depending on the environment.  
In particular, Zed (the white editor) shows the issue — nothing appears while typing, and text only appears after confirming the conversion.


https://github.com/user-attachments/assets/76ecec6a-f3c9-4d2a-a6cf-652d0a50d55a

</details>

Every time a new tool is released, I often get disappointed when I find that I can’t input Japanese properly.  
And I believe this problem isn’t limited to Japanese — it likely affects languages like Korean and Chinese as well.

Ideally, each tool would handle IME (input method) properly, but with the rapid evolution of AI tools, it’s difficult to request fixes for every one of them.  
So, I wanted to create a reliable input form that ensures proper text input regardless of the tool — a kind of safe interface for IME users.

As a secondary goal, I also thought it would be nice if this helped standardize line break behavior (like `Enter` vs `Shift+Enter`) across different tools such as Slack or Teams.  
That part is more of a side effect — though if the Raycast core ever offered a unified input form for all extensions, that would be amazing!

Anyway, if this feels too niche to be included as an extension, that’s totally fine — I can just use it locally for myself.  
Thank you again for your feedback and review!

pernielsentikaer:
The name denshobato, is that something special? What about the greptile feedback, can you look into that?

@LitoMore maybe you have some me input here?

extensions/denshobato/src/denshobato.tsx
```diff
@@ -0,0 +1,130 @@
...
+      await Clipboard.copy(text);
+      await showHUD("Message copied to clipboard (AppleScript permissions required for auto-paste)");
+      return;
+    }
+
+    try {
+      await sendToFocusedApp(text);
+
+      // Also copy to clipboard for user convenience
+      await Clipboard.copy(text);
```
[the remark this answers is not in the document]
Yuki-Sakaguchi: I believe the current Clipboard.copy() is necessary, since the text isn’t copied beforehand — only pasted or handled as a fallback.
So I think keeping it here is appropriate.

Yuki-Sakaguchi:
Hi @pernielsentikaer ,

Thanks for asking!

**About the name**
“Denshobato” is the Japanese pronunciation of **伝書鳩**, which means “carrier pigeon.”
It follows the same naming concept as [Hono](https://hono.dev/docs/) — both are Japanese words written in romanized form.

I chose this name because it perfectly represents the idea of a messenger — something that sits between the sender and receiver to deliver messages reliably.

**About the feedback**
I’ve fixed the parts mentioned in your feedback, so please take another look when you have a moment.

Thanks again for the thoughtful review!

pernielsentikaer:
Did you have time to check yet @LitoMore

extensions/denshobato/docs/ime-denshobato-spec-en.md
```diff
@@ -0,0 +1,645 @@
...
+  try {
+    // Set text to clipboard
+    await Clipboard.copy(text);
+    
+    // Execute cmd+v via AppleScript
+    await execAppleScript(`
+      tell application "System Events"
+        keystroke "v" using command down
+      end tell
+    `);
```
LitoMore, line 241: Can use just use the [`Clipboard.paste`](https://developers.raycast.com/api-reference/clipboard#clipboard.paste) for better compatibility (Windows support)?

pernielsentikaer: That's a good point, @Yuki-Sakaguchi

PR #21742, opened 2025-09-22
Update github extension

bxb100:
## Checklist

### DIFF
--- a/extensions/github/CHANGELOG.md
+++ b/extensions/github/CHANGELOG.md
@@ -1,6 +1,6 @@
 # GitHub Changelog
 
-## [Unified handle notification icon retrieval error] - {PR_MERGE_DATE}
+## [Unified handle notification icon retrieval error] - 2025-10-02
 
 - Notifications Command and Menu Bar: Display a fallback icon when the notification icon retrieval fails.
 


extensions/github/src/helpers/notifications.ts
```diff
@@ -148,6 +148,12 @@ export async function getNotificationIcon(notification: Notification): Promise<{
   return icon;
 }
 
+export async function getNotificationIcon(notification: Notification) {
+  return _getNotificationIcon(notification).catch(() => {
+    return { value: { source: Icon.Warning, tintColor: Color.Red }, tooltip: "Could not load icon" };
+  });
+}
```
LitoMore: Why not use a simple try-catch?

bxb100: `getNotificationIcon` is used by unread-notification and notification, so I think it's the least impactful change.

LitoMore: I mean you can simply add the try-catch to the original function statement. There is no need to use an underscore-prefixed function.

```typescript
export async function getNotificationIcon(notification: Notification) {
	try {
		// Original function body
	} catch {
		// ...
	}
}
```

bxb100: I thought this method was around 80 lines, adding an auxiliary method might be helpful. but never mind, I'll change it ASAP.

bxb100: done, thanks for your review

LitoMore: Thanks. This may not be the most beautiful code, but it will be less costly for other contributors to read.

pernielsentikaer (approved):

PR #21855, opened 2025-09-29
Add guitar-tools extension

narghev:
## Description

A Raycast extension to collect useful guitar tools in.
Currently it only has a tuner feature, however, in the future I'm planning to add more features.

Please note that I fully understand that guitar-tools might be a very generic name for this extensions and am open to renaming it to `Tuner` or `Guitar Tuner`, however, in the future I'm planning to add more features, like fretboard practice helper, that are going to heavily rely on the code of this extensions. That's why I thought making it generic and adding that as a separate command might be better, instead of adding 2 extensions that share big chunk of common code. But I will go with whatever the maintainers/community suggest.

## Screencast

guitar-tools-1

## Checklist


pernielsentikaer:
What about using `List.Emptyview`, a bit like this 

```ts
import { List, Icon, Color } from "@raycast/api";
import { useEffect } from "react";
import { useTuner } from "./hooks/useTuner";
import { centsToDisplayObject } from "./utils/display.utils";

  const { startContinuousListening, stopContinuousListening, detectedNote, error } = useTuner();

  useEffect(() => {
    startContinuousListening();

    return () => {
      stopContinuousListening();
    };
  }, []);

  if (error) {
    return (
      <List>
        <List.EmptyView
          title={error.title}
          description={error.subTitle}
          icon={{ source: Icon.ExclamationMark, tintColor: Color.Red }}
        />
      </List>
    );
  }

  if (detectedNote) {
    const { noteName, cents } = detectedNote;

    const displayObject = centsToDisplayObject(cents);

    return (
      <List>
        <List.EmptyView
          title={noteName}
          description={`${cents > 0 ? "+" : ""}${cents} cents`}
          icon={{ source: displayObject.icon, tintColor: displayObject.color }}
        />
      </List>
    );
  }
  return (
    <List>
      <List.EmptyView title="Play your instrument" icon={{ source: Icon.Heartbeat, tintColor: Color.Blue }} />
    </List>
  );
}
```

narghev:
@pernielsentikaer thanks for the review. I think I tried this exact UI at some point during development, it is better in terms of keeping everything centered. My only problem with it is how small the Note name looks compared to the Icon, which is IMO less important than the note name, but I don't mind going with this suggestion.

PR #21914, opened 2025-10-02
Update sesh extension

sebb3:
## Description
Added a user setting for manually entering the value of PATH.
There are probably several use cases for this, mine was that I have tmux and sesh installed under my home-folder.

## Checklist


pernielsentikaer:
Hi @sebb3👋 

Can you look into the suggestions by Greptile?

@joshmedeski could you check this?

extensions/sesh/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Sesh Changelog
 
+## [Make PATH configurable]
+- Add a user-setting for the PATH variable
+- This is necessary for home-manager installations, because the path includes the users home-directory, and Raycast doesn't expand '~'.
```
joshmedeski: I would remove "This is necessary..." from the changelog

extensions/sesh/package.json
```diff
@@ -26,6 +29,14 @@
           "title": "Open with (required)",
           "description": "Open with",
           "default": "Terminal"
+        },
+        {
+          "name": "searchPath",
```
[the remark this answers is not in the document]
joshmedeski: Make sure to follow this guide :)

extensions/sesh/src/env.ts
```diff
@@ -1,3 +1,15 @@
...
+
+interface Preferences {
+  searchPath: string;
+}
+
+export function getEnv() {
+  const { searchPath } = getPreferenceValues<Preferences>();
+
+  const env = Object.assign({}, process.env, {
+    PATH: searchPath,
```
joshmedeski: So this overwrites the value entirely? What do you think about a comma separated list that can be appended to the path instead? Just an idea.

extensions/sesh/package.json
joshmedeski: What do you think about `envPath` or `userPath` instead? Search feels misleading to me. I get what you were going for but it doesn't feel semantic to me.

joshmedeski (changes requested):
Thanks, this would actually help me as well :)

extensions/sesh/src/env.ts
sebb3: I actually added what you already had there as the default value. I personally would prefer to have full control of it, over something hard-coded that MAY create problems..
What do you think?

sebb3: Updated this to in addition to providing the old value as a default, also merge the old env.PATH into the new value

joshmedeski (approved):
Thanks, looks good to me.

sebb3:
@pernielsentikaer Ready for review

PR #22039, opened 2025-10-08
Add everything-ext extension

LorryGailius:
## Screencast
screenshot

## Checklist

### DIFF
--- a/extensions/everything-ext/README.md
+++ b/extensions/everything-ext/README.md
@@ -1,12 +1,8 @@
 # Everything Extension for Raycast (Windows)
 
-![Demo GIF](./media/demo.gif)
+![Screenshot showing the main search interface](./metadata/everything-search-1.png)
 
-*Quick demo showing the extension in action*
-
-![Extension Screenshots](./metadata/everything-search-1.png)
-
-*Store-ready screenshot showing the main search interface*
+![Quick demo showing the extension in action](./media/demo.gif)
 
 A powerful Raycast extension that integrates with Everything CLI to provide lightning-fast file system search and navigation on Windows. Transform your file discovery workflow with instant search results and comprehensive directory browsing capabilities.
 
@@ -46,24 +42,19 @@ A powerful Raycast extension that integrates with Everything CLI to provide ligh
 
 ## 📦 Installation
 
-### Prerequisites
-1. **Node.js**: Install the latest version
-   ```bash
-   winget install -e --id OpenJS.NodeJS
-   ```
-
-2. **Everything CLI**: Install the command-line interface for Everything
+1. **Install Everything CLI**: Install the command-line interface for Everything
    ```bash
    winget install --id=voidtools.Everything.Cli -e
    ```
 
-3. **Everything Desktop App**: Ensure Everything is installed and running (for indexing)
+2. **Install Everything Desktop App**: Ensure Everything is installed and running (for indexing)
    - Download from [voidtools.com](https://www.voidtools.com/)
-   - Or install via: `winget install voidtools.Everything`
+   - Or install via: 
+   ```bash
+   winget install voidtools.Everything
+   ```
 
-### Extension Installation
-
-This extension is available through the Raycast store. Search for "Everything" in the Raycast store to install it.
+3. **Install Extension**: This extension is available through the Raycast store. Search for "Everything Search" in the Raycast store to install it.
 
 ## 🛠️ Configuration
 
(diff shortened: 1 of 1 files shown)


LorryGailius:
Due to lack of time of @dougfernando, the previous https://github.com/raycast/extensions/pull/21885 PR is outdated, created this PR for the most up to date version of the extension

LorryGailius:
@mathieudutour any news on this?

mathieudutour:
Sorry I missed this. I still think this should be an update to the existing Everything extension. It can be a separate command. I understand the original author doesn't have time to do it but that doesn't mean that we can't accept contributions to it

aventuraviper:
> Sorry I missed this. I still think this should be an update to the existing Everything extension. It can be a separate command. I understand the original author doesn't have time to do it but that doesn't mean that we can't accept contributions to it

Are you suggesting that @LorryGailius could make a pull request against the other extension directly in this repository that adds all this functionality to it via a seperate set of commands? How does that help future maintenenace of either extension as it would automatically make the very people who have stated they don't have time to co-author be co-authors and would just lead to no maintenance of either extension.

Please correct me if I'm wrong. I really want to see this extension come to the store as the native file search cannot (currently) search network drives, but this doesn't seem the right way to bring the extensions together as one.

mathieudutour:
> Are you suggesting that @LorryGailius could make a pull request against the other extension directly in this repository that adds all this functionality to it via a seperate set of commands?

Yes exactly (it wouldn't a set of commands - at the moment each extension have only one). otherwise there would be 2 "Everything" extensions which would be very confusing to the users.

> How does that help future maintenenace of either extension as it would automatically make the very people who have stated they don't have time to co-author be co-authors and would just lead to no maintenance of either extension.

Well if they don't have time to maintain their extension regardless, I guess it won't improve or worsen the maintenance.

But if there are 2 extensions virtually indistinguishable from each other, you can be sure that some users will report bugs or ask features from one for the other

aventuraviper:
@mathieudutour thank you for your response and sorry to clutter this PR. I just think there is a difference in time commitment if you maintain a project such as a raycast extension as one person vs co-authoring.

Maybe I'm biased because I just want this extension in Raycast so I can migrate from Flow Launcher, but I don't think there would be confusion between this extension and the other Everything based one. If a user has installed Everything, they are more than likely capable of reading the short extension descriptions to select the correct one for their needs.

The other extension requires manual config after install, so if you merge these two extensions then the functionality of this one is likely going to persist beyond the intention of the original and it might confuse users who were looking for the functionality of the original as the extension will appear to be working, just not correctly. Perhaps an update to both Readmes to point to each other and help users choose the correct one would be sufficient?

LorryGailius:
I aggree with @aventuraviper that merging the project together in PR for Raycast will degrade the maintainability a lot. One option would be to create a separate repository where these are merged as a whole other "project", mentioning the initial implementations of both @anastasiuspernat and @dougfernando in the readme and crediting them in the configuration of the extension. I am fine with merging them together myself, but only with approval from both of them.

> The other extension requires manual config after install, so if you merge these two extensions then the functionality of this one is likely going to persist beyond the intention of the original

The command from [ `everything-search`](https://github.com/anastasiuspernat/everything-search) needs to be configured before it can be used, but Raycast supports disabling commands by default, so it could be set up that a user configures and enables the command manually if it will be used

phamthaison1207:
As a user, the Everything extension that is currently in the store is confusing to install and doesn't actually give you a straightforward way of using it and it doesn't do what Everything is supposed to do: easy, lightning fast file indexing and querying

That extension requires setting up a server???, as a normal user, I definitely don't know how to do that so I suggest removing that extension because it's not as straightforward as this one during setup phase and using phase.

I have been using this extension (this everything search, not the one currently in the store) and it has been great, lightning fast, queries literally everything and does what I want with minimal fuss.

I don't even see a repo for the Everything extension that's in the store so my opinion regarding that extension would be:
1/5 installation 1/5 ease of use 1/5 simplicity 0/5 for indexing and querying

This current extension however is different:
5/5 indexing 2/5 installation (cause well manually through the terminal cuz it's not in the store) 5/5 ease of use 3/5 simplicity

mathieudutour:
Let's not create a 3rd extension haha

Could you just create a new PR that adds this command to the existing extension? You can add `@dougfernando` as a `pastContributors` if they aren't interested in maintaining it.
We can keep the existing command, move the preference to the command instead of being at the extension level.

LorryGailius:
@mathieudutour a few comments on this:

> Could you just create a new PR that adds this command to the existing extension?

There must be a repository with the full extension being present for maintainability, that is why i suggested creating a 3rd repo, currently having two sources and merging them into one for the store is very counter-intuitive, for example what would the source URL be in the store? Neither repository fully works since it will be missing half of the functionality of the extension. After the release to public there is an increase of users interested in this and there will definetly be people looking forward to maintaining and improving this

mathieudutour:
Oh I mean if you want to create another repo you can - but the source of truth of the extension will be _this_ repo. Some people might contribute to the extension without going to your repo.

LorryGailius:
Created the PR https://github.com/raycast/extensions/pull/23431 which merges this extension with the already existing one. Would really like to move the discussion with you on to there

PR #22130, opened 2025-10-12
Update toggle-fn extension

jcelaya775:
## Checklist

### DIFF
--- a/extensions/toggle-fn/CHANGELOG.md
+++ b/extensions/toggle-fn/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 All notable changes to this project will be documented in this file.
 
-## [1.0.1] - {PR_MERGE_DATE}
+## [1.0.1] - 2025-10-13
 
 ### Fixed
 


Az0s:
@jcelaya775 Great work fixing the issue! Could you explain a bit about how this change resolved it? I was trying to trace the root cause myself but couldn’t quite figure out where the problem was.

jcelaya775:
> @jcelaya775 Great work fixing the issue! Could you explain a bit about how this change resolved it? I was trying to trace the root cause myself but couldn’t quite figure out where the problem was.

Thanks! I'm glad this extension is usable again. Basically, the `runAppleScript` function returned a promise, but it was not being awaited, so the program would exit early, before the script could execute. By returning the promise, the caller of the exported function could properly await it so that the script could finish running (whereas before, the exported function was likely being awaited, but the promise returned from `runAppleScript` was not). Similarly, replacing `return` with `await` would also work.

Edit: After looking into this behavior more, I believe the javascript event loop may keep track of promises, timers, etc., and keeps running until there is nothing left, which could explain why it worked before. I suspect maybe something with js environment or the raycast code affected this behavior. But in any case, making sure to explicitly handle promises should ensure the code behaves as expected.

Az0s:
> > @jcelaya775 Great work fixing the issue! Could you explain a bit about how this change resolved it? I was trying to trace the root cause myself but couldn’t quite figure out where the problem was.
> 
> Thanks! I'm glad this extension is usable again. Basically, the `runAppleScript` function returned a promise, but it was not being awaited, so the program would exit early, before the script could execute. By returning the promise, the caller of the exported function could properly await it so that the script could finish running (whereas before, the exported function was likely being awaited, but the promise returned from `runAppleScript` was not). Similarly, replacing `return` with `await` would also work.
> 
> Edit: After looking into this behavior more, I believe the javascript event loop may keep track of promises, timers, etc., and keeps running until there is nothing left, which could explain why it worked before. I suspect maybe something with js environment or the raycast code affected this behavior. But in any case, making sure to explicitly handle promises should ensure the code behaves as expected.

Thanks for your reply! These kinds of errs caused by async calls are always hard to track down — so it’s even more impressive you spotted it! and yeah, i feel like it might have something to do with raycast’s way of ipc’ing through js extension processes — maybe it kills the process once the main function returns, regardless of any pending promises. i’ll def keep this in mind next time i run into some weird bugs!

PR #22239, opened 2025-10-17
Update parcel extension

aidenkt:
## Description

- I replaced using Universal Links (parcel://) with Parcel's new Add Delivery API. This adds cross-compatibility with Windows as parcel:// relies on the Mac-only Parcel Desktop app. It also makes adding deliveries much faster. 
- I also added form validation to make description and tracking number on the Add Delivery form to be filled in as otherwise Parcel returns an API error. 
- Previously when a delivery was successfully added, the behavior was to return a Success toast and then close the Raycast window immediately after. I switched to a Raycast HUD notification to do this automatically and ensure the user can see the Delivery Added notification as it's hard to see in the current implementation.
- I added API functions to grab Parcel's list of carriers and added it as a field on the form, as the Add Delivery API requires a carrier to be specified.

Note: I wasn't sure to add Windows as a platform in the manifest as I haven't tested with Windows and I'm not sure if that's currently the only block.

## Screencast

Screenshot 2025-10-17 at 12 08 59 AM

## Checklist

### DIFF
--- a/extensions/parcel/CHANGELOG.md
+++ b/extensions/parcel/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Parcel Changelog
 
-## [Add Delivery Overhaul] - {PR_MERGE_DATE}
+## [Add Delivery Overhaul] - 2025-11-06
 
 - Add Delivery now works on both Mac and Windows without requiring the Parcel Mac app to be installed by using Parcel's Add Delivery API
 - The Add Delivery form now prevents leaving required form fields empty


extensions/parcel/src/api.ts
```diff
@@ -75,6 +80,12 @@ export function getApiKey(): string {
   return preferences.apiKey;
 }
 
+export function getUserLanguage(): string {
+  const preferences = getPreferenceValues<Preferences & { locale?: string }>();
+  const locale = preferences.locale || "en";
+  return locale.split("-")[0]; // Extract language code (e.g., "en" from "en-US")
+}
```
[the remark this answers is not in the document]
pernielsentikaer: @aidenkt did you see this comment?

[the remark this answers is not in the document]
aidenkt: Saw it and just resolved it! That was a remnant of when I thought Raycast had multi-lingual support, it should be good to go now.

pernielsentikaer:
@aidenkt I updated the code a bit, bringing `language` back. Can you check that I didn't break anything too? 😊

aidenkt:
All commands seem to work on my end!

pernielsentikaer:
@marcmagn1 / @xilopaint  does it still work for you?

xilopaint:
> @marcmagn1 / @xilopaint does it still work for you?

Yes, but I think we should handle these errors with toast messages:
Screenshot 2025-10-17 at 08 02 31
Screenshot 2025-10-17 at 08 11 10
Also, is there API for deleting deliveries? It would be a good addition for the My Deliveries command.

It would also be beneficial to include a "Create Delivery" action in the action panel of the My Deliveries command.

pernielsentikaer:
Sounds like a good idea to check, @aidenkt can you have a look?

extensions/parcel/src/api.ts
```diff
@@ -105,6 +154,34 @@ export async function fetchDeliveries(filterMode: FilterMode): Promise<Delivery[
...
 
+export async function addDelivery(trackingNumber: string, carrierCode: string, description: string): Promise<void> {
+  const url = "https://api.parcel.app/external/add-delivery/";
+  const response = await fetch(url, {
+    headers: {
+      ...getAPIHeaders(),
+      "Content-Type": "application/json",
+    },
+    method: "POST",
+    body: JSON.stringify({
```
bendrucker, line 140: Confirmation notifications should be supported too (and added to the form):

> send_push_confirmation (bool, optional). Set this to true if you want to get a push notification sent to you once this delivery is added. Default: false.

The [docs](https://parcelapp.net/help/api-add-delivery.html) explain that validation is just that the delivery number matches the carrier's pattern and not any actual status checks before returning 200:

> Deliveries added using this API will not receive tracking information immediately. You will see a "No data available" message for the new delivery in the app until the server first updates it.

> When a request is submitted, the server will check if the tracking number is in a “known” format. If it is not, your request will fail with the relevant error.

extensions/parcel/src/api.ts
```diff
@@ -85,6 +117,23 @@ export function getAPIHeaders(): Record<string, string> {
   };
 }
 
+export function getSupportedCarriersUrl(): string {
+  return "https://api.parcel.app/external/supported_carriers.json";
+}
+
+export async function fetchSupportedCarriers(): Promise<Carrier[]> {
```
bendrucker: Dead code

extensions/parcel/src/api.ts
```diff
@@ -75,6 +83,30 @@ export function getApiKey(): string {
...
 }
 
+function getSystemLanguage(): string {
+  try {
+    const locale = new Intl.DateTimeFormat().resolvedOptions().locale;
+    const languageCode = locale.split("-")[0].toLowerCase();
+    if (languageCode && languageCode.length >= 2) {
+      return languageCode;
+    }
+  } catch {
```
bendrucker: Avoid empty catch blocks, always do something, if only logging a warning

extensions/parcel/src/api.ts
```diff
@@ -75,6 +83,30 @@ export function getApiKey(): string {
   return preferences.apiKey;
 }
 
+function getSystemLanguage(): string {
+  try {
+    const locale = new Intl.DateTimeFormat().resolvedOptions().locale;
+    const languageCode = locale.split("-")[0].toLowerCase();
+    if (languageCode && languageCode.length >= 2) {
```
bendrucker: You should make sure country-specific language variants are supported. The docs say they are not:

> To be provided as a two letter code (ISO 639-1).

extensions/parcel/src/api.ts
```diff
@@ -75,6 +83,30 @@ export function getApiKey(): string {
   return preferences.apiKey;
 }
 
+function getSystemLanguage(): string {
```
bendrucker, line 86: This is quite complicated, is this really the right approach, to detect and use the user's locale country automatically?

> language (string, optional). Language to use for your delivery information. To be provided as a two letter code (ISO 639-1). **Availability depends on a carrier.** Default: en.

Emphasis mine. If your locale is different than what your carrier supports, what happens? Is it just silently ignored and falls back to `en` on the server? If so this is probably fine. But it should really be tested.

Even if it does seem to work, it might be a better idea to use this to set a default form field value rather than magically deriving it without providing any visibility for the user.

extensions/parcel/src/api.ts
```diff
@@ -75,6 +83,30 @@ export function getApiKey(): string {
...
+    const languageCode = locale.split("-")[0].toLowerCase();
+    if (languageCode && languageCode.length >= 2) {
+      return languageCode;
+    }
+  } catch {
+    // Could not get language from Intl.DateTimeFormat().resolvedOptions().locale
+  }
+
+  // Fallback to LC_ALL environment variable
+  const langVar = process.env.LC_ALL;
```
bendrucker: This is not safe/reliable:

https://www.baeldung.com/linux/locale-environment-variables
https://stackoverflow.com/a/30480596


There are a lot of other language environment variables, `LC_ALL` is the broadest/last resort. It can also have special values like `LC_ALL=C` and `LC_ALL=POSIX`.

I would advise removing this entirely, the rabbit hole of language detection in environment variables goes far too deep, it's not something you want to handle manually.

extensions/parcel/src/add-delivery.tsx
```diff
@@ -1,62 +1,46 @@
...
 
 interface FormValues {
   trackingNumber: string;
+  carrierCode: string;
   description: string;
 }
 
 export default function Command() {
-  const [isLoading, setIsLoading] = useState(false);
+  const { carriers, isLoading: carriersLoading } = useCarriers();
```
bendrucker: Error should be handled

aidenkt:
@xilopaint There is no Delete Delivery endpoint in their docs right now, but I will get to work on displaying the errors.

@bendrucker I will add the confirm notification checkbox to the form on my next commit. @pernielsentikaer could I ask you to take a look at the comments Ben made on the language?

bendrucker:
Re: languages, it'll be important to check how the API actually behaves. Using the native locale API and grabbing the first two characters seems reasonable. But even then I'd say it should be a smart default for either a user preference or per-delivery form field or both, and not a magic behavior internal to the extension.

aidenkt:
User preference would make more sense to me, per-form feels excessive as you'd likely choose the same option every time.

pernielsentikaer:
Ssounds like a preference would work better, do we know which contries is wprking as a setting @aidenkt

aidenkt:
@pernielsentikaer Honestly I've yet to test the languages/countries. It's hard to when the API itself is throttled at 20 requests per day per token and since I don't have any valid tracking numbers to test with. Additionally, according to the Parcel API documentation the languages don't work for every carrier and depend on each making it available. I'll reach out to the developer to see if there is a way on determining the available countries per carrier.

I have a commit that fixes the error handling, action menu, and adds the confirmation notifications ready to push. I am having trouble with pushing it for some reason though - I messaged you on Slack about possibly troubleshooting that issue.

aidenkt:
@pernielsentikaer @bendrucker 
I've received a response from the developer of Parcel:
image
With this, I would propose building it in as Per did in his commit rather than making it a user exposed preference, especially if it's not always able to give the language the user requests. What do you guys think?

bendrucker:
Sure that's fine. Getting the users language should be as simple as this:

```js
Intl.DateTimeFormat().resolvedOptions().locale.slice(0, 2)
```

No conditionals/fallbacks needed

aidenkt:
Made the following requested changes:
- Confirmation notification has been added to the form
- Used @bendrucker's simplified language line
- Add Delivery has been added to the Action Menu on the My Delivery pages
- Improved error handling to appear in the toasts (@xilopaint's request)

This PR should now be ready for review! 🎉

pernielsentikaer:
@greptileai, could you please do a fresh review here?

extensions/parcel/src/add-delivery.tsx
pernielsentikaer: did you see this one @aidenkt

pernielsentikaer:
@bendrucker and @xilopaint do you have more feedback?

xilopaint:
No, that's all.

bendrucker:
Getting this so I can't resolve the handled comments:

Screenshot 2025-10-31 at 11 40 50

Still applicable are:

* Dead code (`fetchSupportedCarriers`)
* Unhandled error (from `useCarriers`)

extensions/parcel/src/add-delivery.tsx
aidenkt: Just did and handled it in my new commit

aidenkt:
@bendrucker Just addressed both!

aidenkt:
Hey all, just wanted to check in and see if there were any updates here.

PR #22249, opened 2025-10-17
Update Blockchain Explorer Search to v1.0.0

Namaskar-1F64F:
3 year update! 

finally got back to some fixes with the help of AI 

biggest changes are adding the ability to add new explorers so it doesn't need updated when something changes, it's configurable

saved addresses

better token lookup with routescan


pernielsentikaer:
Hi @Namaskar-1F64F👋 

Could you look into feedback from Greptile?

Issue #22180, opened 2025-10-14
[Harvest] ... Memory Leak


mrpunkin:
### Extension

https://www.raycast.com/eluce2/harvest

### Raycast Version

1.103.3

### macOS Version

26.0.1

### Description

Any time I go to the "New Time Entry" command it renders the page for a few seconds then crashes.

#### Error:
```
Error: Worker terminated due to reaching memory limit: JS heap out of memory

[kOnExit]:worker:314:26
Worker.<computed>.onexit:worker:230:20
```

### Steps To Reproduce

1. Open Raycast
2. Navigate to "New Time Entry" command for extension
3. Wait a second or two

### Current Behaviour

Loads the new timer form for a few moments, then crashes the extension completely. See screenshots.

 
<img width="886" height="586" alt="Image" src="https://github.com/user-attachments/assets/3c676dd8-9893-4eb3-b1e0-b1568497c46b" />

<img width="886" height="586" alt="Image" src="https://github.com/user-attachments/assets/217e84e8-3071-4d33-886d-562888ae9a66" />

### Expected Behaviour

Stay on the form page without crashing.

PR #22256, opened 2025-10-17
Update harvest extension

mrpunkin:
## Checklist

### DIFF
--- a/extensions/harvest/src/services/harvest.ts
+++ b/extensions/harvest/src/services/harvest.ts
@@ -23,7 +23,9 @@ import axios, { AxiosError, AxiosRequestConfig, AxiosResponse } from "axios";
 import { NewTimeEntryDuration, NewTimeEntryStartEnd } from "./requestTypes";
 import dayjs from "dayjs";
 import duration from "dayjs/plugin/duration";
+import isToday from "dayjs/plugin/isToday";
 dayjs.extend(duration);
+dayjs.extend(isToday);
 import { useCachedPromise, useCachedState } from "@raycast/utils";
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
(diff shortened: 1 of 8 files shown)


mrpunkin:
@eluce2 See changelog for everything I touched. I started with the UI updates I thought might benefit others besides just myself, but ended up reworking some of the caching and such to try and fix my memory error I was getting also.

So this not only has some UI changes, but also fixes #22180 for users like myself with lots and lots of projects (too many to store in cache)

mrpunkin:
Sorry I probably should have waited to let @eluce2 mark this as ready for review and gotten his feedback on changes first.

eluce2:
No worries, it tags me anyway as the extension author :) I’ll take a closer look this weekend, thanks so much for the contribution!

eluce2 (changes requested):
After testing this on my machine (and with my data) I do think that the task vs project name being the primary list item title should be a preference, with the old way set as default so as not to change any existing workflows. At my company for example the project name is the differentiator and the task name is usually generic. I'm fine if that becomes an extension-level preference though and applies to all list views.

Otherwise the rest of performance improvements look good to me

extensions/harvest/src/services/harvest.ts
```diff
@@ -92,28 +94,25 @@ export function useActiveClients() {
 
 async function fetchProjects() {
   let project_assignments: HarvestProjectAssignment[] = [];
-  let page = 1;
-  // eslint-disable-next-line no-constant-condition
-  while (true) {
+  let pageParams = {};
+
+  while (pageParams) {
```
eluce2: This should remain while (true)
```suggestion
  while (true) {
```

mrpunkin:
@eluce2 Great idea to put it in a preference, as we both seem to have very valid uses for the format we each chose. I rolled back to the inline eslint config in pagination and implemented the preference for Title Field. Take a look when you have a chance.

eluce2 (approved):

eluce2:
@pernielsentikaer This looks good from my end :)

pernielsentikaer (approved):
Looks good to me, approved 🔥

jerzygangi:
@mrpunkin @pernielsentikaer @eluce2 After this update I'm no longer able to change the Task using the task dropdown. Could we revert to the previous (working) version, or apply a fix?

mrpunkin:
@jerzygangi I am not able to reproduce the issue you are having myself. Perhaps @eluce2 knows more about what might be happening.

eluce2:
@jerzygangi I found this issue and have a [new PR](https://github.com/raycast/extensions/pull/22545) to fix

PR #22271, opened 2025-10-18
Unicode Search extension: Unicode codepoint lookup and reverse lookup

blueset:
## Description

Unicode Search is a Raycast extension to lookup and reverse lookup Unicode characters with their names.

### Search character by description

In “Search Unicode”, type your query (ex. `subscript`) to get a list of characters
matching the keyword.

Press <kbd>Enter</kbd> to paste the character to your active input (ex. `₀`), press <kbd>Ctrl</kbd> + <kbd>Enter</kbd> (or <kbd>Cmd</kbd> + <kbd>Enter</kbd>) to copy the character. Press <kbd>Ctrl</kbd> + <kbd>D</kbd> (or <kbd>Cmd</kbd> + <kbd>D</kbd>) to see more information about the character.

### Search character by code point
Type a sequence of hexadecimal codepoints (ex. `ff10 fffd`) to look up characters by their codepoints.

### Identify characters in a string

Press <kbd>Enter</kbd> to show details about the character, or press <kbd>Ctrl</kbd> + <kbd>Enter</kbd> (or <kbd>Cmd</kbd> + <kbd>Enter</kbd>) to have details shown in a side bar.

## Screencast

u_subscript
u_ff10_fffd
u_details
uid_lenny
uid_lenny_details

## Checklist

  - Duplicated screenshots from `/images` to `/metadata`.


pernielsentikaer:
Hi @blueset👋 

I'm not completely sure if it's the same but couldn't we enhance https://www.raycast.com/mmazzarolo/unicode-symbols by @mmazzarolo

blueset:
Hi @pernielsentikaer,

Thank you for the suggestion. I’ve tried it out and took a look of the source code of the extension. While both extension involves Unicode and may look similar on the surface, I found a few differences between them:

* `mmazzarolo/unicode-symbols` seems to have missing a lot of codepoints, even with the “full data set” enabled.
   |`mmazzarolo/unicode-symbols`| This extension |
   |:--:|:--:|
   |image|image|
* While Unicode has 297,334 assigned characters as of version 17, the “full data set” claims to have 40,000+ characters (as stated in the Settings page), which only covers 13% of Unicode.
* `mmazzarolo/unicode-symbols` loads the entire data set JSON into Node.JS RAM, which is not suitable for full search of ~297k Unicode characters if the data source is to be expanded to cover everything.
* The data source used by `mmazzarolo/unicode-symbols`, [`unidata16` on NPM](https://github.com/j3lte/unidata), omits characters in various blocks (Tangut, Hangul Syllables, CJK Unified Ideographs, just to name a few), rendering it unsuitable for reverse lookup.
* My extension is written with content and search backed by a bundled binary utility (`uni`), which is significantly different from `mmazzarolo/unicode-symbols`. Porting the existing extension to use `uni` as the backend would require a major rewrite.
* As my extension is backed by `uni`, it couldn’t be considered backward compatible with `mmazzarolo/unicode-symbols` in aspects like its [addition of Apple logo to the Unicode Dataset](https://github.com/raycast/extensions/blob/e7808f9208c1ab48ffc1eb5a0da7c41c612341ed/extensions/unicode-symbols/scripts/generate-dataset.ts#L77C1-L81C3) and usage of `fuse.js` for fuzzy search.

From my understanding, this extension would fall into the following category in [the guideline](https://developers.raycast.com/basics/prepare-an-extension-for-store#contributing-to-existing-extensions-vs-creating-a-new-one)

> **When you should consider creating a new extension instead of contributing to an existing one**
> * The changes to an existing extension would be significant and might break other people's workflows.

pernielsentikaer:
What do you think, @j3lte, about this? You have been working a lot lately with the symbols extension

j3lte:
@pernielsentikaer I'll have a look at the source code and see what could be ported. If it requires some rewriting, that shouldn't be a problem. I do foresee it might be problematic to also support Windows then, unless `uni` also has a Windows client.

I'll have a look at it tonight

blueset:
> I do foresee it might be problematic to also support Windows then, unless `uni` also has a Windows client.

I’ve been mainly developing this on Windows (only switched over to macOS for screenshots; the two screenshots from my comment above are taken on Windows), so Windows support is not a problem. In case Raycast will support any other platforms, I’ve also bundled a WASM version for better compatibility.

j3lte:
@pernielsentikaer @blueset 

I have had a look at the PR and the comment, I have a few remarks:

> * While Unicode has 297,334 assigned characters as of version 17, the “full data set” claims to have 40,000+ characters (as stated in the Settings page), which only covers 13% of Unicode.

While you are 'technically' correct as per stated here: https://www.unicode.org/versions/stats/charcountv17_0.html, the `uni` library only includes the 'Alphabetics, Symbols' (which is 45K of characters), which is the same as `unidata`

> * `mmazzarolo/unicode-symbols` loads the entire data set JSON into Node.JS RAM, which is not suitable for full search of ~297k Unicode characters if the data source is to be expanded to cover everything.

See previous comment. The 45K dataset is roughly 3.3Mb (because it is heavily compressed)

> * The data source used by `mmazzarolo/unicode-symbols`, [`unidata16` on NPM](https://github.com/j3lte/unidata?rgh-link-date=2025-10-21T16%3A01%3A33Z), omits characters in various blocks (Tangut, Hangul Syllables, CJK Unified Ideographs, just to name a few), rendering it unsuitable for reverse lookup.

This is incorrect, I wrote the `unidata16` library myself, see [here](https://github.com/j3lte/unidata#-character-sets) for all the included datasets. You would need to select the 'Full dataset' in the `unicode-symbols` extension settings to get the full dataset.

> * My extension is written with content and search backed by a bundled binary utility (`uni`), which is significantly different from `mmazzarolo/unicode-symbols`. Porting the existing extension to use `uni` as the backend would require a major rewrite.
> * As my extension is backed by `uni`, it couldn’t be considered backward compatible with `mmazzarolo/unicode-symbols` in aspects like its [addition of Apple logo to the Unicode Dataset](https://github.com/raycast/extensions/blob/e7808f9208c1ab48ffc1eb5a0da7c41c612341ed/extensions/unicode-symbols/scripts/generate-dataset.ts#L77C1-L81C3) and usage of `fuse.js` for fuzzy search.

We can probably make this work. It does require some rewriting

---

- If we would include this in the `unicode-symbols` extension, I would opt for just putting the tgz zipped binaries in the assets folder (instead of downloading them from Github) and unzipping them based on platform/arch on first run. Some people might have spotty internet, firewall issues etc, so it's best to just include them in the assets folder. Based on the [release](https://github.com/blueset/uni/releases/tag/pre-eb04126) (_Nice workflow script by the way!_) I saw we're talking about maybe adding 8Mb of binaries to the assets folder.
- We could back away from dividing the search results in different blocks like we have now
- I believe the `uni` bin doesn't have a way of searching in a specific block, so this might be a bit of a pain. What you could do is then search for something and filter it based on the Block (requires full json output). That achieves the same, shouldn't be an issue
- The 'Identify Unicode' functionality would be completely new.
- The 'exec' functions should probably be rewritten using the [`useExec`](https://developers.raycast.com/utilities/react-hooks/useexec), because it is cleaner, has a builtin timeout and does proper aborting of the execution.

> Final verdict: Yes, we can include this in the `unicode-symbols` extension. If you'd like me to do it, let me know. When I started using Raycast I also built a similar unicode lookup extension. I got the same feedback back then ('We already have an extension'), so I just started contributing to this extension instead.

blueset:
> This is incorrect, I wrote the unidata16 library myself, see [here](https://github.com/j3lte/unidata#-character-sets) for all the included datasets. You would need to select the 'Full dataset' in the unicode-symbols extension settings to get the full dataset.

I apologize for my misunderstanding here. I did not realize that the full dataset is only available in the grid view, while my screenshot above is set to list view, which lead to me incorrectly drew the conclusion above. Sorry for the confusion.

image 

image

> If we would include this in the unicode-symbols extension, I would opt for just putting the tgz zipped binaries in the assets folder (instead of downloading them from GitHub) and unzipping them based on platform/arch on first run.

I initially opted for bundling the binaries with the extension. However, [the guideline](https://developers.raycast.com/basics/prepare-an-extension-for-store#binary-dependencies-and-additional-configuration) discourages bundling opaque binaries “just added to the source folder,” which is why I changed to the download pathway.

> I believe the uni bin doesn't have a way of searching in a specific block, so this might be a bit of a pain. What you could do is then search for something and filter it based on the Block (requires full json output). That achieves the same, shouldn't be an issue

Currently my extension is using full JSON output for all searches in order to get full metadata for each character. However, there was some issue when the keywords are matching too many characters, which is why I added a hard result limit in the output (baked in to bundled binary, or stdout line counting for external ones). I think a better way work around this if we do indeed need block filtering is to modify the forked `uni` binary.

> The 'exec' functions should probably be rewritten using the [useExec](https://developers.raycast.com/utilities/react-hooks/useexec), because it is cleaner, has a builtin timeout and does proper aborting of the execution.

`useExec` is indeed better for local binaries, but I didn’t manage to figure out a way to make it work with WASM binaries. Also, it’s harder to have more granular control with stdout line counting limit on external binaries for performance reasons mentioned above. So I’ve opted to write my own code for these instead.

> Yes, we can include this in the unicode-symbols extension. If you'd like me to do it, let me know. 

I really appreciate your help here. Refactoring an existing extension in this scale could be much harder than writing a new one from scratch. It would be really helpful to have a maintainer knowledgeable of the design and architecture of the existing extension to be able to successfully merge the two extensions in a successful way.

j3lte:
@blueset thanks for the reply! I think the search, using this binary, is way better than fuse, I have had a hard time getting the correct results. Also the 'Identify Unicode' command is a very cool feature.

I believe we should probably just remove the list view in the `unicode-symbols` extension, as it is way less useful than the grid view.

> I initially opted for bundling the binaries with the extension. However, [the guideline](https://developers.raycast.com/basics/prepare-an-extension-for-store#binary-dependencies-and-additional-configuration) discourages bundling opaque binaries “just added to the source folder,” which is why I changed to the download pathway.

Given that we know where they come from (Github), we should probably include this in the README so the source is known. If that's the case, it shouldn't be an issue.

Give me a couple of days and I'll see if I can cook up a PR for `unicode-symbols` that includes your contribution.

@pernielsentikaer I have two questions:

- Is it OK if we include the binaries in the assets folder? We can reference the release page in README, so it is clear where the binaries are coming from. If not we'll go with the automatic download mechanism that is in this PR.
- Given I have the best knowledge on the architecture of this extension (_by now I think I should be the author..._ 😉😜) I can probably easily create a PR for this, but in that case I would want @blueset to get the credits (and points) for it, as he's made a pretty sizeable contribution here. Let me know if that's possible (or tell me how we can make that happen)

j3lte:
Update: Did a bit of testing, but for Mac it will be a problem to include a binary. Although we can run it from NodeJS, an unsigned binary will be automatically blocked by Mac security settings. 

That means I'll probably take the route of including the `wasm` file in assets (and properly run it using `WebAssembly`) and give you an option to point to the `uni` binary if it is available on your machine (given that you can run it on a Mac, you'll have to manually whitelist it in your security settings)

pernielsentikaer:
Ah, so it won't even work like that?

j3lte:
> Ah, so it won't even work like that?

@pernielsentikaer Nope, not for Mac at least. For Mac the only option is to have it installed yourself, run it in the terminal once, unblock the App (in your Security settings), then you can run it. This PR is not going to work for Mac running an unpacked bin that is downloaded from Github, unfortunately. Only ones that are going to work is if the user has installed `uni` himself, or if you choose the `WebAssembly` option

blueset:
In fact I was suprised that the bundled binary worked for me for the first time on macOS. Maybe I have changed the OS security settings some time in the past. The concern of unsigned binary ban was the reason I put WASM as the default option for macOS.

pernielsentikaer:
How heavy is the wasm?

blueset:
@pernielsentikaer The binary size is pretty much the same, but the perceived performance is quite worse using WASM.

blueset:
@pernielsentikaer @j3lte I added an exploratory implementation to invoke `uni` as a Node.js binding. This should have a better performance than WASM and hopefully still not triggering macOS signature check. I didn’t totally replace WASM/Bundled binary options since I can’t get it to compile for Windows ARM64. (Windows x64, macOS Intel, and macOS Apple Silicon are mostly fine though.)

pernielsentikaer:
Do you have any input @j3lte to above comment?

j3lte:
@pernielsentikaer nope sorry. I'm currently very busy with work and some personal stuff, so I don't have time to deep dive into it, sorry

pernielsentikaer:
Wouldn't it be possible to add this to the other extension? They feel pretty similar, right?

blueset:
> Wouldn't it be possible to add this to the other extension? They feel pretty similar, right?

About half of the new extension feature is more or less similar to the existing one, but the fundementals are quite different, especally on the data source. So a major refactor is needed if the two extensions are to be merged.

blueset:
Bump

blueset:
bump

blueset:
Bump

pernielsentikaer:
Sorry for the delay in getting back to you. I'll check a few things and get back to you 🙂

pernielsentikaer:
We should try to get it in as an enhancement to the current version, as @j3lte suggested; he might also be able to guide you. 

The WebAssembly approach might work.

blueset:
BTW that I have [added `uni` as a Node.js native binding](https://github.com/raycast/extensions/pull/22271#issuecomment-3507075155) which should have better performance, but has some minor platform limitations (Windows ARM64, which should be compensated by the bundled executable).

pernielsentikaer:
What do you think about that @j3lte

j3lte:
@pernielsentikaer sounds good. Let @blueset just create an update to the existing extension and we'll take it from there...

pernielsentikaer:
Sounds like a good idea, let's try that

Thanks @j3lte 🙂

blueset:
@j3lte @pernielsentikaer thank you both for the response. As mentioned previously in the thread, refactoring an existing extension in this scale could be much harder than writing a new one from scratch, and would potentially introduce some breaking changes along the way, especially when both extensions at a considerable scale and architectured fundamentally differently.

I’m not really familiar with the old extension. Any guidance on a starting point to begin with would be appreciated. Alternatively, I can also try begin with my extension and try to incorporate features from the old one before we can replace it, in case that may sound more reasonable to you.

j3lte:
@blueset whatever you like to do is fine, however, as stated a couple of times already, the fact that there already is an extension that does this, means there isn't going to be a new one. It would be confusing to users if there are two extensions that basically do the same thing. I get that you want to publish your own extension (been there, done that), but this one isn't one of them.

blueset:
Thank you for your prompt response, and I apologize if my previous message was unclear. I understand Raycast’s stance in extensinos with similar features and I’m not intending to publish this as a separate one. I would hope to get a consensus on the replacement/refactor approach before carrying on with it, as it could be quite some work, and I’d love to avoid a stiuation where the changes to the old extension gets rejected due to a misunderstanding or a dissent with the current maintainers.

pernielsentikaer:
I agree with @j3lte - we should enhance since it's so similar

blueset:
Thanks for the comments, I will find a time soon to begin with first replacing the current extension with new code and then try to incorporate features from the old one as a second stage.

blueset:
There has been some crash issues with my Raycast install on Windows after wakeup from hibernation, will take some time to get back to the extension

pernielsentikaer:
Feel free to write to our support if you need help debugging 🙂

PR #22273, opened 2025-10-18
Add action to generate and copy private links

runofthemillgeek:
## Description

Adds a new command action in "Share Meeting Links" view that will generate a new one-time use private link and copy it to clipboard. The action can also be triggered via Cmd + s shortcut. The request can be cancelled manually or if the same action is triggered on another meeting. Feedback is shown during the generation process via toasts/HUD.

This API seems to be only available in V2, so new Axios client + helpers are added to support the feature.

Other changes:
- Ignore raycast-env.d.ts as per Raycast docs
- Updated package.json to add contributor
- Fix lint error and add publish script

## Screencast

[CleanShot_2025-10-19T02.46.20_CleanShot-CleanShot.webm](https://github.com/user-attachments/assets/b4ebfcd6-c829-43eb-a84c-1e4fe1adc8a5)

## Checklist

### DIFF
--- a/extensions/cal-com-share-meeting-links/CHANGELOG.md
+++ b/extensions/cal-com-share-meeting-links/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Cal.com Share Meeting Links Changelog
 
-## [Generate private links] - {PR_MERGE_DATE}
+## [Generate private links] - 2025-10-28
 
 - Adds an action inside "Share Meeting Link" to generate and copy a one-time use private link (⌘ + S).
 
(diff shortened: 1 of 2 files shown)


pernielsentikaer:
Hi @runofthemillgeek👋 

Did you see the feedback from Greptile, @eluce2 do you want to check this?

runofthemillgeek:
> Did you see the feedback from Greptile, @eluce2 do you want to check this?

@pernielsentikaer yep, that was a good catch from Greptile and I force-pushed the fix adding that back. I was expecting Greptile would run again and verify it's addressed but that didn't happen for some reason.

eluce2:
I don't use cal.com anymore, so I'm probably not the best one to review this extension

pernielsentikaer:
@eluce2, got it, thanks! Do you want to be removed as an author or keep that? You might still get some notifications 🙂 I'm always ready to step in if you want to move to past_contributors.

eluce2:
@pernielsentikaer I don't mind the notifications, but perhaps "past_contributors" makes the most sense? Not sure how that affects all your automations

PR #22359, opened 2025-10-23
Add rhttp extension

SebastianJarsve:
## Checklist


SebastianJarsve:
## Update: TypeScript Error Fixed ✅

The CI error has been resolved - `onCorruption` handlers are now properly async.

---

## Note on Commit Message

This extension is developed in a monorepo, so the squashed commit message includes references to my other projects. However, this PR **only contains `extensions/rhttp/` files** - the commit message is just messy due to shared Git history.

Happy to close and resubmit with cleaner history if preferred, or you can squash with your own message on merge. Let me know!

SebastianJarsve:
Hi @pernielsentikaer 👋
Thanks for the feedback! I understand the concern about overlap.

I actually built this because I looked at the cURL extension and realized I needed something different for my workflow. As a developer working on apps that integrate multiple APIs across different environments, I found myself needing collections to organize endpoints, environment variables to switch between Dev/Staging/Prod, and cookie management for authenticated sessions.

The contribution guidelines say "If your extension works standalone and brings something new to the Store, it's acceptable to create a new one." This extension brings collections, environment management, cookie handling, request chaining, and GraphQL support. These are new workflow capabilities that don't exist in the cURL extension.

The cURL extension is excellent for quick, stateless testing, while this extension is built for managing API projects over time with persistent storage and configuration. Both serve legitimate developer needs.

Happy to adjust the positioning or documentation to make the distinction clearer if that helps. What do you think?

pernielsentikaer:
I know it might be a big one, but is there a reason why we don't add this to the other extension? 

Out store guidelines also says this

CleanShot 2025-11-04 at 13 04 13@2x

They are pretty similar, but I like the additions for sure

SebastianJarsve:
I appreciate you taking the time to review this thoroughly.

I understand the concern about duplication, but I think the architectural differences here are significant enough that merging wouldn't really work well.

The main issue is that the **cURL** extension uses multiple root commands for quick access to different actions, which makes sense for one-off testing. **rhttp** is built around a single entry point with persistent context - you're always "in" a specific collection and environment. Adding _rhttp's_ features to **cURL** would mean either keeping the multi-command structure (which would be confusing - how do you manage which environment you're in across separate commands?) or consolidating everything into one command, which would be a breaking change for existing users.

Beyond that, **cURL** is mostly stateless - it has request history, but that's about it. **rhttp** is built around persistent project state - collections that organize endpoints, environments with variables, stored responses, request chaining, etc. This isn't just adding new features, it would require rewriting the core architecture from "history of past requests" to "organized project management."

I get that the features overlap on the surface (both make HTTP requests), but they're really serving different workflows in my opinion. **cURL** is great for "let me quickly test this endpoint" while **rhttp** is for "I'm managing 50+ endpoints across multiple environments for an ongoing project."

I'm happy to reach out to @jonathangiardino to discuss, but honestly I think merging these would either break what makes **cURL** simple or create something confusing that doesn't serve either use case well.

That said, I respect the team's decision on this. Just wanted to explain why I think the architectural differences make this more complicated than just "adding features to an existing extension."

pernielsentikaer:
I understand and agree with your points. I'll get back to you tomorrow if we haven't heard from Jonathan 🙂

jonathangiardino:
Hey @SebastianJarsve @pernielsentikaer, sorry but I haven't had much time lately to review and maintain cURL updates, @pernielsentikaer has managed most of the latest contributions.

I had a quick read at the thread, and I am totally fine if you want to just create your own extension. I am sure there was a way to integrate the ideas behind rhttp into cURL but it looks like you have already built your own idea and architecture and at this point integrating it sounds indeed like a pain. Perhaps if you built it from the beginning into cURL it would have worked, but now that it's a complete separate app it may not be convenient.

I see the scope of rhttp is way bigger than what cURL was supposed to be, also because my mindset and my idea of a Raycast extension was to either extend an existing SaaS product like many other extensions or keep it simple like cURL.

rhtttp seems to go more into a direction of a Raycast-first complex app, so it's really up to @pernielsentikaer at this point.

SebastianJarsve:
Hey @jonathangiardino, thank you for taking the time to review this despite your busy schedule! I really appreciate your understanding of the architectural differences and your support for rhttp existing as a separate extension.

cURL is great for its simplicity and quick testing workflow - that's exactly what made me realize there was room for a more project-focused tool. Your work definitely inspired parts of rhttp's design.

To be transparent about my vision: I originally built rhttp for my own API development work and have been refining it through real-world use for quite some time before deciding to share it with the community. I'm building it to be a comprehensive API development platform within Raycast - essentially a Postman/Insomnia alternative that's fast and keyboard-driven. I'm planning features like password manager integration, dedicated auth helpers, and OpenAPI import. It's a different scope and philosophy from cURL's intentional simplicity, which is why keeping them separate makes sense architecturally.

If there's ever anything from rhttp that could benefit cURL users, feel free to reach out. Happy to collaborate if it helps the community.

Thanks again! 🙏

PR #22491, opened 2025-10-28
Update zen-browser extension

Zheckan:
## Description

Added new command for Zen Browser "Create New Window". Creates a new empty window

## Checklist

### DIFF
--- a/extensions/zen-browser/CHANGELOG.md
+++ b/extensions/zen-browser/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Zen Changelog
 
-## [Added Create New Window Command] - {PR_MERGE_DATE}
+## [Added Create New Window Command] - 2025-10-29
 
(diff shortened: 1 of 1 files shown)


pernielsentikaer:
Why does it need to activate Raycast in the AppleScript?

Zheckan:
Hi, my mistake. I had an issue with window focus. When the command executed, Raycast momentarily regained focus and the keystroke fired in `Raycast` instead of `Zen`. I initially fixed it by adding `activate Raycast` (following the pattern in existing code). 

However, the real issue was that I was running the script *before* closing the `Raycast` window. I've now switched the order: close `Raycast` first, then run the script. This removes the need for the `activate Raycast` workaround.

Issue was there: https://github.com/Zheckan/zen-browser/blob/7fd83eaf47bbea7094e875d04f444438f22044b2/src/new-window.ts#L4-L7

PR #22498, opened 2025-10-29
[FileMaker Snippets] Fix: Preserve dynamic fields when reloading snippet from clipboard

chriscors:
## Description

This PR fixes a bug in the FileMaker Snippets extension where dynamic fields configured by users would be deleted when reloading snippet XML from the clipboard using Cmd+R.

## Problem

When users pressed **Cmd+R** to reload a snippet's XML content from the clipboard, any dynamic fields they had configured would be lost upon saving. This occurred because the `saveSnippet` function in `edit-snippet.tsx` was creating a new snippet object from form values but wasn't preserving the `dynamicFields` array or `customXML` flag from the original snippet object.

## Solution

Modified the `saveSnippet` function to explicitly preserve:
- `dynamicFields` - The user's configured dynamic field definitions
- `customXML` - The flag indicating custom XML modifications

Both properties now default to appropriate values (`[]` for dynamicFields, `false` for customXML) to ensure backward compatibility.

## Changes Made

- Updated `src/components/edit-snippet.tsx` to preserve `dynamicFields` and `customXML` when saving snippets
- Added proper default values for backward compatibility

## Testing

Users can now:
1. Add dynamic fields to a snippet
2. Press Cmd+R to reload the XML from clipboard
3. Save the snippet
4. Verify that dynamic fields are still present and functional

## Impact

This fix ensures that users can safely update their snippet XML content without losing their dynamic field configurations, improving the overall user experience and preventing data loss.

### DIFF
--- a/extensions/filemaker-snippets/src/components/edit-snippet.tsx
+++ b/extensions/filemaker-snippets/src/components/edit-snippet.tsx
@@ -103,8 +103,8 @@ export function useFormLogic({ onSubmit, snippet }: EditSnippetProps) {
         ...values,
         tags: typeof values.tags === "string" ? values.tags?.split(",") ?? [] : values.tags,
         snippet: snippetText,
-        dynamicFields: snippet.dynamicFields ?? [],
-        customXML: snippet.customXML ?? false,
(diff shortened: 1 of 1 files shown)


chriscors:
@eluce2 for viz

PR #22566, opened 2025-11-01
Add spotify-music-player extension

jumahl:
## Checklist


YourMCGeek:
Out of curiosity, what makes this code warrant a brand new extension over PRing the windows functionality into the existing extension here https://www.raycast.com/mattisssa/spotify-player?

extensions/spotify-music-player/src/helpers/spotify.api.ts
```diff
@@ -0,0 +1,5122 @@
+/**
+ * Spotify Web API with fixes and improvements from sonallux
+ * 2023.2.27
+ * DO NOT MODIFY - This file has been generated using oazapfts.
+ * See https://www.npmjs.com/package/oazapfts
```
YourMCGeek, line 5: These definitions are almost 2 years old. Probably worth updating them since we're modifying the codebase anyway. Latest version is 2025.5.18 https://github.com/sonallux/spotify-web-api/blob/main/fixed-spotify-open-api.yml

jumahl:
> Out of curiosity, what makes this code warrant a brand new extension over PRing the windows functionality into the existing extension here https://www.raycast.com/mattisssa/spotify-player?

This extension is built specifically for Windows using only the Spotify Web API, while the existing one is Mac-focused and relies on AppleScript. Integrating Windows support would require major changes, so a separate extension is simpler and cleaner for Windows users

extensions/spotify-music-player/src/helpers/spotify.api.ts
jumahl: You're right - I'll update to the latest definitions in a follow-up PR to ensure we have the newest Spotify features and better API compatibility

pernielsentikaer:
I might agree with @YourMCGeek that there shouldn't be too much that is only macOS-specific in the end, and the menu bar is not included in Windows anyway - it could be cool to have the same for both.

PR #22670, opened 2025-11-05
Add generate-password extension

shubhamwhocodes:
## Checklist


pernielsentikaer:
Hi @shubhamwhocodes👋 

We have so many password generators on macOS. Maybe you could add support for Windows for one of them instead of creating yet another one? 🙂

shubhamwhocodes:
Hi @pernielsentikaer 👋

I was actually a bit confused about which password generator to pick, so I decided to create my own. Then I thought… why not try submitting it directly to the people!

I totally get your point, so I’ll go ahead and close my pull request. But before I do, I’d love your thoughts on something that do you have any ideas that could make it stand out?

Right now, I’ve kept it super simple and easy to use. For example, users can quickly toggle options with shortcuts like `Cmd + S` to remove symbols, or similar toggles for numbers, characters, and even password length.

Honestly, I was pretty proud of how smooth it turned out… until I realized it’s still just another password generator 😅

I was also thinking of adding something like an API key or secret key generator, but that might be better as a separate extension. What do you think?

pernielsentikaer:
I totally understand that. The problem is that we already have so many generators in the store, and adding even more would create further confusion as the selection of generators would grow 🫠

PR #22678, opened 2025-11-05
Add display-reinitializer extension

JosNun:
## Description

Adds a new "Reinitialize Displays" command that allows you to reinitialize your computer's displays with a variety of different methods. In my particular use case, I have an external monitor that sometimes doesn't wake from sleep, but a quick reinitialization will fix it.

## Screencast

https://github.com/user-attachments/assets/be119487-4976-4137-b7bc-878c41c26e8d

## Checklist


JosNun:
Regarding the binary, it's worth noting that the binary is built via the `compile-swift` script in package.json, which builds it from scripts/DisplayHelper.swift. I did just notice this note about binaries from the "Prepare an Extension for Store" page which has an example for an NPM binary (slightly different, but roughly similar case). Would you prefer I remove the binary from the PR, then you can build it locally to verify that it is what it's supposed to be? 
image

pernielsentikaer:
You should be able to let Raycast build it directly then? See how it's done in e.g. [Raycast Store: Color Picker](https://www.raycast.com/thomas/color-picker)

JosNun:
Oh, nice! I wasn't aware of the swift support. I'll take a look at that, and see if I can do a pass at reworking it.

JosNun:
Reworked!

pernielsentikaer:
That's awesome, thanks for doing it 🙂

pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #22759, opened 2025-11-08
Spotify Player - Add Windows support

AdrianKokot:
## Description

- Added `"Windows"` to platforms
- Added check for not running menu bar commands on Windows
- Changed all shortcuts to support both macOS and Windows

## Checklist

### DIFF
--- a/extensions/spotify-player/CHANGELOG.md
+++ b/extensions/spotify-player/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Spotify Player Changelog
 
-## [Add Windows support] - {PR_MERGE_DATE}
+## [Add Windows support] - 2025-11-14
 
 - Added Windows to platforms
 - Added check for not running menu bar commands on Windows


mathieudutour:
There's no need to gate the menubar command, it just won't show on Windows.

However, the extension uses a bunch of AppleScript which I don't see how it would work on Windows

AdrianKokot:
@mathieudutour 
> There's no need to gate the menubar command, it just won't show on Windows.

Yes, it won't show, but the command to display the menu bar is called when using 'next' or 'previous,' and when using this command without a gate, an error shows up. So this gate is only to not show the error on Windows.

> However, the extension uses a bunch of AppleScript which I don't see how it would work on Windows

The used AppleScript is only a fallback when an API query fails. I haven't found a way to do similar action in PowerShell. Running the applescript command on Windows just does nothing, so I think that it can stay as it is. It just doesn't do anything on Windows, but primary action with Spotify API works.

mathieudutour:
The API only works for premium users, doesn't it? Has it changed? It would be nice to handle it properly maybe

AdrianKokot:
Yes, API is available only for premium users. On Windows there's no way to manipulate app like AppleScript does (in the backgorund)

mathieudutour:
Could you add an explicit error message on Windows when the API fails because the user isn't premium? It could be really frustrating otherwise

AdrianKokot:
I've added PowerShell scripts where it was possible. For two commands I haven't found a way to mimic applescript behaviour, in those scenarios I've added error message.

PR #22833, opened 2025-11-12
Improve regex pattern used for Jira issue key detection

AlbertoCortina:
## Description

I adapted the regex expression used to identify if a string is a Jira issue key. For the regex, I took into account the validation message I could see while creating a project

image

### DIFF
--- a/extensions/jira-search-self-hosted/src/issue.ts
+++ b/extensions/jira-search-self-hosted/src/issue.ts
@@ -59,7 +59,7 @@ function statusIcon(status: IssueStatus): Image {
  *  - Be case-insensitive (e.g., "ac-23" is valid)
  *
  * You can test this pattern here: https://regex101.com/r/dHHMLe/1
- * 
+ *
  * ✅ Valid examples:
  *  - "AC-23"
  *  - "AC2-23"
@@ -68,10 +68,10 @@ function statusIcon(status: IssueStatus): Image {
  *
  * ❌ Invalid examples:
  *  - "A-23"   → Only one character before dash
- *  - "2A-23"  → Does not start with a letter 
- *  - "123-23" → No letters before dash 
- *  - "-23"    → Missing prefix 
- *  - "A_C-23" → Underscore not allowed 
+ *  - "2A-23"  → Does not start with a letter
+ *  - "123-23" → No letters before dash
+ *  - "-23"    → Missing prefix
+ *  - "A_C-23" → Underscore not allowed
  *  - "A-23B"  → Suffix must be numeric
  *
  * @param query - The string to check.


extensions/jira-search-self-hosted/src/issue.ts
```diff
@@ -48,9 +48,35 @@ function statusIcon(status: IssueStatus): Image {
...
+ *  - "A-23B"   // Suffix must be numeric
+ *
+ * @param query - The string to check.
+ * @returns True if the string matches the issue key pattern, otherwise false.
+ */
 function isIssueKey(query: string): boolean {
-  const issueKeyPattern = /^[a-z]+-[0-9]+$/i;
-  return query.match(issueKeyPattern) !== null;
+  const issueKeyPattern = /^[A-Z][A-Z0-9]+-\d+$/i;
+  return issueKeyPattern.test(query);
```
0xdhrv: > **suggestion**

```ts
  const ISSUE_KEY_PATTERN = /^[A-Z][A-Z0-9]{1,}-\d+$/i;
/**
 * Validates whether a string is a properly formatted issue key.
 *
 * A valid issue key must:
 *  - Start with a letter (A–Z)
 *  - Contain at least two characters before the dash (letters or digits)
 *  - Include a dash (-) followed by one or more digits
 *  - Be case-insensitive (e.g., "ac-23" is valid)
 *  - Check on https://regex101.com/r/dHHMLe/1
 *
 * ✅ Valid examples:
 *  - "AC-23"
 *  - "AC2-23"
 *  - "A2C-23"
 *  - "ABC-123"
 *
 * ❌ Invalid examples:
 *  - "A-23"    → Only one character before dash
 *  - "2A-23"   → Does not start with a letter
 *  - "123-23"  → No letters before dash
 *  - "-23"     → Missing prefix
 *  - "A_C-23"  → Underscore not allowed
 *  - "A-23B"   → Suffix must be numeric
 *
 * @param query - The string to validate.
 * @returns True if the string matches the issue key pattern; otherwise false.
 */
const isIssueKey = (query: string): boolean => {
  return ISSUE_KEY_PATTERN.test(query.trim());
};

```

extensions/jira-search-self-hosted/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Jira Search Self Hosted Changelog
 
+## [Update] - {PR_MERGE_DATE}
+
+Improved search by `key` to include numbers in project key.
```
0xdhrv:
```suggestion
- Improved search by `key` to include numbers in project key.
```

extensions/jira-search-self-hosted/src/issue.ts
AlbertoCortina: Thanks! Applied

0xdhrv:
> note

Thanks @AlbertoCortina for the updates

One minor lint issue exists; could you please take a look?

→ https://github.com/raycast/extensions/actions/runs/19297120047/job/55181786119?pr=22833#step:7:450

pernielsentikaer (approved):

PR #22840, opened 2025-11-12
Jira Search (self-hosted) - Remove self from contributors

rnprest:
## Description

This small PR is just to move myself from the `contributors` field of the `jira-search-self-hosted` extension's manifest to the `pastContributors` field ([thanks xmok!](https://github.com/raycast/extensions/pull/22840#pullrequestreview-3453588582)), as I don't have any plans on being a maintainer (as wonderful as this extension is).

### DIFF
--- a/extensions/jira-search-self-hosted/CHANGELOG.md
+++ b/extensions/jira-search-self-hosted/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Jira Search Self Hosted Changelog
 
-## [Moved contributor] - {PR_MERGE_DATE}
+## [Moved contributor] - 2025-12-03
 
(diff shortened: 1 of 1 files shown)


xmok (commented):
If you prefer, you could move yourself into the `pastContributors` ([reference](https://developers.raycast.com/information/manifest#extension-properties:~:text=to%20this%20extension.-,pastContributors,-An%20array%20of)) field instead; that way you should not receive notifications but your username will show up in the Store still - fun little vanity metric 💯.

Screenshot from Jira Extension:

Image

rnprest:
Thanks @xmok , that seems like a better solution 😁 Updated this MR

PR #22883, opened 2025-11-14
Add clickup-tasks extension

snelling-a:
## Checklist


xmok:
Hi @snelling-a:

What does this extension do differently from https://www.raycast.com/aparandeh/manage-clickup-tasks?

snelling-a:
> Hi @snelling-a:
> 
> 
> 
> What does this extension do differently from https://www.raycast.com/aparandeh/manage-clickup-tasks?

@xmok fair question


I tried using the existing extension a couple times over and was not able to do really anything with it, especially with regard to tasks. I could see folders and spaces and other ClickUp stuff but not a single task.

The goal of the other extensions seems to be to replicate the flow on the ClickUp platform. I personally don't find that flow very intuitive and it often requires a lot of navigation to get where you want to go. 

In developing this extension, the primary objective was to 

1. Access tasks as fast as possible
2. display them with relevant information "at a glance" 
3. having a clean UI 
4. Having a UX consistent with other staple extensions like the [GitHub extension](https://www.raycast.com/raycast/github)

I am open to any other questions, comments, or requests for changes if you see room for improvement. I'm not trying to compete with or outshine the existing extension. If you use it and it works for you, then great! It don't fit into my workflow and from what I could see there was still a gap in the Raycast ecosystem to work with ClickUp tasks.

Try it out and let lmk what you think

xmok:
> I am open to any other questions, comments, or requests for changes if you see room for improvement. I'm not trying to compete with or outshine the existing extension. If you use it and it works for you, then great! It don't fit into my workflow and from what I could see there was still a gap in the Raycast ecosystem to work with ClickUp tasks.

It's not about competition but redundancy - would it not be viable to extend the existing extension with your new commands rather than have a completely separate one? They even seem to have the same `Preferences` so should be possible to add your new commands to that one or modify the existing ones to match your new ones.

It would help existing users as well.

[Reference](https://manual.raycast.com/extensions#:~:text=At%20the%20same%20time%2C%20we%20prefer%20to%20keep%20services%20grouped%20in%20one%20extension%20and%20would%20rather%20increment%20an%20existing%20extension%20than%20create%20a%20new%20one.)

snelling-a:
Thanks @xmok for the thoughtful feedback and for pointing me to the manual guidelines. I really appreciate you taking the time to review this, and I understand the consolidation concern - you've contributed significantly to the Raycast ecosystem and have worked on Tasks & Docs Explorer, so your perspective carries weight.

I'd like to make the case for why these extensions serve fundamentally different workflows rather than being duplicates.

## Different Problem Spaces

**ClickUp - Tasks & Docs Explorer** is designed around **exploration and navigation** - it provides a hierarchical drill-down through Teams → Spaces → Folders → Lists → Tasks, plus quick task creation and docs browsing.

**ClickUp Tasks** is designed around **task consumption and daily workflow** - it provides direct access to tasks you need to work on, with "My Tasks" as the primary entry point.

## Concrete Workflow Differences

**Tasks & Docs Explorer:**

- User navigates through organizational hierarchy to reach tasks
- Focused on exploring structure and creating new items
- Best for: discovering what exists, capturing quick ideas, browsing docs

**ClickUp Tasks:**

- User jumps directly to assigned tasks or pre-configured lists
- Focused on viewing, filtering, and acting on existing work
- Best for: daily standups, checking what's on your plate, drilling into subtask hierarchies

## Why This Matters

I evaluated Tasks & Docs Explorer before building this extension. It's well-designed for its intended purpose, but the workflow didn't match my daily needs. The hierarchical navigation is valuable for exploration and planning, but when I'm in flow and need to quickly answer _"What do I need to work on today?"_ - during standup, between meetings, or when context-switching from another task - the navigation overhead breaks that flow.

Tasks & Docs Explorer solves a real problem for users who need to explore structure and capture ideas. ClickUp Tasks solves a different problem: maintaining focus and minimizing context switching for users who primarily need fast access to their assigned work.

## Precedent in the Ecosystem

The Raycast extension store already has multiple extensions for the same service when they serve distinct workflows.

**Jira** (ClickUp's direct competitor) has 5 extensions:

- `jira` (main): Browse projects, issues, boards
- `jira-search`: Quick search optimized for finding issues fast
- `jira-search-self-hosted`: Same search workflow, different deployment
- `jira-time-tracking`: Time logging workflow
- `jira2git`: Git integration workflow

I acknowledge these have clearer functional distinctions (search vs. time-tracking vs. git). However, I'd argue that **workflow optimization** is equally valid as a separation criterion:

- `jira-search` exists because quick search is a different workflow than hierarchical browsing in main `jira`
- **ClickUp Tasks** exists because "My Tasks" consumption is a different workflow than hierarchical exploration in **Tasks & Docs Explorer**

**GitHub** (16 extensions) and **Google** (15+ extensions):

- While these are broader ecosystems, they demonstrate Raycast's acceptance of workflow-specific extensions
- Many are authored/maintained by the core Raycast team, suggesting this approach is intentional design philosophy
- Examples: `github-review-requests` (workflow-specific) vs. `github` (general purpose)

## Regarding Preferences Overlap

On the point about "identical preferences": both extensions naturally require a ClickUp API token (mandated by ClickUp's API), but the preferences are otherwise workflow-specific:

- **ClickUp Tasks** requires a **default list ID** for direct task access
- **Tasks & Docs Explorer** has optional preferences for **team ID** and **space IDs** to support hierarchical navigation

The API token overlap isn't meaningful - any ClickUp extension would require authentication. The other preferences reflect each extension's different approach to navigation.

## Why Not Consolidate?

Adding **Quick Capture** or **Docs Explorer** to ClickUp Tasks would create the exact redundancy we're trying to avoid. Similarly, adding the **My Tasks** view to Tasks & Docs Explorer would bloat it with a workflow that doesn't align with its hierarchical navigation model.

Each extension should stay focused on its core workflow. Users who need exploration and capture can use Tasks & Docs Explorer. Users who need the "My Tasks" workflow can use ClickUp Tasks. Some users might install both for different use cases - and that's okay.

## Alignment with Raycast's Philosophy

Raycast's [core philosophy](https://www.raycast.com/why) emphasizes **never wasting time** and preserving flow state by reducing context switching. This extension embodies that philosophy:

- **Direct access** to "My Tasks" respects user attention - no drilling through Teams → Spaces → Folders → Lists
- **Focused workflow** keeps users in flow rather than forcing exploration when they just want to check their assigned work
- **Do one thing well** - task consumption, not exploration, capture, and docs all in one

Both extensions serve Raycast's mission, just for different moments in the user's workflow. Tasks & Docs Explorer for discovery and planning. ClickUp Tasks for execution and daily focus.

## Request for Raycast Team Decision

I respect that the Raycast team has strong opinions about extension organization, and I defer to your judgment on whether these workflows are distinct enough to warrant separate extensions.

If the decision is to consolidate, I'm willing to collaborate on a refactor. However, I believe keeping them separate provides more value to users with different ClickUp workflows, and I'd appreciate the team's consideration of that perspective.

Thanks for taking the time to review this, and I'm happy to discuss further or make changes based on your guidance.

pernielsentikaer:
We also had a few times where the "new" version kinda superseeded the the original one, we could check with @Ali-Parandeh 🙂

snelling-a:
@pernielsentikaer we're going on 10 business days, wdyt?

pernielsentikaer:
It should indeed be an update. If the current one by @Ali-Parandeh is 100% broken, then we should indeed enhance and fix that instead of leaving it. If it makes sense, this can potentially also supersede the old one, but that's up to @Ali-Parandeh.

PR #22989, opened 2025-11-19
Update punto extension

ruskoloma:
## Description

Hi @red-avtovo, thanks for the extension.
There was a small issue: phonetic layouts were listed as supported, but they use different key mappings compared to the standard ones, so switching didn’t work correctly for them.

In this PR, I added the proper mapping for phonetic layouts so they work as expected.

## Checklist

### DIFF
--- a/extensions/punto/CHANGELOG.md
+++ b/extensions/punto/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Punto Changelog
 
-## [Phonetic layouts support] - {PR_MERGE_DATE}
+## [Phonetic layouts support] - 2025-11-21
 
 - Fixed mapping for phonetic keyboard layouts
 


red-avtovo:
Hey @ruskoloma! Thanks a lot for your contribution! I'm extremely happy to see that the use case for that plugin is growing, and we're adopting more layouts than I could have thought of in the very beginning.
While I don't think that you should reorganize the `Dict.js` file, I really see that we could generate 2 required maps out of a set of layout templates that we store in a lean, unique way. 
That will increase the runtime preparation cost, but will reduce the plugin footprint, and it will decrease the code complexity.

I'm happy with your change and your generalization approach 🫶 
I just think that it needs an additional optimization touch, either within this PR or another one. Feel free to decide, and if you give this idea a pass, then I will do it myself later (no hard feelings 😉  )

As usual, thanks a lot @pernielsentikaer for moderating the old PR #22966 and taking a look at this one ❤️

ruskoloma:
@red-avtovo thanks for the feedback.

Please check out the Dict.ts in the latest commit, what do you think about this approach for generating the maps?
I moved the layouts into strings and generate the maps dynamically instead of hardcoded maps.

pernielsentikaer:
Awaiting @red-avtovo 🙂

red-avtovo:
@ruskoloma the change is solid! I just wonder if you have a classical ukr layout at your disposal so we could cover both languages completely?

ruskoloma:
@red-avtovo, I can't test with a physical Ukrainian keyboard, but the Ukrainian phonetic layout works fine after I made a minor fix for a few symbols.

pernielsentikaer (approved):

PR #22993, opened 2025-11-19
[shell] Added windows support

Sataros221:


extensions/shell/package.json
```diff
@@ -188,15 +201,18 @@
...
     "eslint-config-prettier": "^8.3.0",
     "prettier": "^2.5.1",
-    "typescript": "^4.4.3"
+    "typescript": "^5.9.3"
   },
   "scripts": {
-    "build": "ray build -e dist",
+    "build": "ray build -e dist -o dist",
+    "build:windows": "bunx ray build -e dist -o dist && (node ./scripts/install-on-windows.cjs || bun ./scripts/install-on-windows.cjs)",
+    "install-windows": "node ./scripts/install-on-windows.cjs || bun ./scripts/install-on-windows.cjs",
```
mathieudutour: what's this about?

Sataros221: This is because Raycast for Windows is still in beta and the extension installation directory is different (raycast-x instead of raycast). I understand that this file doesn't affect anything if you download the extension from the store, but for development purposes, it avoids having to manually move the installation.

mathieudutour: You don't need it, just bump the @raycast/api version and it should just work

Sataros221: Oh, I didn't know that could be the cause. So, I'll just remove that along with the script and everything will be fine for the merge?

mathieudutour: I haven't reviewed it yet - but it'll be in a better shape for the review yet :)

PR #23011, opened 2025-11-20
Add annotely extension

codiini:
## Description

This PR adds a new extension for **Annotely** (annotely.com), allowing users to instantly annotate images from their clipboard, local files, or web URLs.

### 1. Annotate Clipboard Image
- Detects if the clipboard contains an image file or raw image data.
- If raw data, it saves it to a temporary file.
- Spins up a temporary local HTTP server to serve the image.
- Opens Annotely with the local URL (e.g., `https://annotely.com/editor?url=http://127.0.0.1:xxxxx/image.png`).
- Automatically cleans up temporary files and closes the server after 60 seconds.

### 2. Annotate Selected File
- Gets the currently selected image file from Finder/File Explorer
- Uses AppleScript to ensure it works even if Finder/File Explorer is not the frontmost application.
- Serves the file via the local server and opens it in Annotely.

### 3. Annotate Web URL
- Provides a simple form to enter an image URL.
- Auto-fills the URL if the clipboard contains a valid link.
- Opens the image directly in Annotely.

## Technical Details
- **Local Server**: To allow the web-based Annotely editor to access local files (clipboard dumps or Finder/File explorer selections), the extension runs a lightweight Node.js `http` server on a random port.
- **Security**: The server only runs for 60 seconds per request and is only accessible via `localhost`.
- **File Support**: Handles various image formats (PNG, JPG, GIF, WebP, BMP) by dynamically setting the correct `Content-Type` headers.

## Screencast

image

## Checklist


extensions/annotely/src/annotate-selected-file.ts
```diff
@@ -0,0 +1,61 @@
...
+      imagePath = await runAppleScript(`
+        tell application "Finder"
+          set theSelection to selection
+          if theSelection is {} then
+            return ""
+          end if
+          set theItem to item 1 of theSelection
+          return POSIX path of (theItem as alias)
+        end tell
+      `);
```
pernielsentikaer: Would be better to handle with try catch instead of runAppleScript

codiini: @pernielsentikaer issue with that is, it adds friction to the UX. On MacOS, Finder needs to be open as the frontmost app else `getSelectedFinderItems` fails.  Since users might trigger this command while another app is in focus, the AppleScript method is more robust for our use case. What do you think?

codiini: Or what if I structured the try/catch this way?:


```
    try {
      const items = await getSelectedFinderItems();
      if (items.length > 0) {
        imagePath = items[0].path;
      }
    } catch {
      if (process.platform === "darwin") {
        // We use AppleScript as a fallback because getSelectedFinderItems throws if Finder isn't frontmost
        imagePath = await runAppleScript(`
        tell application "Finder"
          set theSelection to selection
          if theSelection is {} then
            return ""
          end if
          set theItem to item 1 of theSelection
          return POSIX path of (theItem as alias)
        end tell
      `);
      }
    }
```

So `runAppleScript` is only a backup.

codiini: I've just updated the code to only use `runAppleScript` as a fallback so there's no UX issue for the user.

pernielsentikaer (approved):
Looks good to me, approved 🔥

Let’s get the first iteration out 🚀

PR #23028, opened 2025-11-21
revert(sound-search): delete extension

clins1994:
No one can use it at the moment so I'll submit again when it is actually usable 🙇 

Extension: https://www.raycast.com/soundraw/sound-search


clins1994:
Hmmm ... Nevermind I'll just update it once it is ready

PR #23035, opened 2025-11-21
Update dust-tt extension

adrsimon:
## Description

This PR solves a bug where users would eventually be unable to re-log when logging out. This is due to not clearing local storage on new login which leads to stale data being re-used on next sessions.

We were creating the API with the URL of the region from the previous connection, which led people from the EU region to be stuck on the US region following the migration of the auth flow 2 PRs ago.

## Checklist

### DIFF
--- a/extensions/dust-tt/CHANGELOG.md
+++ b/extensions/dust-tt/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Dust.tt Changelog
 
-## [Fixes] - {PR_MERGE_DATE}
+## [Fixes] - 2025-11-27
 
 - Tighten auth flow for people with multiple accounts
 - Solved a bug where people would log out but being unable to re-log in on the same account


extensions/dust-tt/src/dust_api/oauth.tsx
```diff
@@ -70,7 +71,28 @@ export const provider = new OAuthService({
...
     currentToken = params.token;
 
-    // Use default US region initially, region will be determined when workspace is selected
+    // Check if this is a fresh log in
+    try {
+      const decoded = jwtDecode<{ sub?: string }>(params.token);
+      const currentUserId = decoded.sub;
+      const storedUserId = await LocalStorage.getItem<string>("currentUserId");
+
+      if (currentUserId && currentUserId !== storedUserId) {
```
Fraggle, line 80: Are you sure it's enought ? Why would the stored userId be different if we relocate ?

adrsimon: it would be different if you previously logged in in the raycast extension with another account on another version than this one (the previous version wouldn't have cleared the id), and logged out in the meantime
really edge case but yk i'm afraid of this thing now

extensions/dust-tt/src/dust_api/oauth.tsx
```diff
@@ -92,20 +114,25 @@ export const provider = new OAuthService({
...
-          return cachedUser;
+          return { user: cachedUser, dustAPI };
         }
 
         const r = await dustAPI.me();
         if (r.isErr()) {
-          return undefined;
+          // Handle authentication errors by clearing stale data
+          if (r.error.type === "user_not_found") {
+            await LocalStorage.removeItem("workspaceId");
```
Fraggle, line 128: maybe you want to always clear regardless of the error ?

extensions/dust-tt/src/dust_api/oauth.tsx
adrsimon: what would you envision instead ?

pernielsentikaer: @Fraggle did you see the question from @adrsimon 🙂

adrsimon: @pernielsentikaer sorry we talked IRL and settled that it was good like this ! I previously resolved the comment

PR #23096, opened 2025-11-23
Update linkding extension

jose-elias-alvarez:
## Description

A fix + a few features for the linkding extension:

- Add confirmation modal before deletion, since deletion is destructive and irreversible
- Add archive action (no confirmation since it's reversible)
- Drop `cheerio` dependency and use a simpler regex-based approach, which is faster, reduces extension size, and (possibly) solves a memory issue I've seeing

Regex-based extraction is technically worse than `cheerio`, which parses the entire document, but it's good enough for this case, where we only want to extract 2 tags (almost always located in the first few KB of the document) and don't care if we fail.

## Screencast

N/A, new commands aren't interesting enough to screenshot

## Checklist

### DIFF
--- a/extensions/linkding/CHANGELOG.md
+++ b/extensions/linkding/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Linkding Changelog
 
-## [Refactor + Features] - {PR_MERGE_DATE}
+## [Refactor + Features] - 2025-11-26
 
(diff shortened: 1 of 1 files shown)


pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #23284, opened 2025-12-01
Add reboot-to-bios extension

spix18:
## Checklist

### DIFF
--- a/src/cancel-reboot.ts
+++ b/src/cancel-reboot.ts
@@ -1,0 +1,73 @@
+import { showHUD } from "@raycast/api";
+import { exec } from "child_process";
+import { promisify } from "util";
+import { existsSync, unlinkSync } from "fs";
+import { tmpdir } from "os";
+import { join } from "path";
+
+const execAsync = promisify(exec);
+
+const TASK_NAME = "RaycastCancelReboot";
+const REBOOT_PENDING_FILE = join(tmpdir(), "raycast-reboot-to-bios-pending");
+
+/** Check if a reboot is currently pending */
+function isRebootPending(): boolean {
+  return existsSync(REBOOT_PENDING_FILE);
+}
+
+/** Clear the pending marker so the reboot command's toast knows to update */
+function clearRebootPending(): void {
+  try {
+    if (existsSync(REBOOT_PENDING_FILE)) {
+      unlinkSync(REBOOT_PENDING_FILE);
+    }
+  } catch {
+    // Ignore errors
+  }
+}
+
+/** Execute shutdown /a via Task Scheduler (bypasses Raycast sandbox) */
+async function executeCancelReboot(): Promise<void> {
+  // Clean up any existing task
+  try {
+    await execAsync(`schtasks /delete /tn "${TASK_NAME}" /f`);
+  } catch {
+    // Task didn't exist, that's fine
+  }
+
+  // Create elevated one-time task to abort shutdown
+  await execAsync(
+    `schtasks /create /tn "${TASK_NAME}" /tr "shutdown /a" /sc once /st 00:00 /sd 01/01/2000 /rl HIGHEST /f`,
+  );
+
+  // Run immediately
+  await execAsync(`schtasks /run /tn "${TASK_NAME}"`);
+
+  // Clear the pending marker so the reboot toast detects cancellation
+  clearRebootPending();
+
+  // Clean up task entry after a short delay
+  setTimeout(async () => {
+    try {
+      await execAsync(`schtasks /delete /tn "${TASK_NAME}" /f`);
+    } catch {
+      // Ignore cleanup errors
+    }
+  }, 2000);
+}
+
+/** Cancel any pending reboot to BIOS */
+export default async function main() {
+  // Check if a reboot was scheduled via this extension
+  if (!isRebootPending()) {
+    await showHUD("ℹ️ No scheduled reboot");
+    return;
+  }
+
+  try {
+    await executeCancelReboot();
+    await showHUD("✅ Reboot cancelled");
+  } catch {
+    await showHUD("⚠️ Failed to cancel reboot");
+  }
+}


spix18:
Sorry, I accidentally opened a new PR instead of updating the existing one. :(

extensions/reboot-to-bios/src/reboot-to-bios.ts
```diff
@@ -0,0 +1,68 @@
...
+async function executeReboot() {
+  const batPath = await ensureBatFile();
+
+  // Show success message before executing, as the system will reboot immediately
+  await showHUD("Rebooting to BIOS...");
+
+  // Execute the batch file - ignore any errors as the system will be rebooting
+  // and explorer.exe may not exit cleanly
+  try {
+    await execAsync(`explorer.exe "${batPath}"`);
```
0xdhrv: > **suggestion**

The command `execAsync(\explorer.exe "${batPath}")`: launching the batch by invoking explorer.exe is unconventional. On some Windows setups (especially 64-bit Windows running from 32-bit processes), starting `explorer.exe` from a batch may just open a new File Explorer window rather than properly executing the batch or triggering shutdown.

extensions/reboot-to-bios/src/reboot-to-bios.ts
```diff
@@ -0,0 +1,68 @@
...
+  await writeFile(batPath, batContent);
+
+  return batPath;
+}
+
+async function executeReboot() {
+  const batPath = await ensureBatFile();
+
+  // Show success message before executing, as the system will reboot immediately
+  await showHUD("Rebooting to BIOS...");
```
0xdhrv: > **query**

Even if shutdown starts, the OS may kill explorer.exe (or other processes) before showing the “Rebooting to BIOS” HUD so the HUD might not have time to render or remain visible.

extensions/reboot-to-bios/src/reboot-to-bios.ts
```diff
@@ -0,0 +1,68 @@
...
+  showConfirmation: boolean;
+}
+
+async function ensureBatFile(): Promise<string> {
+  // Use Raycast's support directory instead of a temporary directory.
+  const batPath = join(environment.supportPath, "reboot_to_bios.bat");
+
+  const batContent = `@echo off
+if not DEFINED IS_MINIMIZED set IS_MINIMIZED=1 && start "" /min "%~f0" %* && exit
+shutdown /r /fw /t 0
```
0xdhrv: > **suggestion**

The use of `shutdown /r /fw` may fail depending on OS version or if firmware supports it. In some cases this method does not work; manual method is recommended (Shift+Restart into recovery → UEFI firmware option) as fallback.

Reference:
- https://learn.microsoft.com/en-us/answers/questions/4046070/shutdown-r-fw-not-working

0xdhrv: > **query**

If the user lacks administrative privileges, the batch file may not run or the shutdown command may fail. Consider checking for elevated privileges or catching errors meaningfully.

0xdhrv (changes requested):
> **note**

Fix suggestions and queries

extensions/reboot-to-bios/src/reboot-to-bios.ts
spix18: Thanks you for the review and you're were right, the explorer.exe approach was unreliable because tbh i had no other choice, raycast sandbox does not allow these kind of command to pass so at that time i had and was forced to use explorer.exe approach.
but now i have completly rewritten it and used Windows Task Scheduler instead, please review and give me your opinion, hopefully its a better approach.

typescript
// Create elevated one-time task
await execAsync(
  `schtasks /create /tn "${TASK_NAME}" /tr "shutdown /r /fw /t ${delay}" /sc once /st 00:00 /sd 01/01/2000 /rl HIGHEST /f`,
);
// Run immediately
await execAsync(`schtasks /run /tn "${TASK_NAME}"`);

basically this approach:

1. Bypasses Raycast's sandbox reliably
2. Works on both 32-bit and 64-bit Windows
3. Properly elevates privileges with /rl HIGHEST
4. No batch files or explorer.exe needed
5. Clean task cleanup after execution
and i have completely removed old explorer and bat file approach

extensions/reboot-to-bios/src/reboot-to-bios.ts
spix18: True i have completely missed this

With the new Windows Task Scheduler implementation it should solves this completely:

Configurable delay - The shutdown command now includes the delay: shutdown /r /fw /t ${delay} (default 10 seconds)
Countdown toast - For delays ≥3 seconds, users see a live countdown toast:
Rebooting to BIOS
⏳ 10 seconds remaining...
[Cancel]
Cancel functionality - Users can cancel via:
The toast's Cancel button
A dedicated "Cancel Reboot" command (can be hotkeyed)
Running shutdown /a manually
The toast persists throughout the countdown, giving users clear visual feedback and ample time to react. The old immediate-reboot-with-no-feedback approach has been removed.

extensions/reboot-to-bios/src/reboot-to-bios.ts
spix18: True The /fw flag indeed only works on UEFI systems.

I've added UEFI detection that runs before attempting the reboot:

typescript
async function isUEFIMode(): Promise<boolean> {
  // Check PEFirmwareType registry: 1 = BIOS, 2 = UEFI
  const { stdout } = await execAsync(
    `powershell.exe -Command "(Get-ItemProperty -Path 'HKLM:\\SYSTEM\\CurrentControlSet\\Control' -Name 'PEFirmwareType').PEFirmwareType"`,
  );
  if (stdout.trim() === "2") return true;
  if (stdout.trim() === "1") return false;
  
  // Fallback: check bcdedit for EFI boot path
  // ...
}
If the system isn't UEFI, users now see a clear error message:

"This feature requires UEFI firmware. Your system appears to use legacy BIOS."

The README also documents alternative methods for unsupported systems:

Shift+Restart → Troubleshoot → Advanced options → UEFI Firmware Settings
Boot key (F2, Del, etc.) during startup
Thanks for the Microsoft reference.

spix18: yea i had this on my mind during the dev and i said maybe i should add a check for administractive privileges 

The new task scheduler implementation handles elevation properly(hopefully ?):

typescript
await execAsync(
  `schtasks /create /tn "${TASK_NAME}" /tr "shutdown /r /fw /t ${delay}" /sc once /st 00:00 /sd 01/01/2000 /rl HIGHEST /f`,
);
Key points:

/rl HIGHEST - Task runs with highest available privileges
Error handling: If task creation/execution fails, users see:
"Failed to schedule reboot. Try running 'shutdown /r /fw /t 10' in an elevated Command Prompt."

No UAC prompts - Task Scheduler handles privilege escalation transparently
If the user's account doesn't have permission to create scheduled tasks (Guest accounts or Employee accounts in corporate environments, Kids' accounts on family PCs basically any account created as "Standard User" instead of "Administrator") (very rare for standard users), they get a meaningful error with manual fallback instructions.

spix18:
i'm really sorry again, i made the same mistakes i did before, i will reopen a new fresh PR again if that ok :(

PR #23359, opened 2025-12-02
Add Remove Background extension

NathanTCode:
## Description

Adds the Remove Background extension, which uses AI-powered processing via the Replicate API to instantly remove backgrounds from images, and it does it at a much lower price than the well‑known [remove.bg](http://remove.bg/) API.

## Screencast

remove-background---replicate-api-1 - Initial screen with file picker
remove-background---replicate-api-2 - Image selection and processing
remove-background---replicate-api-3 - Side-by-side comparison
remove-background---replicate-api-4 - Side-by-side comparison (Processed image selected)
remove-background---replicate-api-5 - Action panel options

## Features

- Automatic Finder integration for macOS users to select images
- Manual file picker fallback for all platforms
- Instant visual feedback with side-by-side image comparison (4:3 aspect ratio)
- Multiple output actions: copy processed image (as data URL), copy image URL, paste URL, open in browser
- Secure API token setup with custom onboarding
- Supports PNG and JPG formats

## Checklist


LitoMore:
There is a Replicate extension in the store: https://www.raycast.com/KevinBatdorf/replicate

What do you think about implementing the Remove Background feature in that extension instead?

NathanTCode:
Thank you for the feedback, it's a valid point. I've reviewed the existing Replicate extension, and here's why I recommend keeping Remove Background as a standalone extension.

**Quick Analysis**

- Replicate Extension: Great for exploring AI models, but it's broad and exploratory.
- Remove background with Replicate : Focused on one specific, frequent task  →  fast background removal.

**Why**

**1. User Needs:** Background removal is a common, quick job (e.g., for social media or design). Users want speed, not model browsing.
**2. Simplicity:** Direct keyboard shortcut, auto image detection, instant results. No extra steps.
**3. Ecosystem Fit:** Raycast works best with specialized tools. This isn't duplication, it's complementary.

So yes, I’d still go with a standalone extension. It hits a clear need and keeps the workflow smooth without complicating the Replicate extension.

extensions/remove-background---replicate-api/src/remove-background.tsx
```diff
@@ -0,0 +1,172 @@
...
+                      style: Toast.Style.Failure,
+                      title: "Invalid file type",
+                      message: "Please select a PNG or JPG image.",
+                    });
+                    return;
+                  }
+                  try {
+                    const imageData = fs.readFileSync(filePath);
+                    const base64 = imageData.toString("base64");
+                    const dataURL = `data:image/${path.extname(filePath).slice(1)};base64,${base64}`;
```
0xdhrv: > **note**

For images larger than 15 MB, I think this will break the extension. Can you try it on your local machine?

NathanTCode: Thanks for taking the time to review my PR!

Good catch, I tested it on my side, and it ends up breaking the extension, basically because the whole file gets turned into a base64 data URL, way too big for the API to handle.

I looked into a few solutions:

**1. Upload to a temporary hosting service (URL-based)**
Upload the image to a service like imgbb or file.io, then pass the URL to Replicate instead of the base64 data.
- Pros: Works for any size
- Cons: Requires another API key + privacy concerns

**2. Auto-resize/compress large images**
Use a library like sharp to automatically resize images over a certain size threshold before processing.
- Pros: Seamless UX
- Cons: Adds a native dependency, may reduce quality

**3. Check file size and show a clear error message**
Add a simple file size check and show a helpful toast if the image exceeds the limit.
- Pros: Simple, no extra dependencies
- Cons: User has to resize manually

I’m leaning toward Solution 3. It keeps things simple, doesn’t add another API key, and the extension stays easy to handle. The user gets clear feedback and can resize their image if they need to.

Let me know what you think @0xdhrv.

NathanTCode:
@0xdhrv commited solution 3  →  [698781d](https://github.com/raycast/extensions/pull/23359/changes/698781d6cd4e8ea2e919091074de51722bd4ce91)

**Hard limit set to 10 MB because Base64 inflates the payload by ~33 %.**
-  10 MB → ~13.3 MB string in memory.
- Empirically, 14.6 MB file → 20.4 MB Base64

**With 10MB as the threshold:**
- 10 MB × 1.33 = ~13.3 MB Base64 string

**This gives a safe buffer below the point where Raycast crashes due to:**
1. Memory allocation for the large string
2. JSON serialization when sending to the Replicate API
3. React state handling the data URL

extensions/remove-background---replicate-api/src/remove-background.tsx
0xdhrv: Yeah, this is a good idea. Considering the normal workflow, most users won't have images larger than 10 MB. But if someone does, it will handle that case well. Thanks for the updates.

extensions/remove-background---replicate-api/src/remove-background.tsx
```diff
@@ -0,0 +1,202 @@
...
+  Grid,
+  showToast,
+  Toast,
+} from "@raycast/api";
+import { showFailureToast } from "@raycast/utils";
+import { useEffect, useState } from "react";
+import fs from "fs";
+import path from "path";
+// Max 10MB to account for base64 encoding overhead (~33% increase = ~13MB final size)
+const MAX_FILE_SIZE_MB = 10;
```
0xdhrv: **suggestion**: move to a constants file

For improved readability

extensions/remove-background---replicate-api/src/remove-background.tsx
```diff
@@ -0,0 +1,202 @@
...
+
+  console.log("[DEBUG] Reading file...");
+  const imageData = fs.readFileSync(filePath);
+  console.log("[DEBUG] File read, converting to base64...");
+  const base64 = imageData.toString("base64");
+  console.log("[DEBUG] Base64 length:", base64.length);
+  return `data:image/${ext};base64,${base64}`;
+}
+
+interface Preferences {
```
0xdhrv: **issue**: no need to define preference type

No need to manually set preference types as an interface since it is autogenerated in `raycast-env.d.ts` when you run the extension by `npm run dev`.

Reference:
- https://developers.raycast.com/api-reference/preferences#signature

extensions/remove-background---replicate-api/src/remove-background.tsx
```diff
@@ -0,0 +1,202 @@
...
+  useEffect(() => {
+    async function loadSelectedFile() {
+      try {
+        const selectedItems = await getSelectedFinderItems();
+        if (selectedItems.length === 0) {
+          return; // No toast, will show Form
+        }
+        const file = selectedItems[0];
+        if (!file.path.match(/\.(png|jpg|jpeg)$/i)) {
+          showToast({
```
0xdhrv: **issue**: use @raycast/utils

Replace this with `showFailureToast` 

Reference:
- https://developers.raycast.com/utilities/functions/showfailuretoast

extensions/remove-background---replicate-api/src/remove-background.tsx
```diff
@@ -0,0 +1,202 @@
...
+          });
+          return;
+        }
+        const dataURL = prepareImage(file.path);
+        setOriginalImage(dataURL);
+        processImage(dataURL);
+      } catch (error) {
+        console.log("[DEBUG] Caught error in loadSelectedFile:", error);
+        const message = error instanceof Error ? error.message : String(error);
+        if (message.toLowerCase().includes("finder") && message.toLowerCase().includes("frontmost")) {
```
0xdhrv, line 88: **question**: edge case for window

How this is handled for Windows?

extensions/remove-background---replicate-api/src/remove-background.tsx
```diff
@@ -0,0 +1,202 @@
...
+    }
+    loadSelectedFile();
+  }, []);
+
+  async function processImage(imageDataURL: string) {
+    try {
+      const response = await fetch(`https://api.replicate.com/v1/models/bria/remove-background/predictions`, {
+        method: "POST",
+        headers: {
+          Authorization: `Bearer ${preferences.replicateApiToken}`,
```
0xdhrv, line 111: **question**: more strict input checking

Do you think this needs more error handling for incorrect or null values?

extensions/remove-background---replicate-api/src/remove-background.tsx
```diff
@@ -0,0 +1,202 @@
...
+        <Form
+          actions={
+            <ActionPanel>
+              <Action.SubmitForm
+                title="Process Image"
+                onSubmit={async (values: { file: string[] }) => {
+                  const filePath = values.file[0];
+                  if (!filePath) return;
+                  if (!filePath.match(/\.(png|jpg|jpeg)$/i)) {
+                    showToast({
```
0xdhrv: **issue**: use @raycast/utils

Replace this with `showFailureToast` 

Reference:
- https://developers.raycast.com/utilities/functions/showfailuretoast

extensions/remove-background---replicate-api/src/remove-background.tsx
```diff
@@ -0,0 +1,202 @@
...
+    );
+  }
+
+  return (
+    <Grid columns={2} aspectRatio="4/3" isLoading={!processedImage}>
+      <Grid.Item title="Original" content={originalImage} />
+      <Grid.Item
+        title={processedImage ? "Processed" : "Processing..."}
+        subtitle={processedImage ? undefined : "Please wait while the image is being processed."}
+        content={processedImage || ""}
```
0xdhrv, line 200: **question**: better fallback

Whether `""` is a suitable fallback here?

extensions/remove-background---replicate-api/src/remove-background.tsx
```diff
@@ -0,0 +1,202 @@
...
+        actions={
+          processedImage ? (
+            <ActionPanel>
+              <Action
+                title="Copy Processed Image"
+                onAction={async () => {
```
0xdhrv, line 206: **suggestion**: use function

If you can abstract this logic into a function, it will be more readable

extensions/remove-background---replicate-api/src/remove-background.tsx
```diff
@@ -0,0 +1,202 @@
...
+                    const buffer = Buffer.from(arrayBuffer);
+                    const base64 = buffer.toString("base64");
+                    const mimeType = response.headers.get("content-type") || "image/png";
+                    const dataURL = `data:${mimeType};base64,${base64}`;
+                    await Clipboard.copy(dataURL);
+                  } catch {
+                    showFailureToast("Failed to copy image");
+                  }
+                }}
+                shortcut={{ modifiers: ["cmd"], key: "c" }}
```
0xdhrv: **suggestion**: use common keyboard shortcuts

Using a common keyboard gives a more consistent experience and preserves muscle memory. Since you’ve added support for Windows, it should handle both cases.

Reference:
- https://developers.raycast.com/api-reference/keyboard#keyboard.shortcut.common

extensions/remove-background---replicate-api/src/remove-background.tsx
```diff
@@ -0,0 +1,202 @@
...
+              />
+              <Action.CopyToClipboard title="Copy Processed Image URL" content={processedImage} />
+              <Action.Paste title="Paste Processed Image URL" content={processedImage} />
+              <Action.OpenInBrowser title="Open Processed Image in Browser" url={processedImage} />
```
0xdhrv: **suggestion**: add keyboard shortcut

If you can add a shortcut here as well, it will improve UX for keyboard-heavy users

extensions/remove-background---replicate-api/project.md
0xdhrv, line 1: **nitpick**: duplicate content or PRD commited

Checking whether this is needed since you already have a README.md

extensions/remove-background---replicate-api/README.md
```diff
@@ -0,0 +1,42 @@
+# Remove Background
+
+Remove image backgrounds instantly using AI.
+
+## Setup
```
0xdhrv, line 5: **praise**: thanks for detailed setup

extensions/remove-background---replicate-api/README.md
0xdhrv, line 1: **suggestion**: add file size clarity

If you can add a brief note about file size limitations to the README so users are aware of them.

Reference:
- https://github.com/raycast/extensions/pull/23359#discussion_r2605525188

NathanTCode: @0xdhrv Sorry for missing that. I’ve just opened a new PR with the changes you suggested.

extensions/remove-background---replicate-api/src/remove-background.tsx
NathanTCode: Fixed @0xdhrv

extensions/remove-background---replicate-api/src/remove-background.tsx
NathanTCode: @0xdhrv, honestly, a simple empty block with a small note telling the user to wait felt totally fine to me. The other option is to set an icon, but in a Grid.Item I can't resize it. Therefore, the icon, when used, appears huge and disproportnnal regarding the context of where it is used. The small note under the Grid.Item seems fine.

extensions/remove-background---replicate-api/src/remove-background.tsx
NathanTCode: Fixed @0xdhrv

extensions/remove-background---replicate-api/project.md
NathanTCode: Removed @0xdhrv

extensions/remove-background---replicate-api/src/remove-background.tsx
NathanTCode: Shortcuts added @0xdhrv

extensions/remove-background---replicate-api/src/remove-background.tsx
NathanTCode: Windows support is dropped for now, I just don’t have the time to look into it any further for now. @0xdhrv

pernielsentikaer:
Why can't this work on Windows?

NathanTCode:
@pernielsentikaer It does run on Windows. That one’s on me. I misunderstood a comment from @0xdhrv and thought adding Windows support would take a lot more work. I rolled back the change and added a rule so it switches to FilePicker whenever Finder integration isn’t available on Windows.

pernielsentikaer:
Let's await @0xdhrv then 🙂

extensions/remove-background---replicate-api/package.json
```diff
@@ -0,0 +1,66 @@
...
+  "$schema": "https://www.raycast.com/schemas/extension.json",
+  "name": "remove-background---replicate-api",
+  "title": "Remove Background",
+  "version": "1.0.0",
+  "description": "Remove image backgrounds instantly using AI Powered by Replicate",
+  "icon": "icon.png",
+  "author": "NathanThomassin",
+  "platforms": [
+    "macOS",
+    "Windows"
```
0xdhrv: **issue**: remove Windows from platform

I see you mentioned that you dropped Windows support, so you can remove it from the platform

NathanTCode, line 11: I misunderstood your first comment, and I thought adding Windows support would take a lot more work. I rolled back the change and added "Windows" in package.json, and I added a rule so it switches to FilePicker whenever Finder integration isn’t available on Windows.

0xdhrv (approved):

PR #23417, opened 2025-12-04
Add mite-time-tracking extension

christianarenz:
## Checklist

### DIFF
--- a/extensions/mite-time-tracking/eslint.config.js
+++ b/extensions/mite-time-tracking/eslint.config.js
@@ -1,20 +1,6 @@
-const js = require("@eslint/js");
-const typescript = require("typescript-eslint");
-const raycast = require("@raycast/eslint-plugin");
-const prettier = require("eslint-config-prettier");
-const globals = require("globals");
+const { defineConfig } = require("eslint/config");
+const raycastConfig = require("@raycast/eslint-config");
 
-module.exports = [
-  js.configs.recommended,
-  ...typescript.configs.recommended,
-  {
-    languageOptions: {
-      ecmaVersion: 2022,
-      globals: {
-        ...globals.node,
-      },
-    },
-  },
-  ...(Array.isArray(raycast.configs.recommended) ? raycast.configs.recommended : [raycast.configs.recommended]),
-  prettier,
-];
+module.exports = defineConfig([
+  ...raycastConfig,
+]);


xmok:
Hi @christianarenz:

Is this extension similar to https://www.raycast.com/paolo_ricciuti/mite?
If so, is it possible to combine them?

Best.

christianarenz:
Hi @xmok 
Thanks for your reply.

I also found and tested Paolo Ricciuti’s extension. From my perspective, it seems to be tailored to a workflow that differs a bit from the one commonly used in mite.

Concepts of "Types" with “Primary” and “Secondary,” as well as “Mite Quick Entry” and “Mite Secondary Quick Entry,” don’t exist in mite itself.

In my extension, I've adapted mite's native functions one-to-one. This includes starting and stopping timers for live tracking, and the ability to retrospectively create a time entry after completing a task, even for previous days.

I've also integrated an overview that displays a summary of time entries for the current day and week. Potential issues, such as a missing entry description, are automatically recognized and highlighted. Entries can be edited directly from this overview.

There's also a feature for marking clients and projects as favorites.

Before submitting my application, I read this article: [Contributing to Existing Extensions vs Creating a New One](https://developers.raycast.com/basics/prepare-an-extension-for-store#contributing-to-existing-extensions-vs-creating-a-new-one) 

Without wishing to detract from the valuable and inspiring work on the existing extension, I believe that my extension offers significant differences and more features, and matches the criteria you outlined in the mentioned link under "When you should consider creating a new extension instead of contributing to an existing one."

Best regards.

xmok:
I understand but can't your commands be added to the existing extension as new commands thereby achieving the goals of both extensions in one extension or is there still significant difference in your timers? It's alright if you can ultimately justify it but the fewer similar extensions we have, the better it is from a maintainability perspective. Only trying to understand :)

pernielsentikaer:
I would say the same, this sounds like it should be an enhancement of [Raycast Store: Mite](https://www.raycast.com/paolo_ricciuti/mite) by @paoloricciuti

christianarenz:
Thank you for your feedback @xmok & @pernielsentikaer.

I don't think it makes much sense, as the existing extension looks like it covers a more uncommon workflow. My intention was to seamlessly integrate the native functions of mite and combine them with familiar functions from the mite UI.

Even if Paolo were willing to discuss his own extension, the average mite user would still be left with avoidable overhead, which would compromise clarity and user experience.

Therefore, It seems we won't be able to agree on this. It's a bit of a pity, but thank you very much nonetheless.

paoloricciuti:
Hey I would be totally fine to integrate the commands in the extension...as you've said I kinda tailored the commands to my workflow (which is to often use the same project every day) but while I do think a "Fast entry" is interesting (and I would not remove it from my extension) I don't see the problem in integrating the other commands to adapt to a different workflow.

WDYT?

christianarenz:
Hey @paoloricciuti Thanks for your reply as well.

If you agree, we'd be happy to try it out.

I usually have several different project entries every day, which we generally record live.
That's why I consider the “Start Timer” and “Stop Timer” commands to be important.

Your “Add Today entry” command seems to have the same function as my “Create Entry” command.
However, I only added a field for minutes, as the minutes transfered to the mite API are automatically converted to hh:mm format.

Your current drop-down menu only allows predefined values and not small or odd values like 00:15 or 01:52. That would be a bit inflexible.

I think the “Show summary” command would be useful, especially if you have many entries spread across the day or week.
And the favorites feature is also set via the summary based on a time entry. The project associated with that time entry is marked as a favorite.

There is another important detail in the project/client drop-down list. I have used sections in the drop-down list to group projects by client.

Without this feature, I currently see many identical project names in your extension with my mite account, such as “Stundenkonto” (see image below). This makes it impossible to find the desired projects.



Should I submit PRs for your extension?

Best

paoloricciuti:
Yes please and feel free to reach out if you need any help or if you want to discuss anything.

PR #23423, opened 2025-12-05
Update single-disk-eject extension

LostViking09:
## Description

Added Windows functionality, ~~with the help of [https://www.uwe-sieber.de/drivetools_e.html](https://www.uwe-sieber.de/drivetools_e.html) (bundled)~~ (← removed, deprecated) with a C# helper (compiles at runtime) and some PowerShell scripts.

Removed unused Linux-specific lines.

## Checklist

### DIFF
--- a/extensions/single-disk-eject/CHANGELOG.md
+++ b/extensions/single-disk-eject/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Single Disk Eject Changelog
 
-## [Windows Support] {PR_MERGE_DATE}
+## [Windows Support] 2026-01-09
 
 - Added windows support with a C# helper (compiles at runtime) and PowerShell
 


johndturn (commented):
I've reviewed the code updates, and here are my thoughts: 

- First, thanks for the bumps in deps and to the updated `Toast` API! I really appreciate it. 
- It appears to me that all of the macOS-based code will still work as expected. 
- Third, and most importantly, I don't have a Windows machine to actually pull this and test it on, so I can't verify that this will work. Additionally, I'm less familiar with the Raycast extension contribution guidelines for Windows, so including the binary `.exe` for ejecting a volume (which I can't review in Github) is a bit off-putting

I'm definitely not saying that this is wrong or shady, only that I'm less familiar with whether or not this is standard for plugins that support both macOS and Windows. And, since I can't verify functionality / see the code for the binary, I'm less inclined to accept it. Can someone from the Raycast team weigh in on it?

mathieudutour:
yeah we aren't going to merge a exe - too hard to verify automatically. Isn't it possible to do it with just some powershell?

LostViking09:
There are ways to do this in powershell, but the one that is actually a safe eject, not just an unmount, works differently than the taskbar applet, and often tells the user the drive is still in use.

I completely understand you don't want to merge this third-party tool. Will look into alternative solutions, maybe I can find some solution that has good UX.

LostViking09:
Finally! I found a solution: using a C# helper as an interface for the Windows API. Powershell compiles it on the fly, works like a charm and still transparent for code review and security reasons.

I have also migrated to using wmic as it is somewhat faster than the previous powershell implementation.

johndturn (approved):

LostViking09:
Weird, the Store does not seem to be updated with this update. Changelog only shows the latest bugfix, and the plugin shows up as MacOS only 0.0

pernielsentikaer:
Triggered a new publish to see if it helps 🙂

pernielsentikaer:
@LostViking09 it looks like there is an error with the package(lock) file, can you maybe update it in a PR?

You can see the error here: https://github.com/raycast/extensions/actions/runs/21431223204/job/61710804178

PR #23431, opened 2025-12-05
Added new functionality 

LorryGailius:
## Description

After discussion over the PR https://github.com/raycast/extensions/pull/22039 trying to add [dougfernando/everything-raycast-extension](https://github.com/dougfernando/everything-raycast-extension) to the Raycast Store. A decision is made that it should be merged together with [anastasiuspernat/everything-search](https://github.com/anastasiuspernat/everything-search) already existing implementation. This PR does exactly that, merges Everything search through CLI and makes it the default, the before in-place command to search on ETP/FTP servers can be enabled and configured through the command preferences in the extension settings

## Screencast

screenshot

## Checklist

- [X] I checked that assets used in the `README` are located outside the metadata folder if they were not generated with our [metadata tool](https://developers.raycast.com/basics/prepare-an-extension-for-store#how-to-use-it)


extensions/everything-search/package.json
```diff
@@ -2,58 +2,221 @@
...
-    "@raycast/api": "^1.100.3",
-    "@raycast/utils": "^1.17.0",
-    "basic-ftp": "^5.0.5"
-  },
   "preferences": [
     {
-      "name": "serversJSON",
-      "title": "Everything Servers Configuration",
-      "description": "JSON array of ETP/FTP server configurations",
+      "name": "esExePath",
```
mathieudutour: let's move the preferences the only apply to the new command to the new command

extensions/everything-search/package.json
```diff
@@ -2,58 +2,221 @@
...
+    },
     {
       "name": "evy",
-      "title": "Search Files by Mask",
+      "title": "Search Files on Servers",
       "subtitle": "Everything",
-      "description": "Find Files from Everything by Mask",
+      "description": "Find Files from Everything ETP/FTP servers by Mask",
       "mode": "view",
+      "disabledByDefault": true,
```
mathieudutour: Is there any point of disabling it by default? We tend to do that when there are a lot of commands but here there are only 2 so it's not hard for users to disable it themselves if they want to

mathieudutour:
nice that looks good

extensions/everything-search/package.json
LorryGailius: Figured that if it needs to be configured manually that it should be disabled by default, but just checked that it prompts to configure before launching command. Very nice work from the Raycast team on this!

extensions/everything-search/package.json
LorryGailius: Sure thing

LorryGailius:
How does the source code link work on the Raycast store website? Can it be updated to [dougfernando/everything-raycast-extension](https://github.com/dougfernando/everything-raycast-extension)? Since I'm added as co-author I could update it to be close to current source

mathieudutour:
No, as I mentioned before, the source of truth (and code) is this repo - so the link will point to this repo (and to the specific commit that the version of the extension is built from)

LorryGailius:
> nice that looks good

@mathieudutour What is the status of approval?

PR #23471, opened 2025-12-06
Move nick318 to the past contributors

nick318:
Move nick318 to the past contributors


pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #23472, opened 2025-12-06
Add Órbita extension

tcelestino:
## Description

This extension brings Órbita, the Brazilian Hacker News curated by https://manualdousuario.net/, directly to Raycast.

Displays the latest posts from Órbita's RSS feed in a clean, searchable list interface, allowing users to:
  - Browse the most recent Órbita posts
  - Open posts directly in the browser
  - Copy post links to clipboard
  - Refresh the feed on demand

## Screencast

orbita-1

## Checklist


pernielsentikaer (changes requested):

tcelestino:
> At the moment, Raycast doesn't support localization and only supports US English. Therefore, please avoid introducing your custom way to localize your extension. If the locale might affect functionality (e.g. using the correct unit of measurement), please use the preferences API.

@pernielsentikaer I updated date information using [this](https://github.com/tcelestino/raycast-extensions/blob/main/extensions/github-repository-search/src/utils/index.ts#L11-L13) example

tcelestino:
@pernielsentikaer could you review this PR again? I've implemented your suggestions.

pernielsentikaer:
Hi 👋

I'll be checking this out now 🔥

Thanks for your patience 🙂

pernielsentikaer (changes requested):
The only thing I found is the naming. Can we come up with another or improved one that doesn't use a foreign language?

---

tcelestino:
@pernielsentikaer, in this case, it's the service's proper name. I asked the Órbita owner for suggestions, and he suggested 'Órbita Companion.' But since it's a brand, can't we just keep 'Órbita'?

pernielsentikaer (changes requested):
There is things like `Manual do Usuário` we need to remove at least 🙂

tcelestino:
> There is things like `Manual do Usuário` we need to remove at least 🙂

@pernielsentikaer, can I keep 'Manual do Usuário' in the description, or should I remove it?

pernielsentikaer:
We don't want other languages than English if possible. Could we see if we can minimize it as much as we can?

pernielsentikaer (changes requested):
Manual do Usuário is still not english I suppose 🙂

tcelestino:
> Manual do Usuário is still not english I suppose 🙂

@pernielsentikaer I've now removed any references to 'Manual do Usuário. 😀

pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #23489, opened 2025-12-07
Add simple-renamer extension

chadgsg:
## Description

Just submitted a new Raycast extension for review called Simple Renamer. It lets you batch-rename selected Finder files using powerful modes like find and replace, sequential numbering, case changes, date stamping, and full template control, all with a real-time preview and safety checks before you run it.

## Screencast
simple-renamer-1
simple-renamer-2
simple-renamer-3
simple-renamer-4
simple-renamer-5
simple-renamer-6
simple-renamer-7
simple-renamer-8

## Checklist

### DIFF
--- a/extensions/simple-renamer/src/rename-files.tsx
+++ b/extensions/simple-renamer/src/rename-files.tsx
@@ -183,8 +183,10 @@ export default function RenameFiles() {
       // Only check if it's not the file itself (case change or same file)
       // Note: We use synchronous existsSync for UI responsiveness in the effect loop
       if (!isSameFile && existsSync(newPath)) {
-        // Double check not in our internal set (though internal set tracks *targets*, this tracks *existing on disk*)
-        externalCollisionCount++;
+        // If it exists, check if it's just a case change of the SAME file
+        if (file.path.toLowerCase() !== newPath.toLowerCase()) {
+          externalCollisionCount++;
+        }
       }
 
       return {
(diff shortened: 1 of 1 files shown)


KevinBatdorf:
I reviewed this PR in case it helps! I'm just a fellow extension developer and not on the raycast team.

The code is fine but I have three comments:

1. Use es-toolkit instead of custom formatters. For example, they have a regex that accounts for additional edge cases ([see here](https://github.com/toss/es-toolkit/blob/main/src/string/words.ts#L18)) such as emojis, which are supported in filenames. This also makes updating it 5-10 years later to cover new stuff (assuming they are likely to maintain support)
2. For the dates, use `replaceAll` in case the user wants to put the date in there twice. You could also use `date-fns` if you wanted to support more features and edge cases (but what you have is fine)
3. I wanted to append a date in the custom template, so I could do `image-## (YYYY-MM-DD)`, but it doesn't look possible. This could be a future feature though. I don't think it's a blocker here.

Otherwise I tried it out on multiple files and it handled them fine. Cool extension

chadgsg:
### Fixed
- Fixed issue where case-only renames (e.g. `file` -> `FILE`) failed on case-insensitive file systems (macOS/Windows).
- Improved error handling and validation.

0xdhrv:

We already have an extension in the Store that provide File Renaming functionality. Could we consider enhancing the existing extension below instead of creating another one?

- https://www.raycast.com/theo_daguier/renaming by @theo_daguier

You can enhance the existing extension by adding support for Windows and add more features.

PR #23494, opened 2025-12-08
Remove self from COEOWNERS

peppy:
I don't have the time to handle these notifications so they are just noise to me. Requesting self removal 🙏.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -758,7 +758,7 @@
 /extensions/git-worktrees                                           @philstainer
 /extensions/gitee                                                   @koinzhang
 /extensions/gitfox                                                  @azlekov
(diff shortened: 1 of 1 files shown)


pernielsentikaer:
The `CODEOWNERS` file is autogenerated. Removing yourself from that file will simply add it back again.

You need to change `package.json` in the extension, where you can either remove yourself completely or move yourself to `pastContributors`, which doesn't get notifications 🙂

pernielsentikaer:
@peppy are. you still around?

peppy:
I am yes, will adjust the change soon, thanks!

peppy:
Superseded by https://github.com/raycast/extensions/pull/23548.

PR #23545, opened 2025-12-09
Update imessage-2fa extension

eluce2:
Continuation of [this PR](https://github.com/raycast/extensions/pull/22496)

### DIFF
--- a/extensions/imessage-2fa/CHANGELOG.md
+++ b/extensions/imessage-2fa/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Messages 2FA Changelog
 
-## [Update] - {PR_MERGE_DATE}
+## [Update] - 2025-12-17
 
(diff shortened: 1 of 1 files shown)


clins1994 (changes requested):
@eluce2 you removed various comments related to the Japanese specific support. Can you make sure your PR adds changes without deleting existing ones (comments included)?

(Relates to the greptile comments)

eluce2:
I don't want to become an active maintainer of this extension. I had a problem and was somehow able to vibe-code a fix that works for me on my computer, but I don't fully understand all the moving parts of this extension other than I got all the tests to pass.

I know that other users have reported the issues that I was trying to fix, but if another maintainer can take over or submit their own PR to fix them, that would be great. I thought I had given write-access to maintainers, but perhaps that only applies to the Raycast team. If that's a blocker I am happy to give write-access to my branch so this fix can get merged.

PR #23612, opened 2025-12-12
Update forked-extensions extension

j3lte:
## Description

- Add support for opening an extension folder with a default application

## Checklist

### DIFF
--- a/extensions/forked-extensions/CHANGELOG.md
+++ b/extensions/forked-extensions/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Raycast Fork Extensions Changelog
 
-## [Improvements] - {PR_MERGE_DATE}
+## [Improvements] - 2025-12-17
 
 - Add support for opening an extension folder with a default application
 


extensions/forked-extensions/src/manage-forked-extensions.tsx
```diff
@@ -106,6 +106,14 @@ function ManageForkedExtensions() {
             <ActionPanel>
               <ActionPanel.Section>
                 <Action icon={Icon.Eye} title="Show Details" onAction={() => setIsShowingDetail(!isShowingDetail)} />
+                {openWith ? (
+                  <Action.Open
+                    application={openWith.bundleId}
```
yusifaliyevpro: bundleIds are undefined on Windows. You can use windowsAppId instead 
 ```ts
 application={isMac ? openWith.bundleId : openWith.windowsAppId}
 ```

j3lte: I'm just going to refer the `openWith` itself, that's an `Application` which should work. Thanks for the comment 👍

yusifaliyevpro: Yep, it will also work

pernielsentikaer:
Thanks for the contribution @j3lte - @LitoMore do you want to check this?

pernielsentikaer (commented):
Thanks for the update @j3lte 

@LitoMore do you want to check this?

kmk142789 (approved):

LitoMore:
image

@j3lte This may break the experience of existing users. Would you mind moving the new action below the OpenWith action and assigning the <kbd>Shift + Enter</kbd> shortcut to it?

j3lte:
@LitoMore this shouldn't break current user experience, as the default doesn't show up if you don't set it. I would assume, from a UX perspective, that someone would prefer to have this as the first option instead of the 'Open With'.

Adding a shortcut key is something I will add, but I recommend keeping the sorting I propose here... (Open With would become Shift + Enter, only when someone has selected a default app) 

I'll update the PR later today.

LitoMore:
~I personally prefer to use <kbd>Command + Enter</kbd> for the OpenWith action. Because I don't always open them with an editor, I also frequently use the terminal and GitHub Desktop.~

Your solution sounds good to me. I can keep the new option empty so that I can keep using <kbd>Command + Enter</kbd>.

pernielsentikaer:
So we're good to go @LitoMore, can you try to approve the PR?

j3lte:
@pernielsentikaer a moment please, I still have to add the extra keyboard shortcut that is missing, then it can be approved

j3lte:
@pernielsentikaer @LitoMore latest commit should have added the `Shift` + `Enter` for "Open With" when a default app is selected. This does not interfere with the existing users that don't have a default selected.

extensions/forked-extensions/src/manage-forked-extensions.tsx
```diff
@@ -18,7 +18,7 @@ import { catchError, handleError } from "./errors.js";
 import * as git from "./git.js";
 import operation from "./operation.js";
 import { ForkedExtension } from "./types.js";
-import { extensionLink, getActualIconPath, userLink } from "./utils.js";
+import { extensionLink, getActualIconPath, userLink, openWith } from "./utils.js";
```
LitoMore:
```suggestion
import { extensionLink, getActualIconPath, openWith, userLink } from "./utils.js";
```

It's a bit of a nitpick. But I have a little bit of OCD on it 😄. Let's get these imports ordered alphabetically.

j3lte: @LitoMore haha you and me both, good catch. I usually implement the [trivago/prettier-plugin-sort-imports](https://www.npmjs.com/package/@trivago/prettier-plugin-sort-imports) in my extensions so I don't have to think about sorting the imports :)

pernielsentikaer (approved):
Looks good to me, approved 🔥

yusifaliyevpro:
Hi @j3lte , The feature does not work on Windows.

image


https://github.com/user-attachments/assets/e24e1801-1d46-49d8-b4a1-2e1f41f7e229

j3lte:
@yusifaliyevpro I believe they broke something in the latest version of Raycast, because the "Open With" (the other command that was there and I didn't touch) also opens the folder in explorer instead of the app....

j3lte:
@yusifaliyevpro I notified Raycast, it is a bug in v0.40.0.0 of Raycast on Windows, I debugged the outcome of the apppicker...

yusifaliyevpro:
You are right. Even "Open With" does not work. I will also notify the team on Slack.

image

PR #23644, opened 2025-12-14
[Visual Studio Code] Fix Windows side Issues

yusifaliyevpro:
## Description

The previous "Windows Support" [PR](https://github.com/raycast/extensions/pull/20940) by @tonka3000 had some issues. I have used some codes from this [PR](https://github.com/raycast/extensions/pull/21022) by @vensauro. 

- Resolved `setLocale` warnings in bash by clearing environment variables.
- Updated commands to support Windows.
- Improved error handling and user feedback in various components.
- Updated dependencies for better compatibility and performance.
- Migrated Eslint configuration file

## Screencast

image

## Checklist

### DIFF
--- a/extensions/visual-studio-code-recent-projects/CHANGELOG.md
+++ b/extensions/visual-studio-code-recent-projects/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Visual Studio Code Changelog
 
-## [Fix: Windows Support] - {PR_MERGE_DATE}
+## [Fix: Windows Support] - 2026-01-08
 
 - Resolved `setLocale` warnings in bash by clearing environment variables.
 - "Commands" command now supports Windows.


yusifaliyevpro:
@greptile-apps Eslint works as expected. runExec and Open with Vscode command works as expected. Clearing env doesn't affect PATH or HOME. Please review again

extensions/visual-studio-code-recent-projects/src/open-with-vscode.ts
```diff
@@ -22,25 +25,53 @@ const getSelectedPathFinderItems = async () => {
...
-      await open(currentPath, bundleIdentifier);
-    } else {
       for (const item of selectedItems) {
         await open(item.path, bundleIdentifier);
       }
     }
 
+    if (isWin) {
+      const cliFilename = getVSCodeCLIFilename();
+      const paths = await getSelectedFileExplorerItems();
```
mathieudutour: you shouldn't need this, `getSelectedFinderItems` also works on Windows. If it doesn't, then it's a bug in the API and we'll fix it but let's not introduce this

extensions/visual-studio-code-recent-projects/src/lib/vscode.ts
```diff
@@ -120,6 +121,14 @@ export class VSCodeCLI {
   uninstallExtensionByIDSync(id: string) {
     child_process.execFileSync(this.cliFilename, ["--uninstall-extension", id, "--force"]);
   }
+
+  async newWindow() {
+    if (isWin) {
+      await runPowerShellScript(`
+        Start-Process -FilePath "${this.cliFilename}" -ArgumentList "--new-window" -WindowStyle Hidden
```
mathieudutour: can't this just be
```
child_process.execFileSync(this.cliFilename, ["--new-window"])
```
? 
It also works on macOS

yusifaliyevpro: I updated the code. child_process.execFileSync produced an error, so I replaced it with our runExec helper. This also prevents Raycast’s environment from interfering with VSCode’s integrated terminal.

But I didn't change open-new-window.ts, macOS uses a different script to open a new window, and I don't want to modify the macOS behavior.

extensions/visual-studio-code-recent-projects/src/open-with-vscode.ts
yusifaliyevpro: I updated the code. `getSelectedFinderItems` didn’t work before, so I thought it will not work again, that's why I never tried. But it seems to work now. I removed the custom `getSelectedFileExplorerItems` function. Please review it again.

extensions/visual-studio-code-recent-projects/src/lib/vscode.ts
yusifaliyevpro: Found the issue with child_process.execFileSync. On Windows, file paths can contain spaces, which this function does not handle correctly in this case. Additionally, .cmd files are not executable directly on Windows, so the shell option must be enabled when running on that platform. Also `Show Installed Extensions` and `Install Extension` commands now work on Windows correctly.

```ts
  constructor(cliFilename: string) {
    this.cliFilename = `"${cliFilename}"`;
  }
  
  child_process.execFileSync(this.cliFilename, ["--new-window"], { shell: isWin });
  ```

Comment (yusifaliyevpro) [reply extensions/visual-studio-code-recent-projects/src/lib/vscode.ts]: Can you test on MacOS side?

**Comment by yusifaliyevpro** [comment]
May I ask who are you @kmk142789 ? I have never seen you before, r you a new raycast team member? I didn't saw raycast_bot mentioning or assigning you on this and other PRs.

**Comment by yusifaliyevpro** [comment]
Hi @mathieudutour , I think this must be handled in Raycast side. Raycast overrides and deletes some environment variables in a way that vscode integrated terminal behaves differently. It would be great to have an access to system environment variables or Raycast not to override and delete some variables from `process.env`. 

```ts
export function runExec(commands: string[], onFinish: (error: string | null) => void) {
  const cmd = commands.map((c) => `"${c}"`).join(" ");
  const cleanEnv = { ...process.env };
  delete cleanEnv.LC_ALL;
  delete cleanEnv.LANG;
  delete cleanEnv.LANGUAGE;
  exec(cmd, { env: cleanEnv }, (error, stdout, stderr) => {
    if (stdout || stderr) {
      console.log("fix me");
    }
    if (error) {
      onFinish(error.message);
    } else {
      onFinish(null);
    }
  });
}
```



## Just normal vscode integrated terminal:
image

## When launching VS Code with the extension (custom script)

image

mathieudutour:
do you have to pass the `env`? Obviously Raycast is going to set some env variables and they are subject to change. I don't think you should rely on its values

yusifaliyevpro:
@mathieudutour , If I don't pass env manually without deleting LC_ALL, LANG

This warnings show up in terminal.

yusifaliyevpro:
The exec functions automatically inherits Raycast env, even if I don't give env option

extensions/visual-studio-code-recent-projects/src/lib/vscode.ts
```diff
@@ -111,14 +111,30 @@ export function getVSCodeCLIFilename(): string {
...
+    delete this.cleanEnv.LANG;
+    delete this.cleanEnv.LANGUAGE;
   }
+
   installExtensionByIDSync(id: string) {
-    child_process.execFileSync(this.cliFilename, ["--install-extension", id, "--force"]);
+    child_process.execFileSync(this.cliFilename, ["--install-extension", id, "--force"], {
+      shell: isWin,
+      env: this.cleanEnv,
+    });
```
mathieudutour: Ah I see, then let's only do it for windows, it seems like it's working fine on macos
```
child_process.execFileSync(this.cliFilename, ["--install-extension", id, "--force"], isWin ? {
      shell: true,
      env: this.cleanEnv,
    } : undefined);
```

yusifaliyevpro: I commited a fix. Should work now. Can you review again?

yusifaliyevpro: Btw, I have resolved env problem as you can see below. 

image

yusifaliyevpro: Only two variables (PATHEXT and PROMPT) are different but I don't think they have any impact.

yusifaliyevpro: @mathieudutour , Found that this implementation doesn't work either. AI chatbots says that there is no possible way to block inheriting from Parent processes.

Normaly when you open VSCode using app icon:
```
winlogon.exe
  └─ userinit.exe
       └─ explorer.exe (which combines all system variables)
            └─ Code.exe (about 92 system variables in my case)
```

If extension is in production (installed or imported or quited dev server) The env is inherited like that.
```
Raycast.exe
  → Node
      → Code.exe (about 53 variables in my case which doesn't have git.path, PNPM_HOME)
```

If I am in Development Mode, the final Code.exe env is inherited like that which works quite normal except (PATHEXT and PROMPT variables):
```
explorer.exe
  └─ Code.exe                         (Raycast extension project opened normally)
       └─ bash.exe                    (VS Code integrated terminal)
            └─ npm                    (npm run dev — raycast dev server)
                 └─ node.exe          (Raycast extension dev server)
                      └─ Raycast.exe
                           └─ node.exe (raycast node)
                                └─ Code.exe (about 92 system variables, works with no impact)
```

Problem is that production does not work. How you managed to avoid this behaiviour in open() function from @raycast/api?

yusifaliyevpro:
@raycastbot Remove stalled label

extensions/visual-studio-code-recent-projects/src/lib/vscode.ts
```diff
@@ -65,60 +67,81 @@ function getNLSVariable(text: string | undefined): string | undefined {
...
+      Windsurf: "/Applications/Windsurf.app/Contents/Resources/app/bin/windsurf",
+    };
+  }
+
+  return cliPaths;
+}
 
 export function getVSCodeCLIFilename(): string {
   if (isWin) {
+    const cliPathsWindows = cliPaths();
```
mathieudutour: this is platform-agnostic now, right? So you should be able to remove the `if (isWin)` branch

extensions/visual-studio-code-recent-projects/src/index.tsx
```diff
@@ -211,8 +204,11 @@ function LocalItem(
           end tell
         end tell
         `);
+        }
       }
-      open(props.uri, bundleIdentifier);
+
+      if (isWin) open(path, editorApp);
+      else open(props.uri, editorApp);
```
mathieudutour:
```suggestion
      open(isWin ? path : props.uri, editorApp);
```

extensions/visual-studio-code-recent-projects/src/lib/vscode.ts
yusifaliyevpro: Yes, good catch

PR #23675, opened 2025-12-15
Add dns-lookup extension

valent1d:
## Checklist


pernielsentikaer:
Hi 👋

Would it make sense to show data in the right side: https://developers.raycast.com/api-reference/user-interface/list#list.item.detail

What about it default shows "all" isntead of just one type?

pernielsentikaer:

valent1d:
Hi @pernielsentikaer 😀

I will look into your recommendation regarding the use of the right side for data and how to implement it with a UX-first approach.

Concerning your second recommendation, are you suggesting displaying all record types at once (A, MX, NS, etc.)? Do you think this would improve the module's usability, or should it perhaps be an option?

In my specific use case, for example, I want to check if a TXT validation (like an ACME challenge) has propagated correctly. In that situation, seeing the CNAME or NS for that record isn't necessarily useful since it is strictly a TXT entry.

That being said, I am open to other perspectives if you see a better UX for this module.

Thanks in advance.

pernielsentikaer:
I guess you're right; we can keep it as is. I'll take another look 🙂

pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #23680, opened 2025-12-15
[Downloads Manager] Add Windows Support

YourMCGeek:
## Description

Follow up of #23395, cc @mathieudutour 

## Checklist

### DIFF
--- a/extensions/downloads-manager/CHANGELOG.md
+++ b/extensions/downloads-manager/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Downloads Manager Changelog
 
-## [Add Windows support] - {PR_MERGE_DATE}
+## [Add Windows support] - 2025-12-19
 
 - Added support for Windows
 


0xdhrv:
> ### Additional Comments (2)
> 1. `extensions/downloads-manager/src/utils.tsx`, line 194 ([link](/raycast/extensions/blob/4c01682d324938648df2ff7539ccf4a0d287ce05/extensions/downloads-manager/src/utils.tsx#L194))
>    **style:** This `accessSync` call is redundant since `hasAccessToDownloadsFolder()` already performs the same check on line 149.
> 2. `extensions/downloads-manager/src/utils.tsx`, line 26-30 ([link](/raycast/extensions/blob/4c01682d324938648df2ff7539ccf4a0d287ce05/extensions/downloads-manager/src/utils.tsx#L26-L30))
>    **logic:** The preference check logic doesn't handle the default value correctly. When `downloadsFolder` is set to the default `"~/Downloads"` in `package.json`, it will always use the preference instead of attempting detection. This means the custom detection for macOS/Windows will never run for users who haven't changed the preference.
>    Consider checking if the preference equals the default value before skipping detection, or update the logic to prioritize detection over the default preference value.
> 
> 4 files reviewed, 2 comments
> 
> [Edit Code Review Agent Settings](https://app.greptile.com/review/github) | [Greptile](https://greptile.com?utm_source=greptile_expert&utm_medium=github&utm_campaign=code_reviews&utm_content=extensions_23680)

@YourMCGeek can you check these comments?

pernielsentikaer:
Hi @YourMCGeek 👋

Can you check comments from Greptile 🙂

YourMCGeek:
Went ahead and removed the redundant call per Greptile's feedback. 

Regarding the preference check logic, I simply removed the default value and changed it to not be required. This means that it'll auto detect the folder, then fall back to the cache detection unless the preference is manually made -- I don't think there's a way to programmatically set the preference and I don't believe this should break anyone's current workflow. 

Let me know if there are concerns, or if I've missed something. Thanks y'all.

extensions/downloads-manager/src/utils.tsx
```diff
@@ -1,19 +1,86 @@
...
+      const result = execSync(
+        `osascript -e 'tell application "Finder" to return POSIX path of (path to downloads folder)'`,
+        { encoding: "utf-8" },
+      );
+      return result.trim();
+    } catch (error) {
+      // Fallback to default location if query fails
+      console.error("Failed to get Downloads folder from Finder:", error);
+      return untildify("~/Downloads");
+    }
```
mathieudutour:
```suggestion
    return untildify("~/Downloads");
```

let's just use `~/Downloads` on macos, there are no other path

mathieudutour (approved):

PR #23762, opened 2025-12-18
Adds a `Move latest download` command to Downloads Manager extension

rmehner:
## Description

My workflow often consists of downloading a bunch of invoices and then needing to move them to a particular folder in Finder. This eases this flow a lot.

## Checklist

### DIFF
--- a/extensions/downloads-manager/CHANGELOG.md
+++ b/extensions/downloads-manager/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Downloads Manager Changelog
 
-## [Add move last download item command] - 2025-12-18
+## [Add move last download item command] - {PR_MERGE_DATE}
 
 - Added the command `Move Latest Download`, which moves the latest download to the foremost active app.
 


clins1994:
@rmehner thanks for the contribution. there's some things to consider here since you are "pasting" and then "deleting" (via trash or permanently) ... What you want to do is "CUT" + "PASTE" behavior 

| Concern | Description |
|--------|--------|
| 1 | Is it only for Finder ? What happens if you paste to a frontmost application that doesn't do anything with it. Should it really delete the file or move to trash? |
| 2 | You are showing toast for both the deletion and the move 'success' which show kind of at once, not good  |
| 3 | Simulating "CUT" + " PASTE" behavior isn't in Raycast's [System Utilities](https://developers.raycast.com/api-reference/utilities). The delete util also moves things to trash, do we really want to move to trash when performing a CUT + PASTE. Also do we really want to permanently delete? Similar to the issue mentioned in 1 we kind of need to back up the data before destroying it and understand how to revert if something goes wrong |

Since we already have the "Paste Latest Download" and "Delete Latest Download" may I suggest a different approach? How about adding arguments to these commands so you can

"Paste Latest X Downloads" or "Delete Latest X Downloads"

Then instead of you running let's say 10 "Move Latest Download" commands which would be running 10 x "Paste Latest Download" + 10 x "Delete Latest Download" (20 commands) you could actually get by with running 2 commands

1 Paste Latest Download with argument "10"
1 Delete Latest Download with argument "10"

What do you think?

Of course it would still be nice to have a Move Latest Download with argument so you would decrease from 2 commands to 1. But you definitely need to add some kind of revert functionality 🤔

pernielsentikaer:
Thanks for the feedback @clins1994, could you look into that @rmehner 

clins1994:
@rmehner for what it is worth: https://github.com/raycast/extensions/pull/23770

rmehner:
Thanks for all the feedback. Since there's no easy and reliable way to detect if the app has accepted the new file and the potential for data loss is real, I'll close this and use another way to fit my need.

PR #23774, opened 2025-12-18
Remove sound-search extension

clins1994:
It has been unusable since the release so I'm removing it. I'll work on it properly when I have more time 😿


pernielsentikaer (commented):
Hi @clins1994 

Can you maybe do something like this extension: https://github.com/raycast/extensions/tree/main/extensions/associated-press

We need to make something like that to make sure that as many as possible gets that before we remove it from the store, then in 1/2 months, then we can remove it from the the store 🙂

clins1994:
@pernielsentikaer i see. i'll make a separate pr

PR #23780, opened 2025-12-18
Add color-contrast extension

alexbourt-bxl:
## Checklist

### DIFF
--- a/extensions/color-contrast/CHANGELOG.md
+++ b/extensions/color-contrast/CHANGELOG.md
@@ -1,5 +1,5 @@
 # Color Contrast Changelog
 
-## [Added Color Contrast] - 2025-12-16
+## [Added Color Contrast] - {PR_MERGE_DATE}
 
 Initial code


alexbourt-bxl:
I’ve updated CHANGELOG.md to use {PR_MERGE_DATE} as indicated.
Let me know if anything else is needed from my side 🙌

alexbourt-bxl:
> We already have an extension in the Store that provide Color Contrast functionality. Could we consider enhancing the existing extension below instead of creating another one?
> 
> * https://www.raycast.com/lekoarts/polished by @LekoArts

Thanks for the suggestion — I did look at the existing extension.

My goal here is a different workflow and a narrower scope: a two-click, pixel-based contrast check (pick foreground + background directly from the screen) and get an immediate WCAG/APCA result.

The extension you linked is broader and requires manually entering/pasting values, which is great for its use case, but it’s not the quick “sample from UI” flow I’m targeting.

Also, this extension is intended to work on macOS and Windows, whereas the other one appears to be macOS-only.

Happy to clarify the overlap/positioning in the README if that helps.

0xdhrv:
> > We already have an extension in the Store that provide Color Contrast functionality. Could we consider enhancing the existing extension below instead of creating another one?
> > 
> > * [raycast.com/lekoarts/polished](https://www.raycast.com/lekoarts/polished) by @LekoArts
> 
> Thanks for the suggestion — I did look at the existing extension.
> 
> My goal here is a different workflow and a narrower scope: a two-click, pixel-based contrast check (pick foreground + background directly from the screen) and get an immediate WCAG/APCA result.
> 
> The extension you linked is broader and requires manually entering/pasting values, which is great for its use case, but it’s not the quick “sample from UI” flow I’m targeting.
> 
> Also, this extension is intended to work on macOS and Windows, whereas the other one appears to be macOS-only.
> 
> Happy to clarify the overlap/positioning in the README if that helps.

Apologies for the late reply. I understand your use case and the need for a quick workflow. Since these are additions to the existing extensions without direct overlap, you can enhance the current extension. For example, add a new command for quick contract checks and add support for Windows. The idea is to make sure we have more discoverability and less duplication.

PR #23806, opened 2025-12-19
Add tui-type extension

MrDoekje:
## Description

A beautiful, customizable, (important-)feature-equivalent MonkeyType inspired experience.

## Screencast

tui-type-1-1

tui-type-6-1

## Checklist

### DIFF
--- a/extensions/tui-type/package.json
+++ b/extensions/tui-type/package.json
@@ -39,22 +39,18 @@
     }
   ],
   "dependencies": {
-    "@raycast/api": "^1.79.0",
-    "@raycast/utils": "^1.16.0",
+    "@raycast/api": "^1.104.1",
+    "@raycast/utils": "^2.2.2",
     "zustand": "^5.0.9"
   },
   "devDependencies": {
-    "@raycast/eslint-config": "^1.0.8",
+    "@raycast/eslint-config": "^2.1.1",
     "@types/node": "20.8.10",
-    "@types/react": "18.2.27",
-    "@types/react-dom": "18.2.12",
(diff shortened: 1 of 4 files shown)


extensions/tui-type/package.json
```diff
@@ -0,0 +1,61 @@
...
+    "@types/node": "20.8.10",
+    "@types/react": "18.2.27",
+    "@types/react-dom": "18.2.12",
+    "eslint": "^8.57.0",
+    "prettier": "^3.3.2",
+    "typescript": "^5.4.5"
+  },
+  "overrides": {
+    "@types/react": "18.2.27",
+    "@types/react-dom": "18.2.12"
```
LitoMore: Can you explain why we need this?

```suggestion
  "overrides": {
    "@types/react": "18.2.27",
    "@types/react-dom": "18.2.12"
  },
```

MrDoekje: I don't remember why I needed it, but it no longer seems to be required, so I've removed it

0xdhrv:
Thanks for your contribution @MrDoekje 🔥

In this case, Raycast’s core features already provide comparable value for Typing Practice. We try to avoid duplicate functionality in the Store when it can be achieved directly with built‑in features.

If there are unique workflows or capabilities you’re aiming to add that go beyond what the core app can do today, we’d love to hear more so we can see whether:
- They can be expressed using the existing native features, or
- They warrant being added as part of a dedicated extension.

> This helps avoid duplication and keeps related functionality consolidated in one place.

PR #23839, opened 2025-12-20
Update zed-recent-projects extension

remorses:
Improvements to zed-recent-projects extension:

- Add colored dots per project for easy visual identification
- Show green "Open" tag for projects currently open in Zed
- Reorder list when opening a project to show it at the top
- Add "Close Project" action to close open Zed windows directly (macOS only)
- Add "open" and "closed" keywords to filter projects by typing in the search bar

### DIFF
--- a/extensions/zed-recent-projects/CHANGELOG.md
+++ b/extensions/zed-recent-projects/CHANGELOG.md
@@ -1,6 +1,6 @@
 #  Zed Recent Projects Changelog
 
-## [Enhancements] - {PR_MERGE_DATE}
+## [Enhancements] - 2026-01-15
 
 - Add colored dots per project for easy visual identification
 - Show green "Open" tag for projects currently open in Zed


remorses:
Fixed Greptile review feedback

remorses:
Conflicts resolved, PR ready, see the screenshots in the diff to see example of how it looks like now

ewgenius:
@remorses thanks for the contribution - I appreciate the effort. I have some concerns about the proposed UX changes.

-  Colored dots replacing icons don’t add much value in my opinion. The folder icon is a recognizable standard, projects are effectively folders, and remote projects use a distinct icon. With multi-folder projects coming (https://github.com/raycast/extensions/pull/23579), replacing informative icons with dots may reduce clarity.

-  The “Open” tag and open/closed filter seem redundant. The command’s purpose is to find a project and open it; if it’s already open, raycast will focus the existing Zed window. Extra tags and filters add UI noise without a strong use case.

- Reorder list when opening a project to show it at the top. I don't think we should do any custom sorting on top of default one (by last opened) from the Zed workspaces table. If project was opened not through Raycast extension, it must be on top.

I prefer changes that clearly improve the core flow. We have a decent install base (>7 500), so we shouldn’t drastically change the UI or add new features without clear benefits or user requests.

Happy to iterate on a narrower set of improvements aligned with these principles.

remorses:
> Colored dots replacing icons don’t add much value in my opinion

Using a different color per project makes it much easier to spot projects, without reading the name each time, your mind associates each color to a project. I am much faster spotting each project to open right now, and I can use this extension as a tab switcher for Zed. I am sure many useres who are not super fast readers like me will find this super useful

> The “Open” tag and open/closed filter seem redundant

I find this super useful. I sometimes have 10 projects open at the same time. If i want to quickly find what projects i have open I can quickly spot them there and close them. Or search for the word open and find them.

> Reorder list when opening a project to show it at the top. I don't think we should do any custom sorting

Reordering outside of sqlite is required to make sure the last opened project is at the top when you do not restart the 
extension. This happens when you open a project via extension, then trigger again the recent zed proejcts extension (I have a shortcut for this). If we do not sort the list again the project at the top will be wrong because the extension did not refresh. Another approach would be to force rerender via a useState but it is less elegant in my opinion.

This PR improved my Zed DX a lot, I can now use this extension as a tab switcher. The previous use case still works really well. I am sure other users will love these improvements. cc @marcjulianschwarz. If not we can revert it in a later PR. Agree?

ewgenius:
Thanks for the detailed follow-up. I agree this workflow works well for you, but we can’t assume it fits everyone. I use the extension differently and don’t find these additions helpful for my use cases - specifically, I’m not a fan of extra colors and tags.

For context: I have a hotkey to open the command, then I either type first letters of the project or select one of the pinned items and hit Enter.

To reiterate: the purpose of “Search Recent Projects” is to quickly find a recent project and open it (or focus it if already open). It’s intentionally narrow in scope, and that scope is the main reason for my pushback on adding extra features.

As a reference: https://www.raycast.com/thomas/visual-studio-code - I created the Zed extension to mirror the simple “recent projects” experience from that VS Code command. Its UX has remained minimal and mostly unchanged since the start, and the extension serves a significantly larger user base.

> Reordering outside of sqlite is required to make sure the last opened project is at the top when you do not restart the extension

Ok, that is bug. Revalidate makes sense 👍

> This PR improved my Zed DX a lot, I can now use this extension as a tab switcher.  

That suggests a different workflow. The command should remain simple with a clear purpose, not an all‑in‑one solution. I’d rather consider a new command (e.g., “Zed Project Switcher”) for that use case, or make extra features configurable so the core flow remains unchanged by default.

> I am sure other users will love these improvements  

We don’t have a reliable feedback channel to validate that assumption.

> If not we can revert it in a later PR. Agree?  

No - I’d like to avoid back-and-forth churn.

Given this, I’m not comfortable merging the current proposal as-is. My suggestions:
-  I’m OK adding open/closed tags and colored accents if they can be disabled via extension preferences (default off), similar to branch tags. Icons should remain the default option.
-  Submit a separate PR that fixes sorting by revalidating on open.

Please send changes as separate PRs. It’s less convenient, but it makes review easier and keeps discussions focused.

pernielsentikaer:
We can guard some of the things by preferences; that way, we might be able to extend it a bit and the colors?

pernielsentikaer:

marcjulianschwarz:
Just my two cents as I was mentioned. I tried to use the extension the same way as @remorses. So like a Zed tab switcher. Colors or different icons would definitely help a lot. 

I think a preference for all of this functionality would be great!

remorses:
Added preferences to use the previous icons styles by default. Added a preference for the open tag visibility too.

extensions/zed-recent-projects/package.json
```diff
@@ -76,15 +76,34 @@
...
+      "label": "Show Colored Icons",
+      "default": false,
+      "description": "Use colored dots instead of folder icons for easier visual identification of projects"
+    },
+    {
+      "name": "showOpenStatus",
+      "type": "checkbox",
+      "required": false,
+      "label": "Show Open Status",
+      "default": true,
```
ewgenius:
```suggestion
      "default": false,
```

extensions/zed-recent-projects/package.json
```diff
@@ -76,15 +76,34 @@
...
+    },
+    {
+      "name": "showColoredIcons",
+      "type": "checkbox",
+      "title": "Visual Customization",
+      "required": false,
+      "label": "Show Colored Icons",
+      "default": false,
+      "description": "Use colored dots instead of folder icons for easier visual identification of projects"
+    },
```
ewgenius, line 97: It is better to use `dropdown` instead of checkbox, with two options "Icons", "Colored Dots". And name this accordingly, something like "Project Icon Style".

extensions/zed-recent-projects/src/search.tsx
```diff
@@ -134,15 +189,34 @@ export function Command() {
...
-  ) : (
-    <Action.Open title="Open in Zed" target={entry.uri} application={app} icon={zedIcon} />
-  );
+
+  if (entry.wsl) {
+    return <Action title="Open in Zed" onAction={openZedInWsl} icon={zedIcon} />;
+  }
+
+  if (allPaths && allPaths.length > 1) {
+    return <Action title="Open in Zed" icon={zedIcon} onAction={() => openMultiFolderInZed(entry, zedBuild)} />;
```
ewgenius: Nice, I didn't realize it adds multi folder support. That should be mentioned in release notes.

extensions/zed-recent-projects/src/search.tsx
```diff
@@ -105,9 +146,23 @@ export function Command() {
...
-                    <OpenInZedAction entry={entry} />
+                    <OpenInZedAction
+                      entry={entry}
+                      allPaths={entry.allPaths}
+                      zedBuild={zedBuild}
+                      onOpen={() => openEntry(workspace)}
+                    />
+                    {isMac && entry.isOpen && (
+                      <Action
+                        title="Close Project"
```
ewgenius: That seems a bit ambiguous to me. Let's clarify that it closes the opened project window.

extensions/zed-recent-projects/src/search.tsx
```diff
@@ -105,9 +146,23 @@ export function Command() {
               <EntryItem
                 key={entry.uri}
                 entry={entry}
+                keywords={[entry.isOpen ? "open" : "closed"]}
```
ewgenius: How does this behave if I have projects with `open` or `closed` in the name?
Let's put it under the preference flag along with open badge.

ewgenius (changes requested):
@remorses, thank you. I left a few comments regarding preferences.

The multi-folder implementation seems more like a workaround than proper support. The extension can detect if the Zed CLI is installed upon loading and can then show or skip multi-folder projects accordingly. Having both `path` and `allPaths` fields is confusing.

Leave this to me. I'm planning to take a shot at this and do some code cleanup this weekend.

Let's keep this PR only for open/closed filtering and colored dots option.
Also since default configuration is not changed, let's keep regular screenshots.

extensions/zed-recent-projects/package.json
remorses: Done, changed to false.

extensions/zed-recent-projects/package.json
remorses: Done, changed to dropdown with "Icons" and "Colored Dots" options, renamed to "Project Icon Style".

extensions/zed-recent-projects/src/search.tsx
remorses: Removed multi-folder support from this PR as requested. You can handle it in a separate PR.

extensions/zed-recent-projects/src/search.tsx
remorses: Done, renamed to "Close Project Window".

extensions/zed-recent-projects/src/search.tsx
remorses: Done, keywords are now only added when showOpenStatus preference is enabled.

ewgenius (approved):

ewgenius:
@remorses thanks!
@pernielsentikaer it is good to go

pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #23856, opened 2025-12-21
Update things extension

faridrashidi:
## Description

This PR introduces a new **Quick Find** command, bringing the native search capabilities of Things into Raycast. Users can now efficiently search across all Areas, Projects, and To-dos from a single command without needing to switch contexts.

### Features
- **Unified Search**: Aggregates results from Inbox, Anytime, Someday, Upcoming, Projects, and Areas into a single list.
- **Fast Navigation**: Open any search result directly in Things.
- **Performance**: Uses cached data fetching for a snappy experience.

### Changes
- Added `src/quick-find.tsx` to implement the search logic.
- Registered the `quick-find` command in `package.json`.

### How to Test
1. Build the extension.
2. Run the **Quick Find** command.
3. Type to search for a known Project, Area, or To-do.
4. Select an item and press `Enter` to reveal it in Things 3.

## Checklist


faridrashidi:
@andreaselia @samuelkraft would you minding reviewing this PR?

bendrucker:
The performance issues with AppleScript automation in macOS Tahoe are probably going to be an impediment to this. The related issues around timeouts that started popping up recently. Querying all of these lists will be expensive.

faridrashidi:
> The performance issues with AppleScript automation in macOS Tahoe are probably going to be an impediment to this. The related issues around timeouts that started popping up recently. Querying all of these lists will be expensive.

Not sure what you mean by "expensive". On my M1 machine, there's no noticeable lag or performance issues at all. Have you tested it?

bendrucker:
https://github.com/raycast/extensions/issues/22504

You can trace related issues/changes from there. 

"Expensive" means slow in this case I don't actually know the root cause of the slowness. I definitely notice it in many views in the extension. Data loading operations that used to be near instant (tens to hundreds of milliseconds) for my Things DB are taking multiple seconds, sometimes 10+. When the data is cached you'll see instant loads. But if every list cache lookup misses, it'll add up. 

Haven't had a chance to run this in dev mode.

bendrucker:
What's unique about this feature is that it accesses multiple lists where existing commands access a single list. The performance regression is pre-existing since Tahoe. I have not benchmarked it thoroughly on Tahoe and certainly don't have any pre-Tahoe benchmarks to reference. 

Not necessarily a blocker to adding this but the tail latency could be pretty bad in light of this perf bug.

faridrashidi:
@bendrucker If you're supposed to review this PR, please test it in dev mode and share your thoughts with me.

bendrucker:
> If you're supposed to review this PR

I'm a volunteer/contributor so your expectations are off here.

If you could include a screencast in the PR, that would be helpful. And if you can try to force clear or even disable the cache and time how long this command takes vs a single list, that would help address the concern or document the expected performance limitation. Again, this is a pre-existing performance regression. It's not your fault, but it is your problem, to consider on the user's behalf. 

I don't have the time/inclination to do this perf investigation myself right now, the code LGTM though.

faridrashidi:
> I'm a volunteer/contributor so your expectations are off here.

You appear to be speculating without testing the current implementation. I can't address speculation until I understand the specific issue. By the way, I'm short on time and don't want to waste it on benchmarking and performance investigation. If you believe this PR isn't suitable, please close it. I thought others might find it helpful.

By the way, if you're a volunteer and don't have time, let the developer or Raycast team review and test it.

Best!

bendrucker:
All of the `Show * List` commands use the `TodoList` component, which loads data like this:

https://github.com/raycast/extensions/blob/be592af954f51d7c2853de343d70d709724d47a2/extensions/things/src/components/TodoList.tsx#L24

I use `Show Today List` constantly, I have it on a `Hyper-T` shortcut. 

In some very unscientific testing, I just ran a stopwatch after hitting that. It loads for **7.6 seconds**, as shown by the side to side pulsing loading bar at the top. n=1, but representative of what I've been seeing since upgrading to macOS 26. When there's a cache hit I have stale data in the UI while that loading happens. But if there's a cache miss it's an empty UI. Also it's pretty noticeable that the latency is correlated if not roughly linear with the size of the list, in number of todos (my today is currently at 28 😬). Anytime pushes over 10s, while my relatively clean inbox is <5s.

A list command gets 1 list, this command would get 5 lists. That's the performance concern, I don't need to run it to flag that, I can take the existing performance issues in a function and infer that calling it repeatedly might be especially slow. I don't think there's a way to mitigate this given the automation interface exposed by Things and the intended functionality of this command. All I'm asking you to do is try to stress test how this command performs with somewhat larger lists.

bendrucker:
Meanwhile, I'm going to focus on taking a fresh look at whether there are performance optimizations that could be made to how the JXA access is happening. I like the idea of `Quick Find`, I just can't get past how degraded the extension is for me in the basic 1 list command, I can't imagine 5x. If the extension didn't serve cache data, all the list commands would be unusable for me.

bendrucker:
https://github.com/raycast/extensions/pull/24286

That would speed up these list calls. It's still going to be noticeably slow to load 5 lists with ~100 todos total.

Also worth noting, if you wanted to faithfully replicate Things native _Quick Find_ feature, you'd need 2 more lists: `Logbook` and `Trash`. Things includes them in its searches.

And so the performance limitations of loading all data into memory become more stark. While those 5 lists are on the order of tens of to-dos for me and probably at most 100s for the majority of users, my logbook sits at 13,523 items. Things is not actually loading all of those entries into memory itself. It's delegating that query to SQLite.

faridrashidi:
@bendrucker, Thanks for the detailed feedback! The performance concerns are valid. I've refactored Quick Find to use a single combined JXA query instead of 6 parallel calls. This should dramatically reduce the AppleScript overhead since the Tahoe latency appears to be per-invocation.

Key changes:

- Consolidated all data fetching into one `getQuickFindData()` call
- Only fetching essential fields (no nested objects, dates, or notes)
- Deduplication happens in JXA to minimize data transfer

It should now make Quick Find's performance roughly comparable to a single list command rather than 5×.

Also agreed on keeping logbook out of scope, that's a fundamentally different query pattern that would need SQLite-level access.

Please let me know your opinion.

bendrucker:
Nice I will review the updated code soon (might be early next week) and then try it locally. Apple Events serializes access btw so parallel loading doesn't help.

pernielsentikaer:
Thanks for helping out @bendrucker - will await your feedback 🔥

bendrucker:
Got distracted with follow ups on #24286, I think #24489 may end up influencing this feature. In that PR I am moving the tag search out of the main list search bar and into its own `searchBarAccessory` to get closer to the way Things works natively and defer the performance penalty of ancestor tag fetching that the list view isn't waiting on it before it can render any todos.

0xdhrv:
Hey @faridrashidi @bendrucker 👋 

bendrucker:
Thanks, I have not had time to test locally, will share notes when I do

faridrashidi:
@bendrucker, did you have a chance to review this?

bendrucker:
Overloaded at the moment, any time I have to work on these extensions I need to give to finishing https://github.com/raycast/extensions/pull/24489 first

bendrucker:
Just gave this a quick try just to be sure this isn't sitting here ready to go. The expected performance issues are definitely there. This sat in a loading state for 15 seconds on initial open for me, when there's no cached data to display.

Screenshot 2026-02-03 at 22 49 02

I would say this is a nonstarter. Maybe there are some clever ways to reduce the time to first content, but you're facing an uphill battle:

* Things provides no search API
* Accessing list data via Apple Events is slow and all access is serialized

faridrashidi:
@bendrucker, You're right that Apple Events is the bottleneck and there's no way to make it fast enough. Even consolidated into a single JXA invocation, each property accessor is still a serialized IPC round-trip.

I've switched Quick Find to read directly from Things' SQLite database (~/Library/Group Containers/JLMPQHK86H.com.culturedcode.ThingsMac/ThingsData-*/Things Database.thingsdatabase/main.sqlite). A single SQL query with JOINs fetches all areas, projects, and open todos in one shot via /usr/bin/sqlite3 -readonly, piped through stdin. Things uses WAL mode so concurrent reads are safe.

Initial load should now be <100ms instead of ~15s. The JXA path is kept as an automatic fallback in case the DB path changes.

bendrucker:
I'll leave this up to the Raycast team, but I think personally this is a line that this extension (and extensions in general) should not cross. The Applescript/JXA/URL interfaces are officially supported, while directly accessing the application's SQLite DB is not. A fallback doesn't really address the core complexity/maintainability concerns and if anything exacerbates the former.

bendrucker:
Coincidentally stumbled onto `whose` while writing other JXA for Things:

https://bru6.de/jxa/basics/working-with-objects/#filtering-lists-arrays-with-whose

Also `things.toDos` will get you all open items across lists reasonably efficiently.

I do think there's a path to building something reliable/maintainable here if you compromise on full feature parity with Quick Find in Things.

pernielsentikaer:
Where are we on this one?

faridrashidi:
> Where are we on this one?

Following @bendrucker comment at https://github.com/raycast/extensions/pull/23856#issuecomment-3863902419, "I'll leave this up to the Raycast team,..."

So, I'm waiting for your reviews. I'm not sure why it was changed to draft last week. Those Copilot reviews are not accurate. If possible, please ask Greptile for a fresh review.

pernielsentikaer:
Now you'll get Greptile review, and I would say if there is an official way it might sound like the right way to do if possible 🙂

faridrashidi:
> Now you'll get Greptile review, and I would say if there is an official way it might sound like the right way to do if possible 🙂

Thanks Per! Now should be ready for your review.

extensions/things/src/api.ts
```diff
@@ -18,6 +21,57 @@ import {
...
+  return join(container, 'Things Database.thingsSQLite');
+}
+
+let _thingsDBPath: string | undefined;
+function getThingsDBPath(): string {
+  if (!_thingsDBPath) _thingsDBPath = findThingsDBPath();
+  return _thingsDBPath;
+}
+
+// Pipe SQL to sqlite3 via stdin to avoid shell/argument parsing issues
```
bendrucker: Much easier solution here, don't use a shell at all, `execFile`, not `exec`

bendrucker:
To provide a little more color:

- There is some precedent for SQLite backdooring into apps in a small # of extensions (https://github.com/search?q=repo%3Araycast%2Fextensions+sqlite3+language%3ATypeScript&type=code&p=1&l=TypeScript)
- When backdooring/reverse engineering is the only option, I'll consider the accompanying   cost in complexity and maintenance. But when there's an official API, I'm inclined to live with it, limitations included. 

Which is to say, as both a user and contributor, this is not for me. But if it's limited to this command, I'll live with that.

bendrucker:
Random other thought that comes to mind, you'll want to document that this requires granting Full Disk Access. Which is part of the reason I'd say this should be strictly limited to this one command, so that requirement isn't imposed on all users.

faridrashidi:
> Random other thought that comes to mind, you'll want to document that this requires granting Full Disk Access. Which is part of the reason I'd say this should be strictly limited to this one command, so that requirement isn't imposed on all users.

Thanks for the feedback! Switched runSqlite to use execFile. Now much cleaner, and equally safe since it doesn't spawn a shell. I've also added a note about the Full Disk Access requirement scoped specifically to the Quick Find command (in both the README and package.json description), so it's clear this doesn't affect other users.

pernielsentikaer:
Thanks for the corrections, can you check that it still works @bendrucker since I don't have things installed - feel free to approve the PR if it works, then it pops up in my list 🙂

pernielsentikaer:
Are you still around @bendrucker

pernielsentikaer:
Pinging @bendrucker a last time 🙂

loris:
If it can help, I have things installed and Raycast already had Full Disk Access granted on my side (for other extensions). Checked out the extension PR and it worked as expected (no delay, immediate result)
CleanShot 2026-03-13 at 09 16 29@2x

pernielsentikaer (commented):
Thanks for giving it a spin, @loris 🔥

Can you both (@loris and @faridrashidi) test it after I push a few changes to ensure I didn't break anything? 🙂

faridrashidi:
> Thanks for giving it a spin, @loris 🔥
> 
> Can you both (@loris and @faridrashidi) test it after I push a few changes to ensure I didn't break anything? 🙂

@pernielsentikaer Works well Per! Thanks!

pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #23867, opened 2025-12-22
Update imessage-2fa extension

hyusap:
## Description

This PR adds support for multiple Gmail accounts to be logged in simultaneously. Users can add, authorize, and manage multiple accounts from an action within the extension.

## Screencast

image

image

image

image

image

image

## Checklist


hyusap:
@greptileai is it fixed?

hyusap:
just wanted to refresh this thread. please let me know if there's anything i need to do on my end!

clins1994:
@hyusap seeing it now 😓 Looks like the PR was closed. Can you submit a fresh one? 

If I don't see it feel free to ping me on Slack community as I might notice quicker: Caio Lins

Thanks for contributing 🥹

pernielsentikaer:
That would break what this extension is for, wouldn't it?

hyusap:
what do you mean it would break it? its the same functionality but with additional accounts.

pernielsentikaer:
Isn't this extension focused on iMessage?

clins1994:
@hyusap could make a new generic 2fa grabber like "mail-2fa" or a specific "gmail-2fa" extension. Pier's got a point. Keep them short and simple!

pernielsentikaer:
I think we have a lot of extensions in the store doing this, don't we?

hyusap:
this pr isn't creating that much new functionality. this extension already supports gmail, im just allowing it multiple gmail accounts at the same time.

pernielsentikaer:
Hey @hyusap — picking this back up.

The multi-account direction is great and mostly well structured. Before we can land it, a few things need fixing:

1. Rules-of-Hooks violation in `src/account-manager.tsx`: the `!preferences.gmailClientId` early return happens before `useCallback`/`useEffect`, which will crash once the `clientId` is configured. Move the check after all hooks, or lift it to a parent.

2. Silent per-account failures in `useGmail`: right now a broken refresh token on one account silently returns `[]` with no UI signal. Please surface per-account failure state in the `Account Manager` so users know which account needs re-auth.

3. Stale account count in `index.tsx`: `gmailAccountCount` is set once on mount and never refreshed, so adding/removing accounts doesn't update the empty-state messaging until the extension is relaunched.

4. Minor: `checkGmailAuth` is dead code (auto-triggers `authorize()` — dangerous); `removeAccount` doesn't evict the cached `PKCE` client from `oauthClients`; `bun.lockb` should not be in the checked-in `.gitignore`.

If you're up for pushing fixes, I'm happy to re-review.

him229:
@hyusap - I just created a PR to fix comments here - https://github.com/hyusap/extensions/pull/1

hyusap:
> @hyusap - I just created a PR to fix comments here - [hyusap#1](https://github.com/hyusap/extensions/pull/1)

merged

pernielsentikaer:
🔔 @yuercl can you give this PR a spin?

him229:
@yuercl bumping this up

pernielsentikaer:
@greptileai can you do a fresh review of this PR?

pernielsentikaer:
Thanks for this — multi-account support is a great addition, and the per-account `PKCEClient` design (separate `providerId`, tokens kept in Raycast's `OAuth` store, only `id`+`name` in `LocalStorage`) is exactly the right approach.

The `account-manager` UI with per-account auth status is really nice. A few things I'd like to sort out before we merge:

1. Background polling can open `OAuth` windows. The 10s poll calls `getGmailMessages` → `authorize()` for every account, and `authorize()` opens a browser when an account has no token. So an account that's added-but-not-authorized (or whose refresh was rejected) will pop an `OAuth` window on a timer. Could you gate the fetch loop on existing tokens (a `getTokens()`/`verifyAccountAuth` check) and only run the interactive flow from the explicit "Authorize" action?

2. Migration for existing users. This switches the `providerId` from "google" to `google-<uuid>`, so users who already authorized will lose their connection and the old "google" token is left orphaned in the keychain. Please at least note this in the `CHANGELOG` ("re-add your Gmail account after updating"), and ideally migrate the old token into a default account or `removeTokens()` on the legacy `id` so nothing dangles.

3. `expiry_date`. `setCredentials({ expiry_date: tokens.expiresIn })` — Raycast's `expiresIn` is a duration in seconds, but `@googleapis` expects an absolute epoch-ms timestamp, so the client treats the token as expired. It probably still works since you pass a fresh `access_token`, but I'd use `Date.now() + (expiresIn ?? 0) * 1000` (or omit it).

4. Polling volume. `useGmail` has its own `setInterval` and `index.tsx`'s `useInterval` also calls `revalidateGmail`, so there appear to be two timers. Each tick also does a full `messages.get` per message per account, which multiplies against the user's `OAuth` quota. Could you consolidate to one timer and consider narrowing the fetch?

5. Stale codes. In `useGmail`, `setData` only runs when `length > 0`, so when codes age out of the window the old ones stay on screen. Setting it unconditionally would fix that.

Minor: there's a fair amount of `console.log` debug output that's worth removing, and `removeAccount` could reuse `getOAuthClient` instead of constructing a throwaway client.

Happy to take another look once these are in — thanks again for the contribution!

PR #23889, opened 2025-12-22
Fix the missing logo from the community badge

LitoMore:
The Slack icon has been deleted from Simple Icons due to a legal issue:

- https://github.com/simple-icons/simple-icons/pull/13503

I've used a Raycast icon instead.


0xdhrv:
@pernielsentikaer 

This looks broken in it's current state, so I think we are need to use Raycast's logo instead of Slack's logo.

| Before | After |
|---|---|
| image | image |

pernielsentikaer:
Looks like this is currently the best solution 🙂

PR #23992, opened 2025-12-27
Update clockify extension

pcamminadi:
## Description

Adds time entry management features to the Clockify extension:

  - Stop Timer at specific time - Stop a running timer at a past time, useful when you forgot to stop before a break. Available via action menu (Cmd+K) on a running timer or Cmd+Shift+S shortcut.
  - Add Time Entry - Create completed time entries with both start and end times for logging past work that wasn't tracked in real-time.
  - Optional start time for new timers - Backdate when a timer started if you forgot to start it on time.

  Also fixes the API request format to use top-level start/end fields per Clockify API specification.

## Screencast

https://github.com/user-attachments/assets/d7e88dcd-d9c6-4c4c-9b13-0b7036971a3d

## Checklist

### DIFF
--- a/extensions/clockify/CHANGELOG.md
+++ b/extensions/clockify/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Clockify Changelog
 
-## [Time Entry Management] - {PR_MERGE_DATE}
+## [Time Entry Management] - 2026-01-14
 
 - Add "Stop Timer at" action to stop running timer at a specific past time
 - Add "Add Time Entry" to create completed entries with start and end times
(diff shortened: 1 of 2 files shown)


xmok (changes requested):
Hi @pcamminadi:

From the video it looks pretty good!

Since you're already refactoring some of the code, I suggest we utilise [useForm](https://developers.raycast.com/utilities/react-hooks/useform) to make the code in some places (like `DatePicker` assign and most `validation`) simpler and more robust.

pcamminadi:
Hi @xmok, I've changed the current implementation to make use of useForm instead of manual validation.
Since raycast's useForm does not support cross-field validation, I left the validation for the `startDate <= endDate` (Add Time Entry feature) in the onSubmit.

xmok:
You actually can do some cross validation by using the `values` of `useForm` but it's not a requirement.
Will review the changes in a bit!
Example, for reference
```
const {handleSubmit, itemProps, values} = useForm...
    validation: {
        startDate(value) {
            if (value < values.endDate) ...
...
```

pcamminadi:
Thank you for the correction, I've tried it and the user experience is actually a lot better since the validation kicks in instantly instead of only after trying to add the entry.
I've changed that as well now.

PR #24006, opened 2025-12-27
Update google-workspace extension

xilopaint:
## Description

- 📂 **Hierarchical Drive navigation** — The extension now loads only files from the root directory by default (navigate through folders to see their contents).
- 🔍 **Global search** — Searching for a term now automatically searches across the entire Drive, while browsing without a search term stays within the current directory.

## Checklist

### DIFF
--- a/extensions/google-workspace/CHANGELOG.md
+++ b/extensions/google-workspace/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Google Workspace Changelog
 
-## [Hierarchical Navigation] - {PR_MERGE_DATE}
+## [Hierarchical Navigation] - 2026-01-05
 
 ### Changed
 


extensions/google-workspace/package.json
```diff
@@ -95,8 +95,8 @@
       "mode": "no-view"
     },
     {
-      "name": "search-google-drive-files",
-      "title": "Search Google Drive Files",
+      "name": "search-google-drive",
```
pernielsentikaer: Can we keep the old `name`? It is the unique ID of the command where ranking, aliases, and hotkeys are saved. You can, of course, change the title as you did 🙂

xilopaint: Done.

PR #24097, opened 2026-01-01
Add rewrite-anywhere extension

jopmiddelkamp:
## Description

**Quick AI text rewriting and translation tool** — Transform any selected text instantly using OpenAI GPT. Perfect for business emails, casual messages, and professional communication.

## Screencast

Screenshot 2026-01-01 at 18 36 17

Screenshot 2026-01-01 at 18 36 39

## Checklist


pernielsentikaer:
Hi 👋

What's the purpose of this extension, there is nothing that's not possible out of the box in Raycast already

jopmiddelkamp:
Yes, there are tools available, but they are part of Raycast AI, which requires payment. I cannot use my own AI API key. Additionally, it's less tailored to my use case because I have to set up many commands to accomplish something that I grouped behind a single hotkey with a simple option menu and one AI call.

pernielsentikaer:
Can't you use your own key or local models?

CleanShot 2026-02-12 at 22 04 17@2x

jopmiddelkamp:
You can think outside the box more than I expected. However, there is no control over tone, etc. Could you also add profiles that we can switch between to impact the writing style? This way, we wouldn't need to use two commands and consume tokens twice.

We should be able to change the profile for the text, just like we can change the model.
image

I wish I could configure different profiles for private and work-related text improvements. For example, in work, I have informal and formal styles.

You can create a work profile that provides the AI with your job title and additional context that I consider necessary. This would make the tool much more flexible to use.

It would also be beneficial to have a setting for automatic translation using the latest profile, or an option to manually select a setting before the translation occurs. This is how I configured it in my own plugin. See my example.

pernielsentikaer:
You could save them as a preset and use the `Continue in Chat with...` option or save them as different commands instead? This might be something we'll explore in the near future, and making it faster and easier is always a good idea 🙂

PR #24133, opened 2026-01-02
Change install button to image link format

chrismessina:
Updated the installation section to use an image link for the install button.

## Description

Improving formatting of embedded image, which leaks HTML in the Raycast Docs extension. 

raydocs 2026-01-03 at 21 18 40

### DIFF
--- a/docs/information/developer-tools/forked-extensions.md
+++ b/docs/information/developer-tools/forked-extensions.md
@@ -13,7 +13,7 @@ This extension leverages the [Git sparse-checkout](https://git-scm.com/docs/git-
 
 The extension is available on the [Raycast Store](https://www.raycast.com/litomore/forked-extensions).
 
-[![](https://www.raycast.com/litomore/forked-extensions/install_button@2x.png?v=1.1)](https://www.raycast.com/litomore/forked-extensions "Install forked-extensions Raycast Extension")
+<p align="center"><a title="Install forked-extensions Raycast Extension" href="https://www.raycast.com/litomore/forked-extensions"><img src="https://www.raycast.com/litomore/forked-extensions/install_button@2x.png?v=1.1" height="64" style="height: 64px;" alt="" /></a></p>
 
 ## Hint
 


docs/information/developer-tools/forked-extensions.md
```diff
@@ -13,7 +13,7 @@ This extension leverages the [Git sparse-checkout](https://git-scm.com/docs/git-
 
 The extension is available on the [Raycast Store](https://www.raycast.com/litomore/forked-extensions).
 
-<p align="center"><a title="Install forked-extensions Raycast Extension" href="https://www.raycast.com/litomore/forked-extensions"><img src="https://www.raycast.com/litomore/forked-extensions/install_button@2x.png?v=1.1" height="64" style="height: 64px;" alt=""></a></p>
+[![](https://www.raycast.com/litomore/forked-extensions/install_button@2x.png?v=1.1)](https://www.raycast.com/litomore/forked-extensions "Install forked-extensions Raycast Extension")
```
LitoMore: You're removing the `align="center"` style from the document.

I want to keep it centered at all times.

chrismessina: Image

Understood — the issue (which I guess is a matter of priority) is that the HTML is leaking in the Raycast Docs extension. 

If we can fix the way Raycast renders embedded HTML in Markdown, then there's no need to change it. 

But this is why I made this PR (I'll add it to the description).

0xdhrv: @chrismessina If you can update it to be in the center? And also if possible can we have height to be `64px` so it looks more consistent.


| main | branch |
|---|---|
| Image | Image |

0xdhrv (commented):
Hi 👋 

https://github.com/raycast/extensions/pull/24133/changes#r2666080056

docs/information/developer-tools/forked-extensions.md
chrismessina: Image

Maybe my comment is awaiting moderation? Here's what I wrote:

> Understood — the issue (which I guess is a matter of priority) is that the HTML is leaking in the Raycast Docs extension. 
> 
> If we can fix the way Raycast renders embedded HTML in Markdown, then there's no need to change it. 
> 
> But this is why I made this PR (I'll add it to the description).

@0xdhrv it sounds like you want to keep the current HTML-in-Markdown, but **the reason for this PR is that it breaks the docs in the Raycast Docs extension**.

0xdhrv: I see, my concern is on the Raycast Docs front, if there's a way to handle both separately that would be ideal, but so far it seems like we can go with you fix.

@pernielsentikaer This seems like a good fix, if possible we track "Raycast renders embedded HTML in Markdown" for long-term maintenance.

LitoMore:
```suggestion
<p align="center"><a title="Install forked-extensions Raycast Extension" href="https://www.raycast.com/litomore/forked-extensions"><img src="https://www.raycast.com/litomore/forked-extensions/install_button@2x.png?v=1.1" height="64" style="height: 64px;" alt="" /></a></p>
```

LitoMore: Let's add a slash to fix the format in the Raycast app.

LitoMore (changes requested):

docs/information/developer-tools/forked-extensions.md
LitoMore: The `Copy Embed Code` button provides a wrong format of the `` tag. It's missing a slash behind.

We can simply add a slash to fix it.

0xdhrv: Thanks for the feedback. Noted on this

chrismessina: I've added the slash to my commit.

Issue #24141, opened 2026-01-03
[Bartender] Extension does not support Bartender 6.3.0


JhonnyAxe:
### Extension

https://www.raycast.com/kmusick/bartender

### Raycast Version

1.104.1

### OS Version

macOS 15.2

### Description

The Bartender Raycast extension fails to work with Bartender version 6.3.0. The extension appears to be designed for an older version of Bartender and is not compatible with the latest release.

When trying to use the extension commands, they either don't execute properly or fail to interact with Bartender 6.3.0.

### Steps To Reproduce

1. Install Bartender 6.3.0
2. Install the Bartender extension from Raycast Store
3. Open Raycast and try to use any Bartender command
4. Observe that the extension doesn't work correctly with Bartender 6.3.0

### Current Behaviour

The extension commands fail to execute or do not properly interact with Bartender 6.3.0.

### Expected Behaviour

The extension should work seamlessly with Bartender 6.3.0, allowing users to control Bartender features through Raycast commands as intended.

kmusick:
Please see #21856 - I wish it was fixable, but I have checked again and confirmed that I am not able to run any AppleScript against Bartender 6 (I have the Setapp version). Hopefully someone is able to figure out a fix, or Bartender gets an update to fix AppleScript support.

```applescript
tell application "Bartender" to return version
```

Returns 6.3.0

```applescript
tell application "Bartender" to toggle bartender
```

and other scripts listed in their dictionary fail with `error "Bartender got an error: The handler some object is not defined." number -1717`

@raycastbot close as duplicate

Issue #22268, opened 2025-10-18
[Bartender]...No Commands Work


BellalaK:
### Extension

https://raycast.com/kmusick/bartender

### Raycast Version

1.103.3

### macOS Version

26.0.1

#### Error:
```
Command failed with exit code 1: osascript
32:51: execution error: Bartender got an error: The handler some object is not defined. (-1717)
```

### Steps To Reproduce

Standard config, none of the commands work.

kmusick:
I believe Bartender 6 is not supporting scripts correctly. See #21856 

@raycastbot close as duplicate

Issue #21856, opened 2025-09-29
[Bartender] Bartender 6 on macOS Tahoe error


texastoland:
### Extension

https://www.raycast.com/kmusick/bartender

### Raycast Version

1.103.2

### macOS Version

macOS 26.0.0

#### Error:

```
Command failed with exit code 1: osascript
32:51: execution error: Bartender got an error: The handler some object is not defined. (-1717)
```

macOS Tahoe requires Bartender 6. I'm currently using the Setapp beta but previously experienced the same behavior in the standalone alpha.

### Steps To Reproduce

1. Upgrade macOS to Tahoe
2. Install Bartender 6
3. Run Raycast command: `Search Menu Bar Apps`

kmusick:
Great bug report, thanks for the clear information. I ran a quick test with the Setapp beta and it looks like this might be a Bartender bug - I wasn't able to run any AppleScript interactions with Bartender 6. I'll need to dig deeper over the next few days.

texastoland:
I'm glad you made this! I asked the Bartender creator for an extensions years ago before he sold it. Your implementation has been perfect for my use case 🙌🏼

kmusick:
Just confirming that this is still an issue. I don't believe Bartender 6 is supporting scripts correctly, but I am running the beta version via Setapp and don't really have the ability to try different versions or confirm if Setapp is causing these issues. Hopefully a future version fixes these problems.

texastoland:
I emailed Bartender and requested someone reply here 🤞🏼

megaroeny:
I'm getting a similar error for the "Open Search Overlay" command as well. ☹️ 

```
Command failed with exit code 1: osascript
34:53: execution error: Bartender 5 got an error: Connection is invalid. (-609)
```

Realized upgrading to v6 still kept my v5 app. Deleting it didn't help.

JhonnyAxe:
Has anyone tried to reach out the bartender support?

megaroeny:
> Has anyone tried to reach out the bartender support?

I totally forgot to. But does this issue qualify support from the Bartender team if this is an extension made by someone not on the team? I guess I don't fully understand the technical issue, so probably better that the author of this extension reach out? @kmusick 

@JhonnyAxe Assuming this is still an issue? I totally forgot about it and been on vacation

kmusick:
[...]

Issue #21797, opened 2025-09-25
[Bartender] error


lonagi:
### Extension

https://raycast.com/kmusick/bartender

### Raycast Version

latest

#### Error:
```
Command failed with exit code 1: osascript
44:53: syntax error: Объект «имя класса (множ. число)» не может следовать за объектом «идентификатор». (-2740)
```

### Steps To Reproduce

<img width="826" height="153" alt="Image" src="https://github.com/user-attachments/assets/7dc13d6a-7914-415b-8b03-db63a062e0a5" />

just select  Search Menu Bar Apps

texastoland:
Not enough info: probably duplicate of #21856.

kmusick:
@raycastbot close as duplicate of #21856

PR #24154, opened 2026-01-03
Updated Bartender package.json description to point out that Bartender 6 is not and cannot be supported

kmusick:
## Description

Bartender 6 has broken AppleScript support. Developer made [one comment](https://www.reddit.com/r/macapps/comments/1oc73it/another_bartender_6_rant/nkrupgi/) on Reddit a few months ago acknowledging the issue. This PR just adds one sentence to the description explaining that Bartender 6 is not supported.

See #24141, #21856, #22268, #21797.

## Checklist

### DIFF
--- a/extensions/bartender/CHANGELOG.md
+++ b/extensions/bartender/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Bartender Changelog
 
+## [Updated Description] - {PR_MERGE_DATE}
+
+- Updated description to point out that Bartender 6 is currently not supported due to broken AppleScript support.
+
 ## [Bug fix] - 2025-03-27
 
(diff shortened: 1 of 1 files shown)


kmusick:
> What if we check which version of macOS the user is on and show a warning about it?

Hoping to get your take on the best way to handle this. Even if Bartender 6 fixes their AppleScript issues, testing across both versions 5 and 6 is difficult because of the OS support gap. Can we limit this extension to Bartender 5, or does Raycast generally discourage version-specific extensions? In a Bartender 5 only world, we could check the version quite easily but I would be unable to fully test that it functions.

Thanks!

pernielsentikaer:
We can call it Bartender 5 in the title, and if 6 will be better supported then we can add that?

PR #24200, opened 2026-01-05
Fix Window Walker Extension

luinbytes:
## Description
Fixes and improvements for the Window Walker extension.

## Fixed
- **Pins**: Pins now work by application name only, so pinning apps like Discord keeps it pinned regardless of which channel you're viewing.
- **Pins**: Fixed bug where pinned windows would not always appear in the list or could be removed incorrectly.

## Improved
- **Refresh Rate**: Window list now auto-refreshes every 1 second (was 10 seconds). The selected index is preserved so refreshes are seamless and unnoticeable.
- **Icon Caching**: Icons are cached on disk and load in background, with automatic cleanup after 7 days.

## Testing
- ✅ Tested on Windows
- ✅ All commands working correctly
- ✅ Build passes
- ✅ Lint passes

### DIFF
--- a/extensions/window-walker/CHANGELOG.md
+++ b/extensions/window-walker/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## [Performance & Pin Fixes] - {PR_MERGE_DATE}
+## [Performance & Pin Fixes] - 2026-01-05
 
 ### Fixed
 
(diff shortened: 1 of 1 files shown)


pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #24215, opened 2026-01-06
Update fancy-text extension

nexxai:
## Description

I've added the ability for the user to set a customizable text decorator that when selected, will surround their generated fancy text on both side.  Not having a decorator will not affect current functionality, and the new functionality is set as a second action, so even if a decorator is configured, it will only be used when specifically requested or intentionally set as the default.

For example: 
- Decorator set to: ✨
- Fancy text generated: 𝓱𝓮𝓵𝓵𝓸
- Copied output when decoration style is "Surrounded": ✨ 𝓱𝓮𝓵𝓵𝓸 ✨

## Screencast

Screenshot 2026-01-05 at 10 02 30 PM

## Checklist

### DIFF
--- a/extensions/fancy-text/CHANGELOG.md
+++ b/extensions/fancy-text/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Fancy Text Changelog
 
-## [Feature] - {PR_MERGE_DATE}
+## [Feature] - 2026-01-06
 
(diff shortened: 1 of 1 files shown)


pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #24224, opened 2026-01-06
Enable sparse checkout for API docs

pernielsentikaer:
This might be enough to work around the file size. Can you check that, @mathieudutour? 

I haven't tested it elsewhere, but I can if you feel it's unsafe 🙂

### DIFF
--- a/.github/workflows/sync-api-docs.yml
+++ b/.github/workflows/sync-api-docs.yml
@@ -16,7 +16,8 @@ jobs:
         with:
           fetch-depth: 2
           sparse-checkout: |
-            docs
+            docs/
+            scripts/
 
       - name: Setup Node
         uses: actions/setup-node@v4


.github/workflows/sync-api-docs.yml
```diff
@@ -15,6 +15,8 @@ jobs:
         uses: actions/checkout@v4
         with:
           fetch-depth: 2
+          sparse-checkout: |
+            docs
```
mathieudutour:
```suggestion
            docs/
            scripts/
```

pernielsentikaer:
I added your suggestion @mathieudutour .

Can you approve and merge it, so we can see if it needs more care? 😇

mathieudutour:
oh that wasn't even the right file... https://github.com/raycast/extensions/pull/24294

PR #24235, opened 2026-01-06
feat(parcel): improve internationalization, date parsing, and carrier name display

damian-duchnowski:
## Description

This PR improves the extension's internationalization support, enhances date parsing accuracy, and improves the user experience by displaying readable carrier names and better-formatted dates throughout the interface.

### Changes

- **Removed hardcoded locale value** - Dates now display correctly for all international users regardless of their system locale
- **Display carrier names instead of codes** - The interface now shows proper carrier names (e.g., "FedEx", "UPS") instead of codes (e.g., "fedex", "ups") for better readability
- **Removed redundant date labels** - Fixed redundant "(Today)" and "(Tomorrow)" labels that appeared when the date already displayed "Today" or "Tomorrow"
- **Improved date parsing for ambiguous formats** - Fixed date parsing to correctly handle ambiguous dot-separated dates (MM.dd.yyyy vs dd.MM.yyyy) by detecting the format based on date values. For example, "01.07.2026" is now correctly interpreted as January 7th instead of July 1st
- **Enhanced date display format** - Recent dates (within 14 days) now display as "Wednesday 09:00" (day name with time), while older dates display as "30 December at 11:15" (full date with "at" before time) for improved readability

### Impact

- Better internationalization: date parsing works correctly for users worldwide
- Improved UX: users see familiar carrier names instead of cryptic codes
- Cleaner interface: removed redundant date labels for a cleaner display
- Accurate date ordering: tracking events now display in correct chronological order
- Better readability: date format adapts based on recency for easier scanning
- Graceful fallback: if a carrier name isn't available, the extension falls back to displaying the uppercase carrier code

### Testing

- Verified date parsing works with various international date formats
- Confirmed carrier names display correctly in list view, detail view, and metadata panels
- Tested fallback behavior when carrier code isn't in the supported carriers list
- Verified redundant date labels are no longer shown when date already indicates "Today" or "Tomorrow"
- Verified ambiguous dates (like "01.07.2026") are correctly parsed based on context
- Confirmed date display format changes based on recency (recent vs older dates)

## Checklist

### DIFF
--- a/extensions/parcel/CHANGELOG.md
+++ b/extensions/parcel/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Parcel Changelog
 
-## [Internationalization and Carrier Name Display] - {PR_MERGE_DATE}
+## [Internationalization and Carrier Name Display] - 2026-01-13
 
 - Removed hardcoded locale value introduced in a previous change - dates now display correctly for all international users
 - Carrier names are now displayed throughout the interface instead of codes (e.g., "FedEx" instead of "fedex") for better readability


pernielsentikaer:
Thanks for the update @damian-duchnowski 🔥

@xilopaint, can you check this? I think you had some issues with date formats in the past, right?

xilopaint:
@damian-duchnowski, could you change your region to **Brazil** and share a screenshot of how that delivery looks in the extension?

Screenshot 2026-01-07 at 07 08 10

damian-duchnowski:
@xilopaint 
That exact one was already gone, but here's another one. No difference after changing the region. Unless there was an extra step after switching it in the settings.

Would any potential issues not be solved by your addition to `DATE_FORMATS`? Since the format is expected, it can be properly parsed.
https://github.com/raycast/extensions/blob/15c50746701663d6b03472337eaadbabae49775a/extensions/parcel/src/my-deliveries.tsx#L35-L36

CleanShot 2026-01-07 at 21 59 06@2x

bendrucker (approved):

xilopaint:
> Would any potential issues not be solved by your addition to DATE_FORMATS? Since the format is expected, it can be properly parsed.

To be honest, I don’t remember exactly what I tried to fix with the hardcoded locale. You can remove it as you already did, and if I notice any issue I’ll submit a new PR and mention you.

At the moment, I’m noticing other issues that you can address in this PR. The images below show discrepancies between the date formats displayed in the application and in the extension that may be a regression introduced by the previous PR:

Screenshot 2026-01-08 at 02 34 44

Screenshot 2026-01-08 at 02 42 12

The tracking numbers are not mine; I obtained them through a web search using Gemini, since I do not have any other delivery registered at the moment to carry out the tests.

xilopaint:
I also notice that in one delivery the extension shows a difference of 3 hours compared to the content displayed in the application, whereas in the other the difference is 4 hours. We need to investigate the reason for this.

pernielsentikaer (commented):
Could be a difference in the API?

pernielsentikaer:
@damian-duchnowski can you check this as the last thing?

aidenkt:
> I also notice that in one delivery the extension shows a difference of 3 hours compared to the content displayed in the application, whereas in the other the difference is 4 hours. We need to investigate the reason for this.

Are these potentially two different carriers?

damian-duchnowski:
> > I also notice that in one delivery the extension shows a difference of 3 hours compared to the content displayed in the application, whereas in the other the difference is 4 hours. We need to investigate the reason for this.
> 
> Are these potentially two different carriers?

**For FedEx:**
- The app and website (web.parcelapp.net) translate timestamps to the user's TZ, because if you check the Network tab and find https://data.parcelapp.net/data.php, you can see the time is 10:10
CleanShot 2026-01-09 at 12 33 22
<br><br>

- Whereas the app and website shows 17:10 and since the Albuquerque is UTC-7 and I'm UTC+0, that explains the difference. @xilopaint I guess that would explain your discrepancies as well.

CleanShot 2026-01-09 at 12 34 55<br>

**But I think this is a bigger issue:**
- For the delivery events, the API provides a date, but doesn't specify the TZ. If I understand correctly, that date is the same as highlighted in the data.php screenshot. However, the data.php also return a Unix timestamp (in blue), which the API doesn't provide, hence we can't get reliable timestamps.
CleanShot 2026-01-09 at 13 13 26
<br><br>

**For UPS:**
- This one's interesting, because the highlighted event provides 10:00 in the date field, whereas the Unix timestamp translates to 9:00, which doesn't make sense because Canada is not UTC-1 (it's UTC-5).
CleanShot 2026-01-09 at 13 09 56
<br><br>

- Whereas the app shows 9:00, so it seems to be using the Unix timestamp the API doesn't provide
CleanShot 2026-01-09 at 13 19 02
<br><br>

- Lack of Unix timestamp in the API surfaces another issue, the date parsing even when updated with the American format, mistakes it for another resulting in 1 Jul instead of 7 Jan, although the previous event was parsed correctly
CleanShot 2026-01-09 at 13 40 45
image
<br><br>

Sorry for the wall of text, hopefully someone will understand, I'm confused as well 😵‍💫
BTW: I think all the above is beyond the scope of this PR. In terms of date parsing, I just removed the hardcoded locale and looking at those deliveries in the current released version, it has the same problems as noted here, just dressed in ptBR locale dates 😆

damian-duchnowski:
@xilopaint

(Almost) Fixed both issues via 415124d
- FedEx screenshot - Fixed the redundant labels in Expected Delivery Date
- UPS screenshot - Added American format MM.dd.yyyy HH:mm, but although correctly parsing 12.30.2025, it incorrectly interpreted the date format in the next event, so will need help with that

pernielsentikaer:
Let's see last review from Greptile and then we can merge and get more feedback, great work 🔥

pernielsentikaer:
Above is intended, right @damian-duchnowski

damian-duchnowski:
> Above is intended, right @damian-duchnowski 

Yes, hardcoding locales should be a no-no. I tested by changing region in system settings and it looked fine. @xilopaint - can you please check yourself so we can be sure? Thank you! 🙏

xilopaint:
Date, time and format don't match between the app and the extension.

Screenshot 2026-01-13 at 08 27 22

pernielsentikaer:
Can you share debug data @xilopaint now you have it open 🙂

damian-duchnowski:
@xilopaint 
Sorry, but times won't match due to what I wrote in the wall of text, because the API doesn't return the UTC Unix timestamp, just the localised timestamp, which might be in different time zones and I don't have a solution for that. I just take the localised one and display it, as it has been done before, nothing changed in that logic.

As for time format, I don't think anyone tried to replicate the format used by the app, and this PR didn't change that (as seen in Greptile reviews).
**EDIT:** Replicated the format via e780c3a.

Only thing I could improve is to fix this issue, which I will do now.
image
**EDIT:** Fixed the issue via e780c3a.

damian-duchnowski:
Final look for this PR 🥳 

CleanShot 2026-01-13 at 17 13 50@2x

Issue #24256, opened 2026-01-07
[Things] AI-generated To-Dos fail to add due to empty fields in URL scheme


ajason:
### Extension

https://www.raycast.com/loris/things

### Raycast Version

1.104.1

### OS Version

macOS 26.2

### Description

When using Raycast AI to create a To-Do in Things 3, the action fails silently.
The issue is that the AI generates a JSON object where unused fields are set to empty strings ("") rather than being omitted. The current implementation includes these empty strings in the Things URL (e.g., deadline=&when=), which causes the operation to fail because Things 3 does not handle empty values for certain parameters (like dates) gracefully.

```
{
  "list": "inbox",
  "deadline": "",
  "titles": "",
  "show-quick-entry": false,
  "notes": "这是一个用于测试的待办事项（由 Raycast AI 创建）。",
  "completed": false,
  "heading": "",
  "checklist-items": "",
  "title": "测试 todo",
  "creation-date": "",
  "when": "",
  "heading-id": "",
  "reveal": false,
  "canceled": false,
  "use-clipboard": "",
  "list-id": "",
  "completion-date": "",
  "tags": "测试"
}
```

The result is:
things:///add?list=inbox&deadline=&titles=&show-quick-entry=false&notes=%E8%BF%99%E6%98%AF%E4%B8%80%E4%B8%AA%E7%94%A8%E4%BA%8E%E6%B5%8B%E8%AF%95%E7%9A%84%E5%BE%85%E5%8A%9E%E4%BA%8B%E9%A1%B9%EF%BC%88%E7%94%B1%20Raycast%20AI%20%E5%88%9B%E5%BB%BA%EF%BC%89%E3%80%82&completed=false&heading=&checklist-items=&title=%E6%B5%8B%E8%AF%95%20todo&creation-date=&when=&heading-id=&reveal=false&canceled=false&use-clipboard=&list-id=&completion-date=&tags=%E6%B5%8B%E8%AF%95

### Steps To Reproduce

* Enter `@Things` in Raycast to add a test to-do.
* The AI invokes the `things add-todo` tool.
* The operation is reported as successful, yet no to-do appears in Things.

janosorcsik:
@raycastbot assign me

janosorcsik:
Hi @ajason,
Please run the `⁠Check For Extension Updates` command in Raycast and verify that the issue is resolved.

ajason:
[...]

PR #24271, opened 2026-01-07
[Update] [Things] Improve Query String Creation

janosorcsik:
## Description

Replaced the `⁠qs` package with `⁠query-string` to automatically exclude empty strings and null values. Possible fix for https://github.com/raycast/extensions/issues/24256

## Checklist

### DIFF
--- a/extensions/things/CHANGELOG.md
+++ b/extensions/things/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Things Changelog
 
-## [Improved Query String Creation] - {PR_MERGE_DATE}
+## [Improved Query String Creation] - 2026-01-08
 
 - Replaced the `⁠qs` package with `⁠query-string` to automatically exclude empty strings and null values when generating Things URLs
 


extensions/things/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Things Changelog
 
+## [Improved Query String Creation] - {PR_MERGE_DATE}
+
+- Replaced the `⁠qs` package with `⁠query-string` to automatically exclude empty strings and null values
```
bendrucker:
```suggestion
- Replaced the `⁠qs` package with `⁠query-string` to automatically exclude empty strings and null values when generating Things URLs
```

extensions/things/src/api.ts
```diff
@@ -343,17 +343,25 @@ export async function silentlyOpenThingsURL(url: string) {
   await asyncExec(`open -g "${url}"`);
 }
 
+const stringifyConfig = {
```
bendrucker: It would be better to define a reusable function in this module (or its own for extra encapsulation) that invokes `queryString.stringify` with this `stringifyConfig` inline as a literal.

Then everything that needs to generate the query portion of a Things URL just calls that with an object rather than always needing to use the upstream function + options in every call.

For example:

```typescript
import queryString, { Stringifiable } from 'query-string';

function thingsQueryString(params: Record<string, Stringifiable>): string {
  return queryString.stringify(params, {
    skipNull: true,
    skipEmptyString: true,
  });
}
```

Then usage becomes:

```typescript
await silentlyOpenThingsURL(`things:///add?${thingsQueryString(todoParams)}`);
```

0xdhrv:
Hey @janosorcsik 👋 


Could you look into the suggestions from @bendrucker ?

extensions/things/src/api.ts
janosorcsik: Thanks for the review! I've created a ⁠`generateQueryString` method to organize the stringifying logic in one place.

PR #24286, opened 2026-01-08
things: optimize JXA performance with `properties()` batching

bendrucker:
## Description

Optimizes JXA performance in the Things extension by using `properties()` batching instead of individual property calls. This reduces Apple Event overhead from ~15+ round-trips per todo to ~3-4, resulting in **~75% faster list loading times**.

This is particularly impactful on macOS Tahoe, where Apple Event latency appears to have increased significantly. Before this change, loading 28 todos took ~5+ seconds. After, it takes ~1-1.5 seconds.

### Changes

- **`api.ts`**: Rewrote `getListTodos` to use `properties()` batching for todos, projects, and areas
- **`api.ts`**: Applied same optimization to `getProjects`, `getAreas`, and `getTagsProjectsAndAreas` mapping templates
- Made `tags` optional in `Area` type since nested areas no longer include them

### Benchmark Results

| List | Items | Baseline | Optimized | Improvement |
|------|-------|----------|-----------|-------------|
| Inbox | 5 | 782ms | 149ms | 81% faster |
| Today | 28 | 5,295ms | 1,141ms | 78% faster |
| Upcoming | 42 | 7,050ms | 1,629ms | 77% faster |
| Anytime | 49 | 8,911ms | 1,993ms | 78% faster |

### Breaking Change: Nested Area Tags Removed

Tags on containing areas (`todo.area.tags`, `todo.project.area.tags`) are no longer fetched. Todo and project data now only includes:

- `todo.tags` — the todo's own tags
- `todo.project.tags` — the project's tags
- `todo.area.name` / `todo.project.area.name` — area names (still available)

Nested area tags required additional Apple Event calls per todo and provided marginal value—users are unlikely to need a tag assigned to a todo's grandparent area. Removing them contributes ~14% additional performance improvement.

## Screencast

N/A - performance optimization with no visual changes.

## Checklist

### DIFF
--- a/extensions/things/CHANGELOG.md
+++ b/extensions/things/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Things Changelog
 
+## [JXA Performance Optimization] - {PR_MERGE_DATE}
+
+- Reduced JXA fetch times by ~75% using `properties()` batching to minimize Apple Event overhead
+- Removed nested area tags from todo data (tags on containing areas are no longer fetched)
+
 ## [Improved Query String Creation] - 2026-01-08
 
 - Replaced the `⁠qs` package with `⁠query-string` to automatically exclude empty strings and null values when generating Things URLs
--- a/extensions/things/CHANGELOG.md
+++ b/extensions/things/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Things Changelog
 
-## [JXA Performance Optimization] - {PR_MERGE_DATE}
+## [JXA Performance Optimization] - 2026-01-12
 
 - Reduced JXA fetch times by ~75% using `properties()` batching to minimize Apple Event overhead
 - Removed nested area tags from todo data (tags on containing areas are no longer fetched)
--- a/extensions/things/src/api.ts
+++ b/extensions/things/src/api.ts
@@ -106,34 +106,50 @@ export const getListTodos = (commandListName: CommandListName): Promise<Todo[]> 
   const things = Application('${preferences.thingsAppIdentifier}');
   const todos = things.lists.byId('${commandListNameToListIdMapping[commandListName]}').toDos();
 
-  return todos.map(todo => ({
-    id: todo.id(),
-    name: todo.name(),
-    status: todo.status(),
-    notes: todo.notes(),
-    tags: todo.tagNames(),
-    dueDate: todo.dueDate() && todo.dueDate().toISOString(),
-    activationDate: todo.activationDate() && todo.activationDate().toISOString(),
-    isProject: todo.properties().pcls === "project",
-    project: todo.project() && {
-      id: todo.project().id(),
-      name: todo.project().name(),
-      status: todo.project().status(),
-      tags: todo.project().tagNames(),
-      dueDate: todo.project().dueDate() && todo.project().dueDate().toISOString(),
-      activationDate: todo.project().activationDate() && todo.project().activationDate().toISOString(),
-      area: todo.project().area() && {
-        id: todo.project().area().id(),
-        name: todo.project().area().name(),
-        tags: todo.project().area().tagNames(),
-      },
-    },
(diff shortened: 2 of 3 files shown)


extensions/things/CHANGELOG.md
0xdhrv, line 41: Could you please update the CHANGELOG entry in the specified format? 

Here's our guidelines about that: 
- [Version History](https://developers.raycast.com/basics/prepare-an-extension-for-store).

0xdhrv:
@greptileai, could you please do a fresh review of this PR?

extensions/things/CHANGELOG.md
bendrucker: PR needed major rebasing as I didn't notice my local was out of date so that was an old unrelated change with the hardcoded date. Fixed.

janosorcsik:
Hi @bendrucker,
Thank you for optimizing the list command.
I'd like to ask if you could apply the same optimization to the other get commands that use JXA to improve their performance as well?

bendrucker:
Done! Applied `properties()` batching to the `mapProjectJxa`, `mapAreaJxa`, `mapProjectTodoJxa`, and `mapAreaTodoJxa` templates as well. These are used by `getProjects`, `getAreas`, and `getTagsProjectsAndAreas`.

pernielsentikaer (approved):
Looks good to me, approved 🔥

janosorcsik:
Hi @pernielsentikaer,
Could you please reopen the PR? The missing tags weren't re-added to the script, and the extension has now lost this feature.

bendrucker:
> Could you please reopen the PR? 

You can't "reopen" merged changes, merged is a terminal state, the side effect already happened on the target branch. 

>  The "missing" tags

Are you talking about the well-documented breaking change or something else? 

If you are now realizing you really want it the old way, the way forward is another PR to make that change. Though I'd advocate for a preference to control whether area tags are fetched. It's a large performance penalty to fetch related data most users won't populate. That's a bit of conjecture on my part, but I'm coming up on 15 years with Things and I've never once tagged an area.

janosorcsik:
Hi @bendrucker,
I meant either reverting the change or fixing it in a follow-up PR.

About the breaking change: just because you documented it doesn't mean it's the right call. The tags were there and working. Removing them just because you haven't used them in 15 years isn't a valid reason. By that logic, we could remove any feature that some users don't need.

The performance improvement is valuable, but it shouldn't come at the cost of removing functionality. If performance is the concern, a preference toggle makes sense, but the feature should stay by default.

I flagged this in my review before the merge. People's workflows are broken now. Can you restore the tags in a follow-up?

bendrucker:
> I flagged this in my review before the merge.

Where? Look, I can do this, but if you want it urgently, just open the revert PR, it doesn't need to be a discussion.

bendrucker:
Re default, it doesn't work this way in Things afaict. My usage is just an example. I proposed the breaking change because i saw a mismatched default behavior with a large perf cost. 

**Edit:** Will document Things native behavior more thoroughly, but the more fundamental issue here is loading up every attribute and allowing full text search of a specific list. That's not how Things search behaves, it's more of a Raycast concept. That is why a performance penalty is being paid on load regardless of whether you depend on ancestor tagging for search. Gonna explore a bigger UI change so that the Raycast search bar isn't so overloaded and filtering works more like Things native UI. 

**Edit 2:** Yeah, inheritance is quite complicated. It's not just todo -> project -> area, tags have parent relationships too. As pondered above I am going to propose a bigger change, since functional parity with Things with the current extension UI would see a perf cost of more like 50%.

pernielsentikaer:
Are we waiting for a fix here?

janosorcsik:
#24489 PR will add the tag filter feature. cc: @pernielsentikaer

bendrucker:
Yes! I might have some time this afternoon to finish ⏫  up.

PR #24335, opened 2026-01-09
Add browser-profile-switcher extension

sajadabedi:
## Checklist

### DIFF
--- a/extensions/browser-profile-switcher/src/switch-to.ts
+++ b/extensions/browser-profile-switcher/src/switch-to.ts
@@ -1,10 +1,7 @@
 import { getPreferenceValues, showHUD } from "@raycast/api";
 import { execSync } from "child_process";
 
-interface Preferences {
-  defaultBrowser: string;
-  showToastMessages: boolean;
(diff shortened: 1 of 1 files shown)


sajadabedi:
@raycastbot Thanks for pointing that out! I definitely want to avoid duplication.

However, the goal of this is to create a different experience rather than just a **list of profiles**. I'm aiming for a workflow similar to **Arc Browser's 'Spaces.'** The idea is to allow users to switch between profiles in their default browser using a single, quick shortcut.

0xdhrv:
> @raycastbot Thanks for pointing that out! I definitely want to avoid duplication.
> 
> However, the goal of this is to create a different experience rather than just a **list of profiles**. I'm aiming for a workflow similar to **Arc Browser's 'Spaces.'** The idea is to allow users to switch between profiles in their default browser using a single, quick shortcut.

Yes. If possible, let's enhance the existing browser profile by adding a new command that matches your workflow. The functionality already exists; we just need to improve the user experience.

PR #24355, opened 2026-01-10
Add DPI Converter extension

Qualigraphe:
## Description

DPI Converter is a utility extension that converts between pixels, centimeters, millimeters, and inches with customizable DPI support.

**Features:**
- Convert single or dual dimensions (width × height)
- Support for pixels (px), centimeters (cm), millimeters (mm), inches (in)
- Customizable DPI resolution (default: 300 DPI)
- Detailed step-by-step calculations
- Quick copy actions with keyboard shortcuts

**Use Cases:**
Perfect for designers, publications editors and printing houses who need to convert units for print design (A4, US Letter…). This extension allows them, for example, to know how many pixels are the minimum required to print at a good resolution.

## Screencast

Screenshots are included in the `metadata/` folder showing:
1. The conversion form with input fields
2. The detailed calculation results with all unit conversions

## Checklist

- [x] I checked that assets used in the `README` are located outside the metadata folder if they were not generated with our metadata

### DIFF
--- a/extensions/dpi-converter/package.json
+++ b/extensions/dpi-converter/package.json
@@ -2,7 +2,7 @@
   "$schema": "https://www.raycast.com/schemas/extension.json",
   "name": "dpi-converter",
   "title": "DPI Converter",
-  "description": "Convert pixels in centimeters, millimeter and inches, and vice-versa! Useful for printers.",
+  "description": "Convert pixels to centimeters, millimeters, inches, and vice-versa! Useful for printers.",
   "icon": "dpi-converter-icon.png",
   "author": "Qualigraphe",
   "owner": "Qualigraphe",
@@ -18,13 +18,12 @@
     {
       "name": "convert-to-px-cm-mm-in",
       "title": "Convert to px, cm, mm, in",
-      "description": "Convert px in cm, mm, in, and vice-versa.",
+      "description": "Convert pixels to centimeters, millimeters, inches, and vice-versa.",
       "mode": "view"
     }
   ],
   "dependencies": {
-    "@raycast/api": "^1.104.1",
-    "@raycast/utils": "^1.17.0"
+    "@raycast/api": "^1.104.1"
   },
   "devDependencies": {
     "@raycast/eslint-config": "^2.0.4",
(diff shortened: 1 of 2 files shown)


Qualigraphe:
Hi @0xdhrv,

Thank you so much for taking the time to review my contribution and for your feedback! 🙏 I really appreciate the thoughtful response and completely understand the concern about avoiding duplication.

I'd love to clarify why I believe DPI Converter addresses a specific workflow gap that Calculator doesn't currently cover, based on my daily work with designers and print professionals.

### Real-World Use Cases
As a designer and trainer working daily with print professionals, these are critical daily workflows:

Example 1: Large Format Print
   - Print shop needs a banner: "400cm × 300cm" (4m × 3m banner viewed from distance)
   - Designer asks: "How many pixels do I need at **75 DPI** for outdoor signage?"
   - Answer: 11,811 × 8,858 px
   - Formula used: px = (cm / 1 inch) × DPI

Example 2: Video Screenshot to Print
   - Designer has a video screenshot: "1920 × 1080 px"
   - Client asks: "What's the maximum print size at **300 DPI** for good quality?"
   - Answer: 16.26 × 9.14 cm (approximately 6.4 × 3.6 inches)
   - Formula used: cm = pixels × (1 inch / DPI)

​These examples demonstrate why DPI context is essential – you need different pixel dimensions based on viewing distance and print technology, and you need to know printable dimensions from existing pixel assets.
​
### Why This is Unique to DPI Converter
This is a unique capability that Calculator does not currently offer. Calculator handles standard unit conversions excellently, but it has no mechanism to perform resolution-dependent conversions where the relationship between pixels and physical units depends on a third parameter (DPI). Queries like those in the example above simply cannot be resolved without knowing the target resolution.

### Potential Integration into Calculator
If Calculator could support this syntax natively, that would be ideal. Maybe by typing something like this (though I'm concerned it might become too complicated for the syntax):
   - `1024 px to cm @300dpi`
   - ​`4m to px at 75dpi`

If no DPI is specified, defaulting to 300 DPI makes sense because it's the universal standard for commercial printers, photo labs, and professional print houses.
   - `1920 px to cm `→ defaults to 300 DPI

​

Qualigraphe:
Hello @0xdhrv, could you please confirm if my explanations were clear?

PR #24440, opened 2026-01-13
Fix browsers-profiles extension chromium profile discovery

nilleb:
## Description

An alternative to #24434 that does not revert the code.

## Checklist

### DIFF
--- a/extensions/browsers-profiles/CHANGELOG.md
+++ b/extensions/browsers-profiles/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Raycast Browsers Profiles
 
+## [Fix Chromium profile discovery logics] - {PR_MERGE_DATE}
+
+- Fallback to original state file discovery if the Preferences folder does not exist
+
 ## [Allow the user to filter out non-interesting profiles] - 2026-01-13
 
 - Adding an option to let the users select only a subset of the available browsers


nilleb:
@pernielsentikaer for your attention in case you're around
and @chingweih for review

pernielsentikaer:
Does anything from Greptile make sense to look into @nilleb 

Also @chingweih, can you check out this solution?

extensions/browsers-profiles/src/lib/chromium.ts
```diff
@@ -59,21 +59,35 @@ export const getChromiumProfiles = (filter: string[]) => {
...
+        browserProfiles.push({
+          type: browser.type,
+          browser: browser.title,
+          app: browser.app,
+          path: directory,
+          name,
+          label: name,
+          uid: directory,
+          icon: browser.icon,
+        });
```
chingweih, line 89: I think we might need to catch if `name` is undefined, otherwise `sortProfiles` will throw error since it uses `.toLowerCase()`

For example

```ts
if (profileName !== undefined) {
  browserProfiles.push({
    type: browser.type,
    browser: browser.title,
    app: browser.app,
    path: directory,
    name: profileName,
    uid: directory,
    label: profileLabel,
    icon: browser.icon,
  });
}
```

nilleb:
> Note that the original Local State solution already fix the profile display name issue.

Okay, let's go with @chingweih implementation! 🤞🏻

PR #24476, opened 2026-01-14
fix(spotify-player): handle disabled menu bar command error

kevintraver:
## Summary
- Fixes error toast appearing when Menu Bar Player command is disabled
- The error check was looking for "must be activated" but Raycast throws "No enabled command" when user disables the command

## Test plan
- [ ] Disable "Menu Bar Player" command in Raycast settings
- [ ] Use any other spotify-player command (e.g., Now Playing)
- [ ] Verify no error toast appears


pernielsentikaer:
Hi @kevintraver 👋 

Could you add an entry in the `CHANGELOG.md` file? Here's our guidelines about that: [Version History](https://developers.raycast.com/basics/prepare-an-extension-for-store).

kevintraver:
This is a follow-up to #23374 which fixed the error toast when Menu Bar Player is disabled.

That fix catches the `"must be activated"` error message, but Raycast also throws `"No enabled command"` in some cases (possibly depending on Raycast version or how the command was disabled). This PR adds handling for that error message as well.

PR #24486, opened 2026-01-14
Add Open, Copy and Download actions to Immich Assets view

julien-wff:
## Description

When viewing an asset, add actions to copy the full-res image to clipboard, and to save it to the `Downloads` folder.

## Screencast

https://github.com/user-attachments/assets/50ec4c1c-82c0-4bf1-a55d-212481dd6ed2

## Checklist

### DIFF
--- a/extensions/immich/CHANGELOG.md
+++ b/extensions/immich/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Immich Changelog
 
-## [Open, Copy and Download assets] - {PR_MERGE_DATE}
+## [Open, Copy and Download assets] - 2026-01-14
 
 - Add Open in Immich, Copy and Download actions for assets in the Asset View
 


xmok (changes requested):
Hi @julien-wff:

Great update!

Only prob is downloading high-res images takes time so it's not evident if something is happening all the time.

I suggest using a stream allowing us to provide a download percentage in `toast`.

Here is a PoC:
```
async function fetchFileToDisk(filePath: string, url: string, onProgress?: (percent: number) => void) {
  const res = await fetch(url);
  if (!res.ok) throw new Error(`Failed to fetch: ${res.status} ${res.statusText}`);

  const total = parseInt(res.headers.get("content-length") || "0");
  const reader = res.body!.getReader();
  const chunks: Uint8Array[] = [];
  let loaded = 0;

  while (true) {
    const { done, value } = await reader.read();
    if (done) break;
    chunks.push(value);
    loaded += value.length;
    onProgress?.(Math.round((loaded / total) * 100));
  }

  const buffer = Buffer.concat(chunks.map(c => Buffer.from(c)));
  await writeFile(filePath, buffer);
}

export async function downloadAndCopyImage(url: string, filename: string) {
  const toast = await showToast(Toast.Style.Animated, "Downloading image", "Please wait...");
  try {
    const path = join(await realpath(tmpdir()), safeFileName(filename));
    await fetchFileToDisk(path, url, (percent) => {
      toast.message = `${percent}%`;
    });
    await Clipboard.copy({ file: path });
    await showHUD("Image copied to clipboard");
  } catch (error) {
    toast.style = Toast.Style.Failure;
    toast.title = "Download failed";
    toast.message = `${error}`;
  }
}
```

pernielsentikaer:
That's a good idea @xmok 

@julien-wff can you look into this?

julien-wff:
Hi @xmok and @pernielsentikaer,

That was a great idea, thanks for the Poc! Now implemented.
I also added an Open to Web action

https://github.com/user-attachments/assets/7ed82c4b-4cf8-468a-b948-c5921b4d04fe

PR #24736, opened 2026-01-21
Update fancy-text extension

nexxai:
## Description

I was trying to figure out why it always took so long to load the fancy fonts every time the extension was opened.  I figured it was reaching out to an external service and it was just the delay from that.  But everything in this extension is done locally, so there's no need for a useEffect or timeout.  

As you can see in the video recording, the first run is with the extension as it currently exists and there is a noticeable delay between selecting "Fancy Text" and the list populating; the second run shows my edited version and it loads nearly instantly.  All of the fonts are stored in the local `src/fonts.json` file which is loaded as an import in the `src/generate-fancy-text.tsx` file so there's no reason to pretend like it's being loaded from a remote source.

## Screencast

https://github.com/user-attachments/assets/8d7e7662-96de-4de0-903a-bfe93486860f

## Checklist

### DIFF
--- a/extensions/fancy-text/CHANGELOG.md
+++ b/extensions/fancy-text/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Fancy Text Changelog
 
+## [Improvement] {PR_MERGE_DATE}
+
+- Removed the unnecessary useEffect for near instant performance
+- Removed the loading state since there's no need for it
+
 ## [Feature] - 2026-01-06
 
 - Added a new action to decorate the generated fancy text


nexxai:
For any reviewer: I updated the CHANGELOG before greptile finished scanning.

ridemountainpig:
@nexxai The useEffect that converts selected text to fancy text is necessary, so please don't remove it. If loading is too slow, maybe we can add a preference to disable the selected-text feature to improve speed.

extensions/fancy-text/src/generate-fancy-text.tsx
```diff
@@ -1,33 +1,15 @@
...
 import FontListItem from "./font-list-item";
 
 export default function Command() {
   const defaultText = "Stay in the flow";
-  const [isLoading, setIsLoading] = React.useState(true);
-  const [searchText, setSearchText] = React.useState("");
+  const [searchText, setSearchText] = React.useState<string>(defaultText);
   const [pinnedFonts, setPinnedFonts] = useCachedState<string[]>("pinned-fonts", []);
 
-  React.useEffect(() => {
```
0xdhrv, line 14: **suggestion**

https://github.com/raycast/extensions/pull/24736#issuecomment-3775993798

0xdhrv:

nexxai:
> The useEffect that converts selected text to fancy text is necessary, so please don't remove it. If loading is too slow, maybe we can add a preference to disable the selected-text feature to improve speed.

Yeah, like I showed in my screen recording, there's an unnecessary delay when opening the extension. We should definitely make the timeout configurable.

pernielsentikaer:
Are you still working on this @nexxai

nexxai:
@pernielsentikaer I was waiting on @ridemountainpig when they said "we can add"

pernielsentikaer:
I think he meant `you` when we said `we` :D

ridemountainpig:
> I think he meant `you` when we said `we` :D

@pernielsentikaer Exactly, thanks for clarifying. @nexxai  I’ve added the resolution to #25146.

PR #24761, opened 2026-01-22
Add timetransfer extension

RoyalXC:
## Checklist

### DIFF
--- a/extensions/timetransfer/.gitignore
+++ b/extensions/timetransfer/.gitignore
@@ -12,3 +12,4 @@ compiled_raycast_rust
 
 # misc
 .DS_Store
+.idea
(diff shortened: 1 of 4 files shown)


ridemountainpig:
Hello, I think there are already many extensions related to timestamp conversion, so perhaps we should update those extensions instead of creating a duplicate.

- [Unix Timestamp Converter](https://www.raycast.com/ride-space/unix-timestamp-converter)
- [Unix Timestamp](https://www.raycast.com/destiner/unix-timestamp)
- [Date Converter](https://www.raycast.com/asportnoy/date-converter)

RoyalXC:
these extensions are not fit chinese style , so i create a new one, It's okay not to accept it, i can use it just by local build




XC
***@***.***

0xdhrv:
> these extensions are not fit chinese style , so i create a new one, It's okay not to accept it, i can use it just by local build XC ***@***.***

PR #24924, opened 2026-01-28
 Avoid overlapping invocations in raytaskwarrior

chromy:
This solves lock errors of the form:
```
Clear working set query: database is locked: Error code 5: The database file is locked
```

Also rename ListActiveTabs to ListActiveTasks matching file name.


pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #24939, opened 2026-01-28
Add control-brightness extension

pavzagor:
## Description

This PR adds a new **Set Brightness** command that allows users to control the brightness of their displays directly from Raycast. It leverages the [Lunar](https://lunar.fyi/) CLI for robust multi-monitor support.

## Key Features

- **Smart Detection**: Automatically detects and selects the display where the cursor is currently located
- **Precise Control**: Set brightness in percent (1-100) for both built-in and external displays
- **Multi-Monitor Support**: Works seamlessly with multiple screens, including XDR/Liquid Retina displays
- **Visual Feedback**: HUD notifications show the brightness change
- **Easy Setup**: Automatically checks for Lunar installation and guides the user to install the CLI if missing

## Screencast

https://github.com/user-attachments/assets/475d5c0d-d3b9-4aa2-8f18-0d3b76efaf31

## Checklist

- [x] I checked that assets used by the README are placed outside of the `metadata` folder


0xdhrv:

We already have an extension in the Store that deals with Brightness Control. Could we consider enhancing the existing extension below instead of creating another one?

- https://www.raycast.com/cali/brightness-control by @CaliCastle

pavzagor:
@0xdhrv I created this exactly because that one wasn't enough and wasn't doing what I wanted to do.

The existing one can only do brightness up and brightness down, which is pretty useless because you have to invoke it, like, multiple times, and it's basically the same as pressing the default Mac buttons to increase and decrease the brightness.

This one can actually detect position on what monitor you are currently and change the brightness only for this monitor.

You can also set the brightness by percentage points, like 100 or 20 or 21 instead of clicking multiple times to set the brightness.

0xdhrv:
> @0xdhrv I created this exactly because that one wasn't enough and wasn't doing what I wanted to do.
> 
> The existing one can only do brightness up and brightness down, which is pretty useless because you have to invoke it, like, multiple times, and it's basically the same as pressing the default Mac buttons to increase and decrease the brightness.
> 
> This one can actually detect position on what monitor you are currently and change the brightness only for this monitor.
> 
> You can also set the brightness by percentage points, like 100 or 20 or 21 instead of clicking multiple times to set the brightness.

That makes sense, but you can improve the existing one by adding or changing the behavior to suit your needs.

pavzagor:
https://github.com/raycast/extensions/pull/25315

PR #24998, opened 2026-01-30
Update bitwarden extension

0xdhrv:
## Description

**Security:** Stops the master password from appearing in error logs when users use “Copy Last Errors” or share bug reports.

- **Cause:** Unlock/login failures were stored as raw exec errors. Execa’s error message includes the full CLI command (`bw unlock <password> --raw`), so the password was included in copied logs.
- **Change:** Unlock and login errors are now sanitized with `treatError()` before being passed to `captureException()`, so only a redacted string is stored. Passwords are replaced with `[REDACTED]`, and special characters in the password are escaped so redaction is reliable.
- **Other:** Resolved npm install peer dependency (react-devtools) and updated shortcut types (`windows` → `Windows`) for current @raycast/api.

## Screencast

N/A — security/logging fix only; no UI or flow changes.

## Checklist

### DIFF
--- a/extensions/bitwarden/CHANGELOG.md
+++ b/extensions/bitwarden/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Bitwarden Changelog
 
+## [Security & Fixes] - {PR_MERGE_DATE}
+
+- **Security:** Redact master password from error logs so it is never exposed when using "Copy Last Errors" or sharing bug reports (unlock/login failures now store a sanitized message)
+- Fix npm install peer dependency conflict by pinning react-devtools to the version expected by @raycast/api
+
 ## [Added PasteUsernameAction] - 2026-01-01
 
 ## [Added support for Windows] - 2025-10-15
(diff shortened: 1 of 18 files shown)


pernielsentikaer:
@0xdhrv is this ready 🙂

@jomifepe can you check this?

jomifepe:
Hey! Thanks for contributing. 

I mean, I'm ok with this change, even though the `Copy Last Errors` action is mainly for the user to debug their own problem. I also warn about the risk of sharing it using the following message:

`Please be mindful of where you share this error log, as it may contain sensitive information. Always analyze it before sharing.`

But if that sounds risky, I'm ok with redacting the value, so LGTM!

0xdhrv:
> Hey! Thanks for contributing.
> 
> I mean, I'm ok with this change, even though the `Copy Last Errors` action is mainly for the user to debug their own problem. I also warn about the risk of sharing it using the following message:
> 
> `Please be mindful of where you share this error log, as it may contain sensitive information. Always analyze it before sharing.`
> 
> But if that sounds risky, I'm ok with redacting the value, so LGTM!

I will remove those, thanks for the review!

PR #25015, opened 2026-01-30
Update notion extension

parterburn:
## Description

Adds support for up to 2 Notion accounts. It's very seamless to enable the multi-account setup, just add a label for your two accounts in the extension settings and when you go to use the extension it will ask you to OAuth into both of them, separately.

If you don't have 2 Notion accounts, you simply leave the second label empty and it works exactly how it does today.

## Screencast

image
image
image
image
image

## Checklist


pernielsentikaer:
Hi 👋

Do you have a screencast of this? Adding another account might add extra complexity?

pernielsentikaer:
Also, could you resolve the merge conflict

pernielsentikaer:
Hi 👋

I really like the idea! Support for multiple accounts isn't something we'll introduce right away, so it might make sense to do something like you're suggesting, which generally looks like a solid way to handle it.

What’s not working right now is the way it's managed. The current method isn't really effective. What if we added an `Action` that pushes to a new view where it shows a list to add new account(s)? That way, it's still a bit hidden but offers a better experience.

Another idea is to create a new command to `List Accounts`, which would do the same as above.

What do you think?

PR #25026, opened 2026-01-31
Add link-groups extension

matheuscoelhomalta:
## Description

Link Groups provides a focused workflow for creating collections of URLs and opening an entire collection at once. After comparing it with [Pins](https://www.raycast.com/HelloImSteven/pins), there is substantial overlap in the core outcome, but the extensions differ in scope and in how users trigger a group.

### Comparison with Pins

| Area | Link Groups | Pins |
| --- | --- | --- |
| Scope | Purpose-built for groups of web URLs | General-purpose pins for URLs, files, folders, apps, Terminal commands, and scripts |
| Primary experience | Focused Raycast list for managing and launching link groups | Menu bar experience plus separate commands for creating and managing pins and groups |
| Group structure | Simple, flat groups | Groups and nested subgroups with richer organization controls |
| Open a whole group | First-class action plus a dedicated `Open Link Group` no-view command | `Open All` from the menu bar/group UI; no dedicated group-specific no-view command |
| Quicklink and deeplink workflow | Creates a group-specific Quicklink backed by a deeplink and stable group ID | No equivalent built-in action for generating a group-specific Quicklink/deeplink |
| Browser selection | Optional browser override shared by every URL in a group | Global preferred browser, with an application override configured per pin rather than per group |
| Bulk entry | Paste newline-separated URLs; titles are generated automatically and invalid URLs are skipped | Imports structured Pins data from JSON, CSV, TOML, YAML, or XML rather than accepting a raw URL list |
| Opening implementation | Opens all URLs concurrently and reports partial failures | Opens the pins in a group sequentially |
| Customization | Intentionally minimal: group title, links, and browser | Rich pin/group options including icons, colors, visibility, sorting, placeholders, expiration, and more |

The main unique workflows in this extension are the group-specific no-view command, one-step Quicklink/deeplink generation, group-level browser override, and fast raw-URL bulk import. The underlying create-group-and-open-all use case remains similar to Pins.

## Checklist


0xdhrv:

We already have an extension in the Store that deals with Link Groups. Could we consider enhancing the existing extension below instead of creating another one?

- https://www.raycast.com/HelloImSteven/pins by @SKaplanOfficial

matheuscoelhomalta:
@0xdhrv I just bothered to check now, but it is really different to the behavior of Pins, maybe it is worthing uploading this one

0xdhrv:
> @0xdhrv I just bothered to check now, but it is really different to the behavior of Pins, maybe it is worthing uploading this one

Thanks for following up 🔥

I took another look at both extensions. While Link Groups has a more focused interface, its core workflow still overlaps significantly with Pins:

* creating named groups of URLs
* opening all links in a group
* importing multiple URLs
* choosing which browser opens the links
* managing groups directly in Raycast

The main differences I can see are the dedicated no-view command for opening a group, deeplink/Quicklink support, and an optional browser choice per group. These are useful improvements, but they seem better suited as additions to the existing Pins extension rather than enough differentiation for a separate Store extension.

Could we consider contributing these workflows to Pins instead?

In particular, the following would be valuable additions:

* an “Open Pin Group” no-view command
* deeplink or Quicklink generation for individual groups
* an optional browser override per group
* a faster bulk URL import flow

This would keep the functionality consolidated, improve discoverability, and avoid having two extensions that provide substantially the same user outcome.

matheuscoelhomalta:
@0xdhrv Thanks for the feedback. I followed the suggestion and opened [PR #30047](https://github.com/raycast/extensions/pull/30047), which adds the missing Link Groups workflows directly to Pins, including group Quicklinks, deeplinks, per-group browser preferences, and bulk URL import.

0xdhrv:
Closing this in favor of https://github.com/raycast/extensions/pull/30047

PR #25032, opened 2026-01-31
fix(ente-auth): don't delete export file before re-exporting

### LINKED ISSUE #20619: [Ente Auth] Import secrets deletes ente_auth.txt export file, then gives no such file or directory
### Extension

https://www.raycast.com/chkpwd/ente-auth

### Raycast Version

1.101.1

### macOS Version

15.5

### Description

Import secrets stopped working; it deletes the file it should import, then fails as the file is not found.

If I change the file flags (`chflags uchg ~/.ente/auth/exports/ente_auth.txt'`), making it immutable, import works correctly.

Thanks

```
Error: ENOENT: no such file or directory, open '/Users/xxx/.ente/auth/exports/ente_auth.txt'

w4:export.tsx:28:31

---
25:     showFailureToast(error, { title: "Export failed" });
26:   }
27: 
28:   const secrets = parseSecrets(getSecrets(EXPORT_FILE_PATH));
29:   const secretsList = secrets
30:     .map((secret) => `- ${secret.issuer.replaceAll("+", " ")}  - \`${secret.username}\`\n`)
31:     .join("");
---

hr:index.js:7:3106
    at ray-navigation-stack (<anonymous>)
Ho:index.js:7:2607
```

### Steps To Reproduce

1. Try to import ente secrets (after export via cli)
2. The import fails

### Current Behaviour

_No response_

### Expected Behaviour

The import should work and should not delete the file before importing.

xela92:
## Summary

Fixes #20619

The `Import Secrets` command calls `exportEnteAuthSecrets()` which first deletes the existing export file via `fse.removeSync()`, then attempts to run `ente export`. If the export CLI fails for any reason, the original file is already gone, causing an `ENOENT: no such file or directory` error when trying to read it.

## Changes

- **`src/helpers/ente.ts`**: Removed the pre-emptive `fse.removeSync(EXPORT_FILE_PATH)` in `exportEnteAuthSecrets()`. The `ente export` CLI can overwrite the file itself — there is no need to delete it beforehand.
- **`src/export.tsx`**: Added missing `return` in the catch block so execution doesn't continue to read a non-existent file after an export failure.

## Root Cause

The race condition was:
1. `fse.removeSync()` deletes `ente_auth.txt`
2. `ente export` fails (for any reason)
3. Import proceeds with "existing file" — but it was already deleted in step 1
4. `ENOENT` 💥

As the I noted, making the file immutable (`chflags uchg`) worked around the bug because `removeSync` would fail silently, preserving the file.

### DIFF
--- a/extensions/ente-auth/CHANGELOG.md
+++ b/extensions/ente-auth/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Ente Auth Changelog
 
-## [Fix import deleting export file] - {PR_MERGE_DATE}
+## [Fix import deleting export file] - 2026-02-01
 
 - Fixed import command deleting the export file before re-exporting, causing ENOENT if export fails
 - Added missing return in export command error handling


chkpwd:
LGTM

PR #25042, opened 2026-02-01
Add FaviconSync extension

alivault:
## Summary

Adds **FaviconSync** - a simple extension to refresh Safari favicons.

### Features
- **Refresh Favicon**: Clears the cached favicon for the current Safari tab and restarts Safari to fetch a fresh icon

### Why?
Safari caches favicons aggressively and doesn't update them when websites change their icons. This results in outdated or broken favicon images in tabs and the start page. FaviconSync fixes this with a single command.

### How It Works
1. Gets the current Safari tab URL via AppleScript
2. Quits Safari
3. Deletes the favicon from Safari's cache directories
4. Reopens Safari (tabs restore automatically)

### Requirements
- macOS 13.0 (Ventura) or later
- Full Disk Access permission for Raycast (to modify Safari's favicon cache)

### Checklist
- [x] Extension passes `ray lint`
- [x] 512x512 PNG icon included
- [x] README with description and setup instructions
- [x] No API keys or secrets required

### DIFF
--- a/extensions/faviconsync/src/refresh-favicon.ts
+++ b/extensions/faviconsync/src/refresh-favicon.ts
@@ -1,15 +1,19 @@
 import { showHUD, showToast, Toast } from "@raycast/api";
 import { exec } from "child_process";
 import { promisify } from "util";
-import { existsSync, readdirSync, unlinkSync, statSync } from "fs";
+import { existsSync, unlinkSync } from "fs";
 import { join } from "path";
 import { homedir } from "os";
+import { createHash } from "crypto";
 
 const execAsync = promisify(exec);
 
 // Safari cache directories
 const TOUCH_ICONS_CACHE = join(homedir(), "Library/Safari/Touch Icons Cache");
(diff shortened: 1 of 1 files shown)


alivault:
CHANGELOG was added

Ernest0-Production:
Hi! @alivault . Thanks for contributions!

Why not add this solution to [Safari Extension](https://www.raycast.com/loris/safari)?

Moreover, I find the name FaviconSync confusing.
- The extension only works for Safari.
- The extension name doesn't comply with the guidelines: https://developers.raycast.com/basics/prepare-an-extension-for-store#extensions-and-commands-naming.
- If I needed to find a similar feature, I probably wouldn't find it by simply searching for "Safari" and "Icon." The package.json file is missing keywords.

IMHO, it would be more convenient to see this action in Safari Extensions. Perhaps in the future, clearing the entire cache wouldn't hurt (and doing it as a separate extension would be just as much overhead).

alivault:
Hi! Thanks for the feedback. You raise good points. I'm happy to close this PR and see if I can add this to Safari Extension.

PR #25311, opened 2026-02-10
Add aqi-air-quality extension

riccardogiorato:
## Description

Adds a menu bar extension that displays real-time Air Quality Index (AQI) information with support for both European (EAQI) and US (EPA) scales. 

The extension automatically detects your location via IP geolocation or allows manual location input, fetching data from multiple sources including Open-Meteo, IQAir (with optional API key), and EEA modelled data for European locations. 

Features include pollutant details (PM2.5, PM10, gases), UV index guidance, and comprehensive provider reporting—all accessible from your Mac menu bar with hourly automatic refresh.

## Checklist


0xdhrv (commented):

We already have an extension in the Store that deals with AQI. Could we consider enhancing the existing extension below instead of creating another one?

- https://www.raycast.com/earthpyy/air-quality by @earthpyy

riccardogiorato:
Hey @0xdhrv this one is much more flexible cause it focuses on having multiple APIs to get data from to not only from a single source of truth. 

That's why I get data from:
1. Open-Meteo
2. IQAir (with optional API key)
3. EEA modelled data for European locations
4. I will add more sources also based on other common API providers from other parts of the world like US or Asia

Also main point of this extension is that we also automatically get the user location from user IP or you can specify a city/name and we will automatically get lat lng to get correct air quality.

The other extension main issue is that it only focuses on nearby stations but this leads to totally different results compared to a more correct avg value for a city.

Also this extension mainly focuses on being inside the menu bar and giving just enough info using user defined scale EuAQI or US AQI (maybe in the future also chinese AQI)


I can bring these features to the other extension but I would need to change it a lot and make the provider choice a dropdown choice to be more flexibile to not only use "World Air Quality Index" but apis we are using here.

Let me know what we can do cause I also don't know if the owner of the other extension wants me to make these changes to his extension maybe? 

Let me know.

earthpyy:
> Hey @0xdhrv this one is much more flexible cause it focuses on having multiple APIs to get data from to not only from a single source of truth.
> 
> That's why I get data from:
> 
> 1. Open-Meteo
> 2. IQAir (with optional API key)
> 3. EEA modelled data for European locations
> 4. I will add more sources also based on other common API providers from other parts of the world like US or Asia
> 
> Also main point of this extension is that we also automatically get the user location from user IP or you can specify a city/name and we will automatically get lat lng to get correct air quality.
> 
> The other extension main issue is that it only focuses on nearby stations but this leads to totally different results compared to a more correct avg value for a city.
> 
> Also this extension mainly focuses on being inside the menu bar and giving just enough info using user defined scale EuAQI or US AQI (maybe in the future also chinese AQI)
> 
> I can bring these features to the other extension but I would need to change it a lot and make the provider choice a dropdown choice to be more flexibile to not only use "World Air Quality Index" but apis we are using here.
> 
> Let me know what we can do cause I also don't know if the owner of the other extension wants me to make these changes to his extension maybe?
> 
> Let me know.

Hey @riccardogiorato, the owner of original extension here.

First, I want to clarify a bit about the Air Quality extension. The nearby station option is just a default choice when a user doesn't specify a particular station or city. It can provides accurate information when a city or a station is selected. The extension also features a menu bar that displays brief information about air quality as well!

However, it still lacks features and has rooms of improvement such as adding multiple providers and additional units, as you mentioned.

Based on the feature you've described, which is very similar to the original one, I would be very happy if you could bring these awesome changes into the original extension and I'm eager to help make it happen.

Let me know on what do you think. Thanks!

PR #25349, opened 2026-02-11
Spotify Player: Fix menu bar unloading before API fetch completes

hugodemenez:
## Description

Fixes the menu bar icon disappearing permanently when "Hide icon while idle" is enabled and Spotify is restarted.

Previously, a mutable ref (`shouldExecute`) controlled when `useCurrentlyPlaying` executes. Since refs don't trigger re-renders, the hook would not re-execute when the ref value changed — causing the menu bar to unload before the API fetch completed and never reappear.

Fixes #25291

## Changes

- **Replace ref-based execution control with computed reactive values**: The `execute` condition is now derived from reactive state (`isSpotifyActive`, `uriChanged`), ensuring the hook re-evaluates when Spotify state changes.
- **Render a loading `MenuBarExtra`** with `isLoading={true}` to keep the menu bar mounted while data is still being fetched (even when the user has "Hide menu bar icon on idle" enabled).
- **Clean up `useEffect` dependencies**: Properly scoped dependency array for the URI sync effect.

## Screencast

https://github.com/user-attachments/assets/c410b81c-fc3a-4b61-824d-34b2050838ef

## Checklist

### DIFF
--- a/extensions/spotify-player/CHANGELOG.md
+++ b/extensions/spotify-player/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Spotify Player Changelog
 
-## [Fix Menu Bar Unloading Before API Fetch Completes] - {PR_MERGE_DATE}
+## [Fix Menu Bar Unloading Before API Fetch Completes] - 2026-02-17
 
 - Fixed the menu bar icon disappearing permanently when "Hide icon while idle" is enabled and Spotify is restarted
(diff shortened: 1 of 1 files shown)


hugodemenez:
@pernielsentikaer if you have the opportunity to take a look at this fix. Thank you 🙏  

The issue primarily stemmed from the early return of null when using the "hide on idle" preference (which unloads directly without waiting for the promises to resolve https://github.com/raycast/extensions/issues/25291#issuecomment-3883385923) and the use of `ref`, which also caused problems.

pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #25361, opened 2026-02-12
Deep Search Claude Sessions and Path Resolution Edge Cases

RoderickJDunn:
## Description

Fixes project path resolution and adds full-content session search to ClaudeCast.

### Bug Fix: Project Path Resolution

Claude Code encodes project directory names by replacing both `/` and `.` with `-`, making the encoding lossy and non-reversible. The previous naive decode (replacing all `-` with `/`) produced incorrect paths for:
- Usernames containing dots (e.g. `steph.curry` → `steph/curry`)
- Project names containing dashes (e.g. `helm-charts` → `helm/charts`)

This caused "Project path no longer exists" errors when trying to resume sessions from Browse Sessions.

**Fix:** Three-tier path resolution:
1. **sessions-index.json** — reads the authoritative `originalPath` written by Claude Code
2. **Filesystem-guided walk** — anchors on `os.homedir()`, then greedily matches path components against real directory entries (longest match first to preserve literal dashes)
3. **Naive decode** — last resort fallback

Also fixed `encodeProjectPath` to correctly encode both `/` and `.`.

### New Command: Deep Search Sessions

Searches through full JSONL session content across all Claude Code conversations.

- Streams through session files incrementally — results appear as they're found
- Searches all message content (user + assistant) case-insensitively
- Debounced input (300ms, min 3 chars) with AbortSignal cancellation when query changes
- Short-circuits on first match per session file for performance
- Same session actions as Browse Sessions (Resume, Fork, View Details, Copy, Delete)

## Checklist

### DIFF
--- a/extensions/claudecast/CHANGELOG.md
+++ b/extensions/claudecast/CHANGELOG.md
@@ -1,6 +1,6 @@
 # ClaudeCast Changelog
 
-## [1.2.0] - {PR_MERGE_DATE}
+## [1.2.0] - 2026-02-20
 
 ### Added
 


extensions/claudecast/src/lib/session-parser.ts
```diff
@@ -46,19 +46,157 @@ const CLAUDE_DIR = path.join(os.homedir(), ".claude");
...
+
+    // Build the component name; a leading empty part means "." prefix
+    // e.g. ["", "claude"] from "--claude" → ".claude"
+    let component = componentParts.join("-");
+    if (component.startsWith("-")) {
+      component = "." + component.slice(1);
+    } else if (component === "") {
+      // Single empty part — skip (a bare "." by itself is not useful)
+      continue;
+    }
```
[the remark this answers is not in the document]
RoderickJDunn: So in conclusion this is fine, its just potentially a bit of extra work

extensions/claudecast/src/deep-search-sessions.tsx
```diff
@@ -0,0 +1,418 @@
...
+      await showToast({
+        style: Toast.Style.Animated,
+        title: "Deleting session...",
+      });
+      try {
+        await deleteSession(session.id);
+        await showToast({
+          style: Toast.Style.Success,
+          title: "Session deleted",
+        });
```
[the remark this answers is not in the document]
RoderickJDunn: fixed

extensions/claudecast/src/deep-search-sessions.tsx
```diff
@@ -0,0 +1,431 @@
...
+        (session) => {
+          if (controller.signal.aborted) return;
+          // Deduplicate by session ID
+          if (seenIds.has(session.id)) return;
+          seenIds.add(session.id);
+          setResults((prev) => [...prev, session]);
+        },
+        controller.signal,
+      );
+    } catch {
```
qazi0: **CRITICAL — Blanket catch swallows all errors**

This catch block treats AbortErrors (expected cancellation) and real bugs (TypeError, RangeError from recursive `walkPathSegments`, etc.) identically — all silently ignored. If the search infrastructure has a real bug, users see zero results with no indication that the search itself failed.

Suggestion: Distinguish between abort (expected) and unexpected errors:
```typescript
} catch (error: unknown) {
  if (controller.signal.aborted) return; // Expected cancellation
  console.error("Deep search failed:", error);
  await showToast({
    style: Toast.Style.Failure,
    title: "Search failed",
    message: error instanceof Error ? error.message : "An unexpected error occurred",
  });
}
```

qazi0: **CRITICAL — Blanket catch swallows all errors**

This catch block treats AbortErrors (expected cancellation) and real bugs (TypeError, RangeError from recursive `walkPathSegments`, etc.) identically — all silently ignored. If the search infrastructure has a real bug, users see zero results with no indication that the search itself failed.

Suggestion: Distinguish between abort (expected) and unexpected errors:
```typescript
} catch (error: unknown) {
  if (controller.signal.aborted) return; // Expected cancellation
  console.error("Deep search failed:", error);
  await showToast({
    style: Toast.Style.Failure,
    title: "Search failed",
    message: error instanceof Error ? error.message : "An unexpected error occurred",
  });
}
```

extensions/claudecast/src/deep-search-sessions.tsx
```diff
@@ -0,0 +1,431 @@
...
+
+const MIN_QUERY_LENGTH = 3;
+const DEBOUNCE_MS = 300;
+
+export default function DeepSearchSessions() {
+  const [isSearching, setIsSearching] = useState(false);
+  const [results, setResults] = useState<SessionMetadata[]>([]);
+  const [searchText, setSearchText] = useState("");
+  const abortRef = useRef<AbortController | null>(null);
+  const debounceRef = useRef<NodeJS.Timeout | null>(null);
```
qazi0, line 45: **IMPORTANT — No cleanup on component unmount**

Neither `debounceRef` nor `abortRef` are cleaned up on unmount. If the user navigates away during a debounce window or active search, `performSearch` will call `setResults`/`setIsSearching` on an unmounted component.

Add a cleanup effect:
```typescript
useEffect(() => {
  return () => {
    if (debounceRef.current) clearTimeout(debounceRef.current);
    abortRef.current?.abort();
  };
}, []);
```

extensions/claudecast/src/deep-search-sessions.tsx
```diff
@@ -0,0 +1,431 @@
...
+  );
+
+  const emptyDescription =
+    searchText.length === 0
+      ? "Type at least 3 characters to search session content"
+      : searchText.length < MIN_QUERY_LENGTH
+        ? `Type ${MIN_QUERY_LENGTH - searchText.length} more character${MIN_QUERY_LENGTH - searchText.length === 1 ? "" : "s"} to start searching`
+        : isSearching
+          ? "Searching session content..."
+          : `No sessions matched "${searchText}"`;
```
qazi0: **Suggestion — Nested ternary**

Triple-nested ternary with an embedded fourth ternary (pluralization). Consider a helper function with early returns for readability:
```typescript
function getEmptyDescription(searchText: string, isSearching: boolean): string {
  if (searchText.length === 0) return "Type at least 3 characters to search session content";
  if (searchText.length < MIN_QUERY_LENGTH) {
    const remaining = MIN_QUERY_LENGTH - searchText.length;
    return `Type ${remaining} more character${remaining === 1 ? "" : "s"} to start searching`;
  }
  if (isSearching) return "Searching session content...";
  return `No sessions matched "${searchText}"`;
}
```

extensions/claudecast/src/deep-search-sessions.tsx
```diff
@@ -0,0 +1,431 @@
...
+}) {
+  const [isLoading, setIsLoading] = useState(true);
+  const [session, setSession] = useState<SessionDetail | null>(null);
+
+  useEffect(() => {
+    (async () => {
+      const detail = await getSessionDetail(sessionId);
+      setSession(detail);
+      setIsLoading(false);
+    })();
```
qazi0, line 327: **HIGH — SessionDetailView has no error handling**

This async IIFE has no error handling. If `getSessionDetail` throws (stream errors, stat failures), the promise rejection is unhandled and the component stays in the loading state forever — infinite spinner with no explanation.

```typescript
useEffect(() => {
  (async () => {
    try {
      const detail = await getSessionDetail(sessionId);
      setSession(detail);
    } catch (err) {
      console.error("Failed to load session detail:", err);
    } finally {
      setIsLoading(false);
    }
  })();
}, [sessionId]);
```
Also: `browse-sessions.tsx` uses a named inner function for this pattern rather than an IIFE — consider matching that style for consistency.

extensions/claudecast/src/lib/session-parser.ts
```diff
@@ -46,19 +46,157 @@ const CLAUDE_DIR = path.join(os.homedir(), ".claude");
...
+      "sessions-index.json",
+    );
+    const content = await fs.promises.readFile(indexPath, "utf8");
+    const index = JSON.parse(content);
+    if (index.originalPath && typeof index.originalPath === "string") {
+      resolvedPathCache.set(encodedDirName, index.originalPath);
+      return index.originalPath;
+    }
+  } catch {
+    // sessions-index.json doesn't exist or is unreadable
```
qazi0: **CRITICAL — Tier 1 catch swallows corrupted JSON and permission errors**

This catch treats `ENOENT` (expected — file doesn't exist), `SyntaxError` (corrupted JSON), and `EACCES` (permissions) identically. For corrupted/permission cases, silently falling through to the filesystem walk may produce a **wrong path that gets cached permanently** in `resolvedPathCache`.

Suggestion: Only silently fall through for `ENOENT`:
```typescript
} catch (error: unknown) {
  if ((error as NodeJS.ErrnoException)?.code !== "ENOENT") {
    console.warn(`Failed to read sessions-index.json for ${encodedDirName}:`, error);
  }
}
```

extensions/claudecast/src/lib/session-parser.ts
```diff
@@ -46,19 +46,157 @@ const CLAUDE_DIR = path.join(os.homedir(), ".claude");
...
+ * so the result may be wrong. Prefer resolveProjectPath() which reads
+ * sessions-index.json for the authoritative original path.
  */
 export function decodeProjectPath(encodedPath: string): string {
   // Replace leading dash and all dashes with forward slashes
   return "/" + encodedPath.slice(1).replace(/-/g, "/");
 }
 
+// Cache resolved paths to avoid repeated fs reads
+const resolvedPathCache = new Map<string, string>();
```
qazi0, line 60: **MEDIUM — Module-level cache never invalidated**

This cache persists for the entire Raycast extension process lifetime (especially the menu bar command which runs continuously). If a user moves a project directory, stale cache entries will keep returning the old path — which is the exact "Project path no longer exists" error this PR fixes.

Consider a TTL (e.g., 5 minutes), or at minimum avoid caching Tier 3 (lossy) results (see line 103).

extensions/claudecast/src/lib/session-parser.ts
```diff
@@ -46,19 +46,157 @@ const CLAUDE_DIR = path.join(os.homedir(), ".claude");
...
+  // 2. Try filesystem-guided resolution
+  const fsResolved = await resolveByFilesystemWalk(encodedDirName);
+  if (fsResolved) {
+    resolvedPathCache.set(encodedDirName, fsResolved);
+    return fsResolved;
+  }
+
+  // 3. Naive decode (last resort)
+  const decoded = decodeProjectPath(encodedDirName);
+  resolvedPathCache.set(encodedDirName, decoded);
```
qazi0: **HIGH — Lossy naive-decode result cached permanently**

The naive decode is documented as "known-lossy" but its result is unconditionally cached as if authoritative. Once cached, even if `sessions-index.json` is later created or permissions are fixed, the wrong answer persists.

Consider not caching Tier 3 results, or only caching them if the decoded path actually exists:
```typescript
const decoded = decodeProjectPath(encodedDirName);
try {
  await fs.promises.access(decoded);
  resolvedPathCache.set(encodedDirName, decoded); // Only cache if path exists
} catch { /* don't poison cache */ }
return decoded;
```

extensions/claudecast/src/lib/session-parser.ts
```diff
@@ -46,19 +46,157 @@ const CLAUDE_DIR = path.join(os.homedir(), ".claude");
...
+        // Last component — any file type is fine
+        return candidatePath;
+      }
+
+      if (stat.isDirectory()) {
+        const resolved = await walkPathSegments(candidatePath, remaining);
+        if (resolved) return resolved;
+      }
+    } catch {
+      // Doesn't exist on disk — try shorter component
```
qazi0: **HIGH — walkPathSegments catches all stat errors, not just ENOENT**

This catches `EACCES` (permission denied), `ELOOP` (symlink loop), and errors from the recursive `walkPathSegments` call — not just "doesn't exist on disk." When `EACCES` is caught, the function tries a shorter component, which may produce a wrong path that gets cached.

Suggestion: Narrow to expected errors:
```typescript
} catch (error: unknown) {
  const code = (error as NodeJS.ErrnoException)?.code;
  if (code !== "ENOENT" && code !== "ENOTDIR") {
    console.warn(`Unexpected error probing ${candidatePath}: ${code}`);
  }
}
```

extensions/claudecast/src/lib/session-parser.ts
```diff
@@ -46,19 +46,157 @@ const CLAUDE_DIR = path.join(os.homedir(), ".claude");
...
+    }
+  }
+
+  return null;
+}
+
 /**
- * Encode a project path to Claude's directory naming format
+ * Encode a project path to Claude's directory naming format.
+ * Claude replaces both / and . with -.
```
qazi0: **MEDIUM — Dot-encoding claim unverified**

This is a high-stakes claim that gates the correctness of the entire resolution pipeline. If wrong, `resolveByFilesystemWalk` would fail its `startsWith(encodedHome)` check for any user whose home directory contains a dot (e.g., `/Users/j.doe`).

Could you add a reference to how this was verified? (e.g., a link to Claude Code source, a test with a dotted path, or the GitHub issue that surfaced this behavior). A version pin in the comment would also help: `"Claude CLI (as of v2.x) replaces both / and . with -"`.

extensions/claudecast/src/lib/session-parser.ts
```diff
@@ -442,6 +580,233 @@ export async function getMostRecentSession(): Promise<SessionMetadata | null> {
...
+      });
+    }
+  }
+}
+
+/**
+ * Search a single session file for a query match.
+ * Reads the entire file (not just first 50 lines) but short-circuits
+ * on first content match, continuing only to collect metadata.
+ */
```
qazi0, line 701: **IMPORTANT — searchSingleSession JSDoc is misleading + performance concern**

The comment says "short-circuits on first match" but the implementation reads the **entire file** after `found = true`. It only skips the `toLowerCase().includes()` comparison — but full `JSON.parse` + content extraction still runs for every remaining line. For large sessions (hundreds of thousands of lines), this is significant.

Two options:
1. Fix the comment to accurately describe what happens: *"Reads the entire file to collect metadata. Sets a match flag on first content hit, skipping redundant string comparisons for subsequent lines."*
2. Actually short-circuit: after finding a match, use approximate metadata (like `parseSessionMetadataFast`'s 50-line approach) and destroy the stream early.

extensions/claudecast/src/lib/session-parser.ts
```diff
@@ -442,6 +580,233 @@ export async function getMostRecentSession(): Promise<SessionMetadata | null> {
...
+  for (const projectDir of projectDirs) {
+    if (signal?.aborted) return;
+    const sessionFiles = await listSessionFiles(projectDir);
+    for (const sessionFile of sessionFiles) {
+      const filePath = path.join(PROJECTS_DIR, projectDir, sessionFile);
+      try {
+        const stat = await fs.promises.stat(filePath);
+        fileInfos.push({ filePath, projectDir, mtime: stat.mtime });
+      } catch {
+        // Skip files we can't stat
```
qazi0: **HIGH — searchSessionContent stat catch silently skips files**

If the system hits `EMFILE` (too many open file descriptors — plausible when iterating hundreds of session files), **every** subsequent `stat` call will also fail. The search silently returns zero/partial results and the user sees "No sessions matched" when in fact the search infrastructure failed.

At minimum, log the error. For `EMFILE`/`ENFILE`, consider aborting with a user-facing error.

qazi0 (commented):
a couple issues

**Critical (2):**
- Blanket catch in `performSearch` swallows all errors including real bugs — users see zero results with no indication of failure
- `resolveProjectPath` Tier 1 catch silently swallows corrupted JSON and permission errors, potentially caching wrong paths

**Important (4):**
- No `useEffect` cleanup for debounce timer and AbortController on unmount
- `searchSingleSession` JSDoc misleadingly claims short-circuit behavior; reads entire file after match
- `SessionDetailView` has no error handling — throws lead to infinite spinner
- Several error handlers discard the error object entirely, hiding I/O failures

**Suggestions (4):**
- Module-level `resolvedPathCache` has no TTL/invalidation
- Dot-encoding claim in `encodeProjectPath` should cite evidence
- Nested ternary → extract helper function
- Use `safeResolve` consistently

See inline comments for details and suggested fixes.

extensions/claudecast/src/lib/session-parser.ts
```diff
@@ -442,6 +580,233 @@ export async function getMostRecentSession(): Promise<SessionMetadata | null> {
...
+
+    rl.on("error", () => {
+      cleanup();
+      signal?.removeEventListener("abort", onAbort);
+      if (!resolved) {
+        resolved = true;
+        resolve(null);
+      }
+    });
+    stream.on("error", () => {
```
qazi0: **HIGH — Stream/readline error handlers discard the error object**

Both error handlers don't even accept the error parameter — they discard it entirely. When a stream encounters I/O failures, encoding errors, or `EMFILE`, the function silently returns `null` ("no match"), making search results silently incomplete.

At minimum, accept and log the error:
```typescript
rl.on("error", (err) => {
  console.warn(`Error reading session ${filePath}:`, err.message);
  cleanup();
  signal?.removeEventListener("abort", onAbort);
  safeResolve(null);
});
```

extensions/claudecast/src/lib/session-parser.ts
```diff
@@ -442,6 +580,233 @@ export async function getMostRecentSession(): Promise<SessionMetadata | null> {
...
+            firstMessage,
+            turnCount,
+            cost: totalCost,
+            model,
+          });
+        }
+      } else {
+        if (!resolved) {
+          resolved = true;
+          resolve(null);
```
qazi0: **MEDIUM — safeResolve used inconsistently in close handler**

`safeResolve` is defined but bypassed in the `close` and `error` handlers, which manually duplicate the `if (!resolved)` guard. Using `safeResolve` consistently would reduce the close handler from 12 lines to 3:
```typescript
rl.on("close", () => {
  signal?.removeEventListener("abort", onAbort);
  safeResolve(found ? { id, summary, firstMessage, turnCount, cost: totalCost, model } : null);
});
```

qazi0 (changes requested):
pls see comments

extensions/claudecast/src/lib/session-parser.ts
RoderickJDunn: Resolved indirectly by another fix

extensions/claudecast/src/lib/session-parser.ts
RoderickJDunn: fixed comment

extensions/claudecast/src/deep-search-sessions.tsx
RoderickJDunn: Added try/catch/finally

RoderickJDunn:
@qazi0 thanks for the thorough review! I think everything is addressed :)

extensions/claudecast/src/lib/session-parser.ts
```diff
@@ -442,6 +595,220 @@ export async function getMostRecentSession(): Promise<SessionMetadata | null> {
   return sessions[0] || null;
 }
 
+/**
```
qazi0, line 629: **Residual from Round 1 — `searchSessionContent` JSDoc still says "short-circuits"**

The `searchSingleSession` JSDoc was correctly updated to say "Reads the entire file to collect full metadata." However, the *caller's* JSDoc here still says "short-circuits on first match per session" — the same misleading claim that migrated upward.

Suggested fix:
```typescript
/**
 * Search all session files for content matching a query string.
 * Iterates through each JSONL file sequentially (most recent first),
 * delegating to searchSingleSession which reads each file in full
 * to collect metadata. Supports cancellation via AbortSignal.
 *
 * @param query - Case-insensitive search string
 * @param onMatch - Called incrementally as matching sessions are found
 * @param signal - AbortSignal to cancel in-flight search
 */
```

extensions/claudecast/src/deep-search-sessions.tsx
```diff
@@ -0,0 +1,457 @@
...
+        </ActionPanel>
+      }
+    />
+  );
+}
+
+function SessionDetailView({
+  sessionId,
+  projectPath,
+}: {
```
qazi0, line 310: **SessionDetailView: error caught but no toast, misleading "Not Found" for load errors**

The try/catch/finally pattern you added here is good — it prevents the infinite spinner. However, when `getSessionDetail` throws (permissions, corrupt JSONL, disk error), the user sees "Session Not Found" which implies the session doesn't exist, when it may just be unreadable.

Consider adding a toast and differentiating the UI state:
```typescript
const [loadError, setLoadError] = useState<string | null>(null);

// In the catch block:
catch (err) {
  console.error("Failed to load session detail:", err);
  setLoadError(err instanceof Error ? err.message : String(err));
  await showToast({
    style: Toast.Style.Failure,
    title: "Failed to load session",
    message: err instanceof Error ? err.message : String(err),
  });
}

// In the render:
if (!session) {
  return (
    <Detail markdown={
      loadError
        ? `# Failed to Load Session\n\nError: ${loadError}`
        : "# Session Not Found\n\nThis session could not be located."
    } />
  );
}
```

extensions/claudecast/src/lib/session-parser.ts
```diff
@@ -46,19 +46,172 @@ const CLAUDE_DIR = path.join(os.homedir(), ".claude");
...
+  // 3. Naive decode (last resort — known-lossy, only cache if path exists)
+  const decoded = decodeProjectPath(encodedDirName);
+  try {
+    await fs.promises.access(decoded);
+    resolvedPathCache.set(encodedDirName, decoded);
+  } catch {
+    // Don't cache lossy results that don't exist on disk
+  }
+  return decoded;
+}
```
qazi0, line 115: **Nit: Tier-3 `fs.promises.access` catch could be narrowed**

Same pattern you already fixed elsewhere — this empty catch could suppress EACCES or EIO in addition to ENOENT:
```typescript
} catch (error: unknown) {
  const code = (error as NodeJS.ErrnoException)?.code;
  if (code !== "ENOENT") {
    console.warn(`Unexpected error checking decoded path "${decoded}":`, error);
  }
  // Don't cache lossy results that don't exist on disk
}
```

Low priority since this is the last-resort fallback, but would be consistent with the narrowing you did in the other catch blocks.

extensions/claudecast/src/deep-search-sessions.tsx
```diff
@@ -0,0 +1,457 @@
...
+      text: `${session.turnCount} turns`,
+      icon: Icon.Message,
+    });
+  }
+
+  if (session.cost > 0) {
+    accessories.push({
+      text: `$${session.cost.toFixed(4)}`,
+      icon: Icon.Coins,
+    });
```
qazi0, line 172: **Suggestion: `handleResume`/`handleFork` missing try-catch around `launchClaudeCode`**

If `launchClaudeCode` throws (terminal app not installed, AppleScript failure), the error propagates as an unhandled rejection. The existing `browse-sessions.tsx` has the same pattern, but since this is new code, consider wrapping:

```typescript
try {
  await launchClaudeCode({ projectPath: session.projectPath, sessionId: session.id });
  await popToRoot();
} catch (error) {
  await showToast({
    style: Toast.Style.Failure,
    title: "Failed to launch Claude Code",
    message: error instanceof Error ? error.message : String(error),
  });
}
```

Not blocking — just would improve UX for edge cases.

qazi0:
hey @RoderickJDunn, the above review is low-priority, these changes arent critical
Ive checked your deep search changes and the search seems to be working, although I when i try to open a session, it says `claude code is not installed, and shows the installation guide`

This doesn't have anything to do with your PR (which is good!) but that currently it doesn't auto-detect claude installed via the native installer, but only checks for the `npx` installed claude. If you could please also make a small change so that it also looks for claude in 
```
/Users/<Username>/.local/bin/claude
```
while searching for claude (or uses `which claude`) that would be amazing. I can then approve this PR.
Thank you for awesome work!

qazi0:
hey @RoderickJDunn, i went ahead and did the final fixes myself. attached is a patch, pls apply it on top of your latest remote commit (`b7be406`) and push. It fixes some other issues:

1. **Fix ENOENT crash** - `Action.ShowInFinder` in `launch-project.tsx` was crashing with `realpath ENOENT` for projects whose directory no longer exists on disk. Added an `existsSync` guard.

2. **Fix JS heap OOM** - `browse-sessions.tsx` called `listAllSessions()` with no limit, loading every session file into memory. Added `limit: 200`. Also rewrote `listProjectSessions` to only parse the target project directory instead of loading all sessions then filtering. Capped `getAllTimeStats` at 1000 sessions.

3. **Add `~/.local/bin/claude`** - Added to the common CLI paths array for users who install via pipx or manual local installs.

### To apply

```bash
# Download the patch
curl -L https://gist.githubusercontent.com/qazi0/0e5780beefd88e6c9d30eeb3e92b39e9/raw/claudecast-fixes.patch -o /tmp/claudecast-fixes.patch

# Apply it (from your extensions repo root)
git am /tmp/claudecast-fixes.patch
```

If `git am` has issues, you can also use:
```bash
git apply /tmp/claudecast-fixes.patch
```

Patch: https://gist.github.com/qazi0/0e5780beefd88e6c9d30eeb3e92b39e9

0xdhrv:
Hey @RoderickJDunn @qazi0 👋 

Moving this to draft, as it's in progress, mark it as ready for review, and I'll review 😊

RoderickJDunn:
@qazi0 thanks for the patch! Applied

PR #25373, opened 2026-02-12
[Skills] Add support for installing/removing skills

DaleSeo:
## Description

The Skills extension currently allows users to browse and search for skills on skills.sh, but it doesn't let them install or manage those skills within Raycast. Users have to copy the install command manually and run it in a terminal, and they can't see what's already installed on their machine.

This PR adds those missing features. Now, users can install skills directly from the search and trending views with just one action. There's also a new "Manage Skills" command that shows all installed skills along with agent information and SKILL.md content. Users can remove skills through a confirmation dialog to prevent accidental deletions.

Behind the scenes, all skill operations are handled by the `npx skills` CLI instead of directly manipulating the filesystem. This keeps the extension in sync with how the CLI manages its internal state, including agent directories and symlinks, and it automatically supports all 30+ agents that the CLI recognizes.

We're parsing the CLI output after removing ANSI codes since the CLI doesn't yet support `--json` flag yet. I've reached out to the Vecel team about this. Once they add support, we can simplify the parsing logic.

## Screencast

skills-5

skills-6

## Checklist


keito4:
LGTM
I think this is a great feature! Thanks for adding it.

0xdhrv:
@greptileai, could you please do a fresh review of this PR?

PR #25411, opened 2026-02-13
Add plex-movie-check extension

leoincedo:
## Description

  This PR adds a new Raycast extension: **Plex Movie Check**.

  ### What it does
  - Adds a new command: **Search Plex Movie**
  - Searches movie availability across Plex libraries
  - Supports:
    - title search
    - optional year filter
    - exact title toggle

  ### Technical details
  - Implemented Plex API integration for:
    - `/library/sections`
    - `/library/sections/{key}/all`
  - Added pagination fallback scan when direct title query does not return reliable results
  - Fixed Korean title matching issues by enabling XML entity decoding (`&#...;` -> actual text)
  - Added preferences:
    - `Plex Base URL`
    - `Plex Token`
    - `Default Exact Title`

  ## Screencast

  Screencast attached in PR comment (search flow: `Seven`, with/without year filter, and result selection).

  ## Checklist

  - [x] I ran `npm run build` and [tested this distribution build in Raycast](https://developers.raycast.com/basics/prepare-an-
  extension-for-store#metadata-and-configuration)


extensions/plex-movie-check/README.md
0xdhrv, line 1: **issue**: localization

Can you update the README to reflect this?

0xdhrv (commented):

We already have an extension in the Store that deals with Plex Media. Could we consider enhancing the existing extension below instead of creating another one?

- https://www.raycast.com/mSarheed/plex by @mSarheed

leoincedo:
The existing Plex plugin module had poor search performance, so it was improved to allow effective searching even with partial input strings.

0xdhrv:
> The existing Plex plugin module had poor search performance, so it was improved to allow effective searching even with partial input strings.

Makes sense. Improve the existing one instead of creating a new extension to avoid confusing users with multiple extensions serving the same purpose.

Issue #25549, opened 2026-02-17
[Bitwarden Vault] Failing to generate password post sign in


mabraithwaite:
### Extension

https://www.raycast.com/jomifepe/bitwarden

### Raycast Version

1.104.6

### OS Version

macOS 26.3

### Description

When searching my vault, everything seems to be working fine, but when trying to generate a password, I just get a "Failed to generate" message. Running the command myself via command line it's working fine. In the configuration I've tried passing the path to my own homebrew cli installation and without. In both cases I'm able to sign in and search the vault but unable to generate passwords.

### Steps To Reproduce

1. In macOS 26.3
2. Configured with Client ID and Client Secret, all other defaults set (also tried with setting my own cli and get the same results)
3. Log in with master password
4. Run 'Generate Password (Quick)'
5. See error "Failed to generate"
6. Run 'Generate Password'
7. Never finishes "Generating" and never shows an error message

jomifepe:
@raycastbot assign me

mabraithwaite:
Hey @jomifepe sorry if you already started working on this, put up a PR for it here https://github.com/raycast/extensions/pull/25553

jomifepe:
> Hey [@jomifepe](https://github.com/jomifepe) sorry if you already started working on this, put up a PR for it here [#25553](https://github.com/raycast/extensions/pull/25553)

Hey, the fix was already released, but thanks for looking into the issue.

I commented in your PR nonetheless.

jomifepe:
@raycastbot close this issue

PR #25553, opened 2026-02-18
Update bitwarden extension

mabraithwaite:
## Description

Open issue: https://github.com/raycast/extensions/issues/25549#event-22869625789

There's a bug when trying to generate passwords. It's looking like we're passing `false` for flags like `--passphrase` when they should instead by omitted from the cli call.

Updated `getPasswordGeneratingArgs()` to only allow args for the either `passphrase` type passwords or regular type passwords, and filtered out flags when the value is false. 

## Checklist

### DIFF
--- a/extensions/bitwarden/src/api/bitwarden.ts
+++ b/extensions/bitwarden/src/api/bitwarden.ts
@@ -216,7 +216,9 @@ export class Bitwarden {
 
         // clear the data.json file to avoid issues with the new binary
         const dataJsonPath = join(supportPath, "data.json");
-        await tryExec(() => unlink(dataJsonPath));
+        if (existsSync(dataJsonPath)) {
+          await tryExec(() => unlink(dataJsonPath));
+        }
       } catch (extractError) {
         toast.title = "Failed to extract Bitwarden CLI";
         throw extractError;
--- a/extensions/bitwarden/src/types/passwords.ts
+++ b/extensions/bitwarden/src/types/passwords.ts
@@ -1,6 +1,10 @@
+import { asOptionKeys } from "~/utils/types";
+
 export type PasswordType = "password" | "passphrase";
 
-export interface PasswordOptions {
+export type PasswordOptions = {
+  /** Do not generate a passphrase */
+  passphrase?: false;
   /** Include uppercase characters */
   lowercase?: boolean;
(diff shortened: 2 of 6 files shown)


extensions/bitwarden/src/api/bitwarden.ts
```diff
@@ -216,7 +216,9 @@ export class Bitwarden {
 
         // clear the data.json file to avoid issues with the new binary
         const dataJsonPath = join(supportPath, "data.json");
-        await tryExec(() => unlink(dataJsonPath));
+        if (existsSync(dataJsonPath)) {
+          await tryExec(() => unlink(dataJsonPath));
+        }
```
jomifepe, line 221: Did you have any issues with this line? Curious if it's a failing point

extensions/bitwarden/src/types/passwords.ts
```diff
@@ -26,9 +44,19 @@ export interface PassphraseOptions {
...
+};
+
+/** Keys for the passphrase options to be used at runtime */
+export const PASSPHRASE_OPTION_KEYS: readonly string[] = asOptionKeys<PassphraseOptions>()([
+  "passphrase",
+  "words",
+  "separator",
+  "capitalize",
+  "includeNumber",
+]);
```
jomifepe, line 56: This is an ok solution, but in the end, the issue was only that boolean args were being passed with a value, as you described. The generate command doesn't really care about password args being mixed with passphrase ones, or vice-versa.

extensions/bitwarden/src/api/bitwarden.ts
mabraithwaite: Yeah when first forking the project it kept failing on build because the json wasn't present

extensions/bitwarden/src/types/passwords.ts
mabraithwaite: yeah though when debugging it gets confusing when all params are passed but only a select few are taking effect.

mabraithwaite: Didn't realize you had closed this already so opened another PR after pulling most recent changes, added additional tests to if you want to take a look. https://github.com/raycast/extensions/pull/25598

PR #25555, opened 2026-02-18
Add temp-check extension

ADCAdams:
## Description

**Temp Check** gives you instant access to your Mac's CPU temperature — right from the menu bar or a detailed Raycast command.

### Why this extension?

There's no built-in way to check CPU temperature on macOS without third-party apps. Existing solutions are either paid, bloated, or require granting elevated permissions. Temp Check seamlessly reads thermal sensors directly via macOS IOKit APIs. This means no sudo, no background daemons, no kernel extensions, and no pop-ups!

### Features

- **Menu Bar Display** — Live CPU temperature with color-coded severity indicator (green/yellow/orange/red)
- **Detailed View** — Full breakdown of all thermal die sensors, chip model, CPU core count, and temperature status
- **Celsius / Fahrenheit Toggle** — Switch units instantly with `Cmd+U` in the list view, or from the menu bar dropdown. Preference persists across sessions via LocalStorage.
- **Configurable Thresholds** — Set custom warning (default 80°C) and critical (default 95°C) temperature thresholds in extension preferences
- **Apple Silicon Aware** — Detects chip model (M1/M2/M3/M4 etc.), reports correct core count, and explains that GPU shares the die on unified architecture chips

### How it works

The extension compiles a lightweight Objective-C binary from source (`tools/temp-reader.m`) during the build step — no opaque binaries are bundled. The binary uses IOKit HID thermal APIs to read all available temperature sensors. It's compiled as a universal binary (arm64 + x86_64) so it runs natively on both Apple Silicon and Intel Macs.

### A note on "Die Sensors" vs CPU cores

macOS exposes thermal die sensors (e.g. `PMU tdie1`), not per-core temperature readings. The number of die sensors does not match the number of CPU cores — for example, an M4 Max has 14 cores but 10 die sensors. The extension clearly labels these as "Die Sensor 1", "Die Sensor 2", etc. and shows the actual core count separately so users understand what they're seeing.

## Screencast

TC1
TC2
TC3
TC4
TC5

## Checklist


0xdhrv:

We already have an extension in the Store that deals with system monitoring. Could we consider enhancing the existing extension below instead of creating another one?

- https://www.raycast.com/hossammourad/raycast-system-monitor by @hossammourad

I think you can enhance the existing extension to add the temp-check feature

ADCAdams:
Thanks for the feedback @0xdhrv — that makes sense. Happy to contribute CPU temperature monitoring to the existing System Monitor extension instead.

The core value-add here is reading thermal sensor data via IOKit HID APIs through a lightweight Objective-C binary compiled from source (no sudo, no kernel extensions, no native node addons). This gives:

- **CPU die temperature** — average, max, and per-sensor breakdown
- **Chip-aware context** — detects Apple Silicon model, core count, and number of die sensors
- **Severity indicators** — color-coded normal/warm/hot/critical thresholds
- **Celsius/Fahrenheit toggle**

System Monitor currently only has battery temperature. I'll open a PR on the System Monitor extension adding a Temperature section to both the list view and menu bar command.

I'll close this PR once that's up.

PR #25577, opened 2026-02-18
System Monitor: Add CPU temperature monitoring

ADCAdams:
## Description

Adds real-time CPU temperature monitoring to the System Monitor extension, as suggested in #25555 by @0xdhrv.

Currently no temp monitoring in raycast store!

### What's new

- **Temperature section in list view** — shows CPU average, max, per-sensor die temperature readings, chip model, core count, and color-coded severity status (normal/warm/hot/critical)
- **Temperature in menu bar** — CPU temperature displayed alongside existing CPU, Memory, Network, and Power metrics
- **"Temperature" option in Default Tab preference** — users can launch directly to the temperature view

### How it works

Reads thermal sensor data via macOS IOKit HID APIs through a lightweight Objective-C binary (`tools/temp-reader.m`) compiled from source during the build step. No sudo, no kernel extensions, no native Node.js addons. The binary is compiled as a universal binary (arm64 + x86_64) via a `prebuild` npm script.

### Die sensors vs CPU cores

macOS exposes thermal die sensors (e.g. `PMU tdie1`), not per-core temperature readings. The number of die sensors does not match the CPU core count — for example, an M4 Max has 14 cores but 10 die sensors. The UI clearly labels these and shows the actual core count separately.

### GPU temperature

On Apple Silicon, GPU shares the die with the CPU (unified architecture). Separate GPU temperature is not exposed by the hardware. The extension explains this in the detail panel.

## Screencast

Screenshots from the standalone version (same UI, now integrated into System Monitor):

Temperature detail view
Die sensor readings

## Checklist

### DIFF
--- a/extensions/system-monitor/CHANGELOG.md
+++ b/extensions/system-monitor/CHANGELOG.md
@@ -1,6 +1,6 @@
 # System Monitor Changelog
 
-## [Added Temperature Monitoring] - {PR_MERGE_DATE}
+## [Added Temperature Monitoring] - 2026-02-19
 
 - Added temperature view under CPU section
 - CPU temperature displayed in menu bar dropdown


JadenMajid:
I think UI wise this might integrate better if the temps were shown under CPU/GPU with less verbosity(ie: showing only min avg max)

ADCAdams:
Thanks for the feedback! Happy to simplify the list view. I can fold temperature into the existing CPU detail panel as avg/max, and drop the per-sensor breakdown.

I'd like to keep a separate line item for CPU temperature in the menu bar dropdown though; temperature and usage are fundamentally different metrics (you can have low usage but high temps from throttling/ventilation). Having both visible at a glance is the main value-add here.

I'll rework the PR.

JadenMajid:
makes sense to me! I made a PR that got merged a while back that added customizability to the menu at entries, but not sure whether adding temp as a new tag(like "\<TEMP\>") or a new "thermals" section with CPU and GPU under it would make more sense, what do you think?

I'm tempted to say that a new tag would be more consistent between the menu bar and raycast views.

sorry for formatting, writing this on mobile

ADCAdams:
all good! I think a separate "Temperature" section in the dropdown makes the most sense. Temperature is a distinct metric from CPU usage, and nesting it under CPU with a <TEMP> tag would feel cramped in a format string that's really designed for a single percentage value.

The current implementation already has it as its own section with "CPU Temperature" as a line item. If GPU temp is available (non-Apple Silicon), it shows there too. Keeps it clean and consistent with how the other sections (Memory, Network, Power) each get their own block.

I've already pushed the rework; the list view now shows temp avg/max under the CPU detail panel, and the menu bar has the separate Temperature section. Let me know if that works or if you'd prefer any changes.

JadenMajid:
sounds good, I'll clone and take a proper look after lunch

JadenMajid:
Tested it out! The changes to both the menubar and raycast view look awesome. I think this would be a great addition to system-monitor. Thank you for your time and effort in contributing.


-  Currently, the temperature binary is compiled by the target computer during the prebuild step. I tried uninstalling xcode tools(which bundles clang) and got an error. I think(but am not sure) that dependency on X-code would break the extension for people without x code cli tools installed :(

To solve this I see two paths:
  - Precompiled binaries could be added to the assets folder(x86, arm) which could be prepackaged as part of the extension(with some check to see which to run). [Raycast allows binaries](https://developers.raycast.com/basics/prepare-an-extension-for-store#binary-dependencies-and-additional-configuration) as long as the source code is visible, so no issue there.
  - The [systeminformation](https://www.npmjs.com/package/systeminformation) npm package also includes temperature. I've been thinking about doing a full backend rewrite of the system-monitor with `systeminformation`, but for now just using it for temperature would be acceptable too IMO. I think a rewrite would be nice because it would offload the data gathering responsibility to a third party tool, and the extension could stick to being a pretty interface. It could also be unacceptably slow and a terrible idea; I've done no testing :P

```bash
ADCAdams/add-temperature-to-system-monitor ✔ $ npm run build

> prebuild
> clang -Wall -O2 -fobjc-arc -arch arm64 -arch x86_64 tools/temp-reader.m -framework IOKit -framework Foundation -o assets/temperature-reader

xcode-select: note: No developer tools were found, requesting install.
If developer tools are located at a non-default location on disk, use `sudo xcode-select --switch path/to/Xcode.app` to specify the Xcode that you wish to use for command line developer tools, and cancel the installation dialog.
See `man xcode-select` for more details.
```

The units system monitor uses are also currently prebaked, I'd like to add a tab to the preferences page that allows users to specify which units they want(F/C, GiB/GB, etc) but that's outside the scope of this PR and I'll make a new one for that in the near(ish) future.

Thanks once again! I'd love to see this added once the compilation issue is resolved.

ADCAdams:
Hm weird since I thought I had went that route. I'll look into it later tonight - cheers

JadenMajid:
I saw `assets/temperature-reader` in the `.gitignore`, so assumed the intention was for the end user to compile themselves at install time.
Cheers

ADCAdams:
Okay. Good catch on the clang dependency. I had the binary in .gitignore thinking it'd compile at install, but that obviously doesn't work without Xcode CLI tools. I've switched to shipping the precompiled universal binary (arm64 + x86_64) in assets/ directly. Source is still in tools/temp-reader.m and the prebuild script is there for anyone who wants to recompile.

Should be good to go now. Let me know :)

extensions/system-monitor/package.json
```diff
@@ -64,8 +64,8 @@
           "type": "textfield",
           "required": false,
           "title": "Network Formatting",
-          "placeholder": "↓ <DOWN>/s ↑ <UP>/s",
-          "default": "↓ <DOWN>/s ↑ <UP>/s"
+          "placeholder": "\u2193 <DOWN>/s \u2191 <UP>/s",
```
JadenMajid: @ADCAdams This is nitpicky but I'd prefer to keep the unencoded symbols in the `package.json` file. Makes the source code a bit more legible IMO

JadenMajid:
Other than my nitpicky comment about the symbol encoding, the change looks good overall to me!

extensions/system-monitor/CHANGELOG.md
JadenMajid, line 1: The changelog should also be updated to include the latest changes, I think these are a bit out of date.

JadenMajid (changes requested):

extensions/system-monitor/CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # System Monitor Changelog
 
+## [Added Temperature Monitoring] - {PR_MERGE_DATE}
+
+- Added CPU temperature monitoring via IOKit thermal sensors
+- Temperature section in list view with die sensor breakdown, chip info, and severity status
```
JadenMajid:
```suggestion
- Added temperature view under CPU section
```

extensions/system-monitor/CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
...
+- CPU temperature displayed in menu bar dropdown
+- Temperature option in Default Tab preference
```
JadenMajid:
```suggestion
```

JadenMajid:
awesome, LGTM🔥

JadenMajid (approved):

PR #25598, opened 2026-02-18
Update bitwarden extension

mabraithwaite:
## Description

Open issue: https://github.com/raycast/extensions/issues/25549#event-22869625789

There's a bug when trying to generate passwords. It's looking like we're passing false for flags like --passphrase when they should instead by omitted from the cli call.

Updated `getPasswordGeneratingArgs()` to only allow args for the either passphrase type passwords or regular type passwords, and filtered out flags when the value is `false`.

Added tests for arg filtering

## Checklist


mabraithwaite:
Saw separate change was made so pulled changes and integrated my changes into it (with tests) if you want to reconsider changes. Just saw your comments on this PR https://github.com/raycast/extensions/pull/25553 so no worries if changes are considered unnecessary, though I'd argue that while the param filtering technically doesn't matter to the cli as the `--passphrase` flag auto ignores the other params that aren't meant for passphrase, it makes it easier for debugging to see which flags are actually being passed and taking affect. Also if in the future the CLI changes and it starts rejecting these types of arguments, this plugin will continue working

mabraithwaite:
@jomifepe

PR #25628, opened 2026-02-19
Paste as Plain Text: Collapse surrounding whitespace when cleaning line breaks

asamofal:
## Summary

- Fix "Clean line breaks" option to also collapse whitespace surrounding line breaks, not just the line break characters themselves

## Problem

When copying text from terminal emulators, lines are often padded with trailing spaces and the next line starts with leading spaces/indentation. The current regex (`/[\r\n]/gm`) replaces only the newline character with a space, leaving all surrounding whitespace intact.

For example, copying terminal-wrapped text like:
```
...end-to-end type
   safety...
```

Results in: `...end-to-end type    safety...` (multiple spaces) instead of the expected `...end-to-end type safety...` (single space).

## Fix

Changed the regex from `/[\r\n]/gm` to `/[ \t]*[\r\n]+[ \t]*/gm` so it consumes spaces and tabs adjacent to line breaks, producing a single clean space.

This only affects whitespace directly surrounding line breaks — intentional multiple spaces elsewhere in the text are preserved.

### DIFF
--- a/extensions/paste-as-plain-text/CHANGELOG.md
+++ b/extensions/paste-as-plain-text/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Paste as Plain text Changelog
 
-## [Enhance Clean Line Breaks] - {PR_MERGE_DATE}
+## [Enhance Clean Line Breaks] - 2026-02-23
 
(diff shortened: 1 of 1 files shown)


pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #25644, opened 2026-02-20
Add ramadan-companion extension

saadhsnain:
## Description

Frame 3

The Ramadan Companion extension to help the Muslim community keep track of Sehri (Fajr) and Iftar (Maghrib) times during Ramadan, directly from Raycast.

The extension ships with 3 commands:

- **Ramadan Companion** – View both Sehri and Iftar times for today, with a live countdown to the next event.
- **Iftar Time** – Instantly see today's Iftar time and the time remaining until it.
- **Sehri Time** – Instantly see today's Sehri time and the countdown.

Prayer times are fetched from the AlAdhan Prayer Times API based on user-configured location and calculation preferences. Users can set their city, country, calculation method, juristic school, and preferred time format (12h/24h) in extension preferences.

The extension also supports inline keyword search! Typing iftar or sehri directly in Raycast surfaces the relevant command instantly.

## Screencast

Frame 5

Frame 4

A few notes on the choices made:
"Ramadan Companion" as the primary command doubles as a dashboard so users don't need to run two separate commands just to see both times.

The countdown is the most useful piece of info during Ramadan, so it's front and center in all three commands.
Keyword aliases (iftar, sehri) keep the UX frictionless — no need to remember the exact extension name.

## Checklist


0xdhrv (commented):

We already have an extension in the Store that deals with Ramadan timing. Could we consider enhancing the existing extension below instead of creating another one?

- https://www.raycast.com/mzaien/adhan-time by @mzaien

Reference:
- https://github.com/raycast/extensions/pull/25613#issuecomment-3927741935

PR #25658, opened 2026-02-20
[Skills] Inline detail Panel

DaleSeo:
## Description

At the Skills Launch Party, we received feedback that moving between the skill list and skill details was a bit inconvenient. Users had to open a separate detail view and then go back, which interrupted their browsing flow.

This PR changes that by replacing the push-to-detail navigation with inline detail panels. Now, skill content loads right next to the list. This way, users can browse skills and read their details without leaving the list view. There's also a toggle (Cmd+D) that lets users show or hide the detail panel based on what they prefer. This UI change also provides more horizontal space for the install command to display without being cut off.

To avoid unnecessary network requests, content is fetched only for the currently selected item instead of for every item in the list.

## Screencast

- Before: Two separative views

2026-02-20 at 12 59 27

2026-02-20 at 12 59 48

- After: Single view with inline panel

2026-02-20 at 12 58 54

## Checklist


keito4:
LGTM!
The inline detail panel is a big UX win, and Cmd+D toggle feels very natural!

DaleSeo:
Hi @0xdhrv, could you please review this PR?

PR #25669, opened 2026-02-21
Fix/trakt-manager user-agent

### LINKED ISSUE #25666: [Trakt Manager] Unable to use any commands and unable to login
### Extension

https://www.raycast.com/thesobercoder/trakt-manager

### Raycast Version

1.104.6

### OS Version

26.3

### Description

All commands do not show anything. 
Uninstalled the extension and tried to sign in again, but it gave the error below and does not login. 

### Steps To Reproduce

I used the extension like normal, so there was no change. I used it even before my macOS was updated to the latest version, and it had the same issue before too.

Error while signing in:
Error while fetching tokens: 403 (Forbidden)
<DOCTYPE html>
<!--[if It IE 7]> < html class= "no-js ie6 oldie"
lang="en-US" > <![endif] →..

### Current Behaviour

_No response_

### Expected Behaviour

_No response_

vishalbalaji3:
## Description

Added User-Agent header to API client to fix 403 errors caused by Trakt's Cloudflare WAF blocking requests without one.

Fixes #25666

### DIFF
--- a/extensions/trakt-manager/CHANGELOG.md
+++ b/extensions/trakt-manager/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Trakt Manager Changelog
 
-## [Fix] - {PR_MERGE_DATE}
+## [Fix] - 2026-02-23
 
 - Added User-Agent header to API client to fix 403 errors caused by Trakt's Cloudflare WAF blocking requests without one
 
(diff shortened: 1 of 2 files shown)


pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #25693, opened 2026-02-22
Claude Code Launcher: Load shell profiles in Ghostty and Alacritty adapters, update icon

jordiparra:
## Summary
- **Load shell profiles in Ghostty and Alacritty adapters**: When these terminals are launched via the macOS `open` command, shell profiles (`~/.zshrc`, `~/.zprofile`, etc.) are not automatically sourced. This causes environment variables like `ANTHROPIC_API_KEY` to be missing, resulting in "Missing API key" errors when launching Claude Code. The fix explicitly sources common shell profile files before running Claude.
- **Update extension icon**: Updated to match current Claude Code branding.

## Details

The Ghostty and Alacritty adapters both use `open -na <App> --args -e $SHELL -l -c <command>` to launch. While `-l` requests a login shell, the `-c` flag causes the shell to run in non-interactive mode, skipping `~/.zshrc` and similar interactive profile files where users commonly set environment variables.

Terminal.app and Warp are not affected — Terminal.app opens a full login+interactive shell via AppleScript, and Warp manages its own shell environment through launch configurations.

The fix sources all common profile files with error suppression (`2>/dev/null`) so missing files are silently ignored:
```bash
source ~/.zprofile 2>/dev/null; source ~/.zshrc 2>/dev/null; source ~/.bash_profile 2>/dev/null; source ~/.bashrc 2>/dev/null; source ~/.profile 2>/dev/null;
```

## Test plan
- [ ] Launch Claude Code via Ghostty with an API key set only in `~/.zshrc` — verify it loads without "Missing API key" error
- [ ] Launch Claude Code via Alacritty with the same setup — verify it works
- [ ] Launch Claude Code via Terminal.app — verify no regression
- [ ] Launch Claude Code via Warp — verify no regression


0xdhrv (commented):
Hi @jordiparra 👋

Thanks for the contribution 🔥

---

Issue #25708, opened 2026-02-22
[Apple Notes] get-note-content and update-note time out when Notes.app is not already running


jasondk:
### Extension

https://www.raycast.com/raycast/apple-notes

### Raycast Version

1.104.6

### OS Version

26.2

### Description

`getNoteBody()` and `setNoteBody()` in `api/applescript.ts` call `runAppleScript()` without specifying a timeout. The default in `@raycast/utils` is 10 seconds (10,000ms). When Notes.app is not already running, the AppleScript `tell application "Notes"` triggers an app launch, which can easily exceed 10 seconds.

The resulting error message is:

```
Error: Command timed out after undefined milliseconds: osascript
```

(The "undefined" is a cosmetic bug in `@raycast/utils` error reporting — `options?.timeout` is undefined because no options were passed, even though the actual timeout used was 10s.)

## Suggested Fix

Pass an explicit timeout to `runAppleScript` in `getNoteBody` and `setNoteBody`:

```typescript
export async function getNoteBody(id: string) {
  return runAppleScript(
    `
    tell application "Notes"
      set theNote to note id "${escapeDoubleQuotes(id)}"
      return body of theNote
    end tell
    `,
    { timeout: 30000 },
  );
}

export async function setNoteBody(id: string, body: string) {
  return runAppleScript(
    `
    tell application "Notes"
      set theNote to note id "${escapeDoubleQuotes(id)}"
      set body of theNote to "${escapeDoubleQuotes(body)}"
    end tell
    `,
    { timeout: 30000 },
  );
}
```

30 seconds provides enough headroom for Notes.app to launch, while still failing reasonably quickly if something is actually wrong.

Note: The "timed out after undefined milliseconds" phrasing in the error message is a separate minor bug in `@raycast/utils` — it reads `options?.timeout` for the error string, but when no options are passed, the timeout is applied internally as a default without being set on the options object.

### Steps To Reproduce

1. Quit Notes.app
2. Ask the AI to read a note's content
3. `get-note-content` calls `getNoteBody()` → `runAppleScript()` with default 10s timeout
4. Notes.app begins launching but doesn't finish within 10 seconds
5. Timeout error

### Current Behaviour

When Notes.app is not already running, `get-note-content` and `update-note` fail with:

```
Error: Command timed out after undefined milliseconds: osascript
```

[...]

PR #25720, opened 2026-02-23
[Apple Notes] Fix AI tool note ID mismatch, timeout, and search filtering

### LINKED ISSUE #25706: [Apple Notes] get-note-content and update-note AI tools fail with error -1728 due to note ID format mismatch
### Extension

https://www.raycast.com/raycast/apple-notes

### Raycast Version

1.104.6

### OS Version

26.2

### Description

The `get-note-content` and `update-note` AI tools frequently fail with AppleScript error -1728 because the AI passes the note's UUID (ZIDENTIFIER) instead of the x-coredata ID that AppleScript requires.

The `search-notes` tool returns notes with two different ID fields:

- `id` / `noteId`: the x-coredata format, e.g. `x-coredata://81243823-DC1E-4D85-81EC-3C075E135B4E/ICNote/p3910`
- `UUID`: the ZIDENTIFIER from the database, e.g. `5AEC22F4-7356-456E-9D0C-112466B97201`

The `get-note-content` and `update-note` tools pass the received `noteId` directly into AppleScript:

```applescript
tell application "Notes"
  set theNote to note id "${noteId}"
  return body of theNote
end tell
```

AppleScript only accepts the x-coredata format. When the AI passes the UUID format (which it does unpredictably, since both are present in the search results and the tool schema gives no guidance), the call fails with:

```
Notes got an error: Can't get note id "5AEC22F4-7356-456E-9D0C-112466B97201". (-1728)
```

## Suggested Fix

Two changes, ideally both applied:

**a) Make the tools accept both ID formats.** Before passing the `noteId` to AppleScript, check if it starts with `x-coredata://`. If not, resolve it via the SQLite database using the existing `executeSQL` utility:

```typescript
// In helpers.ts
import { executeSQL } from "@raycast/utils";

const NOTES_DB = resolve(homedir(), "Library/Group Containers/group.com.apple.notes/NoteStore.sqlite");

export async function resolveNoteId(noteId: string): Promise<string> {
  if (noteId.startsWith("x-coredata://")) {
    return noteId;
  }
  const result = await executeSQL<{ id: string }>(
    NOTES_DB,
    `SELECT 'x-coredata://' || zmd.z_uuid || '/ICNote/p' || note.z_pk AS id
     FROM ziccloudsyncingobject AS note
     LEFT JOIN z_metadata AS zmd ON 1=1
     WHERE note.zidentifier = '${noteId.replace(/'/g, "''")}'
     LIMIT 1`,
  );
  return result?.[0]?.id ?? noteId;
}
```

**b) Improve the tool input schema.** Update the `noteId` description in both tool `Input` types and in the extension's `package.json` instructions to steer the AI toward the correct format:

```typescript
type Input = {
  /** The x-coredata ID of the note (starts with 'x-coredata://'). Use the 'id' field from search results, NOT the 'UUID' field. */
  noteId: string;
};
```

### Steps To Reproduce

1. Use Raycast AI with the Apple Notes extension
2. Ask it to retrieve the content of a specific note (e.g. "What's in my note called [note title]?")
3. The AI calls `search-notes`, gets results containing both `id` and `UUID` fields
4. The AI calls `get-note-content` with the UUID instead of the x-coredata ID
5. AppleScript error -1728

This doesn't happen every time — it depends on which ID field the AI selects. But it's frequent enough to make the tool unreliable.__

### Current Behaviour

When the AI passes the UUID format (e.g. `5AEC22F4-7356-456E-9D0C-112466B97201`) instead of the x-coredata format, `get-note-content` and `update-note` fail with:

```
Notes got an error: Can't get note id "5AEC22F4-7356-456E-9D0C-112466B97201". (-1728)
```

The error occurs because AppleScript's Notes dictionary only accepts the x-coredata URI format for note IDs. Since `search-notes` returns both formats and the tool schema provides no guidance on which to use, the AI selects the wrong one unpredictably.

### Expected Behaviour

`get-note-content` and `update-note` should reliably retrieve/update the correct note regardless of which ID format the AI passes. Either the tools should accept both ID formats (resolving UUIDs to x-coredata IDs before calling AppleScript), or the schema should clearly guide the AI to always use the `id` field (x-coredata format) from `search-notes` results.

Aayush9029:
## Description

Fixes Apple Notes AI note ID resolution, AppleScript timeout handling, and SQL search filtering for reliable macOS behavior; closes #25706, #25707, and #25708.

## Screencast

N/A

## Checklist

### DIFF
--- a/extensions/apple-notes/CHANGELOG.md
+++ b/extensions/apple-notes/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Apple Notes Changelog
 
-## [Bug Fix Update] - {PR_MERGE_DATE}
+## [Bug Fix Update] - 2026-02-26
 
 - Fix `get-note-content` and `update-note` timing out when Notes is not already running by using a longer AppleScript timeout.
 - Fix AI tool failures caused by UUID note IDs by resolving them to x-coredata IDs before calling AppleScript.


0xdhrv (approved):
Looks good to me, approved 🔥

PR #25740, opened 2026-02-23
feat(clipboard-type): add soft newlines preference to use Shift+Enter

wouterds:
## Summary

- Adds a new **Soft Newlines** preference (checkbox, disabled by default) to the Clipboard Type extension
- When enabled, newlines are typed as **Shift+Enter** instead of Enter
- Existing behavior is completely unchanged unless the user opts in

## Why?

Many websites and apps (Slack, Discord, ChatGPT, etc.) treat Enter as "submit/send" rather than "insert a newline". When pasting multiline clipboard content via keystroke simulation, each newline triggers a form submission, splitting your text into multiple messages.

With this option enabled, newlines are sent as Shift+Enter, which inserts an actual line break in most apps without submitting.

## Changes

- `package.json`: Added `softNewlines` checkbox preference (default: `false`)
- `src/type-clipboard.ts`: When enabled, uses `key code 36 using shift down` (Shift+Enter) instead of `key code 36` (Enter) for newline characters

### DIFF
--- a/extensions/clipboard-type/CHANGELOG.md
+++ b/extensions/clipboard-type/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Clipboard Type Changelog
 
-## [Added soft newlines preference] {PR_MERGE_DATE}
+## [Added soft newlines preference] 2026-03-23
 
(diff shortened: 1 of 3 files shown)


0xdhrv (approved):
Looks good to me, approved ✅

PR #25745, opened 2026-02-23
Add holy-quran extension

abdullahnettoor:
## Checklist


0xdhrv (commented):
Thanks for your contribution @abdullahnettoor  🔥

We already have an extension in the Store that deals with Quran. Could we consider enhancing the existing extension below instead of creating another one?

- https://www.raycast.com/BalliAsghar/the-noble-quran by @BalliAsghar

abdullahnettoor:
Hi @0xdhrv ,

I started by looking at the https://www.raycast.com/BalliAsghar/the-noble-quran by @BalliAsghar, but then I noticed that the purpose of the existing one and the one I’ve made are quite different. I tried to use the same approach in the existing one, which would require a lot of changes and refactoring. If you take a look at the list of commands that [this](https://www.raycast.com/BalliAsghar/the-noble-quran) one offers compared to mine, you’ll see that it only includes two commands: just translating the Quran and favouriting any verse. 

My extension has nine commands and a menubar that can handle audio. I mainly focused on creating an A Quran audio player that can do a lot of things, and I’m also planning to add more features in the future. I hope you’ll consider this extension, and we’ll be more helpful.

Thank you

PR #25791, opened 2026-02-24
Update bitwarden extension

mabraithwaite:
## Description
- Replace the separate Card and Identity detail screens with a single detail view that works for all vault item types (Login, Card, Identity, Note, SSH Key)
- Add interactive field components with copy, paste, reveal/hide, and open-in-browser actions per row
- Add keyboard shortcuts for copy/paste password (Opt+P / Cmd+Opt+P) and detail panel toggle (Cmd+D)

### Motivation

[Related Feature Request](https://github.com/raycast/extensions/issues/25790)

Only Card and Identity items had a detail screen, and it was a static markdown view with copy-only actions. and SSH Key items had no detail screens. Sensitive fields like card numbers were shown in plain text with no masking.

This PR introduces a unified, interactive detail view where every item type gets a sectioned field list. Sensitive fields are masked by default, URIs can be opened in the browser, and all fields support both copy and paste.

#### New — Detail screen

- `ItemDetails` — fetches fresh item data from the CLI on open, with loading/error/empty states and retry
- `ItemFieldList` — renders fields grouped into titled sections via `buildFieldSections`
- `ShowItemDetailsAction` — single action replacing `ShowCardDetailsAction` and `ShowIdentityDetailsAction`

#### New — Field components

- `BaseItemField` — shared `List.Item` layout with composable action panel and detail side-panel toggle
- `TextItemField` — plain-text field (copy + paste)
- `HiddenItemField` — masked field with reveal/hide toggle (Cmd+H), transient clipboard copy
- `LinkItemField` — URI field with copy + open-in-browser (Cmd+Return)
- `TotpItemField` — TOTP code copy/paste as primary actions, secret management as secondary

#### New — Reusable actions

- `CopyFieldItemAction` — generic copy with transient clipboard support
- `PasteFieldItemAction` — generic paste with frontmost app name in title

#### New — Types & utilities

- `types/item-field.ts` — discriminated union (`TextField | LinkField | HiddenField | TotpField`) and `FieldSection`
- `buildFieldSections.ts` — pure function mapping an `Item` to ordered `FieldSection[]` for all five item types
- `buildFieldSections.test.ts` — comprehensive test suite (701 lines covering all item types, partial data, edge cases, section ordering)
- `constants/icons.ts` — URI scheme to icon mapping (80+ schemes)
- `utils/uri.ts` — `uriSchemeIcon()` helper
- `constants/passwords.ts` — `SECRETS_MASK` constant
- `utils/strings.ts` — `asPlainTextDetail()` for safe markdown code-fence wrapping

## Screencast

screenshot1
screenshot2_identity1
screenshot3_identity2
screenshot4_identity3
screenshot5_login
screenshot6_ssh
screenshot7_notes

## Checklist


pernielsentikaer:
Can you have a look here @jomifepe

jomifepe:
> Can you have a look here @jomifepe

Is the PR ready? It's still in draft @mabraithwaite.

I can say already that I like the direction, but it's quite a big change, and it might take a bit longer to review.

mabraithwaite:
Ah, sorry, yes it's ready for review

extensions/bitwarden/src/components/searchVault/ItemActionPanel.tsx
```diff
@@ -32,6 +31,9 @@ const VaultItemActionPanel = () => {
 
   return (
     <ActionPanel>
+      <ActionPanel.Section>
+        <ShowItemDetailsAction />
+      </ActionPanel.Section>
```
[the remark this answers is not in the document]
mabraithwaite: The goal of the change is to create a unified experience across all item types so this is intentional

extensions/bitwarden/src/components/searchVault/utils/buildFieldSections.ts
```diff
@@ -0,0 +1,342 @@
...
+  if (login.uris && login.uris.length > 0) {
+    const uriFields: ItemField[] = login.uris
+      .filter((u) => u.uri)
+      .map((u, i) => ({
+        type: "link",
+        id: `login.uri.${i}`,
+        label: login.uris!.length === 1 ? "URI" : `URI ${i + 1}`,
+        value: u.uri!,
+      }));
+    if (uriFields.length > 0) sections.push({ title: "URIs", fields: uriFields });
```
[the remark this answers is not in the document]
mabraithwaite: Added change

extensions/bitwarden/src/components/searchVault/ItemFieldList.tsx
```diff
@@ -0,0 +1,86 @@
...
+                return <TotpItemField key={field.id} item={field} onToggleDetailPanel={toggleDetailPanel} />;
+              default:
+                <BaseItemField
+                  key={(field as ItemField).id}
+                  id={(field as ItemField).id}
+                  label={(field as ItemField).label}
+                  displayValue={(field as ItemField).displayValue ?? (field as ItemField).value}
+                  detailValue={(field as ItemField).value}
+                  onToggleDetailPanel={toggleDetailPanel}
+                />;
```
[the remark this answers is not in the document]
mabraithwaite: Good catch, added change

mabraithwaite:
~@jomifepe not sure what happened with the PR/why so many commits were added. Locally I don't see these changes so not sure what happened. Should I delete and start over?~

0xdhrv (commented):
Hey @jomifepe 👋

Do you have a moment to check out this PR when it's possible? Thanks!

jomifepe:
> Hey @jomifepe 👋
> 
> Do you have a moment to check out this PR when it's possible? Thanks!

Will check it later today.

extensions/bitwarden/src/components/searchVault/ItemFieldList.tsx
```diff
@@ -0,0 +1,88 @@
...
+                return (
+                  <BaseItemField
+                    key={(field as ItemField).id}
+                    id={(field as ItemField).id}
+                    label={(field as ItemField).label}
+                    displayValue={(field as ItemField).displayValue ?? (field as ItemField).value}
+                    detailValue={(field as ItemField).value}
+                    onToggleDetailPanel={toggleDetailPanel}
+                  />
+                );
```
jomifepe: This `default` doesn't seem to make sense here, especially since the sections are controlled and build by an util. There doesn't seem to be a case where this code will be reached, and TypeScript also tells us that.



```suggestion
```

extensions/bitwarden/src/components/searchVault/LinkItemField.tsx
```diff
@@ -0,0 +1,51 @@
...
+ *
+ * The row icon is derived from the URI scheme (e.g. `https` → globe icon) via
+ * {@link uriSchemeIcon}, and a trailing link accessory indicates the value is
+ * openable.
+ *
+ * @param props - See {@link LinkItemFieldProps}.
+ * @param props.item - The link field descriptor containing the URI value.
+ * @param props.onToggleDetailPanel - Callback forwarded to {@link BaseItemField}
+ *        to toggle the side detail panel.
+ * @returns A {@link BaseItemField} element configured for URI display.
```
jomifepe: I appreciate the documentation, but I believe JSDocs is quite verbose and requires maintenance (sure LLMs would take care of that better than us humans), so I would prefer one of the following options:
1. Don't document the props or returns too much, and leave good naming and TypeScript to explain them enough.
2. Move these lines to the type properties themselves, which are more useful and actually appear on the tooltip of the prop (in most editors).

Let's keep comments short and concise, and only explain things if the code doesn't explain itself well enough (e.g., bitwarden-specific/complex logic, specific choices, etc).

This applies to the whole PR.

extensions/bitwarden/src/components/searchVault/TotpItemField.tsx
```diff
@@ -0,0 +1,77 @@
...
+          />
+        </>
+      }
+      additionalActionsSections={
+        <>
+          <ActionPanel.Section title="TOTP Secret">
+            <ActionWithReprompt
+              title={isShowing ? `Hide Secret` : `Reveal Secret`}
+              icon={isShowing ? Icon.EyeDisabled : Icon.Eye}
+              onAction={() => setIsShowing(!isShowing)}
```
jomifepe: Nitpick:

```suggestion
              onAction={() => setIsShowing((current) => !current)}
```

extensions/bitwarden/src/components/searchVault/ItemFieldList.tsx
```diff
@@ -0,0 +1,88 @@
+import { List } from "@raycast/api";
+import { useState } from "react";
+import { Item, ItemType } from "~/types/vault";
+import { BaseItemField } from "./BaseItemField";
+import HiddenItemField from "./HiddenItemField";
+import LinkItemField from "./LinkItemField";
+import TextItemField from "./TextItemField";
+import TotpItemField from "./TotpItemField";
```
jomifepe: Nitpick: I know you're essentially following the examples of the codebase, that's mostly my fault, but I would move all these components to this file. There's no need to extract them if they are short enough and not used anywhere else.

extensions/bitwarden/src/components/searchVault/HiddenItemField.tsx
```diff
@@ -0,0 +1,67 @@
...
+      onToggleDetailPanel={onToggleDetailPanel}
+      mainActions={
+        <>
+          <CopyFieldItemAction label={item.label} content={copyContent} type="password" />
+          <PasteFieldItemAction label={item.label} content={copyContent} />
+          <ActionWithReprompt
+            title={isShowing ? "Hide Value" : "Reveal Value"}
+            icon={isShowing ? showingIcon : hiddenIcon}
+            shortcut={{ macOS: { modifiers: ["cmd"], key: "h" }, Windows: { modifiers: ["ctrl"], key: "h" } }}
+            onAction={() => setIsShowing(!isShowing)}
```
jomifepe: Nitpick: 

```suggestion
            onAction={() => setIsShowing((current) => !current)}
```

extensions/bitwarden/src/components/searchVault/ItemFieldList.tsx
```diff
@@ -0,0 +1,88 @@
...
+ * is the notes body) and closed for all other types.
+ *
+ * @param props - See {@link ItemFieldListProps}.
+ * @param props.item - The fully-loaded vault item whose fields should be rendered.
+ * @param props.folderName - Optional folder name prepended to the navigation title
+ *        (e.g. `"/Work/GitHub"`) for breadcrumb context.
+ * @returns A Raycast `List` with sectioned field rows, or an empty-state view when
+ *          the item has no displayable fields.
+ */
+function ItemFieldList({ item, folderName }: ItemFieldListProps) {
```
jomifepe, line 63: One important thing I believe we should preserve is the direct path to get the password. Currently, you can hit Enter or ⌥+Enter to copy or paste the password. With this change, you either need to ⌥+P or ⌥+⌘+P or open the details and move an item down.

Ideally, for login items, we would pre-select the password (and not the username) field when the details are open, so that users only need to hit enter twice, rather than having to memorize a new shortcut or have to go down to copy it the password.

extensions/bitwarden/src/components/searchVault/BaseItemField.tsx
jomifepe, line 1: ⚠️ We currently require the password prompt twice for items with the "Master password re-prompt" option enabled, the first to open the details and the second to copy a value, reveal a secret, or even to hide it again. Happens for all types of items.

jomifepe, line 1: ⚠️ It seems that if we open an item's details when the vault's items are almost finished loading (real data, not cached), and we click to open the details, it will fetch the item from the vault and not make use of the one that was just loaded, making the user wait longer than before.

extensions/bitwarden/src/components/searchVault/utils/buildFieldSections.ts
```diff
@@ -0,0 +1,341 @@
...
+
+  if (mainLoginFields.length > 0) sections.push({ title: "Login", fields: mainLoginFields });
+
+  if (login.uris && login.uris.length > 0) {
+    const nonEmptyUris = login.uris.filter((u) => u.uri);
+    const uriFields: ItemField[] = nonEmptyUris.map((u, i) => ({
+      type: "link",
+      id: `login.uri.${i}`,
+      label: nonEmptyUris.length === 1 ? "URI" : `URI ${i + 1}`,
+      value: u.uri!,
```
jomifepe, line 45: I'm not a big fan of these non-null assertions. I would prefer that we think about each case, return something else, or even throw an error if it's an impossible state to encounter.

extensions/bitwarden/src/components/searchVault/utils/buildFieldSections.ts
```diff
@@ -0,0 +1,341 @@
...
+    case ItemType.CARD:
+      sections.push(...buildCardSections(item.card!));
+      break;
+
+    case ItemType.IDENTITY:
+      sections.push(...buildIdentitySections(item.identity!));
+      break;
+
+    case ItemType.SSH_KEY:
+      sections.push(...buildSshKeySections(item.sshKey!));
```
jomifepe: Same case as https://github.com/raycast/extensions/pull/25791/changes#r2908749580

extensions/bitwarden/src/components/searchVault/actions/CopyTotpAction.tsx
```diff
@@ -7,7 +7,12 @@ import useGetUpdatedVaultItem from "~/components/searchVault/utils/useGetUpdated
 import { captureException } from "~/utils/development";
 import { showCopySuccessMessage } from "~/utils/clipboard";
 
-function CopyTotpAction() {
+type CopyTotpActionProps = {
+  omitShortcut?: boolean;
+  shortcut?: Keyboard.Shortcut;
```
jomifepe: This two-prop combination is unintuitive. If `omitShortcut=true`, the shortcut prop is ignored silently.

Better: `shortcut?: Keyboard.Shortcut | null` — callers pass `null` to suppress, a value to override, and omit the prop to get the default.

Same for `PasteTotpAction`.

extensions/bitwarden/src/components/searchVault/utils/useItemAccessories.ts
```diff
@@ -7,6 +7,14 @@ import { Folder, Item, ItemType, Reprompt } from "~/types/vault";
...
 
 type ListItemAccessory = NonNullable<List.Item.Props["accessories"]>[number];
+export const ITEM_TYPE_TO_COLOR_MAP: Record<ItemType, Color> = {
+  [ItemType.LOGIN]: Color.Blue,
+  [ItemType.CARD]: Color.Green,
+  [ItemType.IDENTITY]: Color.Orange,
+  [ItemType.NOTE]: Color.PrimaryText,
+  [ItemType.SSH_KEY]: Color.SecondaryText,
+};
+
```
jomifepe: This doesn't seem to be used anywhere.

```suggestion
```

extensions/bitwarden/src/components/searchVault/BaseItemField.tsx
```diff
@@ -0,0 +1,76 @@
...
+}: BaseItemFieldProps) {
+  return (
+    <List.Item
+      id={id}
+      title={label}
+      subtitle={displayValue}
+      icon={icon ?? Icon.Clipboard}
+      accessories={accessories}
+      detail={
+        typeof detailValue === "string" ? <List.Item.Detail markdown={asPlainTextDetail(detailValue)} /> : detailValue
```
jomifepe: According to TypeScript, this condition doesn't make sense, as detailValue can only be a `string`.

jomifepe (changes requested):
Good job on the implementation; it's working fairly well, but not quite ready. 

Found a couple of issues and left some comments, let me know if you need anything 🙂

extensions/bitwarden/src/constants/icons.ts
```diff
@@ -0,0 +1,89 @@
+import { Icon } from "@raycast/api";
+
+export const URI_SCHEME_ICONS: Record<string, Icon> = {
```
jomifepe: This can live alongside the `uriSchemeIcon` function, let's reduce the number of unnecessary files.

extensions/bitwarden/src/components/searchVault/ItemFieldList.tsx
mabraithwaite: This is just usually something I'll add as a bit of defensive coding in case a new field type gets introduced in the future it will still render without having to update every layer of it. Good to remove if you'd like though.

extensions/bitwarden/src/components/searchVault/LinkItemField.tsx
mabraithwaite: Yeah this tends to be where I use AI the most is documentation. It's cheaper now to keep the documentation up to date so I tend to have it get more descriptive, will dial it back.

extensions/bitwarden/src/components/searchVault/TotpItemField.tsx
mabraithwaite: will update

extensions/bitwarden/src/components/searchVault/ItemFieldList.tsx
mabraithwaite: Will move

extensions/bitwarden/src/components/searchVault/utils/buildFieldSections.ts
mabraithwaite: In this particular case it's only required because of how the `Uris` interface in `src/types/vault.ts` is set up. In the line above it `const nonEmptyUris = login.uris.filter((u) => u.uri);` this should already have filtered out all falsy values so `u.uri!` is more just based on the interface definition.

extensions/bitwarden/src/components/searchVault/utils/buildFieldSections.ts
mabraithwaite: I think the most appropriate fix here would be changing the Item type/interface definition to be a union of sub-types (login, note, etc). Then there would be an expectation that that login value be present instead possibility of it being missing because `login` won't be on all item types

extensions/bitwarden/src/components/searchVault/actions/CopyTotpAction.tsx
mabraithwaite: Agreed, more intuitive this way, will change

extensions/bitwarden/src/components/searchVault/utils/useItemAccessories.ts
mabraithwaite: yeah sorry, remains from trying something else, will remove

extensions/bitwarden/src/components/searchVault/BaseItemField.tsx
mabraithwaite: Updated the type to allow react nodes for more custom layouts

extensions/bitwarden/src/constants/icons.ts
mabraithwaite: Will move

extensions/bitwarden/src/components/searchVault/BaseItemField.tsx
mabraithwaite: Switched all the copy/paste/display actions to just `@raycast/api/Action` since the detail display action is already wrapped by reprompt

mabraithwaite: Might be worth adding this as a future update/backlog item? Currently the item is being fetched the same way `CopyUsernameAction`, `CopyPasswordAction`, and other item actions are being fetched, so while it potentially may be an issue, it's not a new issue/different from what is already present

extensions/bitwarden/src/components/searchVault/ItemFieldList.tsx
mabraithwaite: I'd like to push back on this a bit.
1. I think it can be a bit jarring to have different ux for different items. Since the every item will now be sharing this standard detail card, I think they should all share a standard path of getting there.
2. Using ⌥+P will bring more parity with the current mac bitwarden desktop app that uses ⌘+P to copy the password (just like ⌥+U => ⌘+U and ⌥+T => ⌘+T)

extensions/bitwarden/src/components/searchVault/utils/buildFieldSections.ts
jomifepe: Yes, I do agree that we should have a discriminated union, but maybe not in this PR, as it touches more parts of the codebase. Here, I believe it would be better to be defensive and do something like `if (!item.login) break` instead of the null assertion.

extensions/bitwarden/src/components/searchVault/utils/buildFieldSections.ts
jomifepe: I would say to add a type predicate to the filter, but honestly, here it's irrelevant, so it looks good 👍

extensions/bitwarden/src/components/searchVault/BaseItemField.tsx
jomifepe: What I was experiencing was definitely a UX regression, but it's not happening anymore in the latest version, so I guess no change is needed.

extensions/bitwarden/src/components/searchVault/ItemFieldList.tsx
jomifepe: Still, I find the ergonomics of ⌥+⌘+P quite awkward, and it's also not working for me 😕

0xdhrv:
Hey @mabraithwaite 👋 

Thanks for your contribution!

Thanks for the review as well @jomifepe , if you can approve once you feel it's ready for the merge. 

extensions/bitwarden/src/components/searchVault/ItemFieldList.tsx
mabraithwaite: Ah, I see what you mean by the pre-select. I moved the password field to be first in the login list so it's first selected. Does that work or are you wanting username first but password selected?

As for ⌥+⌘+P not working for you, might be transient? I've tried it with multiple passwords, including reprompts and it's working fine for me

extensions/bitwarden/src/components/searchVault/utils/buildFieldSections.ts
mabraithwaite: Added falsy checks

mabraithwaite:
Hey @jomifepe, added requested changes, ready for additional review

extensions/bitwarden/src/components/searchVault/utils/buildFieldSections.ts
```diff
@@ -0,0 +1,270 @@
...
+  }
+  if (login.username) {
+    mainLoginFields.push({
+      type: "text",
+      id: "login.username",
+      label: "Username",
+      value: login.username,
+      icon: Icon.PersonCircle,
+    });
+  }
```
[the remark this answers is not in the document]
mabraithwaite: Good catch, pushed fix, all tests pass

extensions/bitwarden/src/components/searchVault/ItemFieldList.tsx
jomifepe: Maybe I have something conflicting, but good to know that it works.

Looks good, the ordering of the fields was nice (username first), but I think it's a good tradeoff to reduce the amount of keystrokes/clicks to get the password.

jomifepe (approved):
Alright! Quite a large PR, but it looks good!

Nice work on refreshing the UX of the extension 🙂

0xdhrv (approved):
Looks good to me, approved ✅

PR #25805, opened 2026-02-25
Add worktree-manager extension

rodrigoea:
## Description

This is a [Raycast](https://raycast.com) extension to manage [Git worktrees](https://git-scm.com/docs/git-worktree): search worktrees across one or more root folders and create new ones from the same interface.

With this extension you can:

- **Search Worktrees** – List all worktrees from a configured root path. Search by path, branch, or repo name. Open in editor (Cursor, VS Code, etc.), Show in Finder, or Copy path.
- **Create Worktree** – Pick a repo and branch; the new worktree is created in your default worktree folder (e.g. `reponame-branch`). Supports branches that exist only on the remote (creates local branch in the new worktree).

Setup is done via Extension Preferences:
- Root path (where to scan for repos)
- Default worktree path (where new worktrees are created)
- Open with (app to open worktrees)

## Screenshots

Screenshots are in the `metadata` folder.

image

image

## Checklist


0xdhrv (commented):

We already have an extension in the Store that deals with Git Worktrees. Could we consider enhancing the existing extension below instead of creating another one?

- https://www.raycast.com/canac/worktrees by @canac

rodrigoea:
@0xdhrv

I appreciate you taking a look :) 

The main reason for this PR was because I found the existing plugins less "plug-and-play", especially [this one](https://www.raycast.com/philstainer/git-worktrees) that doesn't work well if you already have Worktrees created, it requires the workflow of creating worktrees to pass through the plugin, otherwise it won't identify any existing worktrees.

The plugin you commented on was also not listing worktrees -- but in this case, i'd say it was more an error handler missing. I had a broken worktree and when I tried to use the plugin to list worktrees, internally the command would break in one of these worktrees I had (sorry, I don't remember the exact error right now, but it was probably due to the worktree not being available on the remote anymore).

With that being said, i'll just close this PR. I might try to open a PR to fix that thing on https://www.raycast.com/canac/worktrees by @canac

0xdhrv:
Thanks @rodrigoea ❤️ 

Let me know if you need anything else. Happy to help!

PR #25818, opened 2026-02-25
[Window Walker] Fix window icons not shown on first load

### LINKED ISSUE #25798: [Window Walker] Window icons not shown
### Extension

https://www.raycast.com/nazzy_wazzy_lu/window-walker

### Raycast Version

0.46.2.0

### OS Version

Windows 11 Professional 10.0.26100.0 amd64

### Description

"Navigation", the built-in window switcher, shows window icons (but doesn't show some of the windows). Window Walker shows all the windows but doesn't show window icons.
Both PowerToys Command Palette and Flow Launcher show all the windows and their icons.

### Steps To Reproduce

Open Window Walker.

### Current Behaviour

_No response_

### Expected Behaviour

_No response_

luinbytes:
Fixes #25798

## Problem
Window Walker doesn't show window icons on first load. Icons only appear after a second manual refresh.

## Root Cause
Icons are extracted asynchronously in background via `setTimeout(extractMissingIcons, 100)`, but the UI never refreshes to show the newly extracted icons.

## Solution
Remove the `setTimeout()` wrapper and call `extractMissingIcons()` synchronously. Icons now extract immediately before the window list returns.

## Trade-offs
- ✅ Icons appear on first load (better UX)
- ⚠️ Brief pause on first load (~0.5-1s) while extracting
- ✅ Subsequent loads are instant (icons cached)

## Testing
- `npm run lint` - ESLint & Prettier passed
- Icons extract synchronously on first window list load
- Cached icons return instantly
- No functional changes to icon extraction logic

## Changes
- Modified `extensions/window-walker/src/lib/windows.ts`
- Removed `setTimeout()` wrapper around `extractMissingIcons()`
- Icons now extract synchronously before window list returns

## Note
This is a minimal fix that solves the immediate problem. An alternative approach would be to add a callback to refresh the UI when async extraction completes, but that requires more extensive changes to the React component.

Co-Authored-By: Lumi <luinbytes>

### DIFF
--- a/extensions/window-walker/src/lib/windows.ts
+++ b/extensions/window-walker/src/lib/windows.ts
@@ -363,11 +363,9 @@ export function refreshWindowListCache(): WindowInfo[] {
   cachedWindowList = windows;
   cacheTimestamp = Date.now();
 
-  // Schedule background icon extraction for windows without cached icons
+  // Extract icons synchronously for windows without cached icons
   if (windowsNeedingIcons.length > 0) {
-    setTimeout(() => {
-      extractMissingIcons(windowsNeedingIcons);
-    }, 100);
+    extractMissingIcons(windowsNeedingIcons);
   }
 
   return windows;


0xdhrv (approved):
Looks good to me, approved 🔥

PR #25845, opened 2026-02-26
[Cron Manager] Fix tasks disappearing from UI & permission handling

TahaZahit:
## Summary

Fixes #25811 — Tasks disappearing from UI despite being present in crontab, and recurring permission prompt loop.

### 1. Non-Raycast crontab content destroyed on every save (Critical)
`writeCrontab()` was overwriting the **entire** system crontab with only Raycast-managed entries. All non-Raycast content (environment variables, other cron jobs, comments) was silently wiped on every save operation.

**Fix:** `readCrontab()` now returns a `CrontabReadResult` containing both parsed jobs **and** `nonManagedContent`. `writeCrontab()` accepts and preserves this content verbatim, prepending it before the Raycast-managed block.

### 2. TCC / Permission errors crashed silently (High)
Only `"no crontab"` errors were caught. macOS TCC permission-denied errors threw generic exceptions, causing the job list to render as empty with no guidance.

**Fix:** Added explicit detection for `"Operation not permitted"` / `"permission denied"` with an actionable error message pointing to System Settings → Privacy & Security → Full Disk Access.

### 3. Handler stabilization (Medium)
Wrapped `handleUpdateJob`, `handleDeleteJob`, and `handleToggleStatus` in `useCallback` with stable dependency on `saveToSystem`. Stored non-managed crontab content in a `useRef` to preserve it across component re-renders and round-trip writes.

## Files Changed
- `extensions/cron-manager/src/utils/crontabSync.ts` — Rewritten parser/serializer with non-managed content preservation
- `extensions/cron-manager/src/cron.tsx` — Updated to use new `CrontabReadResult` API

## Testing
- `npm run build` passes with zero TypeScript errors
- Verified crontab round-trip preserves non-Raycast entries

### DIFF
--- a/extensions/cron-manager/CHANGELOG.md
+++ b/extensions/cron-manager/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## [1.0.2] - {PR_MERGE_DATE}
+## [1.0.2] - 2026-02-27
 
 - Fix: Preserve non-Raycast crontab content (env vars, other cron jobs, comments) across every save operation.
 - Fix: Correct `no crontab` detection and macOS TCC permission error handling after switching from `exec` to `spawn`.


pernielsentikaer:
@greptileai can you do a fresh review of this PR?

PR #25932, opened 2026-03-01
[Skills] Use explicit npx path resolution on macOS

raulgg:
## Description

This change updates the CLI execution path to function more reliably on macOS setups where `npx` is not available in the default shell environment. In my case, the extension was not working because it couldn't resolve `npx` properly.

The previous implementation relied on zsh login-shell execution. To make this more robust, this change follows the same general approach used by other Raycast extensions, and specifically I followed the approach in [ccusage](https://github.com/raycast/extensions/tree/main/extensions/ccusage): resolve `npx` from known Node installation locations and version-manager paths instead of depending on shell startup behavior.

Additionally, the changes allow users with uncommon setups to specify a custom path for `npx`, enabling them to direct the extension to the exact `npx` path they wish to use.

## Checklist

### DIFF
--- a/extensions/skills/CHANGELOG.md
+++ b/extensions/skills/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## [Improve macOS `npx` Path Resolution] - {PR_MERGE_DATE}
+
+- Run the Skills CLI without spawning a login shell by building an explicit PATH for Homebrew and common Node.js version-manager installs
+- Add a custom `npx` path preference for non-standard setups
+- Show clearer recovery guidance for `npx` and Skills CLI failures, including a shortcut to open Extension Preference
+
 ## [Fix Incomplete Agent List] - 2026-03-17
 
 - Use `skills list --json` for structured output instead of parsing ANSI text
--- a/extensions/skills/CHANGELOG.md
+++ b/extensions/skills/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## [Improve macOS `npx` Path Resolution] - {PR_MERGE_DATE}
+## [Improve macOS `npx` Path Resolution] - 2026-03-20
 
 - Run the Skills CLI without spawning a login shell by building an explicit PATH for Homebrew and common Node.js version-manager installs
 - Add a custom `npx` path preference for non-standard setups
--- a/extensions/skills/eslint.config.js
+++ b/extensions/skills/eslint.config.js
@@ -1,5 +1,15 @@
 const { defineConfig } = require("eslint/config");
 const raycastConfig = require("@raycast/eslint-config");
+const nodePlugin = require("eslint-plugin-n");
 
-module.exports = defineConfig([...raycastConfig]);
-
+module.exports = defineConfig([
+  ...raycastConfig,
+  {
+    plugins: {
+      n: nodePlugin,
+    },
+    rules: {
+      "n/prefer-node-protocol": ["error", { version: ">=16.0.0" }],
+    },
+  },
+]);
--- a/extensions/skills/src/manage-skills.tsx
+++ b/extensions/skills/src/manage-skills.tsx
@@ -1,6 +1,7 @@
-import { List, Icon, Detail, ActionPanel, Action, Color } from "@raycast/api";
+import { List, Icon, Detail, ActionPanel, Action, Color, openExtensionPreferences } from "@raycast/api";
 import { useState } from "react";
 import { useInstalledSkills } from "./hooks/useInstalledSkills";
+import { isNpxResolutionError } from "./utils/skills-cli";
 import { InstalledSkillListItem } from "./components/InstalledSkillListItem";
 import { UpdateSkillAction } from "./components/actions/UpdateSkillAction";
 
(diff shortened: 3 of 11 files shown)


raulgg:
@keito4 @DaleSeo these changes are fixing issues I've been experiencing with the extension in my computer.

image

As mentioned in the description, the extension will now search for common paths. If it still doesn't work, the user can configure a custom path in the extension preferences.

This should work well for most macOS users. However, I haven't been able to verify whether everything functions correctly on Windows. No significant changes were made for Windows, except for adding support for the custom path, which should work on both platforms.

Are you able to take a look to these changes?

extensions/skills/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## [Improve macOS `npx`Path Resolution] - {PR_MERGE_DATE}
```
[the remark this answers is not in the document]
raulgg: Fixed!

extensions/skills/src/manage-skills.tsx
```diff
@@ -10,14 +11,46 @@ export default function Command() {
...
+**Error:** ${error?.message}
+
+---
+
+This is an npx resolution issue in the local CLI runtime.
+
+1. Run \`which npx\` in Terminal.
+2. Open Extension Preferences (\`Cmd+Shift+,\`).
+3. Set **Custom npx Path** to the path from step 1, then retry.
+`;
```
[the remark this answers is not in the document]
raulgg: Good catch! Done ✅

extensions/skills/src/utils/node-path-resolver.ts
```diff
@@ -0,0 +1,123 @@
...
+  const versionManagerPaths = resolveVersionManagerPaths();
+
+  const systemPaths = ["/usr/bin", "/bin"];
+
+  if (process.env.HOME) {
+    systemPaths.push(`${process.env.HOME}/.npm/bin`, `${process.env.HOME}/.yarn/bin`);
+  }
+
+  // Preserve the current PATH first, then prepend common package-manager and version-manager locations
+  const allPaths = [process.env.PATH || "", ...platformPaths, ...versionManagerPaths, ...systemPaths];
```
DaleSeo: Would it be more efficient to search `platformPaths` and `versionManagerPaths` first since it's unlikey that `npx` is in `process.env.PATH`?

extensions/skills/src/utils/exec-options.ts
```diff
@@ -0,0 +1,45 @@
...
+    if (!process.env.FNM_DIR) {
+      const fnmBaseDir = resolveFnmBaseDir(home);
+      if (fnmBaseDir) {
+        env.FNM_DIR = fnmBaseDir;
+      }
+    }
+
+    if (!process.env.npm_config_prefix) {
+      env.npm_config_prefix = `${home}/.npm-global`;
+    }
```
DaleSeo: How about only setting this if the directory actually exists?


```suggestion
    const npmGlobalDir = `${home}/.npm-global`;
    if (!process.env.npm_config_prefix && existsSync(npmGlobalDir)) {
      env.npm_config_prefix = npmGlobalDir;
    }
```

extensions/skills/src/utils/node-path-resolver.ts
```diff
@@ -0,0 +1,123 @@
...
+export const resolveVersionManagerPaths = (): string[] => {
+  const versionedPaths: Array<{ path: string; version: string }> = [];
+  const staticPaths: string[] = [];
+  const home = process.env.HOME;
+
+  if (!home) return [];
+
+  const nvmVersionsDir = join(home, ".nvm", "versions", "node");
+  try {
+    const nodeVersions = readdirSync(nvmVersionsDir);
```
DaleSeo: Have you conisdered using async functions from `node:fs` like `readdir` insetad of `readdirSync` and making `resolveVersionManagerPaths` asynchronous? I'm concerned if the user has many nvm/fnm versions installed, this could cause a UI freeze on first load.

extensions/skills/src/utils/skills-cli.ts
```diff
@@ -1,23 +1,41 @@
...
+
+function buildSkillsCliCommand(npxCommand: string, args: string[]): string {
+  return [npxCommand, "-y", "skills@latest", ...args].map(shellEscape).join(" ");
+}
+
+async function runSkillsCli(args: string[]): Promise<string> {
+  const npxCommand = getCustomNpxPath() ?? "npx";
+  try {
+    const { stdout } = await execAsync(buildSkillsCliCommand(npxCommand, args), getExecOptions());
+    return stdout.toString();
```
DaleSeo:
```suggestion
    return stdout;
```

DaleSeo (commented):
Thanks for improving reliability for macOS users, @raulgg! I've left minor comments below.

extensions/skills/eslint.config.js
```diff
@@ -1,5 +1,15 @@
...
-module.exports = defineConfig([...raycastConfig]);
-
+module.exports = defineConfig([
+  ...raycastConfig,
+  {
+    plugins: {
+      n: nodePlugin,
+    },
+    rules: {
+      "n/prefer-node-protocol": ["error", { version: ">=16.0.0" }],
```
DaleSeo, line 12: Cool!

pernielsentikaer:
Hi @raulgg 👋 

Can you look into feedback from @DaleSeo

keito4:
Thanks for the additional context. I reviewed the latest changes and the macOS path resolution approach looks good to me.
Looks good to me!

Non-blocking follow-up: the npx resolution guidance in the error UI is still a bit macOS-specific (which npx, Cmd+Shift+,), so we can make that platform-aware later if needed.

pernielsentikaer:
We also have a merge conflict that we need to resolve @raulgg 😊

extensions/skills/src/utils/node-path-resolver.ts
raulgg: Initially I was more in favor of preserving the precedence than worrying about a minimal performance improvement. However you raise a really good point, and is true that finding the right path to `npx` in `process.env.PATH` is very unlikely in the Raycast extension, therefore I agree with your approach to it and I moved it to the last position, just as a fallback.

extensions/skills/src/utils/node-path-resolver.ts
raulgg: Good suggestion to use async, thank you! I had not considered such scenario. I have done a refactor to make the path resolution async and hopefully help to improve performance by a little bit.

raulgg:
> Thanks for the additional context. I reviewed the latest changes and the macOS path resolution approach looks good to me. Looks good to me!
> 
> Non-blocking follow-up: the npx resolution guidance in the error UI is still a bit macOS-specific (which npx, Cmd+Shift+,), so we can make that platform-aware later if needed.

Thank you  so much @keito4! Indeed the error is very macOS-specific. I'll do a follow up to improve that.

raulgg:
> We also have a merge conflict that we need to resolve @raulgg 😊

Feedback and conflicts addressed 🙏

raulgg:
@greptile-apps do a fresh review

extensions/skills/src/preferences.ts
```diff
@@ -0,0 +1,20 @@
...
+  if (customPath === "~") {
+    return homedir();
+  }
+
+  if (customPath.startsWith("~/") || customPath.startsWith(`~${sep}`)) {
+    return homedir() + customPath.slice(1);
+  }
+
+  return customPath;
+};
```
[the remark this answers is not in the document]
raulgg: I prefer to leave it as it is, as I believe it would be better for the custom path to fail if it's incorrect. In a follow-up, I will enhance the error message to better guide the user toward a resolution.

pernielsentikaer (commented):
Can you approve this @DaleSeo

pernielsentikaer:
CC @keito4 🙂

keito4 (approved):
Thanks for addressing the earlier feedback. From my side, the main macOS path resolution change still looks good.

I do think the bare ~ case Greptile pointed out is a real edge case, but it’s non-blocking for me.

LGTM! thanks

pernielsentikaer:
Let's get this out and then we can look into the other issue 🙂

pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #25935, opened 2026-03-01
Add identify-phone-utils extension

degouville:
##  Description                                                                             
                                                                                          
Phone Number Utils is a zero-dependency extension for working with international phone numbers. No accounts, no API calls/keys needed.
                                                                                          
It ships 6 commands:                                                                  
                                               
View commands (pre-fill from selected text, fallback to clipboard):
- Identify Phone Country — paste or type any number to instantly see its country flag,
  name, and dial code
- Format Phone Number — shows the same number in four formats side by side: E.164, US
  (XXX) XXX-XXXX, RFC 3966, and digits-only
- Extract Phone Numbers — finds every phone number in a block of text (clipboard or
  selection), with country identification for each
- Look Up Country Code — search ~230 countries by name or dial code; sorted by relevance

Background commands (no UI, act on clipboard directly):
- Format as US Number — rewrites clipboard as (XXX) XXX-XXXX, pastes result
- Remove Formatting — strips all spaces, dashes, and parentheses from clipboard number

##  Screencast
identify-phone-utils-1
identify-phone-utils-2
identify-phone-utils-3

##  Checklist

[x] I read the https://developers.raycast.com/basics/prepare-an-extension-for-store
[x] I read the https://developers.raycast.com/basics/publish-an-extension
[x] I ran npm run build and https://developers.raycast.com/basics/prepare-an-extension-for-store#metadata-and-configuration
[x] I checked that files in the assets folder are used by the extension itself
[x] I checked that assets used by the README are placed outside of the metadata folder


degouville:
what-do-you-think-gif

pernielsentikaer (changes requested):
Hi @degouville 👋

Could you look into feedback from Greptile?

---

degouville:
here-we-go-again-gif

degouville:
all done @pernielsentikaer 

waiting for @greptile-apps wisdom

pernielsentikaer:
@degouville it's somehow already possible in the search country, isn't it?

country-lookup 2026-03-06 at 13 55 05

Maybe it could live in that to make it even better?

degouville:
> @degouville it's somehow already possible in the search country, isn't it?

I hadn't considered looking for "search country" in the store for phone numbers.

Eventually could it be renamed it from "Country Lookup" to "Phone Number Lookup"?

I need something memorable for every time I see a phone numbers

pernielsentikaer:
Would you be up for improving the current, only thing missing might be that you can search for full phone number?

degouville:
> Would you be up for improving the current, only thing missing might be that you can search for full phone number?

I'm fine with improving the existing but I also believe a key missing element is that the "Country Lookup" feature provides **no indication that it helps with phone number** identification

Is it ok to update the name and icon of Country Lookup ?

I think the current name is misleading

(I know we can use aliases as a quick fix, but it would really help since Country Lookup doesn't show up right now when you're searching for things like the Phone Utils extension)

Screenshot 2026-03-09 at 10 05 49

Screenshot 2026-03-09 at 10 09 10_bordered

pernielsentikaer:
We can also update the `description` and `keywords` in `package.json` - could that work?

PR #25954, opened 2026-03-02
[Skills] Add support to parse and display security audits data

raulgg:
## Description

Installing a skill from the extension is very easy today, which is great and has huge value but also means users can install something risky with very little friction.

The site skills.sh already exposes security audit information for each skill, but that data is not currently available through their API. This change bridges that gap by parsing the audit information from the skills.sh website directly and displaying it inside the extension so users can make a more informed decision before installing a skill.

This is intentionally a pragmatic first step. The current implementation relies on parsing the website HTML, which is fragile and may break if the site structure changes in the future; fallback handling was added for cases in which parsing the information is not possible. Even with the limitation, showing this information in the extension is better than leaving users unaware when a skill has failed audits.

### What changed

- Fetch and parse skill security audit data from the skills.sh detail page
- Show audit status in the skill detail view
- Add quick actions to open the original audit links
- Warn users before installing a skill with failed audits
- Keep users informed when audit data is unavailable or could not be parsed

## Screencast

screen1
screen2
screen3
screen4

## Checklist

### DIFF
--- a/extensions/skills/CHANGELOG.md
+++ b/extensions/skills/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## [Security Audits Data] - {PR_MERGE_DATE}
+## [Security Audits Data] - 2026-03-11
 
 - Parse security audit data from skills.sh and show their status in the skill's details
 - Warn before installing skills with failed security audits


0xdhrv:
Hey @keito4 @DaleSeo 👋 


Can you please take a look?

keito4:
@raulgg 

Really nice step forward — surfacing skills.sh security audit signals before install is a pragmatic UX/security win, and the detail view + quick links are a great touch.

Before approving, could we harden a couple things? This runs on the install path and relies on scraped HTML, so I’d like to make sure it never throws and degrades gracefully:
- In fetchSkillAudits, please wrap both response.text() and parseSecurityAuditsFromHtml(...) in try/catch so this can’t throw and always returns a consistent SkillAuditsResult (fetch-error / parse-error).
- In InstallSkillAction.handleInstall, add a catch around fetchSkillAudits (currently try/finally) so audit lookup failures don’t crash/abort the flow; degrade gracefully to “couldn’t verify security audits” and let the user decide.
- The confirm title Install unsafe ... feels a bit too definitive for a best-effort signal. Would you mind rephrasing to something more neutral, e.g. “Install … despite failed security audits?” (for confirmed failures) / “Install … without verified security audits?” (when audits can’t be checked).

If you address the items above, I’m happy to re-review/approve.

extensions/skills/src/components/actions/InstallSkillAction.tsx
```diff
@@ -37,5 +63,47 @@ export function InstallSkillAction({ skill }: InstallSkillActionProps) {
...
+        await loadingToast.hide();
+      }
+    }
+
+    if (!auditResult) {
+      auditResult = {
+        audits: [],
+        availability: "fetch-error",
+      };
+    }
```
DaleSeo: This fallback logic looks unnecessary because `auditResult` will always be a `SkillAuditsResult` at this point. The `fetchSkillAudits` function takes care of all error cases internally and never throws an error. So, the second null-guard after the `try/finally` block will never be reached.

extensions/skills/src/utils/skill-audits.ts
```diff
@@ -0,0 +1,161 @@
...
+export type SkillAuditsResult = {
+  audits: SkillAudit[];
+  availability: SkillAuditsAvailability;
+};
+
+const PROVIDERS_BY_SLUG = {
+  "agent-trust-hub": "agent-trust-hub",
+  socket: "socket",
+  snyk: "snyk",
+} as const;
```
DaleSeo: Each key corresponds to its own value, which makes it an identity lookup. This would be a better use case for `Set` instead, and it can help avoid repetition.

```suggestion
const KNOWN_PROVIDERS = new Set<AuditProvider>(["agent-trust-hub", "socket", "snyk"]);
```

And update the usage in the parsing loop:

```ts
    if (!KNOWN_PROVIDERS.has(slug as AuditProvider)) continue;
    const provider = slug as AuditProvider;
```

extensions/skills/src/utils/skill-audits.ts
```diff
@@ -0,0 +1,161 @@
...
+ */
+function extractSecurityAuditSection(mainHtml: string): string | undefined {
+  const lower = mainHtml.toLowerCase();
+  const start = lower.indexOf("security audits");
+  if (start < 0) return undefined;
+
+  const after = mainHtml.slice(start);
+  const afterLower = lower.slice(start);
+  const installedOn = afterLower.indexOf("installed on");
+  return installedOn >= 0 ? after.slice(0, installedOn) : after;
```
DaleSeo: Since the PR already calls out the fragility of HTML parsing, it would be helpful to add a comment explaining why this specific string was chosen. This way, future maintainers will understand that it serves as a heuristic boundary marker, not a stable selector.

```suggestion
  // "installed on" marks the start of the installs section that follows security audits on skills.sh.
  // This is a heuristic boundary — update if the page structure changes.
  const installedOn = afterLower.indexOf("installed on");
  return installedOn >= 0 ? after.slice(0, installedOn) : after;
```

0xdhrv:
Hey @raulgg 👋 

Can you please look into the suggested changes from @DaleSeo and @keito4?

---

pernielsentikaer:
We also need to resolve the merge conflicts mentioned above 🙂

raulgg:
@keito4 thank your so much for your kind review comments.

You were right that errors needed better handling to never block the install path. I have done a deeper refactor on my original code to better handle errors and ensure we are never blocking installs due to any unexpected issue while scrapping the skills.sh content to obtain the audits data. I also updated the copy in the confirm alert as you suggested, so we are not too definitive.

While working on the error handling I also improved the report action, so users can now use it from the toast actions too to report errors that they experience while parsing the audit data.

When you have a chance, let me know what you think.

raulgg:
@DaleSeo, thank you for your feedback and suggestions. I have removed the unnecessary fallback, replaced the object with the Set, and added a comment explaining the purpose of the "installed on" string, which is subject to change at any time. Your input there was spot on!

keito4:
Thanks for the updates — the points I raised look addressed now. Approving from my side!

DaleSeo:
@raulgg Looks like Greptile is flagging a duplicate error toast issue in https://github.com/raycast/extensions/pull/25954#discussion_r2915134641

DaleSeo:
Hey @raulgg, since this PR has already been merged, I put up a quick follow-up: #26236

raulgg:
> Hey @raulgg, since this PR has already been merged, I put up a quick follow-up: #26236

Thank you for tackling that!

PR #25989, opened 2026-03-03
Update ChatGPT extension with codexauth sign-in and model handling

edoedac0:
## Summary
- add dual authentication support so users can sign in with either an API key or ChatGPT codexauth, with local session storage and sign-in/sign-out actions
- route ChatGPT-authenticated chats through the Codex Responses backend and keep API-key flows intact, including abort/stream handling and clearer auth errors
- enforce auth-gated access across extension command screens so unauthenticated users consistently see the sign-in view
- restrict image-understanding commands to API-key auth only and show an API-key-required gate when launched without a key
- add ChatGPT/Codex model support filtering for codexauth sessions, automatic fallback for unsupported models, and include codex-supported models in API-key model options
- set `gpt-5.2` as the default model across models and AI commands, migrate legacy `gpt-4o-mini` defaults, and keep `chatgpt-4o-latest` last in selection order
- make the `apiKey` preference optional so codexauth-only usage works while preserving API-key functionality for supported features

## Validation
- run `npm run lint`
- run `npm run build`


pernielsentikaer (changes requested):
Hi @edoedac0 👋

I'm a bit unsure about this one for some reasone like:

* Authenticating against `chatgpt.com/backend-api` is using an unofficial/internal API. This is fragile and could break anytime OpenAI changes their backend, which would leave the extension broken for `codexauth` users
* It's essentially scraping `ChatGPT`'s session auth — unclear if this is within OpenAI's ToS
* The complexity is significant: `PKCE` flow, `token refresh`, `session storage`, `dual routing paths`, `model compatibility mapping` — a lot of surface area to maintain
* Missing `CHANGELOG`

What do you think @abielzulio?

---

edoedac0:
> Hi @edoedac0 👋
> 
> I'm a bit unsure about this one for some reasone like:
> 
>     * Authenticating against `chatgpt.com/backend-api` is using an unofficial/internal API. This is fragile and could break anytime OpenAI changes their backend, which would leave the extension broken for `codexauth` users
> 
>     * It's essentially scraping `ChatGPT`'s session auth — unclear if this is within OpenAI's ToS
> 
>     * The complexity is significant: `PKCE` flow, `token refresh`, `session storage`, `dual routing paths`, `model compatibility mapping` — a lot of surface area to maintain
> 
>     * Missing `CHANGELOG`
> 
> 
> What do you think @abielzulio?
> 

Thanks for the feedback @pernielsentikaer !

My intention with this implementation was to allow users to authenticate with their
own ChatGPT account instead of requiring an API key.

This follows the same pattern used by tools like OpenCode (which implementation was made with the help of openai team, source: https://x.com/thdxr/status/2009742070471082006) and some Codex-based
CLI integrations, where authenticated ChatGPT sessions send requests to the same
backend used by the ChatGPT web application. 

- According to OpenAI’s Codex documentation, the system supports
authentication flows like “Sign in with ChatGPT” and is designed
to power multiple clients such as CLI tools, IDE extensions,
and other applications. (https://developers.openai.com/codex/app-server/) also statement from openai employee confirming the use of chatgpt/codex usage in 3rd pary apps https://x.com/thsottiaux/status/2009742187484065881 , https://x.com/thsottiaux/status/2009714843587342393)

Because of that, this PR mirrors that same pattern so Raycast
users can authenticate with their ChatGPT subscription instead
of requiring an API key.

The extension does not bypass authentication or share credentials with the
extension developer — the user signs in directly and the session is stored locally.

pernielsentikaer:
Hi @edoedac0 👋 

Can you resolve the merge conflicts so we can have a fresh look?

edoedac0:
> Hi @edoedac0 👋
> 
> Can you resolve the merge conflicts so we can have a fresh look?

Yeah, here it is

extensions/chatgpt/src/hooks/useCommand.tsx
```diff
@@ -78,6 +78,11 @@ export function useCommand(): CommandHook {
           if (!storedCommands[defaultKey]) {
             allCommands[defaultKey] = DEFAULT_COMMANDS[defaultKey];
           }
+
+          const storedCommand = allCommands[defaultKey];
+          if (storedCommand?.model === "gpt-4o-mini") {
+            allCommands[defaultKey] = { ...storedCommand, model: "gpt-5.2" };
```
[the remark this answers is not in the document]
edoedac0: > [P2](#) **Migration is ephemeral — LocalStorage value is never updated**
> 
> The `gpt-4o-mini` → `gpt-5.2` migration updates the in-memory `allCommands` map via `setData(allCommands)`, but the patched value is never written back to `LocalStorage`. As a result, every time the extension loads, `storedCommands` still contains `gpt-4o-mini` and the migration re-runs needlessly. More importantly, if a user explicitly saved a command with model `gpt-4o-mini`, their preference silently reverts to `gpt-5.2` on each load without ever becoming permanent.
> 
> The same one-shot migration pattern in `useModel.tsx` (lines 139–148) has the same issue. Both should either persist the migrated value, or use a versioned migration flag stored in `LocalStorage` so the write happens exactly once.

They both let the user know that it switched the model, and the user can make it permanent by switching the model to the desired one 
image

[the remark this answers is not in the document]
edoedac0: it doesn’t, since the default model for all ai commands is gpt 5.2, so the user would need to explicitly set the model to gpt 4o mini and see that it would get replaced by gpt 5.2 because they need an api key.

pernielsentikaer (changes requested):
Thanks for all the updates, there is still some things we need to look into before this is ready as I see it:

- No token refresh — `refreshToken` is stored but never used anywhere. When the access token expires, `ChatGPT-auth` users will get opaque errors until they manually sign out and back in. This is a significant UX gap for a popular extension.  
Proxy broken for `ChatGPT` path — `codex-responses.ts` imports `cross-fetch` and passes `agent` in request options, but `cross-fetch` silently ignores the `agent` option. Proxy users on the `Codex` auth path won't have their proxy applied. `Greptile` also flagged this.

- Auth priority is backwards — If a user has both an API key AND a `ChatGPT` session, `resolveAuthStatus` returns `provider: "chatgpt"` (`ChatGPT` wins). This silently routes API-key users who try the sign-in through the `Codex` backend and hides their non-codex models. There's no way to switch back without signing out. API key should either take precedence or the user should be able to choose.

- Migration is ephemeral — The `gpt-4o-mini` → `gpt-5.2` migration in `useCommand.tsx` updates in-memory state via `setData()` but `Greptile` flagged that it never persists to `LocalStorage`, meaning it re-runs every launch. The author argued it's fine because "default model for all AI commands is `gpt-5.2`" — but any user who explicitly saved `gpt-4o-mini` as their model will have it silently overridden on every launch without persistence. This should either persist the migration or not do it at all.

- Loading flash for existing users — Every command now shows `<List isLoading={true} />` while `resolveAuthStatus()` runs (a `LocalStorage` read). This didn't exist before. Minor UX regression for the majority of users who already have an API key.

- "Working..." label on preferences button — The `isBusy` state from sign-in/sign-out operations changes the unrelated "Open Extension Preferences" button text to "Working…".

- `cross-fetch` unnecessary — Both `auth.ts` and `codex-responses.ts` import `cross-fetch`. Node 21+ has built-in `fetch`. `Greptile` flagged this too.

---

pernielsentikaer:
Hi 👋

I'm still investigating a bit about whether we're going to break the ToS, which we don't want with this extension 🙂

edoedac0:
yeah sure, but openai specifically allows this in their tos, and they also open-sourced their codex app server so people can use their codex/chatgpt subscription wherever they want, and build on top of that, as i did with this extension. they are officially (their employees and them) saying exactly this all around twitter lol

pernielsentikaer (commented):
I have been on contact with the team behind Codex and they said that it shouldn't make requests to that directly as it does.

This documents the sanctioned way to do it which is through app server: https://developers.openai.com/codex/app-server

edoedac0:
> I have been on contact with the team behind Codex and they said that it shouldn't make requests to that directly as it does.
> 
> This documents the sanctioned way to do it which is through app server: https://developers.openai.com/codex/app-server

Thanks, I don’t think the documentation existed when I was first updating the extension, so i just copied the opencodes implementation at the time. This is better, and i’ve updated it to follow their docs 100%.

I have also added support for gpt 5.4 models

pernielsentikaer:
Thanks for the update, I’ll do another review 🙂

pernielsentikaer (commented):
Hi @edoedac0 👋

It seems to work as expected for me, but I'm not sure we have the best UX for this yet. It will continue to work for everyone who has it installed today since they will have the key. New users will always be forced to use the new method until they add the key, right?

edoedac0:
> Hi @edoedac0 👋
> 
> It seems to work as expected for me, but I'm not sure we have the best UX for this yet. It will continue to work for everyone who has it installed today since they will have the key. New users will always be forced to use the new method until they add the key, right?

Hey @pernielsentikaer 👋,
I’m not sure what you mean by “forced to use the new method.” I made it so new users who install the extension have the option to authenticate via the API key or sign in with ChatGPT (Codex method). I intentionally made the ChatGPT method the preferred one, as I think users would use it more since it draws from their usage, which is far cheaper than API costs and is included in all ChatGPT subscriptions (Free, Plus, Pro Lite, and Pro). Users who are already authenticated with an API key can sign in with ChatGPT and also use their usage and API key only for things that I did not make Codex support, like image understanding, which I can also add in future PRs. But I want to see how this will work first. At any time, users can remove either the ChatGPT or API method and use only one. If you want, or if you’re not sure about something, I can change it. I also pushed one commit that should make the extension with the app server method respond much faster.

pernielsentikaer (changes requested):
We should make update the `README` with the new changes explaining what's happening

edoedac0:
> We should make update the `README` with the new changes explaining what's happening

Updated, I think it should be good now

pernielsentikaer (commented):
@0xdhrv can you give this a spin on Windows too?

pernielsentikaer (commented):
Can you check the update @edoedac0 - we had to update it a bit, can you test it on Windows?

edoedac0:
> Can you check the update @edoedac0 - we had to update it a bit, can you test it on Windows?

Hey, sorry for the late response. I do have a Windows machine, but I don't have Raycast on it and don't use it much, so I think it would be better if somebody who actually uses Raycast on Windows could test it.

pernielsentikaer:
@0xdhrv did you test this?

edoedac0:
> @0xdhrv did you test this?

hey, any updates?

0xdhrv:
> > @0xdhrv did you test this?
> 
> hey, any updates?

Hey, this got lost, I will test this today

0xdhrv:
> > @0xdhrv did you test this?
> 
> hey, any updates?

Hey @pernielsentikaer @edoedac0 

I tested this with the latest update. I logged out completely and logged in again, and also tested it with a few queries. These two work fine for me on Windows.

However, I see only default models in the dropdown.

image

---

Also, on macOS, I see the following warning and can't ask any questions.

Dhruvs- 07022026144719@2x

edoedac0:
> > > @0xdhrv did you test this?
> 
> > 
> 
> > hey, any updates?
> 
> 
> 
> Hey @pernielsentikaer @edoedac0 
> 
> 
> 
> I tested this with the latest update. I logged out completely and logged in again, and also tested it with a few queries. These two work fine for me on Windows.
> 
> 
> 
> However, I see only default models in the dropdown.
> 
> 
> 
> image
> 
> 
> 
> ---
> 
> 
> 
> Also, on macOS, I see the following warning and can't ask any questions.
> 
> 
> 
> Dhruvs- 07022026144719@2x
> 
> 

Hey, so regarding the first issue, that was the behavior of the extension before, where you have a default model that you can change in the settings. The other issue was a Codex issue that was present on the day you were testing it. That should be fixed by now

pernielsentikaer (commented):
@edoedac0 did you push any fixed and @0xdhrv can you test again?

edoedac0:
> @edoedac0 did you push any fixed and @0xdhrv can you test again?

No, the bug was on their side but they fixed it and it should automatically work now

pernielsentikaer (commented):
@0xdhrv can you check this again?

edoedac0:
hey, any updates on this?

extensions/chatgpt/src/utils/codex-app-server.ts
```diff
@@ -0,0 +1,637 @@
...
+  return `https://registry.npmjs.org/${encodedName}/-/${name.split("/").pop()}-${version}.tgz`;
+}
+
+async function downloadFile(url: string, destinationPath: string): Promise<void> {
+  const response = await fetch(url);
+  if (!response.ok || !response.body) {
+    throw new Error(`Download failed (${response.status} ${response.statusText})`);
+  }
+
+  await pipeline(Readable.fromWeb(response.body as never), createWriteStream(destinationPath));
```
0xdhrv, line 590: **[BLOCKING]** Could we verify the integrity of the downloaded Codex archive before extracting and executing it?

This path downloads an executable runtime from the network and later launches the binary, but currently only checks `response.ok`. A compromised or unexpected archive would therefore be trusted automatically.

Please validate the downloaded archive against a pinned checksum/integrity value before extraction.

extensions/chatgpt/src/utils/codex-app-server.ts
```diff
@@ -0,0 +1,637 @@
...
+    await fs.mkdir(extractRoot, { recursive: true });
+    await tar.x({
+      file: archivePath,
+      cwd: extractRoot,
+      strip: 1,
+    });
+
+    await fs.rm(runtimeRoot, { recursive: true, force: true });
+    await fs.mkdir(path.dirname(runtimeRoot), { recursive: true });
+    await fs.cp(extractRoot, runtimeRoot, { recursive: true });
```
0xdhrv, line 566: Could we make this installation atomic?

Right now we remove `runtimeRoot` before copying the newly extracted runtime into place. If the copy fails or Raycast exits in between, the previously working runtime is lost.

It would be safer to fully extract/validate into a temporary directory first and then atomically rename it into place.

extensions/chatgpt/src/utils/codex-app-server.ts
```diff
@@ -0,0 +1,637 @@
...
+  if (!name || !version) {
+    throw new Error(`Invalid Codex package specifier: ${packageSpecifier}`);
+  }
+
+  const encodedName = name.startsWith("@") ? `@${encodeURIComponent(name.slice(1))}` : encodeURIComponent(name);
+  return `https://registry.npmjs.org/${encodedName}/-/${name.split("/").pop()}-${version}.tgz`;
+}
+
+async function downloadFile(url: string, destinationPath: string): Promise<void> {
+  const response = await fetch(url);
```
0xdhrv, line 585: Could we add a timeout/`AbortSignal` to this download?

A stalled registry/network request can currently leave runtime setup and therefore ChatGPT sign-in waiting indefinitely.

extensions/chatgpt/src/utils/codex-app-server.ts
```diff
@@ -0,0 +1,637 @@
...
+  reject: (error: Error) => void;
+}
+
+const CLIENT_INFO = {
+  name: "raycast_chatgpt_extension",
+  title: "Raycast ChatGPT Extension",
+  version: "0.1.0",
+};
+
+const CODEX_VERSION = "0.122.0";
```
0xdhrv, line 48: `CODEX_VERSION` and the platform/package mapping are duplicated in both this file and `vendor-codex-runtime.mjs`.

Could we keep this configuration in a single shared source? Otherwise a future Codex upgrade can easily update the build-time version while leaving the runtime fallback on a different version.

extensions/chatgpt/src/utils/codex-app-server.ts
```diff
@@ -0,0 +1,637 @@
...
+
+const PLATFORM_PACKAGE_BY_TARGET: Record<string, string> = {
+  "x86_64-apple-darwin": "@openai/codex@0.122.0-darwin-x64",
+  "aarch64-apple-darwin": "@openai/codex@0.122.0-darwin-arm64",
+  "x86_64-pc-windows-msvc": "@openai/codex@0.122.0-win32-x64",
+  "aarch64-pc-windows-msvc": "@openai/codex@0.122.0-win32-arm64",
+};
```
0xdhrv, line 55: Could we clarify how all supported platforms are expected to receive a bundled runtime?

The vendoring script only packages the runtime matching the machine running the build, while this runtime code supports four targets. That means the other architectures/platforms appear to depend on the network-download fallback.

We should either bundle all supported targets or make the intended distribution strategy explicit.

extensions/chatgpt/scripts/vendor-codex-runtime.mjs
```diff
@@ -0,0 +1,90 @@
...
+  if (await fileExists(archivePath)) {
+    return;
+  }
+
+  await fs.mkdir(assetsDir, { recursive: true });
+
+  const { stdout } = await execNpm(["pack", packageSpecifier, "--json", "--pack-destination", assetsDir]);
+
+  const [{ filename }] = JSON.parse(stdout);
+  await fs.rename(path.join(assetsDir, filename), archivePath);
```
0xdhrv, line 38: **[BLOCKING]** This only vendors the Codex archive for the current build platform.

If publishing happens on macOS ARM, for example, only `aarch64-apple-darwin` is added to the extension assets even though the extension declares support for Intel macOS and Windows as well.

Could we vendor every supported runtime during publishing rather than relying on users on other platforms to download executable code after install?

extensions/chatgpt/scripts/vendor-codex-runtime.mjs
```diff
@@ -0,0 +1,90 @@
...
+  const packageSpecifier = PLATFORM_PACKAGE_BY_TARGET[targetTriple];
+  if (!packageSpecifier) {
+    console.warn(`Skipping Codex runtime vendoring for unsupported platform ${process.platform} (${process.arch}).`);
+    return;
+  }
+
+  const archivePath = path.join(assetsDir, `${targetTriple}-${CODEX_VERSION}.tgz`);
+  if (await fileExists(archivePath)) {
+    return;
+  }
```
0xdhrv, line 31: Could we validate the existing archive before returning here?

At the moment any file at this path is treated as a valid Codex runtime archive purely because it exists. A partially written or stale archive could therefore survive subsequent builds.

extensions/chatgpt/src/utils/model-support.ts
```diff
@@ -0,0 +1,118 @@
...
+  "gpt-5.1-codex-max",
+  "gpt-5.4-mini",
+  "gpt-5.3-codex",
+  "gpt-5.2",
+  "gpt-5.1-codex-mini",
+] as const;
+
+export const MODEL_OPTION_LAST_IN_SELECTION = "chatgpt-4o-latest";
+
+export const DEFAULT_CHATGPT_CODEX_MODEL = "gpt-5.4-mini";
```
0xdhrv, line 16: **[BLOCKING]** The default model is inconsistent across this PR.

This constant uses `gpt-5.4-mini`, while the PR summary/changelog describe `gpt-5.2` as the new default.

Could we pick one default and make the implementation, changelog, README, and migration behavior consistent?

extensions/chatgpt/src/utils/model-support.ts
```diff
@@ -0,0 +1,118 @@
...
+
+export const CHATGPT_CODEX_SUPPORTED_MODELS = [
+  "gpt-5.4",
+  "gpt-5.2-codex",
+  "gpt-5.1-codex-max",
+  "gpt-5.4-mini",
+  "gpt-5.3-codex",
+  "gpt-5.2",
+  "gpt-5.1-codex-mini",
+] as const;
```
0xdhrv, line 12: Could the Codex `model/list` response be the primary source of truth here?

We already query supported models from the app-server, but also maintain a static list. The hard-coded list will become stale as Codex model availability changes.

A static list would make more sense as a fallback when model discovery fails.

extensions/chatgpt/src/utils/model-support.ts
```diff
@@ -0,0 +1,118 @@
...
+
+export const MODEL_OPTION_LAST_IN_SELECTION = "chatgpt-4o-latest";
+
+export const DEFAULT_CHATGPT_CODEX_MODEL = "gpt-5.4-mini";
+
+export function resolveModelOptionForAuth(
+  option: string,
+  authProvider: AuthProvider,
+  availableOptions?: string[],
+): string {
```
0xdhrv, line 22: `authProvider` is passed into this function but currently isn't used.

Either the provider should influence resolution, or we should remove the argument so callers don't assume this function handles provider-specific behavior that it actually ignores.

extensions/chatgpt/src/utils/auth.ts
```diff
@@ -0,0 +1,227 @@
...
+    expiresAt: Date.now() + AUTH_STATUS_CACHE_MS,
+  };
+
+  void LocalStorage.setItem(
+    AUTH_STATUS_STORAGE_KEY,
+    JSON.stringify({
+      account,
+      expiresAt: Date.now() + AUTH_STATUS_CACHE_MS,
+    }),
+  );
```
0xdhrv, line 184: Do we need to persist the user's email and plan type to `LocalStorage` for a 15-second auth cache?

This information is only used briefly for auth status and could potentially remain in memory. If persistence is required, it'd be good to minimize the stored account metadata.

extensions/chatgpt/src/utils/auth.ts
```diff
@@ -0,0 +1,227 @@
...
+    return cachedChatGPTAccount.account;
+  }
+
+  try {
+    const account = await withCodexAppServer((client) => readChatGPTAccount(client));
+    primeCachedChatGPTAccount(account);
+    return account;
+  } catch {
+    return null;
+  }
```
0xdhrv, line 156: This catches every app-server/auth error and converts it into "not signed in".

That makes runtime setup failures, corrupted Codex installations, JSON-RPC errors, and genuinely logged-out sessions indistinguishable to callers.

Could we preserve actionable runtime errors while only treating an actual unauthenticated account response as `null`?

extensions/chatgpt/src/utils/auth.ts
```diff
@@ -0,0 +1,227 @@
...
+  });
+}
+
+export async function clearCodexAuthSession(): Promise<void> {
+  await withCodexAppServer(async (client) => {
+    await client.request("account/logout", {});
+  });
+  clearCachedChatGPTAccount();
+  await LocalStorage.removeItem(AUTH_STATUS_STORAGE_KEY);
+}
```
0xdhrv, line 143: Could we clear the local auth cache even if `account/logout` fails?

Right now an app-server error prevents `clearCachedChatGPTAccount()` and `LocalStorage.removeItem()` from running. A `finally` block would avoid leaving stale local auth state behind.

extensions/chatgpt/src/utils/codex-responses.ts
```diff
@@ -0,0 +1,501 @@
...
+  }
+
+  const message = error.message?.trim() || "ChatGPT request failed in Codex app-server.";
+  const details = error.additionalDetails?.trim();
+  return details ? `${message} ${details}` : message;
+}
+
+function supportsReasoningEffort(model: string): boolean {
+  return model.startsWith("gpt-5");
+}
```
0xdhrv, line 334: Checking `model.startsWith("gpt-5")` seems too broad for deciding whether `effort` is supported.

Could we derive this from the model metadata returned by Codex, or maintain an explicit capability check instead? Otherwise a future GPT-5 model without this parameter may receive an unsupported option.

extensions/chatgpt/src/utils/codex-responses.ts
```diff
@@ -0,0 +1,501 @@
...
+  return (
+    (message.includes("thread") || message.includes("rollout")) &&
+    (message.includes("not loaded") ||
+      message.includes("not found") ||
+      message.includes("unknown") ||
+      message.includes("closed") ||
+      message.includes("unavailable") ||
+      message.includes("no rollout found"))
+  );
+}
```
0xdhrv, line 239: Could we avoid deciding retry behavior by matching fragments of error-message text?

This is fairly brittle against Codex wording changes/localization. If the app-server exposes structured JSON-RPC error codes/data, it'd be safer to retry based on those.

extensions/chatgpt/src/utils/codex-responses.ts
```diff
@@ -0,0 +1,501 @@
...
+      stream: options.stream,
+      signal: options.signal,
+      onDelta: options.onDelta,
+    });
+
+    return { text, threadId };
+  } catch (error) {
+    if (!threadId || !shouldRetryWithFreshThread(error)) {
+      throw error;
+    }
```
0xdhrv, line 159: Could we ensure abort errors never enter the fresh-thread retry path?

A cancellation should stop immediately rather than potentially creating a new thread and starting another turn because the error text happened to match the retry predicate.

extensions/chatgpt/src/utils/codex-responses.ts
```diff
@@ -0,0 +1,501 @@
...
+      result.push({ type: "text", text: part.text, text_elements: [] });
+      continue;
+    }
+
+    if (part.type === "image_url" && "image_url" in part) {
+      const imageUrl = resolveImageUrl(part.image_url);
+      if (imageUrl.trim()) {
+        result.push({ type: "image", url: imageUrl });
+      }
+    }
```
0xdhrv, line 452: The PR says image understanding is API-key-only, but the Codex request path still accepts and forwards image inputs.

Could we either remove image handling from this path or explicitly guarantee at this boundary that Codex-auth requests can never contain image content?

extensions/chatgpt/package.json
```diff
@@ -358,24 +357,25 @@
...
   },
   "scripts": {
-    "build": "ray build -e dist",
-    "dev": "ray develop",
+    "build": "node ./scripts/vendor-codex-runtime.mjs && ray build -e dist",
+    "dev": "node ./scripts/vendor-codex-runtime.mjs && ray develop",
     "fix-lint": "ray lint --fix",
     "lint": "ray lint",
-    "publish": "ray publish",
+    "publish": "node ./scripts/vendor-codex-runtime.mjs && ray publish",
```
0xdhrv, line 378: Making `ray develop` run `npm pack` for the Codex runtime means development now has an implicit network/download step.

Could we separate runtime vendoring for publishing from normal development, or cache/verify this in a way that doesn't make `ray develop` dependent on npm registry availability?

extensions/chatgpt/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## [Feature] - 2026-03-17
```
0xdhrv, line 3: Please use `{PR_MERGE_DATE}` here instead of `2026-03-17`.

The PR is still open, so this date is already stale and would produce an incorrect release date when merged.

extensions/chatgpt/src/hooks/useModel.tsx
```diff
@@ -131,6 +77,15 @@ export function useModel(): ModelHook {
...
         }
+        if (!modelsById[DEFAULT_MODEL.id]) {
+          modelsById[DEFAULT_MODEL.id] = DEFAULT_MODEL;
+        } else if (!CHATGPT_CODEX_SUPPORTED_MODELS.includes(modelsById[DEFAULT_MODEL.id].option?.trim() as never)) {
+          modelsById[DEFAULT_MODEL.id] = {
+            ...modelsById[DEFAULT_MODEL.id],
+            option: DEFAULT_MODEL.option,
+            updated_at: new Date().toISOString(),
+          };
+        }
```
0xdhrv, line 88: Could we persist this model migration rather than only updating the in-memory value?

`useCommand.tsx` now writes hydrated commands back to `LocalStorage`, but this path still only `setData()`s the rewritten default model. The stored model can therefore remain `gpt-4o-mini` / an unsupported option, so the same migration/fallback can run again on later launches.

The stored and displayed model should converge after migration — please persist the updated models here as well.

0xdhrv (changes requested):
Thanks for the Codex app-server rewrite. Before this is ready, a few **blocking** items:

1. Downloaded Codex archives are executed after only checking `response.ok` — please pin and verify a checksum before extract/launch.
2. The vendoring script only packs the build machine's runtime, so Intel macOS and Windows appear to depend on a post-install network download of an executable.
3. Default model is inconsistent: code uses `gpt-5.4-mini`, while the PR summary/changelog still say `gpt-5.2`.

Next up: `readChatGPTAccountSafe()` swallows all runtime errors as "not signed in", and the model migration should persist so stored and displayed values converge.

Inline comments cover the rest (atomic install, download timeout, duplicated version config, auth cache, retry/abort, image inputs on the Codex path, `ray develop` registry dependency, and `{PR_MERGE_DATE}`).

PR #26025, opened 2026-03-04
System Monitor: Fix stale temperature readings in menubar

ADCAdams:
## Summary
- Having prev. made a standalone temperature extension, I noticed temp. reads were going stale by comparison
- Moved temperature sensor polling to its own dedicated 3-second interval, separate from the other stats
- The temperature reader is an external binary that talks to IOKit HID sensors, so it's inherently slower than the in-process stats (CPU, memory, network, etc.). When it was bundled into the same 1-second interval as everything else, revalidation calls would stack up and the readings would go stale.
- No change to polling frequency for other stats

## Test plan
- [ ] Pin CPU Temperature to the menubar
- [ ] Verify the reading updates consistently without freezing or lagging behind
- [ ] Verify other pinned stats (CPU %, memory, etc.) still update at normal speed

### DIFF
--- a/extensions/system-monitor/CHANGELOG.md
+++ b/extensions/system-monitor/CHANGELOG.md
@@ -1,6 +1,6 @@
 # System Monitor Changelog
 
-## [Fix Temperature Polling] - {PR_MERGE_DATE}
+## [Fix Temperature Polling] - 2026-03-04
 
 - Moved temperature sensor polling to a dedicated 3s interval to prevent stale readings
 


JadenMajid:
@ADCAdams the pinned value on my system is going stale unless I click the icon, have you noticed this too? I don't have the time to fix it myself but wanted to let you know.

ADCAdams:
Yes I noticed. Was embarrassed doing fix #2 on the same issue lol, I'll fix today.

JadenMajid:
No worries! Hope all is well

PR #26032, opened 2026-03-04
[Copy-Path] Add add URL cleanup preference for tracking and query parameters

cdanyl:
This PR introduces a new **URL Cleanup** preference that allows users to automatically sanitize URLs before they are copied to the clipboard. This is particularly useful for removing bulky tracking parameters or stripping queries entirely for a cleaner sharing experience.

### ✨ Features

*   **New Preference `urlCleanupMode`**: Added a dropdown in the extension settings with three modes:
    *   **None**: Copies the URL as-is.
    *   **Remove Tracking Params**: Strips common marketing and analytics trackers (e.g., `utm_*`, `gclid`, `fbclid`, `igshid`, etc.) while preserving legitimate query parameters.
    *   **Remove Query + Fragment (Default)**: Strips everything after the path (search params and hashes) for the cleanest possible URL.

image

### DIFF
--- a/extensions/copy-path/CHANGELOG.md
+++ b/extensions/copy-path/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Copy Path Changelog
 
+## [Add URL cleanup preference] - {PR_MERGE_DATE}
+
+- Add URL cleanup preference with modes to remove tracking params or strip query + fragment.
+- Default cleanup now strips query parameters and fragments for copied URLs.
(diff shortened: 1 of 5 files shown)


pernielsentikaer (changes requested):
I can see that you changes default behaviour with the use of `removeQueryAndFragment` - I don't think we should that change the default behaviour but let the user decide

---

PR #26038, opened 2026-03-04
Improved project filtering with client name

CylosoftGit:
## Description

- Changed: Project dropdown in New Time Entry and Favorites forms now displays client name alongside project name (e.g. "Acme Corp – My Project"), allowing you to filter by client or project in a single dropdown
- Changed: Projects are now sorted alphabetically within each client group
- Changed: Client groups are now sorted alphabetically

### DIFF
--- a/extensions/harvest/CHANGELOG.md
+++ b/extensions/harvest/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Harvest Changelog
 
-## [Improved Project Filtering With Clients Included] - {PR_MERGE_DATE}
+## [Improved Project Filtering With Clients Included] - 2026-03-09
 
 - Added: New preference "Show Client Name in Project Dropdown" (default off) — when enabled, the project dropdown additionally displays entries with client name for easier filtering by client name
 - Changed: Projects and client groups are now sorted alphabetically in the project dropdown


eluce2:
The client name is intentionally omitted from that title. For my team's use case it would be redundant to put it there because the code includes a shortened client name, but it is in the keywords so you can still find it when filtering.

If you think it is important to put it there, there would need to be a preference in the extension settings that is disabled by default so that there is no change for existing users unless they go and turn on the preference.

CylosoftGit:
> The client name is intentionally omitted from that title. For my team's use case it would be redundant to put it there because the code includes a shortened client name, but it is in the keywords so you can still find it when filtering.
> 
> If you think it is important to put it there, there would need to be a preference in the extension settings that is disabled by default so that there is no change for existing users unless they go and turn on the preference.

The problem with the existing code is it assumes that multiple clients don't have the same project name. You can end up with something like this when filtering. Raycast doesn't allow for the client titles to stick when filtering that's I've found (how Harvest web app would do).

image

eluce2:
Your screenshot shows that you aren't using project codes in Harvest, which could also distinguish them. If we can maintain the client's name in the group headings after filtered results, that would also be agreeable. It may also be that we could show the client name or the project code in the title, but never both. I'm open to that idea as well.

CylosoftGit:
I might just be missing it but I can't find that Raycast would allow for the group headings in filtered results. Or that the title can be changed conditionally based on filtering.

eluce2:
Yeah, it may not be possible. I know I looked at this a long time ago, but I'm not sure if the API has changed since.

I think a fair compromise is to show the client name or the project code, but never both, or have a preference control what it shown

CylosoftGit:
> Yeah, it may not be possible. I know I looked at this a long time ago, but I'm not sure if the API has changed since.
> 
> I think a fair compromise is to show the client name or the project code, but never both, or have a preference control what it shown

Added client name as an optional display, default off.

PR #26103, opened 2026-03-07
Quick Open Project: add Windows support

iml885203:
## Description

Add Windows support to the Quick Open Project extension. The `open` function now uses `cmd /c start` on Windows and `open -a` on macOS. Also removed the empty `env: {}` object from the spawn call which could cause issues with some apps.

## Screencast

Snipaste_2026-03-07_14-06-12

## Checklist

### DIFF
--- a/extensions/quick-open-project/CHANGELOG.md
+++ b/extensions/quick-open-project/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## [1.1.4] - {PR_MERGE_DATE}
+
+- Add Windows support for opening projects in editor and terminal
+
 ## [1.1.3] - 2025-02-24
 
 - Stop transferring environment variables to opened applications
--- a/extensions/quick-open-project/CHANGELOG.md
+++ b/extensions/quick-open-project/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## [1.1.4] - {PR_MERGE_DATE}
+## [1.1.4] - 2026-03-17
 
 - Add Windows support for opening projects in editor and terminal
 
--- a/extensions/quick-open-project/package-lock.json
+++ b/extensions/quick-open-project/package-lock.json
@@ -9,8 +9,8 @@
       "version": "1.1.0",
       "license": "MIT",
       "dependencies": {
-        "@raycast/api": "^1.86.1",
-        "@raycast/utils": "^1.17.0",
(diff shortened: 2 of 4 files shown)


iml885203:
@greptileai

iml885203:
@greptileai

iml885203:
@greptileai Thanks for the review! Here's what I've addressed:

1. **`.unref()` for detached child process** — Fixed in 264edb4. Added `stdio: "ignore"` and `.unref()` to the Windows `spawn` call.

2. **macOS-only default app paths** — The `appPicker` is a Raycast UI picker where users select an app from a list. If the default path doesn't exist, Raycast handles it gracefully and the user simply picks their preferred app on first use. These defaults are from the original author and outside the scope of this PR.

canac:
> Also removed the empty `env: {}` object from the spawn call which could cause issues with some apps.

@iml885203 Can you tell us more about the issues `env: {}` was causing? I added it because Raycast's `NODE_ENV=production` was being transferred to VS Code, which was being transferred to all terminal processes VS Code launched, which was causing jest test failures. See #15957 and #17158.

But the Windows support looks great! Thanks for adding that!

iml885203:
@canac Good catch — you are right, removing `env: {}` was an oversight on my part. I have restored it and also applied it to the Windows branch for consistency.

I renamed it to `envWithoutRaycastVars` so the intent is clear from the variable name alone (Raycast injects `NODE_ENV=production` which would leak into spawned editors/terminals).

Thanks for the context on #15957 and #17158!

iml885203:
Also updated the Windows branch to use filtered env instead of empty `{}` — on Windows, spawned processes directly inherit the provided env, so an empty object would strip PATH/TEMP/USERPROFILE and break the terminal session.

macOS keeps `env: {}` since `open -a` delegates to Launch Services and the app gets its env from launchd, not from the spawn call.

Reference: https://nodejs.org/api/child_process.html (see "options.env" behavior per platform)

pernielsentikaer (commented):
Thanks for the update @iml885203  🔥

@zerowidth do you want to check this PR?

iml885203:
Hi @zerowidth! Wondering if you've had a chance to take a look at this PR? Happy to address any feedback. Thanks! 🙏

pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #26122, opened 2026-03-07
Add compare-lists extension

gokhan:
## Description

Adds a new Raycast extension, `compare-lists`, for comparing two pasted newline-separated lists and reviewing items only in List A, only in List B, and shared between both. It supports case-sensitive and case-insensitive comparison, preserves first-seen order, and includes copy actions for each result set and the full report.

## Screencast

Screenshots are included for the input form, summary, A-only items, B-only items, shared items, and shared item actions.

## Checklist

- [x] I ran `npm run build` and tested the distribution build locally


0xdhrv (commented):

We already have an extension in the Store that deals with Diff Checking for List. Could we consider enhancing the existing extension below instead of creating another one?

- https://www.raycast.com/destiner/diff-checker by @Destiner 

The existing extension can be extended to support the features you want to add.

gokhan:
Hi @0xdhrv,

Thanks for the feedback.

I agree there is some overlap with [Diff Checker](https://www.raycast.com/destiner/diff-checker) by @Destiner, and I understand the concern about duplication.

That said, this extension is intended for a more specific list-comparison workflow and includes functionality that is central to its use case, not just incremental additions. In particular:

1.  optional case-sensitive comparison
2. a clearer result summary, for example:
                    In List A
                    Only in List A
                    In List B
                    Only in List B
                    Difference
                    In Both Lists
3. keyboard-first actions to copy:
      -   items only in List A
      -   items only in List B
      -  items present in both lists

My goal was not to recreate the existing extension, but to provide a more focused tool for comparing two lists and acting on the results immediately.

I’m open to discussion if you think these features would fit naturally into the existing extension.

0xdhrv:
> Hi @0xdhrv,
> 
> Thanks for the feedback.
> 
> I agree there is some overlap with [Diff Checker](https://www.raycast.com/destiner/diff-checker) by @Destiner, and I understand the concern about duplication.
> 
> That said, this extension is intended for a more specific list-comparison workflow and includes functionality that is central to its use case, not just incremental additions. In particular:
> 
> 1. optional case-sensitive comparison
> 2. a clearer result summary, for example:
>    In List A
>    Only in List A
>    In List B
>    Only in List B
>    Difference
>    In Both Lists
> 3. keyboard-first actions to copy:
>    
>    * items only in List A
>    * items only in List B
>    * items present in both lists
> 
> My goal was not to recreate the existing extension, but to provide a more focused tool for comparing two lists and acting on the results immediately.
> 
> I’m open to discussion if you think these features would fit naturally into the existing extension.

I understand. Since JSON Compare already had unique features that were merged into the same extension, it's better to add this functionality to the existing extension as a new command with a slightly different UX. That way users can find the right tool for their use case more easily when it's all in one extension.

PR #26147, opened 2026-03-09
Update toothpick extension

merchako:
### Motivation
- Provide a single action to "refresh" Bluetooth when a device becomes unresponsive by briefly reconnecting a device or toggling Bluetooth power, as requested in issue https://github.com/raycast/extensions/issues/26144.

### Implementation Details
- Add a per-device "Refresh" action in the Manage Bluetooth Connections panel that runs connect → wait → disconnect by introducing `src/core/devices/handlers/refresh-device.ts` and wiring it into `Device` actions (`src/core/devices/devices.model.tsx`).
- Add a `refresh-device` command and three favorite shortcuts (`refresh-favorite-device-1/2/3`) that reuse the existing fuzzy name/MAC lookup flow (`src/refresh-device.ts`, `src/refresh-favorite-device-1.ts`, `src/refresh-favorite-device-2.ts`, `src/refresh-favorite-device-3.ts`).
- Add a `refresh-all` command to toggle Bluetooth power off/on via a new handler `src/core/devices/handlers/refresh-bluetooth.ts` and command `src/refresh-all.ts`.
- Extend the devices service interface with `refreshBluetooth()` and implement it in blueutil only (`blueutil -p 0 && blueutil -p 1` in `src/core/devices/devices.service.blueutil.ts`).
- Register the new commands in `package.json` and add consistent user-facing messages via existing `showAnimatedMessage` / `showSuccessMessage` / `showErrorMessage` utilities.

### Limitations
- "Refresh All" requires `blueutil`

### Testing
- Passed `npm run lint`
- Hand-tested on macOS

### AI Disclosure
- Used ~12-15 prompts to [Codex](https://chatgpt.com/codex/tasks/task_e_69adee7705348331a79c11336a2ea1ee) and hand-modified the end result.

## Screencast
https://github.com/user-attachments/assets/d3a4ab68-6c89-4c09-942f-ed3f402db9c2

## Checklist

- [x] ~~I checked that files in the `assets` folder are used by the extension itself~~ not changed
- [x] ~~I checked that assets used by the `README` are placed outside of the `metadata` folder~~ not changed

### DIFF
--- a/extensions/toothpick/CHANGELOG.md
+++ b/extensions/toothpick/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Toothpick Changelog
 
-## [Refresh Device and Refresh All Commands] - {PR_MERGE_DATE}
+## [Refresh Device and Refresh All Commands] - 2026-05-05
 
 - Added a **Refresh** action in Manage Bluetooth Connections which disconnects and reconnects a device.
 - Added commands: **Refresh Device**, **Refresh Favorite Device #1/#2/#3**.
--- a/extensions/toothpick/CHANGELOG.md
+++ b/extensions/toothpick/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Toothpick Changelog
 
+## [Refresh Device and Refresh All Commands] - {PR_MERGE_DATE}
+
+- Added a **Refresh** action in Manage Bluetooth Connections which disconnects and reconnects a device.
+- Added commands: **Refresh Device**, **Refresh Favorite Device #1/#2/#3**.
+- Added command **Refresh All** (`blueutil` only).
+
 ## [Keyboard Shortcuts to Copy] - 2026-01-08
 
 - Added `Keyboard` `Shortcut`s to copy `Action`s
--- a/extensions/toothpick/package.json
+++ b/extensions/toothpick/package.json
@@ -12,7 +12,8 @@
     "rspeicher",
     "roele",
     "AntonNiklasson",
-    "xmok"
+    "xmok",
+    "merchako"
   ],
   "repository": {
     "type": "git",
@@ -97,6 +98,30 @@
       "disabledByDefault": true
     },
     {
+      "name": "refresh-device",
+      "title": "Refresh Device",
+      "subtitle": "Toothpick",
+      "description": "Shortcut for refreshing a specific device connection.",
+      "mode": "no-view",
+      "arguments": [
+        {
+          "name": "nameOrMacAddress",
+          "placeholder": "Name or Mac Address",
+          "type": "text",
+          "required": true
+        }
+      ],
+      "disabledByDefault": true
+    },
+    {
+      "name": "refresh-all",
+      "title": "Refresh All",
+      "subtitle": "Toothpick",
+      "description": "Turns Bluetooth off and on again.",
+      "mode": "no-view",
+      "disabledByDefault": true
+    },
+    {
       "name": "connect-favorite-device-1",
       "title": "Connect Favorite Device #1",
       "description": "Shortcut for connecting to favorite device #1.",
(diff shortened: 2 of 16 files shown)


pernielsentikaer:
Thanks for the update @merchako 

@VladCuciureanu can you give this a spin?

VladCuciureanu:
> Thanks for the update @merchako
> 
> @VladCuciureanu can you give this a spin?

I'm afraid not. Schedule doesn't allow me to work on side-projects.

pernielsentikaer (changes requested):
Hi @merchako 👋

I have some feedback when I checked the code 🙂

- Empty `.gitkeep` file added — extensions/toothpick/.gitkeep is a new empty file with no purpose. Should be removed 🙂

- `refresh-all` is instant off→on with no delay — `blueutil -p 0 && blueutil -p 1` turns BT off then immediately back on. For devices that are "stuck", this may not give enough time for the radio to fully power down. A small `sleep 1` between the two commands would be more reliable: `blueutil -p 0 && sleep 1 && blueutil -p 1`.

- Hardcoded 2-second delay in `refresh-device.ts` — setTimeout(resolve, 2000) is a magic number. Should at minimum be a named constant. Could also consider making it configurable or at least documenting why 2s was chosen.

- `refreshDevice` duplicates connect/disconnect logic — refresh-device.ts calls devicesService.disconnectDevice() and devicesService.connectDevice() directly rather than reusing connectDevice() and disconnectDevice() from the existing handlers. This means it doesn't get the closeOnSuccessfulConnection behavior consistently and has its own message flow. It does handle closeOnSuccessfulConnection at the end, but the disconnect path doesn't respect it.

- `connect-device.ts` side effect: now returns `boolean` — The existing connectDevice handler was changed to return !!result, and disconnectDevice similarly. These return values aren't used by existing callers. The change to showErrorMessage instead of showSuccessMessage on failure is a good bugfix, but the return value addition is dead code outside of the new refresh path (and refresh doesn't even use these handler functions).

- Greptile flag is valid — The "Refresh" action in devices.model.tsx is inserted at index 1 (between Connect/Disconnect and Copy MAC Address). This extension's action list is ordered, and Refresh should probably go after Copy MAC Address, before ...additionalActions.

- `refresh-device.ts` uses raw arrow function export — export default async (props) => {...} instead of a named function like the other command files (export default async function Command()). Minor consistency issue.

- AppleScript backend `refreshBluetooth()` silently returns `false` — No user-facing message explains that Refresh All isn't supported on the AppleScript backend. The refresh-bluetooth.ts handler hardcodes getDevicesService("blueutil") which is good, but if someone calls refreshBluetooth() on the AppleScript service directly through the Manage panel, it silently fails. Not a real issue since refresh-all forces blueutil, but worth noting for interface completeness.


---

merchako:
Thanks for the thorough review! Here's what was addressed:

**Resolved:**

- **`.gitkeep`** — Removed.

- **`sleep 1` delay in refresh-all** — Added. Used `/bin/sleep 1` (full path) instead of `sleep 1` — Raycast's subprocess runs with a restricted `PATH` that doesn't include `/bin`, so `sleep` wasn't found at runtime.

- **Hardcoded 2s delay** — Extracted to `const RECONNECT_DELAY_MS = 2000` in `refresh-device.ts`.

- **`refreshDevice` duplicates connect/disconnect logic** — Refactored to call the `connectDevice` and `disconnectDevice` handler functions directly. As a side effect, `closeMainWindow` is now only called if the reconnect actually succeeds, since it's handled inside `connectDevice` rather than unconditionally at the end of `refreshDevice`.

- **Return values as dead code** — Since `refreshDevice` now calls the handler functions, the `boolean` return values from `connectDevice` and `disconnectDevice` are actually used. No longer dead code.

- **Refresh action position** — Moved after Copy MAC Address, before `...additionalActions`.

- **Raw arrow function export** — Changed to `export default async function Command`.

**Not addressed:**

- **AppleScript `refreshBluetooth()` silently returns false** — Agreed it's not a real issue. For context: if blueutil isn't installed, Refresh All shows an error telling the user to install it via Homebrew. If blueutil is installed but AppleScript is set as the active backend in preferences, Refresh All still works — the handler hardcodes blueutil regardless of the preference setting.

merchako:
Requesting a re-open @pernielsentikaer

merchako:
Anybody able to reopen this for me?  @VladCuciureanu @briankosw @VladCuciureanu @sxn @pernielsentikaer @rspeicher @VladCuciureanu @AntonNiklasson @xmok

pernielsentikaer (changes requested):
Hi @merchako 👋

Thanks for getting back to me, sorry I didn't catch it before now 🙂

Two small things:
- `refresh-favorite-device-1/2/3.ts` use `export default async () =>` instead of named function `Command()` like the other command files. Could you align these for consistency?
- Minor: unused `DevicesService` type import in `refresh-bluetooth.ts` (only `getDevicesService` is needed)

Could you look into above?

---

pernielsentikaer (approved):
Looks good to me, approved 🔥

merchako:
@raycastbot Fixed. Try again. Can I tell you to try again? Probably not.

PR #26148, opened 2026-03-09
Fix: Silent mode no longer activates Obsidian window [Obsidian]

byheaven:
### Problem

When running **Append to Daily Note**, **Append Task**, or **Run Custom Action** with **silent mode enabled**, Obsidian still jumps to the foreground and becomes the active window.

The root cause: Raycast's `open()` triggers macOS-level URL scheme handling (`LSOpenWithURL`), which always activates the target application — regardless of the `&openmode=silent` parameter in the Advanced URI URL. The `silent` parameter only controls Obsidian's internal navigation behavior (whether it opens the note), but has no effect on macOS window activation.

### Fix

Introduce a `src/utils/open.ts` utility (`openUrl`) that routes through macOS `open -g` (the background flag) via Node.js `child_process.execFile` when silent mode is on. This tells macOS to handle the URL without bringing the app to the foreground. When silent mode is off, it falls back to Raycast's `open()` as before.

- `execFile` is used (not `exec`) because it passes arguments as an array, bypassing shell interpretation entirely — no command injection risk.
- A `os.platform() === "darwin"` guard ensures Windows compatibility: non-macOS platforms fall back to `open()` since `open -g` is macOS-only.
- All `onAction` callbacks properly `await` `openUrl` so cleanup operations (`clearCache`, `popToRoot`, `closeMainWindow`) only run after the URL has been opened.

### Bug fix included

The multi-vault selection path in **Append Task** was using `<Action.Open>` without passing `silent` to `Obsidian.getTarget()`. This meant Obsidian would always open without the silent flag when the user had multiple vaults and needed to select one. This is now fixed.

### Changes

| File | Change |
|---|---|
| `src/utils/open.ts` | **New** — `openUrl(url, { background })` utility with platform guard |
| `src/dailyNoteAppendCommand.tsx` | Use `openUrl` with `await` in single-vault and multi-vault paths |
| `src/appendTaskCommand.tsx` | Use `openUrl` with `await` + fix missing `silent` in multi-vault path |
| `src/tools/appendNote.ts` | Use `openUrl` with `background: input.silent ?? true` |
| `src/runActionCommand.tsx` | Use `openUrl` with `background: action.silent` |

## Screencast

_Silent mode is a background operation — no visible UI change is the expected behavior. Before this fix, Obsidian would steal focus every time; after this fix it stays in the background._

## Checklist


pernielsentikaer (commented):
Thanks for the update @byheaven 

Do you want to check this @marcjulianschwarz

marcjulianschwarz:
@byheaven this does not seem to work. I tried all three commands and no matter what I do, Obsidian is still opening and moving to the front.

pernielsentikaer (commented):
Thanks for checking @marcjulianschwarz 🤗

byheaven:
> @byheaven this does not seem to work. I tried all three commands and no matter what I do, Obsidian is still opening and moving to the front.

What are your system and Obsidian versions? I've tested this on macOS 26.3.1 and Obsidian 1.12.5, and it works as expected.

pernielsentikaer (commented):
It works for me too in `26.4 Beta (25E5218f)` with `1.12.4 (Installer 1.9.10)` of Obsidian

@marcjulianschwarz, could you check once again 🙂

pernielsentikaer (commented):
@byheaven, can you fix the merge conflict too? 🙂

byheaven:
@marcjulianschwarz  I'm guessing you're using Windows? As I mentioned in the description, the behavior on Windows hasn't changed because the command `open -g` isn't supported.

pernielsentikaer:
awaiting @marcjulianschwarz 🙂

pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #26181, opened 2026-03-09
Squeeze: Add 'Paste after squeeze' extension preference

maxbeizer:
## Description

Adds an extension-level preference ("Paste after squeeze") that allows users to paste squeezed text directly into the active app instead of only copying to the clipboard.

**Current behavior:** All no-view commands read from clipboard → process → `Clipboard.copy()` back to clipboard. User must then ⌘V manually.

**New behavior (when preference enabled):** Commands use `Clipboard.paste()` instead, which pastes into the frontmost app *and* updates the clipboard — a single-step squeeze-and-paste workflow.

## Changes

- **`package.json`** — Added a top-level `preferences` array with an `autoPaste` checkbox (default: off)
- **`src/utils.ts`** *(new)* — Shared `outputResult()` helper that checks the preference and calls `Clipboard.paste()` or `Clipboard.copy()` accordingly
- **All no-view commands** (`brute-squeeze`, `natural-speech`, `trim-leading-whitespace`, `trim-trailing-whitespace`, `trim-both-whitespace`, `remove-line-breaks`) — Replaced `Clipboard.copy(processed)` with `outputResult(processed)`
- **`advanced-squeeze.tsx`** — Unchanged; it already has its own paste/copy actions in the ActionPanel

## Why an extension preference?

A single toggle avoids duplicating every command. Users who want paste-after-squeeze enable it once and all commands gain the behavior. The default remains copy-only, so existing behavior is unchanged.

Closes #26179


0xdhrv (approved):
Looks good to me, approved ✅

PR #26193, opened 2026-03-10
Add openclaw-channels extension

Kris77z:
## Description

This extension adds a **channel-first OpenClaw workflow** in Raycast.

Why:
- The default chat flow is great for quick prompts, but daily usage in OpenClaw benefits from stable session keys and channel-like context separation.
- This extension is designed to align with OpenClaw session semantics while keeping Raycast fast for daily use.

What’s included:
- **Chat with OpenClaw** (channel-first):
  - Multiple context types (`main`, `dm`, `group`, `channel`, `topic`)
  - Persistent session-key-based continuity
  - `Start New Round` via `/new` on the same session key
  - Clone channel to another profile
- **Profile-aware workflow**:
  - In-chat profile switching
  - Profile management in command UI (create / edit / test / delete / switch)
- **Gateway Status**:
  - Multi-instance status view in one screen
  - Per-profile health/latency and active-profile switch
- **Open Webchat**:
  - Open current profile web UI with session context

Notes:
- Extension id/title are set as `openclaw-channels` / `OpenClaw Channels` to avoid conflict with existing naming.

## Screencast
command overview
chat overview
create profile
manage profile
edit channel
chat view

## Checklist


pernielsentikaer (changes requested):
Hey, thanks for the contribution 🔥

However we already have an `OpenClaw` extension in the store: https://www.raycast.com/leveragedrobot/openclaw by @leveragedrobot — it already covers `Chat`, `Gateway Status`, and `Open Webchat`.  

Rather than publishing a separate extension, it'd be better to contribute your improvements (`multi-profile support`, `session contexts`, `clone-to-profile`, etc.) as a `PR` to the existing extension. That way users don't end up with two competing `OpenClaw` extensions in the store.  

A few things from reviewing the code that would be good to address when you contribute to the existing extension:  

* `chat.tsx` is ~1400 lines with 5+ components, persistence, and business logic all in one file — would be good to split that up into smaller modules

* There's no fetch timeouts — if the gateway is unresponsive the extension just hangs. Adding `AbortSignal.timeout()` especially on the health check would help  

* Same for streaming — if you navigate away mid-stream there's no `AbortController` to cancel the in-flight `SSE` reader
  
* Might want to add a per-conversation message limit since with 80 max conversations and no message cap, `LocalStorage` could grow pretty large over time  

Would you be open to working with the existing extension author on getting these features in there instead?

---

PR #26229, opened 2026-03-11
google-chrome-profiles: Add Google Chrome Canary support

NirRosh:
## Summary

The extension currently only supports Google Chrome Stable. This PR adds support for **Google Chrome Canary** by introducing a browser configuration abstraction that replaces all hardcoded Chrome paths, AppleScript app names, and binary paths.

### Changes

- Added a `BrowserConfig` interface and predefined configs for Chrome Stable and Chrome Canary
- Added a **Browser** dropdown preference so users can select which Chrome variant to use
- Updated `openGoogleChrome` to use the selected browser's app name and binary path in all AppleScript and shell fallback scripts
- Updated profile and bookmark file paths to use the selected browser's data directory
- Updated the "Open in Google Chrome" action label to reflect the selected browser name
- Added an informative toast when "Open Profile with Context" is launched directly without a quicklink context (previously it silently did nothing)

### Files changed

- `src/util/types.ts` -- `BrowserConfig` interface, `BROWSERS` map, `getSelectedBrowser()` helper
- `src/util/util.ts` -- `openGoogleChrome` now accepts a `BrowserConfig` parameter
- `src/index.tsx` -- Threads browser config through all components and actions
- `src/open-profile.tsx` -- Uses browser config, shows feedback on missing context
- `package.json` -- New "Browser" dropdown preference

### DIFF
--- a/extensions/google-chrome-profiles/CHANGELOG.md
+++ b/extensions/google-chrome-profiles/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Google Chrome Profiles Changelog
 
-## [Feature] - {PR_MERGE_DATE}
+## [Feature] - 2026-03-12
 
 - Add Google Chrome Canary support via a new Browser preference dropdown
 - Show informative toast when "Open Profile with Context" is launched without a quicklink context


extensions/google-chrome-profiles/src/util/types.ts
```diff
@@ -80,6 +82,29 @@ export type GoogleChromeBookmarkFile = {
...
+export interface BrowserConfig {
+  readonly name: string;
+  readonly dataPath: string;
+  readonly appName: string;
+  readonly binaryPath: string;
+}
+
+export const BROWSERS: Record<string, BrowserConfig> = {
+  chrome: {
+    name: "Google Chrome",
```
frouo: what do you think if we remove `name` and keep `appName`. For the moment they both have the same value in the two entries.

I suggest we keep `appName`.

extensions/google-chrome-profiles/src/util/util.ts
```diff
@@ -179,21 +180,19 @@ export const openGoogleChrome = async (
...
   const action = target.action;
   const url = action === "openUrl" ? target.url : undefined;
 
-  // Escape all user-controlled input to prevent AppleScript injection
   const escapedProfileName = escapeAppleScriptString(profile.name);
   const escapedUrl = url ? escapeAppleScriptString(url) : undefined;
+  const escapedAppName = escapeAppleScriptString(browser.appName);
 
-  // Use menu bar item 8 for Profiles menu (language-independent position)
-  // Chrome menu bar: 1=Apple, 2=Chrome, 3=File, 4=Edit, 5=View, 6=History, 7=Bookmarks, 8=Profiles, 9=Tab, 10=Window, 11=Help
```
frouo, line 191: I noticed that most comments have been removed. I think we should keep them to give (us and AI) more context, it always helps.

frouo:
Apparte: do you think we could add a "custom" browser entry where Raycast users would set up the "Local State" file path, binary path and app name. Making this extension compatible with any Chromium browser?

extensions/google-chrome-profiles/src/util/types.ts
NirRosh: Good call -- removed `name` and kept `appName` only. Since both values were identical for both browser entries, there was no reason to have a separate field. Fixed in 2c0b7f0.

extensions/google-chrome-profiles/src/util/util.ts
NirRosh: You are right, apologies for that. All the original comments have been restored -- the AppleScript injection warning, the menu bar item layout reference, the inline AppleScript comments, and the fallback explanations. Fixed in 2c0b7f0.

NirRosh:
That is a great idea. A "Custom" entry in the browser dropdown that unlocks text-field preferences for the Local State path, binary path, and app name would make this work with Brave, Vivaldi, Arc, or any Chromium-based browser. The BrowserConfig abstraction we introduced here should make that straightforward to add. Happy to tackle it as a follow-up PR if you think it fits the scope of this extension.

extensions/google-chrome-profiles/src/util/util.ts
```diff
@@ -282,6 +286,6 @@ export const openGoogleChrome = async (
     await willOpen();
     await runAppleScript(fallbackScript);
   } catch (error) {
-    // Handle errors silently
+    console.error("Fallback shell script failed:", error);
```
frouo: Sorry I come back with another review that could help support. 

We have some users saying "the extension is not working", would you mind showing an [alert](https://developers.raycast.com/api-reference/feedback/alert) **with both the "Profiles menu approach" error + this error**, instead of silently fail?

I think an alert best fit so the user has the time to read it, eventually understand why the extension is not working and ultimately report the errors in a issue.

Thank you very much

NirRosh: Sure, Its my pleasure.
When both the Profiles menu AppleScript and the fallback shell script fail, we now show a confirmAlert with the profile name, both error messages, and a hint about the binary path. This should give users enough context to understand the failure and report it. Fixed in 09d9dde.

extensions/google-chrome-profiles/src/index.tsx
```diff
@@ -66,23 +76,19 @@ const ProfileItem = (props: { index: number; profile: Profile }) => {
...
 export default function Command() {
+  const browser = getSelectedBrowser();
   const [localState, setLocalState] = useState<GoogleChromeLocalState>();
   const [error, setError] = useState<Error>();
 
   useEffect(() => {
     async function listProfiles() {
       try {
-        // for google-chrome-profiles-1.png:
-        // 1. comment the code below:
```
frouo, line 76: Also, please restore these comments

extensions/google-chrome-profiles/src/index.tsx
```diff
@@ -66,23 +76,19 @@ const ProfileItem = (props: { index: number; profile: Profile }) => {
...
-        const path = join(homedir(), "Library/Application Support/Google/Chrome/Local State");
+        const path = join(homedir(), browser.dataPath, "Local State");
         const localStateFileBuffer = await readFile(path);
         const localStateFileText = localStateFileBuffer.toString("utf-8");
         setLocalState(JSON.parse(localStateFileText));
-        // 2. uncomment function _createDataSetForScreenshot1() at the bottom of the file
-        // 3. uncomment code below:
-        // setLocalState(_createDataSetForScreenshot1());
```
frouo, line 83: and also restore these comments 🙏🏻

extensions/google-chrome-profiles/src/index.tsx
NirRosh: Restored -- the screenshot data helper comments are back. Fixed in 3d6dba8.

extensions/google-chrome-profiles/src/index.tsx
NirRosh: Restored in the same commit (3d6dba8).

pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #26234, opened 2026-03-11
[Skills] Highlight outdated skills

DaleSeo:
## Description

Previously, outdated skills looked identical to up-to-date ones in the installed skills list. This PR change makes them immediately visible by tinting their hammer icon orange and improves the "Updates Available" section with a count subtitle so users always know how many pending updates exist.

## Screencast

2026-03-07 at 11 35 00

## Checklist

### DIFF
--- a/extensions/skills/CHANGELOG.md
+++ b/extensions/skills/CHANGELOG.md
@@ -1,10 +1,9 @@
 # Changelog
 
-## [Per-Skill Update] - {PR_MERGE_DATE}
+## [Highlight Outdated Skills] - {PR_MERGE_DATE}
 
 - Highlight outdated skills with an orange hammer icon in the installed skills list
-- Show individual "Updates Available" section with count when updates exist
-- Update a single skill directly from its action panel (Cmd+Shift+U)
+- Show "Updates Available" section with count when updates exist
 
 ## [Fix Duplicate Audit Error Toast] - 2026-03-11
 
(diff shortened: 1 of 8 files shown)


keito4:
This looks good overall. One thing I wanted to check, though: are we sure the released `skills@latest` CLI already supports per-skill update?

From what I can tell, the current published flow still documents `update` as update-all, and the released CLI seems to ignore positional args for `update`. If that is right, `npx -y skills@latest update <skillName> -y` would still behave like update-all today, so the UI might imply a level of granularity that is not actually available yet.

Since there is already upstream work for single-skill update, maybe we should either wait for that to be released, or keep this PR focused on outdated indicators + update-all for now. What do you think?

DaleSeo:
@keito4 Good catch! I was told that the skills CLI would ship single-skill update support by the end of this week, but it seems like the PR is still under review. I'll scope this PR down to just the outdated skill highlighting for now and re-add per-skill update in a follow-up once the CLI catches up.

keito4:
Thanks for scoping this down. Focusing this PR on outdated-skill highlighting for now makes sense, and the resulting UX changes look straightforward. LGTM!!

pernielsentikaer:
@keito4, if you actually use the "Approve" feature in GitHub, the bot will set a label and move it to the front of the queue. That's just a tip for next time ;)

pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #26271, opened 2026-03-12
Add clean-agent-text extension

rhukster:
## Checklist


0xdhrv (commented):
Thanks for your contribution @rhukster  🔥

We already have an extension in the Store that deals with Clean Text. Could we consider enhancing the existing extension below instead of creating another one?

- https://www.raycast.com/GenuineCheddar/clean-text by @MoisesCervera

rhukster:
Thanks @0xdhrv! I did look at Clean Text before submitting. I think these two actually solve pretty different problems, even with the name overlap.

Clean Text is aimed at prose hygiene: quotes, whitespace, invisible characters, capitalization, merging paragraphs. Great for stuff pulled from PDFs, OCR, or websites.

Clean Agent Text is aimed at a specific developer annoyance I hit daily. When you copy output from terminal AI agents like Claude Code, Aider, or Codex, you get all the surrounding Unicode box drawing characters (`╭─╮`, `│`, `╰─╯`) and pipe borders along with the content, plus everything is hard wrapped to the terminal width. The extension strips the borders, reflows the content, and detects whether the inside is prose or code so it handles each correctly (collapsing wrapped lines for prose, keeping line breaks and indentation for code).

The two don't really overlap in functionality. Clean Text doesn't handle box art, pipe stripping, or code detection, and Clean Agent Text doesn't touch quotes, capitalization, or invisible characters. The UX is also quite different: Clean Agent Text is a one-shot no-view command (alias `cat`) that just runs instantly with a HUD, no toggles or combos to pick.

Happy to rename it if the "Clean" in the title is the concern. Something like "Detui" or "Strip Agent Result" would make the distinction clearer. But I don't think the functionality naturally fits inside Clean Text.

BTW, I have pushed the greptile fixes already.

0xdhrv (approved):
Looks good to me, approved ✅

PR #26286, opened 2026-03-13
Add blip-raycast extension

astroboii47:
## Summary

Adds a Raycast extension for sending files and folders to Blip on macOS.

## What It Does

- Send the current Finder selection to Blip
- Pick any file or folder in Raycast and send it to Blip
- Prefill the picker from Finder when a file is already selected

## Important Note

Blip does not expose a public macOS file-send API, so this extension uses Blip's Finder `Services` integration.

Because of that, the extension requires **Accessibility** permission for **Raycast** in:

`System Settings > Privacy & Security > Accessibility`

This is required so Raycast can trigger Finder's `Services > Blip…` menu item.

## Screenshot

Raycast command UI with the Accessibility requirement shown before sending to Blip.

Blip Raycast command

## Testing

Tested locally on macOS with:
- `Send to Blip`
- `Send Selected Finder Item to Blip`

Verified that:
- Blip opens correctly
- the selected file is attached to the send flow
- missing Accessibility permission shows setup guidance


0xdhrv (approved):
Looks good to me, approved ✅

PR #26329, opened 2026-03-14
Add raycast-triage extension

vovarevenko:
### Motivation

I noticed that many people (myself included) struggle with capturing quick thoughts during focused work. I was using Quick Reminder which takes several clicks, my colleagues use Notes or even Telegram chats for the same purpose. All of these break the flow.

Triage solves this: type the reminder, press Enter, keep working. No windows, no forms, no context switching. You capture it now, triage it later.

### What it does

Triage is a no-view command for quickly capturing reminders to Apple Reminders without leaving what you're doing.

**Text argument** — the reminder itself. Supports inline priority markers (`!`, `!!`, `!!!`) and automatic URL cleanup (strips protocol and trailing slash).

**Due date argument** (optional) — flexible date input:
- relative: `20m`, `2h`, `1d`, `1w`, `1y`;
- absolute: `20.09`, `20.09.26`, `20.09.2026`;
- with time: `2d 14:30`, `20.09 14:00`.

The target Reminders list is configurable in preferences (default: "Triage"). If the list doesn't exist, it's created automatically.

Uses AppleScript via `runAppleScript()` to interact with Apple Reminders — no third-party services, no network requests.

## Checklist


vovarevenko:
Applied Greptile's suggestions

0xdhrv (commented):
Thanks for your contribution @vovarevenko 🔥

We already have an extension in the Store that deals with Apple Reminders. Could we consider enhancing the existing extension below instead of creating another one?

- https://www.raycast.com/raycast/apple-reminders by @raycast

I think it would be nice to have add this feature to the existing extension.

PR #26371, opened 2026-03-16
Update models-dev extension

carlesandres:
## Description

This PR contains bug fixes, performance improvements, and a UX refinement for the models-dev extension.

### Behavior Change

**Compare AI Models - Simplified to 2-model comparison**

- Previously: Users could select up to 3 models, then manually trigger comparison
- Now: Users select 2 models, and comparison view opens automatically when the second model is selected
- Rationale: Simpler UX, reduces cognitive load, and was necessary to fix the OOM crash (fewer models = less memory)
- The comparison view now shows a clean side-by-side of exactly 2 models

### Bug Fixes

**Critical: JS Heap Out of Memory Crash**

- The "Compare AI Models" command was crashing when comparing models due to view stacking
- Root cause: Using `push(<ComparisonView>)` stacked the Detail view on top of the ~3,800 item List, keeping both mounted
- Fix: Switched to conditional rendering (`showComparison` state) to unmount the List when showing comparison
- Commits: `da13dd1`, `0377ad1`, `d23ac5e`, and related

**Critical: React Version Mismatch Runtime Error**

- All commands were experiencing `Cannot read properties of null (reading 'useRef')` errors
- Root cause: Multiple React instances in dependency tree (@raycast/api used 19.0.0, @raycast/utils used 19.2.4)
- Fix: Added npm `overrides` to force all packages to use React 19.2.4
- Commit: `4eb87ad`

**Search Filtering Issue**

- Search was matching capability keywords (reasoning, vision, etc.) causing confusing results
- User expectation: Search should only match what's visible on screen (model name + provider)
- Fix: Simplified keywords array to only include `[model.providerId, model.providerName, model.family]`
- Commit: `3c44d1b`

### Performance Improvements

- Optimized component re-renders with memoization
- Stabilized event handlers to prevent render cascades
- Removed redundant cache layers

### Testing & Quality

- Added E2E validation harness using AppleScript + Node orchestrator
- Added automated memory validation script
- Both can be run via npm scripts: `npm run validate:e2e` and `npm run validate`

### Verification

- ✅ `npm run build` passes
- ✅ `npm run lint` passes
- ✅ Tested distribution build in Raycast
- ✅ All 5 commands function correctly
- ✅ Memory usage is stable (no crashes after 10+ comparison cycles)
- ✅ Search results match user expectations
- ✅ Extension icon: 512x512px PNG
- ✅ Screenshots: 2000x1250px PNG (6 images)

## Screencast

No screencast provided. The main visible change is the simplified "Compare AI Models" flow (2 models with auto-navigate instead of 3 models with manual trigger). Happy to provide a screencast if reviewers would like to see it in action.

## Checklist

---

### Why npm `overrides` for React version?

- Standard solution for peer dependency conflicts
- Fixes actual runtime bug (not cosmetic)
- Safe change (19.0.0 → 19.2.4 is patch version)
- @raycast/utils explicitly allows >=19.0.0

### Why add test/validation files?

- Prevent regression of OOM crash (critical bug)
- Dev dependencies only (not bundled in distribution)
- Zero impact on extension size or performance

### DIFF
--- a/extensions/models-dev/CHANGELOG.md
+++ b/extensions/models-dev/CHANGELOG.md
@@ -1,6 +1,6 @@
 # models.dev Changelog
 
-## [Bug Fixes] - {PR_MERGE_DATE}
+## [Bug Fixes] - 2026-03-16
 
 ### Changed
 
--- a/extensions/models-dev/CHANGELOG.md
+++ b/extensions/models-dev/CHANGELOG.md
@@ -1,5 +1,18 @@
 # models.dev Changelog
 
+## [Bug Fixes] - {PR_MERGE_DATE}
+
+### Changed
+
+- **Compare AI Models**: Simplified to 2-model comparison with auto-navigate (previously 3 models with manual trigger)
+
+### Fixed
+
+- Fixed JS heap out of memory crash in compare-models command caused by view stacking
+- Fixed React version mismatch causing useRef runtime errors
+- Fixed search filtering to only match visible properties
+- Improved memory management with conditional rendering in model comparison
+
 ## [Initial Version] - 2026-02-05
 
 ### Added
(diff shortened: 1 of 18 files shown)


pernielsentikaer (approved):
Looks good to me, approved 🔥

carlesandres:
> Looks good to me, approved 🔥

Thanks @pernielsentikaer

PR #26387, opened 2026-03-16
Add RayEngineer extension, Jira Time Tracking with Menu Bar

fratoro:
## Description

Raycast extension for Jira time tracking. Tracks time locally and logs worklogs to Jira.

  Commands:
  - Browse Tasks — lists your assigned/recent Jira issues, start/stop timers inline
  - Quick Start — reads a Jira issue key or URL from your clipboard and starts a timer
  - Stop Timer — review tracked time, adjust if needed, log the worklog to Jira
  - Menu Bar Timer — shows the running timer in the menu bar with pause/stop actions, always visible

  Auto-rounds time to 10min blocks. Can auto-transition issues to "In Progress" and auto-assign on timer start. "Done" on timer stop. 

## Screencast

rayengineer 2026-03-16 at 23 56 45
screenshot-20260316-gTdmJFLB@2x

## Checklist


0xdhrv (commented):

We already have an extension in the Store that deals with Tempo Logging. Could we consider enhancing the existing extension below instead of creating another one?

- https://www.raycast.com/darchen_gautier/tempo by @gdarchen

I think you can these features into the existing extension instead of creating a new one.

fratoro:
Hi @0xdhrv  

This extension does not rely on tempo for time logging, instead it uses the built-in time tracking feature of JIRA to log work time on a task/project. As far as I understand, `tempo` does not work without a Tempo account. 

If I'm wrong about that, I'd be happy to incorporate my adjustments into the existing extension. But to me, it looks like these are two different types of time logging.

0xdhrv:
> Hi @0xdhrv
> 
> This extension does not rely on tempo for time logging, instead it uses the built-in time tracking feature of JIRA to log work time on a task/project. As far as I understand, `tempo` does not work without a Tempo account.
> 
> If I'm wrong about that, I'd be happy to incorporate my adjustments into the existing extension. But to me, it looks like these are two different types of time logging.

Tempo is part of the Atlassian suite, so it's linked to Jira, I am not sure if there's a separate Jira time tracking.

PR #26390, opened 2026-03-17
[Coffee] Fix display sleep on battery for macOS 26+

### LINKED ISSUE #25868: [Coffee] ...
### Extension

https://www.raycast.com/mooxl/coffee

### Raycast Version

1.104.7

### OS Version

Macos 26.3

#### Error:
```
Error: Command failed with exit code 1: osascript
40:99: execution error: Not authorised to send Apple events to System Events. (-1743)

P:run-applescript.ts:127:17

---
124:     timedOut,
125:     command: "osascript",
126:     options,
127:     parentError: new Error(),
128:   });
129: }
130: 
---

caffeinateWhile.js:4:2125
```

### Steps To Reproduce

When I run the command Caffinate while it is not loading

### Current Behaviour

It is showing something went wrong

### Expected Behaviour

_No response_

jparkypark:
## Description

On macOS 26+, battery power management can override `caffeinate`'s `-d`/`-i` assertions, causing the display to sleep and the lock screen to engage even when caffeinated. This is because the system sleep timer (`pmset -b sleep`) takes precedence over idle sleep prevention assertions on battery power.

This fix adds a periodic user activity assertion (`caffeinate -u -t 1`) inside the existing `status` background command, which already runs every 15 seconds. The `-u` flag declares user activity and resets the idle timer, which prevents both display sleep and lock screen activation — without requiring elevated privileges or spawning additional processes.

### Changes

- `status.ts`: When caffeinate is running, spawn a detached `caffeinate -u -t 1` to assert user activity and reset the idle timer
- `index.tsx`: Filter out `caffeinate -u` processes from `ps` output so they don't interfere with the time-remaining display
- Simplified redundant `if/else if` into a single condition
- Updated `CHANGELOG.md`

### Why this approach

- **Zero persistent processes** — piggybacks on the existing `status` command (runs every 15s); the `-u -t 1` process exits after 1 second
- **No sudo required** — `caffeinate -u` doesn't need elevated privileges
- **No side effects** — `index.tsx` filters out `-u` processes so toggle, status detection, and time-remaining display are unaffected
- **Harmless on older macOS** — the `-u` assertion is a no-op when the idle timer isn't fighting the user

## Screencast

N/A — the fix is behavioral (prevents display sleep on battery). Verified locally on macOS 26.3.1 on battery power with `displaysleep 2` and `sleep 1` pmset settings.

## Checklist

Fixes #25868

### DIFF
--- a/extensions/coffee/CHANGELOG.md
+++ b/extensions/coffee/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Coffee Changelog
 
-## [Fix] - {PR_MERGE_DATE}
+## [Fix] - 2026-03-17
 
 - Fixed caffeination not preventing display sleep on macOS 26+ when running on battery power.
 
--- a/extensions/coffee/CHANGELOG.md
+++ b/extensions/coffee/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Coffee Changelog
 
+## [Fix] - {PR_MERGE_DATE}
+
+- Fixed caffeination not preventing display sleep on macOS 26+ when running on battery power.
+
 ## [Fix] - 2026-01-27
 
 - Fixed memory leak caused by zombie processes when starting caffeination.
--- a/extensions/coffee/src/index.tsx
+++ b/extensions/coffee/src/index.tsx
@@ -35,7 +35,13 @@ function useCaffeinateInfo(execute: boolean) {
         return { isRunning: false, timeRemaining: null };
       }
 
-      const lines = stdout.split("\n");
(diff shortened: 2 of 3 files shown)


jparkypark:
Both issues addressed in `75675a28`:

1. **`execSync` blocking** — Replaced with `spawn` + `detached: true` + `unref()`. Also added a filter in `index.tsx` to exclude `caffeinate -u` processes from the time-remaining parser.
2. **Changelog date** — Changed to `{PR_MERGE_DATE}`.

pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #26402, opened 2026-03-17
Add Clipboard QR Decoder extension

pchkauu:
## Summary
- add a new Raycast extension that decodes the first QR code from an image in the clipboard
- support noisy QR images with decoding fallbacks and normalize JSON-style quotes before copying
- use a toast-only no-view UX and include store-facing metadata, README, changelog, and icon assets

## Test plan
- [x] `npm run lint`
- [x] `npm run build`
- [x] verify QR payload is copied back to the clipboard
- [x] verify noisy QR sample decodes after fallback processing
- [x] verify failure toasts for missing image or unreadable QR


0xdhrv (commented):
Thanks for your contribution  

- https://www.raycast.com/StevenRCE0/qr-code-scanner by @pchkauu 

You can extend the existing extension to add support for the current extension’s functionality.

PR #26403, opened 2026-03-17
Update kagi-news extension

mickaphd:
## Checklist


mickaphd:
@greptileai Can you check again?

pernielsentikaer (changes requested):
Hi 👋

It looks like there are still some issues to resolve:

- Unreachable `No Chaos Index Data` fallback (confirmed bug) — In `useCategoryFeed.ts`, when `isChaosIndex` is true and `contentData` is present, the hook always returns a non-null `chaosIndex` object (with `score: 0`, `description: ""` as defaults). So the `chaosIndex ? <ListItem> : <EmptyView title="No Chaos Index Data">` check in both `kagi-news.tsx` and `time-travel.tsx` can never reach the empty view branch. If the API omits chaos fields, users would see `Score: 0/100` with blank analysis instead of the intended empty state. Fix: return `chaosIndex: null` when all three API fields are missing/falsy.

- Duplicated favorites action JSX — The same favorites toggle `<Action>` block (with `isFavorite` check, `Icon.Star`/`Icon.StarDisabled`, `toggleFavorite`) is copy-pasted 4 times in `kagi-news.tsx` (chaos index, events section x2, articles section). This could be extracted into a small component or variable. Not blocking but it's a lot of identical code.

- Chaos Index unconditionally added in `time-travel` — `time-travel.tsx` always appends the `Chaos Index` category, but older batches may not have chaos data. Unlike `hasOnThisDay` which is conditionally gated, chaos is always shown. For old batches without chaos, users would see `Score: 0/100` (per issue #1 above). If the API can indicate chaos data presence, it should be gated similarly.

---

pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #26434, opened 2026-03-18
[g-cloud] Add configuration management, logging improvements, and stability hardening

ojowwalker77:
## Description

Absorbs the functionality from #24645 (gcloud-config-manager) into the existing g-cloud extension, as discussed with @pernielsentikaer — enhancing rather than adding a separate extension.

  New **Settings & Configuration** command for managing named gcloud configurations (list, switch, create, duplicate, delete). Extension settings moved from the main hub into this command.

  Logging gets a **time range filter**, **open in IDE** action (with configurable app picker preference), and new **Default Region** / **Default IDE** / **Default Terminal** preferences.

  Includes a stability pass: cache eviction, exec timeouts, safe JSON parsing, friendly error messages, and several bug fixes.
  
## Screencast

Configuration : 

Screenshot 2026-03-18 at 12 13 55

Open on IDE :

Screenshot 2026-03-18 at 12 14 26

Set default IDE and Terminal options: 

Screenshot 2026-03-18 at 12 15 11

## Checklist


pernielsentikaer (changes requested):
Hi 👋 

- `searchText` escaping is incomplete — Line 9615: searchText.replace(/[\\"]/g, "\\{context}") escapes backslashes and quotes, but GCP log filter syntax has other special characters. A user typing ( or ) could break the filter query or inadvertently alter its structure. Should sanitize for GCP filter syntax specifically.

- `configurations.tsx` uses `Preferences` type correctly — but LogsView.tsx uses getPreferenceValues<Preferences>().preferredIDE at module scope (line 9535), which gets evaluated once at import time. If preferences change, the extension needs a restart. Acceptable but worth noting.

---

ojowwalker77:
Hi @pernielsentikaer 

both addressed 😉

  - searchText escaping: Switched from =~ (regex) to : (substring contains), so special characters like (, ), | are treated as literals.
  - preferredIDE at module scope: Moved inside the component body so it's read fresh on each render.

ojowwalker77:
Previous fix was a oversight, now should be good to go!  : )

pernielsentikaer (changes requested):
Hey, nice work on this — love that you absorbed the `config manager` into the existing `extension` instead of a separate one.

Two things before we merge:

- When creating or duplicating a `config`, if the name already exists it silently overwrites the existing `config`'s properties. Could you add a check against the existing `configs` list in both `CreateConfigForm` and `DuplicateConfigForm` to show an error if the name is taken? The `configs` list is already available in both components, just needs a validation check before submit.

- I noticed `Jest` and `ts-jest` are added as `devDependencies` but the test files are `gitignored` (`src/__tests__/` in `.gitignore`). Could you either include the tests in the `PR` or remove the test infra from `package.json`? Having test tooling without visible tests is a bit confusing.  
Everything else looks good — security is solid, no existing commands changed, and the stability improvements are welcome.

---

pernielsentikaer (changes requested):
Found a few more things that might be worth looking into?

- Broken cancellation pattern in `index.tsx` (bug) — The cancelled boolean is passed as a function parameter to checkGcloudInstallation(cancelled), checkAuthStatus(cancelled), and loadServiceCounts(cancelled). Since JS passes primitives by value, the inner functions get a copy of false. When the cleanup sets the outer cancelled = true, the inner functions never see it. All if (cancelled) return; guards after await calls are dead code. Unmounted components will still call setState. Fix: remove the parameter and close over the variable directly.

- `preferredTerminal` preference declared but unused — Defined in package.json but no code references it anywhere in the diff. Dead configuration.

- Server-side log search removed (functional regression) — Users can now only filter the fetched 100 log entries client-side. Previously they could search the full log corpus via gcloud filters. The changelog doesn't mention this. Trade-off for security is reasonable, but should be documented.

- Pre-existing shell injection in `gcloud.ts` — The quotePath helper only wraps paths with spaces in double quotes. Doesn't escape $, backticks, ;, etc. Since gcloudPath comes from user preferences, this is a theoretical injection vector. New code in ConfigurationsService.ts correctly uses execFile — the old code should be migrated too. Not blocking since it's pre-existing.

---

pernielsentikaer (changes requested):
A few more things I stumbled upon:

- `@raycast/api` ^1.55.2 and `@raycast/utils` ^1.9.0 — critically outdated. Both must be bumped to ^1.100.0+ and ^2.x respectively. Same applies to all devDependencies (`@raycast/eslint-config` ^1.0.6, `eslint` ^8, `typescript` ^5.2.2).  

- `.prettierrc` missing — no `.prettierrc` file found on the branch. Required with exactly { "printWidth": 120, "singleQuote": false }.  

- `shellSplit` in `gcloud.ts` — the `executeGcloudCommand` function takes a raw command string and splits it with a custom `shellSplit()` helper before passing to `execFile`. This partially defeats the purpose of using `execFile` — if any argument contains spaces or special chars that aren't double-quoted by the caller, they'll be mis-parsed. A safer API would accept `string[]` args directly and skip `shellSplit` entirely.  

Minor / non-blocking:

- Temp file not cleaned up in `LogsView` when `open()` throws (Greptile's P2 finding — valid but low risk since `tmpdir()` is OS-managed)

- Missing `fetchLogs` in `useEffect` dependency array in `LogsView`

`configurations.tsx` at the `src/` root — this appears to be a stub/re-export; confirm it's intentional

Do you want to address above or do you want me to help?

---

pernielsentikaer (changes requested):
Hi 👋

Could you look into below issues?

- Missing ```.prettierrc``` — no Raycast standard prettier config was added

- Module-scope ```getPreferenceValues()``` calls — ```preferredIDE``` was moved inside the component (good), but ```defaultRegion``` is still at module scope in 4 files (```CreateFunctionForm```, ```CloudRunView```, ```CreateConfigForm```, ```StorageBucketView```) and ```CONFIGURED_GCLOUD_PATH``` at module scope in 2 files (```configurations.tsx```, ```index.tsx```). These will read stale values if preferences change mid-session.

Minor nits:
- ```didMount``` ref workaround in ```LogsView``` for suppressing initial ```onSearchTextChange``` is fragile but functional
- Background refresh has ```.catch(() => {})``` silently swallowing errors (acceptable but not ideal)

---

ojowwalker77:
Hi @pernielsentikaer, thanks for the re-review!

1. Module-scope preference reads — moved all 6 inside the component bodies: defaultRegion in CreateFunctionForm, CloudRunView (CreateServiceForm), CreateConfigForm, StorageBucketView, and configuredGcloudPath in configurations.tsx and index.tsx. They'll now read fresh on each render.
2. .prettierrc — this one might be a false alarm on my end: extensions/g-cloud/.prettierrc is present on the branch with the exact expected contents ({ "printWidth": 120, "singleQuote": false }). It predates this PR (added in #18093) so there's no diff for it here — could you double-check your side? Happy to re-add it if you're seeing something different.

 ray lint, ray build, and prettier --check all pass locally.

pernielsentikaer:
So, is this ready for another review? If it is, please mark it as ready for review 🙂

pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #26444, opened 2026-03-18
Add approval check for merging pull requests

pernielsentikaer:


.github/workflows/pull_request_merge.yml
```diff
@@ -1,5 +1,9 @@
 name: Pull Request - Merge
 
+permissions:
```
grzegorzkrukowski: This should not be required as we are using `secrets.PERSONAL_ACCESS_TOKEN` anyways and the `permissions` as here only affects the default token generated for each workflow. 
So I think we can just remove it.

.github/workflows/pull_request_merge.yml
```diff
@@ -12,7 +16,67 @@ on:
         required: true
 
 jobs:
+  require_admin_or_write_approval:
+    runs-on: ubuntu-latest
+    name: Require admin/write approval
+    env:
+      GITHUB_TOKEN: ${{ secrets.PERSONAL_ACCESS_TOKEN }}
+    outputs:
+      approved_by: ${{ steps.check.outputs.approved_by }}
```
grzegorzkrukowski: What's the idea for this output? It's never used in the "merge" step - not sure if it was suppose to be used.
It can stay like that if we want to use it the future.

grzegorzkrukowski (changes requested):
I left some comments inline, but one thing I am a bit thinking about is the Slack messaging.
With the current implementation it won't be triggered if `require_admin_or_write_approval` fails as it's part of the `merge` step which won't be executed if the `require_admin_or_write_approval` fails due to `exit 1` approach.

Not sure if that's just missed or intented to not spam on Slack, but for my taste, we should move Slack to a separate job that `needs: [require_admin_or_write_approval, merge]` and runs-on ubuntu and generate a proper error message based on which task has failed for better information about this fact.

Unless we want to silently skip things like that.

pernielsentikaer (commented):
Can you give it another spin 🙂

grzegorzkrukowski (approved):
Looking good now, sorry for the delay in review

Issue #20456, opened 2025-07-15
[Coffee] Raycast slows down over time


aaronjensen:
### Extension

https://www.raycast.com/mooxl/coffee

### Raycast Version

1.100.3

### macOS Version

15.5

### Description

I reported to Raycast that Raycast was getting slower over time, specifically typing in the search and most noticeably when backspacing the last character from Raycast. They suggested uninstalling Coffee and since doing that I have not been able to reproduce the slow down. It may be related to the menu bar item, which I was using.

### Steps To Reproduce

Install Coffee. Enable the menu bar item. Use Raycast for several days to a week and notice it get slower with time.

sammyolo:
Experiencing the same issue, I don't know if you want comments like this, but mine was slow as well (everything not just typing also window switching) and found this issue, uninstalled coffee. First impressions is significant performance increase.

raulelba:
I have also experienced this issue. Raycast's memory usage in my computed was increasing exponentially over time, eventually reaching nearly 4GB, which slowed down my computer. After reviewing the extension diagnostics and consulting with support, it seems that this extension may have been misbehaving and causing a memory leak in Raycast, so I also uninstalled it.

However, I cannot confirm that Coffee is the only cause for Raycast's memory leak I've been experiencing, as I still observe an increase in memory usage even after uninstalling the extension, although at a much lower rate than before.

pernielsentikaer:
You might be able to see something in `Extension Diagnostics` if anything is acting up. I think the coffee extension makes a single call once a minute to test whether it's caffeinated and shouldn't be "heavy"

garrucho:
I was suffering from a problem where macOS wouldn't run any new processes, apps, browser tabs or the terminal. On the latter, these errors were presented while opening a session:

```
zsh: fork failed: resource temporarily unavailable
```

```
[forkpty: Resource temporarily unavailable]
[Could not create a new process and open a pseudo-tty.]
```


This happened after at least 5 days of uptime, where the computer was sleeping on the last 2 days (maybe waking up from time time, as it's a MacOS behavior).

Influeced by [this old post](https://stackoverflow.com/questions/48986937/cant-fork-process-resource-temporarily-unavailable-issue) on StackOverflow, I've found the following:

```
$ ps -e | awk '{print $4" "$5" "$6}' | sort | uniq -c | sort -n

(...)
  11 /usr/sbin/cfprefsd agent 
[...]

Issue #9410, opened 2023-11-24
[Coffee] The Caffeinate/Decaffeinate command creates zombie processes


faultables:
### Extension

https://www.raycast.com/mooxl/coffee

### Description

I use this extension very often (big thanks for making it!) and just noticed that every time I run `caffeinate` and `decaffeinate`, it creates a zombie process with "Raycast Helper (Extensions)" as the parent process.

![CleanShot 2023-11-25 at 00 50 47@2x](https://github.com/raycast/extensions/assets/6561394/fa0c5f65-913b-42a8-8958-0900724dea7e)

I don't know exactly who is responsible for this issue (the extension developer? the extension API itself?) and the workaround is of course by killing the "Raycast Helper (Extensions)" process which is most likely harmless.

### Steps To Reproduce

1. Install the [Coffee](https://www.raycast.com/mooxl/coffee) extension
2. Launch Raycast, select "Caffeinate"

### Current Behaviour

The coffee command is executed, works as intended (prevent mac from sleeping) and the `<defunct>` process will appear (`ps -ef | grep defunct`)

### Expected Behaviour

No defunct processes are created

pernielsentikaer:
Do you have a clue what the might be @erics118

erics118:
I have no clue why the process is kept hanging. I tried a few things to attempt to fix it, including adding a `&` or a `&; disown` to the end of the `(start|end)Caffeinate`, which didn't seem to change anything.

faultables:
After playing around for a while if we use `execSync` instead of async `exec`, the problem seems to go away?

![CleanShot 2023-11-26 at 14 28 34](https://github.com/raycast/extensions/assets/6561394/17462037-7199-4e80-86cc-72c91874c112)

The patches:

```diff --git a/extensions/coffee/src/utils.ts b/extensions/coffee/src/utils.ts
index 7018c7f..394aca1 100644
--- a/extensions/coffee/src/utils.ts
+++ b/extensions/coffee/src/utils.ts
@@ -1,5 +1,5 @@
 import { getPreferenceValues, launchCommand, LaunchType, showHUD } from "@raycast/api";
-import { exec } from "node:child_process";
+import { execSync } from "node:child_process";
 
 function preventArguments(args?: string | undefined) {
   const preferences = getPreferenceValues<Preferences>();
@@ -42,7 +42,7 @@ export async function stopCaffeinate(updateMenubar = true, hudMessage?: string)
     await showHUD(hudMessage);
   }
 
[...]

PR #26464, opened 2026-03-19
[Coffee] Fix zombie process leak in status command

jparkypark:
## Description

Hardens the `status` background command's idle-timer reset (`caffeinate -u -t 1`) against zombie process creation in the Raycast extension daemon.

Multiple users [reported](https://www.reddit.com/r/raycastapp/comments/1rx6kzx/raycast_creates_defunct_processes_continually/) zombie processes accumulating under `Raycast Helper (Extensions)` after #26390 was merged. Credit to u/thepante for tracing the root cause to that PR.

### Investigation

- The current code on `main` uses `spawn()` + `child.on("exit", () => {})` + `child.unref()` (without `detached: true`). In standalone Node.js testing — including with forced GC — this pattern does **not** produce zombies.
- The original #26390 diff included `detached: true`, which was subsequently removed. Users who received the extension update before this change may still be running the version with `detached: true`, which is the likely cause of the zombie accumulation.
- We were unable to reproduce the zombie flood with the current `main` code, but the `spawn` + `unref` pattern remains fragile — its safety depends on Node.js/libuv internals and the Raycast runtime's event loop behavior.

### Fix

Replace `spawn()` + `unref()` with `execFile()` + callback:

```ts
// Before:
const child = spawn("/usr/bin/caffeinate", ["-u", "-t", "1"], { stdio: "ignore" });
child.on("exit", () => {});
child.unref();

// After:
execFile("/usr/bin/caffeinate", ["-u", "-t", "1"], () => {});
```

`execFile()` with a callback maintains an internal reference to the child process and guarantees `waitpid()` is called when it exits. This eliminates any possibility of zombie creation regardless of runtime behavior.

### Why this approach

- **Guaranteed reaping** — `execFile` callback-based lifecycle ensures `waitpid()` always runs
- **Non-blocking** — doesn't stall the background command
- **Same functionality** — still asserts user activity via `caffeinate -u -t 1`
- **Simpler** — one line vs five

Related: #9410, #20456, #26420 (same pattern in System Monitor)

## Checklist

- [x] I ran `npm run build` and tested this distribution build

Fixes #26463

### DIFF
--- a/extensions/coffee/src/status.ts
+++ b/extensions/coffee/src/status.ts
@@ -1,5 +1,5 @@
 import { LocalStorage, updateCommandMetadata } from "@raycast/api";
-import { spawn } from "node:child_process";
+import { execFile } from "node:child_process";
 import { Schedule, startCaffeinate, getSchedule, stopCaffeinate, isCaffeinateRunning } from "./utils";
 
 async function handleScheduledCaffeinate(schedule: Schedule): Promise<boolean> {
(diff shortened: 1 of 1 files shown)


jparkypark:
Closing this PR after further investigation.

The zombie accumulation reported by Reddit users was likely caused by the earlier version of #26390 that included `detached: true` in the `spawn()` call. That has already been corrected on `main` (removed `detached: true`, added `child.on("exit", () => {})`).

We tested the current `main` code extensively — including with forced GC via `--expose-gc` and rapid spawn cycles — and could not reproduce any zombie accumulation. The current pattern properly reaps child processes.

While `execFile()` is a marginally simpler pattern, it's not fixing an active bug. Closing to avoid noise.

PR #26469, opened 2026-03-19
[Mole] Add residual file cleanup to Uninstall command

jlrochin:
Fixes https://github.com/raycast/extensions/issues/26467

## Summary

- Added deep residual file scanning when uninstalling apps (searches `~/Library`, `/Library`, containers, and vendor directories)
- Matches residual files by bundle ID, app name, display name, and executable name
- Shows a detailed confirmation dialog listing all found residuals with sizes before removal
- Falls back to admin privileges via osascript for protected system files
- Hides removed apps from the list after successful uninstall

## Code quality improvements

- Migrated all error handling from manual `Toast.Style.Failure` to `showFailureToast` from `@raycast/utils`
- Replaced manual Preferences type with auto-generated `ExtensionPreferences` from `raycast-env.d.ts`
- Updated `uninstall` command description to reflect new residual scanning functionality
- Removed unused `exec` import
- Fixed Prettier formatting

## Type of change

- [x] Bug fix
- [x] New feature

## Test plan

- [ ] Open "Uninstall App" command and search for an app
- [ ] Press Enter to uninstall — verify residual files are listed in the confirmation dialog
- [ ] Confirm uninstall — verify app + residuals are moved to Trash
- [ ] Verify app disappears from the list after uninstall
- [ ] Test with an app that has no residuals — verify "No residual files found" message
- [ ] Test canceling the confirmation — verify nothing is removed

### DIFF
--- a/extensions/mole/CHANGELOG.md
+++ b/extensions/mole/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Mole Changelog
 
-## [Uninstall Residual Cleanup] - {PR_MERGE_DATE}
+## [Uninstall Residual Cleanup] - 2026-03-20
 
 - Added deep residual file scanning when uninstalling apps (searches ~/Library, /Library, containers, and vendor directories)
 - Shows a confirmation dialog listing all found residuals with sizes before removal
--- a/extensions/mole/CHANGELOG.md
+++ b/extensions/mole/CHANGELOG.md
@@ -8,7 +8,7 @@
 - Migrated error handling to `showFailureToast` from `@raycast/utils`
 - Replaced manual Preferences type with auto-generated `ExtensionPreferences`
 
-## [1.0.0] - {PR_MERGE_DATE}
+## [1.0.0] - 2026-03-18
 
 - Added System Status command with real-time health monitoring
 - Added Clean System command with streaming scan progress
--- a/extensions/mole/CHANGELOG.md
+++ b/extensions/mole/CHANGELOG.md
@@ -1,6 +1,14 @@
 # Mole Changelog
 
-## [1.0.0] - 2026-03-18
+## [Uninstall Residual Cleanup] - {PR_MERGE_DATE}
+
+- Added deep residual file scanning when uninstalling apps (searches ~/Library, /Library, containers, and vendor directories)
+- Shows a confirmation dialog listing all found residuals with sizes before removal
+- Hides removed apps from the list after successful uninstall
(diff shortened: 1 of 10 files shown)


extensions/mole/CHANGELOG.md
```diff
@@ -1,6 +1,14 @@
...
-## [1.0.0] - 2026-03-18
+## [Uninstall Residual Cleanup] - {PR_MERGE_DATE}
+
+- Added deep residual file scanning when uninstalling apps (searches ~/Library, /Library, containers, and vendor directories)
+- Shows a confirmation dialog listing all found residuals with sizes before removal
+- Hides removed apps from the list after successful uninstall
+- Migrated error handling to `showFailureToast` from `@raycast/utils`
+- Replaced manual Preferences type with auto-generated `ExtensionPreferences`
+
+## [1.0.0] - {PR_MERGE_DATE}
```
pernielsentikaer: This should not be changed

pernielsentikaer (changes requested):
Hi 👋

There is a few things we should look into before merging this PR

* False positive matching — The substring match on normalized names is too aggressive for a file-deletion feature. At minimum, require exact match (not substring) for terms under ~8 chars. `Notes`, `Pages`, `Numbers` would otherwise match unrelated system files.

* Binary plist reading — `readFileSync(…, "utf-8")` on binary plists is unreliable. Should use `defaults read` or `plutil -convert xml1` like the rest of the code already does for `readPlistKey`. One-line fix.

* `/usr/local/bin` should be removed from scan paths — Trashing CLI binaries that happen to share a name with a GUI app is too risky. Homebrew-managed tools, standalone CLIs, etc. could get caught.

---

jlrochin:
Hi! Thanks for the review. All points have been addressed:

- **False positive matching** — Raised the substring match threshold from 5 to 8 characters. Short app names like `Notes`, `Pages`, `Numbers` now require an exact match only.
- **Binary plist reading** — Replaced `readFileSync(…, "utf-8")` with `readPlistKey` (which uses `defaults read`) to properly handle binary plists in container metadata. Also removed the now-unused `readFileSync` import.
- **`/usr/local/bin` removed** — Dropped from scan paths entirely to avoid trashing Homebrew-managed CLI binaries.
- **CHANGELOG** — Restored the original `[1.0.0] - 2026-03-18` entry, only added the new version entry above it.

Ready for another look when you get a chance!

pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #26471, opened 2026-03-19
Music: Performance optimization for large libraries

espenbye:
## Description

Performance improvements for the Music extension, especially for users with large libraries (10K+ tracks):

- **Play Library Album** now opens instantly and searches on demand, instead of loading the entire library upfront — fixing out-of-memory crashes for large libraries
- **Play Library Track** search is faster thanks to optimized AppleScript data fetching
- **Start Playlist** and **Add to Playlist** now use `useCachedPromise` with client-side filtering — reopening is instant after the first load, and switching the dropdown filter no longer re-fetches data
- **Favorite Track** confirms within 3 seconds instead of up to 10 seconds (reduced polling interval and timeout)
- **Menu Bar Player** scroll animation uses fewer re-renders (400ms vs 200ms interval)
- Fixed `ScriptError` type guard typo that prevented it from ever matching
- Exported `PlaylistKind` enum for type-safe reuse across modules

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [x] Performance improvement

## Checklist

### DIFF
--- a/extensions/music/src/play-library-album.tsx
+++ b/extensions/music/src/play-library-album.tsx
@@ -1,54 +1,32 @@
 import { Action, ActionPanel, closeMainWindow, Icon, List, showToast, Toast, useNavigation } from "@raycast/api";
-import { flow, pipe } from "fp-ts/lib/function";
-import * as O from "fp-ts/Option";
+import { pipe } from "fp-ts/lib/function";
 import * as S from "fp-ts/string";
 import * as T from "fp-ts/Task";
 import * as TE from "fp-ts/TaskEither";
-import { useState } from "react";
+import { useCallback, useEffect, useState } from "react";
 
 import { Album, Track } from "./util/models";
-import { fromEmptyOrNullable } from "./util/option";
-import { parseResult } from "./util/parser";
 import * as music from "./util/scripts";
 import { handleTaskEitherError } from "./util/utils";
-import { usePromise } from "@raycast/utils";
 
 const EMPTY_TEXT = " "; // Visually empty but non-empty to prevent jumping around
 
 export default function PlayLibraryAlbum() {
-  const [albums, setAlbums] = useState<readonly Album[] | null>(null);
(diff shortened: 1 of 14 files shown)


extensions/music/src/util/models.ts
```diff
@@ -44,7 +41,7 @@ export interface ScriptError extends Error {
 }
 
 export const ScriptError = {
-  is: (error: Error): error is ScriptError => "shortMessaage" in error,
+  is: (error: Error): error is ScriptError => "shortMessage" in error,
```
rawnly, line 44: nice catch!

extensions/music/src/play-library-album.tsx
```diff
@@ -12,22 +13,36 @@ import { handleTaskEitherError } from "./util/utils";
...
+    () =>
+      pipe(
+        music.albums.getAll,
+        TE.matchW(
+          () => {
+            showToast(Toast.Style.Failure, "Could not get albums");
+            return [] as ReadonlyArray<Album>;
+          },
+          (albums) => albums,
+        ),
```
rawnly: `matchW` it's useless here also `W` on fp-ts stands for `widening` you should use it only when you're mapping data

```ts
        TE.mapLeft(
          () => {
            showToast(Toast.Style.Failure, "Could not get albums");
            return [] as ReadonlyArray<Album>;
          },
        ),
```

byheaven:
> * **Favorite Track** confirms within 3 seconds instead of up to 10 seconds (reduced polling interval and timeout)

In my actual use, even when I mark a song as favorite directly inside the music app, there is still a delay exceeding an average of more than 3 seconds. Of course, I think this may be because I enabled the music app setting “Add songs to library when marking them as favorite.” That is also why I set the timeout to 10 seconds.

0xdhrv (commented):
Hey @espenbye 👋 

Can you please look into the merge conflicts?

PR #26494, opened 2026-03-20
Add quickly-add-to-reminders extension

tcyeee:
## Description

Quickly add tasks to Apple Reminders using prefix autocomplete — type `!` for priority, `@` for due date, `/` for list, and `#` for tags, all without opening the Reminders app.

### Prefix syntax

Prefixes can appear in any order and are space-separated. `!`, `@`, `/` prefixes are stripped before the remainder becomes the reminder title. `#tag` tokens are kept in the title as plain text.

| Prefix | Description | Example |
|--------|-------------|---------|
| `!` / `!!` / `!!!` | Priority — low / medium / high | `!! Buy groceries` |
| `@[date/time]` | Due date (bare `@` defaults to today) | `@tomorrow Fix bug` |
| `/[list]` | Target Reminders list (defaults to default list) | `/Work Refactor auth` |
| `#[tag]` | Tag — kept in title as plain text; supports autocomplete from history | `#work Submit report` |

## Screencast

https://github.com/user-attachments/assets/903b5aba-e593-4233-b3e9-b0f3a2cc746b

## Checklist


0xdhrv (commented):
Thanks for your contribution  

- https://www.raycast.com/raycast/apple-reminders

tcyeee:
Thanks for the reminder and for pointing out the duplication guideline.

This extension was originally created because the existing Apple Reminders extension couldn’t fully cover my workflow, but I agree it’s better to improve the existing one instead of adding a separate extension. I’ll extract the unique parts of this extension and open a PR to add them to the ‎`apple-reminders` extension instead.

PR #26502, opened 2026-03-20
[System Monitor] Eliminate zombie process accumulation in menubar command

### LINKED ISSUE #26480: [System Monitor] Menubar interval spawns zombie processes, exhausting macOS process limit
### Extension

https://www.raycast.com/hossammourad/raycast-system-monitor

### Raycast Version

1.104.10

### OS Version

MacOs Tahoe 26.3.1

## Bug

Since commit 17c452b (2026-03-16), the menubar command gained `"interval": "10s"` in `package.json`. This causes the extension to spawn child processes (`system_profiler`, `nettop`, `ioreg`) every 10 seconds without properly reaping them, resulting in thousands of zombie (defunct) processes accumulating over time.

## Impact

After a few hours of uptime, the macOS process limit (~2666 for my system) is completely exhausted by defunct processes parented to `Raycast Helper (Extensions)`. This makes it impossible to open new terminal sessions:
```
[forkpty: Resource temporarily unavailable]
[Could not create a new process and open a pseudo-tty.]
```

## Root Cause

The `"interval": "10s"` added in the recent commit triggers repeated shell command execution. The parent process (`Raycast Helper (Extensions)`) is not calling `wait()` on the child processes, leaving them as zombies indefinitely.

## Workaround

Disable the Menubar System Monitor command in Raycast Extensions settings.

## Steps to Reproduce

1. Install System Monitor extension
2. Enable Menubar System Monitor
3. Leave running for several hours
4. Run: `ps aux | awk '$8=="Z"' | wc -l`
5. Observe thousands of zombie processes

### Current Behaviour

After a few hours with Menubar System Monitor enabled, thousands of zombie 
(defunct) processes accumulate parented to Raycast Helper (Extensions). 
The macOS process limit is exhausted, making it impossible to open new 
terminal sessions or spawn any new process. Error: "forkpty: Resource 
temporarily unavailable / Could not create a new process and open a pseudo-tty."

Confirmed via: ps aux | awk '$8=="Z"' | wc -l → 2300+ defunct processes
All parented to Raycast Helper (Extensions) PID.

### Expected Behaviour

Child processes spawned by the menubar interval (system_profiler, nettop, 
ioreg) should be properly reaped after execution. Zombie processes should 
not accumulate over time regardless of how long the extension runs.

unirt:
## Summary

Fully eliminate zombie process accumulation in the menubar command by replacing all `exec()` calls with `execFile()`, removing libraries that internally spawn shell processes, and rearchitecting the command to align with Raycast's menu-bar lifecycle.

## Background

PR #26466 reduced zombie accumulation by ~70% with revalidation guards and longer polling intervals, but zombies continued to leak at ~1 every 15 seconds. This PR addresses the remaining root causes for a complete fix.

Related: #26489 (proposed reverting `interval` entirely — this PR preserves background refresh while fixing the underlying issues)

## Root causes

Three sources of zombie processes remained after #26466:

1. **`exec()` spawns `/bin/sh` for every command** — Each `exec()` call creates a shell process in addition to the target binary. With ~12 commands per cycle, this doubled the child process count. `execFile()` spawns binaries directly, eliminating the shell overhead.

2. **`systeminformation` library uses `exec()` internally** — `si.osInfo()` and `si.fsSize()` were called every cycle. The library contains 147 `exec()` calls across its source. Replaced with direct `sw_vers` and `df -kP` calls via `execFile()`.

3. **`system_profiler` fire-and-forget** — The background fetch for battery condition/capacity was not `await`ed, so the child process was never reaped. Replaced with Raycast Cache API — `system_profiler` runs once (with `await`), results are cached to disk.

## Architecture change

The previous implementation used `useInterval()` for in-process polling, but this conflicts with Raycast's menu-bar command lifecycle:

> Menu-bar commands are not long-lived processes. Raycast loads them into memory on demand, executes their code, waits for `isLoading` to become `false`, and then unloads them.
> — [Raycast docs: Menu Bar Commands](https://developers.raycast.com/api-reference/menu-bar-commands#lifecycle)

`useInterval()` running during background `interval` launches prevented `isLoading` from settling, causing 9-second timeouts. The new design:

- **Background launch** (`interval: "10s"`): Fetch all data in a single `Promise.all`, render, unload. No `useInterval`.
- **User click** (menu open): Command stays in memory while the menu is open. `useInterval` polls every 2s for live updates, gated by `environment.launchType === LaunchType.UserInitiated`.
- **Cache**: Previous data is persisted via Raycast `Cache` API so the menubar title renders instantly on the next interval restart (no flicker).

## Changes

| File | Change |
|------|--------|
| `src/utils.ts` | Replace `exec()`-based `execp` with `execFile()`-based `execf`. Add optional `maxBuffer` param. |
| `src/Cpu/CpuUtils.ts` | `execp` → `execf` |
| `src/Memory/MemoryUtils.ts` | `execp` → `execf`, parallelize with `Promise.all`, parse `vm_stat` in JS (removes `awk` pipe) |
| `src/Network/NetworkUtils.ts` | `execp` → `execf` |
| `src/Power/PowerUtils.ts` | `execp` → `execf`, parallelize `ioreg`+`pmset`, cache `system_profiler` via Raycast Cache API, increase `maxBuffer` for `pmset -g log` |
| `src/SystemInfo/SystemUtils.ts` | Remove `systeminformation` dependency, use `df -kP` and `sw_vers` directly |
| `src/menubar-system-monitor.tsx` | Single `usePromise` + `Promise.all` for all data, `Cache` API for persistence, `useInterval` only during user-initiated launches, replace `os-utils` with `os.cpus()` |
| `CHANGELOG.md` | Add entry |

## Test plan

- [x] Enable **Menu Bar System Monitor** command
- [x] Run `watch -n5 'ps -eo ppid,stat | awk "\$2 ~ /^Z/" | wc -l'`
- [x] Verify zombie count stays at 0 over 2+ minutes ✅
- [x] Verify menubar stats update every ~10 seconds (background)
- [x] Open menubar dropdown → verify stats update every ~2 seconds (live)
- [x] Verify pinned stats don't flicker between updates
- [x] Verify no `timed out (9s max)` errors in `npm run dev` console
- [x] Open **System Monitor** view command → verify all tabs work (CPU, Memory, Power, Network, System Info)
- [x] `npm run build` passes cleanly

Fixes #26480

### DIFF
--- a/extensions/system-monitor/CHANGELOG.md
+++ b/extensions/system-monitor/CHANGELOG.md
@@ -1,5 +1,16 @@
 # System Monitor Changelog
 
+## [Eliminate Zombie Processes] - {PR_MERGE_DATE}
+
+- Replace all `exec()` calls with `execFile()` to avoid spawning shell processes
+- Remove `systeminformation` and `os-utils` dependencies (both spawn child processes internally)
+- Rearchitect menubar command to align with Raycast's menu-bar lifecycle (load → render → unload)
+- Cache menubar data to disk via Raycast Cache API to prevent flicker between interval restarts
+- Cache `system_profiler` results for battery condition/capacity (rarely changes, slow to fetch)
+- Poll for live updates only while the menu is open (2s interval via `useInterval`)
+- Parallelize system data fetches with `Promise.all` for faster command execution
+- Fix `pmset -g log` buffer overflow in Power Monitor view
+
 ## [Fix Zombie Process Accumulation] - 2026-03-20
 
 - Add revalidation guards to prevent overlapping child process spawns in the menubar command
--- a/extensions/system-monitor/CHANGELOG.md
+++ b/extensions/system-monitor/CHANGELOG.md
@@ -1,6 +1,6 @@
 # System Monitor Changelog
 
-## [Eliminate Zombie Processes] - {PR_MERGE_DATE}
+## [Eliminate Zombie Processes] - 2026-03-20
 
 - Replace all `exec()` calls with `execFile()` to avoid spawning shell processes
 - Remove `systeminformation` and `os-utils` dependencies (both spawn child processes internally)
--- a/extensions/system-monitor/src/Cpu/CpuUtils.ts
+++ b/extensions/system-monitor/src/Cpu/CpuUtils.ts
@@ -1,4 +1,4 @@
-import { execp } from "../utils";
+import { execf } from "../utils";
 
 const UNITS = {
   year: 24 * 60 * 60 * 365,
@@ -10,7 +10,7 @@ const UNITS = {
 };
 
 export const getTopCpuProcess = async (count: number): Promise<string[][]> => {
-  const output = await execp("/bin/ps -Aceo pcpu,comm -r");
+  const output = await execf("/bin/ps", ["-Aceo", "pcpu,comm", "-r"]);
   const processList: string[] = output
     .trim()
     .split("\n")
--- a/extensions/system-monitor/src/Memory/MemoryUtils.ts
+++ b/extensions/system-monitor/src/Memory/MemoryUtils.ts
@@ -24,14 +24,15 @@ export const getTopRamProcess = async (): Promise<string[][]> => {
 };
 
 export const getMemoryUsage = async (): Promise<MemoryInterface> => {
-  const [pHwPagesize, pMemTotal, pVmPagePageableInternalCount, pVmPagePurgeableCount, vmStatOutput] =
-    await Promise.all([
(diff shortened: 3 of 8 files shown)


pernielsentikaer:
@unirt I don't have any write access, could you lint the code 🙂

pernielsentikaer:
Also since we can't touch the PR, then you need to write todays date instead of {PR_MERGE_DATE}

pernielsentikaer (changes requested):
Could you look into above @unirt 

---

unirt:
Fixed — ran `ray lint --fix` and replaced `{PR_MERGE_DATE}` with `2026-03-20`. Thanks for the quick review!

pernielsentikaer (approved):

PR #26515, opened 2026-03-20
fix(year-in-progress): prevent LocalStorage overwrite during initial …

saivall:
**Fix: User preferences (pins, menubar) reset on extension restart**

The save useEffect fires with default state before the async load from LocalStorage completes, overwriting user preferences (pinned items, menubar selections) on every extension 
restart (lock/unlock, sleep/wake).

Adds an isLoading guard to skip persisting until stored state has loaded.


pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #26523, opened 2026-03-21
fix(coffee): keep caffeinate child referenced until exit

saschabuehrle:
## Bug
https://github.com/raycast/extensions/issues/26512 — the Coffee status command spawns `caffeinate -u -t 1`, but `child.unref()` can drop the handle before process exit, which can leave unreaped zombie processes.

## Fix
Removed `child.unref()` in `extensions/coffee/src/status.ts` so the child process stays referenced long enough for Node to observe process exit and reap it.

## Testing
- Verified the Coffee command still compiles with the change (single-line lifecycle update only).
- Tried to run extension lint (`pnpm -C extensions/coffee lint`), but local tooling requires the Ray CLI (`ray`) which is not available in this environment.

Happy to address any feedback.

Greetings, saschabuehrle

### DIFF
--- a/extensions/coffee/CHANGELOG.md
+++ b/extensions/coffee/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Coffee Changelog
 
+## [Fix] - {PR_MERGE_DATE}
+
+- Prevent zombie caffeinate child processes by keeping the spawned process referenced until exit.
 ## [Fix] - 2026-03-19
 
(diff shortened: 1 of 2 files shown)


saschabuehrle:
Added the CHANGELOG.md entry with the `{PR_MERGE_DATE}` placeholder. Good catch — thanks for the review.

Greetings, saschabuehrle

0xdhrv (approved):
Looks good to me, approved ✅

PR #26539, opened 2026-03-21
Add Display Music extension

guymgn:
Shows Apple Music album artwork in the menu bar with track info, playback controls, and copy link feature.


0xdhrv (commented):

We already have an extension in the Store that deals with Apple Music Menu Bar. Could we consider enhancing the existing extension below instead of creating another one?

- https://www.raycast.com/fedevitaledev/music by @rawnly

You can enhance the existing extension to add support for the album art in the menu bar.

PR #26552, opened 2026-03-22
[skills] Fix fnm base directory resolution on macOS

raulgg:
## Description

This change improves `fnm` path resolution when building the environment for running the Skills CLI through `npx`.
 
On macOS, `fnm` may be installed under `~/Library/Application Support/fnm`, which the previous logic did not detect. That could leave `FNM_DIR` pointing to the wrong location.

The fix updates `resolveFnmBaseDir()` to look for `fnm` in the same locations and preference order used by the [official install script](https://github.com/Schniz/fnm/blob/master/.ci/install.sh).

## Checklist

- [X] I checked that assets used in the `README` are located outside the metadata folder if they were not generated with our [metadata tool](https://developers.raycast.com/basics/prepare-an-extension-for-store#how-to-use-it)

### DIFF
--- a/extensions/skills/src/utils/node-path-resolver.ts
+++ b/extensions/skills/src/utils/node-path-resolver.ts
@@ -4,6 +4,7 @@ import { join } from "node:path";
 import semver from "semver";
 
 const isWindows = process.platform === "win32";
+const isMacOS = process.platform === "darwin";
 
 let cachedPaths: string | null = null;
 let cachedPathsPromise: Promise<string> | null = null;
@@ -65,14 +66,23 @@ const scanVersionedNodePaths = async (
 export const resolveFnmBaseDir = async (home = process.env.HOME): Promise<string | null> => {
   if (!home) return null;
 
-  const legacyFnmPath = join(home, ".fnm");
-  if (await pathExists(legacyFnmPath)) {
-    // Older fnm installs keep everything under ~/.fnm instead of XDG data directories
-    return legacyFnmPath;
+  const fnmBaseDirCandidates = process.env.FNM_DIR ? [process.env.FNM_DIR] : [];
+  fnmBaseDirCandidates.push(join(home, ".fnm"));
+  if (process.env.XDG_DATA_HOME) {
+    fnmBaseDirCandidates.push(join(process.env.XDG_DATA_HOME, "fnm"));
+  }
+  if (isMacOS) {
+    fnmBaseDirCandidates.push(join(home, "Library", "Application Support", "fnm"));
+  }
+  fnmBaseDirCandidates.push(join(home, ".local", "share", "fnm"));
+
+  for (const dir of fnmBaseDirCandidates) {
+    if (await pathExists(dir)) {
+      return dir;
+    }
   }
 
-  const xdgDataHome = process.env.XDG_DATA_HOME || join(home, ".local", "share");
-  return join(xdgDataHome, "fnm");
+  return null;
 };
 
 export const resolveVersionManagerPaths = async (): Promise<string[]> => {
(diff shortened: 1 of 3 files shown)


extensions/skills/src/utils/node-path-resolver.ts
```diff
@@ -65,14 +66,23 @@ const scanVersionedNodePaths = async (
 export const resolveFnmBaseDir = async (home = process.env.HOME): Promise<string | null> => {
   if (!home) return null;
 
-  const legacyFnmPath = join(home, ".fnm");
-  if (await pathExists(legacyFnmPath)) {
-    // Older fnm installs keep everything under ~/.fnm instead of XDG data directories
-    return legacyFnmPath;
+  const fnmBaseDirCandidates = process.env.FNM_DIR ? [process.env.FNM_DIR] : [];
+  fnmBaseDirCandidates.push(join(home, ".fnm"));
```
DaleSeo: If `resolveFnmBaseDir` were ever called from a context where `FNM_DIR` is set, the current logic would add it as the first candidate and then potentially return a different directory. I mean the one that exists on disk but isn't the one the user explicitly configured. This could silently override the user's intent.

```suggestion
  const fnmBaseDirCandidates: string[] = [join(home, ".fnm")];
```

DaleSeo (commented):
Thanks for this fix, @raulgg! I have a comment.

pernielsentikaer (changes requested):
Thanks for the review 🙂

@raulgg can you look into the suggestion?

---

keito4 (approved):
Nit: should we align this with fnm’s current runtime resolution rather than the install script? Otherwise we could end up picking a different base dir than fnm itself when multiple locations exist.

pernielsentikaer (commented):
CC @DaleSeo and @keito4 🙂

pernielsentikaer (commented):
@DaleSeo do you have any final thoughts on this?

DaleSeo:
@pernielsentikaer My comment is a minor edge case. It only matters if `FNM_DIR` is set and points somewhere different from `~/.fnm`. I'm fine with merging as-is. We can tighten it later if needed.

extensions/skills/src/utils/node-path-resolver.ts
raulgg: Thank you for the review @DaleSeo! And I'm sorry for taking a long time to check your comments, as I've been busy.
I'm not sure if I understand your concern here regarding overriding the user's intent. I added `FNM_DIR` here to honor it if present when setting the `PATH` in [`getExecOption`](https://github.com/raycast/extensions/blob/d0f455d1585733a0b2da9354083a8274fc331c0c/extensions/skills/src/utils/exec-options.ts#L10). Actually I think that is not possible for this env variable to be set during the Raycast execution, so given this I could just event ignore the `FNM_DIR` here as you suggest.

However, I don't see the code paths from which we would override the explicit user config; I may be missing something. If the user configures the custom path in the extension preferences, then we would be using that explicitly in [runSkillsCli](https://github.com/raycast/extensions/blob/d0f455d1585733a0b2da9354083a8274fc331c0c/extensions/skills/src/utils/skills-cli.ts#L33).

DaleSeo: @raulgg You're right that `FNM_DIR` likely won't be set in Raycast's context, which is exactly why including it as a candidate doesn't add value. My concern was more about defensive correctness: if it were set, the function iterates candidates and returns the first that exists on disk, which might not match `FNM_DIR`. But since you agree it's unlikely to matter here, just dropping it keeps things simple. Thanks for looking into it!

DaleSeo (approved):

extensions/skills/src/utils/node-path-resolver.ts
raulgg: Thank you for your clarification! I completely agree with you on it.

pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #26554, opened 2026-03-22
Ghostty: Fix "Open with Ghostty" to open selected Finder item

lederniermagicien:
## Description

The "Open with Ghostty" command currently ignores the selected file in Finder. It only detects a single selected **folder**, and falls back to the Finder window's current directory for everything else (files, multiple selection, no selection). This means hovering over or selecting a file and invoking the command opens the wrong directory.

### Root cause

The existing implementation uses a pure AppleScript approach (`openGhosttyWindowAtFinderLocation`) that:
1. Only matches when exactly 1 item is selected **and** it is a folder
2. Falls back to `Finder's insertion location` (the window directory) otherwise
3. Uses keystroke injection (`keystroke "cd " & ...`) which is fragile and requires accessibility permissions

### Fix

This rewrite mirrors the approach used by the [Warp extension's `open-selected-finder-item`](https://github.com/raycast/extensions/blob/main/extensions/warp/src/open-selected-finder-item.tsx):

- Uses `getSelectedFinderItems()` from the Raycast API for accurate selection detection
- Resolves files to their parent directory via `path.dirname()`
- Uses Ghostty's **native AppleScript API** ([`new surface configuration`](https://ghostty.org/docs/features/applescript) + `initial working directory`) instead of keystroke injection
- Opens a new tab in existing Ghostty, or a new window if Ghostty is not running
- Supports multiple selections and Path Finder
- Falls back to the current Finder window directory only when nothing is selected

### Changes

| File | Change |
|------|--------|
| `src/open-with-ghostty.ts` | Rewritten to use Raycast API + Ghostty native AppleScript |
| `src/utils/scripts.ts` | Removed dead `openGhosttyWindowAtFinderLocation` script |
| `CHANGELOG.md` | Added entry |

### Testing

- Select a **file** in Finder → opens Ghostty at the file's parent directory
- Select a **folder** in Finder → opens Ghostty at that folder
- Select **multiple files** in different directories → opens one tab per unique directory
- No selection, Finder frontmost → opens Ghostty at the Finder window's current directory
- Ghostty not running → launches Ghostty with a new window
- Ghostty already running → opens a new tab

### DIFF
--- a/extensions/ghostty/.eslintrc.json
+++ b/extensions/ghostty/.eslintrc.json
@@ -1,4 +1,0 @@
-{
-  "root": true,
-  "extends": ["@raycast"]
-}
--- a/extensions/ghostty/CHANGELOG.md
+++ b/extensions/ghostty/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Ghostty Changelog
 
-## [Fix] - {PR_MERGE_DATE}
+## [Fix] - 2026-03-25
 
 - Fix "Open with Ghostty" to open the selected Finder item (file or folder), not just the current window directory
 - Support files (opens parent directory), folders, multiple selections, and Path Finder
--- a/extensions/ghostty/CHANGELOG.md
+++ b/extensions/ghostty/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Ghostty Changelog
 
+## [Fix] - {PR_MERGE_DATE}
+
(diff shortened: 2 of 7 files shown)


pernielsentikaer (changes requested):
Hi 👋 

- Could you try to update `@raycast/utils` and import `runAppleScript` from that instead, it shouild not be nessecary to have `run-applescript` as dependency at all 🙂

---

pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #26562, opened 2026-03-22
[HomeAssistant] Fix "Show Entity IDs" preference being completely non-functional

rspeicher:
## Description

Two bugs combined to make the "Show Entity IDs" checkbox do nothing:

1. `shouldDisplayEntityID()` read `preferences.instance` (the HA URL textfield, always truthy) instead of `preferences.showEntityId` (the actual checkbox).

2. The subtitle logic was inverted: `if (shouldDisplayEntityID())` suppressed the entity ID instead of showing it.

These cancelled out in practice — the always-true result meant entity IDs were always hidden, which matched the default unchecked state. But toggling the checkbox had no effect either way.

> [!NOTE]
> Might conflict with https://github.com/raycast/extensions/pull/26561. This one should probably be merged first.

## Checklist

### DIFF
--- a/extensions/homeassistant/src/components/state/list.tsx
+++ b/extensions/homeassistant/src/components/state/list.tsx
@@ -89,7 +89,7 @@ export function StateListItem(props: { state: State }): React.ReactElement {
         icon = ha.urlJoin(ep);
       }
     }
-    if (shouldDisplayEntityID()) {
+    if (!shouldDisplayEntityID()) {
       return extra;
     }
     if (extra) {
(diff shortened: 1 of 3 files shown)


extensions/homeassistant/src/components/state/list.tsx
rspeicher, line 93: `extra` in this case is only a media player's content information. So if we _don't_ want to display entity IDs, we just return that early.

If we _do_ want to display entity IDs, the two get merged below.

It's kind of a weird way to structure this, but I'm just making the minimal change.

pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #26563, opened 2026-03-22
Fix potential search error in the Linkwarden Raycast extension

trucke:
### Fix potential search error in the Linkwarden Raycast extension

When no collection is selected, the search command was sending an empty collectionId= to the API — which expects an integer, so it responds with a 400 Bad Request. On top of that, search text wasn't being URL-encoded, so typing characters like &, #, or = would break the query string and cause similar errors.

Changes:
- Only include collectionId param when a collection is actually selected
- Use URLSearchParams to properly encode all query parameters
- Surface fetch errors to the user via toast (previously silently swallowed)
- Strip trailing slashes from the instance URL to avoid double-slash API paths

Note: GET /api/v1/links is marked as deprecated in the Linkwarden API docs (https://docs.linkwarden.app/api/retrieve-a-list-of-links) in favor of GET /api/v1/search.

## Checklist

### DIFF
--- a/extensions/linkwarden/CHANGELOG.md
+++ b/extensions/linkwarden/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Linkwarden Changelog
 
-## [Fix Potential Search Errors] - {PR_MERGE_DATE}
+## [Fix Potential Search Errors] - 2026-03-23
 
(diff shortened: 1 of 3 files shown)


pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #26577, opened 2026-03-23
Update coffee extension

IKatsuba:
## Checklist

### DIFF
--- a/extensions/coffee/CHANGELOG.md
+++ b/extensions/coffee/CHANGELOG.md
@@ -2,9 +2,9 @@
 
 ## [Fix] - {PR_MERGE_DATE}
 
-- Fixed zombie process accumulation by double-forking the caffeinate process via shell backgrounding, reparenting it to launchd (PID 1) which automatically reaps exited children.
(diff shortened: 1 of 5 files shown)


pernielsentikaer (changes requested):
Hey @IKatsuba, thanks for looking into this! The zombie fix is a real issue worth solving.

However, switching from `spawn` to `execSync` with string interpolation opens up a shell injection surface through `additionalArgs`. Even if current callers pass safe values, it's a fragile pattern.  

I think the actual fix is simpler — the previous attempt with `spawn` + `detached: true` was on the right track, it was just missing `.unref()`. Could you try this approach instead?

```javascript
const args = ["-u", ...generateArgs(additionalArgs).split(/\s+/).filter(Boolean)];
const child = spawn("/usr/bin/caffeinate", args, { detached: true, stdio: "ignore" });
child.unref();
```  

This detaches and reparents to `launchd` the same way, without needing a shell. The rest of your changes (removing the periodic spawner, adding `-u` to the main process) are good — just the execution method that needs changing.

---

pernielsentikaer (commented):
Let's try to merge and see how it goes 🙂

pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #26590, opened 2026-03-23
ext/plexamp Critical library selection fix

kendaniels:
## Summary
- **Fix music library detection**: Try all available server connections concurrently (`Promise.any`) instead of only the preferred one, so unreachable local connections no longer silently fail
- **Hide servers without music**: Filter out servers that have no music libraries (or where all connections fail) from the selection list
- **Fix search hang**: Replace sequential connection fallback with concurrent racing so the UI no longer hangs for minutes when connections timeout
- **Progressive loading**: Stream library results into the UI as each server resolves instead of waiting for all servers, with local servers prioritized first
- **Remove UI noise**: Remove LAN badge and "Shared By" accessories from library selection items
- **Fix status screen flash**: Prevent the "Sign in to Plex" screen from flashing on the status command while the initial async load is in progress

## Testing
- Manual testing of library selection flow with local and remote servers
- Type checking passes (`tsc --noEmit`)
- ESLint passes on all changed files

## Post-Deploy Monitoring & Validation
No additional operational monitoring required: client-side Raycast extension with no server-side components.

### DIFF
--- a/extensions/plexamp/CHANGELOG.md
+++ b/extensions/plexamp/CHANGELOG.md
@@ -1,3 +1,13 @@
+## [Library Selection Fixes] - {PR_MERGE_DATE}
+
+- Fixed music library detection failing silently when the preferred server connection is unreachable by trying all connections concurrently.
+- Fixed the library selection screen hanging for minutes when remote server connections timeout.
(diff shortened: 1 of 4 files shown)


pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #26623, opened 2026-03-24
Plexamp: Save verified connection URL during library selection

kendaniels:
## Summary
- Fixed "All promises were rejected" error on Browse Library, Search Library, and Status commands by saving the connection URL that actually loaded the user's chosen library, instead of an untested ranked-preferred URL that may be unreachable.
- Fixed the Now Playing Menubar not showing track information by using the server address from Plexamp's timeline response.

## What changed
During setup, Plex servers expose multiple connection URLs (local IP, remote domain, relay). Previously `saveSelectedServer` saved the ranked-preferred URL (local > remote > relay) regardless of which one actually worked. Now `getMusicSectionsForServer` returns the URI of the connection that successfully loaded libraries, and that URI is passed through to `saveSelectedServer` when the user selects their library.

## Test plan
- [ ] Sign in and select a music library from library selection
- [ ] Verify Browse Library and Search Library work without "All promises were rejected" errors
- [ ] Verify Now Playing Menubar displays current track info
- [ ] Verify choosing a different library via Change Library command updates the saved connection

### DIFF
--- a/extensions/plexamp/CHANGELOG.md
+++ b/extensions/plexamp/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Plexamp CHANGELOG
 
-## [Server Connection Fixes] - 2026-03-24
+## [Server Connection Fixes] - {PR_MERGE_DATE}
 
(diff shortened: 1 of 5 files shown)


0xdhrv (approved):
Looks good to me, approved ✅

PR #26632, opened 2026-03-25
password-store: Fix insert hanging due to execFile stdin

dportalesr:
## Summary

- Fix `pass insert` command hanging indefinitely when called with stdin input via `execFile`'s `input` option
- Use `spawn` with explicit `stdin.write()`/`stdin.end()` when input is provided, which properly signals EOF to the `pass` bash script's GPG encryption pipeline

## Problem

When inserting a new password, `runPassCmd` calls `execFile("pass", ["insert", "-m", ...], { input })`. The `execFile` `input` option does not properly close stdin in a way that the `pass → gpg` pipeline recognizes as EOF. The child process never exits, causing the `await` to hang forever — the success toast never shows and `pop()` (navigate back) never runs.

Passwords are written to disk successfully, but the UI remains stuck on the insert form.

This is the same class of bug identified in #25809 for `pass rm` (where the fix was to use `--force` to skip stdin). The `insert -m` case has no equivalent flag, so it requires proper stdin handling.

## Fix

When `input` is provided, use `spawn` instead of `execFile` with manual stdin handling:
- `child.stdin.write(input)` sends the data
- `child.stdin.end()` explicitly closes the stream, sending a proper EOF
- Listen for the `close` event to resolve the promise

When no `input` is provided, the original `execFile` path is unchanged.

## Test plan

- [ ] Insert a new password via the extension — form should close and show success toast
- [ ] Insert a password with metadata — both password and metadata should be saved
- [ ] Existing commands (decrypt, generate, rename, delete) continue working

### DIFF
--- a/extensions/password-store/CHANGELOG.md
+++ b/extensions/password-store/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Password Store Changelog
 
+## [Fix] - 2026-03-25
+
+- Fixed insert password hanging indefinitely due to stdin not closing properly
+
 ## [Improvements] - 2026-03-23
 
 - Able to insert password via `pass insert`
--- a/extensions/password-store/CHANGELOG.md
+++ b/extensions/password-store/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Password Store Changelog
 
-## [Fix] - 2026-03-25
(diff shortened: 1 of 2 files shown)


pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #26669, opened 2026-03-26
Update linkinize extension

NaderIkladious:
## Description

Fixes broken or missing bookmark icons in the Linkinize extension.

Previously, bookmarks with an invalid or unavailable favicon URL could display a broken image. This update:

- Validates bookmark favicon URLs in the background
- Displays a generated avatar using the bookmark’s first letter when its favicon is missing or invalid
- Uses a consistent avatar color based on the bookmark name
- Caches invalid favicon URLs to avoid repeating failed requests on subsequent loads
- Bumps the extension version to 0.1.2 and updates the changelog

## Checklist

### DIFF
--- a/extensions/linkinize/CHANGELOG.md
+++ b/extensions/linkinize/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Linkinize Changelog
 
+## [Favicon Fallback] - {PR_MERGE_DATE}
+
+- Added initial letter avatars as fallback for bookmarks with broken or missing favicons
+- Favicon validation results are cached for faster subsequent loads
+
 ## [Update Contributors] - 2026-01-20
 
 - Updated extension metadata (author, contributors, owner)
--- a/extensions/linkinize/CHANGELOG.md
+++ b/extensions/linkinize/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Linkinize Changelog
 
-## [Favicon Fallback] - {PR_MERGE_DATE}
+## [Favicon Fallback] - 2026-07-31
 
 - Added initial letter avatars as fallback for bookmarks with broken or missing favicons
 - Favicon validation results are cached for faster subsequent loads
--- a/extensions/linkinize/package-lock.json
+++ b/extensions/linkinize/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "linkinize",
-  "version": "0.1.1",
+  "version": "0.1.2",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "linkinize",
-      "version": "0.1.1",
+      "version": "0.1.2",
       "license": "MIT",
       "dependencies": {
         "@raycast/api": "^1.104.1",
--- a/extensions/linkinize/package.json
+++ b/extensions/linkinize/package.json
@@ -1,7 +1,7 @@
 {
   "$schema": "https://www.raycast.com/schemas/extension.json",
   "name": "linkinize",
-  "version": "0.1.1",
+  "version": "0.1.2",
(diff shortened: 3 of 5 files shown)


0xdhrv (commented):
.

NaderIkladious:
Hi! Reopening this PR as the changes are ready for review.

I’ve updated the description to better explain the main improvement: bookmarks with missing or broken favicons now display a generated initial-based avatar instead of a broken image. Invalid favicon results are also cached to avoid repeated failed requests.

The branch has been updated, and the extension has been built and tested in Raycast. Thanks!

0xdhrv:
Hey @NaderIkladious 👋 

Can you please fix the merge conflicts?

NaderIkladious:
Hey @0xdhrv 👋 Merge conflicts are fixed — the branch is rebased onto latest `main` and now shows as mergeable.

The only conflict was in `package-lock.json`. I took the upstream lockfile (keeping your picomatch update) and regenerated it, so the lock now differs from `main` by just the `0.1.1` → `0.1.2` version bump with no dependency drift.

I also addressed all four issues from the Greptile review:

1. **Transient failures became permanent** — favicon validation now distinguishes three outcomes. Only a definitive **4xx** is cached as invalid; timeouts, aborts, network errors and 5xx fall back for the current session and are retried on the next launch, so a recovered endpoint is no longer suppressed forever.
2. **Avatar text broke SVG markup** — the initial is now XML-escaped (`&`, `<`, `>`, `"`, `'`). I also switched to code-point splitting so a name starting with an emoji renders the whole emoji instead of a broken surrogate half.
3. **Unused dependency** — `@raycast/utils` removed from both `package.json` and the lockfile.
4. **Unbounded validation** — requests are now deduplicated to distinct favicon URLs (bookmarks frequently share a host), capped at 8 concurrent, bounded with `maxContentLength`, and wired to an `AbortController` that fires on unmount.

One extra fix while in there: some hosts answer a missing favicon with an HTML error page at status 200, which the previous code treated as valid. Validation now also requires an `image/*` content-type.

`npm ci`, `npm run build` and `npm run lint` all pass, and the distribution build was tested in Raycast. Thanks for the review!

NaderIkladious:
Good catch on the 4xx classification — `408` and `429` are indeed transient and my previous change would still have cached them permanently.

Fixed in `8c44e40`: `408 Request Timeout`, `425 Too Early` and `429 Too Many Requests` are now excluded from the definitive set and classified as `unavailable`, so they fall back for the current session and are retried on the next launch. Only a 4xx that genuinely means "this favicon does not exist" (404, 403, 410, …) is persisted to the invalid cache.

NaderIkladious:
Fair point, and rather than keep narrowing the status allowlist (409, then 401/403, and so on) I've addressed the underlying class of problem in `1e1c5f3`: **the invalid-favicon cache now expires.**

The cache is a `url -> timestamp` map with a 7-day TTL instead of a plain list. Expired entries are dropped on read and pruned from storage on the next write. That means no response, whatever its status, can permanently suppress a favicon — a URL that returned 409/401/403 because of temporary conditions is simply re-validated on a later launch and recovers on its own.

The status classification is still there, but its role is now much narrower: it only decides whether a failure is worth *not re-checking for a while* (a 4xx, likely genuinely missing) versus retried on the very next launch (timeouts, rate limits, aborts, network errors and 5xx, which say nothing about the favicon itself).

Verified the cache behaviour across 12 cases including expiry, storage pruning, empty writes, corrupt JSON, and the legacy array format — all of which fall back to re-validating rather than suppressing. `npm run build` and `npm run lint` pass.

0xdhrv (approved):
Looks good to me, approved ✅

PR #26701, opened 2026-03-27
Bitwarden: Update CLI to v2026.2.0 to fix "Invalid session token" error

### LINKED ISSUE #26143: [Bitwarden Vault] update issue
### Extension

### Raycast Version

0.48.0.0

### OS Version

Windows 11 25h2

### Description

After the latest Bitwarden Vault extension update for Raycast on Windows, the extension stopped functioning. The update caused the UI to remain stuck on the “Restart Raycast” prompt, which persisted after restarting Raycast, rebooting the system, and reinstalling the extension while preserving settings.

Removing the extension without keeping settings led to a different failure: the built‑in Bitwarden CLI auto‑installer began throwing errors. I was able to restore functionality only by manually installing the Bitwarden CLI through PowerShell. During this process, I found that Chocolatey was no longer installed, and the previously installed Bitwarden CLI had also been removed. It’s unclear whether the update process removed these components, but the behavior suggests a dependency issue triggered by the extension update.

I cannot confirm reproducibility, but I’m reporting this in case other users encounter the same dependency removal or CLI installation failures.

### Steps To Reproduce

Install the Bitwarden Vault extension in Raycast for Windows.
Run any command that prompts for the Bitwarden Client ID and Client Secret.
Press Ctrl + Enter to trigger the automatic Bitwarden CLI installation.
Wait for the installation to complete to view the error.

### Current Behaviour

Needs manual Bitwarden CLI installation.

### Expected Behaviour

Shouldn't require troubleshooting after updating the extension.

edobry:
## Description

Update bundled Bitwarden CLI from v2025.11.0 to v2026.2.0 to fix "Invalid session token" error when unlocking the vault.

**Root Cause:** CLI v2025.12.0 introduced automatic KDF (Key Derivation Function) upgrades, enforcing a minimum of 600,000 PBKDF2 iterations per OWASP guidelines. Users whose KDF settings were upgraded server-side (triggered by logging in through any Bitwarden client) can no longer unlock their vault with the old CLI — `bw unlock --raw` returns empty stdout instead of a session token.

Also fixes a minor bug where vault status was saved as "unlocked" after API key login, when it should be "locked" (the vault still requires `bw unlock` after `bw login --apikey`).

Fixes #26143

## Checklist

- [x] I updated the CHANGELOG.md

> **Note:** This PR was authored with assistance from [Claude Code](https://claude.com/claude-code) (`claude-opus-4-6`). The diagnosis, code changes, and SHA256 hash verification were performed collaboratively.


jomifepe (approved):

0xdhrv (approved):
Looks good to me, approved ✅

PR #26759, opened 2026-03-30
[Prompts.chat] Fix crash from API field rename (content -> contentPreview)

tmchow:
## Description

Fixes the prompts.chat extension crashing when searching or loading prompts.

## Root Cause

The prompts.chat API renamed the `content` field to `contentPreview`. In `cache.ts`, `downloadAllPrompts()` maps `item.content` which is now `undefined`. When `searchPrompts()` calls `.toLowerCase()` on this undefined value, the extension crashes.

Verified by calling the API directly: all 1595 prompts have `contentPreview`, none have `content`.

## Fix

In `cache.ts`:
- `downloadAllPrompts()`: map `item.contentPreview ?? item.content ?? ""` (handles both new and cached data)
- `searchPrompts()`: null-safe `(p.content ?? "").toLowerCase()` in the filter

This contribution was developed with AI assistance (Claude Code).

Fixes #26607

### DIFF
--- a/extensions/prompts-chat/CHANGELOG.md
+++ b/extensions/prompts-chat/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## [Fix] - 2026-03-30
+
+- Fix crash from upstream API field rename (content -> contentPreview)
 ## [Initial Version] - 2026-01-26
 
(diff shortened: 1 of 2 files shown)


pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #26795, opened 2026-03-31
[Skills] Remove broken trending command

DaleSeo:
## Description

A user at a recent event mentioned that the skills displayed in the Trending Skills command are quite different from those on the skills.sh website. After looking into it, I found that this command uses `/api/search?q=skill`, which returns skills that include the word "skill" in their name. Also, this endpoint is designed for searching and doesn't sort skills by popularity, which is misleading.

I reached out to the Vercel team for a workaround, but they said they don’t plan to support endpoints for trending. I think this command is useful if it works as intended, but if it doesn’t, it just confuses users. The Search Skills command already has all the features that the trending command offers. So, this PR removes the command entirely to avoid any confusion for users.

## Screencast

Skills.sh|Extension
-|-
2026-03-31 at 09 21 22|2026-03-31 at 09 20 07

## Checklist


pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #26801, opened 2026-03-31
Fix Dust extension: SSE streaming and workspace picker

chriscamicas:
## Summary
- **Fix SSE streaming hang**: The Dust client's `streamAgentAnswerEvents()` uses `ReadableStream.getReader()` which is incompatible with undici's fetch polyfill in Raycast's Node.js environment, causing the stream to hang indefinitely. Replaced with direct undici fetch + Node.js async iterable (`for await...of`) which properly delivers SSE chunks.
- **Fix AbortController lifecycle**: The previous `useEffect` had `conversationId` in its dependency array, causing the effect cleanup to abort the stream mid-flight when `setConversationId` triggered a re-render. Moved `AbortController` creation inside the effect and removed `conversationId` from deps.
- **Fix workspace picker fallback**: When `user.organizations` is empty, fall back to `user.workspaces` from the `/me` endpoint, using the region from the JWT token claim (`https://dust.tt/region`) instead of a hardcoded value.

## Test plan
- [x] Ask Dust a question → streaming response appears incrementally
- [x] Chain of thought / thinking toast updates during generation
- [x] Final answer displays correctly with success toast
- [x] Workspace picker shows workspaces when organizations list is empty
- [x] Region is correctly detected from JWT token

### DIFF
--- a/extensions/dust-tt/src/answerQuestion.tsx
+++ b/extensions/dust-tt/src/answerQuestion.tsx
@@ -214,6 +214,7 @@ async function answerQuestion({
           title: "Thinking...",
         });
 
+        let streamError = false;
         const processEvent = (eventData: Record<string, unknown>) => {
           const event = eventData.data as Record<string, unknown> | undefined;
           if (!event || !event.type) return;
@@ -223,12 +224,14 @@ async function answerQuestion({
               const error = event.error as { code: string; message: string };
               console.error(`User message error: code: ${error.code} message: ${error.message}`);
               setDustAnswer(`**User message error** ${error.message}`);
+              streamError = true;
               break;
             }
             case "agent_error": {
               const error = event.error as { code: string; message: string };
               console.error(`Agent message error: code: ${error.code} message: ${error.message}`);
               setDustAnswer(`**Dust API error** ${error.message}`);
+              streamError = true;
               break;
             }
             case "agent_action_success": {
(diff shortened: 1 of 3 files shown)


extensions/dust-tt/src/answerQuestion.tsx
```diff
@@ -168,92 +170,139 @@ async function answerQuestion({
...
-                break;
-              }
-              case "agent_message_success": {
-                answer = processAction({ content: event.message.content ?? "", action, setDustDocuments });
-                showToast({
-                  style: Toast.Style.Success,
-                  title: `@${agent.name} answered your question`,
-                  message: question,
+        let streamError = false;
+        const processEvent = (eventData: Record<string, unknown>) => {
```
adrsimon, line 218: note to self:
compared to sdk's implem, we're notably losing the `lastEventId` tracking
the SDK resumes from the last event on reconnection, avoiding duplicate/missed events + the reconnection logic on some http errors

we also lose the typing for AgentEvent types

not asking to implem here at all, thanks for taking a stab at those fixes! 
just noting for future us so we clean this up a bit, but looks like it works well enough as is 👍

adrsimon (approved):
thanks a lot for taking a stab at those, looks legit 👍

0xdhrv (approved):
Looks good to me, approved ✅

PR #26809, opened 2026-04-01
Fix {date} template variable using UTC instead of local timezone

shohei-majima:
## Summary
- The `{date}` template variable in Slack Templated Message uses `toISOString().split("T")[0]`, which returns a UTC-based date
- This causes the date to be off by one day for users in positive UTC offset timezones (e.g., JST/UTC+9) between midnight and 09:00 local time
- Changed to use `getFullYear()`/`getMonth()`/`getDate()` which respect the local timezone, consistent with how `{time}` already uses `toTimeString()`


pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #26832, opened 2026-04-01
Update prism-launcher

TakenMC:
## Description

- Added instance versions info to instace picker

## Screencast

image

sadly the images are broken on the windows version of the app so it's not showing tho it should show an icon for each version present so you can see what each instance is using

## Checklist


tomjpeg:
I'm still against adding extra conditions to obtain versions from specific software types like Forge and Fabric. We can consider adding something in the future, but that would require us to change the entire structure of the extension, rather than just patching it in one spot.

When we take that approach, the current code quickly starts to look strange and becomes difficult to maintain.

We already mentioned this in your last 4 PRs.

lockieluke:
What model did you use to generate the code?

0xdhrv:
Hey @lockieluke @tomjpeg 👋 

Thoughts on this?

tomjpeg:
I’m still strongly against this kind of integration because the extension should remain simple and adhere to the principles of the CLI tool of Prism. If we start adding such features, it will quickly become a mess. However, if he has such an interest in this, we can merge it. But personally, I would say no.

PR #26838, opened 2026-04-01
Add Obsidian WikiLinker extension

geodeo42:
## Description

Search your Obsidian vault by note name and paste a `[[wikilink]]` directly into whatever app you're typing in — no Obsidian window needed.

## What the extension does

- Fuzzy-searches all `.md` files in your vault
- Three actions: **Paste Wikilink** `[[Note Name]]`, **Paste Wikilink with Path** `[[Folder/Note Name]]`, **Copy Wikilink**
- Vault path is configurable via extension preferences (default: `~/vaults/Notes`)
- Excludes `.trash`, `.smart-env`, and `Templates` folders automatically
- No Obsidian needs to be running — reads the vault directly from disk

## Checklist

- [x] I have a 512x512 icon
- [x] I have a 1280x800 screenshot
- [x] The extension has a `version` field in `package.json`
- [x] No private data in the code


0xdhrv (commented):

- https://www.raycast.com/marcjulian/obsidian by @marcjulianschwarz

PR #26849, opened 2026-04-02
[hop] Add Ghostty support and generic custom terminal handler

danmartuszewski:
## Summary
- Added Ghostty as a first-class terminal option in the dropdown preferences
- Added Ghostty-specific AppleScript handler
- Made the `default` case in `getAppleScript()` work as a generic handler for arbitrary custom terminals instead of falling back to Terminal.app

## Fixes
Closes #26799 — custom terminals (like Ghostty) selected via "Other (specify below)" fell through to Terminal.app because there was no matching case and the default was hardcoded to Terminal.

## Test Plan
- [ ] Select Ghostty from the terminal dropdown — should open Ghostty and execute SSH command
- [ ] Select "Other" and type a custom terminal name — should activate that app and type the command via System Events
- [ ] Verify existing terminals (Terminal, iTerm, Warp, Alacritty, kitty) still work as before

### DIFF
--- a/extensions/hop/CHANGELOG.md
+++ b/extensions/hop/CHANGELOG.md
@@ -1,5 +1,16 @@
 # Changelog
 
+## [Ghostty Support] - 2026-04-02
+
+### Features
+
+- Added Ghostty as a first-class terminal option
+- Generic handler for custom terminals — "Other (specify below)" now works with any terminal app
+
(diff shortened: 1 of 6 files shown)


0xdhrv (approved):
Looks good to me, approved ✅

PR #26882, opened 2026-04-03
Fix extend-display to support Chinese macOS system language

blankhub:
## Description

Fixed the issue where the extend-display extension failed to work on Chinese macOS systems.

## Problem

The extension used hardcoded English text "Mirror or extend to" to locate the display menu in System Settings. On Chinese macOS systems, this menu is labeled "镜像或扩展至", causing the extension to fail with the error that it cannot find the display menu.

## Solution

Added system language detection to automatically use the appropriate menu text based on the system language:
- English systems: "Mirror or extend to"
- Chinese systems: "镜像或扩展至"

The fix checks the `LANG` and `LANGUAGE` environment variables to determine the system language.

## Changes

- Modified `src/utils/connect.ts` to detect system language and select appropriate menu text
- Modified `src/utils/displays.ts` with the same fix

## Testing

- [ ] Tested on English macOS (should continue to work as before)
- [ ] Tested on Chinese macOS (should now work correctly)

## Related Issues

Fixes compatibility issues for Chinese macOS users.


0xdhrv (commented):
Hey @blankhub 👋 

At the moment, Raycast doesn't support localization and only supports US English. Therefore, please avoid introducing your custom way to localize your extension. Also for parsing it would be same case.

PR #26876, opened 2026-04-03
[codex] Follow up on skills extension maintainability

keito4:
## Summary
- add hook-level tests for `useSkillContent` and restore lower-level `skill-content` tests
- expand `skills-cli` coverage around CLI error normalization and agent id mapping
- unify search and manage empty/error states with shared retry UI
- document the maintainer workflow and add a manual smoke checklist

## Why
- follow up on the recent maintainability work for the `skills` extension
- make regressions easier to catch in both tests and routine manual verification

## Validation
- `npm run validate`
- `npm run build`
- `npm run test:api-live`

## Notes
- `extensions/skills/MAINTAINABILITY_FOLLOWUP_PLAN.md` is intentionally left out of this PR

PR #26865, opened 2026-04-02
Improve maintainability of the skills extension

keito4:
## Summary
- add a small test harness for the `skills` extension with unit tests and an opt-in live API test
- extract search/content-fetching logic into testable helpers without regressing the current CLI and content-loading behavior
- improve a few empty states in the search and manage flows while keeping the current upstream functionality intact

## Testing
- npm run validate
- npm run build
- npm run test:api-live

PR #26895, opened 2026-04-03
[Skills] Remove test infra and reduce complexity

DaleSeo:
## Description

After reviewing #26865 and #26876 in detail, I found several issues that make the added code a net negative for maintainability which is the opposite of its stated goal. This PR reverses the test infrastructure and over-abstractions introduced, which added maintenance burden without providing meaningful value. The legitimate code extractions (`api.ts`, `skill-content.ts`) are preserved since they genuinely simplify the hooks.

### Tests that don't test anything meaningful

The `useSkillContent.test.ts` hook test fully mocks `useCachedPromise` and `fetchSkillContent`, then asserts that the mock's return value passes through unchanged. For example:

```ts
vi.mocked(useCachedPromise).mockReturnValue({
  data: { body: "# Skill Body", frontmatter: { title: "Example" } },
  isLoading: false,
} as ReturnType<typeof useCachedPromise>);

const result = useSkillContent(sampleSkill);
expect(result.content).toBe("# Skill Body"); // just checking the mock's own value
```

This verifies that mocks return what you told them to return. It catches zero real regressions. The same pattern applies to most of the other test files.

### Incomplete test stub that will break on any growth

The `@raycast/api` stub (`test/raycast-api.ts`) only exports `environment` and `getPreferenceValues`. Any test that touches `showToast`, `Icon`, `List`, `Action`, or any other Raycast API will fail immediately. This was already evident in #26876 where `@raycast/utils` had to be separately mocked with `vi.mock()`. The stub creates a false sense of test coverage while being fundamentally fragile.

### Over-abstracted UI wrappers

`CommandStates.tsx` introduced three components: `RetryAction`, `CommandErrorDetail`, and `CommandEmptyView`, each of which is a trivial wrapper:

```tsx
// The entire RetryAction component
export function RetryAction({ onAction, title = "Retry" }: RetryActionProps) {
  return <Action title={title} onAction={onAction} icon={Icon.RotateClockwise} />;
}
```

These add a layer of indirection without reducing duplication in any meaningful way. When reading the command files, you now have to jump to another file to understand what is just a standard Raycast `Action` or `List.EmptyView`. Inlining these makes the code more readable.

### Unnecessary export surface for testing

Nine internal functions were changed from private to `export` solely so tests could import them: `stripAnsi`, `shellEscape`, `normalizeCliError`, `parseSkillsListJson`, `agentDisplayNameToId` in `skills-cli.ts`, and `buildOwnerCounts`, `filterSkillsByOwner`, `resolveOwnerSelection` in `useOwnerFilter.ts`, and `buildSkillContentUrls` in `skill-content.ts`. Exporting internal functions for testing weakens module encapsulation and signals to other developers that these are part of the public API when they are not.

### No ecosystem support, no maintenance path

The Raycast docs don't mention unit testing anywhere. The PR review checklist only asks for `npm run build` and manual testing in Raycast, and the repo CI runs build, lint, and AI checks. Not extension-level test scripts. So even if tests exist, nothing enforces them. `@raycast/api` is not designed to be stubbed, which is why #26865 had to hand-roll a fake in `test/raycast-api.ts` that only covers two of its exports. The default extension template ships with no test setup, and almost no extension in the store has tests. Building a bespoke vitest setup for one extension means future contributors have to understand and maintain something nobody else in the ecosystem uses, with no CI to catch breakage. `TESTING.md` (a 43-line manual smoke checklist) has the same problem. It will go stale the moment someone changes the UI without updating the doc.

## Checklist


keito4:
@DaleSeo 

Sorry for the delayed reply, and thanks for laying out the context so clearly.

I think that’s a fair assessment. My goal with #26865 / #26876 was to add some regression coverage around logic in this extension that felt easy to break, not to introduce a bespoke testing workflow unless it was clearly paying for itself. Looking back, though, the mocked hook tests, custom API stubs, test-only exports, and extra process pushed things further than they should have.

You’re also right that changes with this kind of impact on contributor workflow should have had more discussion up front.

So I’m okay treating this PR as the reset point for the heavier parts of that setup. I still think there may be room later for a much narrower set of low-maintenance tests around code that’s intentionally structured for it, but that should be a separate conversation and should start from clear value rather than more scaffolding.

Thanks again for stepping back and reevaluating the whole setup!

0xdhrv (approved):
Looks good to me, approved ✅

PR #26917, opened 2026-04-04
Show installed badge in skills search results

keito4:
## Summary

- Show a green "Installed" tag on search results for skills that are already installed locally
- Helps users quickly identify which skills they already have when browsing search results
- Badge is visible in both detail panel and compact list views

## Screenshot

Installed badge in search results

## Implementation

- Add `useInstalledSkillNames` hook that fetches installed skill names via the CLI and exposes them as a `Set`
- Match search results against installed skills using `skillId`, which corresponds to the CLI's `name` field per the existing `discoverAgents` contract
- Build `accessories` array inline in `SkillListItem` for simplicity

## Test plan

- [x] `npm run build`
- [x] Verified in Raycast: installed skills show "Installed" tag, non-installed skills do not
- [x] Verified both detail panel on/off states


pernielsentikaer (changes requested):
Hi @keito4 👋

The previous changelog entry mentions adding hook-level tests for `useSkillContent`. The new `useInstalledSkillNames` hook follows the same pattern but ships without tests. At minimum, a test verifying the `Set` is correctly built from CLI output would catch the name/skillId mismatch.

---

pernielsentikaer (commented):
This PR adds an "Installed" badge to skill search results. The new useInstalledSkillNames hook fetches locally installed skills via listInstalledSkills() and collects their .name into a Set. However, the consumer in search-skills.tsx checks membership using skill.skillId — a different field from the search result model.

The Skill type clearly has distinct name, id, and skillId fields (visible in the existing keywords array). If s.name from the CLI doesn't match skill.skillId from the API, the badge will silently never render. This is the core correctness concern.

extensions/skills/src/hooks/useInstalledSkillNames.test.ts
```diff
@@ -0,0 +1,99 @@
...
+
+    useInstalledSkillNames();
+
+    const [loader] = vi.mocked(useCachedPromise).mock.calls[0];
+
+    await expect(loader()).resolves.toEqual(["skill-a", "skill-b"]);
+    expect(listInstalledSkills).toHaveBeenCalled();
+  });
+
+  it("matches by skillId only — id and name alone must not match", () => {
```
DaleSeo: I'm not sure what the purpose of this test is. It's titled "matches by skillId only," but the hook doesn't have any matching logic. If someone changed the matching key from `skill.skillId` to `skill.name` in `search-skills.tsx`, this test would still pass.

extensions/skills/src/components/SkillListItem.tsx
```diff
@@ -179,6 +187,13 @@ export function SkillListItem({ skill, rank, isSelected, isShowingDetail, onTogg
       ? { source: Icon.Trophy, tintColor: rank <= 3 ? Color.Yellow : Color.SecondaryText }
       : { source: Icon.Hammer };
 
+  const accessories = useMemo(() => {
```
DaleSeo: I'm afraid using `useMemo` here just complicates things without any real benefit. The Raycast list items will re-render when the selection changes, so this memoization will break often.

DaleSeo (commented):
This is a UI change, but there are no screenshots showing what it looks like. A quick screenshot would really help with the review and make it easier to reference later. Could you please follow the repo's PR template, @keito4?

pernielsentikaer (changes requested):
@keito4 can you look into the suggestions

---

pernielsentikaer (changes requested):
Hi @keito4 👋

You used the wrong background of the metadata image, it needs to have the same background as the rest of them

---

DaleSeo:
@keito4 Could you please fix the broken screenshot in the PR description?

2026-04-11 at 13 55 48

To get consistent screenshots, consider using the built-in screenshot tool in Raycast. Just go to Raycast's Settings, then Advanced, and scroll down to Window Capture:

image

keito4:
Sorry for the late response!

- Updated all metadata screenshots with a consistent background
- Replaced skills-4.png → skills-1.png (no need for a 4th image)
- Fixed the broken screenshot in the PR description
- Merged main and removed test files (vitest was removed upstream)

Note: skills-2 and skills-3 are shared with #26936, so they include the Repo Activity feature from that PR. They will be consistent regardless of merge order.

Ready for re-review!

pernielsentikaer (approved):
Looks good to me, approved 🔥

Issue #25664, opened 2026-02-20
[Toggl Track] ...


TheTechPastor:
### Extension

https://www.raycast.com/franzwilhelm/toggl-track

### Raycast Version

1.104.6

### Description

since September Toggl has created reduced their API access giving only some calls every hour.  As a paid Toggl user I get 240 calls an hour.  your extension hits the 240 limit in 8 minutes after install.  So the extension has to be modified to be do a steady query of API calls, else it cannot be used

### Steps To Reproduce

1. install the toggl extension 
2. insert the API code
3. wait

### Current Behaviour

maxes out the toggl API which locks my toggles account

### Expected Behaviour

only query the API when I do something with the extension

PR #26932, opened 2026-04-05
fix(toggl-track): comprehensive API rate limit optimization

speckofthecosmos:
> **Note:** The rate limit retry fix (items 4 & 5) was extracted to #26979 and has been merged. This PR now focuses on the API optimization changes only.

## Description

A deeper fix for #25664. PR #25696 helped (bumping the interval and cache TTL), but the extension was still burning through API quota fast — especially for free-tier users at 30 req/hour. This PR goes after the root causes: too many parallel requests, redundant fetches, and a retry loop that actually makes rate limiting worse.

### 1. Resource fetch consolidation (5 API calls → 1)
Opening any time entry form fires requests for workspaces, projects, clients, tags, and tasks — 5 parallel calls that can trigger 429 concurrency errors immediately. These now go through a single `/me?with_related_data=true` bootstrap call. Concurrent callers share the same in-flight promise, and results get seeded into the existing cache.

### 2. Menu bar polling (3m → 10m)
At 3-minute intervals, background polling alone eats ~20 req/hour — that's two-thirds of a free-tier user's entire budget before they've even touched the extension. Bumped to 10m (~6 req/hour), which leaves room for actually using the thing.

### 3. Optimistic running entry state
After starting or stopping a timer, the API response already contains the full time entry. Previously this was thrown away and then re-fetched via `GET /me/time_entries/current`. Now the response is cached immediately — no redundant round-trip. Cold starts still hit the API to catch timers started from web/mobile.

## Impact

| Scenario | Before | After |
|---|---|---|
| Background polling (menu bar) | ~20 req/hour | ~6 req/hour |
| Opening a time entry form | 5 requests | 1 request |
| Starting/stopping a timer | 1 redundant fetch | 0 (cached) |

## Checklist

### DIFF
--- a/extensions/toggl-track/CHANGELOG.md
+++ b/extensions/toggl-track/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Toggl Track Changelog
 
-## [Bug Fixes] - {PR_MERGE_DATE}
+## [Bug Fixes] - 2026-04-15
 
 - Consolidated 5 parallel resource API calls into a single `/me?with_related_data=true` bootstrap request, reducing API usage from 5 calls to 1 when opening time entry forms
 - Reduced Menu Bar background refresh interval from 3 minutes to 10 minutes to further reduce API consumption for free-tier users (30 calls/hour limit)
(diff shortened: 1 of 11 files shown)


pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #26935, opened 2026-04-05
Add raycast-bluetooth-devices extension

athNdev:
## Description

Manage Windows Bluetooth devices directly from Raycast. 
- Connect
- Disconnect
- Pair, Unpair
- Switch audio output for bluetooth devices
- Toggle Bluetooth

Extension also bundles a .NET 8 C# application which is a CLI for accessing Windows-native Bluetooth APIs, the extension is a frontend for it,

## Screencast

image

## Checklist


0xdhrv (commented):

- https://www.raycast.com/VladCuciureanu/toothpick by @VladCuciureanu

I suggest we enhance the existing extension instead of creating a new one. You can add support for Windows to the existing extension.

athNdev:
@0xdhrv This extension contains a bundled windows CLI for communicating with native windows API for bluetooth connections. I think it'd be a bad idea to bundle all that into the MacOS extension.

0xdhrv:
> @0xdhrv This extension contains a bundled windows CLI for communicating with native windows API for bluetooth connections. I think it'd be a bad idea to bundle all that into the MacOS extension.

You can have conditions in the code, like the [Color Picker extension](https://www.raycast.com/thomas/color-picker) does have completely different code for Windows and macOS

PR #26952, opened 2026-04-05
1Password: Enable cross-field search

st0mich:
## Summary

- Search queries now match across **all item fields** (title, username/email, URL hostnames, vault name) instead of only matching against title or username individually.
- Each space-separated token is matched independently, so queries combining parts of different fields now return the expected results.
- Applied to both v7 and v8 item lists.

## Motivation

Previously, searching for a combination of email and service name returned no results because Raycast's built-in filtering only matched tokens within single fields. This made it hard to find specific logins when you remembered parts of both your email and the service name.

## Implementation

- Replaced Raycast's built-in `List` filtering with custom `matchesSearch()` that splits the query into tokens and checks each token against all searchable fields.
- Set `filtering={false}` and `onSearchTextChange` on `List` to handle filtering manually.
- Extracted URL hostnames from `item.urls` to make domains searchable.

## Test plan

- [ ] Search by title only → works as before
- [ ] Search by email only → works as before
- [ ] Search by combination of email + service name → now returns matching items
- [ ] Search by URL hostname → now returns matching items
- [ ] Category filter still works alongside search
- [ ] Empty search shows all items


pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #26978, opened 2026-04-06
fix(toggl-track): Fix negative duration in today's total

speckofthecosmos:
## Summary
- Running time entries have negative `duration` values (Toggl API convention: `-1 * unix_start_time`). The previous code used `.slice(1)` to skip the running entry, assuming it was always at array index 0. When the `at`-based sort order placed it elsewhere, its large negative duration leaked into the sum, producing displays like `Today: -2465982:-20:-55`.
- Replaced position-based skipping with a `duration >= 0` filter — the API-guaranteed way to identify completed entries.
- Added `Math.max(0, seconds)` guard in `formatSeconds` as defense in depth.

## Test plan
- [ ] Start a timer → verify "Today: HH:MM:SS" shows correct positive duration
- [ ] Stop the timer → verify total updates correctly
- [ ] With multiple entries today → verify sum is accurate
- [ ] Menu bar and list view both display correctly

### DIFF
--- a/extensions/toggl-track/CHANGELOG.md
+++ b/extensions/toggl-track/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Toggl Track Changelog
 
-## [Bug Fixes] - {PR_MERGE_DATE}
+## [Bug Fixes] - 2026-04-13
 
 - Fix negative duration display in today's total when a running time entry is not first in the array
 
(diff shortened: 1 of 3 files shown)


pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #26979, opened 2026-04-06
fix(toggl-track): Fix infinite retry loop on API rate limit

speckofthecosmos:
## Summary
- The HTTP client retries infinitely on HTTP 429 with a fixed 1-second delay and no retry cap. Each retry counts against Toggl's leaky-bucket rate limit (30 req/hr for free tier), preventing recovery and silently draining the user's entire hourly API budget with no UI feedback.
- Adds exponential backoff (1s, 5s, 15s) with a 3-retry cap, then throws so Raycast can surface an error toast.
- Adds a distinct handler for HTTP 402 (quota exhausted) as a non-retryable error.

## Test plan
- [ ] Trigger rate limiting (exhaust free-tier quota) → verify error toast appears after 3 retries instead of silent infinite loop
- [ ] Normal API calls still work without delay
- [ ] Extension recovers cleanly after rate limit window expires


pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #26987, opened 2026-04-07
Dia: Add broken URL detection and fix

Berckan:
## Summary

URLs copied from terminals, emails, or narrow viewports often get split across multiple lines, making them impossible to open directly. This PR adds automatic detection and repair:

- **Fix Broken URL** standalone command: reads the clipboard, joins broken lines into a clean URL, and copies it back. Works silently with a HUD confirmation.
- **Inline detection in Search Dia**: when a multi-line URL is pasted into the search bar, a "Fixed URL" section appears at the top with the repaired link ready to open in Dia or copy.

The detection logic (`fixBrokenURL` in `utils.ts`) checks if joining lines produces a valid URL pattern, so it only triggers when there's actually a broken URL to fix.

## Type of Change

- [x] New feature (non-breaking change that adds functionality)

## Test Plan

- Copy a long URL, paste it into a text editor, add line breaks in the middle, copy that broken version
- Open Search Dia and paste it: "Fixed URL" section should appear with the correct URL and a wand icon
- Run "Fix Broken URL" command: clipboard should now contain the clean URL, confirmed by HUD
- Paste a normal search query: no "Fixed URL" section should appear
- Paste a single-line URL: only "Open URL" section should appear, not "Fixed URL"

### DIFF
--- a/extensions/dia/CHANGELOG.md
+++ b/extensions/dia/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Dia Changelog
 
-## [Broken URL detection and fix] - 2026-04-07
+## [Broken URL detection and fix] - {PR_MERGE_DATE}
 
 - **Fix Broken URL**: New no-view command that detects broken URLs in clipboard (split across lines from terminals, emails, or narrow viewports), joins them, and copies the clean URL back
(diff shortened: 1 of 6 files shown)


pernielsentikaer (changes requested):
Hi 👋

I kinda like the idea of changing it inside the current command, but I’m not sure the new command "Fix Broken URL" is too much since it feels like more of a general thing; don’t you think so?

---

PR #27003, opened 2026-04-08
Update evm-toolkit extension

imqdee:
## Description

This update reshapes how users reach the extension's commands and broadens coverage of two areas where a single command was getting cramped.

**Open Commands hub.** The extension now ships ten commands. Pinning each one individually pollutes the Raycast root, but pinning none makes the extension hard to discover. The new `Open Commands` view is a searchable list grouped by purpose (Explore, Intel, Profiles, Tools, Reference) that users can pin as a single entrypoint. Commands that need a network pick open an inline form with the clipboard pre-filled and the network dropdown defaulting to the user's preferred network, so the hub is not just a launcher but a place where a full command can complete without leaving the hub. Individual commands stay at the Raycast root so power users keep their existing muscle memory; users who prefer the hub can hide the ones they don't want via Extension preferences.

**Intel split into three commands.** The old `Open Intel` was tied to Arkham only. This PR splits it into `Open Arkham Intel`, `Open MetaSleuth Intel`, and `Open Bubblemaps Intel`. Arkham stays network-agnostic; MetaSleuth and Bubblemaps each take a network dropdown because their URLs are network-scoped, and merging them into a single "tool + network" command would have made the argument list inconsistent across tools.

**Profile split into DeBank and Zerion.** DeBank and Zerion are the two portfolio trackers users actually want to pin, and each has strong opinionated users. Rather than force one as the default and bury the other, this PR exposes both as first-class commands (`Open DeBank Profile`, `Open Zerion Profile`). Both are network-agnostic so no dropdown is needed.

**Local CI parity.** A `lefthook.yml` runs `npm run lint:ci` and `npm run build` on pre-commit so the same checks the Raycast CI runs are enforced locally, avoiding "fix lint" round-trips during review.

## Checklist


0xdhrv (approved):
Looks good to me, approved ✅

PR #27012, opened 2026-04-08
Google Calendar: Fix DST offset for ET/CT/MT/PT timezone abbreviations

alexpriest:
## Summary
- Short US timezone abbreviations (ET, CT, MT, PT) were hardcoded to standard time offsets (e.g., ET → EST/UTC-5), causing events created with Quick Create Event to land 1 hour off during daylight saving time
- Now resolves these abbreviations dynamically via IANA timezone names (`America/New_York`, etc.) so the offset correctly reflects DST status at the time of the event
- Explicit abbreviations (EST, EDT, CST, CDT, etc.) are unaffected since they already specify standard vs daylight

## Test plan
- [ ] Type "Test event 4pm ET" during DST (March–November) — event should land at 3pm CT / 1pm PT
- [ ] Type "Test event 4pm EST" — event should land at 4pm EST regardless of DST
- [ ] Type "Test event 4pm PT" during DST — should resolve to PDT (UTC-7)
- [ ] Type "Test event 4pm CT" outside DST — should resolve to CST (UTC-6)

### DIFF
--- a/extensions/google-calendar/CHANGELOG.md
+++ b/extensions/google-calendar/CHANGELOG.md
@@ -1,5 +1,14 @@
 # Google Calendar Changelog
 
+## [1.4.0] - 2025-12-23
+
+- Add new "Quick Create Event" command with natural language parsing
+  - Calendar selector with `/calendarname` syntax and fuzzy matching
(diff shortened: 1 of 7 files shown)


0xdhrv (commented):

---

PR #27046, opened 2026-04-09
fix(google-calendar): correct timezone offset for half-hour timezones

tanuj-b:
## Description

Fixes incorrect timezone offset calculation in `toISO8601WithTimezoneOffset()` that affects any timezone with a non-zero minute component (IST +05:30, Nepal +05:45, Iran +03:30, Newfoundland -03:30, etc.).

The bug is in the order of operations when computing `offsetHours`:

```js
// Before — floors first, then takes absolute value
// For IST (offset = -330): abs(floor(-330/60)) = abs(floor(-5.5)) = abs(-6) = 6 ❌
const offsetHours = Math.abs(Math.floor(offset / 60))

// After — absolute value first, then floors
// For IST (offset = -330): floor(abs(-330)/60) = floor(330/60) = floor(5.5) = 5 ✅
const offsetHours = Math.floor(Math.abs(offset) / 60)
```

This caused events created via the Google Calendar extension to be scheduled one hour off in affected timezones, because the formatted timezone string was `+06:30` instead of the correct `+05:30`.

**Change in `src/lib/utils.ts`:** one-line fix — swap `Math.abs(Math.floor(offset / 60))` → `Math.floor(Math.abs(offset) / 60)`.

Fixes #21062

## Checklist

### DIFF
--- a/extensions/google-calendar/CHANGELOG.md
+++ b/extensions/google-calendar/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Google Calendar Changelog
 
-## [1.4.2] - {PR_MERGE_DATE}
+## [1.4.2] - 2026-04-09
 
 - Fix timezone offset calculation for half-hour timezones (e.g. IST +05:30) that caused events to be scheduled one hour off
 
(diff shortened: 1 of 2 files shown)


0xdhrv (approved):
Looks good to me, approved ✅

PR #27074, opened 2026-04-10
Add reading-time extension

fromsergio:
## Checklist


0xdhrv (commented):
Thanks for your contribution @fromsergio 🔥

- https://www.raycast.com/itsmingjie/word-count by @itsmingjie

I think, you can add more features like WPM preferences and also update the description to make it more clear to increase the discoverability for the use case of reading time.

fromsergio:
@0xdhrv thanks for checking this out so quickly. Word-count is very different from reading-time. Not only does reading-time take input from the clipboard avoiding unnecessary copy and paste,  but it gives estimates of reading time in seconds. On the word-count plugin if I count 2 words or 100 it gives me the same outcome which is 1 min. 
Honestly it's not even close in funcionality, I cannot substitute reading-time with word-count. 

I understand what you mean in terms of avoiding duplication but this really is not the case. There are actual unnecessary duplications in the extension store such as" timezone converter " and "time zone converter". This is not one. Either way i'm only sharing for the community's benefit, I'm already running this locally. 

Lmk!

0xdhrv:
> @0xdhrv thanks for checking this out so quickly. Word-count is very different from reading-time. Not only does reading-time take input from the clipboard avoiding unnecessary copy and paste, but it gives estimates of reading time in seconds. On the word-count plugin if I count 2 words or 100 it gives me the same outcome which is 1 min. Honestly it's not even close in funcionality, I cannot substitute reading-time with word-count.
> 
> I understand what you mean in terms of avoiding duplication but this really is not the case. There are actual unnecessary duplications in the extension store such as" timezone converter " and "time zone converter". This is not one. Either way i'm only sharing for the community's benefit, I'm already running this locally.
> 
> Lmk!

That makes sense, thanks for the explanation. Let me go ahead and review and update as needed

0xdhrv (approved):
Looks good to me, approved ✅

fromsergio:
@0xdhrv Thank you! It's currently assigned to @sergio which is not me, my Raycast account is @fromserg 

Have a nice one 🫡

extensions/reading-time/package.json
```diff
@@ -0,0 +1,52 @@
+{
+  "$schema": "https://www.raycast.com/schemas/extension.json",
+  "name": "reading-time",
+  "title": "Reading Time",
+  "description": "Estimate clipboard reading time as X min read",
+  "icon": "command-icon.png",
+  "author": "sergio",
```
0xdhrv, line 7: Hey @fromsergio 

Can you make a new PR to update this one?

0xdhrv:
> @0xdhrv Thank you! It's currently assigned to @sergio which is not me, my Raycast account is @fromserg
> 
> Have a nice one 🫡

I have created a new pull request for the same https://github.com/raycast/extensions/pull/27142

fromsergio:
I missed this and went with #27143

PR #27081, opened 2026-04-10
Google Meet: Add Dia browser support

wunnle:
## Summary
- Adds [Dia browser](https://dia.google/) to the list of supported browsers
- Uses `Cmd+Shift+C` keystroke simulation (via AppleScript System Events) to copy the current tab URL, since Dia doesn't expose tab URLs through standard AppleScript
- Adds `wunnle` as a contributor

## How it works
Similar to the existing Firefox workaround, the Dia script:
1. Focuses the Dia browser window
2. Simulates `Cmd+Shift+C` to copy the URL to the clipboard
3. Reads the URL from the clipboard via `pbpaste`

## Test plan
- [ ] Set Dia as the preferred browser in extension preferences
- [ ] Run "Create Meet" command and verify the meeting link is copied to clipboard
- [ ] Run "Create Meet with Specified Profile" and verify the link is copied without query parameters

### DIFF
--- a/extensions/google-meet/CHANGELOG.md
+++ b/extensions/google-meet/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Google Meet Changelog
 
+## [Improvement] - {PR_MERGE_DATE}
+
+- Add Dia browser support
+
+## [Improvement] - 2026-02-13
+
 ## [Improvement] - 2026-04-11
 
 - Add Dia browser support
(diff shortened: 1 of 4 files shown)


0xdhrv (commented):
Hey @wunnle 👋

Thanks for your contribution!

Could you resolve the merge conflict?

---

PR #27133, opened 2026-04-13
[Cal.com] Major expansion: availability management, out of office, and View Bookings redesign

andrewbenson:
## Summary

This PR expands the **Cal.com Share Meeting Links** extension into a more general **Cal.com** workflow surface. It includes the v1 → v2 API migration (originally opened as #27118 — this PR supersedes it), plus three substantial new feature areas and a fix for a long-standing pagination bug in View Bookings.

> **Naming note for reviewers:** the extension's `name` slug is unchanged (`cal-com-share-meeting-links`) so existing users keep their stored API key and cached state. The `title` is bumped from `Cal.com - Share Meeting Links` to just `Cal.com` since the scope no longer fits the original name. Open to feedback on whether to keep that title or pick something else.

## Why one PR for all this

The features are intentionally cohesive: they share the same API layer (`src/api/cal.com.ts`), reuse the same `useCachedPromise` cache-namespace fix, follow the same forms/actions/optimistic-update patterns, and depend on common helpers in `src/lib/`. Splitting them into separate PRs would either force artificial dependencies or duplicate the shared infrastructure across reviews. The rebrand also only makes sense once the full feature set lands. Happy to split if reviewers prefer.

### 1. v1 → v2 API migration (was #27118)

- All API calls migrated from Cal.com v1 to v2 (v1 was permanently shut down on April 8, 2026)
- Authentication switched from query parameter to Bearer token header
- Required `cal-api-version` headers added per endpoint
- Replaced unified booking status update with separate confirm/decline endpoints
- Cancel booking switched from DELETE to POST with body
- Adopted v2 field names (`meetingUrl`, `location`, `lengthInMinutes`, etc.)

### 2. New: Availability schedule management

A new top-level `View Availability` command:

- List all of your schedules (default sorted to top)
- Sectioned detail view: Working Hours (Mon–Sun), Date Overrides, Settings (timezone, default flag, name)
- Edit working hours per day (up to 3 ranges with overlap validation)
- Add / edit / delete date overrides (date picker + time range, or "Unavailable all day")
- Edit timezone with a humanized, searchable IANA dropdown (`(GMT-7) America/Los Angeles`, searchable by city)
- Rename schedule, set default, and a `Cmd+Shift+T` "Set to Device Timezone" action available on every relevant surface
- Glance pane shows working hours by day, **upcoming** date overrides (with weekday context), and which event types use this schedule (as a `TagList`)
- All cache mutations propagate across views via a hoisted-fetcher pattern (fixes a real "edit doesn't refresh parent" bug discovered during QA)

### 3. New: Out of Office

Two new commands: `Out of Office` (browse/manage) and `Create Out of Office` (top-level form with native Raycast draft support):

- List current + upcoming OOO entries with reason-tinted icons matching cal.com's web emojis (🏝 vacation, 🛫 travel, 🤒 sick, 📅 public holiday, 🕒 unspecified)
- Each row shows the date range, weekday range, and accessories for the redirect target (with circular avatar) and active state
- Create / edit / delete entries with date range, reason enum, optional notes, and a searchable team-member redirect picker
- Team-member dropdown shows circular 32px avatars (with Cal.com URL normalization) and supports search by name, email, or team
- Self-filtered out of the redirect list; gracefully degrades when the user has no teammates or insufficient permission to list memberships
- Native Raycast draft persistence on the standalone "Create Out of Office" command (auto-save + resume prompt + discard)
- Quick links to the OOO and General Account settings pages on cal.com (the latter covers scheduled timezone change, which has no public API)

### 4. View Bookings — bug fix + redesign

**The bug:** `useBookings()` was calling `GET /v2/bookings` with no params, which defaults to `take=100, skip=0` in the API's internal order — effectively the oldest 100 bookings. Users with longer histories never saw recent or upcoming bookings.

**The fix + redesign:**

- Three section-specific hooks (Pending, Upcoming, Past, Cancelled) with the correct `status` + `sortStart` query params
- Past bookings lazy-load on scroll via Raycast's `useCachedPromise` pagination (50 per page)
- New layout: **Pending Confirmation → Upcoming → Past** (Cancelled hidden by default behind `Cmd+H`)
- `Cmd+Shift+A` toggle: **Show Only My Bookings** (hosts include you) vs **Show All Bookings** (includes team bookings hosted by others); default is filtered to yours, with a filter-aware empty state
- Safer destructive UX: Enter is now `Open Booking in Browser`, Cancel uses `Ctrl+X`, Accept `Cmd+Y`, Decline `Cmd+Shift+X`. The cancel-confirm flow is unchanged.
- New `Cmd+Shift+R` "Request Reschedule" action on Pending and Upcoming bookings — pushes a form where the user enters a reason, confirms, and the extension calls `POST /v2/bookings/{uid}/request-reschedule`. Cal.com cancels the original booking and emails the attendee a link to pick a new time.
- Booking URL fixed to use `https://app.cal.com/bookings/upcoming?uid=...` (the logged-in user view) instead of the public-facing `https://cal.com/booking/...`

## Implementation notes

- Multi-component caches share a stable function reference so `useCachedPromise`'s namespace (derived from `object_hash(fn)`) doesn't partition data across components — see comments on `fetchSchedules`, `fetchOOOEntries`, and `fetchPastBookingsPaginator`
- Type updates to `CalEventType.scheduleId` to enable the View Availability "Used by" linkage
- `src/lib/` now houses pure helpers per feature area (`schedule.ts`, `ooo.ts`, `bookings.ts`) — easily unit-testable later, no test framework added in this PR

## Test plan

- [x] `npx ray lint` clean (one pre-existing title-casing warning on the new "Create Out of Office" command — the suggested "Create out of Office" reads oddly so left as-is; happy to change)
- [x] `npx ray build -e dist` clean across all 5 entry points (`index`, `view-bookings`, `view-availability`, `out-of-office`, `create-out-of-office`)
- [x] Manual QA against a real Cal.com account verified each feature end-to-end (including a team account with multiple teammates for the OOO redirect picker)

## Breaking changes

None for end users. The `name` slug is unchanged so existing installs preserve their stored API key and `useCachedState` data. The `useBookings()` API helper is removed but it was internal only.

## Open items / known limitations

1. **Scheduled timezone change** has no public API. The `Cmd+Shift+A` action in the OOO command opens the relevant cal.com settings page as an escape hatch.
2. **Team-member listing** for the OOO redirect picker uses `/v2/teams` + `/v2/teams/{id}/memberships`. Returns full membership lists for org owners (verified) and gracefully degrades to an empty dropdown for users without permission. Other listing endpoints might exist that work for non-admins — would appreciate Cal.com guidance.
3. **Title casing** of the new command "Create Out of Office" — Raycast's lint suggests "Create out of Office" which feels wrong. Open to whatever reviewers prefer.


andrewbenson:
Addressed all three findings in fb2b955a4:

1. **`pop()` in `finally` (5 forms)** — fixed in `edit-day-hours`, `edit-override`, `edit-timezone`, `rename-schedule`, and `edit-ooo`'s `apply` function. They now `return` after `showFailureToast` and only `pop()` on success, matching the `handleDelete` pattern. Users keep their input and can retry on error.

2. **OOO past-entry filter documented** — added a comment on `fetchOOOEntries` explaining the intentional "current + upcoming only" scope and what to change if a "Show Past" toggle gets added later.

3. **Optimistic `toUser` on OOO create** — the synthetic entry now looks up the selected teammate from `useTeammates()` and embeds the matching `toUser` object so the avatar/name renders immediately, not just after the API responds.

pernielsentikaer (changes requested):
Hi 👋 

* Outdated deps — `@raycast/api` ^1.79.1 (should be ^1.100+) and `@raycast/utils` ^1.16.1 (should be ^2.x). These need bumping.

* Both `moment` and `dayjs` in dependencies — the existing code uses `moment`, the new code uses `dayjs`. Would be cleaner to pick one, but not blocking.

* No `.prettierrc` — standard Raycast prettier config is missing.

---

andrewbenson:
@pernielsentikaer Thanks for the quick feedback!

Outdated deps — fixed in `c2ceef021`:
- `@raycast/api` bumped from ^1.79.1 → ^1.104.12
- `@raycast/utils` bumped from ^1.16.1 → ^2.2.3
- `@types/react` pinned to 19.0.10 to match the version bundled inside `@raycast/api`
- Build passes clean across all 5 entry points.

`moment` + `dayjs` — acknowledged. Both are inherited from the original extension's code. Happy to consolidate in a follow-up, leaving as-is for now since it's not blocking.
         
`.prettierrc` — it's actually already tracked in the repo (since the original extension, commit `601a6e8ab`):
``` 
{
    "printWidth": 120,
    "singleQuote": false
  }
```
Let me know if you'd like a different config — happy to adjust.

I've marked the PR as ready for review but please let me know if there's any additional feedback!

pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #27148, opened 2026-04-13
Update TypeWhisper dictation tracking and Windows support

SeoFood:
## Description

This updates the TypeWhisper Raycast extension to use dictation session IDs for exact transcript lookup and marks the extension as available on Windows.

Previously, `Show Last Transcription` had to rely on the latest `/v1/history` entry, which could be ambiguous when multiple transcriptions or automations happened close together.

With this change:
- `Start Dictation` stores the session ID returned by the TypeWhisper API
- `Show Last Transcription` first checks `/v1/dictation/transcription?id=<uuid>` for the exact session result
- it falls back to `/v1/history?limit=1` when no tracked session is available
- the manifest metadata allows the extension on both macOS and Windows

This makes transcript retrieval deterministic and lets Raycast for Windows use the same local HTTP API integration.

## Screencast

No UI changes. This is an API integration and extension metadata update for the existing commands.

## Checklist

### DIFF
--- a/extensions/typewhisper/CHANGELOG.md
+++ b/extensions/typewhisper/CHANGELOG.md
@@ -1,5 +1,11 @@
 # TypeWhisper Changelog
 
+## [Improved Dictation Session Tracking] - 2026-04-13
+
+- Updated Start Dictation and Show Last Transcription to use dictation session IDs for exact transcript lookup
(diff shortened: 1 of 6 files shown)


0xdhrv (approved):
Looks good to me, approved ✅

PR #27181, opened 2026-04-14
[Brightness Control] Fix Brightness Up/Down failing when triggered by hotkey

### LINKED ISSUE #27085: [Brightness Control] Bug: Increase and decrease brightness don't work when triggered by hotkey.
### Extension

https://www.raycast.com/cali/brightness-control

### Raycast Version

1.104.12

### OS Version

26.3

### Description

The brightness up and down commands don't work when triggered via hotkeys, yet they work when triggering the command directly from the Raycast launcher.

### Steps To Reproduce

1. Install the extension
2. Set a hotkey to the commands for decrease and increase brightness. I.e. `Shift`+`→` 
3. Trigger the command via raycast (not using the hotkey). See that the command works and the brightness does change.
4. Trigger the command via hotkey. See that the toast alert says "Brightness Increased" or decrease. Yet no change has occurred.

I've tried this on two separate macs with the same results. I also use a studio display and have the same results with that.

### Current Behaviour

_No response_

### Expected Behaviour

_No response_

pavzagor:
## Description

Fixes #27085.

When **Brightness Up** or **Brightness Down** is bound to a hotkey (e.g. `Shift+→`), the command's HUD confirmed the action but the actual brightness did not change. Triggering the same command from the Raycast launcher worked correctly.

### Root cause

`src/script.ts` synthesized a brightness keypress via AppleScript (`tell application "System Events" to key code 144/145`). When invoked via a hotkey, the user is still physically holding the hotkey modifier(s) at the moment the AppleScript runs, so macOS merges the held modifiers with the synthetic key code and the event becomes `Shift + F14/F15` rather than bare `F14/F15`. macOS does not interpret modified brightness keys as brightness adjustments, so nothing happens — but the HUD still fires because the AppleScript itself succeeded.

### Fix

Replace the AppleScript key-injection path with Lunar's relative brightness CLI, which already powers `Set Brightness` and `Max Brightness` in this extension. No synthetic key events, so held modifiers can't interfere.

- `src/up.ts` / `src/down.ts`: call a new `adjustCursorBrightness(±10)` helper in `src/utils/lunar.ts` that runs `lunar displays <serial> brightness +/-N` against the cursor display.
- `src/script.ts` and the `run-applescript` dependency are removed.
- HUD now shows the resulting display name and brightness percentage (e.g. `Studio Display: 70%`) instead of a static "Brightness increased".
- `Brightness Up` / `Brightness Down` now require Lunar (matches `Set Brightness` / `Max Brightness`); Lunar is auto-installed via Homebrew on first use by the existing `ensureLunarReady` flow.
- CHANGELOG and README updated.

### Verification

- Trigger via launcher → brightness changes, HUD shows new percentage.
- Trigger via `Shift+→` / `Shift+←` hotkey → brightness changes, HUD shows new percentage (the failing case from #27085).
- Repeated hotkey presses at 0% / 100% are clamped by Lunar — no errors.
- Cursor on secondary display → that display's brightness changes, not the main.
- Lunar uninstalled → existing `ensureLunarReady` toast appears with install action (unchanged behavior).

Screencast of the original failure is in [the issue](https://github.com/raycast/extensions/issues/27085#issuecomment-4233473380).

## Checklist

- [x] I ran \`npm run build\` and tested this distribution build in Raycast
- [x] I checked that assets used by the \`README\` are placed outside of the \`metadata\` folder


pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #27196, opened 2026-04-15
Add spotify-to-apple-music extension

stepankuzmin:
## Description

Adds `spotify-to-apple-music`, a no-view Raycast extension for macOS.

The command:

- reads the currently playing Spotify track via AppleScript
- copies the Spotify track URI to the clipboard
- searches Apple Music for the closest song match
- opens the matched track in the Music app
- falls back to the Apple Music search page when an exact match is not found

This extension is macOS-only because it depends on AppleScript integration with Spotify and Music.app.

## Screencast

No screencast attached.

This is a no-view command with no custom Raycast UI. Running it immediately resolves the current Spotify track and opens the best Apple Music match in Music.app.

## Checklist


stepankuzmin:
Bumping this in hopes of getting a human review

0xdhrv (commented):
Thanks for your contribution  

- https://www.raycast.com/Quentin23Soleil/music-link-converter

stepankuzmin:
Well, yeah. I used that extension for some time, but it bugged me that you first have to obtain the URL of the current track somehow, and the underlying song.link service was down quite often (it wasn't working for me for several months), hence this tiny alternative. But if it's preferred to keep more generic extensions in the Store, that also makes sense, so we could close this, no hard feelings :)

0xdhrv:
> Well, yeah. I used that extension for some time, but it bugged me that you first have to obtain the URL of the current track somehow, and the underlying song.link service was down quite often (it wasn't working for me for several months), hence this tiny alternative. But if it's preferred to keep more generic extensions in the Store, that also makes sense, so we could close this, no hard feelings :)

Thanks for understanding. You can contribute to the existing extension, and I think it would be a great update.

PR #27207, opened 2026-04-15
New Extension: Claude Code Cleaner

adigold:
## Description

Clean Clode removes pipes, box-drawing characters, broken line wraps, and excessive whitespace from Claude Code and Codex terminal output, turning mangled text into clean, readable content instantly.

## Commands

- **Clean Text** — paste mangled terminal output into a form, hit ⌘↩, get clean text copied to clipboard. Includes searchable local history.
- **Clean Clipboard** — no-view command that cleans the current clipboard and copies the result back. Ideal for assigning a hotkey.

## Motivation

Copying text from the Claude Code terminal often produces broken, pipe-filled output that is hard to reuse. This extension solves that in one keystroke.

## Screenshots

Included in `metadata/`.

## Checklist

- [x] Extension builds (`npm run build`)
- [x] Linting passes (`npm run lint`)
- [x] 512x512 icon included
- [x] Screenshots at 2000x1250 included
- [x] README included
- [x] No sensitive data, no network requests
- [x] Author handle verified (`adigold1`)


adigold:
Hi @pernielsentikaer 👋

Thank you for the review, and hope everything is ok on your side!

I went back to the drawing board on your feedback about ending up at the Raycast root with a toast. You were right — the old flow pushed a `Detail` view on top of the form, so submitting cascaded back through three screens. Felt wrong every time.

**What changed in this round:**

- **Inline result, no more navigation hop.** The `Clean Text` form now re-renders itself as the cleaned-output view instead of `push(<Detail/>)`. Submit → result appears in the same screen. Esc takes you back to the history list, not all the way out. Inspired by the flat one-screen UX of `Link Cleaner` you pointed at — that flow is so clean.
- **"Clean Another" action (⌘N)** on the result view to flip back to the form without leaving the command.
- **Greptile P1: `showHUD` in `clean-clipboard.ts`** — fixed (was `showToast`, which is invisible in no-view mode). Thanks Greptile 🙏
- **Greptile P1: stale closure in "Paste from Clipboard & Clean"** — fixed by passing the clipboard text directly to the cleaner instead of relying on `setState` + `setTimeout`.
- **Greptile P2: unused `@raycast/utils` devDependency** — removed.

I kept the form-mode for `Clean Text` because it hosts the local history and the "Collapse extra spaces" toggle, but the navigation no longer dumps the user back at root. The `Clean Clipboard` no-view command is still there as the zero-friction hotkey path.

Marking ready for review. Happy to keep iterating if anything still feels off — and tak for the patient feedback! 🙏

0xdhrv (commented):
Thanks for your contribution  

- https://www.raycast.com/rhuk/clean-agent-text

---

PR #27217, opened 2026-04-15
✨ feat(espanso): add edit match action and search by replacement text

### LINKED ISSUE #22967: [Espanso] Edit Matches
### Extension

https://www.raycast.com/kud/espanso

### Description

It would be cool if i could edit a match from the "search matches". 
From the Actions Menu i could select _Edit Match_ and it would open the selected match in the _add match_ form.

Image

People who have to adjust their expansions from time to time.

### Anything else?

_No response_

kud:
## Description

Closes #22967.

Two improvements to the Espanso extension:

**Edit Match action**

Each match in the Search Matches list now has an "Edit Match" action (shortcut: `⌘E`) for matches that are editable (i.e. plain text replacements — regex, image, and package matches are excluded). Activating it opens a pre-populated form with the match's current trigger(s), label, and replacement text. On save, the YAML file is updated in-place using the YAML Document API, which preserves comments and unsupported fields such as `vars`. The list auto-refreshes after a successful edit.

A Vitest test suite with 11 cases covers the `updateMatchInFile` utility across scenarios including multi-trigger matches, matches with labels, and edge cases where the match is not found.

**Search by replacement text**

The replacement text is now included as a `keywords` prop on `List.Item`, so users can find matches by their replacement text in addition to their triggers and labels.

## Screencast

No visual changes for the search-by-replacement improvement.

The Edit Match action opens a form pre-populated with the match's current data and writes the updated values back to the YAML file on save — a screencast is not included but the flow is straightforward and can be validated using the steps below.

## Checklist

- [x] I checked that assets used in the README are located outside the metadata folder if they were not generated with our metadata tool

### DIFF
--- a/extensions/espanso/AGENTS.md
+++ b/extensions/espanso/AGENTS.md
@@ -9,6 +9,7 @@ The Espanso Raycast extension provides a search interface for Espanso text expan
 ### Core Components
 
 #### 1. Main Index (`src/index.tsx`)
+
 - **Purpose**: Primary search interface and state management
 - **Key Features**:
   - Fetches and displays Espanso matches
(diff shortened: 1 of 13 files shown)


0xdhrv (approved):
Looks good to me, approved ✅

PR #27238, opened 2026-04-16
[Skills] Add support for updating individual skills

DaleSeo:
## Description

Follow-up to #26234, which added the outdated-skill highlight but deferred the per-skill update action because `skills update` only supported updating all installed skills at once.

The skills CLI now accepts a skill name argument (https://github.com/vercel-labs/skills/issues/528), so this PR wires it up to the extension. When a user presses the update action on a single outdated skill, only that skill is updated instead of all of them.

## Screencast

2026-04-16 at 16 02 41

## Checklist

- [x] I checked that files in the `assets` folder are used by the extension's commands
- [x] I checked that assets used by the extension itself are placed outside of the `assets` folder

### DIFF
--- a/extensions/skills/CHANGELOG.md
+++ b/extensions/skills/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## [Update Individual Skills] - {PR_MERGE_DATE}
+## [Update Individual Skills] - 2026-04-17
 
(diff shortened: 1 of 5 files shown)


0xdhrv (approved):
Looks good to me, approved ✅

PR #27266, opened 2026-04-18
fix(memos): restore memo listing for newer memos instances

duxiaoxiong:
## Summary

This PR restores memo listing for newer Memos instances.

It updates the extension to:
- resolve the current user via `/api/v1/auth/me`, with a fallback to the legacy `/api/v1/auth/sessions/current` endpoint for older deployments;
- list memos with the authenticated user's resource name as the `parent` query parameter instead of parsing a numeric ID from `user.name` or using a `creator == ...` filter expression;
- remove the render loop in `MemosListCommand` by deriving the filtered list from fetched data and caching attachment markdown separately.

## Why listing failed before

The previous implementation assumed an older Memos API shape in two places:
- it fetched the current user from `/api/v1/auth/sessions/current`, which newer Memos deployments no longer expose;
- it parsed `user.name` as if it always contained a numeric row id and then rebuilt `parent=users/<id>` from that parsed value.

While investigating this against a current Memos instance, I also found that using a `creator == ...` filter expression for list requests returned an error payload instead of a memo list, whereas `parent=<authenticated user resource name>` returned the expected results. That meant the extension could silently render an empty list even with valid credentials.

On top of that, once list data was available, `MemosListCommand` kept a second derived `filterList` state and updated it from effects, which could trigger a render loop.

## What changed

- Added a compatibility wrapper for the current-user request that prefers `/api/v1/auth/me` and falls back to the legacy endpoint on 404.
- Switched memo list and recent-tag requests to use the authenticated user's resource name directly as `parent`.
- Stopped parsing a numeric user id from `user.name`.
- Removed the extra derived list state from `MemosListCommand` and replaced it with a small markdown cache keyed by memo name.
- Added compatibility tests covering:
  - auth endpoint fallback;
  - resource-name based user handling;
  - `parent`-based memo listing;
  - the absence of the render-loop pattern.

## Verification

- `node --test extensions/memos/tests/compatibility.test.mjs`
- `npm run type-check` in `extensions/memos`
- `npm run lint` in `extensions/memos`
- verified against a live Memos instance that:
  - `/api/v1/auth/me` returns the authenticated user resource name;
  - `/api/v1/memos?parent=<user resource name>` returns memo history;
  - `creator == ...` filtering does not return the expected memo list on that instance.

### DIFF
--- a/extensions/memos/CHANGELOG.md
+++ b/extensions/memos/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Send To Memos Changelog
 
-## [Fix] - {PR_MERGE_DATE}
+## [Fix] - 2026-04-20
 
 - fix: restore memo listing for newer Memos instances by resolving the current user via `/api/v1/auth/me` and listing memos with the authenticated user's `parent` resource name.
 - fix: remove the render loop in `MemosListCommand` by deriving filtered items from fetched data instead of storing a second derived list in component state.
--- a/extensions/memos/CHANGELOG.md
+++ b/extensions/memos/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Send To Memos Changelog
 
+## [Fix] - 2026-04-18
+
+- fix: restore memo listing for newer Memos instances by resolving the current user via `/api/v1/auth/me` and listing memos with the authenticated user's `parent` resource name.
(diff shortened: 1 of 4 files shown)


0xdhrv (approved):
Looks good to me, approved ✅

PR #27291, opened 2026-04-20
Add ability to move to bottom or hide target language that matches source language in quick translate

brucemakes012:
### Problem
As I'm studying Japanese right now, I often use the Quick Translate command to convert from English to Japanese and vice versa. I have English set as the first language in my quick translate output list as I usually want to see what something means in Japanese, but I often want to see how to say something in Japanese as well, and when I type in English I don't want to see it translated into English again. 

### Solution
I added an extension preference for "when target language matches source language", which defaults to "do nothing" so existing users would not be affected unintentionally. The other two options are "move target language to bottom" and "hide target language". In the former, if a target language matches the source language, then that language will be moved to the bottom of the list so the user can browse the other target languages first. The latter option hides the matching target language completely. 

### Coding changes
Changes were made to package.json to add the extension preferences. Changes to simple-translate.tx were necessary in order to get the auto-detected language before the translated text are all loaded. The remaining changes are for implementing the functionality, as well as other minor fixes like resetting the item selection to the first item in "move to bottom" mode, and preventing blank output if target list only has one language in "hide" mode. 

## Screencast

https://github.com/user-attachments/assets/805ddaab-1ea2-4517-a6a3-773702311317

Here's a demo of 1. "do nothing" with current functionality, 2. "move to bottom" mode where english and japanese were moved to the bottom depending on the source language, and 3. "hide" mode with similar test case. Thanks for reviewing this and let me know if anything needs to be changed!

## Checklist

### DIFF
--- a/extensions/google-translate/src/QuickTranslate/QuickTranslateListItem.tsx
+++ b/extensions/google-translate/src/QuickTranslate/QuickTranslateListItem.tsx
@@ -36,6 +36,7 @@ export function QuickTranslateListItem(props: {
   if (isLoading) {
     return (
       <List.Item
+        id={langTo.code}
         title={`Translating to ${langTo.name}...`}
         accessories={[
           {
@@ -58,6 +59,7 @@ export function QuickTranslateListItem(props: {
 
   return (
     <List.Item
+      id={langTo.code}
       key={langTo.code}
       title={result.translatedText}
       accessories={[
--- a/extensions/google-translate/src/QuickTranslate/QuickTranslateListItem.tsx
+++ b/extensions/google-translate/src/QuickTranslate/QuickTranslateListItem.tsx
@@ -1,7 +1,8 @@
 import React from "react";
 import { ActionPanel, List, Toast, showToast } from "@raycast/api";
+import { useEffect } from "react";
 import { usePromise } from "@raycast/utils";
-import { supportedLanguagesByCode } from "../languages";
+import { supportedLanguagesByCode, LanguageCode } from "../languages";
 import { simpleTranslate } from "../simple-translate";
 import { LanguageCodeSet } from "../types";
 import { ConfigurableCopyPasteActions, OpenOnGoogleTranslateWebsiteAction, ToggleFullTextAction } from "../actions";
@@ -12,6 +13,7 @@ export function QuickTranslateListItem(props: {
   isShowingDetail: boolean;
   setIsShowingDetail: (isShowingDetail: boolean) => void;
   setIsLoading: (isLoading: boolean) => void;
+  onDetectedSourceLanguage: (language: LanguageCode) => void;
 }) {
   let langFrom = supportedLanguagesByCode[props.languageSet.langFrom];
   const langTo = supportedLanguagesByCode[props.languageSet.langTo[0]];
@@ -32,6 +34,16 @@ export function QuickTranslateListItem(props: {
       });
     },
   });
+
+  useEffect(() => {
+    if (!result) {
+      return;
+    }
+
+    if (props.languageSet.langFrom === "auto") {
+      props.onDetectedSourceLanguage(result.langFrom);
+    }
+  }, [result, props.languageSet.langFrom, props.onDetectedSourceLanguage]);
 
   if (isLoading) {
     return (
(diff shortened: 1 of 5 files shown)


extensions/google-translate/src/quick-translate.tsx
```diff
@@ -29,16 +76,18 @@ export default function QuickTranslate(): ReactElement {
       isLoading={isAnyLoading}
       isShowingDetail={isShowingDetail}
       searchBarAccessory={<LanguageDropdown />}
+      selectedItemId={selectedItemId}
```
gebeto, line 79: do we really need to move to controlled List value? What will happen if we will remove this and all the `setSelectedItemId` logic? Isn't it going to just select first element by default all the time?

extensions/google-translate/src/QuickTranslate/QuickTranslateListItem.tsx
```diff
@@ -33,9 +35,20 @@ export function QuickTranslateListItem(props: {
     },
   });
 
+  useEffect(() => {
+    if (!result) {
+      return;
+    }
+
+    if (props.languageSet.langFrom === "auto") {
```
gebeto, line 43: use `AUTO_DETECT` variable instead of magic string

extensions/google-translate/src/quick-translate.tsx
```diff
@@ -1,22 +1,69 @@
...
+
+  const effectiveSourceLanguage = sourceLanguage === AUTO_DETECT ? detectedSourceLanguage : sourceLanguage;
+
+  const processedTargetLanguages = useMemo(() => {
+    if (effectiveSourceLanguage && targetLanguages.includes(effectiveSourceLanguage)) {
+      if (outputMatchesInput === "hide" && targetLanguages.length > 1) {
+        return targetLanguages.filter((lang) => lang !== effectiveSourceLanguage);
+      }
+
+      if (outputMatchesInput === "moveToBottom") {
```
gebeto, line 37: as I remember, raycast dev server is generating types from preferences, can you please check if there are generated enums and use it instead of magic strings?

gebeto:
UX looks a bit strange, it is flickering and as I understand it is doing duplicate translations

OVerall is it so big deal that you can just not ignore one extra translation? It is doing translation anyway to detect language, so as a result it is not doing any optimization

extensions/google-translate/src/quick-translate.tsx
```diff
@@ -29,16 +76,18 @@ export default function QuickTranslate(): ReactElement {
...
-        ? targetLanguages.map((targetLanguage) => (
+        ? processedTargetLanguages.map((targetLanguage) => (
             <QuickTranslateListItem
               key={targetLanguage}
               debouncedText={debouncedText}
               languageSet={{ langFrom: sourceLanguage, langTo: [targetLanguage], proxy }}
               isShowingDetail={isShowingDetail}
               setIsShowingDetail={setIsShowingDetail}
               setIsLoading={(isLoading) => setIsLoading(targetLanguage, isLoading)}
+              onDetectedSourceLanguage={handleDetectedSourceLanguage}
```
gebeto, line 90: to really simplify it all, according to your problem, we can just solve your problem and add single boolean property, that going to hide source language translation if needed

it will require way less code to achieve this, as you will just need to check if that preference enabled, and return `null` from `QuickTranslateListItem` if translation language equals to source language. What about this?

If somebody will need(right now nobody need it), we will be able to do some other solution for it, but overall it is unnecessary and adding way more complexity

pernielsentikaer:
Drafting until @brucemakes012 has looked into this

extensions/google-translate/src/quick-translate.tsx
brucemakes012: Thanks for reviewing the PR! Yes, that's a totally fair assessment. This was just my weekend project to solve a very niche problem of my own, and I used vscode copilot to help me so the code is probably not written very cleanly. I thought I'd try submitting a PR to see if it's a helpful addition, but I can certainly use my version as a local extension too. I'll close this PR for now, and thank you @gebeto and @pernielsentikaer for your time!

gebeto: When I'll have some spare time, Ill try to add that preferences checkbox to hide source language from translations, it should be pretty simple solution, I think it can be useful

PR #27332, opened 2026-04-21
[Skills] Fix node homebrew versioned formula resolution

### LINKED ISSUE #27297: [Skills] node: No such file or directory
### Extension

https://raycast.com/keito4/skills

### Raycast Version

Version 1.104.12

### OS Version

macOS Tahoe 26.4

#### Error:
```
Error: Command failed: 'npx' '-y' 'skills@latest' 'list' '-g' '--json'
env: node: No such file or directory

    at ChildProcess.exithandler (node:child_process:414:12)
    at ChildProcess.emit (node:events:518:28)
    at ChildProcess.emit (node:domain:489:12)
    at maybeClose (node:internal/child_process:1101:16)
    at Socket.<anonymous> (node:internal/child_process:456:11)
    at Socket.emit (node:events:518:28)
    at Socket.emit (node:domain:489:12)
    at Pipe.<anonymous> (node:net:351:12)
```

Image

### Steps To Reproduce

Manage Skills

### Current Behaviour

_No response_

### Expected Behaviour

_No response_

raulgg:
## Description

Fixes #27297

Fix two related `skills` extension issues around local Node.js path resolution.

- Add support for versioned Homebrew Node formula paths such as `/opt/homebrew/opt/node@24/bin` when building the CLI `PATH`. This fixes cases where Raycast can find Homebrew `npx` but not the matching `node` binary.
- Validate the optional `Custom npx Path` before spawning the CLI and show a clearer error when the configured path is incorrect, instead of surfacing a misleading downstream command failure.

## Checklist

### DIFF
--- a/extensions/skills/src/utils/node-path-resolver.ts
+++ b/extensions/skills/src/utils/node-path-resolver.ts
@@ -3,30 +3,17 @@ import { cpus } from "node:os";
 import { join } from "node:path";
 import semver from "semver";
 
+type VersionedPath = {
+  path: string;
+  version: string;
+};
+
 const isWindows = process.platform === "win32";
 const isMacOS = process.platform === "darwin";
 
-let cachedPaths: string | null = null;
-let cachedPathsPromise: Promise<string> | null = null;
 let cachedIsAppleSilicon: boolean | null = null;
-
-const getAppleSiliconStatus = (): boolean => {
-  if (cachedIsAppleSilicon !== null) return cachedIsAppleSilicon;
-  cachedIsAppleSilicon = cpus()[0]?.model?.includes("Apple") ?? false;
-  return cachedIsAppleSilicon;
-};
-
-const parseVersion = (version: string): number => {
-  const coerced = semver.coerce(version);
-  if (!coerced) return 0;
-
-  // Convert semver to a sortable number so directory names can be ordered newest-first
-  return coerced.major * 1000000 + coerced.minor * 1000 + coerced.patch;
-};
-
-const sortPathsByVersion = (paths: Array<{ path: string; version: string }>): string[] => {
-  return paths.sort((a, b) => parseVersion(b.version) - parseVersion(a.version)).map((item) => item.path);
-};
+let cachedPathsPromise: Promise<string> | null = null;
+let cachedPaths: string | null = null;
 
 export const pathExists = async (path: string): Promise<boolean> => {
   try {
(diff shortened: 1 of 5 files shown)


DaleSeo:
Thanks for stepping in to fix the issue, @raulgg. Since the PR is in draft, I'll take a look when it's ready for review.

raulgg:
> Thanks for stepping in to fix the issue, @raulgg. Since the PR is in draft, I'll take a look when it's ready for review.

Thank you, @DaleSeo! I was verifying different scenarios. I hope these changes will resolve the issue reported by the user. I would appreciate your comments on the fix.

extensions/skills/src/utils/skills-cli.ts
```diff
@@ -61,6 +79,37 @@ function normalizeCliError(error: unknown, npxCommand: string): Error {
...
+    if (fileStats.isDirectory()) {
+      throw new InvalidCustomNpxPathError(invalidPathMessage);
+    }
+  } catch (error) {
+    if (error instanceof InvalidCustomNpxPathError) {
+      throw error;
+    }
+
+    throw new InvalidCustomNpxPathError(invalidPathMessage);
+  }
```
DaleSeo, line 124: How about we simplify this?

```suggestion
  let fileStats;
  try {
    fileStats = await stat(customNpxPath);
  } catch {
    throw new InvalidCustomNpxPathError(invalidPathMessage);
  }
  if (fileStats.isDirectory()) {
    throw new InvalidCustomNpxPathError(invalidPathMessage);
  }
```

extensions/skills/src/utils/node-path-resolver.ts
```diff
@@ -107,21 +142,28 @@ export const resolveVersionManagerPaths = async (): Promise<string[]> => {
...
   const systemPaths = ["/usr/bin", "/bin"];
 
   if (process.env.HOME) {
     systemPaths.push(`${process.env.HOME}/.npm/bin`, `${process.env.HOME}/.yarn/bin`);
   }
 
-  const allPaths = [...platformPaths, ...versionManagerPaths, ...systemPaths, process.env.PATH || ""];
+  const allPaths = [
+    ...homebrewPaths,
+    ...homebrewVersionedNodePaths,
```
DaleSeo: Worth considering whether versioned paths should come before the default bin so explicit version choices win.

extensions/skills/src/utils/skills-cli.ts
```diff
@@ -25,14 +26,31 @@ export function isNpxResolutionError(error: unknown): boolean {
...
 function buildSkillsCliCommand(npxCommand: string, args: string[]): string {
   return [npxCommand, "-y", "skills@latest", ...args].map(shellEscape).join(" ");
 }
 
 async function runSkillsCli(args: string[]): Promise<string> {
-  const npxCommand = getCustomNpxPath() ?? "npx";
+  const customNpxPath = getCustomNpxPath();
+  if (customNpxPath) {
+    await validateCustomNpxPath(customNpxPath);
+  }
```
DaleSeo, line 51: Every call to `runSkillsCli` issues fresh syscalls for the same custom path. Since the preference value is captured at module load via `getPreferenceValues<Preferences>()`, the result won't change mid-session. A tiny memoization similar to `cachedPaths` could be helpful.

DaleSeo (approved):
Thanks for tackling this, @raulgg! I only have non-blocking suggestions!

pernielsentikaer (commented):
@raulgg do you want to have a look at this before we merge it?

extensions/skills/src/utils/skills-cli.ts
raulgg: Good suggestion! I applied this. Thank you!

extensions/skills/src/utils/skills-cli.ts
raulgg: You are correct! I have implemented a simple memoization for the custom path validation as suggested.

extensions/skills/src/utils/node-path-resolver.ts
raulgg: Yes, this is a good point. I'm also considering that we should prefer the `versionManagerPaths` over homebrew ones, as usually those would be initialized in the shell and have precedence over the versions installed through homebrew. I pushed the change for this. Thank you for raising this concern and making me reconsider my initial choice.

raulgg:
> @raulgg do you want to have a look at this before we merge it?

Sure! The feedback is addressed now.

pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #27337, opened 2026-04-22
Update microsoft-teams-calling extension

leec-666:
## Description

Prevents Teams prompting for permission every time the extension performs an action.  

- Resolves #25453
- Resolves #23710
- Resolves #21743
- Resolves #24585

Updated the New Teams API client to persist the refreshed pairing token before no-view commands exit.

The fix tracks the tokenRefresh save operation from Teams and waits for it during pairing, with a short timeout so already-paired commands stay responsive. This prevents Teams from treating every Raycast command as a new connection request after the first approval. The websocket connection URL now also uses URLSearchParams so the auth token and app identity query parameters are encoded consistently.

Also closes the websocket after direct no-view commands complete, giving those commands a cleaner lifecycle.
 
## Checklist

### DIFF
--- a/extensions/microsoft-teams-calling/CHANGELOG.md
+++ b/extensions/microsoft-teams-calling/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Microsoft Teams Changelog
 
+## [Fix Teams API 2.0 Permissions] - {PR_MERGE_DATE}
+
+- Stops Teams seeking user permission every time the extension performs an action.
+
 ## [Fix Notifications] - 2024-04-04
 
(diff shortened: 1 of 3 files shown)


0xdhrv (approved):
Looks good to me, approved ✅

montysawi:
Thank you for the update. The issue is fixed, but the "unmuted" and "muted" messages on the screen are still showing incorrectly. When muted, it is showing as unmuted and vice versa.

leec-666:
@montysawi What about the camera messages?

montysawi:
> @montysawi What about the camera messages?

Camera has the same problem. the message is wrong.

PR #27359, opened 2026-04-22
Add bash-commands extension

AlessandroCaula:
## Description

A quick access reference for bash commands. You can search across names, syntax, and descriptions, filter by category, and copy or paste commands instantly. Supports pinning favorites and tracks recently used commands.

## Screencast

bash-commands-1
bash-commands-2
bash-commands-3

## Checklist


AlessandroCaula:
@greptileai can you review it

AlessandroCaula:
Can someone review this PR? Happy to make any changes needed.

0xdhrv (commented):
Thanks for your contribution ⁣⁠‌⁣‌⁣‌­‍⁣­­‌⁠⁣­⁣⁣‍‍‍⁣⁣‌‌‍‍‍‌‍‍⁠⁠⁣ 🔥

We already have a extension  in the Store that provide similar functionality around shell/terminal command references and snippets. Could we consider enhancing an existing extension instead of creating another one?

- https://www.raycast.com/pomdtr/tldr

This PR mainly provides:

* searchable bash command cheatsheets
* categorized commands
* copy/paste actions
* pinned/recent commands
* inline notes/examples

The overall value proposition appears quite close to existing command/snippet/reference style extensions already available in the Store.

If there are unique workflows or differentiators you’re aiming for (for example: interactive execution, AI-assisted explanations, shell-specific tooling, learning flows, or dynamic/man-page integrations), we’d love to hear more and explore whether those can be integrated into an existing extension to avoid duplication and improve discoverability.

AlessandroCaula:
Hi @0xdhrv , thanks for taking the time to look at this PR.

I did actually check out the tldr extension, and I get why the comparison is there. But the way I thought about it was similar to how git-commands ([https://www.raycast.com/manumorante/git-commands](https://www.raycast.com/manumorante/git-commands)) sits alongside more general CLI reference tools. It's not trying duplicate or replace something like tldr, it's just scoped to one specific domain and built with a different goal in mind.

tldr it's pulling from a big repository covering thousands of tools. What I was going for here is a tighter, more curated set of searchable bash fundamentals commands and notes that are written to actually explain what's going on rather than just show the syntax. 
More of a "learning bash" tool than a general CLI lookup. For example in this new extension, the search works across names, commands and description, so if you don't know the command for copying a file, you can just search "copy" and find "cp". This feels more natural for someone who doesn't know or remember the exact syntax.

Also the pinning and recents features are something I added intentionally, the idea being that someone who's actively learning or regularly reaches for the same handful of commands can build up their own personalized shortlist.

I understand the concern about duplication. But I do feel like the use case is distinct enough to stand on its own. Let me know what do you think about.

0xdhrv (approved):
Looks good to me, approved ✅

Issue #66, opened 2021-10-16
[API Feature Request] Focus specific element in a Form


mooxl:
During the development of Deepcast I noticed that there is no possibility to set a focus on a specific element within a form after the render. This would simplify the input of texts to be translated, since the respective field does not have to be selected separately. 

This could be made possible by specifying the element to be focused on in the Form component, or by setting the focus to true using a boolean in a Form.x child component.

PitNikola:
Good idea, we'll consider adding it soon.

sxn:
available as of **v1.33.0** 😸

PR #27373, opened 2026-04-23
Update zen-browser extension

thomasbecker:
## Description

Bumps `@raycast/utils` from `^2.2.0` (lockfile: `2.2.1`) to `^2.2.3`.

Search Bookmarks and Search History break with "Cannot query the data, database is locked" when Zen is running. Until recently, Raycast's Node runtime didn't expose `node:sqlite`, so `useSQL` in `@raycast/utils` fell through to the `sqlite3` CLI path. The CLI's error text contains `(5)`, which matches the existing lock detector and triggers the copy-to-temp fallback — so searches worked.

Raycast now bundles Node 22.22.2, which does expose `node:sqlite`. `useSQL` switches to the native path, and its error comes back as:

```
{ errcode: 5, message: "database is locked", code: "ERR_SQLITE_ERROR" }
```

No `(5)` in the message. The detector in `@raycast/utils` 2.2.1/2.2.2 only used `message.match("(5)") || message.match("(14)")`, so the lock went undetected, the fallback never ran, and the error bubbled up to the user.

[`@raycast/utils` 2.2.3 (PR #66)](https://github.com/raycast/utils/pull/66) added `error.errcode === 5 || error.errcode === 14` to the detector, which handles the native `node:sqlite` error shape. Bumping the pin fixes the regression.

No source changes — dependency bump only, plus lockfile and changelog entry.

## Screencast

Before (Zen running, installed store version):

> `Cannot query the data  database is locked`

After (Zen running, this branch via `npm run dev`):

> Bookmarks return as expected.

## Checklist

### DIFF
--- a/extensions/zen-browser/CHANGELOG.md
+++ b/extensions/zen-browser/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 - Bumped `@raycast/utils` from `^2.2.0` to `^2.2.3` to pick up the fix for `useSQL` not falling back to a DB copy when `node:sqlite` reports `errcode: 5` ("database is locked"). Previously, searching bookmarks or history while Zen was running surfaced "Cannot query the data — database is locked" because the lock detector only matched error message text containing `(5)`/`(14)`, which the native `node:sqlite` path does not produce.
 
-## [Add Kagi Search Engine] - {PR_MERGE_DATE}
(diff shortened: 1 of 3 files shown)


0xdhrv (approved):
Looks good to me, approved ✅

Issue #26574, opened 2026-03-23
[Bitwarden Vault] Vault passwords are shown in clipboard history (Windows)


petewilkins:
### Extension

https://www.raycast.com/jomifepe/bitwarden

### Raycast Version

v0.71.0.0

### OS Version

Windows 11

### Description

Passwords are showing in the clipboard history, even with "Copy Values Temporarily" set to Always.

### Steps To Reproduce

1. Open Bitwarden Extension
2. Unlock Vault
3. Copy a password
4. See the password is stored in clipboard history

### Expected Behaviour

I expect passwords not to be shown in clipboard history.

petewilkins:
The latest update to the extension (#26997) unfortunately hasn't fixed the issue. I have reinstalled the extension but still no joy.

Would really appreciate if anyone could take a look @clins1994 @jomifepe, currently having to type in my password 20+ times a day 🤣

Just to reiterate:
- Only happens on Windows (behaviour is as expected on MacOS)
- Doesn't matter what I have the vault timeout set to
- Opening either vault or authenticator requires my master password everytime
- Doesn't matter if I'm using local Bitwraden CLI or not, same result

petewilkins:
The latest update to the extension (#27362) unfortunately hasn't fixed the issue

jomifepe:
Hey, sad to hear that nothing fixed it yet 😔 Although none of the fixes were targeting this issue.

Honestly, I'm not sure where to start with this issue, seems very setup/installation related. 

I'll try to take a look.

jomifepe:
Hey @petewilkins! Let me know if the latest update solved the issue!

I'm also trying to report an issue for the Raycast API. I've been able to reproduce this on my Windows machine, and it seems like it just doesn't load the values properly from the storage. I don't see how it could be a bug on the extension's code.

petewilkins:
Hi @jomifepe, really appreciate you looking into this. The latest update #27391 does seem to have mostly worked, it's no longer causing the vault to be locked on every open of the extension, so thank you so much for squashing that!

One bug that is still happening though is the passwords are still showing in the clipboard history (even with "Copy Values Temporarily" set to `Always`).

Do you want me to make a new GitHub issue for that and close this one? Or leave it here open?

jomifepe:
> One bug that is still happening though is the passwords are still showing in the clipboard history (even with "Copy Values Temporarily" set to Always).

Hey! Good to hear! Unfortunately that should be more or less the same issue, will look into it.

[...]

PR #27391, opened 2026-04-24
[Bitwarden] Update raycast packages

jomifepe:
## Description

While looking into #26574, I ran into a weird bug on Windows with the `LocalStorage` API. For some reason, it seems to randomly fail to load values (logs `undefined`) when they were just written to storage, without any apparent reason. I logged the value before a `setItem` and after a `getItem`, and it the value is not there.

Eventually the issue stopped happening, but I cannot associate it with the package updates, because it wasn't fixed right after updating. Although it's worth giving it a shot, might fix the issue.

Is this a known or already reported bug with the Raycast API? CC @andreaselia

## Checklist

### DIFF
--- a/extensions/bitwarden/CHANGELOG.md
+++ b/extensions/bitwarden/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Bitwarden Changelog
 
+## [Update packages] - {PR_MERGE_DATE}
+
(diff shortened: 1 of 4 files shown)


0xdhrv (approved):
Looks good to me, approved ✅

PR #27413, opened 2026-04-24
Fix aerospace window switcher search to match on window title

squiter:
## Summary

- The window switcher filter previously only matched the **beginning of the app name**, ignoring the window title entirely
- This made it impossible to find windows by their title (e.g. typing "huddle" wouldn't surface a Slack Huddle window, even though the title is visible in the list)
- The filter now checks both `app-name` and `window-title` using substring (`includes`) matching, so the full visible text in the list is searchable

## Test plan

- [ ] Open "Switch Apps in Workspace" command
- [ ] Type a word that appears only in a window title (e.g. "huddle", "issues", a document name) and confirm the window appears in results
- [ ] Confirm app-name search still works as before

### DIFF
--- a/extensions/aerospace/CHANGELOG.md
+++ b/extensions/aerospace/CHANGELOG.md
@@ -1,5 +1,9 @@
 # aerospace Changelog
 
+## [Fix] - 2026-04-24
+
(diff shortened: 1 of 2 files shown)


limonkufu:
LGTM, thanks!

pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #27436, opened 2026-04-25
Update markdown-converter extension

ewilderj:
## Description

This update adds smart clipboard detection to the "Convert to" commands and fixes email HTML conversion.

**Smart clipboard detection (fixes #26238)**
The "Convert to Google Docs", "Convert to Word", "Convert to HTML", and "Convert to Slack" commands now detect when the clipboard contains rich text instead of Markdown/Org. When rich text is detected, the command automatically round-trips it through Markdown first, then converts to the target format — so users get clean output in one step instead of silently degraded results.

Toast messages now show the conversion path taken (e.g. "Rich text → Markdown → Google Docs") so the user knows what happened.

**Email layout table handling**
Email HTML from Gmail, Outlook, etc. uses table elements for visual layout rather than data. These layout tables previously caused content loss and broken Markdown table delimiters (| --- |). The converter now detects layout tables  (role="presentation", border="0" with nested tables, th with block content) and unwraps them before conversion, preserving text, links, and formatting.

## Screencast

n/a no UI changes

## Checklist


pernielsentikaer (changes requested):
Is above something that's relevant @ewilderj 

---

ewilderj:
> Is above something that's relevant @ewilderj

Yes I think this is a valid (minor) finding, will bring it in

pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #27443, opened 2026-04-26
[Skills] Improve installed status in search, add source subtitle, and auto-refresh on install

raulgg:
## Description

The problem this addresses is that "Search Skills" command previously treated installed skills as installed by `skillId` only. That meant multiple search results could show as "Installed" even when they were technically different skills from different sources and only shared the same name. In the `react-doctor` example below, several distinct skills appeared installed while only one was.

Current:
Metadata-4

Proposed change:
skills-3
skills 2026-04-30 at 00 53 00

This is risky because the Skills CLI currently overwrites installs for skills with the same name without asking the user to confirm that they understand they are replacing a different source. The UX changes in this PR try to make that behavior visible in the extension and guide users through install decisions before they accidentally replace an existing skill.

Changes include:

- Replace the green "Installed" text tag in Search Skills with a green check-circle indicator and green skill icon to make more room in the list to display the subtitles
- Show the skill source, for example `owner/repo`, as the subtitle in both "Search Skills" and "Manage Skills" commands to distinct skill results from different sources at first glance.
- Replace name-only installed matching with source-aware matching using the installed skills list plus the metadata in `.skill-lock.json`.
- Detect when a skill with the same `skillId` is installed from a different source, show a warning indicator, and offer a "Replace Installed Skill" action to better represent the outcome of the action.
- Refresh installed indicators in "Search Skills" command immediately after installing or replacing a skill.
- Add Cmd+R refresh actions in "Manage Skills".
- Extract shared install/remove/update action handling into `withSkillAction` for more consistent confirmations, toasts, and failure handling.

## Checklist


raulgg:
@keito4 @DaleSeo I'm keeping this as a draft for now, but it's ready for you to review. I would love to hear your thoughts and feedback on the change I'm proposing before publishing the PR.

The skills CLI has a behavior I dislike when adding skills from different sources that share the same name: it replaces the old skill with the new one without any warning. This can be unexpected, and it also feels surprising in the extension. After the recent change to display the "Installed" badge in search results, I thought it would be helpful to clarify the potential outcomes when trying to install a skill that reuses the name of a currently installed one from a different source.

I would appreciate any thoughts on this proposed change, whether you think it's excessive or if you have any suggestions for improvements.

DaleSeo:
Thanks for sharing your proposal, @raulgg!

I agree with the overall direction. I'm glad you're coloring the icons of installed skills green. I was planning to make that change to align with how outdated skills are displayed in the other command.

The piece I'd push back on is squeezing the sources next the skill names, which could lead to truncation as shown in the screenshot. Skill names are what users primarily scan for, so I think ellipsing them could create more problems than the duplicate-name confusion we're trying to solve. Collisions are the minority case and can usually be avoided by typing more search terms.

Could we explore options that preserve the full skill name? I'm happy to discuss the trade-offs. Just want to make sure we don't fix one DX issue by introducing another.

@keito4 Please feel free to chime in if you have any other thoughts.

raulgg:
Thank you for your feedback @DaleSeo! This is the type of feedback I was hoping for.

You're right that truncating the skill name is far from ideal, and we must avoid that; so I'll try to find an alternative. I was really hoping to be able to show the source in the list since it's also the default behavior of the CLI when using `skills find` and the search in the skills.sh site:

image
image

Unfortunately, Raycast does not currently provide a method to ensure that the title is not truncated, at least to my knowledge.

The only compromise I can think of at this moment is to display the subtitle only when the details are hidden. However, the challenge with this approach is that we still lack a mechanism to persist the mode across sessions (we can discuss wether to add this in a later iteration is worth it or not). For my personal workflow, I would prefer to have the subtitle consistently displayed alongside the skill name, even if that means to have the detail hidden by default, as this would help me identify which skill I wish to navigate to, whether by scrolling through the list or by typing further in the search. However, until we can agree on a more effective solution, I am happy with this solution of displaying it only in full width mode, as I agree that truncating the skill name is a significant drawback.

image
image

What do you think?

DaleSeo:
@raulgg Thanks for thinking this through. Great call on the default change! I'm on board with hiding details. The UI looks cleaner and there's less chance skill names will be truncated.

Just for context, we initially made detail-on default because the list itself didn't carry enough information to identify a skill at a glance. I think your subtitle work changes that. Having the source, install count, and conflict indicators in the list view makes scanning much more efficient than inspecting. These are separate pages on skills.sh as well.

One UI tweak I'd suggest is using a single indicator instead of two. The hammer icon on the left and the check/warning icon on the right convey the same status. Dropping the right-side indicator and letting the left-side indicator carry status via icon/tint alone would free up the row and keep the install counts aligned.

If we switching the default, would it help sidestep the persistence issue? If the default fits most cases, I guess there's less pressure to save state? Correct me if I'm mistaken.

raulgg:
> @raulgg Thanks for thinking this through. Great call on the default change! I'm on board with hiding details. The UI looks cleaner and there's less chance skill names will be truncated.
>
> Just for context, we initially made detail-on default because the list itself didn't carry enough information to identify a skill at a glance. I think your subtitle work changes that. Having the source, install count, and conflict indicators in the list view makes scanning much more efficient than inspecting. These are separate pages on skills.sh as well.

Perfect! I'll implement this change then.

> One UI tweak I'd suggest is using a single indicator instead of two. The hammer icon on the left and the check/warning icon on the right convey the same status. Dropping the right-side indicator and letting the left-side indicator carry status via icon/tint alone would free up the row and keep the install counts aligned.

I actually thought about doing what you suggested, dropping the icon accessory from the list to simplify it and relying solely on the icon's color. However, I considered that having the warning icon is a good way to draw attention from the user. Also, the orange color used to represent the conflicts is the same as that used for updates in the "Manage skills" command. So I was a bit unsure about what would be best.

> If we switching the default, would it help sidestep the persistence issue? If the default fits most cases, I guess there's less pressure to save state? Correct me if I'm mistaken.

Yes, I think that persisting the mode would not be necessary in this case.

DaleSeo:
> I actually thought about doing what you suggested, dropping the icon accessory from the list to simplify it and relying solely on the icon's color. However, I considered that having the warning icon is a good way to draw attention from the user. Also, the orange color used to represent the conflicts is the same as that used for updates in the "Manage skills" command. So I was a bit unsure about what would be best.

You might be overthinking this, @raulgg? I thought we could just change both the icon and its color on the left. I picked the hammer icon, but I don't think it has much meaning. Personally, I find the red "Replace Installed Skill" button is eye-catching enough. I don't feel strongly about my suggestion, so it's your call! 😀

raulgg:
> > I actually thought about doing what you suggested, dropping the icon accessory from the list to simplify it and relying solely on the icon's color. However, I considered that having the warning icon is a good way to draw attention from the user. Also, the orange color used to represent the conflicts is the same as that used for updates in the "Manage skills" command. So I was a bit unsure about what would be best.
> 
> You might be overthinking this, @raulgg? I thought we could just change both the icon and its color on the left. I picked the hammer icon, but I don't think it has much meaning. Personally, I find the red "Replace Installed Skill" button is eye-catching enough. I don't feel strongly about my suggestion, so it's your call! 😀

I'm probably overthinking it! I'm good with changing the icon too. I'll push these changes soon.
Thank you so much for your feedback!

keito4:
@raulgg Sorry for the late reply! I agree with the direction you and @DaleSeo have discussed!!

extensions/skills/src/components/SkillDetailView.tsx
```diff
@@ -0,0 +1,209 @@
...
+function getAuditFallbackText(isLoading: boolean, availabilityState?: SkillAuditsAvailabilityState): string {
+  if (isLoading) return "Loading...";
+  if (availabilityState === "parse-error" || availabilityState === "fetch-error") return "Unable to verify";
+  return "Pending";
+}
+
+interface SkillDetailViewProps {
+  resultSkills: Skill[];
+  initialResultIndex: number;
+  onViewedSkillChange: (skillId: string) => void;
```
keito4: onViewedSkillChange, resultSkills, and initialResultIndex — resultSkills[initialResultIndex] is only used to get a single Skill, and onViewedSkillChange is declared but not destructured or used. If there's no plan for in-view result navigation, could we simplify the interface to accept skill: Skill directly and drop these three props?

extensions/skills/src/utils/with-skill-action.ts
```diff
@@ -0,0 +1,50 @@
...
+    style: Toast.Style.Animated,
+    title: toastOpts.animatedTitle,
+  });
+
+  try {
+    const result = await operation();
+    toast.style = Toast.Style.Success;
+    toast.title = toastOpts.successTitle;
+    if (toastOpts.successMessage) toast.message = toastOpts.successMessage;
+    await onSuccess?.(result);
```
keito4: onSuccess runs inside the same try/catch as operation, so if the post-install refresh (e.g. search API revalidation) throws, the success toast is hidden and replaced with "Failed to install skill" even though the install itself succeeded. Consider running onSuccess outside the try/catch, or catching its errors separately.

extensions/skills/src/components/SkillDetailView.tsx
raulgg: Good catch! I was trying to implement the navigation on the details view, but ended up causing unexpected behaviors in the list view, so I reverted it. I just didn't do the full cleanup properly 🙈

extensions/skills/src/utils/with-skill-action.ts
raulgg: Great catch! The failure to refresh shouldn't block the success toast after install. I have moved the call to `onSuccess` out of the try-catch.

0xdhrv (commented):
Hi @raulgg 👋

Thanks for the contribution 🔥

And, also look into the merge conflicts?

---

extensions/skills/src/hooks/useInstalledSkillMatches.ts
```diff
@@ -0,0 +1,65 @@
...
+
+  const installedByName = useMemo(() => {
+    const map = new Map<string, InstalledSkill>();
+    for (const record of data ?? []) {
+      map.set(normalizeSkillIdentityPart(record.name), record);
+    }
+    return map;
+  }, [data]);
+
+  const getInstalledMatch = useCallback(
```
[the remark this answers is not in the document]
raulgg: This is intentional. If the source is unknown we are unable to determine whether the searched skill is the installed one or not, so the safest choice to me seems to display the conflict warning and the "unknown source" description.

raulgg:
> Hi @raulgg 👋
> 
> Thanks for the contribution 🔥
> 
> Could you look into the suggestions from Greptile?
> 
> And, also look into the merge conflicts?
> 

Thank you @0xdhrv! I have responded to the P1 from Greptile. IMO there's no action to be taken there as that behavior was intentional. If the user installs the skill from the search results marked as conflict they would override the one they have locally from the unknown source.

I have resolved the merge conflitc.

DaleSeo (approved):
Looks good to me!

0xdhrv (approved):
Looks good to me, approved ✅

PR #27445, opened 2026-04-26
Add support for bunx on skills

kayotimoteo:
## Description
- Added initial support for `bunx`, it is only called optimally if `npx` is not found

## Checklist


extensions/skills/src/utils/skills-cli.ts
```diff
@@ -40,23 +44,54 @@ export function isInvalidCustomNpxPathError(error: unknown): boolean {
   return error instanceof InvalidCustomNpxPathError;
 }
 
-function buildSkillsCliCommand(npxCommand: string, args: string[]): string {
-  return [npxCommand, "-y", "skills@latest", ...args].map(shellEscape).join(" ");
+function buildSkillsCliCommand(runner: PackageRunner, args: string[], executable: string = runner): string {
+  const runnerArgs = runner === "npx" ? ["-y", "skills@latest"] : ["--silent", "--no-cache", "skills@latest"];
```
DaleSeo: Using `bunx --no-cache` will re-download the skill from the registry each time. This will make it feel slower than `npx -y`, which doesn't force-refresh and uses its package cache.

```suggestion
  const runnerArgs = runner === "npx" ? ["-y", "skills@latest"] : ["--silent", "skills@latest"];
```

extensions/skills/.gitignore
```diff
@@ -1,5 +1,8 @@
 *.log
 
+# Dependencies
+node_modules/
+
```
DaleSeo, line 5: This is already ignored at the repo root, so we don't need it here.

https://github.com/raycast/extensions/blob/b174ce6b8b990bc745ca1537ef2861be97423b08/.gitignore#L73-L74

DaleSeo:
Thanks for your contribution, @kayotimoteo! I left a couple of comments above. Could you also address the feedback from Greptile?

extensions/skills/.gitignore
kayotimoteo: While developing on Windows, I had an issue where node_modules was being recognized by Git, which prevented me from publishing. In other plugins I had already contributed to, it was also present, so I thought I could add it without any problems.

DaleSeo: OK, no harm in keeping it then.

DaleSeo (approved):
@kayotimoteo It looks good now, but we need to resolve the conflicts.

0xdhrv (approved):
Looks good to me, approved ✅

Issue #27273, opened 2026-04-18
[Todoist] Allow/Fix "Group by" and "Sort by" in filter views


yashpatel21:
### Extension

https://www.raycast.com/doist/todoist

### Raycast Version

1.104.12

### OS Version

MacOS Tahoe 26.3.1

### Description

When using the "group tasks by" and "sort tasks by" feature in the Todoist extension, it only works in the default views like Inbox/Today/Upcoming and not in any of my custom filters. I have certain filters that I use very often, and Todoist in its native app allows me to group and sort the tasks in any filter. The "group tasks by" and "sort tasks by" options are actually available within a filter view in the actions menu, but running those commands doesn't actually do anything. As a result I'm not sure if this is more of a bug (since the option is clearly there, it just doesn't work) or a feature request, but nonetheless, I would like to be able to group and sort tasks in my filters from within the extension.

### Steps To Reproduce

1. Create any filter in Todoist. For example, one of my filters is `(today | overdue | no date) & (@focus | no label) & !subtask`. In Todoist's native app, I usually like to group by label and sort by date for this filter.
2. Open that filter's view using the `My Tasks` todoist command in Raycast, and selecting the filter from the dropdown menu on the top right.
3. Open the actions menu and try "group tasks by" or "sort tasks by" for labels, date, etc. 

### Current Behaviour

By following "Steps to Reproduce", you should be able to see that running the sort and group actions does not do anything (unless of course this is somehow a bug or issue on my end only, and in that case I would like some assistance in troubleshooting/getting it to work on my end). Let me know if further information is required from my end.

### Expected Behaviour

"Sort tasks by" and "Group tasks by" actions within a filter view should correctly sort and group the tasks by the chosen task metadata (label, date, etc.)

PR #27446, opened 2026-04-26
Update todoist extension

yashpatel21:
## Description

Fixes a bug in the Todoist `My Tasks` filter view where **Group Tasks by** and **Sort Tasks by** actions appeared in the action menu but had no effect.

In `FilterTasks`, the view now renders the grouped/sorted output from `useViewTasks` (instead of always rendering the original fetched sections), and passes cached sync data into `useViewTasks` so grouping/sorting options that depend on metadata (projects/labels/assignees) work correctly.

This is a fix for [#27273](https://github.com/raycast/extensions/issues/27273).

## Checklist

### DIFF
--- a/extensions/todoist/CHANGELOG.md
+++ b/extensions/todoist/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [Fix group and sort actions in filter views] - {PR_MERGE_DATE}
 
 - In **My Tasks**, when a custom Todoist filter is selected, **Group tasks by** and **Sort tasks by** now update the list (they were previously no-ops because the UI ignored grouped/sorted output for filters).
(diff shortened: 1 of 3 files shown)


yashpatel21:
agreed on the P2 by Greptile. I’ve updated the default Group by path for filter views with multiple comma-separated sub-queries so we keep the original "one section per sub-query" layout, while still using useViewTasks for Group / Sort when grouping isn’t default.

I didn’t use reference equality on task objects as Greptile suggested. I bucket sortedTasks by task.id with a Set per section, so the global sort order is preserved per section in a way that’s stable if instances ever differ.

I tested this in Raycast, marking ready for review.

jakoss (approved):

0xdhrv (approved):
Looks good to me, approved ✅

PR #27451, opened 2026-04-26
[Open With App] Add Bloom support

errepuntopenco:
Thanks @fturcheti for the original extension.

## Description

Extends `Open With App` to read the user's selection from Bloom in
addition to Finder. The frontmost application is detected at runtime
via `getFrontmostApplication()` and the selection is queried from the
matching provider; Finder remains the fallback in every other case,
so existing behaviour is unchanged.

To keep it easy to add more file managers later (ForkLift, Path Finder,
etc.), selection retrieval is moved behind a small `FileManagerProvider`
interface in `src/file-managers/`. Adding a new manager is a single
class plus one entry in `registry.ts`.

### Implementation notes

- Bloom's bundle id is `com.asiafu.Bloom`.
- Bloom's scripting dictionary doesn't implement standard suite messages
  on its `windows` collection (`count`, `exists` both raise -1708), so
  the provider goes straight to `selection of front window` and lets the
  command's outer try/catch handle the no-window edge case.
- `selection of front window` returns `«class furl»` references (single
  item or list depending on selection size); the script normalizes to a
  list and coerces each item via `as alias` → `POSIX path of`, with a
  direct `POSIX path of` fallback.

## Type of change

- [x] New feature

## Checklist

- [x] I checked the linter (`npm run lint`) and code is formatted correctly
- [x] This change doesn't break existing functionality — Finder selection still works exactly as before

### DIFF
--- a/extensions/open-with-app/.gitignore
+++ b/extensions/open-with-app/.gitignore
@@ -5,9 +5,6 @@
 
 # Raycast specific files
 raycast-env.d.ts
-.raycast-swift-build
-.swiftpm
-compiled_raycast_swift
 
 # misc
 .DS_Store
--- a/extensions/open-with-app/CHANGELOG.md
+++ b/extensions/open-with-app/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Open With App Changelog
 
-## [Add Bloom support] - {PR_MERGE_DATE}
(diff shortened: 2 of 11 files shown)


0xdhrv (approved):
Looks good to me, approved ✅

PR #27471, opened 2026-04-27
Google Meet: Add profile-refocus command and fix Dia clipboard bug

artisticmedic:
## Description

- Adds a new **"Create Meet and Refocus with Specified Profile"** view command — mirrors the existing profile picker and refocuses the previous app after copying the link to the clipboard.
- Fixes the **"Couldn't copy to clipboard"** error on Dia. Dia's sdef rejects the flat `active tab of front window` form used for most Chromium browsers. Routed Dia through the nested `tell front window` form (previously only used for Arc; both browsers are from The Browser Company and share the same scripting quirk). Removed Dia from the Firefox-keystroke branch.
- Fixes **frontmost-browser detection**. The previous `lsappinfo metainfo | grep | head -1` approach returned the first supported-browser name found anywhere in the metadata dump — including background Chrome helper processes and Electron apps whose bundle paths contain "Google Chrome Framework" — misidentifying the actual frontmost browser on machines where the user's default browser was not the one lsappinfo happened to surface first. Replaced with a System Events query for the actual frontmost process name.
- Refocus failure in the profile commands no longer reports a clipboard failure when the link was already copied. Aligned the profile-list failure toast with the one used by the refocus command.

## Note

This is a re-submission of #27330, which was auto-closed after my fork was inadvertently deleted during a local cleanup. Same commit (`5466586`), no changes from the original PR.

## Screencast

(Same as the original PR.)

## Checklist

- [x] I confirmed that the extension doesn't require the `--remote-debugging-port` Chromium flag to work
- [x] I tested the experience after my changes

### DIFF
--- a/extensions/google-meet/CHANGELOG.md
+++ b/extensions/google-meet/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Google Meet Changelog
 
+## [New Command & Bug Fix] - {PR_MERGE_DATE}
+
+- Add "Create Meet and Refocus with Specified Profile" command — creates a meeting using a selected profile, copies the link, and refocuses the previous app
+- A refocus failure (e.g. the keystroke can't be sent) no longer reports a clipboard failure when the link was already copied. Also aligned the profile-list failure toast with the one used by the refocus command for consistency.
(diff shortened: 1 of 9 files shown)


artisticmedic:
Addressed in d2eb1eb — extracted the shared body into `components/ProfileCommand.tsx` parameterised by an optional `refocus` prop. Both `manual-profile.tsx` and `manual-profile-and-refocus.tsx` are now 4-line wrappers, so future tweaks to the profile picker only need to land in one place.

0xdhrv (approved):
Looks good to me, approved ✅

PR #27515, opened 2026-04-29
Add opencode extension

mormubis:
## Description

I use OpenCode across multiple projects daily and end up with hundreds of sessions. Finding a past conversation or switching to the right session means digging through terminal tabs or remembering which project it was in. I wanted something faster — a way to jump to any session from anywhere without leaving what I'm doing.

Raycast extension for managing [OpenCode](https://opencode.ai) projects and sessions. OpenCode is an open source AI coding agent for the terminal.

**Commands:**
- **Search Projects** — browse all OpenCode projects with session counts, drill into sessions grouped by working directory
- **Search Sessions** — find conversations by title or message content using multi-word scored search

**Features:**
- Detects which sessions are actively running or open in a terminal (Active/Open badges)
- Focuses the existing iTerm2 tab when resuming a session that's already open
- Supports iTerm2 and Terminal.app via preferences
- Shows helpful error when OpenCode is not installed

**How it works:**
Uses the OpenCode SDK to start a lightweight server for API access (projects, todos, messages). Falls back to SQLite for cross-project session listing and content search. Session liveness is detected by scanning running processes and checking recent database activity.

## Checklist


0xdhrv (commented):
Thanks for your contribution  

- https://www.raycast.com/mike182uk/opencode-sessions by @mike182uk

mormubis:
I checked the existing extension before submitting. They solve different problems @0xdhrv.

`opencode-sessions` is a session browser/inspector — view transcripts, see stats (cost, tokens, model), copy things, delete sessions. Mine is a session launcher — find sessions fast across projects, detect what's running, and jump into them in your terminal.

| | `opencode` (this PR) | `opencode-sessions` (existing) |
|---|---|---|
| **Data source** | OpenCode SDK server + SQLite fallback | SQLite only |
| **Session search** | Multi-word scored search over title + message content | Raycast built-in filtering (title only) |
| **Project browsing** | Dedicated command, grouped by working directory | Dropdown filter |
| **Session liveness** | Detects Active/Open via process scanning | No |
| **Resume session** | Opens terminal directly, focuses existing tab | Copy command to clipboard |
| **Terminal support** | iTerm2, Terminal.app, Warp, Ghostty, Kitty | N/A |
| **Transcript viewer** | No | Yes, with markdown rendering |
| **Session stats** | No | Cost, tokens, model, files changed |
| **AI summary** | No | Yes |
| **Delete sessions** | No | Yes |

The features are complementary. I'm open to contributing to the existing extension if that's the preferred path, but the scope and intent are pretty different. What do you think?

0xdhrv:
> I checked the existing extension before submitting. They solve different problems @0xdhrv.
> 
> `opencode-sessions` is a session browser/inspector — view transcripts, see stats (cost, tokens, model), copy things, delete sessions. Mine is a session launcher — find sessions fast across projects, detect what's running, and jump into them in your terminal.
> 
> `opencode` (this PR)	`opencode-sessions` (existing)
> **Data source**	OpenCode SDK server + SQLite fallback	SQLite only
> **Session search**	Multi-word scored search over title + message content	Raycast built-in filtering (title only)
> **Project browsing**	Dedicated command, grouped by working directory	Dropdown filter
> **Session liveness**	Detects Active/Open via process scanning	No
> **Resume session**	Opens terminal directly, focuses existing tab	Copy command to clipboard
> **Terminal support**	iTerm2, Terminal.app, Warp, Ghostty, Kitty	N/A
> **Transcript viewer**	No	Yes, with markdown rendering
> **Session stats**	No	Cost, tokens, model, files changed
> **AI summary**	No	Yes
> **Delete sessions**	No	Yes
> The features are complementary. I'm open to contributing to the existing extension if that's the preferred path, but the scope and intent are pretty different. What do you think?

Thanks for the clarification; I agree the implementations and feature sets differ, and I can see the additional work around terminal integration, liveness detection, and project navigation here.

That said, I still think this falls into the same core extension domain as `opencode-sessions`: helping users discover, manage, and resume OpenCode sessions. The workflows are adjacent enough that, from a Store perspective, this feels more like an evolution/expansion of the existing extension rather than a separate category of tool.

The Extensions Guidelines generally push us toward consolidating related functionality where possible, even when the implementations or UX approaches differ substantially. The concern is less “do these have identical features?” and more “would users reasonably expect these capabilities to live in one extension?”.

In this case, I think the answer is probably yes:

* both extensions operate on the same underlying OpenCode session data
* both center around session discovery and interaction
* the differences are mostly in *how* users interact with sessions afterward (inspect vs resume/manage)

Features like:

* terminal launching
* active-session detection
* improved search/navigation
* project grouping

all feel like capabilities that could strengthen the existing extension rather than justify a parallel one in the Store.

I understand that integrating this into the current extension may require significant restructuring or coordination with the maintainer, but that’s generally the direction we prefer when the primary problem space overlaps this closely.

I’d recommend reaching out to `opencode-sessions`’ maintainer to explore collaboration and whether these workflows can be merged into the existing extension instead of introducing a second OpenCode session manager.

PR #27517, opened 2026-04-29
[Skills] Fix bunx error on windows

### LINKED ISSUE #27509: [Skills]...
### Extension

https://raycast.com/keito4/skills

### Raycast Version

v0.55.0.0

### OS Version

Windows 11 Pro / 10.0.26200 Build 26200

#### Error:
```
Error: Command failed: "bunx" "--silent" "skills@latest" "add" "greptileai/skills@greploop" "-g" "-a" "claude-code" "cursor" "opencode" "-y"
'"bunx"' is not recognized as an internal or external command,
operable program or batch file.

    at ChildProcess.exithandler (node:child_process:417:12)
    at ChildProcess.emit (node:events:519:28)
    at maybeClose (node:internal/child_process:1101:16)
    at ChildProcess._handle.onexit (node:internal/child_process:304:5)
```

### Steps To Reproduce

Try to install Skills on Windows with custom NPX path set

### Current Behaviour

_No response_

### Expected Behaviour

_No response_

DaleSeo:
## Description

Fixes #27509.

The Skills extension is supposed to try `bunx` first and then silently switch to `npx` if Bun isn't installed. However, on Windows, the fallback never happened. The shell-escape helper wraps every argument in double quotes for `cmd.exe`, so when `bunx` is missing, the error shows up as `'"bunx"' is not recognized as an internal or external command`. But the resolution-failure matcher only checked for the unquoted version, which is `'bunx' is not recognized...`, so the error went to the user instead of triggering the `npx` fallback. This change adds the cmd.exe-quoted version to the matcher, ensuring the fallback works as it should.

## Checklist

### DIFF
--- a/extensions/skills/CHANGELOG.md
+++ b/extensions/skills/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Skills Changelog
 
+## [Fix Windows bunx Fallback] - {PR_MERGE_DATE}
+
+- Fix the Skills CLI throwing `'"bunx"' is not recognized as an internal or external command` on Windows when Bun is not installed; the extension now correctly falls back to `npx` as intended
+
 ## [Add bunx support] - 2026-04-28
 
 - Added initial support for `bunx`, it is only called optimally if `npx` is not found
--- a/extensions/skills/CHANGELOG.md
+++ b/extensions/skills/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Skills Changelog
 
-## [Fix Windows bunx Fallback] - {PR_MERGE_DATE}
+## [Fix Windows bunx Fallback] - 2026-04-30
 
 - Fix the Skills CLI throwing `'"bunx"' is not recognized as an internal or external command` on Windows when Bun is not installed; the extension now correctly falls back to `npx` as intended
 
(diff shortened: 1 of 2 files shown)


raulgg (approved):
Looks good to me! Thank you for the fix!

raulgg:
Thank you for the fix @DaleSeo!

I just realized now about the change to include `bunx` as a package runner in the extension. I just resolved, I hope, an issue I was having with GitHub not notifying me properly because I missed the PR that I was reviewing now, and I would have loved to leave some feedback there.

Not to be tackled in this PR, but here are some thoughts: #27445 stated that `bunx` would be used if `npx` was unavailable, but that is not the case. The fallback chain is: custom npx path => bunx => npx. In my opinion, we should not force `bunx` as the primary package runner and should always use `npx` first. However, we can add a preference to the extension to prioritize `bunx`, in which case we could reverse the fallback chain.

Additionally, the errors now can feel like they push `bunx` even for users with `npx`. I can tackle that; I already started working on fixing error message instructions for Windows, so they don't feel too macOS specific.

Regardless, that's unrelated to the current fix, which I already approved. I just wanted to engage in some discussion over those concerns.

DaleSeo:
@raulgg Thanks for sharing your thoughts. One thing worth flagging on reversing the fallback chain is that most users who have `bunx` also have `npx` installed because Bun users typically keep Node around too, so using `npx` first would mean `bunx` is essentially never reached in practice.  That makes the preference toggle you mentioned seem like the real solution to me.

0xdhrv (approved):
Looks good to me, approved ✅

PR #27538, opened 2026-04-30
[Browser Bookmarks] Add Windows support for Chromium browsers

haneul486:
## What Changed

- added Windows support for Chromium-based bookmarks
- added support for Chrome, Edge, and Brave on Windows
- added automatic bookmark refresh when Chromium bookmark files change
- preserved existing macOS browser identifiers for storage compatibility

## Testing

- npm run lint
- npm run build
- tested in Raycast for Windows
- verified Chrome bookmark search and open flow
- verified bookmark title updates are picked up after file changes

### DIFF
--- a/extensions/browser-bookmarks/CHANGELOG.md
+++ b/extensions/browser-bookmarks/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Browser Bookmarks Changelog
 
+## [Windows Support for Chromium Browsers] - {PR_MERGE_DATE}
+
+- Added Windows support for Chrome, Edge, and Brave bookmarks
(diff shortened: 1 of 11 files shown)


haneul486:
Addressed the current review suggestions in commit `1243eb3`.

Updated:
- tightened Windows app-name matching to exact display-name equality while keeping executable-path matching
- wrapped the Chromium refresh `mutate` function with `useCallback` and added it to the polling effect dependencies
- avoided rewriting the stored `browsers` value unless legacy ID normalization actually changes it

Also re-ran `npm run lint` and `npm run build` after the changes.

0xdhrv (approved):
Looks good to me, approved ✅

PR #27544, opened 2026-04-30
Update proton-pass extension

### LINKED ISSUE #27511: [Proton Pass] Not searching secrets in all Vaults
### Extension

https://www.raycast.com/izyuumi/proton-pass

### Raycast Version

1.104.13

### OS Version

26.5

### Description

When using the "search items" commands only items from one vault are shown, even if "All Vaults" is selected at the top.
Both vaults are listed in the list but the other one is considered empty when selecting it.

Using pass-cli v2.0.2 (installed with the official curl snippet).

### Steps To Reproduce

1. Using "Search items" command
2. Item list is only showing one Vault even if "All Vaults" is selected
3. The drop-down shows all my Vaults, but only one of it has items in it when changing.

### Current Behaviour

_No response_

### Expected Behaviour

- "All Vaults" should list all items across all vaults.
- Changing the Vault in drop-down should switch the vault correctly and list its items.

izyuumi:
## Description

Fixes #27511.

The Proton Pass `Search Items` command could show items from only one vault even when "All Vaults" was selected. The vault list itself loaded correctly, but item loading is performed per vault with `pass-cli item list --share-id`. Some scoped item-list responses can omit an item-level `share_id`; the extension was treating that as invalid item data, dropping that vault's results during the all-vault aggregation.

This change extracts the `pass-cli` JSON normalization into a small testable module and uses the requested vault `shareId` as a fallback when normalizing scoped item-list and item-detail responses. It also keeps the per-vault error log useful by including the error message if another vault fails to load.

## Validation

- `npm test`
- `npm run build`
- `npm run lint`
- `npm run publish`

## Screencast

N/A - bug fix for CLI JSON normalization and vault aggregation behavior.

## Checklist


pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #27565, opened 2026-04-30
Inspire Search: add Copy URL to Clipboard action

errepuntopenco:
## Description

Mirrors the URL opened by ⌘↩ so users can copy the record link without leaving Raycast. Bound to ⌘C.


0xdhrv (approved):
Looks good to me, approved ✅

PR #27568, opened 2026-04-30
Fix Clean Agent Text clipboard read on Windows

### LINKED ISSUE #27566: Clean Agent Text
### Extension

https://www.raycast.com/rhuk/clean-agent-text

### Raycast Version

v0.56.1.0

### OS Version

Windows 11

### Description

Instead of outputting cleaned text, it outputs a file name. 

### Steps To Reproduce

1. Select Text from Gemini in browser (ms edge)
2. Control-C, 
3. Alt-Space to launch Raycast
4. type CAT then press Enter. Message confirms text is cleaned. 
5. open notepad window. 
6. Control-V. 

windows 11

### Current Behaviour

swaps clipboard text with a single file name. 

### Expected Behaviour

paste original text without all of the extra lines.

rhukster:
## Description

Fixes [#27566](https://github.com/raycast/extensions/issues/27566) — on the Raycast Windows beta, copying a selection from a Chromium browser (Edge) and running Clean Agent Text was replacing the clipboard with a single filename instead of the cleaned text.

The selection from a browser puts multiple formats on the OS clipboard at once: plain text, HTML, and often a file reference for any inline images in the selection. `Clipboard.readText()` on the Windows beta appears to be returning that file reference rather than the plain text slot, so the cleaner was running on a filename and writing it back.

This change:

- Switches from `Clipboard.readText()` to `Clipboard.read()` and uses the `text` field explicitly.
- Adds a clearer HUD message ("Clipboard contains a file, not text") when the clipboard genuinely only holds a file, instead of silently writing a filename back.

No behavioral change on macOS — verified locally.

## Checklist

### DIFF
--- a/extensions/clean-agent-text/CHANGELOG.md
+++ b/extensions/clean-agent-text/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Clean Agent Text Changelog
 
+## [Fix Windows Clipboard Reading] - {PR_MERGE_DATE}
+
+- Read the clipboard's text slot explicitly so a file reference on the clipboard (common when copying from web pages on Windows) no longer gets pasted back as a filename
+- Show a clearer HUD message when the clipboard holds a file instead of text
+
 ## [Initial Version] - 2026-04-15
 
 - Clean box-drawing characters from clipboard text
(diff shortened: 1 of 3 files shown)


rhukster:
still waiting on your approval

rhukster:
Good process! updates closed because you didn't get to them.

PR #27583, opened 2026-05-01
Fix VS Code recent projects after storage changes

kzheart:
## Description

Fixes the Visual Studio Code Recent Projects command when newer VS Code versions no longer keep newly opened folders in the `history.recentlyOpenedPathsList` global storage entry.

The command now keeps the existing SQLite-backed fast path, then performs a background sync from VS Code's `workspaceStorage/*/workspace.json` files. Missing entries are merged into `history.recentlyOpenedPathsList` and persisted, so subsequent launches keep using the existing recent-projects path.

## Testing

- `npm run lint`
- `npm run build`


0xdhrv (commented):
Hey @kzheart 👋

Thanks for your contribution!

Also, could you look into the suggestions from Greptile

---

PR #27598, opened 2026-05-02
Remove fingertip extension

mblode:
## Summary

- Remove the `extensions/fingertip` Raycast extension directory.
- Remove the matching `.github/CODEOWNERS` entry and `.github/extensionName2Folder.json` mapping.

## Validation

- `node -e "JSON.parse(require('fs').readFileSync('.github/extensionName2Folder.json','utf8')); console.log('JSON ok')"`

## CI note

The `Extensions / Check` workflow currently fails for full extension-directory deletion PRs because Raycast's action derives `extensions/fingertip/` from the deleted files and then tries to `cd` into the removed directory.

This appears to be the same deletion-workflow limitation seen on the merged deletion PR https://github.com/raycast/extensions/pull/25841.


pernielsentikaer (changes requested):
Hi 👋

First of all, why does this need to be removed? I didn't test it 🙂

We need to do this in multiple steps. Otherwise, the extension will get stuck on a stale version for everyone who has it installed.

Step 1 is to trim the extension down to a minimum by showing an EmptyView saying that it's deprecated for around a month so as many people as possible will see it, and then we can look into removing it from the store

---

pernielsentikaer:
Did you see my feedback here @mblode

pernielsentikaer:
Closed in favor of #28707

PR #27612, opened 2026-05-03
Add gemini-tts extension

semantic-craft:
## Description

Adds **Gemini TTS**, a Raycast extension for reading selected macOS text aloud with Gemini text-to-speech.

This extension is optimized for academic and long-form listening rather than generic one-shot TTS:

- Direct Gemini REST API integration with `gemini-3.1-flash-tts-preview` and `gemini-2.5-flash-preview-tts`
- Quick Read from selected text, with clipboard fallback
- Resume, restart, stop, playback speed controls, and menu-bar reading status
- Gemini-native prompt structure using Audio Profile, Scene, Director's Notes, and Transcript sections
- Smart Auto reading mode that chooses Legal Text Mode, English Paper Reader, Mandarin Lecture, or Bilingual Academic Reader based on selected text
- Mixed Chinese / English guidance so Chinese is read as Mandarin while embedded English terms, citations, names, and acronyms remain English
- Smart Academic Pauses using safe `[short pause]` audio tags between paragraphs
- Voice picker for Gemini's 30 prebuilt voices, with academic recommendations highlighted
- README guidance for Google AI Studio API key creation, free-tier quota, AI Studio Billing, and the distinction between Google AI Pro benefits and Gemini API project billing
- Rainbow scholarly-speaker icon with light/dark assets

Voice cloning is intentionally not included because the Gemini TTS API currently provides prebuilt voices rather than a voice-clone endpoint.

## Screencast

No screencast is included because the extension requires a user-provided Gemini API key. The UI is built from standard Raycast commands and lists. I validated the extension locally in Raycast development mode and through the Raycast CLI checks below.

## Validation

- `npm run build`
- `npm run lint`
- `npx tsc --noEmit`
- `git diff --check`
- Ran `npm run dev` / `ray develop` locally and confirmed the extension builds and loads in Raycast.

## Checklist


extensions/gemini-tts/assets/command-icon.source.svg
0xdhrv, line 1: **nitpick:** unused SVG file

This looks like it may no longer be needed. If it's safe to remove, deleting it could help keep it cleaner.

extensions/gemini-tts/assets/command-icon@dark.png
0xdhrv, line 1: **suggestion (non-blocking):** Would it make sense to use the same extension icon for both dark and light mode here?

This seems like the two icons are identical, so keeping both may not add much value.

extensions/gemini-tts/assets/command-icon@dark.source.svg
0xdhrv, line 1: **nitpick:** unused SVG file

This looks like it may no longer be needed. If it's safe to remove, deleting it could help keep it cleaner.

extensions/gemini-tts/src/api/types.ts
```diff
@@ -0,0 +1,102 @@
+export type GeminiTTSModel =
+  | "gemini-3.1-flash-tts-preview"
+  | "gemini-2.5-flash-preview-tts"
+  | "gemini-2.5-pro-preview-tts";
+export type GeminiLanguageMode = "auto" | "cmn" | "en" | "mixed-cmn-en";
```
0xdhrv, line 5: **question:** multiple language support

Is it possible to support multiple languages here?

0xdhrv (changes requested):
Hey @xwzhangSZU 👋 

---

extensions/gemini-tts/assets/command-icon.source.svg
semantic-craft: Done in `50cd3c84` — the manifest and README only reference `command-icon.png`, the source SVG was an editable artifact from the design pass.

extensions/gemini-tts/assets/command-icon@dark.png
semantic-craft: Consolidated in `50cd3c84` — the dark variant only differed by one or two background shades, so I removed it and Raycast falls back to `command-icon.png` for both modes.

extensions/gemini-tts/assets/command-icon@dark.source.svg
semantic-craft: Done in `50cd3c84` — removed alongside the light source SVG.

extensions/gemini-tts/src/api/types.ts
semantic-craft: The extension already handles multiple languages. `languageMode` defaults to `auto`, which passes text straight through to Gemini 3.1 Flash TTS Preview (the extension's default model) — that model supports 70+ languages with automatic input-language detection, so pasted text in any supported language is spoken correctly without extra config.

The `cmn` / `en` / `mixed-cmn-en` values aren't a language whitelist — they're optional delivery hints that bias pronunciation and pacing for the extension's primary Chinese + English legal-academic use case (citations, mixed-script handling, code-switching). I kept the explicit list narrow on purpose since `auto` plus the model's built-in detection already covers the rest. Happy to broaden it if per-language hints would be valuable.

semantic-craft:
Hi @0xdhrv — addressed your review and ready for another look.

- `50cd3c84` — assets cleanup (your 3 nitpicks/suggestion) + replaced the MiniMax-leftover metadata screenshots with real Gemini TTS captures (the latter was also Greptile-flagged on earlier rounds)
- `20c863fc` — separate fix discovered while regenerating the screenshots: `Read with Voice Selection` could deadlock on a same-view re-read because `stopExternalPlayback` treated this process's own session lock as an external holder, so `waitForSessionLockRelease` waited on a lock we held → permanent "Stopping previous reading" toast. Released our own lock first; also cleared STOP_FILE in the cleanup `finally` (Greptile's prior point).

Inline replies above. Thanks for testing!

semantic-craft:
Pushed `d60e8c83` — pure follow-up that closes the remaining Greptile P1 items the earlier rounds left open. No reviewer-facing logic changed, just hardening:

- **`restart-reading.tsx` / `resume-reading.tsx`** — added the `waitForSessionLockRelease()` guard that `read-with-voice.tsx` already uses, so the running session actually observes STOP_FILE before the new command clears it. Previously the stop signal lived ~0 ms (back-to-back sync calls) and the new `playReadingSession` returned "Another reading is already in progress".
- **`reading-runner.ts`** — moved `writePlaybackSpeed()` and the stopPoll `setInterval` inside the `try` block, so an early throw at session start can't leak the lock and block every subsequent reading until Raycast restarts.
- **`gemini-tts.ts`** — `shouldRetry()` now also retries transient `fetch` failures (ECONNRESET / ETIMEDOUT / EAI_AGAIN / ENOTFOUND / EPIPE / undici socket errors / `TypeError: fetch failed`), not just HTTP 429/5xx. A single dropped packet during a long reading no longer aborts the session.
- **`CHANGELOG.md`** — dropped the misleading "Removed" section (Greptile P2) since the extension never shipped MiniMax features to end users; documented the three fixes above.

One Greptile suggestion intentionally **not applied**: the "title case" P2 asking to rewrite `"Read with Voice Selection"` → `"Read With Voice Selection"` and `"Speed up Reading"` → `"Speed Up Reading"`. Raycast's own linter (`ray lint`) actively warns *against* that form and expects the AP-style casing (short prepositions stay lowercase), so I left the titles as-is to keep `ray lint` clean.

`ray build` is green on the new HEAD. Ready for another look @0xdhrv — thanks!

extensions/gemini-tts/package.json
```diff
@@ -0,0 +1,420 @@
...
+      "mode": "no-view"
+    },
+    {
+      "name": "speed-up-reading",
+      "title": "Speed up Reading",
+      "subtitle": "Gemini TTS",
+      "description": "Increase reading speed by 0.25× for the next segment.",
+      "mode": "no-view"
+    },
+    {
```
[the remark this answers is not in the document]
semantic-craft: Not applying this one — Raycast's official `ray lint` warns *against* this form and expects `"Read with Voice Selection"` / `"Speed up Reading"` (AP-style: short prepositions like "with" / "up" stay lowercase). Reverted the title-cased variants I briefly had locally so `ray lint` stays clean. Heads-up in case the lint rule used here is older than the current Raycast convention.

0xdhrv (approved):
Looks good to me, approved ✅

Issue #13382, opened 2024-07-09
[Google Chrome Profiles] ...


alebialowas:
### Extension

https://www.raycast.com/frouo/google-chrome-profiles

### Description

Hi, I've tried your extension, and when creating a quicklink for a profile (through the app), this quick link opens a blank tab (okay, on the right profile), but I believe that's a bug. Also, it would be great if you could implement the possibility of creating quicklinks for certain bookmarks within a profile.

Cheers

Issue #5656, opened 2023-03-29
[Google Chrome Profiles] ...


daverad:
### Extension

https://www.raycast.com/frouo/google-chrome-profiles

### Description

I'd like to be able to switch chrome profile via the extension, without it loading a new tab. (If that is already possible apologies, just couldnt figure it out >.<)

thanks!

### Who will benefit from this feature?

anyone who uses this extension to frequently switch profiles without the need to create a new tab or new search

frouo:
Hello, unfortunately it is not possible :( At least I didn't find a way to for the moment and time available.

PR #27615, opened 2026-05-03
google-chrome-profiles: fix silent failure via detached osascript spawn, fix bookmark crash

drye:
## Description

Updated 2026-05-12: this PR has been rewritten end to end. The earlier description below the line is no longer accurate. See [my reply](https://github.com/raycast/extensions/pull/27615#issuecomment-4427796551) for the full investigation.

### TL;DR

Two real bugs in `google-chrome-profiles`:

1. **All profile actions silently no-op for users who haven't already granted Raycast AppleEvents permission for `System Events.app`.** `runAppleScript` from `@raycast/utils` spawns `osascript` without `detached: true`, so it inherits the extension's Node process group. Raycast tears that group down ~40ms after the action handler returns to React, killing `osascript` mid-flight before the script can drive Chrome. The first-run TCC prompt for `kTCCServiceAppleEvents → com.apple.systemevents` is queued on a future tick and gets cancelled along with the subprocess, so the user never sees the dialog and has no path to grant the permission.
2. **`TypeError: Invalid URL` crash in the bookmark list** when any bookmark has an opaque-origin scheme (`chrome://`, `about:`, etc.), because `getFavicon(new URL(url).origin)` is `null` for those.

### Fix

- **Keep** the AppleScript Profiles-menu design from #24147 intact (the bring-to-front + tab-focus UX is the right one; nothing else preserves it without an extra `about:blank` tab on already-running Chrome).
- **Route every AppleScript through a small `runDetachedAppleScript` helper** that writes the script to a temp file and spawns `osascript` with `{ detached: true, stdio: "ignore" }` + `child.unref()`. The subprocess survives Raycast's teardown, the TCC prompt renders, the script runs to completion. Spawn errors and the child's `error` event surface via `showToast` instead of being swallowed.
- **Same treatment for the `newWindow` path** (which uses `do shell script` for Chrome flags) — it was failing for the same lifecycle reason.
- **Favicon guard**: only resolve favicons for `http(s)` bookmarks; use `Icon.Globe` otherwise.

### Reproduction

Reproduced on a fresh macOS user with no prior Raycast → System Events grant. With file-based logging and 200ms heartbeats: 24 attempts across `focus`, `newTab`, `newWindow` for two profiles, every one logging the script-about-to-run, then `process:exit code=0` at ~38ms. Zero successes, zero errors, zero heartbeats, no TCC dialog. Same AppleScript run from a terminal `osascript` completes in 836ms and focuses Chrome correctly, confirming the script itself is fine and the failure is purely the subprocess lifecycle.

### Behaviour after the fix

- `Bring to Front` focuses the existing window in the target profile, no extra tab.
- `New Tab` opens `chrome://newtab/` in the target profile.
- `New Window` opens a new window in the target profile via `--profile-directory`.
- `Open URL` opens the URL in the target profile.
- On a system without the System Events grant, the macOS TCC dialog now actually renders on first action; once granted, all subsequent actions work.
- Bookmark list opens cleanly with `chrome://` and `about:` bookmarks present.

Tested on macOS 25.4, two profiles, Slovenian system locale, English Chrome menu.

### Upstream

`@raycast/utils.runAppleScript` arguably should use `detached: true` by default, or at minimum document the teardown gotcha. I'll open a separate issue on `raycast/utils` once this lands.

## Screencast

n/a, behavioral fix. Visible difference: actions now reliably hit the right profile, and the first-run macOS permission dialog actually appears.

## Checklist

### DIFF
--- a/extensions/google-chrome-profiles/CHANGELOG.md
+++ b/extensions/google-chrome-profiles/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Google Chrome Profiles Changelog
 
+## [Fix] - {PR_MERGE_DATE}
+
+- Fix silent failure of all profile actions (Bring to Front, New Tab, New Window, Open URL) for users who have not previously granted Raycast `AppleEvents` permission for `System Events.app`. `@raycast/utils.runAppleScript` spawns `osascript` without `detached: true`, so it inherits the extension's Node process group. Raycast tears that group down ~40ms after the action handler returns control to React, which kills `osascript` mid-flight and also cancels the asynchronous TCC permission prompt that macOS tries to render on first run, leaving no path for the user to actually grant the permission. Run AppleScript via a detached `child_process.spawn("/usr/bin/osascript", [...], { detached: true, stdio: "ignore" })` + `child.unref()` so the subprocess survives teardown, the TCC prompt renders, and the script runs to completion.
+- Fix bookmark favicon crash on `chrome://` / `about:` URLs: `new URL(...).origin` is `null` for opaque-origin schemes; passing that to `getFavicon` threw `TypeError: Invalid URL` and broke the bookmarks list. Only resolve favicons for `http(s)` bookmarks; use the globe icon for everything else.
+
 ## [Feature] - 2026-04-08
 
 - Add "New Window" action to open a new Chrome window for a profile
--- a/extensions/google-chrome-profiles/CHANGELOG.md
+++ b/extensions/google-chrome-profiles/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Google Chrome Profiles Changelog
 
-## [Fix] - {PR_MERGE_DATE}
+## [Fix] - 2026-05-26
 
 - Fix silent failure of all profile actions (Bring to Front, New Tab, New Window, Open URL) for users who have not previously granted Raycast `AppleEvents` permission for `System Events.app`. `@raycast/utils.runAppleScript` spawns `osascript` without `detached: true`, so it inherits the extension's Node process group. Raycast tears that group down ~40ms after the action handler returns control to React, which kills `osascript` mid-flight and also cancels the asynchronous TCC permission prompt that macOS tries to render on first run, leaving no path for the user to actually grant the permission. Run AppleScript via a detached `child_process.spawn("/usr/bin/osascript", [...], { detached: true, stdio: "ignore" })` + `child.unref()` so the subprocess survives teardown, the TCC prompt renders, and the script runs to completion.
 - Fix bookmark favicon crash on `chrome://` / `about:` URLs: `new URL(...).origin` is `null` for opaque-origin schemes; passing that to `getFavicon` threw `TypeError: Invalid URL` and broke the bookmarks list. Only resolve favicons for `http(s)` bookmarks; use the globe icon for everything else.
--- a/extensions/google-chrome-profiles/src/index.tsx
+++ b/extensions/google-chrome-profiles/src/index.tsx
@@ -334,7 +334,11 @@ function ListBookmarks(props: { profile: Profile; browser: BrowserConfig }) {
               key={index}
               title={b.title}
               subtitle={b.subtitle}
-              icon={getFavicon(b.iconURL, { fallback: Icon.Globe, mask: Image.Mask.Circle })}
+              icon={
+                b.iconURL
+                  ? getFavicon(b.iconURL, { fallback: Icon.Globe, mask: Image.Mask.Circle })
+                  : { source: Icon.Globe, mask: Image.Mask.Circle }
+              }
               actions={
                 <ActionPanelForTarget
                   profile={props.profile}
--- a/extensions/google-chrome-profiles/src/util/util.ts
+++ b/extensions/google-chrome-profiles/src/util/util.ts
@@ -209,7 +209,16 @@ export const escapeAppleScriptString = (str: string): string => {
  */
 const runDetachedAppleScript = (script: string): void => {
   const scriptPath = join(tmpdir(), `raycast-google-chrome-profiles-${randomUUID()}.applescript`);
-  writeFileSync(scriptPath, script);
+  try {
+    writeFileSync(scriptPath, script);
+  } catch (writeError) {
+    showToast({
+      style: Toast.Style.Failure,
+      title: "Could not write script file",
+      message: String(writeError),
+    });
+    return;
+  }
 
   let child;
   try {
--- a/extensions/google-chrome-profiles/src/util/util.ts
+++ b/extensions/google-chrome-profiles/src/util/util.ts
@@ -1,6 +1,10 @@
 import { URL } from "url";
-import { confirmAlert, Icon } from "@raycast/api";
-import { runAppleScript } from "@raycast/utils";
+import { writeFileSync, unlinkSync } from "fs";
+import { tmpdir } from "os";
+import { join } from "path";
+import { spawn } from "child_process";
+import { randomUUID } from "crypto";
+import { showToast, Toast } from "@raycast/api";
 import { BrowserConfig } from "./types";
 
 export type ChromeTarget =
@@ -17,13 +21,21 @@ export const ChromeAction = {
 };
 
 export const createBookmarkListItem = (url: string, name?: string) => {
-  const urlOrigin = new URL(url).origin;
   const urlToDisplay = url.replace(/(^\w+:|^)\/\//, "");
+  let iconURL: string | undefined;
+  try {
+    const parsed = new URL(url);
+    if (parsed.protocol === "http:" || parsed.protocol === "https:") {
+      iconURL = parsed.origin;
+    }
+  } catch {
+    // opaque or invalid URL; fall through to globe icon
+  }
   return {
     url: url,
     title: name ? name : urlToDisplay,
     subtitle: name ? urlToDisplay : undefined,
-    iconURL: urlOrigin,
+    iconURL,
   };
 };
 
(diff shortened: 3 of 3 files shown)


frouo:
Thank you very much @drye for your contribution!

We recently switched from the CLI approach to AppleScript to improve the overall user experience (see https://github.com/raycast/extensions/pull/24147). With the CLI approach, we weren’t able to bring to front an existing window without the opening a new tab trick (`about:blank`). This led to several feature requests and bug reports in the past, for example:
- https://github.com/raycast/extensions/issues/5656
- https://github.com/raycast/extensions/issues/13382 (“opens a blank tab (okay, on the right profile), but I believe that's a bug”)
- and others

You’re absolutely right that assuming “Profiles” is always the 8th menu item is fragile. If I understand correctly, you ran into the issue where the extension couldn’t find the “Profiles” menu in your version of Chrome, right?

I’ve personally been using this extension since day1 with the CLI approach, and I can say the AppleScript version has been a real UX improvement. In particular, it allows us to:

- properly bring an existing window to the front (without opening a blank tab)
- focus an already opened tab if the URL matches

Do you think we could make the AppleScript more robust so it can reliably find the “Profiles” menu in your setup?

0xdhrv:

frouo:
@drye, did you have time to look at my reply? If the AppleScript doesn't work on your computer, I would love to benefit from your developer skills to investigate if there is an AppleScript fix.

In the meantime, I had another idea: we could keep both methods: `openGoogleChromeViaAppleScript` (default) and `openGoogleChromeViaCli` (fallback + make it default via Preferences).

drye:
@frouo you were right. Spending a morning reproducing my original "none of the actions work" against the pre-PR code, the real bug is not where I thought it was, and your AppleScript design from #24147 is the right one once the bug that was hiding it goes away.

Here is what I found. The action handler eventually calls `runAppleScript(...)` from `@raycast/utils`. That helper spawns `osascript` via Node's `child_process.spawn` without `detached: true`, so the subprocess inherits the extension's Node process group. Raycast tears that group down around 40ms after `onAction` hands control back to React, regardless of whether the handler awaits the Promise. The exit is clean, code 0, no signal. Any `await` after that point quietly never resolves and the action looks like it did nothing.

For first-time users that has a second nasty consequence. When the AppleScript runs `tell application "System Events" to tell process "Google Chrome" ...`, macOS triggers an asynchronous TCC prompt, `"Raycast.app wants access to control System Events.app"`, which is the `kTCCServiceAppleEvents` row for target `com.apple.systemevents` (separate from the one for `com.google.Chrome` we already had). That prompt is rendered on a future tick. By the time it would appear, the `osascript` subprocess has already been killed along with Node, the prompt is cancelled, and the user never sees a dialog. Granting that permission from inside the extension is impossible because the extension dies before the prompt can be answered.

I instrumented the pre-PR `util.ts` with file logging and 200ms heartbeats and reproduced this 24 times across `focus`, `newTab`, and `newWindow` for two profiles. Every attempt logs the script about to run, then `process:exit code=0` at around 38ms. Zero successes, zero errors, zero heartbeats. The exact same AppleScript run from a terminal `osascript` completes in 836ms and focuses Chrome correctly, so the script itself is fine. My TCC state confirmed Raycast has `AppleEvents` allowed for `com.google.Chrome` since early 2025 but no row for `com.apple.systemevents`.

To confirm, I bypassed `runAppleScript` in the dev build and instead used:

```ts
const child = spawn("/usr/bin/osascript", [scriptPath], { detached: true, stdio: "ignore" });
child.unref();
```

First retry, the macOS prompt finally appeared on screen for the first time. I clicked Allow, Chrome focused to the requested profile, and all three actions then worked cleanly for both profiles. Your `#24147` design behaves exactly as you described, including bring-to-front without an `about:blank` tab.

My best guess for why this never bit you is that your Raycast row for `com.apple.systemevents` got granted by some other script or extension a long time ago, so for you the TCC check resolves synchronously inside Raycast's teardown window. For anyone who has not pre-granted it (which is probably most users), the extension is silently broken.

Given that, the shape I would like to land on this branch is much smaller than what is there now. Keep your AppleScript Profiles-menu script verbatim, plus the existing `do shell script Chrome --new-window` for `newWindow`, no changes to the AppleScript itself. Replace each `runAppleScript(...)` call with a small helper that writes the script to a temp file and runs `spawn("/usr/bin/osascript", [path], { detached: true, stdio: "ignore" })` + `child.unref()`. Errors from spawn or the child's `error` event get surfaced via `showToast`. Keep the favicon / opaque-URL guard for `chrome://` and `about:` bookmarks. Drop the `open -na` rewrite entirely. End result is around 100 lines added, 70 removed, and the user-facing behaviour matches your design intent.

One more thing worth flagging upstream: `@raycast/utils.runAppleScript` arguably should use `detached: true` by default, or at minimum document the teardown gotcha. I would be surprised if this is not biting other extensions silently.

I will force-push the new shape onto this branch and retitle to something like `google-chrome-profiles: fix silent failure via detached osascript spawn, fix bookmark crash`. Sorry for the long detour through the wrong fix.

0xdhrv:
Hey @frouo 👋 

Thanks for reviewing 😄 

Is this good to be merged?

frouo (approved):

pernielsentikaer (approved):
Looks good to me, approved 🔥

frouo:
@drye sorry for the long wait. I have tested the pull request, it works fine so far.

For the record, I asked Claude Opus 4.7 to review the MR. My prompt:

```
Please can you review google-chrome-profiles last 2 commits (664e880, 087cefa) please? Any security concern? Any performance concern?
``` 

Opus' reply:

---

I have enough context. Here's my review.

# Review: google-chrome-profiles commits `664e880` and `087cefa`

## Summary

- **`664e880`**: Replaces `@raycast/utils.runAppleScript` with a manual detached `osascript` spawn to fix silent failures caused by Raycast tearing down the Node process group ~40ms after the action handler returns. Also fixes an `Invalid URL`/opaque‑origin crash in the bookmarks list.
- **`087cefa`**: Tiny follow‑up — wraps the `writeFileSync` of the temp script file in try/catch so write failures surface as a toast instead of an uncaught exception.

Both changes are well‑motivated, the root‑cause analysis is rigorous, and the implementation is generally sound. There are a few real (but minor) issues worth flagging.

## Security

### 1. Temp file written world‑readable in a shared `/tmp` — low risk, worth tightening
`src/util/util.ts:211-213`

```ts
const scriptPath = join(tmpdir(), `raycast-google-chrome-profiles-${randomUUID()}.applescript`);
writeFileSync(scriptPath, script);
```

- The path uses `randomUUID()`, which is unguessable, so a local attacker cannot predict the filename to swap content (no TOCTOU vector here).
- However, `writeFileSync` on macOS defaults to mode `0666 & ~umask` (typically `0644`). The file contains a fully interpolated AppleScript that embeds `profile.name`, `profile.directory`, `binaryPath`, `appName`, and `url`. For most users that's not sensitive, but a bookmark URL passed via `ChromeAction.openUrl` can include tokens/credentials in query strings. Any local user (or process in another sandbox) can read `/tmp` during the brief window before `unlinkSync`.
- **Recommendation:** pass `{ mode: 0o600 }` to `writeFileSync`, or use `fs.mkdtempSync` to create a private directory first. Cheap, defensive, no downside.

### 2. AppleScript injection surface unchanged — OK
All user‑controlled inputs (`profile.name`, `profile.directory`, `browser.binaryPath`, `browser.appName`, `url`) continue to flow through `escapeAppleScriptString` before string interpolation. The escape function handles `\`, `"`, `\n`, `\r`, `\t` — sufficient for AppleScript double‑quoted string literals. The `newWindow` path additionally uses `quoted form of` for the shell layer. No regression here.

### 3. URL validation for bookmarks — improved
The previous code did `new URL(url).origin` blindly and crashed on `chrome://`/`about:`. The fix correctly restricts favicon resolution to `http:`/`https:`, which also avoids leaking `chrome-extension://<id>` origins to the favicon service. Good.

### 4. Loss of the `confirmAlert` failure dialog — UX, not security
`664e880` removes the previous fallback chain (Profiles menu → shell script → `confirmAlert` with detailed error). Now failures from the AppleScript itself (e.g., Profiles menu not present, profile not found) are silently absorbed because `osascript`'s stderr is `stdio: "ignore"` and the parent never waits for exit. Only spawn‑level and write‑level errors surface. **This is a genuine regression in observability**, although it's the price paid for the detach. See "Other findings" below for a mitigation.

## Performance

### 1. Synchronous filesystem calls on the React render/handler thread
`writeFileSync` and `unlinkSync` are blocking. They're called per action, on a path under `/tmp` (local APFS) with payloads <4 KB — typically sub‑millisecond, so the practical impact is negligible. Async (`fs.promises.writeFile`) would be cleaner but is not a real concern.

### 2. Per‑invocation `spawn` cost
Each action spawns a fresh `osascript` process. That's unchanged from before (Raycast's `runAppleScript` did the same), so no regression.

### 3. Temp file cleanup race when parent dies first
The cleanup is only triggered by `child.on("exit", ...)`. If Raycast tears the parent down before `osascript` exits — which is exactly the scenario the fix exists to handle — the `exit` listener never fires and the `.applescript` file is leaked into `/tmp`. The comment acknowledges this and leans on macOS's `/tmp` maintenance, which is fine, but over heavy use it accumulates small files until the next periodic cleanup. Not a real performance issue, but worth noting.

## Correctness / other findings

### A. `await willOpen()` runs unconditionally before potential early returns
`src/util/util.ts:282`

`willOpen()` was previously called inside each branch; now it runs once up front. If `runDetachedAppleScript` fails at the write or spawn step, `willOpen()` (typically closing the Raycast window / showing a HUD) has already executed. The user sees the window dismiss, then a failure toast — minor UX awkwardness, not a bug.

### B. `let child;` is implicitly `any`
`src/util/util.ts:223`

Minor — declaring it as `ChildProcess | undefined` (or assigning directly inside the `try` and moving the listeners inside) would be cleaner and keep `strict` typings.

### C. `showToast` returns a Promise that's not awaited
Throughout `runDetachedAppleScript`. Functionally fine (fire‑and‑forget), but means an unhandled rejection could be raised if the toast API ever rejects. Low risk.

### D. Loss of structured error reporting for in‑script failures
As mentioned in Security #4, AppleScript runtime errors (profile not found, Chrome not running, menu structure changed, accessibility permission revoked) are no longer surfaced at all. A pragmatic compromise: redirect `stdio` to a temp `.log` file (or use `stdio: ['ignore','ignore','pipe']` + drain stderr without holding the parent), and on non‑zero exit show a toast. That preserves the detach guarantee while restoring user feedback. The exit handler currently runs but ignores `code` and `signal`.

### E. `child.on("error", ...)` after spawn returns is mostly redundant
With `detached: true` and `stdio: "ignore"`, once `spawn` has returned a child object the only remaining `error` events are unusual (e.g., ENOMEM at fork). The synchronous `try/catch` already covers the common failures. Harmless to keep.

### F. The Profiles‑menu approach is now the only path
The previous "shell‑script fallback" branch is gone for `focus`/`newTab`/`openUrl`. If Chrome's menu bar structure changes (localization edge cases, future Chrome versions reordering menus), there's no graceful degradation. The commit message argues — correctly — that the fallback path was actually breaking UX (about:blank tab, no bring‑to‑front), so removing it is a deliberate trade. Just worth being aware of.

## Verdict

- **Security:** acceptable. One easy hardening win: set `mode: 0o600` on the temp script file, since bookmark URLs can carry secrets in query strings.
- **Performance:** no concerns; sync I/O is negligible at this size and frequency.
- **Main regression:** silent failure for in‑AppleScript errors. The fix solves a much worse silent‑failure problem, but it would be worth restoring some level of error surfacing via captured stderr + exit code.
- **`087cefa`** is a clean, correct hardening of the write step. No concerns.

Would you like me to draft concrete patches for (a) `mode: 0o600`, and (b) stderr capture with an exit‑code toast, so we can apply them once we leave plan mode?
---

What do you think @drye of this review?

drye:
Thanks for the praise and thanks for the review too! I agree both suggestions make sense to apply. I'll apply as soon as I have some bandwidth.

I updated my Raycast profile, looking forward to the swag 😁🙃

PR #27634, opened 2026-05-04
Fix VSCode Windows recent projects & CLI, WSL, Icon, Shortcut

jkker:
## Summary

- read modern VS Code recent entries from 
recently.opened while keeping compatibility with the older history.recentlyOpenedPathsList key
- fall back to storage.json metadata (backupWorkspaces and profileAssociations) so Windows and remote recent projects still appear when the DB list is incomplete
- preserve encoded remote URIs for Windows CLI opens, support both user/system VS Code install paths, and replace hardcoded shortcuts with shared Raycast shortcut helpers
- fixed Windows remote workspace opening to pass encoded `vscode-remote://` URIs through the VS Code CLI unchanged.
- added shared shortcut definitions based on Raycast common shortcuts where possible, with explicit Windows mappings for custom bindings.
- improved VS Code CLI path detection for both user and system installs on Windows.

## Why
Recent upstream VS Code builds on Windows now write recent-workspace data under 
recently.opened, while older code paths still reference history.recentlyOpenedPathsList. Upstream source also persists related workspace metadata in backupWorkspaces and profile mappings in profileAssociations, which helps recover recent entries when the primary list is incomplete.

## Validation

- verified on Windows against a live `%APPDATA%\\Code\\User\\globalStorage\\state.vscdb` and storage.json that contain local and WSL recent entries


0xdhrv (commented):
Hi @jkker 👋

Thanks for the contribution 🔥

---

jkker:
Addressed the Greptile feedback in commit 4cbf5c612:

- memoized the `storage.json` fallback read inside `useRecentEntries()` so it no longer re-reads synchronously on every render
- prevented false-success removals for entries backed by `storage.json`, which could otherwise reappear after deletion
- switched the SQLite write query to `INSERT OR REPLACE`
- added defensive URI parsing for remote display formatting
- added user-visible error handling around Windows CLI remote opens

jkker:
Addressed the db.ts removal-guard feedback in commit bd0164096.

The guard now blocks removal only for entries that are exclusively fallback-backed from `storage.json`. Entries that exist in both the DB and fallback sources remain removable from the DB, which prevents the Windows false-negative behavior called out in review.

jkker:
Addressed the latest db.ts review feedback in commit aafd2b598.

Changes in this pass:
- `Remove All` now clears DB-backed recent entries instead of being blocked whenever `storage.json` also has entries
- storage-backed duplicates are suppressed for the current Raycast session after `Remove` / `Remove All`, which avoids the earlier immediate reappearance problem
- the storage-only fallback path now supports the same session-level hide behavior instead of silently no-oping
- trimmed the unnecessary `packageManager` field and restored the original `es2020` tsconfig baseline by removing the remaining `replaceAll` dependency

ready for review! thanks @0xdhrv

0xdhrv (commented):

---

jkker:
merge conflicts resolved and ready for review - thanks @0xdhrv!

jkker:
Rebased onto current `main` and resolved the new conflicts from #27733 / #27732.

Conflict summary:
- kept #27733's Windows versioned `product.json` lookup for VS Code 1.118+ installs
- kept this PR's user/system VS Code CLI path detection and Windows remote URI opener
- extended the Windows install-dir lookup so Code / Insiders can resolve from LocalAppData, Program Files, or Program Files (x86) before applying the versioned resources fallback
- kept both changelog entries in date order

Local verification after the rebase:
- `npm run lint`
- `npm run build`

GitHub now reports the branch as mergeable again. Ready for review when checks finish - thanks @0xdhrv!

jkker:
Follow-up after Greptile's latest pass:

- collapsed the duplicated Windows VS Code path resolver
- preserved the existing `ctrl+shift+x` macOS shortcut for Unpin All
- memoized `getStateDatabasePath()` inside `useRecentEntries()`
- avoided double-decoding remote paths in `getRemoteDisplay()`
- made `Remove All` clear both recent-entry DB keys (`recently.opened` and `history.recentlyOpenedPathsList`)
- switched the retroactive changelog entry to `{PR_MERGE_DATE}` per repo convention

Local verification remains green:
- `npm run lint`
- `npm run build`

All Raycast/GitHub/Security checks are passing. Greptile is still pending on the final changelog-only push, but its previous pass was safe-to-merge with only the changelog placeholder note, which is now addressed.

0xdhrv (approved):
Looks good to me, approved ✅

PR #27640, opened 2026-05-04
feat: consolidate osint-toolkit into osint-web-check

jergensturdley:
## Summary

Consolidates the two previously-merged OSINT Raycast extensions (`extensions/osint-toolkit/` and `extensions/osint-web-check/`) into a single `osint-web-check` extension on the `osint-toolkit-raycast-clean` branch.

- **Single command surface**: `osint-web-check` (no longer `search-ioc`) accepts a URL, domain, IP, IPv6, or file hash in its search bar and renders both surfaces:
  1. A **Web Check** section — one row whose detail panel runs the 12 per-section web-checks (DNS, SSL, ports, headers, etc.).
  2. An **External OSINT Lookups** section — one row per enabled source (VirusTotal, Shodan, AbuseIPDB, etc.).
- **Single entry point**, one mental model; the `search-ioc` command is removed.

### Removed
- `extensions/osint-toolkit/` (entire directory)
- `extensions/osint-web-check/src/search-ioc.tsx`
- The `search-ioc` command from `extensions/osint-web-check/package.json`

### Added
- `extensions/osint-web-check/src/components/CheckDetails.tsx` — extracted the deep-dive detail panel; props `{ url, enabled }`; composed from the existing 12 per-section components.
- `extensions/osint-web-check/src/components/renderMode.ts` — pure `getRenderMode(raw, prefs): RenderMode` helper, no React imports.

### Modified
- `extensions/osint-web-check/src/osint-web-check.tsx` — single default export `OsintWebCheck`; search bar drives both sections via the debounced `useEffect` + `requestId` ref pattern.
- `extensions/osint-web-check/package.json` — `commands` array reduced to one entry (`osint-web-check`) with an optional `url` argument; preferences block now carries 14 `enable_*` booleans and the renamed `copy_on_open` preference.
- `extensions/osint-web-check/CHANGELOG.md` — new top entry.

### Preserved (untouched)
- `src/osint-ioc/**` (`ioc-types.ts`, `utils/{ioc-detection,osint-sources,url-builder,favorites}.ts`)
- `src/utils/{checkUrl,addHttps,useCheckDetail,types}.ts`
- `src/UrlInput.tsx` — still used as the SWR-failure fallback inside `CheckDetails`.
- All 12 per-section components (`DnsInfo.tsx`, `SSLCheck.tsx`, `OpenPorts.tsx`, `Headers.tsx`, `Redirects.tsx`, `WhoIs.tsx`, `Hsts.tsx`, `SecurityTxt.tsx`, `DnsSec.tsx`, `Firewall.tsx`, `CrawlRules.tsx`, `UrlIp.tsx`).

## Behavior preserved

- Detection: `detectIOCType(refangIOC(input))` — `ip`, `ipv6`, `domain`, `url`, `hash`, `unknown`.
- IOC sources filtered by `enable_<id>` preferences (default `true`).
- Defang/refang unchanged.
- Favorites still persist in LocalStorage under key `osint-favorites`.
- Deep-dive section hidden for `hash` / `unknown` IOC types (no URL to probe).
- `SourceDetailView` (markdown detail panel for an OSINT source) preserved verbatim.

## Note for reviewers — source count correction

The earlier review mentioned 18 sources; the consolidated `OSINT_SOURCES` array in this branch is **15 sources**, filtered to **14 enabled-by-default** (Kaspersky OpenTIP / `opentip` is `hash`-only and won't appear for non-hash queries). Please re-verify against `extensions/osint-web-check/src/osint-ioc/utils/osint-sources.ts` rather than the prior description.

## Follow-up commit (review fix)

`5a2093f490 fix(osint-web-check): rename copy_on_select preference to copy_on_open`

The preference was previously named `copy_on_select` but the actual behavior (driven by the `onOpen` callback on `Action.OpenInBrowser`) fires when the browser opens the URL — not when the user selects the row. Hard-renamed `copy_on_select` → `copy_on_open` in both `package.json` and the two call sites in `osint-web-check.tsx` so the identifier matches the behavior.

This is a **hard rename, no migration** — anyone who had this preference set to `true` will need to re-enable it after upgrading (the preference key changes, so the old value is discarded). Acceptable here because (a) the extension is new in this branch, (b) the default is `false`, so most users have nothing to migrate, and (c) keeping a dual-read shim would have left a misleading key in the manifest.

## Validation

- `npm run build` — passes; single entry point `src/osint-web-check.tsx`.
- `npm run lint` — 0 errors; 4 pre-existing `prefer-title-case` warnings in `CrawlRules.tsx`, `Headers.tsx`, `SSLCheck.tsx`, `UrlIp.tsx` (carried over from upstream, not introduced here).
- Manual UI checks (`npm run dev`) — empty input prompts, `github.com` resolves to domain type with deep-dive + 14 sources, `8.8.8.8` resolves to IP with IP-only sources, 64-hex SHA256 hides deep-dive section, garbage input shows "Invalid IOC", `enable_*` toggles hide sources, favorites persist across launches.

## Test plan

- [ ] Launch with no arg → empty view prompts for input.
- [ ] Type `github.com` → "Web Check" section + domain-filtered OSINT lookups.
- [ ] Type `https://example.com` → "Web Check" + URL-type lookups.
- [ ] Type `8.8.8.8` → "Web Check" + IP-only sources.
- [ ] Type a SHA256 hash → "Web Check" hidden, OSINT lookups only.
- [ ] Type garbage → "Invalid IOC" empty view.
- [ ] Toggle any `enable_*` preference off → that source disappears.
- [ ] "Add to Favorites" / "Open All Favorites" → favorites persist (LocalStorage `osint-favorites`).
- [ ] Force a SWR failure (e.g. `not-a-real-domain-xyz123.com`) → `UrlInput` form fallback appears.
- [ ] Launch via `osint-web-check https://example.com` → loads directly into "Web Check" + lookups.


jergensturdley:
Heartbeat

0xdhrv (commented):
Thanks for your contribution @jergensturdley  🔥

- https://www.raycast.com/gksander/osint-web-check by @gksander

The existing osint-web-check extension already covers OSINT lookups in the Security / Developer Tools space, and the functionality you've built here could be added to it as additional commands rather than shipping a separate extension. For example:

- A new Search IOC command alongside the existing OSINT Web Check command, with the auto-detection logic from ioc-detection.ts (IPv4, IPv6, domains, URLs, MD5/SHA1/SHA256 hashes, defang/refang handling)
- The multi-provider URL builder in url-builder.ts covering the 18 threat-intel platforms (VirusTotal, Shodan, AbuseIPDB, AlienVault OTX, URLScan, GreyNoise, Hybrid Analysis, Joe Sandbox, MalwareBazaar, Censys, crt.sh, etc.)
- The favorites system, per-source toggle preferences, and "open all sources" action
Worth noting that osint-web-check is already one of the 18 sources you reference in osint-sources.ts, so the two extensions are clearly in the same problem space — consolidating them would 

jergensturdley:
> Thanks for your contribution @jergensturdley 🔥
> 
> We already have an extension in the Store that deals with this. Could we consider enhancing the existing extension below instead of creating another one?
> 
> * https://www.raycast.com/gksander/osint-web-check by @gksander
> 
> The existing osint-web-check extension already covers OSINT lookups in the Security / Developer Tools space, and the functionality you've built here could be added to it as additional commands rather than shipping a separate extension. For example:
> 
> * A new Search IOC command alongside the existing OSINT Web Check command, with the auto-detection logic from ioc-detection.ts (IPv4, IPv6, domains, URLs, MD5/SHA1/SHA256 hashes, defang/refang handling)
> * The multi-provider URL builder in url-builder.ts covering the 18 threat-intel platforms (VirusTotal, Shodan, AbuseIPDB, AlienVault OTX, URLScan, GreyNoise, Hybrid Analysis, Joe Sandbox, MalwareBazaar, Censys, crt.sh, etc.)
> * The favorites system, per-source toggle preferences, and "open all sources" action
>   Worth noting that osint-web-check is already one of the 18 sources you reference in osint-sources.ts, so the two extensions are clearly in the same problem space — consolidating them would
> 
> If there are unique features or workflows you’re aiming to add, we’d love to hear them and see if they can be integrated into this to avoid duplication and improve discoverability.
> 
> > This would help avoid duplication and keep related functionality consolidated in one place.
> > As mentioned in our [extension guidelines here ↗](https://manual.raycast.com/extensions#:~:text=There%20is%20an%20active%20extension,extension%20providing%20very%20similar%20value)

Thank you for this input, I'll contribute to the existing extension 😀

jergensturdley:
Per feedback from @0xdhrv, this branch now does the following:

  - Single `osint-web-check` command auto-detects IOCs (IPv4/IPv6, domains, URLs, MD5/SHA1/SHA256, defang/refang) and shows a Web Check detail panel alongside per-source OSINT rows (VirusTotal, AbuseIPDB, Shodan, AlienVault OTX, URLScan, Pulsedive, Kaspersky OpenTIP, ipinfo, GreyNoise, WHOIS, MalwareBazaar, Censys, crt.sh, WebCheck).
  - The separate `osint-toolkit` extension and its `Search IOC` command are removed.
  - Favorites, per-source toggle preferences, Open All Favorites/Sources actions, and Copy IOC on Open all live inside
  `osint-web-check`.

0xdhrv (approved):
Looks good to me, approved ✅

PR #27646, opened 2026-05-04
Add Enpass extension

leolionart:
## Description
Adds an Enpass extension for searching a local Enpass vault through enpass-cli. It supports credential search, field inspection, copy/paste actions, and CLI-backed sorting by modified, created, recently used, usage, title, and category.

## Screencast
Metadata screenshots are included in `extensions/enpass/metadata`.

## Checklist
- [x] I ran `npm run build`
- [x] I tested this distribution build in Raycast


leolionart:
Addressed the Greptile feedback in commit 099e130:

- Guarded the `Copy Username / Email` action so it is only shown when a login exists.
- Updated the changelog heading to `## [Initial Release] - {PR_MERGE_DATE}`.
- Removed the manual `EnpassPreferences` interface and switched to the generated `Preferences` type, with `raycast-env.d.ts` included in `tsconfig.json`.

Verified locally with `ray build -e dist` and `ray lint`.

leolionart:
Addressed the updated Greptile feedback in commit 9f17a89:

- Removed duplicate copy actions in `EntryDetail` by excluding fields already handled by dedicated Username/Password/URL actions.
- Removed redundant `ENPASS_VAULT` environment passing and now rely on the existing `-vault` CLI argument.
- Updated the expect wrapper to send the master password once and wait for EOF instead of replying to every prompt containing `password`.

Verified locally with `ray build -e dist` and `ray lint`.

leolionart:
Addressed the latest Greptile feedback in commit e521d41:

- Omit the `-vault` argument when no vault path is configured, allowing the CLI default behavior instead of passing an empty path.
- Only show the password form for likely authentication/locked/password errors; configuration and CLI failures now remain visible as error toasts.

Verified locally with `ray build -e dist` and `ray lint`.

leolionart:
Addressed the latest Greptile PIN retry feedback in commit 8e20940:

- Authentication-like errors now keep the PIN form available even after a submitted PIN is rejected, so users can retry instead of being returned to an empty list.

Verified locally with `ray build -e dist` and `ray lint`.

leolionart:
Addressed the latest Greptile feedback and refreshed the PR after the stale marker:

- Removed the trailing empty positional argument from the sorted `list` invocation.
- Removed `storeValue` from the controlled sort dropdown so Raycast persisted dropdown state cannot race the initial preference-derived load.
- Switched detail lookup to prefer the Enpass item UUID when available, with the existing title/login fallback kept for entries without UUIDs.

Verified locally from the PR source with `npm run build` and `npm run lint`.

extensions/enpass/scripts/enpass_wrapper.exp
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/expect -f
```
0xdhrv, line 1: **suggestion:** Use the `useExec` hook and consider Windows support

Would it make sense to rephrase this a bit for clarity? Something like: “You can use this via Raycast with the `useExec` hook, and it may also be possible to extend support for Windows.”

Reference:
- https://developers.raycast.com/utilities/react-hooks/useexec

extensions/enpass/package.json
```diff
@@ -0,0 +1,119 @@
...
+      "title": "Search Credentials",
+      "subtitle": "Enpass",
+      "description": "Search, inspect, and copy Enpass credential fields",
+      "keywords": [
+        "enpass",
+        "password",
+        "credential",
+        "login"
+      ],
+      "mode": "view"
```
0xdhrv, line 24: **issue**: metadata image style

The current metadata screenshot seems to be generated and not of Raycast.

Reference:

0xdhrv (commented):
Hey @leolionart 👋 

---

PR #27653, opened 2026-05-04
[GitLab] Add pipeline triggers, My Pipelines and job controls

danilorequena:
## Description

Closes some gaps in the GitLab extension's CI/CD coverage. Today pipelines and jobs are only reachable as a sub-view of a project, with retry as the only mutation. This PR adds:

- **`My Pipelines`** — new top-level command listing recent pipelines from up to 20 of your most recently active projects, grouped by project. Auto-refreshes on the existing `CI Auto Refresh Interval` preference. Falls back to a clear "All projects inaccessible" empty view when every project returns 404 (signals SAML SSO enforcement on a group token).
- **`Trigger New Pipeline`** action — opens a form with branches/tags dropdown and up to 3 key/value variables (env_var). Hits `POST /projects/:id/pipeline`.
- **`Play Job` / `Cancel Job`** actions — only render when the job's status allows it (`manual` for play; `pending`/`running`/`created`/`scheduled`/`preparing` for cancel). Hit `POST /jobs/:id/play` and `POST /jobs/:id/cancel`.
- **`Show Log`** action — opens a `Detail` view rendering the job trace from `GET /jobs/:id/trace` as a code block. Strips ANSI escapes, truncates at 100 KB with a header pointing to the browser, auto-refreshes (min 5 s) while the job is non-terminal.

The `RetryFailedPipelineJobsAction` status check is now case-insensitive so it works for both the GraphQL flow (uppercase `FAILED`) and the REST flow used by `My Pipelines` (lowercase `failed`).

## API surface added to `src/gitlabapi.ts`

- `triggerPipeline(projectID, ref, variables)`
- `getProjectBranches(projectID, search?)`, `getProjectTags(projectID, search?)`
- `playJob(projectID, jobID)`, `cancelJob(projectID, jobID)`
- `getJobTrace(projectID, jobID)` — uses `getFetcher()` directly because the trace endpoint returns `text/plain`, mirroring the pattern in `getProjectReadme`.
- `getMyRecentPipelines({ perProject, maxProjects })` — REST-based aggregation (GraphQL nested `pipelines.nodes` over many projects hits the gitlab.com complexity limit). Iterates membership projects ordered by `last_activity_at`, parallel via `Promise.allSettled` so one project's permission gap doesn't block the rest. Returns `{ projects, scanned, inaccessible }` so the UI can distinguish "no pipelines" from "all projects blocked by SSO".

## Files

New:
- `src/pipeline_my.tsx` (command entry)
- `src/components/pipeline_my.tsx`
- `src/components/pipeline_trigger_form.tsx`
- `src/components/job_log.tsx`

Modified:
- `package.json` — adds `pipeline_my` to `commands`, adds contributor
- `CHANGELOG.md`
- `src/gitlabapi.ts`
- `src/components/pipeline_actions.tsx`
- `src/components/job_actions.tsx`
- `src/components/jobs.tsx`

## Validation

- `npm run lint` clean for added/modified files
- `npm run build` succeeds
- Manually validated end-to-end: `My Pipelines` (cold + auto-refresh), `Show Jobs`, `Show Log` (running + strip ANSI + truncation), `Trigger New Pipeline` (default branch + variables, error on bad ref).


danilorequena:
Thanks @greptile-apps for the review. Addressed all three findings in `ea1384db2`:

- **P1 — tag ref pre-selection** (`pipeline_trigger_form.tsx`): `defaultRef` now also matches against fetched tags, so pipelines that ran on a tag pre-select that tag instead of falling back to the default branch.
- **P2 — `inaccessible` 403 counting** (`gitlabapi.ts`): the counter now also catches `Forbidden` / `403` messages, not just `Not found`. The `allBlocked` heuristic now fires for SAML SSO setups that return 403 instead of 404.
- **P2 — type cast in `pipeline_my.tsx`**: aligned `PipelineCardShape.projectId` to `string` to match `Pipeline.projectId`. Dropped the `as unknown as` cast — it's now a single-step `as Pipeline`.

Build, lint and Prettier are clean.

danilorequena:
Round 2 addressed in `85e3335b0`:

- **P1 — `perProject` limit ignored** (`gitlabapi.ts`): switched `getMyRecentPipelines` to call `getFetcher()` directly with a literal query string, mirroring `getProjectReadme`/`getJobTrace`. `this.fetch()` was overriding `per_page` to 50; now the URL carries the requested value (default 5). Worst-case payload drops from ~1000 to ~100 pipelines per refresh. Also catches 404/403 explicitly so the `inaccessible` heuristic stays consistent.
- **P2 — OSC escape stripping** (`job_log.tsx`): regex now also covers OSC sequences (`ESC ] … BEL` and `ESC ] … ST`) — common in `cargo`, `git`, and CI hyperlinks. Kept inline (no new dependency).

Build, lint and Prettier are clean.

pernielsentikaer (approved):
Looks good to me, approved 🔥

Issue #16573, opened 2025-01-23
Clipboard History can not record large text


magiceses:
### Description

when i copy a large file contents, which have 70w char

<img width="748" alt="Image" src="https://github.com/user-attachments/assets/248bf0f5-ddad-4a0d-afc6-130ef2c68b3d" />

but nothing was recorded in raycast clipboard history

ksbrar:
+1

LitoMore:
See https://raycastcommunity.slack.com/archives/C01A0R0NXGQ/p1736234410995599.

The max length of clipboard content is limited to 32678 characters.

ksbrar:
Unfortunate

mathieudutour:
This issue is not related to the API or extensions. Could you please use the `Send Feedback` command inside Raycast or use the #help or #bugs channel in our [Slack Community](https://www.raycast.com/community) 😊

winzamark123:
for anyone looking for a workaround, hopefully this gets merged and becomes an extension! #27670

imranq:
Hi just wondering if someone is working on this, this would be very helpful to be able to tune limits yourself

PR #27670, opened 2026-05-05
Add large-clipboard-history extension

winzamark123:
## Description
Raycast's built-in clipboard history caps each entry at 32,768 characters. So anything larger silently never makes it in. This extensions helps to address this by only storing content that is over the limit of raycast history and doesn't overlap the content in the native Raycast Clipboard History. 

Since it only captures the history of text over 32,768 characters it doesn't overlap with Raycast Clipboard!

Useful for people who frequently copy log dumps, full file contents, multi-page prompts, or long AI conversations, a meaningful slice of clipboard activity is never recoverable from history.

This helps address #16573 !
## Screencast
https://github.com/user-attachments/assets/680f8f07-3086-4cb5-9f67-fd0907837ae8

An example of the use case above here where i copy a large text of Lorem Ipsum
## Checklist


0xdhrv (commented):
Thanks for your contribution @winzamark123 🔥

We already have a built-in clipboard history feature that handles this. With Raycast V2, we’ve introduced no character limit for the clipboard history.

PR #27675, opened 2026-05-05
Add ai-translate extension

semantic-craft:
## Description

Adds **AI Translate**, a Raycast extension focused on fast screenshot OCR translation.

Version 1.0.0 is intentionally scoped: take a screenshot quickly, extract visible text reliably, and translate it with as little friction as possible. It is not trying to become a full translation suite. The main use case is the common moment where users can see text on screen but cannot select it.

Why not just use Raycast's built-in translation? Raycast's built-in flow is useful when text can be selected or passed as plain text, but it does not cover OCR translation. In many real workflows, text is locked inside app UI, images, PDFs, slides, videos, remote desktops, protected documents, or web pages with broken selection. In those cases, screenshot capture is the most reliable input surface.

Raycast's built-in AI features also do not provide this level of model routing for translation. AI Translate lets users choose the provider, model ID, base URL, and API key themselves, so translation can run through the exact Token Plan, Coding Plan, or provider-specific subscription they already pay for.

AI Translate layers OCR and AI translation on top of that screenshot-first workflow. It prioritizes cost-effective, high-quality providers such as DeepSeek, Xiaomi MiMo, MiniMax, and Kimi through Anthropic-compatible `/v1/messages` endpoints, while still supporting OpenAI / ChatGPT and Gemini. Kimi now defaults to `https://api.kimi.com/coding/` as its Anthropic-compatible coding base URL.

The default system prompt is sense-for-sense rather than literal: it asks the model to write as a native speaker of the target language would naturally express the same idea, while preserving the source meaning, tone, facts, and level of formality. For Chinese targets, it explicitly asks for natural Chinese expression instead of English syntax rewritten with Chinese words.

The extension also supports configurable translation prompts. Users can choose a built-in `Prompt Profile` such as Screenshot OCR, Technical / Developer, Academic Writing, Legal / Policy, Subtitle / Conversation, or Custom Only, then add reusable `Custom Prompt Instructions` for terminology, audience, tone, and formatting preferences. These instructions are included with every translation request while the source text stays in a separate `Text:` block.

The extension provides four commands:

- `Translate Screenshot`: captures a screen region, runs OCR, and opens the recognized text in the translation view.
- `Extract Text from Screenshot`: captures a screen region, runs OCR, and opens an editable result view with copy, compact-copy, translate, and retake actions.
- `Copy Text from Screenshot`: captures a screen region, runs OCR, and copies the recognized text to the clipboard without opening a result view.
- `Translate Selected Text`: translates selected text, typed text, or a fallback argument when selection is available.

OCR options include macOS Vision, Tesseract, Baidu OCR, and PaddleOCR HTTP services.

Official API documentation links are included in the README for each provider:

- DeepSeek: https://api-docs.deepseek.com/guides/anthropic_api
- Xiaomi MiMo: https://platform.xiaomimimo.com/static/docs/api/chat/anthropic-api.md and https://platform.xiaomimimo.com/static/docs/tokenplan/quick-access.md
- MiniMax: https://platform.minimaxi.com/docs/api-reference/text-anthropic-api and https://platform.minimaxi.com/docs/token-plan/quickstart
- Gemini: https://ai.google.dev/gemini-api/docs/text-generation and https://ai.google.dev/gemini-api/docs/api-overview
- Kimi: https://platform.kimi.ai/docs/guide/agent-support and https://platform.kimi.ai/docs/api/overview
- OpenAI / ChatGPT: https://platform.openai.com/docs/api-reference/chat/create
- Baidu OCR: https://cloud.baidu.com/doc/OCR/s/zk3h7xz52, https://cloud.baidu.com/doc/OCR/s/1k3h7y3db, and https://cloud.baidu.com/doc/AI_REFERENCE/s/um3zhy50e
- PaddleOCR HTTP: https://paddlepaddle.github.io/PaddleOCR/main/en/version3.x/deployment/serving.html and https://www.paddleocr.ai/main/en/version3.x/pipeline_usage/OCR.html

## Screencast

Store screenshots are included in:

- `metadata/ai-translate-1.png`
- `metadata/ai-translate-2.png`
- `metadata/ai-translate-3.png`

## Checklist


semantic-craft:
Addressed the Greptile feedback in the latest update.

Changes made:
- Fixed Auto language detection so Japanese kana and Korean Hangul are excluded before treating CJK ideographs as Chinese.
- Added safe JSON parsing for OCR API responses so non-JSON error bodies surface readable HTTP / invalid JSON errors instead of raw SyntaxError crashes.
- Removed unreachable Kimi / MiMo provider workarounds from the OpenAI-compatible path now that those providers use Anthropic-compatible requests.
- Removed the redundant Raycast List throttle and kept the existing internal debounce to reduce translation startup latency.

Latest extension commit in this PR: 70d1ae0.

semantic-craft:
### Update: v2 UX redesign (commits b041156, 383dba7)

This is a significant UX overhaul based on real usage feedback. Summary of changes:

#### New commands
- **Translation Settings** — dedicated Form for model tier, prompt profile, style, and custom instructions.
- **Screenshot Translate** rewritten from `no-view` to `view` mode — OCR result is shown for review before translation, with multi-provider comparison in the same view.
- **Screenshot OCR** replaces both `Extract Text from Screenshot` and `Copy Text from Screenshot` — editable Form with new text processing actions.

#### Removed commands
- `Extract Text from Screenshot` → merged into `Screenshot OCR`
- `Copy Text from Screenshot` → merged into `Screenshot OCR` (Copy action)

#### Model tier system
Replaces manual model name text fields with a Fast / Pro / Custom tier selector. Each provider has a curated model catalog. Custom tier preserves the ability to enter any model ID.

#### TTS read-aloud
Gemini 3.1 Flash TTS integration for reading translations and source text aloud (⌘S / ⌘⇧S). Raw PCM response is wrapped in a WAV header for `afplay` playback.

#### OCR improvements
- Removed PaddleOCR support (simplified to Local / Tesseract / Baidu).
- Baidu OCR: `language_type` changed from text field to dropdown (10 languages + auto detect), added `paragraph` detection checkbox.
- New text processing: strip line breaks (⌘L), auto-paragraph (⌘⇧P).

#### Bug fixes from UX audit
- `closeMainWindow()` before screenshot capture to avoid capturing the Raycast window.
- Screenshot cancel/fail now shows retry UI instead of dead EmptyView.
- TTS uses async `exec` instead of blocking `execSync`.
- Runtime settings inherit from Preferences on first load.
- Translation Settings form uses key-based remount for correct reset behavior.
- Removed `arguments` from translate command to prevent unwanted input prompt on launch.

#### Other
- Model name shown in result accessories and detail pane.
- Source text displayed as quote block instead of code fence.
- Toast feedback on model/profile/style switch.
- README updated to reflect all changes.

All changes build and lint cleanly. Existing provider protocols and API integration are unchanged.

semantic-craft:
Addressed the two remaining Greptile style suggestions in the latest update.

Changes made:
- Updated `CHANGELOG.md` to use the Raycast convention `Initial Release`.
- Updated `src/tts.ts` to use the generated `Preferences` type with `getPreferenceValues<Preferences>()`.

Local verification passed: Prettier, TypeScript, Raycast lint, and Raycast build.

PR #27677, opened 2026-05-06
Slack: accent-insensitive search in Open Channel and Send Message

buzali:
## Description

Folds names through Unicode NFD + diacritic stripping before matching, so typing `Angeles` finds `Ángeles` (and the same for any other accented characters).

**Bug:** the `matchesAllWords` helper in both `src/search.tsx` and `src/send-message.tsx` only lowercased — accented names were unfindable unless you typed the exact diacritic. Even typing the accented form could miss when the source string and the input were in different normalization forms (NFC `U+00E1` vs NFD `a` + `U+0301`).

**Fix:** new `foldForSearch(s)` helper applied to both sides of the comparison:

```ts
function foldForSearch(s: string): string {
  return s
    .normalize("NFD")
    .replace(/\p{Diacritic}/gu, "")
    .toLowerCase();
}
```

Both commands (`Open Channel` and `Send Message`) had a copy of `matchesAllWords` and both are updated.

## Screencast

Tested locally against a workspace containing a member named "Ángeles" — typing `angeles`, `Angeles`, `ángeles`, and `Ángeles` all match the entry now (previously only `Ángeles` matched, sometimes).

## Checklist

### DIFF
--- a/extensions/slack/CHANGELOG.md
+++ b/extensions/slack/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Slack Changelog
 
-## [Accent-insensitive search] - {PR_MERGE_DATE}
+## [Accent-insensitive search] - 2026-05-07
 
 - "Open Channel" and "Send Message" now match channel and user names regardless of diacritics, so typing `Angeles` finds `Ángeles` (and the same for any accented characters).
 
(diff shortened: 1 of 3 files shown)


0xdhrv (approved):
Looks good to me, approved ✅

PR #27685, opened 2026-05-06
Fix Google Chrome history search when Chrome is running

### LINKED ISSUE #27493: [Google Chrome] Search History fails with “Database is locked” while Chrome is open
### Extension

### Raycast Version

1.104.13

### OS Version

26.4.1

### Description

The Google Chrome extension’s Search History command fails with “Database is locked” while Chrome is open.

The command works after quitting Chrome, but the same error is reproduced once Chrome is opened again.

### Steps To Reproduce

1. Open Google Chrome
2. Run the Raycast Google Chrome extension command: Search History
3. See the error: “Database is locked”
4. Quit Chrome completely
5. Run Search History again
6. The command works
7. Open Chrome again
8. Run Search History again
9. The error is reproduced

### Current Behaviour

Search History fails while Chrome is open with “Database is locked”.

### Expected Behavior

Search History should work while Chrome is open.

### Likely Cause

The extension appears to read Chrome’s live History SQLite database while Chrome has it locked.

### Suggested Fix

Query a temporary copy of the Chrome History database instead of the live database. Ideally copy the related SQLite files together before querying:

- History
- History-wal
- History-shm

Jeffrey95:
## Description

Bumps `@raycast/utils` from `^2.2.2` to `^2.2.4` to pick up the `useSQL` lock fallback fix for `node:sqlite`.

When Chrome is running, searching Chrome history can fail with `database is locked`. The extension already uses `useSQL`, which has a fallback that copies the database to a temporary file when SQLite reports the source DB is locked. However, older `@raycast/utils` versions only detected lock errors from message text like `(5)`/`(14)`. With Raycast's newer Node runtime, `node:sqlite` reports the lock as `errcode: 5` with message `database is locked`, so the fallback was not triggered.

This matches the root cause fixed for Zen Browser in #27373. Updating `@raycast/utils` makes `useSQL` detect the `node:sqlite` error code and use the existing DB-copy fallback.

Fixes #27493

## Verification

- Reproduced the failure locally with `@raycast/utils` 2.2.2 while Chrome was running.
- Verified that the same extension works after bumping to `@raycast/utils` 2.2.4.
- Ran `npx tsc --noEmit`.
- Ran `npm run lint`.
- Ran `npm run build`.

## Checklist

### DIFF
--- a/extensions/google-chrome/CHANGELOG.md
+++ b/extensions/google-chrome/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Google Chrome Changelog
 
-## [Fix Chrome History Search When Chrome Is Running] - {PR_MERGE_DATE}
+## [Fix Chrome History Search When Chrome Is Running] - 2026-05-07
 
 - Fix searching Chrome history while Chrome keeps the history database locked.
 
--- a/extensions/google-chrome/CHANGELOG.md
+++ b/extensions/google-chrome/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Google Chrome Changelog
 
+## [Fix Chrome History Search When Chrome Is Running] - {PR_MERGE_DATE}
+
+- Fix searching Chrome history while Chrome keeps the history database locked.
+
 ## [Add Name Window Command] - 2026-03-04
 
 - Add Name Window command to name the currently active Google Chrome window.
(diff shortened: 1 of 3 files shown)


0xdhrv (approved):
Looks good to me, approved ✅

PR #27687, opened 2026-05-06
Add force-quit extension

2dubu:
## Description

Force Quit lets you quit running apps and processes from Raycast, mirroring macOS' built-in ⌥⌘⎋ — but without leaving the keyboard.

### Force Quit
Lists running GUI applications sorted by memory usage. Press `Enter` on an item to force quit after a confirmation prompt.

### Force Quit Process
Same flow but lists every process on the system, including background daemons. Top-level GUI apps show their native icons; helpers and daemons fall back to a terminal icon. Useful when an app is unresponsive but isn't visible in the standard list.

## Behavior
스크린샷 2026-05-18 오전 9 45 12

- The confirm dialog uses the destructive button style.
- Confirming sends `SIGKILL` immediately — equivalent to macOS' built-in Force Quit.
- A HUD confirms success or failure (`✓ <name> force quit` / `✗ Failed to quit <name>`).
- `⌘R` refreshes manually. The list does not auto-poll.
- `⌘.` copies `<name> (PID <pid>)` to the clipboard for use in other tools.

## Implementation notes

- **Data source**: `lsappinfo list` (foreground apps with bundle paths) + `ps -axo pid=,rss=,comm=` for memory and the all-processes list. No external dependencies, no native binaries, no AppleScript.
- **Performance**: switched from AppleScript (`every application process whose background only is false` + per-process property reads) to `lsappinfo` after measuring ~17s on 20 apps from the Apple Event roundtrips. New path: ~25ms.
- **Permissions**: none beyond the public process-listing tools. No Automation permission, no Accessibility permission, no full-disk access.

## Testing

- Tested on macOS 14 (Apple Silicon) via \`npm run dev\`.
- Force quit verified on Chrome, Slack, Notion, and helper subprocesses.
- \`EPERM\` from privileged processes (e.g., \`launchd\`) surfaces as the \`✗\` HUD without crashing the extension.

## Screencast
force-quit-1
force-quit-2

## Checklist


0xdhrv (commented):
Hey @2dubu 👋 

This is something you can natively do in Raycast.

Image

---

2dubu:
Hey @0xdhrv 👋 Thanks for the thoughtful review — totally fair to check for overlap with the native Quit Applications command.

The reason I built this is a workflow the native command doesn't cover: **when my Mac suddenly slows down or beachballs, I want a single keystroke to see _what's actually eating memory_ and kill it — without opening Activity Monitor.** A few things make this distinct from the native quit:

1. **All processes, not just foreground apps.** The `Force Quit Process` command lists every running process — background daemons, helpers, menu-bar agents, CLI tools — via `ps -ax`, which the native Quit Applications command (and the macOS ⌥⌘⎋ dialog) can't reach at all.
2. **Sorted by memory, shown inline.** Both commands sort by RSS descending and show each item's memory usage, so the actual resource hog surfaces at the top. The native command shows no memory and no ordering — so it can't answer "what's slowing me down."
3. **Real force quit (`SIGKILL`).** Native quit sends a graceful terminate that hung/unresponsive apps ignore; this sends `SIGKILL` (behind a destructive confirmation), matching the macOS Force Quit dialog semantics.

So the framing is less "quit an app" and more "**a lightweight, memory-sorted triage + force-kill console**" for diagnosing slowdowns. If consolidation is still the preference, I'm happy to **narrow the extension to just the process-level command** (the part that has no native equivalent) and drop the foreground-apps list to avoid overlap — let me know which direction you'd prefer.

I'll also address the automated review nits (locale tag, absolute `ps` path) before marking it ready again. Thanks again! 🙏

0xdhrv:
> Hey @0xdhrv 👋 Thanks for the thoughtful review — totally fair to check for overlap with the native Quit Applications command.
> 
> The reason I built this is a workflow the native command doesn't cover: **when my Mac suddenly slows down or beachballs, I want a single keystroke to see _what's actually eating memory_ and kill it — without opening Activity Monitor.** A few things make this distinct from the native quit:
> 
> 1. **All processes, not just foreground apps.** The `Force Quit Process` command lists every running process — background daemons, helpers, menu-bar agents, CLI tools — via `ps -ax`, which the native Quit Applications command (and the macOS ⌥⌘⎋ dialog) can't reach at all.
> 2. **Sorted by memory, shown inline.** Both commands sort by RSS descending and show each item's memory usage, so the actual resource hog surfaces at the top. The native command shows no memory and no ordering — so it can't answer "what's slowing me down."
> 3. **Real force quit (`SIGKILL`).** Native quit sends a graceful terminate that hung/unresponsive apps ignore; this sends `SIGKILL` (behind a destructive confirmation), matching the macOS Force Quit dialog semantics.
> 
> So the framing is less "quit an app" and more "**a lightweight, memory-sorted triage + force-kill console**" for diagnosing slowdowns. If consolidation is still the preference, I'm happy to **narrow the extension to just the process-level command** (the part that has no native equivalent) and drop the foreground-apps list to avoid overlap — let me know which direction you'd prefer.
> 
> I'll also address the automated review nits (locale tag, absolute `ps` path) before marking it ready again. Thanks again! 🙏

Thanks for clarification 🔥

We noticed that there is already an extension in the Store that appears to cover a very similar use case:

* [Kill Process](https://www.raycast.com/rolandleth/kill-process) by @rolandleth

Both extensions seem focused on helping users discover running processes and terminate them directly from Raycast. From a user's perspective, the primary workflow—finding a process and killing it without leaving Raycast—appears largely the same.

Could you help us understand what unique workflows or capabilities your extension provides compared to the existing one? If there are significant differences in functionality, process management features, filtering capabilities, monitoring, or overall user experience, it would be helpful to highlight those in the PR description.

If the goal is to improve the process-killing experience, it may also be worth exploring whether those enhancements could be contributed to the existing extension instead, helping keep related functionality consolidated and easier for users to discover.

> This would help avoid duplication and ensure we provide the best possible experience through a single, well-maintained extension when the underlying use cases are similar.

2dubu:
Hey @0xdhrv 👋 I installed Kill Process and used it for a while, then read through its source — you're right that the core "find a process and kill it" path overlaps, and it's a great, mature tool. Where Force Quit differs is its scope and default behavior:

**It's an app-first Force Quit, not a process manager.** The primary `Force Quit` command mirrors macOS' ⌥⌘⎋ window: it shows one row per foreground app, sourced from `lsappinfo` (`type == Foreground`) — exactly what macOS itself treats as a foreground app. Kill Process's default is the full flat process table, so a browser spreads across 20+ separate helper rows (`Google Chrome Helper (Renderer)` ×N) and you end up scanning processes, not apps — Force Quit shows you Chrome once. (Kill Process also classifies app-vs-binary heuristically by regex-matching `.app` in the `ps` path via `getProcessType`; Force Quit reads the OS' own app registry instead.) The second command, `Force Quit Process`, covers the rarer "it's a daemon/helper" case.

**It actually force-quits privileged processes.** When an unprivileged `SIGKILL` hits `EPERM` on a root-owned process, Force Quit falls back to an `osascript` admin prompt (the native auth dialog) and kills it. Kill Process's force path shells out to `sudo kill -9`, which needs passwordless sudo or a TTY and otherwise fails silently from a GUI launch.

Beyond that I've kept it deliberately lightweight and keyboard-native (memory/CPU sort, a destructive-style confirm, HUD feedback) rather than a full manager with restart / kill-tree / AI tools.

Individually these are small, but together they add up to a different product for a different user: Kill Process is built for power users doing process-level surgery, while Force Quit targets the everyday, app-centric user — someone who just wants a familiar Force-Quit-style list to spot an app and quit it, not a process manager to learn. I think that audience is distinct enough for it to stand on its own, so I'd really appreciate you weighing these differences with that user in mind.

0xdhrv:
> Hey @0xdhrv 👋 I installed Kill Process and used it for a while, then read through its source — you're right that the core "find a process and kill it" path overlaps, and it's a great, mature tool. Where Force Quit differs is its scope and default behavior:
> 
> **It's an app-first Force Quit, not a process manager.** The primary `Force Quit` command mirrors macOS' ⌥⌘⎋ window: it shows one row per foreground app, sourced from `lsappinfo` (`type == Foreground`) — exactly what macOS itself treats as a foreground app. Kill Process's default is the full flat process table, so a browser spreads across 20+ separate helper rows (`Google Chrome Helper (Renderer)` ×N) and you end up scanning processes, not apps — Force Quit shows you Chrome once. (Kill Process also classifies app-vs-binary heuristically by regex-matching `.app` in the `ps` path via `getProcessType`; Force Quit reads the OS' own app registry instead.) The second command, `Force Quit Process`, covers the rarer "it's a daemon/helper" case.
> 
> **It actually force-quits privileged processes.** When an unprivileged `SIGKILL` hits `EPERM` on a root-owned process, Force Quit falls back to an `osascript` admin prompt (the native auth dialog) and kills it. Kill Process's force path shells out to `sudo kill -9`, which needs passwordless sudo or a TTY and otherwise fails silently from a GUI launch.
> 
> Beyond that I've kept it deliberately lightweight and keyboard-native (memory/CPU sort, a destructive-style confirm, HUD feedback) rather than a full manager with restart / kill-tree / AI tools.
> 
> Individually these are small, but together they add up to a different product for a different user: Kill Process is built for power users doing process-level surgery, while Force Quit targets the everyday, app-centric user — someone who just wants a familiar Force-Quit-style list to spot an app and quit it, not a process manager to learn. I think that audience is distinct enough for it to stand on its own, so I'd really appreciate you weighing these differences with that user in mind.

Hey, you can still add a command for apps or create a preference so the kill extension shows only apps.

The overlap seems significant for the domain

PR #27713, opened 2026-05-07
[Skills] Fix support directory initialization on first launch

keito4:
## Summary

- Prevent "Could not create extension support directory" errors by ensuring `environment.supportPath` exists before `useCachedPromise` hooks run
- Add `ensureSupportDir()` utility that calls `mkdirSync` with `recursive: true` at module scope in both `search-skills` and `manage-skills` commands

## Details

The Raycast framework expects the extension support directory to exist when `Cache` (used internally by `useCachedPromise`) is initialised. On first launch or after a Raycast update, this directory may not yet exist, causing the error.

The fix is a best-effort `mkdirSync` call that:
- Runs at module scope (before any React hooks)
- Is a no-op when the directory already exists (`recursive: true`)
- Silently catches errors, deferring to the framework's own error handling

## Test plan

- [ ] Fresh install: delete the extension support directory, launch "Search Skills", verify no error
- [ ] Existing install: launch both commands, verify no regression
- [ ] Build passes: `npm run build` succeeds

### DIFF
--- a/extensions/skills/CHANGELOG.md
+++ b/extensions/skills/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Skills Changelog
 
+## [Fix Support Directory Initialization] - {PR_MERGE_DATE}
+
+- Ensure the extension support directory exists before cached hooks run, preventing "Could not create extension support directory" errors on first launch
+
 ## [Add Newly Supported Agents] - 2026-05-05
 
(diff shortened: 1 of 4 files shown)


pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #27724, opened 2026-05-07
gather: support Gather V2 alongside V1

Ethnical:
## Description

Gather's v2 desktop client installs at `/Applications/GatherV2.app` rather than `/Applications/Gather.app`, so the existing `existsSync` check (and the subsequent `activate application "Gather"` call) silently fail for v2-only users — every command shows the "Gather is not installed" toast. This PR adds v2 detection and uses POSIX-file activation so the extension works on whichever client is installed.

## Changes

- `gatherInstalled()` now returns true if either `/Applications/GatherV2.app` or `/Applications/Gather.app` is present, preferring v2 when both coexist.
- Activate the app via its POSIX file path (`tell application (POSIX file "${GATHER_APP_PATH}" as alias) to activate`) so we don't depend on the bundle's `CFBundleName` matching the `.app` filename — robust whether v2's bundle is named "Gather", "Gather V2", or "GatherV2".
- Keystrokes (⌘⇧A / ⌘⇧V / ⌘U) are unchanged — they remain Gather's documented defaults; any rebinding is the user's choice in `Gather → Preferences → Keyboard Shortcuts`.
- Version bump 1.0.1 → 1.0.2 and a CHANGELOG entry.

## Checklist

- [x] Behavior unchanged for V1 users (`/Applications/Gather.app`)
- [x] No new dependencies; minimum-diff (~6 net code lines)
- [x] CHANGELOG and version bumped
- [ ] V2 verification on `/Applications/GatherV2.app` — requested from extension author / Raycast reviewer with v2 installed

### DIFF
--- a/extensions/gather/CHANGELOG.md
+++ b/extensions/gather/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Gather Changelog
 
+## [Gather V2 support] - {PR_MERGE_DATE}
+Detect `/Applications/GatherV2.app` in addition to `/Applications/Gather.app` and activate the app via its POSIX path so the extension works on Gather V2.
+
 ## [1.0.1] - 2022-07-11
 Add a proper README.md file and rewrite the extension description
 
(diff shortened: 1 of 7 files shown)


0xdhrv (approved):
Looks good to me, approved ✅

PR #27729, opened 2026-05-07
[KEF Control] Fix setData calls and surface API errors

cjse:
## Description

Fix Set Source and Set Volume in the KEF Control extension. Both commands silently no-op'd: they sent `GET /api/setData?...&value={...}` to the speaker, but the LSX II / LS50 Wireless II HTTP API returns 405 Method Not Allowed on GET for `setData`, and 415 Unsupported Media Type on form-encoded POST. It only accepts POST with a JSON body containing `path`, `roles`, and `value`.

The fetch results were also discarded, so failures never reached the user — the AI tool reported success and the form just dismissed.

Changes:
- `setData` calls now use POST with `Content-Type: application/json` and `path`/`roles`/`value` in the JSON body.
- All three tools throw on non-2xx responses, so failures surface via the existing `showFailureToast` in `useVolume` instead of being silently swallowed.

Verified end-to-end against an LSX II: `getData` returns the expected JSON, `setData` flips physical source / volume, and the Raycast `Set Source` form correctly drives the speaker.

## Screencast

N/A — bug fix, no UI changes.

## Checklist

- [x] I ran \`npm run build\` and tested this distribution build in Raycast
- [x] I checked that assets used in the \`README\` are located outside the metadata folder if they were not generated with our metadata tool


0xdhrv (approved):
Looks good to me, approved ✅

PR #27741, opened 2026-05-08
Add ai-usage-dashboard extension

dawidmatyjasik:
## Description

Adds AI Usage Dashboard, a Raycast extension for viewing local AI coding usage across Claude Code, Codex, and AMP. It shows usage summaries, recent daily activity, costs, token totals, and Claude billing block status with offline-friendly cached results.

## Screencast
Screenshot 2026-05-08 at 12 57 45
Screenshot 2026-05-08 at 12 58 12
Screenshot 2026-05-08 at 12 58 17

## Checklist


0xdhrv (commented):
Thanks for your contribution  

- https://www.raycast.com/thuggyduck/agent-usage

---

dawidmatyjasik:
Thanks for the feedback @0xdhrv  😊

I checked the existing extension and I think there’s definitely some overlap, but my main goal was a bit different

The current extension seems more focused on usage tracking itself, while mine is mainly about tracking AI spending and understanding costs over time, things like:

- how much money is actually being spent,
- trends by day/month,
- breakdown by model/provider,
- budget monitoring,
- and later maybe alerts or spending insights.

So it’s more of an “AI costs dashboard” than just token usage tracking

To be honest, I mainly called it ai-usage-dashboard because I thought the naming would be more discoverable/searchable in the store 😅

I’d be totally okay with changing the name if that helps avoid confusion or overlap

0xdhrv:
Hey @dawidmatyjasik, apologies, this seems to have gotten lost in the busy times.

Thanks for clarifying the direction.

The distinction makes sense: you’re positioning this more as an AI cost dashboard, focused on spend, trends over time, and model/provider breakdowns, rather than only raw usage tracking.

From the Store perspective, though, we still see the core user intent as overlapping with the existing `agent-usage` extension: users looking to understand AI/agent usage and related costs in Raycast would likely encounter both extensions for a very similar purpose. Even if the emphasis is different, having two extensions in the same space can make discoverability harder and split related functionality across multiple places.

Renaming could reduce some confusion, but it would not fully address the overlap in the underlying workflow. Our preferred path is still to consolidate this into the existing extension rather than introduce a parallel one.

A possible next step would be to explore adding the cost-focused views and breakdowns to `agent-usage`, potentially as a larger update if needed. We’re happy to help reach out to the current maintainer and see whether they’d be open to collaborating on that.

PR #27773, opened 2026-05-10
Copy Path: Support QSpace Pro

arjanflac:
## Summary
- add QSpace Pro as a supported file manager in Copy Path
- read selected items and the current location through QSpace Pro's AppleScript interface
- support selected items on QSpace Desktop and keep the existing multi-path separator and tilde preferences

## Testing
- `npm run lint`
- `npm run build`
- Tested locally in Raycast development mode with QSpace Pro frontmost, including copying the current folder from a QSpace window.


0xdhrv (approved):
Looks good to me, approved ✅

PR #27814, opened 2026-05-12
[codex] Fix stale Skills search results

keito4:
## Summary

- guard debounced Skills search state updates so only the latest query can update results, errors, and loading state
- clear stale results when the current search fails
- abort in-flight search requests when the query becomes too short

## Why

A failed search after a previous successful search could leave the old results in state. The UI would then render stale results under the newer query text instead of showing the error state.

## Validation

- npm run lint
- npm run build


0xdhrv (approved):
Looks good to me, approved ✅

PR #27816, opened 2026-05-12
Update brew extension

alexi-build:
### [Add Keyboard Shortcuts]

- Added keyboard shortcuts to common Brew actions, including opening package pages, opening homepages, copying URLs, and running terminal commands.
- Standardized shortcut usage with Raycast common shortcuts where appropriate.
## Checklist

### DIFF
--- a/extensions/brew/src/components/actionPanels.tsx
+++ b/extensions/brew/src/components/actionPanels.tsx
@@ -1,4 +1,4 @@
-import { Action, ActionPanel, Detail, Icon } from "@raycast/api";
+import { Action, ActionPanel, Detail, Icon, Keyboard } from "@raycast/api";
 import {
   brewInstallCommand,
   brewInstallPath,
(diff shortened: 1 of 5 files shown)


0xdhrv (approved):
Looks good to me, approved ✅

extensions/brew/src/components/actionPanels.tsx
```diff
@@ -123,16 +136,33 @@ export function CaskActionPanel(props: {
           <Action
             title={`Run Install in ${terminalName}`}
             icon={terminalIcon}
+            shortcut={{ modifiers: ["cmd"], key: "return" }}
```
nhojb, line 139: Might be better to make all the terminal actions the same and use `Cmd+Shift+Return` rather than a mix?

extensions/brew/src/components/actions.tsx
```diff
@@ -67,6 +67,7 @@ export function FormulaUpgradeAllAction(props: { onAction: (result: boolean) =>
     <Action
       title="Upgrade All"
       icon={Icon.Hammer}
+      shortcut={{ modifiers: ["cmd", "opt"], key: "u" }}
```
nhojb, line 70: This should be `Cmd+Return` (secondary action)

And _Upgrade_ should be `Return` (primary action)

nhojb: Same applies to the _Install_ action for the search command. It should use `Return` and not `Cmd+I`.

nhojb (commented):
Thanks for the contribution!

I notice this is already merged, but I added a few possible further improvements which you could address in a separate PR?

PR #27846, opened 2026-05-13
Add prompt-launcher extension

uchimanajet7:
## Description

Prompt Launcher is a Raycast extension for using Markdown files as reusable prompts from local folders.

It lets users configure up to three Prompt Sets, search all enabled Prompt Sets, preview prompt files, copy raw content, copy content with supported Dynamic Placeholders, and open selected files in their preferred editor.

Prompt files stay in the user's existing local folders, so users can continue editing and versioning them with their usual tools.

## Screencast

prompt-launcher-1
prompt-launcher-2
prompt-launcher-3

## Checklist


uchimanajet7:
Hi Raycast team,

I just wanted to follow up on this PR.

I have addressed the previous review comments, and I believe the requested changes have been completed. The Greptile review threads are resolved, and the latest checks appear to be passing.

Could you please take another look when you have a chance?

Thank you for your time and for maintaining the Raycast extensions repository.

0xdhrv (commented):

We noticed that there is already an extension in the Store that serves a very similar purpose:

* [Prompt Stash](https://www.raycast.com/renzo/prompt-stash) by @renzo4web

Both extensions appear to focus on managing, organizing, and quickly accessing prompt libraries from Raycast. One interesting aspect of this PR is the support for local files as a source of prompts, which could be a valuable addition for users who prefer to keep their prompts under version control or synced through their existing filesystem workflows.

Given that Prompt Stash already has a significant user base, have you considered whether local file support and any other unique capabilities introduced here could be added to the existing extension instead? This would allow current users to benefit from the new functionality while keeping prompt management consolidated in a single, more discoverable extension.

If there are broader differences in workflow or goals beyond local file support, we'd love to hear more about them so we can better understand how this extension differs from the existing offering.

uchimanajet7:
@0xdhrv 
Thanks for the thoughtful review.

I understand the concern about overlap with Prompt Stash. I looked into Prompt Stash again and I agree that, at a high level, both extensions are related to quickly accessing reusable prompts from Raycast.

However, I think the core product model and target workflow are different.

Prompt Stash appears to be a Raycast-native prompt manager. Users create prompts inside Raycast, store them in the extension, organize them with tags/favorites, and manage the prompt collection from Raycast. It also supports import/export for the extension’s own prompt data.

Prompt Launcher is intentionally not a prompt manager. It does not create, edit, tag, favorite, import, export, or store prompts in Raycast-specific storage. Instead, it treats existing local Markdown files as the source of truth and uses Raycast only as a fast search, preview, copy, and open layer on top of those files.

The main difference is the source of truth:

- Prompt Stash: the prompt database lives inside the Raycast extension.
- Prompt Launcher: the prompt library already exists as local Markdown files, and Raycast does not own or duplicate that data.

This is why I initially created Prompt Launcher as a separate extension rather than adding the feature to Prompt Stash. Adding local Markdown files as a source of truth would not just be a small additional action; it would introduce a second data model and a different workflow into Prompt Stash:

- Prompt Stash organizes prompts with app-level metadata such as tags and favorites.
- Prompt Launcher organizes prompts through the filesystem: folders, filenames, relative paths, and Git history.
- Prompt Stash is optimized for users who want to manage prompts inside Raycast.
- Prompt Launcher is optimized for users who already manage prompts outside Raycast and only want Raycast to launch/copy/open them.
- Prompt Stash’s import/export flow moves data into or out of the extension.
- Prompt Launcher avoids migration entirely because the Markdown files remain the canonical data.

For users who maintain prompt repositories, use Git history/code review, sync files through existing filesystem tools, or edit prompts in VS Code/Obsidian/another editor, importing those prompts into a Raycast-native database would create a second source of truth. Prompt Launcher is meant to avoid that.

That said, I agree that the current README, extension description, and Store metadata may not make this distinction clear enough. If the separate-extension direction makes sense, I can update them to emphasize that Prompt Launcher is a local Markdown prompt file launcher, not a Raycast-native prompt manager. I can also add a short “How this differs from prompt managers” section focused on the file-based source-of-truth workflow, Git/editor usage, and the lack of Raycast-specific prompt storage.

Before I make further changes, would this distinction be enough to justify a separate focused extension, or would you prefer that I first discuss adding local Markdown file support to Prompt Stash with its maintainers?

uchimanajet7:
Thanks again. I’ve left the PR in its current state and marked it ready for review so the Raycast team can advise on the direction.

At this point, I’m looking for guidance on whether the file-based source-of-truth distinction is enough to justify a separate focused extension, or whether I should first discuss adding local Markdown file support to Prompt Stash with its maintainers.

uchimanajet7:
Thanks for the review.

After re-evaluating the positioning, I agree that the current Prompt Launcher name and description make the extension look too close to a prompt manager.

The extension’s actual purpose is broader and more specific: it helps users find local Markdown-backed text blocks they already manage and copy them for reuse, while keeping those files as the source of truth.

Because this is a product positioning change rather than a small metadata update, I’m going to close this PR and submit a new one under a clearer name: Local Copy Blocks.

PR #27876, opened 2026-05-14
Add QuickCite extension

michaelmjhhhh:
## What this extension does

QuickCite converts a DOI or article URL into an MLA-style citation from Raycast. It is designed for students, researchers, and writers who need a fast citation draft without leaving their keyboard.

## Features

- Accepts DOI values and article URLs.
- Extracts citation metadata from DOI.org, Crossref, and page metadata where available.
- Formats the result as an MLA-style citation.
- Provides a polished Raycast form with clearer states, validation, colorful but restrained accessories, and readable result details.
- Includes actions to copy the citation and continue working quickly.

## Implementation notes

- Uses Raycast APIs for the command UI, form validation, result rendering, actions, icons, toasts, and clipboard behavior.
- Keeps dependencies minimal; `@raycast/utils` was removed because the extension does not import it.
- Uses the latest two store screenshots in `metadata/`.
- Legacy screenshot files were removed to keep the extension package tidy.

## Checklist

- [x] I checked that files in the assets folder are used by the extension itself.
- [x] I removed unused dependencies.
- [x] I ran lint/build/type checks locally.

## Screenshots

QuickCite input form

QuickCite generated citation


0xdhrv (commented):

We noticed that there is already an extension in the Store that appears to cover a very similar use case:

* [Citation Generator](https://www.raycast.com/AbhinavPalacharla/citation-generator) by @AbhinavPalacharla

Both extensions focus on generating MLA-style citations from a URL. QuickCite adds some useful improvements, such as DOI support, richer metadata extraction, and multiple output formats, but the core workflow still appears to overlap with the existing Citation Generator extension.

Since the existing extension already serves this area, could we consider extending it with these additional capabilities instead? Features like DOI support, Crossref/doi.org metadata lookup, HTML metadata extraction, and Markdown/HTML/plain-text output could be valuable improvements for existing users while keeping citation generation consolidated in one place.

If there are unique workflows or technical differences that make this a separate extension, we’d love to better understand them. Otherwise, contributing these improvements to Citation Generator would likely help avoid duplication and improve discoverability.

PR #27894, opened 2026-05-14
Update Portless: add custom executable path preference

gravitinos:
## Description

The extension's `getExtendedPath` only knows how to find Node binaries installed under `~/.nvm`. Users on other version managers (fnm, volta, asdf, mise, …) see "No active routes" even though `portless list` works fine in their terminal — Raycast spawns commands with a minimal PATH that doesn't include the version manager's bin dir, so the `portless` binary can't be located.

I hit this on fnm. Rather than add per-tool detection logic, this PR mirrors the [`brew` extension's `customBrewPath` pattern](https://github.com/raycast/extensions/blob/main/extensions/brew/package.json#L26-L32) and adds a simple textfield preference for an explicit binary path. The description tells users to run `which portless` to find it.

Existing nvm + homebrew auto-detection in `getExtendedPath` is unchanged, so this is purely additive — users on supported setups see no change.

## Changes

- **`package.json`**: add `customPortlessPath` textfield preference.
- **`src/list.tsx`**: read the preference; if set (and non-empty after trim), use it as the binary path in `useExec`; otherwise call `"portless"` as before.

3 files changed, 22 insertions(+), 2 deletions(-).

## Checklist

### DIFF
--- a/extensions/portless/src/list.tsx
+++ b/extensions/portless/src/list.tsx
@@ -1,7 +1,11 @@
 import { existsSync, readdirSync } from "fs";
 import { join } from "path";
-import { ActionPanel, Action, Icon, List } from "@raycast/api";
+import { ActionPanel, Action, getPreferenceValues, Icon, List } from "@raycast/api";
 import { useExec } from "@raycast/utils";
+
+interface Preferences {
+  customPortlessPath?: string;
+}
 
 function parsePortlessListOutput(stdout: string): string[] {
   const urls: string[] = [];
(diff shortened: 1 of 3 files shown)


extensions/portless/src/list.tsx
```diff
@@ -15,36 +15,46 @@ function parsePortlessListOutput(stdout: string): string[] {
...
-      }
-    } catch {
-      // Ignore nvm resolution errors
-    }
+// Raycast spawns commands with a minimal PATH that often excludes the user's
+// shell setup, so version managers like nvm, fnm, volta, asdf, and mise are
+// invisible. We resolve PATH by asking the user's login+interactive shell
+// (where those tools install their hooks) and fall back to a broad list of
+// well-known install locations if that fails.
+function getShellPath(): string | null {
```
gravitinos: Maybe this should just be a config? I'll update so maybe we have smart defaults + a config

nbaglivo:
Thanks for taking care of this, @gravitinos! When I first hit this, it felt like something many extensions probably run into. @0xdhrv or anyone from the Raycast team — any thoughts? Could getting the path be a Raycast API/primitive? Or maybe we could ship that utility as a package, if it is actually a common problem

PR #27910, opened 2026-05-15
Add super-copy-paste extension

jmgpp:
## Checklist


extensions/super-copy-paste/README.md
```diff
@@ -0,0 +1,26 @@
+# Super Copy/Paste
+
+Super Copy/Paste is a Raycast extension that lets you apply regular expressions (Regex) sequentially to text when copying or pasting. 
+
+![Manage Rules]
```
pernielsentikaer, line 5: broken image: ![Manage Rules] with no URL

pernielsentikaer (commented):
Hi 👋

Did you check [Raycast Store: Advanced Replace](https://www.raycast.com/rocksack/advanced-replace) by @Gitsack - it feels pretty similar or could be extended with what's missing instead of creating a new one?

extensions/super-copy-paste/README.md
jmgpp: Hi Pern, I din't knew the Advance Replace extension, it shares some concepts for sure but the execution is completely different. Because Raycast extensions can't create commands on runtime I specifically focused on creating 6 commands beforehand that the user can setup and create shortcuts. This completely removes the need interact with the app, just assign the regex you want to "ctrl + alt + C" and that speeds up its use in specific workflows immensely changing it from "nice to have" to essential.

pernielsentikaer (changes requested):
That makes sense 🙂

Some feedback for you:

- Regex flags are hardcoded to gm in engine.ts — users can't opt into case-insensitive (i) or single-line. Fine for v1, but limiting; consider a per-rule flags field later.

- No way to test/preview a rule inside Manage Rules before running it on real text.

Could you look into this?

---

PR #27924, opened 2026-05-15
Add browser extension

AadeshMndr:
## Description

This extension allows us to setup browser profiles in raycast and open website with our preferred choice of profile and browser. 
The search query and profile id are arguments to the function so a quick link for a commonly used argument set can be created.

## How it works

Type your query as the command argument when you launch Browser. The command then shows a profile picker, with your default browser at the top and any saved profiles below it.

image

You can also pass an optional profile argument (for Quicklinks). The picker uses that value as an initial filter and preselects the closest profile match, while still letting you pick another profile from the list.

image

## Profile setup

image

1. Type the query you want to search
2. Type out the name of your profile
3. If you haven't setup your profile, it won't show up yet, so press Enter to set it up
4. Set your Profile name
5. Set the name of the browser you want to use for it
6. Set the profile id for your profile within that browser
7. That's it, for the next prompt, you should see your new profile in the dropdown

image

## Notes

This profile launcher uses the browser app name plus the profile directory string. For Chromium-based browsers, that is usually something like `Default` or `Profile 1`.

## Checklist


0xdhrv (commented):
Thanks for your contribution @AadeshMndr 🔥

We already have an extension in the Store that deals with opening URLs in specific browser profiles. Could we consider enhancing the existing extension below instead of creating another one?

- [Open Browsers Profiles](https://www.raycast.com/skydiver/browsers-profiles) by [@skydiver](https://github.com/skydiver)

`browsers-profiles` auto-discovers profiles from disk rather than requiring manual setup, which is arguably a superior UX — worth highlighting that the existing extension already solves the core "open with profile" need. There are also browser-specific options (google-chrome-profiles, microsoft-edge, vivaldi, iridium) if the author wants to contribute upstream into a specific one.


If there are unique features or workflows you're aiming to add (e.g., the query/profile arguments for Quicklinks, or the in-extension profile editor), we'd love to hear them and see if they can be integrated into this to avoid duplication and improve discoverability.

> This would help avoid duplication and keep related functionality consolidated in one place. As mentioned in our [extension guidelines here ↗](https://manual.raycast.com/extensions#:~:text=There%20is%20an%20active%20extension,extension%20providing%20very%20similar%20value)

AadeshMndr:
Ok sure !

PR #27925, opened 2026-05-15
Add string-unescape extension

gkpln3:
## Description

Easily unescape special characters such as \n or \t in strings.

## Screencast

Screenshot 2026-05-15 at 13 16 09
Screenshot 2026-05-15 at 13 16 03

## Checklist


0xdhrv (commented):

We noticed that there are already extensions in the Store that cover similar text transformation and encoding workflows:

* [Ray Boop](https://www.raycast.com/nathan_schwermann/ray-boop) by @nschwermann
* [Escape RegExp Characters](https://www.raycast.com/HerbertLu/escape-regexp-characters) by @lryong

Both the existing extensions and this one operate in the broader area of transforming, escaping, and unescaping text directly within Raycast. While this extension provides a more focused experience around string escaping and unescaping, the overall use case appears closely related to functionality that is already available today.

Could we consider extending one of the existing extensions with these capabilities instead of creating another extension in the same space? A dedicated escape/unescape workflow could be a valuable addition while keeping related text transformation functionality consolidated in one place.

PR #27929, opened 2026-05-15
[Gitlab] Fix OOM errors on "My Projects"

jessedobbelaere:
## Description

The My Projects command crashes with `Worker terminated due to reaching memory limit: JS heap out of memory` on GitLab instances with thousands of projects. Our instance has ~3800 projects, and the command has been unusable for me since a week. 

```

[kOnExit]:worker:316:26
```

2026-05-15 at 13 25 10@2x

Additionally, for the past year, when the cache gets invalidated (every 7 days) it would take 3-5min for the list of projects to be refetched again. This was always painfully slow for me, and i couldn't use Raycast meanwhile. 

### Root causes

**1. Sequential API pagination** — All pages were fetched one-by-one, accumulating raw JSON responses in memory before mapping to Project objects. With 39 pages of 100 projects, this took 60+ seconds and held all raw data simultaneously.
**2. Rendering all items** — Every project was rendered as a `<List.Item>` with ~18 action components in the ActionPanel. At 3800+ items, this created ~70,000 React elements, exceeding the worker memory limit.
**3. Fuse.js indexing** — Building a Fuse.js search index on 3800+ items consumed too much memory on top of the data and rendered components.

### What this PR does

**- Parallel batch fetching** — Pages are now fetched in parallel batches of 5, with a mapPage callback that transforms raw API responses to slim Project objects per-page (so raw JSON can be garbage collected between pages). Includes a single retry per failed page. Reduces initial fetch from ~60s to ~10-15s.
**- Render cap** — Only the top 100 matching projects are rendered at a time. The full dataset stays in memory for client-side filtering. The section subtitle shows e.g. "100 of 3847" when the cap is active.
**- Lightweight search** — Replaced Fuse.js with a multi-term case-insensitive substring filter for the project list. Typing data pipeline matches Showpad / Data / Pipeline Service. Zero memory overhead. Other commands (epics, events, labels) still use Fuse.js since their datasets are small.
**- Sort by activity** — Projects are now sorted by `last_activity_at` so the most relevant ones appear first, because of the cap of 100 items before search.
**- "Search active projects only" preference** — New checkbox on the My Projects command (consistent with `Search Projects`) to optionally exclude archived projects.
**- Background refetch interval** — Changed from the 5-minute default to 1 day. With thousands of projects, a background refetch takes 20 seconds — too expensive to trigger every 5 minutes. Users can still force-refresh via the existing "Clear Cache" action.
**- Cleanup** — Removed dead getNextPageNumber function, simplified useMyProjects hook.

**Note:** 
- I think the GitHub extension only allows server-side searching, but I didn't want to revamp going from client-side cached entries to full server-side search in the Gitlab extension. I tried to optimize the fuse.js indexing but i kept getting OOM's so reverted to simple substring search... WDYT?
- Another option would be to switch to the **graphql API** to keep the responses lighter only picking the project fields we need? I tried with the `simple=true` url param on the current API endpoint, but it excludes specific fields we need like the archived status. 

With these fixes in place, it currently works great again for me 👍 

## Checklist

### DIFF
--- a/extensions/gitlab/CHANGELOG.md
+++ b/extensions/gitlab/CHANGELOG.md
@@ -1,6 +1,6 @@
 # GitLab Changelog
 
-## [Fix OOM crash in My Projects for large GitLab instances] - {PR_MERGE_DATE}
+## [Fix OOM crash in My Projects for large GitLab instances] - 2026-05-16
 
 - Fix out-of-memory crash when fetching thousands of projects by using parallel batch fetching with per-page mapping
 - Replace Fuse.js search with lightweight multi-term substring matching for the project list to reduce memory usage
--- a/extensions/gitlab/CHANGELOG.md
+++ b/extensions/gitlab/CHANGELOG.md
@@ -1,5 +1,14 @@
 # GitLab Changelog
 
+## [Fix OOM crash in My Projects for large GitLab instances] - {PR_MERGE_DATE}
+
+- Fix out-of-memory crash when fetching thousands of projects by using parallel batch fetching with per-page mapping
+- Replace Fuse.js search with lightweight multi-term substring matching for the project list to reduce memory usage
+- Cap rendered list items to 100 to prevent memory exhaustion from complex ActionPanels
+- Sort projects by last activity date so most relevant projects appear first
+- Add "Search active projects only" preference to My Projects command (consistent with Search Projects)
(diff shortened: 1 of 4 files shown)


0xdhrv (approved):
Looks good to me, approved ✅

PR #27963, opened 2026-05-16
Fix Apple Developer Docs search API

dustPyrotechnic:
## Description

- replace the deprecated Apple Developer search_data.php endpoint with the current search API
- normalize documentation, devsite, and developer search results into the existing Raycast result shape
- keep the existing result type filter support for documentation, sample code, video, and general results

## Validation

- npm run build
- npm run lint

### DIFF
--- a/extensions/apple-developer-docs/CHANGELOG.md
+++ b/extensions/apple-developer-docs/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Apple Developer Docs Changelog
 
-## [Fix Search API] - {PR_MERGE_DATE}
+## [Fix Search API] - 2026-05-18
 - Updated the Apple Developer search request to use the current production endpoint.
 
 ## [Search History] - 2022-08-02
--- a/extensions/apple-developer-docs/CHANGELOG.md
+++ b/extensions/apple-developer-docs/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Apple Developer Docs Changelog
 
+## [Fix Search API] - {PR_MERGE_DATE}
+- Updated the Apple Developer search request to use the current production endpoint.
+
 ## [Search History] - 2022-08-02
 - Added search history for searched items.
 - Reduced the often-truncated description clutter for list items.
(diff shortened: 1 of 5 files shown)


0xdhrv (commented):
Hi @dustPyrotechnic 👋

Thanks for the contribution 🔥

---

dustPyrotechnic:
Updated the PR to address the review feedback:

- Removed the duplicate `?q=` query string and now send the search term only in the POST body as `text`, matching the current Apple search page request.
- Added a changelog entry for the extension fix.
- Verified the endpoint against the production `https://developer.apple.com/search/scripts/search.js`, which calls `https://devintserv.msc.sbz.apple.com/api/v1/search` from `fetchSearchResults`.

Validation:
- `npm run build`
- `npm run lint`

dustPyrotechnic:
Updated again based on the latest Greptile suggestions:

- Added `q` back to the `useFetch` URL so search text changes trigger a new request. The API request still sends the actual search term via the POST body `text`, matching Apple’s production script.
- Added an inline comment explaining that the URL query is only the `useFetch` revalidation key.
- Changed the changelog entry date to `{PR_MERGE_DATE}`.

Validation:
- `npm run build`
- `npm run lint`

dustPyrotechnic:
Handled the latest Greptile suggestion:

- Updated `first()` to return `undefined` for missing values so `createSearchResult` fallbacks like `config.rootUrl` and `Untitled` can apply correctly.

Validation:
- `npm run build`
- `npm run lint`

dustPyrotechnic:
Handled the latest Greptile suggestions:

- Normalized both sides of the searched-history type filter comparison so stored values like `General` still match the `general` filter.
- Added `results: config.maxResults` to the Apple search POST body to preserve the previous result limit behavior when supported by the API.
- Rebasing onto the latest `main` is included in this update.

Validation:
- `npm run build`
- `npm run lint`

0xdhrv (approved):
Looks good to me, approved ✅

PR #27976, opened 2026-05-16
Claude/fix figma search extension vi2 cu

egorchuzhin:
## Description

Fixed 4 bugs in the Figma File Search extension:

**1. Cache data loss on partial refresh**
When the cache TTL expired for only some projects, `resolveAllFiles()` was overwriting the entire cache with partial data, causing files from other teams/projects to disappear. Fixed by merging refreshed project data with the existing cache at the project level.

**2. Starred file detection by name instead of key**
`FileGridItem` was comparing `item.name === file.name` to check if a file is starred. This caused false matches when multiple files shared the same name. Fixed to use `item.key === file.key`.

**3. Broken project section title when files are undefined**
The condition `project.files?.length != 0` evaluates to `true` when `project.files` is `undefined` (JavaScript quirk: `undefined != 0`), causing sections to render with the title "(undefined Files)" and no items inside. Fixed with `(project.files?.length ?? 0) > 0`.

**4. User profile not shown in Raycast**
`OAuthService` was missing an `onAuthorize` callback, so `setUserInfo` was never called after login. Raycast had no user data to display in extension preferences. Fixed by fetching `/v1/me` from the Figma API on authorize and passing the user's name and avatar to `setUserInfo`.

## Checklist

- [x] Tested the extension locally
- [x] No new dependencies added
- [x] Existing functionality preserved

### DIFF
--- a/extensions/figma-files/src/api.ts
+++ b/extensions/figma-files/src/api.ts
@@ -102,7 +102,7 @@ async function fetchFiles(): Promise<ProjectFiles[][]> {
               // Update TTL for successfully fetched project
               await updateProjectTTLs([project.id]);
 
-              return { name: project.name, files: result.files ?? [] };
+              return { id: project.id, name: project.name, files: result.files ?? [] };
             }),
           );
           return projects;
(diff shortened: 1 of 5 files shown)


michaelwschultz:
> Adds an onAuthorize callback that fetches /v1/me from the Figma API and calls setUserInfo to populate the user profile in Raycast; network failures are silently swallowed so they don't degrade the OAuth flow.

This won't work because the scope is not available in our oauth setup. I can change this though.

The rest seems good to me. Not sure how many people are still using this extension due to the massive changes to the Figma API call limit. 🤷

michaelwschultz:
I'm submitted an update to our oauth permissions.

0xdhrv (approved):
Looks good to me, approved ✅

PR #27994, opened 2026-05-17
[ClaudeCast] Preserve session cwd when resolving project paths

### LINKED ISSUE #27904: [ClaudeCast] Bug: Paths starting with an underscore (_) are incorrectly replaced with a slash (/) in past sessions
### Extension

https://www.raycast.com/qazi0/claudecast

### Raycast Version

1.104.17

### OS Version

26.4

### Description

In the "past sessions" history, file or directory paths that originally start with an underscore (e.g., `_a`) are incorrectly parsed and displayed as starting with a forward slash (e.g., `/a`). 

This causes severe path resolution issues. For example, a path that should be `/_a` gets converted into `//a`. Because the path is altered, the extension can no longer locate the correct files, breaking the functionality for those specific past sessions.

### Steps To Reproduce

1. Create or use a session that involves a file or directory path containing an underscore prefix (e.g., `/_folderName` or `_a`).
2. Open the ClaudeCast extension and navigate to the past sessions list.
3. Locate the session with the aforementioned path.
4. Observe that the underscore `_` has been replaced by a slash `/` (resulting in something like `//folderName`).
5. Attempt to interact with or load the session, which will fail because the path no longer exists.

### Current Behaviour

Any path segment starting with `_` is incorrectly rendered as starting with `/`. This frequently results in double slashes (e.g., `//a` instead of `/_a`), making the path invalid and the session unfindable.

### Expected Behaviour

_No response_

mturac:
## Description

Fixes #27904.

Claude session project directory names are lossy: path separators, dots, and underscores can all be encoded as dashes. When a session JSONL entry includes the original `cwd`, use that authoritative path for session metadata/details before falling back to the existing encoded-directory resolver. This prevents paths with underscore-prefixed segments from being displayed or resumed as slash-prefixed paths.

## Screencast

N/A - parser-only bug fix.

## Verification

- `npm run lint`
- `npm run build`
- `git diff --check`

## Checklist

- [ ] I read the extension guidelines
- [ ] I read the documentation about publishing
- [ ] I ran `npm run build` and tested this distribution build in Raycast

Note: I ran the distribution build successfully, but did not manually load the build in Raycast.

### DIFF
--- a/extensions/claudecast/CHANGELOG.md
+++ b/extensions/claudecast/CHANGELOG.md
@@ -10,6 +10,7 @@
 
 ### Fixed
 
+- **Session Path Resolution**: Session metadata, details, and search results now prefer the session JSONL `cwd` field when present, avoiding lossy fallback decoding for project paths with underscore-prefixed segments.
 - **Cost Calculation: Streaming Chunk Deduplication**: Anthropic streams response chunks where each chunk's `usage` is cumulative. Naive summing inflated session totals by 2x to 4x. Now deduped by `(message.id, requestId)` so per-message totals reflect the final cumulative value once per request.
 - **Cost Calculation: Sonnet 200K Token Tier**: Above 200K input tokens per message, Sonnet rates double across all token types (input, output, cache read, cache write). The pricing now applies a flat per-request high tier keyed off the message's input token count, matching Anthropic's billing.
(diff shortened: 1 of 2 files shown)


mturac:
Addressed the search path case as well: search results now also prefer the session JSONL cwd before falling back to encoded directory resolution. Added a changelog entry and reran lint/build.

qazi0 (approved):
lgtm🚀
cc @0xdhrv

pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #28003, opened 2026-05-17
[Visual Studio Code] Resolve Code installs in user Applications

### LINKED ISSUE #27737: [Visual Studio Code] macOS app installation in any  directory for resources/app/product.json resolve bug
### Extension

### Raycast Version

1.104.15

### OS Version

macOS. 15.7.3

### Description

My Code app is installed under $HOME/Applications, and the Code installation package is a ZIP archive. After extraction, it should theoretically be possible to place it in any directory, so the paths define 
`'Code: "/Applications/Visual Studio Code.app/Contents/Resources/app/bin/code" ` to resove resources/app/product.json  is not working, and cause the  empty recent projects problem. 

### Steps To Reproduce
macOS

Open Search recent projects
No Results

#### Current Behaviour
No Results

#### Expected Behaviour
Show some folders

mturac:
## Description
Fixes #27737.

The extension resolved the Code build from `/Applications/Visual Studio Code.app`, so `product.json` could not be found when VS Code was installed from the zip into `~/Applications`. This adds a small macOS path resolver for the Code build that checks both `/Applications` and `~/Applications`, while keeping the existing `/Applications` path as the fallback.

## Screencast
N/A; path resolution fix.

## Testing
- npm run lint --prefix extensions/visual-studio-code-recent-projects
- npm run build --prefix extensions/visual-studio-code-recent-projects
- git diff --check


pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #28016, opened 2026-05-17
[Goodreads] Restore book search results

### LINKED ISSUE #27847: [Goodreads] Fails to show book results
### Extension

https://www.raycast.com/puneeth/goodreads

### Raycast Version

0.58.0.0

### OS Version

Windows 11 Core 10.0.26200.0 amd64

### Description

When attempting to search for a book using the extension, nothing shows up.  I am not sure if it requires an API token, but without one, it should still work right?

### Steps To Reproduce

1. Type alias or hotkey if any is set
2. Search for a book
3. Nothing shows up

### Current Behaviour

Nothing shows up, no error message or anything

### Expected Behaviour

Results of the book search

mturac:
## Description
- Switch book search to Goodreads auto-complete JSON results because the HTML search page now returns a WAF challenge with an empty body.
- Map the JSON response into the existing book result shape so details still open through the current Goodreads pages.

Fixes #27847

## Screencast
N/A

## Checklist
- [x] I read the contribution guidelines
- [x] I ran `npm run lint --prefix extensions/goodreads`
- [x] I ran `npm run build --prefix extensions/goodreads`
- [x] I ran `git diff --check`
- [x] I verified `https://www.goodreads.com/book/auto_complete?format=json&q=dune` returns book results

### DIFF
--- a/extensions/goodreads/CHANGELOG.md
+++ b/extensions/goodreads/CHANGELOG.md
@@ -1,6 +1,6 @@
 # goodreads Changelog
 
-## [Bugfix] - {PR_MERGE_DATE}
+## [Bugfix] - 2026-05-21
 
 - Restore book search results.
 
--- a/extensions/goodreads/CHANGELOG.md
+++ b/extensions/goodreads/CHANGELOG.md
@@ -1,5 +1,9 @@
 # goodreads Changelog
 
+## [Bugfix] - {PR_MERGE_DATE}
+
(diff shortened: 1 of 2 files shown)


shettypuneeth:
@mturac Thanks for the fix. LGTM.

0xdhrv (approved):
Looks good to me, approved ✅

PR #28027, opened 2026-05-17
zed-recent-projects: Fix non-POSIX shell detection

FollowTheProcess:
## Description

Fixes the non-POSIX shell detection in the `zed-recent-projects` to correctly classify more than just `fish` as non POSIX. I use [nushell](https://www.nushell.sh) and this extension doesn't open Zed when I hit enter on the project in the picker.

It didn't open Zed but also didn't flag an error or anything so I just assumed it must just not work or maybe Zed changed since this was released or something but I got curious and on looking into it, it's easily fixable :)

I also added an explicit error when it fails to open in the `openSingleFolder` case, mirroring the same approach as the `openMultiFolder` above

### Build and Run

image

image

### Picking the Extensions Repo

image

### And here it is!

image

## Checklist

### DIFF
--- a/extensions/zed-recent-projects/CHANGELOG.md
+++ b/extensions/zed-recent-projects/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Zed Recent Projects Changelog
 
-## [Fix Non-POSIX Shell Support] - {PR_MERGE_DATE}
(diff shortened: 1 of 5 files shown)


0xdhrv (approved):
Looks good to me, approved ✅

PR #28035, opened 2026-05-17
Fix react-native-directory extension

shubh73:
## Description

Fixes React Native Directory in Raycast Beta: https://x.com/shubhporwal24/status/2055901712154849670?s=20

Raycast Beta crashes when a `List.Item` subtitle is passed as a tooltip-only object. This update passes a valid `{ value, tooltip }` subtitle shape and only shows row descriptions when the detail pane is collapsed, since the expanded detail view already displays the selected library description.

Also updates Raycast dependencies to pinned current versions and shows a loading state while package author metadata is fetched.

## Screencast

Raycast 2026-05-17 18 46 21

## Checklist

### DIFF
--- a/extensions/react-native-directory/.eslintrc.json
+++ b/extensions/react-native-directory/.eslintrc.json
@@ -1,4 +1,0 @@
-{
-  "root": true,
-  "extends": ["@raycast"]
-}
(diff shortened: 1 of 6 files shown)


shubh73:
api/runtime mismatch i noticed while fixing this:

`List.Item.subtitle` currently allows an object where both fields are optional:

```ts
subtitle?: string | {
  value?: string | null;
  tooltip?: string | null;
}
```

so this type-checks:

```tsx
subtitle={{ tooltip: "description" }}
```

but in raycast beta, it crashes with react error: `object with keys {tooltip}`

0xdhrv (approved):
Looks good to me, approved ✅

PR #28049, opened 2026-05-18
[Downloads Manager] Add "Copy Path" action to Manage Downloads

### LINKED ISSUE #28048: [Downloads Manager] Add "Copy Path" action to Manage Downloads
### Extension
https://www.raycast.com/thomas/downloads-manager

### Description
"Manage Downloads" already has a "Copy File" action, but that copies the file itself (so you can paste it into Finder etc.), not the path. There's no way to get the absolute path as a string. When I want to paste a download's path into a terminal or editor, I currently have to drop into Finder and use ⌥⌘C.

A "Copy Path" action on the focused download would cover that case.

Anyone who pastes download paths into a terminal, editor, or script.

### Anything else?
_No response_

fdarian:
## Summary
Adds a "Copy Path" action to the Manage Downloads command. It copies the focused download's absolute path as plain text, separate from the existing "Copy File" action (which puts the file itself on the clipboard).

Shortcut: `⌘⇧.` on macOS, `Ctrl+Shift+.` on Windows.

Resolves #28048

## Screencast / Screenshots
_n/a — text-only action change_

## Test plan
- [ ] Focus a file in Manage Downloads, run "Copy Path" → clipboard contains the absolute path as a string
- [ ] Focus a folder, run "Copy Path" → clipboard contains the folder path
- [ ] `⌘⇧.` triggers the action
- [ ] Existing "Copy File" still copies the file itself (regression check)

### DIFF
--- a/extensions/downloads-manager/CHANGELOG.md
+++ b/extensions/downloads-manager/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Downloads Manager Changelog
 
-## [Add Copy Path action] - {PR_MERGE_DATE}
+## [Add Copy Path action] - 2026-05-18
 
 - Added a "Copy Path" action to the Manage Downloads command that copies the focused download's absolute path as text (shortcut: `⌘⇧.` on macOS, `Ctrl+Shift+.` on Windows).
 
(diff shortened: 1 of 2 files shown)


pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #28055, opened 2026-05-18
[1Password] Reduce memory usage for large item lists

### LINKED ISSUE #27540: [1Password] Out of Memory when above 10.000 Passwords
### Extension

### Raycast Version

1.10.4.15

### OS Version

sequioa 15.7.5 (32GB RAM)

### Description

When opening "My passwords" i get the error

[kOnExit]:worker:316:26

This has happened after migration of many passwords (+10.000)

### Steps To Reproduce

Have above 10.000 entries (not sure what the exact limit is) and open Raycast / My Passwords.

### Current Behaviour

It shows the first passwords, then dies after a second or two.

### Expected Behaviour

All passwords should load in / be searchable.

mturac:
## Description

Fixes #27540.

Adds an opt-in **Reduce item list memory usage** path for 1Password item lists with very large vaults. When enabled, the list uses the smaller summary payload from `op item list` instead of `--long`, avoids keeping an extra derived item-list state array, filters across the full result set, and renders the first 200 matching items to keep Raycast worker memory bounded.

Full item details are still fetched lazily when an action needs them. For example, **Open in Browser** fetches the full login item before opening the website if the lightweight list item does not include a URL.

In reduced-memory mode, username/email subtitles and username/email search are unavailable until a full item is fetched by an action. The changelog calls out that tradeoff.

Validation:
- `npm run lint`
- `npm run build`
- `git diff --check`
- Latest extension CI job is green

## Screencast

Not included; this is a memory-usage fix for large local item lists.

## Checklist

### DIFF
--- a/extensions/1password/CHANGELOG.md
+++ b/extensions/1password/CHANGELOG.md
@@ -1,5 +1,9 @@
 # 1Password Changelog
 
+## [Bug Fix] - {PR_MERGE_DATE}
+
+- Reduce memory usage when loading large item lists.
+
 ## [Enhancements] - 2026-04-16
 
 - Improved search: queries now match across all item fields (title, username/email, URLs, vault name). For example, searching "m@ goo" now finds a Google login with email "m@example.com".
--- a/extensions/1password/CHANGELOG.md
+++ b/extensions/1password/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## [Bug Fix] - {PR_MERGE_DATE}
 
-- Reduce memory usage when loading large item lists while preserving list subtitles and search metadata.
+- Reduce memory usage when loading large item lists by avoiding the full item payload until it is needed.
 
 ## [Enhancements] - 2026-04-16
 
--- a/extensions/1password/CHANGELOG.md
+++ b/extensions/1password/CHANGELOG.md
@@ -2,8 +2,8 @@
 
 ## [Bug Fix] - {PR_MERGE_DATE}
 
-- Reduce memory usage when loading large item lists by avoiding the full item payload until it is needed.
-- The item list now uses the summary payload, so username/email subtitles and username/email search are unavailable until full item details are fetched by an action.
+- Reduce memory usage when loading large item lists by avoiding the full item payload until it is needed. Opt in via **Reduce item list memory usage** in extension preferences.
+- When enabled, the item list uses the summary payload, so username/email subtitles and username/email search are unavailable until full item details are fetched by an action.
 
 ## [Enhancements] - 2026-04-16
 
--- a/extensions/1password/CHANGELOG.md
+++ b/extensions/1password/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [Bug Fix] - {PR_MERGE_DATE}
 
 - Reduce memory usage when loading large item lists by avoiding the full item payload until it is needed.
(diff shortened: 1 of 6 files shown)


pernielsentikaer (changes requested):
Hi 👋

What about making this a preference instead? So if it fails, we'll show some information about this and tell the user to enable the new feature in settings, so it will stay as it is now for most of the users.

---

mcinicola:
@mturac please have look at the suggestion from Per

pernielsentikaer (commented):
Would you be up for giving this a test @mcinicola

jkofod:
Hi @pernielsentikaer 
I just tested by enabling "Reduce item list memory usage", then opening My Password in Raycast. It seems there are still issues.
```
~/Documents/git/extensions/extensions/1password fix/27540-1password-memory ❯ npm run dev                                                                                  4m 51s

> dev
> ray develop

info  - entry points ["src/item-list.tsx","src/renew-auth.ts","src/vault-list.tsx","src/generate-password.tsx"]
info  - compiled entry points
info  - generated extension's TypeScript definitions
ready  - built extension successfully
07:24:37

[kOnExit]:worker:316:26
```

image

Note: Im not a developer :) So I may have tested this the wrong way. Please let me know!

jkofod:
Hi @mturac,

Again, I'm not a developer, but I used gemini to propose a fix. It works locally for me after testing. Apologies if this is a bad solution! I'm really just trying to make my collegues' work lives easier by getting this fixed asap :)

Gemini said:
----
We can completely resolve the memory issue and drastically improve keypress filter performance by **slicing the rendered list to 200 items** and **optimizing the search URL parser using a regex** (to avoid massive `new URL()` allocations on every keypress).

Here is the suggested fix for `extensions/1password/src/v8/components/Items.tsx`:
```js
function getHostname(href: string): string {
  try {
    const match = href.match(/^(?:https?:\/\/)?(?:[^@\n]+@)?(?:www\.)?([^:/\n?]+)/i);
    return match ? match[1] : href;
  } catch {
    return href;
  }
}

function getSearchableStrings(item: Item): string[] {
  const strings = [item.title];
  if (item.additional_information) strings.push(item.additional_information);
  if (item.urls) {
    for (const url of item.urls) {
      strings.push(getHostname(url.href));
    }
  }
  if (item.vault?.name) strings.push(item.vault.name);
  return strings;
}
```

And update the render block to slice the passwords array:
```js
<List.Section subtitle={${passwords.length}} title="Items">
  {passwords.length
    ? passwords.slice(0, 200).map((item) => (
        <List.Item
```

mturac:
Thanks for testing this. I pushed a follow-up that keeps the memory reduction behind the preference, filters across the full result set, then renders only the first 200 matching items in that mode. It also avoids allocating URL objects while matching hostnames in the lightweight path.

The extension lint/build pass locally, and the extension CI job is green on the latest commit. A retest with **Reduce item list memory usage** enabled would be very helpful when you have a chance.

jkofod:
@mturac I tested your change and it works as expected locally.

mturac:
Thanks for testing this, @jkofod.

@pernielsentikaer the reduced-memory path is now behind the opt-in preference, and the follow-up caps rendering in that mode while keeping filtering across the full result set. The extension CI is green, and the latest local retest above confirms the preference path works as expected. Ready for another look whenever you have bandwidth.

pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #28072, opened 2026-05-18
[Visual Studio Code] Open Windows projects through editor URLs

### LINKED ISSUE #27056: [Visual Studio Code] 'Open in Code' opens folder in Explorer
### Extension

### Raycast Version

0.53.0.0 (Windows)

### OS Version

25H2 (26220.8148)

### Description

Since today or yesterday, when using the _Open in Code_ command from the extension, projects won't open in VS Code anymore, but instead, their folder is opened in an Explorer window. 

### Steps To Reproduce

1. Run 'Open in Visual Studio Code' on a Windows computer.
2. Pick a project.
3. See the extension open the project folder in Explorer.

### Current Behaviour

_No response_

### Expected Behaviour

_No response_

### LINKED ISSUE #27061: Open in VS Code is not working [Visual Studio Code]
### Extension

### Raycast Version

0.53.0.0

### OS Version

Windows 11 CoreSingleLanguage 10.0.26200.0 amd64

### Description

when i select a folder and press `enter` it doesn't directly open in vs code ? 

### Steps To Reproduce

just search for : search recent project, click any project, press enter

Image

### Current Behaviour

_No response_

### Expected Behaviour

_No response_

mturac:
## Description
- Launches Windows recent projects through the selected editor URL handler instead of Raycast opening the folder target directly in Explorer.
- Uses the existing VS Code build scheme mapping, so Code, Insiders, Cursor, Windsurf, and other configured builds stay aligned.
- Removes the Windows path through the Finder selection API in the Open With VS Code command.

Fixes #27056
Fixes #27061
Fixes #23628

## Validation
- npm run lint
- npm run build
- git diff --check

### DIFF
--- a/extensions/visual-studio-code-recent-projects/CHANGELOG.md
+++ b/extensions/visual-studio-code-recent-projects/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Visual Studio Code Changelog
 
-## [Fix: Windows project opening] - {PR_MERGE_DATE}
+## [Fix: Windows project opening] - 2026-06-17
 
 - Fixed Windows project opening so recent projects are launched through the selected editor URL handler instead of Explorer.
 
--- a/extensions/visual-studio-code-recent-projects/CHANGELOG.md
+++ b/extensions/visual-studio-code-recent-projects/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Visual Studio Code Changelog
 
+## [Fix: Windows project opening] - {PR_MERGE_DATE}
+
+- Fixed Windows project opening so recent projects are launched through the selected editor URL handler instead of Explorer.
+
 ## [Fix: macOS user Applications path] - 2026-06-05
 
 - Fixed `product.json` resolution when Visual Studio Code is installed in `~/Applications`.
--- a/extensions/visual-studio-code-recent-projects/src/index.tsx
+++ b/extensions/visual-studio-code-recent-projects/src/index.tsx
@@ -37,6 +37,7 @@ import {
   isValidHexColor,
   isWin,
   isWorkspaceEntry,
+  openPathInVSCode,
 } from "./lib/utils";
 import { Shortcut } from "./lib/shortcuts";
 import { getEditorApplication } from "./utils/editor";
@@ -209,7 +210,11 @@ function LocalItem(
         }
       }
 
-      open(isWin ? path : props.uri, editorApp);
+      if (isWin) {
+        await openPathInVSCode(path);
+      } else {
+        open(props.uri, editorApp);
+      }
     };
   };
 
--- a/extensions/visual-studio-code-recent-projects/src/index.tsx
+++ b/extensions/visual-studio-code-recent-projects/src/index.tsx
@@ -211,7 +211,11 @@ function LocalItem(
       }
 
       if (isWin) {
-        await openPathInVSCode(path);
+        if (isWorkspaceEntry(props.entry)) {
+          open(props.uri, editorApp);
+        } else {
+          await openPathInVSCode(path);
+        }
       } else {
         open(props.uri, editorApp);
       }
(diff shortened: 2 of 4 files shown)


yusifaliyevpro:
Hi @mturac, this was tried before. Since Raycast does not have all environment variables in its own process, `child_process` will only get some of the user's environment variables. That leads to many errors in the integrated terminal.

That's why I added `open()` a few months ago to fix this.

If you mean that some IDEs don't work with the `open()` function, it would be better to ask the Raycast team to add support for those IDEs.

yusifaliyevpro:
https://app.slack.com/client/T0194883WF8/C0851ELJMUH

You may need to connect Raycast account to join Slack.

mturac:
Thanks for the context. I reworked this away from the CLI/child_process path and now open Windows projects through the selected editor URL handler instead, so it should avoid the integrated terminal environment regression while still avoiding the Explorer fallback.

pernielsentikaer:
Is this ready to be merged or is there any outstanding issues @yusifaliyevpro / @mturac

mturac:
Updated this on top of current main and marked it ready again.

The branch now keeps the Windows fix on the URL-handler path rather than the earlier CLI/child_process approach, so it should avoid the integrated-terminal environment issue discussed above while still preventing local projects from falling back to Explorer.

Validation:
- npm run lint
- npm run build
- git diff --check

I could not manually exercise the Windows shell handoff from this macOS environment, but the changed path is limited to the Windows branches and the selected editor scheme still comes from the existing build preference mapping.

pernielsentikaer (changes requested):
Thanks for reworking this @mturac — moving off `child_process` to the `editor URL handler` is the right call, and it nicely sidesteps the integrated-terminal env issue @yusifaliyevpro raised earlier. Using `getBuildScheme()` so the right editor (`Cursor` / `Insiders` / `VSCodium` / …) gets the URL is a good touch too.

Before we merge, a couple of edge cases I'd like to confirm, since the Windows path now always builds `<scheme>://file/<path>` and ignores `props.uri`:

`Workspaces` — for a `.code-workspace` entry, path is the workspace config file, so the `file/` handler may open it as a plain text file instead of restoring the multi-root workspace. On macOS we still pass `props.uri`, which preserves the folder/file/workspace distinction. Could you special-case workspace entries on Windows (or confirm the handler restores them correctly)?

`Remote entries` — `vscode-remote://` recents (`WSL`/`SSH`/`devcontainers`) would get flattened to a local `file/` path. Is that something you've tested, or should we note it as a known limitation?

`UNC paths` — `\\server\share\…` normalizes to `//server/share/…`, which after `split/join` looks like it produces a triple slash (…://file///server/…). Minor, but might be worth handling for network shares.  

`Folder` and `plain-file` entries look good to me.

Happy to approve once the workspace case is either handled or explicitly documented. Thanks again!

---

mturac:
Thanks for the careful edge-case pass. I pushed a follow-up on the current head to cover the Windows cases you called out:

- Workspace entries now keep the existing app-based open path on Windows, so `.code-workspace` files are not routed through the folder-style `file/` URL handler.
- Remote entries stay on the existing remote path and are still opened through the VS Code CLI helpers, so `vscode-remote://` recents are not flattened into local file URLs.
- UNC paths are normalized before encoding so `\\server\\share` becomes the editor file URI form without the extra slash issue.

Validation re-run:
- `npm run lint`
- `npm run build`
- `git diff --check`

I left this ready for review.

pernielsentikaer (approved):
Looks good to me, approved 🔥

yusifaliyevpro:
@mturac After this is merged, VSCode opens the projects in the same window and closes the previous one. I think this is not the intended behavior. I will probably need to open a PR and revert this change or find a way to fix it without removing editor URLs.

yusifaliyevpro:
> * Launches Windows recent projects through the selected editor URL handler instead of Raycast opening the folder target directly in Explorer.

This was actually an internal Raycast issue for a while, and they have solved it.

Maybe sometimes when the system is under a heavy load, Windows can open it in File Explorer, but it is rare. The `open()` function normally works well.

yusifaliyevpro:
> * Removes the Windows path through the Finder selection API in the Open With VS Code command.

getSelectedFinderItems() actually works on Windows. So I am reverting this one.

image

PR #28094, opened 2026-05-18
Add filemover extension

niborrenrew:
## Checklist


0xdhrv (commented):

We already have an extension in the Store that deals with moving and copying selected Finder files and folders. Could you clarify how this extension differs from the existing one below, which also includes a move command?

* [Finder File Actions](https://www.raycast.com/pa1ar/finder-file-actions) by @pa1ar

Could we consider enhancing the existing extension instead of creating another one?

niborrenrew:
Hi @0xdhrv ,

Thank you for taking the time to review my extension and for the suggestion! 

I had a look at the "Finder File Actions" extension. While it's a great all-around tool, I strongly believe that my extension serves a different use case and provides a significantly better, highly specialized user experience that justifies a standalone extension. 

Here is why I believe it shouldn't be merged into the existing one:

**1. Laser-Focused UX vs. General Purpose:** 
"Finder File Actions" is a "Swiss Army knife" for various file operations. My extension is purpose-built *exclusively* for moving files. Because of this strict focus, the UI is much faster and more streamlined for this specific task.

**2. Key Features (Favorites & Recents):** 
The core value proposition of my extension is the built-in "Favorites" and "Recents" functionality. This makes moving files significantly more convenient, as users can instantly send files to their most used directories without navigating the file system every time. Integrating this specialized workflow into a general-purpose extension would likely clutter it.

**3. Discoverability & Clarity:** 
For a user who just wants a fast way to move files, a dedicated "File Mover" is much easier to discover. The description of the "Finder File Actions" extension isn't very explicit about the file moving capabilities, making it harder for users to find what they need.

**4. Technical Architecture & Edge Cases:**
Under the hood, my implementation handles specific edge cases that are crucial for a smooth file-moving experience, which are hard to maintain in a broad extension:
- **Background Selection State:** My extension polls `getSelectedFinderItems`, meaning users don't need the Finder to be the active, frontmost window to trigger the move command. This removes a lot of friction compared to extensions that throw an error if Finder isn't actively in focus.
- **Intelligent Duplicate Handling:** The `safeMoveOrCopy` function automatically resolves file name conflicts by appending a counter (e.g., `file (1).txt`), preventing crashes or failed operations without interrupting the user.
- **Filtered Search:** The `mdfind` implementation actively strips out system folders, `node_modules`, and `.git` directories, ensuring users only see relevant destination folders and don't accidentally break local projects.
- **Performance:** I extracted components like the `FolderActionPanel` to avoid re-mounting on every render, keeping the UI snappy even during live file system searches.

I believe the Raycast store benefits from having both: a broad tool for general file actions, and a highly optimized, specialized tool for power users who frequently move files between specific folders.

Given these distinct workflow advantages and the specialized codebase, would you be open to merging this as a standalone extension?

pa1ar:
hey @niborrenrew, do you have a screen recording of your extension available somewhere? i’d be happy to see the UI improvements you’re proposing.

based on your description, functionality-wise it seems very similar to what Finder File Actions already does. i didn’t quite understand the part about Finder not needing to be the active window. what is being moved in that case if the selected file isn’t clearly visible? maybe i’m misunderstanding the workflow.

also, polling selection state doesn’t sound ideal from a performance/reliability perspective.

anyway, happy to hear concrete feedback on FFA if you have specific feature or UX improvement ideas.

niborrenrew:
Hey @pa1ar ! Thanks for getting back to me. I've attached a demo GIF so you can see the UI improvements and the workflow in action.

To clarify the "Finder not needing to be active" part: As far as I know, Finder File Actions requires you to actively be in a Finder window. Filemover works system-wide. I can select multiple files on my Desktop, switch to a completely different app like a browser, and simply hit my Raycast shortcut Cmd + M. It smoothly grabs all selected files at once without me having to bring Finder back to the front.

Filemover is also heavily optimized for repetitive sorting. It immediately presents a unified list of "Favorites" and "Recents", and uses direct keyboard shortcuts (like Hyperkey + D to copy or Hyperkey + N for a new folder) right from the main view, saving you the time of navigating through sub-menus.

Regarding the polling state: The setInterval only runs while the Raycast extension window is actively open and visible (it's inside a React useEffect). It doesn't run in the system background at all. It just exists to dynamically update the file preview if the selection changes while the extension is open, so it hasn't caused any performance issues.

Hope the GIF helps to clarify the workflow!

CleanShot 2026-06-08 at 20 05 26

pa1ar:
@niborrenrew from the screen recording, the user flow looks exactly like in FFA. maybe i am not seeing something, but as far as i can tell, you can just assign a shortcut to "move to folder" to CMD+M, and the experience with FFA will be 1-to-1 the same as you suggest here with the filemover.

besides the somewhat questionable part with polling and moving file without finder window being active window, which i don't really see the point of, but maybe i am not familiar with some kind of use case that you have in mind.

so in short, only the conflict resolving can be probably improved in FFA. yet it is somewhat opinionated too. 

CleanShot 2026-06-09 at 12 03 48

niborrenrew:
Hi @0xdhrv , @pa1ar ! Thank you so much for taking the time to review the initial version and for your honest feedback. 

I thought a lot about your recommendation and the comparison to *Finder File Actions* (FFA). You were absolutely right that the initial version was too similar. I took your feedback to heart, went back to the drawing board, and completely overhauled the extension. 

FFA is a fantastic tool for what it does! However, I realized that my own daily workflow required a highly specialized "power user" tool for massive, high-volume file routing. Since these complex features wouldn't fit well into FFA's streamlined UI, I decided to evolve **File Mover** into a standalone tool with a different focus. 

Here is how File Mover has evolved into a unique toolkit:

### 1. Advanced Batch Rename Engine (`Cmd + R`)
File Mover now integrates a powerful batch-rename form directly into the move action. Users can select messy files and instantly apply:
- Complete name replacements
- Prefixes or Suffixes
- Find & Replace functionality (e.g., swap `IMG_` for `Holiday_`)
- Dynamic padded indexing (`_01`, `_02`)
- Optional Date stamping in 4 different formats (ISO, EU, US, UNIX)

### 2. Built-in Safety: Action History & System-Wide Undo
File routing can be error-prone when handling many files. File Mover now tracks the last 20 operations (moves, copies, renames). It introduces a dedicated `File Move History` command to review past actions, and a standalone `Undo Last File Move` command that users can bind globally (e.g., `Hyperkey + Z`) to instantly revert accidental file operations system-wide with a single keystroke.

### 3. Bulletproof Search Architecture
Sometimes the local Spotlight index (`mdfind`) can hang or return empty due to macOS TCC permission issues. **File Mover** now implements a robust hybrid search: it uses `mdfind` for speed, but if Spotlight is unavailable, it instantly falls back to an internal Node-based traversal of the user's critical directories (`Desktop`, `Downloads`, `Documents`, `Development`). The search *always* works.

### 4. "Zero-Drill" UI Optimized for Speed
File Mover is designed to avoid deep menu navigation. Upon launch, users immediately see a unified list of their Custom Favorites and their 4 Most Recently Used folders. Actions are performed instantly via top-level keyboard shortcuts (`Enter` to move, `Cmd + D` to copy, `Cmd + N` to create a new folder on the fly). 

Thank you again for pushing me to improve this! I really believe File Mover serves a different, more demanding target audience now. Let me know what you think about the new feature set!

0xdhrv:
> Hi @0xdhrv , @pa1ar ! Thank you so much for taking the time to review the initial version and for your honest feedback.
> 
> I thought a lot about your recommendation and the comparison to _Finder File Actions_ (FFA). You were absolutely right that the initial version was too similar. I took your feedback to heart, went back to the drawing board, and completely overhauled the extension.
> 
> FFA is a fantastic tool for what it does! However, I realized that my own daily workflow required a highly specialized "power user" tool for massive, high-volume file routing. Since these complex features wouldn't fit well into FFA's streamlined UI, I decided to evolve **File Mover** into a standalone tool with a different focus.
> 
> Here is how File Mover has evolved into a unique toolkit:
> 
> ### 1. Advanced Batch Rename Engine (`Cmd + R`)
> File Mover now integrates a powerful batch-rename form directly into the move action. Users can select messy files and instantly apply:
> 
> * Complete name replacements
> * Prefixes or Suffixes
> * Find & Replace functionality (e.g., swap `IMG_` for `Holiday_`)
> * Dynamic padded indexing (`_01`, `_02`)
> * Optional Date stamping in 4 different formats (ISO, EU, US, UNIX)
> 
> ### 2. Built-in Safety: Action History & System-Wide Undo
> File routing can be error-prone when handling many files. File Mover now tracks the last 20 operations (moves, copies, renames). It introduces a dedicated `File Move History` command to review past actions, and a standalone `Undo Last File Move` command that users can bind globally (e.g., `Hyperkey + Z`) to instantly revert accidental file operations system-wide with a single keystroke.
> 
> ### 3. Bulletproof Search Architecture
> Sometimes the local Spotlight index (`mdfind`) can hang or return empty due to macOS TCC permission issues. **File Mover** now implements a robust hybrid search: it uses `mdfind` for speed, but if Spotlight is unavailable, it instantly falls back to an internal Node-based traversal of the user's critical directories (`Desktop`, `Downloads`, `Documents`, `Development`). The search _always_ works.
> 
> ### 4. "Zero-Drill" UI Optimized for Speed
> File Mover is designed to avoid deep menu navigation. Upon launch, users immediately see a unified list of their Custom Favorites and their 4 Most Recently Used folders. Actions are performed instantly via top-level keyboard shortcuts (`Enter` to move, `Cmd + D` to copy, `Cmd + N` to create a new folder on the fly).
> 
> Thank you again for pushing me to improve this! I really believe File Mover serves a different, more demanding target audience now. Let me know what you think about the new feature set!

Thanks for the detailed follow-up and for iterating on the extension.

I still don’t think this justifies a standalone extension in its current form. The updated positioning mostly confirms that File Mover combines capabilities that already exist across Finder File Actions and Renaming, rather than introducing a sufficiently distinct extension.

The core routing flow - move/copy selected Finder items to folders, use favorites/pinned folders, recent folders, folder search, and create/wrap folders - belongs very naturally in Finder File Actions and is already largely covered there. The rename functionality is also already covered by the Renaming extension, including advanced batch rename operations. Combining those flows into one command is a UX preference, but I don’t think that alone is enough reason to add another extension with substantial overlap.

The main genuinely new capability seems to be action history + undo. That is useful, but it feels like an improvement to Finder File Actions rather than the basis for a separate extension. If the goal is safer high-volume file routing, I’d rather see undo/history, better conflict handling, or rename-before-move added to the existing extension than introduce another file mover that duplicates a lot of the same surface area.

Windows support is noted, but I also don’t think it changes the duplication concern for the macOS store experience. If Windows support is the differentiator, the extension should probably be scoped and positioned around that specifically, or the existing extension should be extended where possible.

So my recommendation would still be not to accept this as a new standalone extension. The useful parts should be proposed as contributions to the existing extensions instead, especially Finder File Actions for move/copy/history/undo and Renaming for the rename-specific functionality.

niborrenrew:
Hi @0xdhrv, thank you for the continued feedback and for clarifying the store positioning perspective.

I completely understand the concern regarding surface area overlap on macOS. However, I’d love to advocate for this extension from an **architectural** and **cross-platform usability** standpoint, which makes proposing these changes to *Finder File Actions* (FFA) technically impractical.

### 1. The Technical Barrier to Contributing to FFA
FFA is fundamentally built around macOS paradigms and Finder APIs. My extension is built entirely on Node.js filesystem APIs (`fs`) specifically to achieve cross-platform parity (macOS + Windows). 

To add Windows support, Node-traversal fallback search, and system-wide undo/history to FFA, I wouldn’t just be contributing a feature—I would have to completely rewrite FFA’s core routing architecture and replace its underlying APIs. I don’t think it is appropriate (or fair to the original author) to propose an overhaul of their entire codebase via a PR.

### 2. The Cross-Platform Usability Vision
With Raycast expanding to Windows, many power users work in mixed-OS environments. Requiring users to use *Finder File Actions* on macOS and a completely different extension on Windows breaks muscle memory and creates a fragmented UX. 

Having a single, unified codebase that offers the exact same speed-optimized UI, shortcuts, history, and safety net across both macOS and Windows is a massive usability win.

### 3. Proposed Solution: Repositioning & Rebranding
I want to fully respect the store guidelines and avoid any user confusion on macOS. Would you be open to accepting this as a standalone extension if I **rebrand and scope it explicitly around cross-platform workflows**? 

For example, I could rename it to **"Universal File Router"** (or **"Cross-Platform File Actions"**) and update the store description/subtitle to clearly position it as a cross-platform toolkit for users working across OS boundaries or needing pure Node-based execution independent of macOS Finder/Spotlight APIs.

Let me know if this repositioning works for the team. I'd be more than happy to adjust the metadata and branding right away!

0xdhrv:
> Hi @0xdhrv, thank you for the continued feedback and for clarifying the store positioning perspective.
> 
> I completely understand the concern regarding surface area overlap on macOS. However, I’d love to advocate for this extension from an **architectural** and **cross-platform usability** standpoint, which makes proposing these changes to _Finder File Actions_ (FFA) technically impractical.
> 
> ### 1. The Technical Barrier to Contributing to FFA
> FFA is fundamentally built around macOS paradigms and Finder APIs. My extension is built entirely on Node.js filesystem APIs (`fs`) specifically to achieve cross-platform parity (macOS + Windows).
> 
> To add Windows support, Node-traversal fallback search, and system-wide undo/history to FFA, I wouldn’t just be contributing a feature—I would have to completely rewrite FFA’s core routing architecture and replace its underlying APIs. I don’t think it is appropriate (or fair to the original author) to propose an overhaul of their entire codebase via a PR.
> 
> ### 2. The Cross-Platform Usability Vision
> With Raycast expanding to Windows, many power users work in mixed-OS environments. Requiring users to use _Finder File Actions_ on macOS and a completely different extension on Windows breaks muscle memory and creates a fragmented UX.
> 
> Having a single, unified codebase that offers the exact same speed-optimized UI, shortcuts, history, and safety net across both macOS and Windows is a massive usability win.
> 
> ### 3. Proposed Solution: Repositioning & Rebranding
> I want to fully respect the store guidelines and avoid any user confusion on macOS. Would you be open to accepting this as a standalone extension if I **rebrand and scope it explicitly around cross-platform workflows**?
> 
> For example, I could rename it to **"Universal File Router"** (or **"Cross-Platform File Actions"**) and update the store description/subtitle to clearly position it as a cross-platform toolkit for users working across OS boundaries or needing pure Node-based execution independent of macOS Finder/Spotlight APIs.
> 
> Let me know if this repositioning works for the team. I'd be more than happy to adjust the metadata and branding right away!

Hey @pa1ar 👋 

When you have a moment, can you take a look at this?

pa1ar:
@niborrenrew

- Cross-platform support may be valuable, but I haven’t seen demand for Windows support and can’t test it properly, so I won’t implement or maintain it. Windows would mainly require adapters for file selection and folder search, not a rewrite of FFA’s core. Feel free to propose a PR.
- FFA already uses Node.js for file operations.
- History could be useful and was on my roadmap, but I never implemented it because demand remains unclear.

I’m open to safe undo, but on macOS it should integrate with Finder’s native undo instead of maintaining separate extension history. I’ve implemented this locally and am testing it. I’ll update FFA once it works reliably.

PR #28097, opened 2026-05-18
Reclaim: Fix error boundary to use proper Raycast List component

SandyGifford:
## Summary

Fixes the error boundary fallback component to use Raycast's proper `List` and `List.EmptyView` components instead of a plain React fragment. This ensures consistent UI rendering and follows Raycast's component patterns when errors occur.

**Changes:**
- Updated `RAIErrorBoundary.tsx` fallback from fragment to `List.EmptyView`
- Maintains the same error message: "Something went wrong! Please contact support."

## Testing

Tested locally by triggering error boundary conditions in the extension. The error state now renders properly with Raycast's standard empty view UI instead of rendering raw text.

### DIFF
--- a/extensions/reclaim-ai/CHANGELOG.md
+++ b/extensions/reclaim-ai/CHANGELOG.md
@@ -8,7 +8,7 @@
 ## [Update] - 2025-06-04
 - Update to the package.json description to include Outlook as a valid calendar provider.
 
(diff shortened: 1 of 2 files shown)


0xdhrv (approved):
Looks good to me, approved ✅

PR #28116, opened 2026-05-19
[Day One] Resolve CLI from common macOS paths

### LINKED ISSUE #28113: [Day One] Says CLI missing when it is installed and working
### Extension

https://www.raycast.com/AntonNiklasson/day-one

### Raycast Version

1.10.4.7

### OS Version

26.5

### Description

Says I "Day One CLI Missing" but it is installed and working (Screenshot attached):

Image

### Steps To Reproduce

CLI Installed
Able to add an entry via the CLI in the terminal
Plugin Installed
Try to add an entry through Raycast
Get error "Day One CLI Missing" 

Have uninstalled, reinstalled. Day One is password protected, have removed that and tried as well. No change.

### Current Behaviour

Does not work

### Expected Behaviour

Should Work

mturac:
## Description

Fixes #28113.

Raycast does not always inherit the same PATH as an interactive terminal, so the Day One extension can report the `dayone2` CLI as missing even when it is installed and works in a shell. This keeps the existing command flow but adds common macOS binary locations when running Day One CLI commands.

## Screencast

N/A - small CLI lookup fix.

## Checklist

- [x] I ran `npm run lint`
- [x] I ran `npm run build`

Türkiye’den sevgiler

### DIFF
--- a/extensions/day-one/CHANGELOG.md
+++ b/extensions/day-one/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Day One Changelog
 
+## [Bug fix] - 2026-05-19
+
+- Fixed CLI detection when Raycast does not inherit the terminal PATH.
+
 ## [Bug fix] - 2025-03-17
 
 - Fixed an issue where adding an entry when the journal name contains a space character causes an error.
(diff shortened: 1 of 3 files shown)


pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #28150, opened 2026-05-20
Keep Bitwarden session active after sync failures

### LINKED ISSUE #28126: [Bitwarden] Syncing vault cancels authentication
### Extension

### Raycast Version

1.104.7

### OS Version

macOS Tahoe 26.5

### Description

The images below describes the steps before the bug happens, but it's not possible to show the bug itself.

As described in the "Steps To Reproduce" section, I open the Vault, type my password, submit and the sync starts.
After some seconds of this background task, the session is stopped and I'm taken back to authentication.

Obs.: I tried to rotate my credentials (and updated in the extensions config) and it got worse: now it just don't recognize my credentials anymore, as if it's invalid.

Image
Image
Image

### Steps To Reproduce

1. Install the extension
2. Set proper Client ID and Client Secret on the extension's settings
3. Open Raycast
4. Type "Open Vault"
5. Type your password and submit
6. Authentication works, starts to sync
7. After some seconds, the vault is locked again and my session is invalid

### Current Behaviour

I authenticate, then, after a few seconds, I'm not authenticated anymore and can't handle my vault items. It happens while it tries to sync my vault in Raycast with the server.

### Expected Behaviour

Usually, after the authentication, the vault should be timing out only after 15 minutes and I don't have to re-authenticate. The sync should work well too.

mturac:
## Description

- avoid clearing the active Bitwarden session when `bw sync` reports an invalid session token
- surface the sync failure while still attempting to load the local vault data
- keep the existing logout behavior for later vault operations that prove the session is unusable

Fixes #28126

## Screencast

N/A

## Checklist

- [x] I ran `npm run lint`
- [x] I ran `npm run build`


marln94:
I couldn't try this out on my own machine, but I'm hoping this fix will finally put an end to this annoying bug! 🙏

pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #28153, opened 2026-05-20
Ignore Raycast windows when tiling layouts

### LINKED ISSUE #28032: [Window Layouts] ... not responding at all in Raycast V2 beta
### Extension

https://www.raycast.com/teemu_suvinen/window-layouts

### Raycast Version

v0.60.1.0

### OS Version

MacOS Tahoe 26.5 (25F71)

### Description

The extension doesn't work. When I select any kind of layout, nothing happens. I'm running V2 Raycast, so it probably has to do with the beta version.

### Steps To Reproduce

1. Installed the extension in Raycast V2 Beta.
2. Tried running any of the layouts.
3. A popup appears very briefly (more like it's flashing) indicating it's trying to tile the windows, but in reality nothing happens.

### Current Behaviour

_No response_

### Expected Behaviour

It should tile the windows as requested (for instance 50/50, grid of 4, horizontal 3 columns, etc.), but nothing works.

### LINKED ISSUE #28130: [Window Layouts]Extension not working in Raycast v2 beta
### Extension

https://www.raycast.com/teemu_suvinen/window-layouts

### Raycast Version

Raycast v2 public beta

### OS Version

macOS 26.5 Tahoe, Apple Silicon

### Description

I’m using the new Raycast v2 public beta.

This extension works fine in Raycast v1, but in v2 it no longer works as expected.

According to the Raycast website, v2 is a separate beta app that runs alongside v1 and some features are still being built, but extensions from the Store are expected to keep working.

On v2, when I try to run this extension, it either does nothing or fails with an error (see steps and current behaviour below).

The same version of the extension, with the same configuration, works normally in Raycast v1 on the same machine.

### Steps To Reproduce

1. Open Raycast v2 public beta on macOS Tahoe.

2. Make sure the extension is installed and configured the same way as in Raycast v1.

3. Trigger the extension from the root search (or via its hotkey / command palette).

4. Observe that in v2 the extension does not run correctly, while in Raycast v1 it still works as expected.

### Current Behaviour

_No response_

### Expected Behaviour

_No response_

mturac:
## Description

- exclude Raycast and Raycast Beta windows from the tiling window set
- avoid sending layout bounds to Raycast's own extension/webview windows, which can reject window-management updates on v2 beta

Fixes #28032
Fixes #28130

## Screencast

N/A

## Checklist

- [x] I ran `npm run lint`
- [x] I ran `npm run build`

### DIFF
--- a/extensions/window-layouts/CHANGELOG.md
+++ b/extensions/window-layouts/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Window Layouts Changelog
 
+## [Fix Raycast Beta windows] - {PR_MERGE_DATE}
+
+- Fixed layout commands trying to tile Raycast's own windows on Raycast v2 beta.
+
 ## [Improvements] - 2026-05-06
 
 - Improved the error message shown when the Raycast `WindowManagement` API is not available, clarifying that the extension requires both Raycast Pro and Accessibility permission
--- a/extensions/window-layouts/CHANGELOG.md
+++ b/extensions/window-layouts/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Window Layouts Changelog
 
-## [Fix Raycast Beta windows] - {PR_MERGE_DATE}
+## [Fix Raycast Beta windows] - 2026-05-20
 
 - Fixed layout commands trying to tile Raycast's own windows on Raycast v2 beta.
 
--- a/extensions/window-layouts/src/utils/get-desktop-context.ts
+++ b/extensions/window-layouts/src/utils/get-desktop-context.ts
@@ -7,6 +7,11 @@ export type DesktopContext = Readonly<{
   windows: WindowManagement.Window[];
 }>;
 
+function isRaycastWindow(window: WindowManagement.Window) {
+  const appName = window.application?.name?.toLowerCase();
+  return appName === "raycast" || appName === "raycast beta";
+}
+
 export async function getDesktopContext(): Promise<DesktopContext | null> {
   if (!environment.canAccess(WindowManagement)) {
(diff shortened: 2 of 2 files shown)


pernielsentikaer (approved):
Looks good to me, approved 🔥

agrarico:
Hi team,

Unfortunately, the fix doesn't work on my Mac. I still get the same error. When I try to run a command in the extension, the screen flashes, but nothing happens. The extension doesn't respond as it should. I was the original poster of this error. See error: #28032.

I think is this is the error:

#### Error:
```
No resizable windows found
```

What do you need from me to continue investigating the issue?
I'm running a MacBook Pro M3 with the latest version of macOS Lake Tahoe. I've tried the extension on both an external screen and my laptop screen, but it makes no difference. I've already uninstalled the extension, closed Raycast, restarted everything, and rebooted my laptop. Nothing seems to work.

If you'd like to take another look or have any tips, I'd be happy to hear them.

teemusuvinen:
Hi @agrarico ! I'm the original creator of the extension and I can verify your findings. 

This seems to be an unfortunate bug in the Raycast extension api. We are using `WindowManagement.getWindowsOnActiveDesktop()` function and it seems to return an empty array even on the latest version of the API package on Raycast Beta. There are also some additional errors visible when running the extension in dev mode. I will try to contact the Raycast team to see if this issue could be resolved

teemusuvinen:
@agrarico the issue has been fixed in the newest version of Raycast Beta (v0.62.2.0)

Screenshot 2026-06-03 at 21 08 32@2x

It's working at least for me now

PR #28182, opened 2026-05-20
Add flashcards extension

Paulito12209:
## Checklist


0xdhrv (commented):

We already have a couple of extensions in the Store that deal with flashcards / study workflows:

* [LeitnerBox](https://www.raycast.com/gzyannick/leitnerbox) by @GZyannick
* [Anki](https://www.raycast.com/anton-suprun/anki) by @anton-suprun

This extension looks similar in domain, but it also seems to introduce distinct workflows, especially around multiple-choice cards, tags, and Markdown import/export.

Could you clarify how you see this extension differing from the existing ones, particularly LeitnerBox? It would be helpful to understand whether the goal is a simpler general-purpose flashcard system, an alternative to Leitner-style spaced repetition, or something else.

If there are unique features or workflows you’re aiming to add, we’d love to capture that clearly so users can understand when to pick this extension over the existing options.

> This would help avoid duplication and keep related functionality easy to discover.

Paulito12209:
Thanks again for the review and for pointing me to the existing flashcard
extensions. I'd like to clarify what Flashcast is actually optimized for,
because its main purpose is different from both LeitnerBox and Anki.
Flashcast is built to make creating flashcards fast, not just reviewing
them.
The core idea is a simple, human- and machine-readable Markdown syntax for
cards. Because the format is plain Markdown, you don't have to enter cards
one by one through a form. You can generate a whole set at once — for
example, ask any LLM or a Raycast AI command to "create 50 vocabulary cards
for different fruits, French to German" or "turn these notes into a
multiple-choice quiz," get the output directly in Flashcast's syntax, and
import the entire batch in one step.
That's the key difference from the existing extensions:
Anki (anton-suprun) is a frontend for the Anki desktop app: it requires
Anki running plus the anki-connect add-on, and cards are added individually
through the Add form.
LeitnerBox (GZyannick) is a Leitner spaced-repetition tool where each
question is created one at a time via a "Create a Question" form.
Neither is designed around bulk, syntax-based authoring. Flashcast is:
define or paste the Markdown once, import many cards instantly, no external
app or setup (everything lives in Raycast LocalStorage). On top of that,
review is fast too — a fully keyboard-driven quiz flow with focused modes
(wrong cards, new cards, by tag, all cards) and auto-graded multiple-choice
cards.
One thing I want to make explicit in the README: I'll document the card
syntax directly, including how to use it with AI. The goal is that a user
can hand this format to any AI assistant and get import-ready cards back.
Concretely:
Standard card:
Question or term == Answer #tag1 #tag2
Multiple-choice card (up to 3 options):
Question ==< 1: Option A 2: Option B 3: Option C -- correct: 2 #tag1
(
correct:
uses
true:
in English mode /
richtig:
in German mode.)
With that documented, typical AI prompts become trivial, e.g.:
"Create 50 cards, French → German, for common fruits, using this Flashcast
syntax."
"Turn this text into 10 multiple-choice cards (3 options each) in Flashcast
syntax, tag them #biology."
"Generate translation cards for these 30 vocabulary words in Flashcast
syntax."
So Flashcast isn't meant to replace Leitner-style spaced repetition — it's
a lighter, general-purpose flashcard tool whose unique value is effortless
bulk card creation via Markdown import (including AI-generated decks) and
very fast, keyboard-first review. I'll add both the positioning and the
syntax reference to the README so users know exactly when and how to use it.


*Kind regards*

*Paul Angeles Chaquire*
*                ---*





Am Mo., 8. Juni 2026 um 07:52 Uhr schrieb Dhruv Suthar <
***@***.***>:

0xdhrv (approved):
Looks good to me, approved ✅

PR #28189, opened 2026-05-20
Preserve Brew search cache during rebuilds

mturac:
## Description

Fixes Brew search cache rebuilds so the previous working chunked index stays available until the replacement index has been built successfully. If rebuilding fails, the extension can keep using the stale cache instead of leaving search with no usable chunked cache.

Closes #26023

## Screencast

N/A

## Checklist

- [x] I ran `npm run lint`
- [x] I ran `npm run build`

### DIFF
--- a/extensions/brew/CHANGELOG.md
+++ b/extensions/brew/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Brew Changelog
 
+## [Fix Search Cache Recovery] - {PR_MERGE_DATE}
+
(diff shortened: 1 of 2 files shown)


nhojb:
@mturac thanks for the PR. I actually have my own PR which addresses the same issue, plus adds a few other improvements to the fetch/cache code. So if you don't mind I will close this one.

mturac:
Sounds good, thanks for taking care of it.

PR #28209, opened 2026-05-20
Fix/table records database views

robertoalvarezalonso:
## Description

Follow-up fixes for the ServiceNow extension after the OAuth/Windows release.

### Fixes
- **Explore Records** now works on database views (`v_*`) and other tables without a `sys_updated_on` column. Before, the `ORDERBY sys_updated_on` failed the whole request (HTTP 500), and an unparseable update date could crash the list. The order-by is now skipped upfront for views and falls back on error for the rare non-view table that also lacks the column.
- **Find Record by Sys ID** and **Find Record References** now distinguish a sign-in failure from a missing admin role, and offer an **Edit Profile** action to fix the profile and retry in place. When an instance uses SSO, the message points you to switch the profile to OAuth, since Basic Auth can't run the background scripts these commands rely on.

### Internals
- Switched **Explore Records** to `sysparm_display_value=all` (shows display values, keeps the raw UTC `sys_updated_on` for the relative-date labels).
- Updater avatars show display name + login in the tooltip and derive initials from real name words, ignoring decorators like "(Admin)".
- Removed unused response types and the unused `serviceNowFetch` / `ServiceNowApiError` helpers.

## Checklist

### DIFF
--- a/extensions/servicenow/CHANGELOG.md
+++ b/extensions/servicenow/CHANGELOG.md
@@ -1,5 +1,11 @@
 # ServiceNow Extension Changelog
 
+## [Database View Support & Clearer Errors] - {PR_MERGE_DATE}
+
+- **Explore Records** now works on database views and other tables that don't track an update time, instead of failing to load.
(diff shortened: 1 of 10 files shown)


0xdhrv (approved):
Looks good to me, approved ✅

PR #28241, opened 2026-05-22
Update scheduler extension

csymons:
## Description

- Refactors the Scheduler configuration UI to make command setup clearer and easier to scan.
- Fixes Raycast 2.0 draft restoration when creating scheduled commands by opening the create form as a top-level command.
- Refreshes Scheduler dependencies and lockfile.

## Checklist

### DIFF
--- a/extensions/scheduler/CHANGELOG.md
+++ b/extensions/scheduler/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Scheduler Changelog
 
-## [Fix Raycast 2.0 Drafts] - {PR_MERGE_DATE}
+## [Fix Raycast 2.0 Drafts] - 2026-05-22
 
 - Fixed schedule form drafts in Raycast 2.0 by opening the create form as a top-level command so draft values are restored correctly.
 
--- a/extensions/scheduler/CHANGELOG.md
+++ b/extensions/scheduler/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Scheduler Changelog
 
+## [Fix Raycast 2.0 Drafts] - {PR_MERGE_DATE}
+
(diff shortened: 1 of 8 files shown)


0xdhrv:
Hey @csymons 👋

Thanks for your contribution.

Can you share more details about this?

> Fixes Raycast 2.0 draft restoration when creating scheduled commands by opening the create form as a top-level command.

This isn't a blocking comment. I'm just trying to understand the issue you mentioned.

0xdhrv (approved):
Looks good to me, approved ✅

csymons:
> Can you share more details about this?
> 
> > Fixes Raycast 2.0 draft restoration when creating scheduled commands by opening the create form as a top-level command.
> 
@0xdhrv  Sure. 

It seems that maybe Raycast 2.0 changed or tightened the behavior around Form drafts. The extension was showing the create schedule form by pushing it from the main Scheduled Commands list command using Action.Push. That meant Raycast saved the draft under the list command, not under the standalone create-new-scheduled command.

So when selecting the draft from Raycast’s root search, Raycast reopened Scheduled Commands, which only renders the saved schedules list. The unsaved half-finished schedule was never passed back into the form, so it looked like the draft was lost.

The fix was to make schedule creation happen through the top-level create-new-scheduled command instead of a nested pushed form. The list action now calls:

launchCommand({ name: "create-new-scheduled", type: LaunchType.UserInitiated })
Then enableDrafts is only enabled on that standalone create command, where LaunchProps.draftValues are available and passed into the form as initial values. Nested edit/create forms no longer enable drafts, avoiding drafts being associated with the wrong command context.

In short: drafts need to belong to the command that owns and hydrates the form. I moved “create new schedule” to launch the dedicated command, so Raycast can restore the saved draftValues correctly.

PR #28249, opened 2026-05-22
Add copy flight number action

effektsvk:
## Description

Adds a Copy Flight Number action to the Flighty extension action panel, with a Command + . shortcut.

## Screencast

Not applicable; action-only change.


0xdhrv (approved):
Looks good to me, approved ✅

PR #28261, opened 2026-05-22
Update leafcast extension

syropian:
## Description

This is an update to the existing **Leafcast** extension. The changes break down into three groups: new ways to control the panels, smarter device discovery during pairing, and a number of fixes and cleanups.

### Added

- **AI tools** for natural-language control via Raycast AI — `set-power`, `set-brightness`, `set-color`, `set-effect`, `list-effects`, `get-state`, and `identify`. Example: *"@leafcast set them to a warm orange at 30%"*.
- **Identify Device** command (`PUT /identify`) — briefly flashes the panels so the user can confirm which device is paired. Available as a no-view command, as an AI tool, and as an action on the pairing success screen.
- Five hotkey-bindable no-view commands: **Toggle Power**, **Next Effect**, **Previous Effect**, **Brightness Up** (+10%), **Brightness Down** (-10%).
- Multi-device discovery — when more than one Nanoleaf is on the network, a picker is shown so the user can choose which device to pair with.

### Changed

- Network discovery now uses **mDNS / Bonjour** (`_nanoleafapi._tcp`) instead of parsing the ARP table for known MAC prefixes. It works regardless of recent network activity and surfaces multiple devices in one pass.
- The **Pair Device** command auto-scans on open — no need to trigger the scan manually.
- Pairing instructions are now inline on the pair screen instead of a separate `confirmAlert` modal.
- Pairing failures surface a specific cause (connection timeout, wrong IP, device not in pairing mode) instead of a generic error toast.

### Fixed

- IPv4 validation no longer accepts malformed addresses or behaves inconsistently across repeated submissions (the previous regex had unescaped dots and a `/g` flag causing `lastIndex` retention).
- Auto-filled addresses from the network scan now clear stale validation errors immediately.
- `pairDevice` no longer silently succeeds when the device responds without an `auth_token` — it throws explicitly.

### Internal

Light refactor for maintainability: extracted `src/lib/nanoleaf-client.ts` so the React hook, the no-view commands, and the AI tools all share a single source of truth for API calls. Migrated to flat ESLint config (required by `@raycast/eslint-config@2` / ESLint 9+). No public API changes.

## Checklist


0xdhrv (approved):
Looks good to me, approved ✅

PR #28270, opened 2026-05-23
Multi Links: 2.0.0 — expanded extraction, preferences, safety dialog, history & filter views

artisticmedic:
## Description

Major update to the **Multi Links** extension (v2.0.0). Builds on the unpublished v1.1.0 clipboard work into a complete "open everything in this text" tool.

### New commands
- **Open Multiple Links History** (view) — browse the last 100 batches you've opened, with type breakdown + timestamp per entry. Actions: Open All Again, Copy URLs, Pin (protects from FIFO eviction), Delete. `LocalStorage`-backed, FIFO trim at 100, per-entry items capped at 20.
- **Filter Multiple Links** (view) — reads URL-rich text from the clipboard, groups extracted links by host (web) and by type (local files, `mailto:`, custom schemes). Multi-select with `Tab`, then Open Selected — preview before opening, or open only the GitHub links from a long thread.

### Expanded extraction
The two existing no-view commands now match far more than `https?://`:
- `www.` bare hosts and bare domains with an allowlisted TLD
- `mailto:` / `tel:` / `sms:`
- `file://` URIs and absolute local paths (`/Users/...`, `~/...` with `os.homedir()` expansion)
- Markdown `[text](url)` syntax
- Arbitrary custom URI schemes (`obsidian://`, `vscode://`, `notion://`, …)
- Files with common extensions in plain text (gated by an allowlist)

### Preferences
- **Browser** picker (web URLs only; local files / custom schemes use the macOS default handler)
- **Open delay (ms)** between opens (`0` = parallel, chunked at 10)
- **Open all URI types** toggle
- **Confirm on many links** toggle + **threshold** (default 10)

### Safety + robustness
- Confirmation dialog with count, per-type breakdown, and first-5 preview when a batch hits the threshold
- Parallel opens via `Promise.allSettled` chunked at 10 (`delay=0`); sequential with `setTimeout` (`delay>0`)
- Friendly error messages — no more raw error objects in the HUD; partial-failure Toast says "Opened X of Y"
- Every command exposes an "Open Extension Preferences" action

### Housekeeping
- Bumped `@raycast/api` to `^1.104.0`, `@raycast/utils` to latest, `@types/react` to `^19`
- Added `platforms: ["macOS"]`
- Added Vitest with 104 tests (URL extraction + open pipeline behavior)
- TypeScript strict mode; `ray build` and `ray lint` clean

## Screencast

Screenshots in `metadata/` (root search with all four commands, History view, Filter view). Smoke-tested via `ray develop`: a clipboard containing a `http://localhost` URL plus a `file:///…/index.html` path now opens both (v1.1.0 silently dropped the `file://`).

## Checklist

- [x] I ran \`npm run build\` and [tested this distribution build in Raycast](https://developers.raycast.com/basics/prepare-an-extension-for-store#metadata-and-configuration)
- [x] I checked that assets used by the \`README\` are placed outside of the \`metadata\` folder


artisticmedic:
🙌  such a cool extension that I've been using for a while, glad to finally add something back to it and replace the other scripts I've been hobbling along with :)

artisticmedic:
Note on the recurring "Copy Urls" → "Copy URLs" suggestion: I'm intentionally keeping `"Copy Urls"`, because `"Copy URLs"` trips Raycast's own `@raycast/prefer-title-case` ESLint rule (`ray lint --fix` rewrites it back to `"Copy Urls"`), so the lowercase-acronym form is the lint-clean one for this codebase.

artisticmedic:
commenting to prevent from going stale!

artisticmedic:
keeping fresh

pernielsentikaer (changes requested):
Hi 👋

Thanks for the update, I have a few questions and pointers:

Main concern — opening local executables/installers from extracted text  

`fileExtAllowlist.ts` includes `app`, `dmg`, `pkg`, and `sh`. `openOne` just calls `open(path)`, so an extracted `/path/to/Something.app` launches it (and `.pkg`/`.dmg` open installers). Since clipboard/selected text can come from an untrusted source, "open all links in this text" silently launching an app is a real escalation. And the confirm dialog only fires at `confirmThreshold` (default `10`) — a single malicious `.app`/`.pkg` path opens with no confirmation. I'd recommend dropping executable/installer extensions from the allowlist, or always confirming for local executables regardless of threshold.  

Related — broad defaults for non-web types  
`extractCustomScheme` opens any `scheme://…` (`obsidian://`, `vscode://`, `raycast://`, etc.), and `openAnyUriType` defaults to `true`, so by default the extension opens file paths and arbitrary custom schemes from pasted text — again only gated by the ≥`10` confirm threshold. Worth considering a safer default (non-web types off by default, or always-confirm for custom-scheme / local-path). A reviewer/security will likely raise this; flagging it proactively in the PR will help.  

Correctness — bare-domain vs filename ambiguity  
Several entries in `tldAllowlist.ts` (`app`, `io`, `dev`, `me`, `tv`, `co`) collide with filename suffixes. `extractBareDomain` will turn `index.app` or `notes.me` into `https://index.app` and open it in a browser. Tests likely cover the happy path, but this will produce surprising web opens on real-world text — worth tightening or documenting.  

Lower priority  
`extractFileExt.ts` uses a variable-width negative lookbehind with `[^\s]{0,500}`. V8 supports it, but evaluated across a large clipboard it could get expensive — worth a quick check on big inputs (I'm inferring from the pattern, haven't profiled).  
`open(item.url, browser.bundleId ?? browser)` falls back to passing the Application object where `open` expects a string — fine if `bundleId` is always populated, but worth confirming.  

Naming: "Copy Urls" → "Copy URLs" (`history.tsx`). Using `Tab` as an action shortcut (`filter.tsx`) is non-standard and may conflict with Raycast's navigation.

---

artisticmedic:
Thanks for the thorough pass, @pernielsentikaer; point by point:

**Executables/installers — fixed, wider than the allowlist.** The vector is broader than `fileExtAllowlist`: `extractAbsPath`/`extractFileUri` also emit `local-path`, so `/Applications/Foo.app` or `file:///…/Setup.pkg` launches regardless of that list. I moved the guard to the `open()` boundary — any item ending in a code-running extension (`app, dmg, pkg, mpkg, xip, sh, command, bash, zsh, scpt, scptd, workflow, terminal, inetloc, fileloc, webloc`) warns before opening, independent of the threshold and the confirm toggle, even for a single item. Declining skips the executables and still opens the safe links.

**Non-web default — keeping it on.** Opening `obsidian://`, `vscode://`, and local paths is one of the core problems v2 set out to solve; web-only is one toggle away. With the executable guard above, the rest is click-equivalent. Documented in the README.

**Bare-domain vs filename — good catch.** No clean tightening that doesn't drop hugely-common single-label domains (`fly.io`, `vercel.app`, `twitch.tv`); the failure mode is a stray tab, not anything unsafe. README notes to wrap it in a path (`./index.app`) to force file interpretation.

**Regex lookbehind** — ~0.8 ms per extraction on a 50 KB clipboard (100-iteration avg). Left as-is.

**`open(url, browser.bundleId ?? browser)` — safe.** Signature is `open(target: string, application?: Application | string)`, so the `Application` fallback is valid, not a type error. `appPicker` populates `bundleId` anyway.

**"Copy Urls" vs "Copy URLs" — keeping "Urls".** `@raycast/prefer-title-case` special-cases only the singular `url → URL`; the plural `urls` title-cases to `Urls`, and `ray lint --fix` rewrites "Copy URLs" back. "Urls" is the lint-clean form here.

**Tab → ⌘T.** Switched it; not fully sold on ⌘T specifically, but it clears the Tab/navigation overlap for now.

Tests green (113), lint clean.

pernielsentikaer (commented):
Thanks for all the updates @artisticmedic, @Aayush9029 do you want to check this?

pernielsentikaer:
@wrick17 do you want to check this?

artisticmedic:
Commenting to prevent from going stale. @pernielsentikaer I noticed that you've been moving this back into "draft" from "ready for review;" is that a workflow preference on your part, or a commentary on the state? I want to maximize the chances for review/merge.

wrick17:
Thank you @artisticmedic for using and liking the extension. Thanks @pernielsentikaer for checking with me. Sorry for the delay, I had been away for some personal reasons. I’m happy with the direction of the update, and the always-confirm behavior for executables addresses the main safety concern from my perspective. No objections to proceeding with the Raycast review. I have briefly used it and didn't find any issues. Will take a deeper look in the future and open a PR if I find myself wanting to add anything more.

pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #28275, opened 2026-05-23
Add portkill extension

Blueemi:
## Description

**PortKill** lists local TCP listeners and stops the process bound to a port — for the “port already in use” moment. Search by port, app name, or PID, see endpoint details, and kill one process or every unique PID with confirmation.

Cross-platform via system binaries only: `lsof` / `kill` on macOS, `netstat` + `tasklist` / `taskkill` on Windows, `ss` (fallback `lsof`) / `kill` on Linux. Manifest targets **macOS** and **Windows**.

Single command (view mode), no API keys, no bundled binaries.

## Screencast

Store screenshots (also in `extensions/portkill/metadata/`):

| Listener list | Search filter |
| --- | --- |
| Default listener list | Search filtered to node |
| **Detail pane with Open in Browser** | **Kill confirmation alert** |
| Detail pane on port 3000 | Kill Process confirmation alert |

## Checklist


0xdhrv (commented):

We already have extensions in the Store that deal with this. Could we consider enhancing one of the existing extensions below instead of creating another one?

* https://www.raycast.com/lucaschultz/port-manager
* https://www.raycast.com/dleteliers_/ports
* https://www.raycast.com/RensHoogendam/plexus
* https://www.raycast.com/tavlean/dev-servers

From the Store perspective, the core user intent overlaps significantly with the existing extensions above. Users who want to list TCP-listening ports and kill the bound process will likely encounter multiple extensions for the same purpose. `port-manager` covers macOS with named ports, a menu bar command, and kill-by-port-argument; `ports` covers Windows with the same list-and-kill loop; `plexus` lists localhost web servers with process kill and WSL support across macOS and Windows; and `dev-servers` provides a project-grouped dashboard with kill, restart, and browser-open actions on macOS. Having another extension in this space can reduce discoverability and split related functionality across multiple places.

If there are unique features or workflows you're aiming to add (such as unified macOS/Windows/Linux support in a single extension, or the graceful SIGTERM → SIGKILL escalation), we'd love to hear them and see if they can be integrated into one of the existing extensions to avoid duplication and improve discoverability.

> As mentioned in our [extension guidelines here](https://manual.raycast.com/extensions#:~:text=There%20is%20an%20active%20extension,extension%20providing%20very%20similar%20value)

PR #28295, opened 2026-05-24
[Open With App] Fix default selection on Raycast 2

errepuntopenco:
## Description

Fixes two related selection bugs in the Open With App extension on Raycast 2 (the bug doesn't manifest in Raycast 1):

1. The command auto-selected the first item of "Other Applications" instead of the first item of "Recommended Applications" — so pressing Enter would open the file with an unrelated app (e.g., 1Password) instead of the user's expected default (e.g., Preview).
2. After fixing #1 with a controlled `selectedItemId`, the first arrow-down press caused a one-row scroll that hid the "Recommended Applications" section header.

### Implementation notes

- Give each `List.Item` an explicit `id` (`app.path`) so Raycast no longer auto-generates UUIDs.
- Seed `selectedItemId` to the first Recommended app on initial load and on search changes via a `useRef` that tracks the "default first id", so it only overrides when the default actually changes.
- Deliberately do **not** wire `onSelectionChange`. Re-applying `selectedItemId` on every selection change made Raycast 2 nudge the viewport by one row on the first arrow press. Leaving the prop seeded only on initial load / search lets Raycast handle arrow-key navigation internally, and the section header stays visible until the selection naturally moves past it.

## Type of change

- [x] Bug fix

## Checklist

- [x] I checked the linter (`npm run lint`) and code is formatted correctly
- [x] Verified in Raycast 2 (beta) — first Recommended app is selected, no scroll glitch on first arrow-down
- [x] Verified in Raycast 1 (stable) — existing behavior unchanged

### DIFF
--- a/extensions/open-with-app/CHANGELOG.md
+++ b/extensions/open-with-app/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Open With App Changelog
 
-## [Fix default selection on Raycast 2] - {PR_MERGE_DATE}
+## [Fix default selection on Raycast 2] - 2026-05-25
 
 - Fix: default the highlighted row to the first Recommended Application instead of the first Other Application
 - Fix: prevent a one-row scroll the first time the user presses the down arrow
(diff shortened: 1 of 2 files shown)


0xdhrv (approved):
Looks good to me, approved ✅

PR #28300, opened 2026-05-24
Add remove-newlines extension

stefankosticc:
## Description

This extension removes line breaks from text. Supports 2 commands:
- Replace Without Newlines - Replace selected text without line breaks
- Paste Without Newlines - Paste copied text without line breaks

## Checklist


0xdhrv (commented):

* https://www.raycast.com/koinzhang/paste-as-plain-text by @koinzhang

From the Store perspective, the core user intent overlaps significantly with the existing `paste-as-plain-text` extension. Users who want to strip newlines from clipboard content and paste will likely find both extensions for the same purpose. Even if the emphasis differs `remove-newlines` exposes newline stripping as a dedicated always-on command, while `paste-as-plain-text` offers it as an opt-in preference alongside other paste formats (plain text, JSON, URL, Markdown) two extensions in the same space can reduce discoverability and split related functionality.

If you plan to add unique features or workflows, such as a `replace-without-newlines` command that operates on selected text rather than the clipboard, please share them. We’d like to see if they can be integrated into the existing extension to avoid duplication and improve discoverability.

> As mentioned in our [extension guidelines here](https://manual.raycast.com/extensions#:~:text=There%20is%20an%20active%20extension,extension%20providing%20very%20similar%20value)

PR #28310, opened 2026-05-25
Add better-world-clock extension

madeyexz:
## Description

Better World Clock is a Raycast extension for quickly checking the current time in the cities and time zones you care about.

It is designed around a few common world-clock workflows:

- See multiple favorite cities at a glance
- Switch between a compact list and a visual grid
- Toggle 12-hour / 24-hour time without leaving the command
- Toggle digital / analog clock cards
- Add or remove time zones from the `⌘K` action panel
- Use friendly city names like `San Francisco`, `New York`, `London`, `Tokyo`, or advanced IANA IDs like `Europe/London`

The extension uses JavaScript's `Intl.DateTimeFormat` with IANA time zones, so UTC offsets and daylight saving time are handled by the runtime time zone database.

### List view

List view

### Digital grid view

Digital grid view

### `⌘K` actions menu

Actions menu

### Analog cards and in-command settings

Analog cards and settings

## Checklist


0xdhrv (commented):

We already have an extension in the Store that deals with world clocks and timezone lookup:

* [World Clock](https://www.raycast.com/koinzhang/world-clock) by @koinzhang

This extension appears to overlap significantly with the existing one. Both focus on checking current time across cities / time zones, support grid and list-style views, and let users manage favorite or frequently used time zones.

The main differences I can see are the digital / analog clock cards, friendlier city-name input, and some in-command display toggles. Those seem like useful improvements, but they may fit better as enhancements to the existing World Clock extension instead of creating a second extension for the same core workflow.

Could we consider contributing these UX improvements to the existing extension instead?

If there are unique workflows this extension enables beyond the existing World Clock extension, we’d love to understand them more clearly.

PR #28339, opened 2026-05-26
Update dev-servers extension

tavlean:
## Description

**3x faster, windows port preparation, support for git branches & worktrees, search by project and branch.**

- Rewrote process detection from an embedded shell pipeline to TypeScript. Output is unchanged; the new path is roughly 3x faster, removes a class of shell-parsing bugs, and lays the groundwork for a future Windows port.
- Group multiple git worktrees of the same repo into one project section. Each row shows its current branch next to the link, and per-row actions (Open in Terminal, Show in Finder) still target the specific worktree on disk.
- Show the current git branch on every project's rows, so you can tell at a glance which branch a long-running dev server is on, even for single-worktree projects.
- Search by project name or branch directly in the Raycast search bar, Raycast's built-in filter previously only matched the port number and subtitle.
- New preferences to toggle each row accessory (uptime, git branch, framework tag) independently.
- Stylized framework names (SvelteKit, Astro, Next.js, esbuild, etc.) in the row tag and the filter dropdown.
- More reliable restarts: the old process is force-killed and confirmed exited before the replacement spawns, closing a rare race where the new server could fail to bind because the old one still held the port.

## Checklist

### DIFF
--- a/extensions/dev-servers/CHANGELOG.md
+++ b/extensions/dev-servers/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Dev Servers Changelog
 
-## [Worktree grouping and detection rewrite] - {PR_MERGE_DATE}
+## [Worktree grouping and detection rewrite] - 2026-05-26
 
 3x faster, windows port preparation, support for git branches & worktrees, search by project and branch.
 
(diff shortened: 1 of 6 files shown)


0xdhrv (approved):
Looks good to me, approved ✅

PR #28344, opened 2026-05-26
Fix useState crash in Withings Sync (sync-to-garmin command)

thebruge:
## Description

Fixes a crash affecting all users of the **Sync to Garmin** command.

**Root cause:** `CustomDateRangeForm` was defined as a function inside the `SyncToGarmin` component. React requires all components to be defined at module scope — when a component is nested inside another, React sees a brand-new function reference on every render, which destroys and recreates the component's fiber, causing hooks to run on a null context:

```
TypeError: Cannot read properties of null (reading 'useState')
```

This crash occurs on every launch of the command, not just when the date range form is used.

**Fix:** Extracted `CustomDateRangeForm` to module scope with an explicit `CustomDateRangeFormProps` interface. State and handlers are passed as props.

## Checklist

- [x] Only one file changed (`src/sync-to-garmin.tsx`) — no other modifications
- [x] `npm run build` passes clean
- [x] Fixes reported issue: [extension-issues/thebruge/withings-sync/7503194464](https://www.raycast.com/extension-issues/thebruge/withings-sync/7503194464)

### DIFF
--- a/extensions/withings-sync/CHANGELOG.md
+++ b/extensions/withings-sync/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Withings Sync Changelog
 
+## [Bug Fix] - 2026-05-26
+
+- Fixed crash on launch of Sync to Garmin command (TypeError: Cannot read properties of null (reading 'useState'))
+
 ## [Initial Version] - 2026-02-05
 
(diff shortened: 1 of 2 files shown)


0xdhrv (approved):
Looks good to me, approved ✅

PR #28353, opened 2026-05-26
Update README.md

tborychowski:
## Description

Raycast 2.0 includes the same functionality, without using 3rd party uninstallers, thus making this extension obsolete.

How do I unpublish my extension?

### DIFF
--- a/extensions/appcleaner/CHANGELOG.md
+++ b/extensions/appcleaner/CHANGELOG.md
@@ -1,5 +1,10 @@
 # App Cleaner Changelog
 
+## [Updated README] - {PR_MERGE_DATE}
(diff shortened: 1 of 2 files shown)


pernielsentikaer (commented):
The best way to remove the extension from the store would be to first make all the commands into view commands with `List.EmptyView` explaining that the app is obsolete, and then after a few months, we can remove it from the store completely, this was most of the users hacing it installed will have the update and remove it themselves

pernielsentikaer (commented):
Do you want to look into that in this release?

tborychowski:
Thanks! Will look into this.

PR #28367, opened 2026-05-27
[Slack] Fix slack:// deep links on Windows

artistro08:
## Description

On Windows, the Slack extension's "Open in Slack" action and the Quicklinks created from "Open Channel" / "Open Chat" launch Slack but always land on the default view — the requested channel or user never opens.

### Root cause

`Action.Open` and `Action.CreateQuicklink` are configured with `application="Slack"` plus a `slack://...` URI as the target. On macOS this resolves to `open -a Slack <slack://...>` and the URI reaches Slack.app's registered URL handler, which navigates correctly. On Windows the same `application` hint causes Raycast to launch `Slack.exe` directly, but the URI is not forwarded to it as a command-line argument — Slack just opens to its default view.

### Fix

Gate the `application` hint and the macOS-only `fileIcon: "/Applications/Slack.app"` icon behind `process.platform === "darwin"`. On Windows the hint is omitted, so `Action.Open(target="slack://...")` is routed through the OS's registered protocol handler — which is `Slack.exe` — and the URI is forwarded intact. macOS behavior is unchanged.

Same fix applied to the `Action.CreateQuicklink` calls in `search.tsx` so saved Quicklinks also work on Windows.

## Screencast / proof

_To be added by the author after install — please verify on your Windows + macOS setups before merging._

## Checklist

- [x] I checked that files in the `assets` folder are used by the extension's code
- [x] I checked that assets used by the extension itself are reasonably sized
- [x] The title of this PR follows the [extension title format](https://developers.raycast.com/basics/prepare-an-extension-for-store#title) (`[Slack] Fix slack:// deep links on Windows`)

### DIFF
--- a/extensions/slack/CHANGELOG.md
+++ b/extensions/slack/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Slack Changelog
 
+## [Fix Windows deep links] - {PR_MERGE_DATE}
+
+- "Open in Slack" and "Open Channel"/"Open Chat" Quicklinks now open the requested channel or user on Windows. Previously the `application="Slack"` hint caused Raycast to launch Slack.exe without forwarding the `slack://` URI; the hint is now macOS-only so Windows routes the URI through the registered protocol handler.
+
 ## [Accent-insensitive search] - 2026-05-07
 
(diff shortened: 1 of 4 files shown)


0xdhrv (approved):
Looks good to me, approved ✅

PR #28374, opened 2026-05-27
Add EPUB download actions to Anna's Archive

seanwessmith:
## Description

Enhances the existing Anna's Archive extension with EPUB-focused filtering, ranking, and download actions instead of adding a duplicate extension.

- Keeps the existing `Search Archives` command and broad all-format search behavior intact
- Adds a persisted `File Type` dropdown with `All` plus Anna's Archive file formats such as PDF, EPUB, MOBI, AZW3, TXT, DOCX, ZIP, and others
- Uses `ext=<selected file type>` for non-All searches, with an extra client-side guard for matching extensions
- Adds EPUB ranking heuristics for exact titles, English fiction books, normal EPUB sizes, and likely wrong/bundle results
- Reorders detail metadata so high-signal fields come first: title, author, and size before lower-level metadata like filename and ID
- Adds EPUB-only actions for opening the official slow-download page, copying the slow-download URL, copying a clean filename, and optional member fast download
- Renames the member API preference to `annaSecretKey` and keeps the secret in Raycast preferences only
- Refactors download helpers around URL builders, the official fast-download API request, file download/save, and filename sanitization

Fast download is intentionally scoped to Anna's official member fast-download JSON API. Slow download remains a browser-open action only; this PR does not automate timers, waitlists, mirrors, or page clicks. If Raycast policy prefers avoiding automated downloads, I'm happy to remove only the fast-download action and keep the EPUB filter/ranking plus `Open Slow Download Page`.

This follows the review direction from #27734 and the existing extension author's suggestion to integrate the EPUB functionality here.

## Screenshot

Added `extensions/anna-s-archive/metadata/anna-s-archive-4.png`, showing the `File Type` dropdown set to `EPUB` with an EPUB result selected.

## Checklist

- [x] I ran `npm run build`
- [x] I ran `npm run lint`


j3lte:
Cool addition! I'll check it out tonight. Two questions:

- You removed some of the metadata in the details panel. Is that no longer relevant? As far as I can see in my current version, those types of data are still relevant and I would like to see them if available.
- The fast download option is cool, but this requires a key. You have made this optional in the extension preferences, but when you haven't filled it in, it will add an action to each item to set this key. Wouldn't it make more sense to say that you can fill in this key to enable fast download, and only when it is possible, you add this action? That way (I don't have an API key) it stays the same for everyone else.

Besides that, I think it would be good to then also add a small section to the README (which is visible in the store) that explains how you enable fast download.

j3lte:
@seanwessmith see previous comments, I got one more:
- Don't forget to add yourself as a contributor to package.json 😄 
```json
  "contributors": [
    "seanwessmith"
  ],
```

0xdhrv (approved):
Looks good to me, approved ✅

seanwessmith:
> ```json
>   "contributors": [
>     "seanwessmith"
>   ],
> ```

Sorry for the delay! I was OOO for the week. thanks for merging!

PR #28378, opened 2026-05-27
feat(skills): add Default Agents preference

### LINKED ISSUE #28377: [Skills] Add preference field that allows for default "Additional Agents" selections
### Extension

https://www.raycast.com/keito4/skills

### Description

This extension could benefit from introducing a field in preferences where the user sets their default coding agents, preventing the necessity of confirming installation for the same coding agent every time.

Image

Image

Users who only use a few coding agents that don't get covered on default install. Users who use Claude Code, Windsurf, etc. must add them as Additional Agents via the "Select agents..." form on every install

### Anything else?

I'd be willing to implement this. Thought I'd open it as an issue first.

alastairsounds:
## Description

Adds a "Default Agents" preference to the Skills extension where users can enter comma-separated agent display names (e.g. `Claude Code`) that will be automatically pre-checked in the install picker, so users don't have to manually select the same agents each time they install a skill.

Closes #28377

## Screencast

image

image

## Checklist

The extension still successfully installs skills

### DIFF
--- a/extensions/skills/src/components/actions/InstallSkillAction.tsx
+++ b/extensions/skills/src/components/actions/InstallSkillAction.tsx
@@ -176,7 +176,7 @@ function AgentPickerInstallForm({
   const selectableAgents = agents.filter((a) => !installedAgents.has(a));
   const defaultAgentNames = getDefaultAgents();
   const [selected, setSelected] = useState<Set<string>>(
-    new Set(selectableAgents.filter((a) => defaultAgentNames.includes(a))),
+    new Set(selectableAgents.filter((a) => defaultAgentNames.some((d) => d.toLowerCase() === a.toLowerCase()))),
   );
   const allSelected = selectableAgents.length > 0 && selected.size === selectableAgents.length;
   const isReplacing = installedMatch.type === "conflict";
--- a/extensions/skills/src/components/actions/InstallSkillAction.tsx
+++ b/extensions/skills/src/components/actions/InstallSkillAction.tsx
@@ -174,10 +174,10 @@ function AgentPickerInstallForm({
   const installedAgents = new Set<string>(installedAgentNames);
   const replacementAgentNames = installedMatch.type === "conflict" ? installedAgentNames : [];
   const selectableAgents = agents.filter((a) => !installedAgents.has(a));
-  const defaultAgentNames = getDefaultAgents();
-  const [selected, setSelected] = useState<Set<string>>(
-    new Set(selectableAgents.filter((a) => defaultAgentNames.some((d) => d.toLowerCase() === a.toLowerCase()))),
-  );
+  const [selected, setSelected] = useState<Set<string>>(() => {
+    const defaultAgentNames = getDefaultAgents().map((d) => d.toLowerCase());
+    return new Set(selectableAgents.filter((a) => defaultAgentNames.includes(a.toLowerCase())));
+  });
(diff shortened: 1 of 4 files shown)


keito4:
@alastairsounds 
Thanks for adding this! The feature makes sense!

One concern, though I haven’t verified the actual behavior locally: the current textfield approach requires users to type agent display names exactly as shown. If matching is case-sensitive, a mismatch like `claude code` vs `Claude Code`, or a small typo, could silently result in no agents being pre-selected.

Could we consider using separate checkbox preferences instead? Since the supported agent list seems finite, one checkbox per agent would avoid display-name parsing and case-sensitive matching issues.

If we keep the textfield, I think we should at least make the comparison case-insensitive and safely ignore unknown names.

alastairsounds:
Just implemented the case-insensitive fix. I'll create a separate pull request for the preference checkbox implementation, see which one y'all like better!

alastairsounds:
Created another implementation that uses checkboxes per agent, and places the checkboxes as a preference under the "Search Skills" command. See [28422](https://github.com/raycast/extensions/pull/28422)

image

extensions/skills/src/components/actions/InstallSkillAction.tsx
```diff
@@ -173,7 +174,10 @@ function AgentPickerInstallForm({
   const installedAgents = new Set<string>(installedAgentNames);
   const replacementAgentNames = installedMatch.type === "conflict" ? installedAgentNames : [];
   const selectableAgents = agents.filter((a) => !installedAgents.has(a));
-  const [selected, setSelected] = useState<Set<string>>(new Set());
+  const defaultAgentNames = getDefaultAgents();
+  const [selected, setSelected] = useState<Set<string>>(
+    new Set(selectableAgents.filter((a) => defaultAgentNames.some((d) => d.toLowerCase() === a.toLowerCase()))),
+  );
```
DaleSeo: The `getDefaultAgents()` and the `selectableAgents.filter(...).some(...)` scans run every time the form re-renders, and the resulting `Set` gets discarded. If we wrap the initializer in a function, it will only run on the first mount.

```suggestion
  const [selected, setSelected] = useState<Set<string>>(() => {
    const defaultAgentNames = getDefaultAgents().map((d) => d.toLowerCase());
    return new Set(selectableAgents.filter((a) => defaultAgentNames.includes(a.toLowerCase())));
  });
```

DaleSeo (approved):
Thanks for this, @alastairsounds! I only have a non-blocking comment.

keito4 (approved):
Sorry for the delayed response here, and thanks for addressing the feedback.

I think this is the better short-term approach. The checkbox implementation is better from a discoverability standpoint, but I hadn’t fully considered that we might need to maintain both the source code and package.json when agents are added or changed. I’ll look into a dynamic selection UI later and create a follow-up PR if needed. Thanks for your help.

The case-insensitive matching addresses my main concern with the text field approach, and ignoring unknown names is acceptable for this iteration.

This looks good to me!

alastairsounds:
Applied @DaleSeo's suggested fix. Rebuilt with `npm run lint`, `npm run dev`, and `npm run build` and confirmed it's working. Tested the changes and the extension works case-insensitively.

image

image

0xdhrv (approved):
Looks good to me, approved ✅

PR #28381, opened 2026-05-28
Add clipboard-qrcode extension

Reluxer:
## Checklist


0xdhrv:

We already have extensions in the Store that cover this functionality:

* [QR Code Generator](https://www.raycast.com/Melvynx/qrcode-generator) by @melvynx

This extension appears to overlap directly with the existing QR Code Generator extension. It already has a dedicated “Generate QR Code from Clipboard” command, along with QR generation from typed text / URLs and selected text.

The existing QR Code Generator extension also provides more complete functionality, including saving QR codes, showing them inline, copying them to the clipboard, PNG / SVG output, custom colors, link shortening, and UTM parameters.

The only small difference I can see is that this extension shows the clipboard content preview underneath the generated QR code. That seems useful, but it would likely fit better as an improvement to the existing QR Code Generator extension rather than as a separate extension.

Could we consider contributing this workflow or the content preview to the existing QR Code Generator extension instead?

> This would help avoid duplication and keep related functionality easier to discover.

PR #28385, opened 2026-05-28
Add restarter extension

b1shoy:
## Description

New extension: **Restarter** — restart or force-quit any running macOS app
from a fuzzy-searchable Raycast list.

It lists every running app (including menu-bar utilities like BetterDisplay
or BetterMouse, which the usual app-listing approaches skip) and lets you:

- Restart with `Enter` or force-quit with `⌘Enter` — both close the window
  immediately. `⇧` variants keep the window open so you can act on several
  apps in a row.
- Pin the apps you restart most often with `⌘⇧P`. The Pinned section sits
  at the top of the list, in your custom order, and is reorderable with
  `⌘⇧↑` / `⌘⇧↓`.
- Hide noisy apps you never want to see with `⌘H`. A second command,
  **Manage Hidden Apps**, lets you unhide them later.
- Live updates: the list polls every 500ms so apps disappear when they quit
  and reappear once they're back up.

App data comes from `lsappinfo`, which gives reliable results across both
foreground apps and `UIElement` menu-bar utilities without needing
Automation permission.

## Checklist


0xdhrv (commented):
Thanks for the contribution 🔥

This overlaps with Raycast’s built-in app management features: System Commands already include Quit All Apps / Quit All Apps Except Frontmost with exclusions, and Auto Quit supports quitting selected apps from Root Search / Settings. See manual.raycast.com/system-commands and manual.raycast.com/auto-quit.

The main difference here seems to be restarting or force-quitting a selected running app from a fuzzy list, including menu-bar utilities. Could you clarify why this should exist as a Store extension rather than relying on the built-in Raycast functionality?

PR #28393, opened 2026-05-28
Add curator extension

semantic-craft:
## Description

**Curator** is a command center for the skills installed for Claude Code and Codex. Three commands:

- **Search Skills** — fuzzy-search every installed skill (across both agents) by name, purpose, or trigger phrase; preview its `SKILL.md`; copy its name / `/command` / a trigger; open it in your editor. Skills shared across both agents merge into one row with surface badges.
- **Skill Doctor** — surface health issues (broken symlinks, missing/unparseable `SKILL.md`, name ≠ directory, Claude↔Codex drift, stale plugin cache), each with a copy-to-clipboard fix command. The extension is **read-only on the filesystem** — it never edits or deletes files itself.
- **Recommend Skills** — describe a task in natural language and an AI ranks the most relevant installed skills, each with a one-line reason. Uses **Raycast AI** by default, or any **OpenAI-compatible** provider via the user's own key (set in preferences).

It scans the standard Claude Code (`~/.claude/skills`, plugin marketplaces/cache) and Codex (`~/.codex/skills`) locations, and degrades gracefully when those don't exist (empty state) or when AI isn't available (points to Search Skills). The pure core is unit-tested (vitest); the UI is three thin `view` commands.

## Screencast

Screenshots are in `metadata/` — `curator-1.png` (Search Skills) and `curator-2.png` (Recommend Skills results). They use **fabricated demo skills**, not real data, for privacy.

## Checklist


semantic-craft:
Thanks @greptile-apps — addressed the review:

- **Generated `Preferences` type**: replaced the hand-written `Prefs` interface in `src/lib/llm.ts` with Raycast's generated `Preferences` type (tsconfig now includes `raycast-env.d.ts`; CI type-checks via `ray build`).
- **Empty model → provider 400**: the custom (OpenAI-compatible) path now also requires a model, so a blank "Custom Model" no longer sends `model: ""`.
- **Duplicated `firstSentence`**: extracted to `src/lib/text.ts` and shared by `recommend.ts` and `search-skills.tsx`, so both apply the same (full-width `！？`-aware) logic.
- **"Silent errors in `load()`"**: already resolved — Search Skills and Skill Doctor now load via `@raycast/utils` `usePromise` + `showFailureToast`.

semantic-craft:
Follow-up addressed in `e14dc29ea`:

- Guarded `Open in Editor` actions in Search Skills and Recommendations so broken symlink or missing `SKILL.md` entries no longer open an invalid editor target.

Validation: `npm test`, `npm run build`, `npm run lint`, `git diff --check` from `extensions/curator`. Store checks are green and there are no unresolved review threads.

0xdhrv (commented):

We already have an extension in the Store that deals with agent skills management. Could we consider enhancing the existing extension below instead of creating another one?

- [Skills](https://www.raycast.com/keito4/skills) by @keito4 

Your submission does have some standout features worth calling out — specifically `Skill Doctor` (local health checks for broken symlinks, missing/unparseable `SKILL.md`, name/directory drift, Claude↔Codex drift, and stale plugin cache detection) and `Recommend Skills` (AI-powered ranking of installed skills for a described task). These are distinct from what the existing extension offers and could be valuable additions.

If these workflows can be integrated into the existing extension, that would be ideal. But if the filesystem-direct, read-only auditing architecture makes integration impractical, we'd love to hear more about that so we can evaluate accordingly.

semantic-craft:
Thanks @0xdhrv — I checked the existing Skills extension and agree that avoiding duplicate skill-management surfaces is important.

Curator is intentionally scoped differently from that extension. The existing Skills extension is centered on Skills CLI workflows: search installed skills, manage/remove them, and update them. Curator is a read-only local audit and recall tool for Claude Code + Codex skill surfaces. Its main workflows are:

- `Search Skills`: fuzzy search across local Claude and Codex skill/plugin surfaces, including symlinked user skills and cached plugin skills.
- `Skill Doctor`: health checks for things that otherwise fail silently, such as broken symlinks, missing or unparseable `SKILL.md`, folder/name drift, Claude↔Codex drift, and stale plugin-cache versions.
- `Recommend Skills`: AI ranking grounded only in the local installed catalog, with validation that returned skill names actually exist.

Curator does not install, remove, update, or mutate skills. For fixes, it only shows/copies commands for the user to inspect and run themselves. That read-only audit boundary is a big part of the design, especially because it spans both Claude Code and Codex local layouts rather than only the Skills CLI surface.

So I think keeping it separate is cleaner: the existing Skills extension remains the install/manage/update tool, while Curator is the local doctor/recommend/search layer for users with multi-agent skill setups. If Raycast strongly prefers consolidation, I’m happy to coordinate with @keito4, but I wanted to explain why this submission is not just another installer/manager.

keito4:
Thanks for looping me in.

I took a look at the PR and @semantic-craft’s explanation. I agree that Curator is not a simple duplicate of the existing Skills extension: it adds local filesystem search across Claude Code and Codex skill locations, read-only health checks via `Skill Doctor`, and local-catalog-based recommendations.

At the same time, from a user perspective, I share the concern about having two skill-related extensions — especially if both expose a `Search Skills` command with different sources of truth. I think consolidating the user-facing workflows into the existing Skills extension would likely give users the clearest mental model, assuming @semantic-craft is open to exploring that path.

One possible integration path could be:

- keep the current skills.sh / CLI install-manage-update flows as they are
- add `Skill Doctor` and `Recommend Skills` as separate commands/modules
- either merge Curator’s local `search-skills` workflow into the existing search command, or rename/scope it clearly as local skill search
- preserve Curator’s read-only behavior for health-check fix actions

I think this would preserve the value of Curator’s implementation while avoiding duplicate skill-management surfaces.

Happy to coordinate with @semantic-craft on the integration path if that’s the direction Raycast prefers.

semantic-craft:
Thanks @keito4 for taking the time to review this so carefully and for laying out such a concrete integration path 🙏 I'm glad we agree the Skill Doctor / Recommend / local-search value is worth preserving.

I also share your concern about two extensions each exposing a `Search Skills` command with different sources of truth — from a user's mental-model perspective that's genuinely confusing, and I'd rather not add that friction to the Store.

So I'm open to exploring consolidation into the existing Skills extension along the lines you sketched:

- Keep your skills.sh / CLI install-manage-update flows exactly as they are — Curator never touched installation and shouldn't.
- Add `Skill Doctor` and `Recommend Skills` as separate commands/modules, preserving their read-only boundary (they only show/copy commands for the user to run, never mutate anything).
- For search, fold Curator's local filesystem scan into the existing search command — either as an additional source or a clearly-scoped "local skills" mode — so there's a single Search Skills surface rather than two.

The one thing I'd want to carry over is the cross-agent scope: Curator's Doctor and search span both `~/.claude` and `~/.codex` layouts (symlinked user skills, cached plugin skills, Claude↔Codex drift), not just the Skills CLI surface. If we can keep that under the consolidated extension, users get the best of both.

@0xdhrv — does routing this into @keito4's Skills extension (rather than a separate Curator submission) match what Raycast would prefer? If so, @keito4 and I can coordinate on the integration and I'll close this PR in favor of that path. Happy to do the implementation work on the Doctor / Recommend / local-search modules.

PR #28416, opened 2026-05-29
Add responsay extension

semantic-craft:
## Description

**Say It Right** is a new extension that turns any selected English text into a quick pronunciation lesson — for non‑native speakers, presenters, and anyone polishing their spoken English.

Select a sentence anywhere on macOS (or type one) and it shows **how to say it**: sentence stress, rising/falling intonation per thought group, rhythm, and connected‑speech linking, aligned right above the words, with General American IPA. It then reads the sentence aloud with neural TTS as a model you can **slow down (0.75× / 0.5×, pitch‑preserved), loop for shadowing, save to Downloads, and step through sentence‑by‑sentence.**

Highlights:
- **Two commands** — *Analyze Selection* (acts on the OS text selection, great with a hotkey) and *Practice a Sentence* (type/paste).
- **Bring‑your‑own key** — works with **OpenAI** *or* **Qwen (Alibaba DashScope)**; whichever key is configured is used automatically, and you can switch providers in‑view to compare.
- **Private by design** — no account, no telemetry, no file import. Keys live in Raycast preferences and are sent only to the provider you choose.

Full feature list and setup are in the README.

## Screencast

Two store screenshots (2000×1250) are included in `metadata/`: the analysis view (Stress & Intonation + Rhythm + IPA) and the extension in use.

## Checklist


semantic-craft:
Addressed the Raycast UI text nit: the manual intent translation placeholder is now in English ("Type what you mean in Chinese..."). Validation: npm test, npm run build, npm run lint; Store PR updated with ray publish.

semantic-craft:
Follow-up addressed in `ef31d1617`:

- Fixed Mouse Jiggle startup so a stale sentinel file no longer blocks restart after an unexpected Swift process death. The start command now verifies a live `mouse-jiggle.*jiggle.swift` process before returning “Already Running”; if none exists, it removes the stale sentinel and proceeds.

Validation: `npm run build`, `npm run lint`, `git diff --check` from `extensions/mouse-jiggle`. Store checks are green and there are no unresolved review threads.

semantic-craft:
Heads up for reviewers: I've renamed this extension from **Say It Right** (`say-it-right`) to **Responsay** (`responsay`) to align it with the product brand. The latest commit is a pure rename with no functional changes:

- Folder `extensions/say-it-right` → `extensions/responsay`
- `package.json`: name `say-it-right` → `responsay`; title and all command subtitles → `Responsay`
- README / CHANGELOG / in-app practice header / saved-audio filename prefix
- Metadata screenshots → `responsay-{1,2}.png`
- `.github/CODEOWNERS` and `.github/extensionName2Folder.json` entries updated and re-sorted into the correct alphabetical position

Local validation passes: `npm test` (154 tests), `tsc --noEmit`, `ray lint`, and `ray build`.

pernielsentikaer (commented):
Hi 👋

I guess, since it's not possible to use the extension without a key, then we should make the preferences required?

pernielsentikaer (commented):
Also feels a bit related to [Raycast Store: Say - Text to Speech](https://www.raycast.com/litomore/say) by @LitoMore

semantic-craft:
Thanks @pernielsentikaer — both points make sense.

On required preferences: Responsay supports four BYOK providers, and users only need to configure the provider they actually use. Making every API key required would force Qwen, MiMo, Gemini, and OpenAI credentials even for a single-provider setup. The extension validates this at runtime instead: it detects configured providers for the action-panel switchers, and if the selected/default provider is missing its key it shows a clear missing-key error. So the real requirement is "at least one usable provider key", which Raycast preferences cannot express as a single required field across several alternatives.

On the relation to Say: I agree there is a TTS touchpoint, but Responsay is not meant to be a general text-to-speech utility. Say uses the built-in macOS speech engine to read text aloud. Responsay is focused on spoken-English coaching: it analyzes stress, intonation, rhythm, thought-group pauses, IPA, and connected speech; turns Chinese intent or rough wording into natural English; explains why the phrasing works; then uses neural TTS only as the model voice for slow playback and shadowing practice.

Happy to tighten the README/metadata copy further if that distinction should be more explicit for Store review.

pernielsentikaer:
CC @LitoMore - what do you think here?

PR #28419, opened 2026-05-29
Update dev-servers extension

tavlean:
Adds a `Start Dev Server` command for spinning up dev servers without leaving Raycast. Works from a Finder selection, from a list of recently-seen projects, or from a native folder picker.

### Start Dev Server

- **From Finder**: select a project folder (or any file inside one) and run **Start Dev Server**. The extension walks up to the nearest `package.json`, detects the package manager (npm / pnpm / yarn / bun), picks the right script, and spawns it with the same PATH-aware login-shell pattern used by Restart. The dashboard opens immediately and shows a "Starting…" toast that transitions to "X is running" the moment the server binds a port.
- **From recents**: run the command with nothing selected in Finder and you get a picker over the projects the extension has seen running recently. Recents auto-populate from the dashboard's polling loop, with no explicit bookmarking. LRU-bounded at 30 entries. Running projects are hidden from this list (they live in the dashboard); they reappear once stopped.
- **From anywhere**: the picker also exposes a **Choose Folder…** entry that opens the native macOS folder dialog directly, with no intermediate screen.
- **From the dashboard**: the empty state offers a primary **Start Dev Server** action (just press `↵` from a fresh dashboard to land in the picker). Each running-server row's action panel also carries `Start Dev Server` (`⌘N`), so spinning up another project never requires bouncing back to root search.
- Each picker row shows last-seen, git branch when applicable, and a framework tag inferred from `package.json` dependencies. Cached favicons appear inline once the dashboard has seen the project running, so even stopped projects keep their real icon. Per-row actions: Start, Open in Terminal (`⌘T`), Show in Finder (`⌘⇧F`), Copy Path (`⌘C`), Remove from Recents (`⌃X`).
- Folders that no longer exist on disk are hidden this render but kept in storage so they reappear when (for example) an external drive remounts.
- **Startup logs**: every spawned server's stdout+stderr is captured to a per-project log. If a server doesn't bind a port within 15s, the toast escalates to a failure with a **View Startup Log** action instead of silently disappearing, so a misconfigured or custom setup (e.g. portless needing sudo, a missing binary, a crashing build) is diagnosable from inside Raycast. Every running-server row also carries a **View Startup Log** action (`⌘L`) for inspecting output on demand.

### Behavior

- Script picker tries `dev` → `start` → `develop` first, then scans script values for known dev-server tools (Vite, Next, Astro, Nuxt, Webpack, Parcel, Gatsby, Remix, Turbo, Bun watch/hot, nodemon, tsx watch, ts-node-dev, serve, http-server, live-server). Monorepo conventions like `dev:web` and `start:dev` resolve out of the box.
- Already running on that folder? You get one consolidated alert: `X is already running. Restart?` for a single target, `All 3 already running. Restart them?` when every selected folder is running, or `2 of 3 already running. Restart these, then start the other one?` for the mixed case. No more N-alert cascades for N-folder selections.
- The Finder selection is only honored when Finder is the frontmost app. Previously a folder selected earlier (to start one server) lingered in Finder's selection, so running the command later from another app, e.g. the browser, would silently treat that stale folder as the target and surface a spurious `already running. Restart?`. Now, unless you're actually in Finder, the command goes straight to the recents picker, where you can start whatever project you meant.
- After a server is started (from Finder, the picker, or a recent), the dashboard moves the selection onto that new row, so pressing `↵` acts on the server you just launched rather than re-opening whatever was previously selected. Restart (`⌘⇧R`) likewise re-focuses the replacement once it binds.
- Multi-folder selection prompts for confirmation by default, useful for monorepo siblings or a "frontend + backend" startup, with an opt-out preference for users who do this regularly.
- New **Open in browser when the port binds** preference auto-opens the URL once the new server starts listening. Off by default; a one-time hint surfaces it in the in-flight toast for the first few starts.

### Under the hood

- The dashboard is the controller for the entire spawn flow. The launching command resolves a target list and hands off via `launchContext`, which lets the user land on the dashboard immediately and watch the spawn happen there, rather than waiting on a blank loading view for a pre-spawn `fetchServers` call.
- Spawn lifecycle is a clean state machine on the dashboard: `idle → pending → confirming → spawning → done`. The "Starting…" toast lives on the dashboard so it's visible the whole time the user is waiting, and transitions to a green "running" state the moment every expected cwd appears in the polling loop.
- All filesystem paths flow through `canonicalCwd` (a `realpathSync` wrapper) so symlinked project paths compare equal between Finder selections, the recents store, and `lsof`'s view of running processes.
- Extracted `startDevServer(cwd)` and `killServer(pid)` so the new command and the existing restart flow share one spawn path. Restart is now `killServer + startDevServer`.
- The spawn passes the package manager and chosen script as separate arguments rather than building a shell string, so projects with unusual script names (spaces, punctuation) start reliably and the launch surface stays free of shell-interpolation surprises.
- Shared `tool-display.ts` so the framework tag styling stays consistent across the dashboard and the picker.
- Favicons cached onto recents for stopped-project icons are size-capped, keeping the recents store small; the live dashboard always renders the real favicon regardless.

### Preferences

- **Terminal App** is now a single shared preference that applies to both **Dev Servers** and **Start Dev Server**: set the terminal `⌘T` opens once, and both commands honor it.

## Checklist

### DIFF
--- a/extensions/dev-servers/CHANGELOG.md
+++ b/extensions/dev-servers/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Dev Servers Changelog
 
-## [Start Dev Server] - {PR_MERGE_DATE}
+## [Start Dev Server] - 2026-06-01
 
 Adds a `Start Dev Server` command for spinning up dev servers without leaving Raycast. Works from a Finder selection, from a list of recently-seen projects, or from a native folder picker.
 
--- a/extensions/dev-servers/CHANGELOG.md
+++ b/extensions/dev-servers/CHANGELOG.md
@@ -1,24 +1,61 @@
 # Dev Servers Changelog
 
-## [Custom domain detection] - 2026-05-26
+## [Start Dev Server] - {PR_MERGE_DATE}
+
+Adds a `Start Dev Server` command for spinning up dev servers without leaving Raycast. Works from a Finder selection, from a list of recently-seen projects, or from a native folder picker.
+
+### Start Dev Server
+
+- **From Finder**: select a project folder (or any file inside one) and run **Start Dev Server**. The extension walks up to the nearest `package.json`, detects the package manager (npm / pnpm / yarn / bun), picks the right script, and spawns it with the same PATH-aware login-shell pattern used by Restart. The dashboard opens immediately and shows a "Starting…" toast that transitions to "X is running" the moment the server binds a port.
+- **From recents**: run the command with nothing selected in Finder and you get a picker over the projects the extension has seen running recently. Recents auto-populate from the dashboard's polling loop, with no explicit bookmarking. LRU-bounded at 30 entries. Running projects are hidden from this list (they live in the dashboard); they reappear once stopped.
+- **From anywhere**: the picker also exposes a **Choose Folder…** entry that opens the native macOS folder dialog directly, with no intermediate screen.
+- **From the dashboard**: the empty state offers a primary **Start Dev Server** action (just press `↵` from a fresh dashboard to land in the picker). Each running-server row's action panel also carries `Start Dev Server` (`⌘N`), so spinning up another project never requires bouncing back to root search.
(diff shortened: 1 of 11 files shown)


0xdhrv (approved):
Looks good to me, approved ✅

PR #28422, opened 2026-05-29
feat(skills): add per-agent checkbox preferences

alastairsounds:
## Description

This feature adds preference checkboxes to pre-select non-standard coding agents (Claude Code, Windsurf, etc..) The checkboxes are nested under the "Search Skills" command in preferences for the following reasons: 1.) "Search Skills" is the only command that utilizes the agent picker, and 2.) it allows the 54 checkboxes to be kept in a dropdown, rather than muddying up the entire preferences area.

**Note:** `ray lint` requires a `description` field per checkbox. I opted for each description to follow the pattern `"Pre-select [Agent] in the install agent picker"`, but can adjust if we want to make it less verbose.

**Possible README/store update:** This feature might be worth updating in the README so users can see it on the Store page. Something like `"Search for agent skills (pre-select default agents in Settings → Extensions → Skills → Search Skills)"`. If not, users may still naturally find it via the preferences menu.

## Screencast

image

image

## Checklist

### DIFF
--- a/extensions/skills/CHANGELOG.md
+++ b/extensions/skills/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Skills Changelog
 
+## [Default Agents Preference] - {PR_MERGE_DATE}
+
+- Adds per-agent checkboxes in extension preferences to pre-select agents in the install picker
+
 ## [Show Installed Skill Audit Details] - 2026-05-20
 
 - Add an "Open on skills.sh" action and `skills.sh` detail link for published installed skills
--- a/extensions/skills/package.json
+++ b/extensions/skills/package.json
@@ -28,7 +28,495 @@
       "title": "Search Skills",
       "subtitle": "Skills",
       "description": "Search for agent skills",
-      "mode": "view"
+      "mode": "view",
+      "preferences": [
+        {
+          "name": "agent_adal",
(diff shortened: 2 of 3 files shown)


DaleSeo:
> **Possible README/store update:** This feature might be worth updating in the README so users can see it on the Store page. Something like `"Search for agent skills (pre-select default agents in Settings → Extensions → Skills → Search Skills)"`. If not, users may still naturally find it via the preferences menu.

@alastairsounds Good call on documenting it. How about adding a Features bullet? Also, could you address Greptile's review? Thanks!

keito4:
Thanks for putting together this alternative implementation, and sorry for the extra back and forth here.

As noted in my review on #28378, I think we should move forward with the text field approach for now:
https://github.com/raycast/extensions/pull/28378#pullrequestreview-4402584240

The checkbox implementation is better from a discoverability standpoint, but I hadn’t fully considered the maintenance cost of keeping both the source code and package.json in sync whenever agents are added or changed. Since #28378 addresses the main concern with case-insensitive matching and keeps the implementation smaller, I think it would be best to close this PR in favor of #28378.

I’ll look into a dynamic selection UI later and create a follow-up PR if needed. Thanks again for your help with this.

PR #28441, opened 2026-05-30
Update Panicly extension

6crucified:
## Description

Updates Panicly Requests so setup uses a normal Panicly project API key instead of a browser session cookie. The View Requests command calls Panicly's read-only Raycast requests endpoint, shows recent gateway request decisions, project/route/provider/model context, tokens, cost estimates, timestamps, and request detail views with copy/open actions.

Configuration is now intentionally small: Panicly API Key is the only required preference. Panicly Base URL and Request Limit are optional preferences with defaults.

## Screencast

Not attached. This command requires an authenticated Panicly project API key with request traffic to display live data.

## Checklist

- [x] I ran `npm run lint`
- [x] I ran `npm run build`
- [x] I ran `npm run dev` and verified the command compiles/imports in Raycast development mode
- [ ] I tested with a live authenticated Panicly project API key


6crucified:
Backend endpoint is now deployed on production at `https://panicly.vercel.app/api/raycast/requests`. Unauthenticated and invalid-key requests return JSON 401 responses, so the extension no longer relies on browser cookies.

PR #28449, opened 2026-05-31
Add Windows support to Claude Code Usage (ccusage) extension

owendavidprice:
## Description

Adds Windows support to the **Claude Code Usage (ccusage)** extension so it runs on the Raycast for Windows beta. No new features — purely platform-compatibility fixes, and macOS behaviour is unchanged.

The extension shells out to `npx ccusage` via Raycast's `useExec`, which calls `child_process.spawn` directly. Three things broke on Windows:

- **`spawn("npx", …)` throws `ENOENT`** — Windows needs the `npx.cmd` shim, which is only found when the command runs through a shell. Now passes `shell: true` on Windows so `cmd.exe` resolves it via `PATHEXT`.
- **`PATH` was Unix-only** — built with the `:` separator and hard-coded Homebrew/nvm paths. Now uses `path.delimiter` and `os.homedir()`, with a Windows branch that adds `%APPDATA%\npm`.
- **Relied on `$HOME`** (unset on Windows) — switched to `os.homedir()` throughout.

Unix-only version-manager (nvm/fnm/n/volta) probing and env vars are gated to non-Windows. `keychain-access` already fell back to reading `~/.claude/.credentials.json` (works on Windows via `homedir()`), and the usage-limits client uses `fetch`, so both were already cross-platform.

## Changes

- `src/utils/exec-options.ts` — `shell: true` on Windows; `path.delimiter`; `os.homedir()`; gate nvm/fnm env vars to Unix
- `src/utils/node-path-resolver.ts` — Windows PATH branch (`%APPDATA%\npm`, nodejs dir); `path.delimiter` / `homedir()`; skip Unix version-manager probing on Windows
- `package.json` — add `Windows` to `platforms`
- `CHANGELOG.md` — new entry

## Checklist

- [x] `npm run build` succeeds (verified on Windows)
- [x] Verified `ray develop` loads the extension into Raycast on Windows and shows live ccusage data across all views
- [x] Updated `CHANGELOG.md` with a new entry using the `{PR_MERGE_DATE}` placeholder
- [x] No macOS regressions — all platform-specific code is gated behind `process.platform === "win32"`

### DIFF
--- a/extensions/ccusage/src/utils/exec-options.ts
+++ b/extensions/ccusage/src/utils/exec-options.ts
@@ -1,6 +1,9 @@
 import { getEnhancedNodePaths, resolveFnmBaseDir } from "./node-path-resolver";
-import { dirname } from "path";
+import { delimiter, dirname } from "path";
+import { homedir } from "os";
 import { getCustomNpxPath } from "../preferences";
+
+const isWindows = process.platform === "win32";
 
 export const getExecOptions = () => {
   const env: Record<string, string> = {
@@ -12,28 +15,34 @@ export const getExecOptions = () => {
   const customNpxPath = getCustomNpxPath();
   if (customNpxPath) {
     const customDir = dirname(customNpxPath);
-    env.PATH = `${customDir}:${env.PATH}`;
+    env.PATH = `${customDir}${delimiter}${env.PATH}`;
   }
 
-  // Add HOME-dependent paths only if HOME is available
-  if (process.env.HOME) {
+  // Version-manager env vars (nvm/fnm/npm-global) only apply to the
+  // Unix layout. Windows uses %APPDATA%\npm and PATH directly.
+  const home = homedir();
+  if (!isWindows && home) {
     if (!process.env.NVM_DIR) {
-      env.NVM_DIR = `${process.env.HOME}/.nvm`;
+      env.NVM_DIR = `${home}/.nvm`;
     }
     if (!process.env.FNM_DIR) {
-      const fnmBaseDir = resolveFnmBaseDir(process.env.HOME);
+      const fnmBaseDir = resolveFnmBaseDir(home);
       if (fnmBaseDir) {
         env.FNM_DIR = fnmBaseDir;
       }
     }
     if (!process.env.npm_config_prefix) {
-      env.npm_config_prefix = `${process.env.HOME}/.npm-global`;
+      env.npm_config_prefix = `${home}/.npm-global`;
     }
   }
 
   return {
     env,
     timeout: 30000,
-    cwd: process.env.HOME || process.cwd(),
+    cwd: home || process.cwd(),
+    // On Windows, `npx`/`ccusage` are `.cmd` shims that Node's spawn cannot
+    // execute directly (ENOENT). Running through a shell lets cmd.exe resolve
+    // them via PATHEXT. Our args are simple tokens, so shell quoting is safe.
+    ...(isWindows ? { shell: true } : {}),
   };
 };
--- a/extensions/ccusage/src/utils/exec-options.ts
+++ b/extensions/ccusage/src/utils/exec-options.ts
@@ -11,8 +11,7 @@ export const getExecOptions = () => {
 
   // PATH is case-insensitive on Windows and conventionally spelled `Path`.
   // Resolve the real key so we overwrite the existing variable instead of
-  // adding a second, ambiguous `PATH` entry. cross-spawn (used by execa and
-  // useExec) reads the same key when resolving the command.
+  // adding a second, ambiguous `PATH` entry.
   const key = pathKey({ env });
   let path = getEnhancedNodePaths(env[key] ?? "");
 
@@ -45,5 +44,11 @@ export const getExecOptions = () => {
     env,
     timeout: 30000,
     cwd: home || process.cwd(),
+    // `useExec` (the views) spawns via raw `child_process.spawn`, which on
+    // Windows throws ENOENT for the `npx.cmd` / `ccusage.cmd` shims unless run
+    // through a shell that honours PATHEXT. Enabling the shell on Windows fixes
+    // the views; execa (the AI tools) tolerates it. Our args are fixed tokens
+    // and validated dates, so there is no shell-quoting risk.
+    shell: isWindows,
   };
 };
--- a/extensions/ccusage/src/utils/exec-options.ts
+++ b/extensions/ccusage/src/utils/exec-options.ts
@@ -40,9 +40,12 @@ export const getExecOptions = () => {
     env,
     timeout: 30000,
     cwd: home || process.cwd(),
-    // On Windows, `npx`/`ccusage` are `.cmd` shims that Node's spawn cannot
-    // execute directly (ENOENT). Running through a shell lets cmd.exe resolve
-    // them via PATHEXT. Our args are simple tokens, so shell quoting is safe.
-    ...(isWindows ? { shell: true } : {}),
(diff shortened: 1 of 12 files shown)


extensions/ccusage/src/utils/exec-options.ts
```diff
@@ -1,7 +1,10 @@
 import { getEnhancedNodePaths, resolveFnmBaseDir } from "./node-path-resolver";
-import { dirname } from "path";
+import { delimiter, dirname } from "path";
+import { homedir } from "os";
 import { getCustomNpxPath } from "../preferences";
 
+const isWindows = process.platform === "win32";
+
 export const getExecOptions = () => {
```
bendrucker, line 9: The extension has two exec paths that behave differently on Windows. The views call `useExec`, which wraps execa and uses `cross-spawn` to resolve `.cmd` / `PATHEXT` shims and to read PATH case-insensitively via `path-key`. The AI tools in `src/tools/` call raw `child_process.exec`, which goes through none of that. Every Windows branch in this PR exists to hand-build, on the second path, what the first path already gets from the dependency.

I'd suggest routing the AI tools through the same execa call the views use (`execa("npx", ["ccusage@latest", ...])`, or `cross-spawn` directly) instead of `child_process.exec` with a manually assembled environment. That removes the divergence at the source. Concretely, it lets this PR drop:

- The `shell: cmd.exe` option (line 49). `child_process.exec` already runs through cmd.exe via `ComSpec`, and execa resolves `.cmd` shims through `PATHEXT`, so forcing a shell is redundant and only adds cmd.exe quoting risk ([cmd command reference](https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/cmd)).
- The `{ ...process.env, PATH }` casing hazard (line 9). On Windows the variable is conventionally `Path`, names are case-insensitive, and the environment block is sorted case-insensitively, so spreading `process.env` and then adding a second `PATH` key leaves two entries for one logical variable with ambiguous precedence ([environment variable names are case-insensitive on Windows](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_environment_variables), [the environment block is sorted case-insensitively](https://learn.microsoft.com/en-us/windows/win32/procthread/changing-environment-variables)). `cross-spawn` handles this for you; if you still inject PATH yourself, write it back to the existing key (for example via `path-key`) rather than adding a new one.
- The `isWindows` branches in `node-path-resolver.ts` and the `getWindowsNodePaths` helper, including its `ProgramFiles\nodejs` entry (which the inherited PATH already covers, and which the helper's own comment does not mention).

The one Windows detail likely worth keeping is `%APPDATA%\npm`, the npm global prefix where global `.cmd` shims live, if Raycast does not inherit the full user PATH. If you find that's needed, prepend just that one directory rather than reconstructing the PATH per platform.

I'd want the env-casing behavior confirmed on a real Windows box regardless of which approach we take, since it determines whether the tools resolve `ccusage.cmd` at all.

bendrucker (changes requested):
The Windows changes here mostly re-implement, on a second code path, behavior that a dependency already in the tree provides. The inline comment describes that in more detail. I'm not a Windows user so any references to behavior there is just from public sources, please do verify them.

There are other code quality comments I'd leave but most of them will fall away when cross-platform details are delegated to dependencies.

owendavidprice:
Heads-up for anyone skimming the run history: the first CI run on this PR went red on a TypeScript error (`shell: true` isn't assignable to `child_process.exec`'s `ExecOptions.shell`, which is `string`-only). That was resolved in the follow-up commit `9828148c`, which sets the shell to the `cmd.exe` path (`process.env.ComSpec`, a string) — type-compatible with both `exec` and `useExec`, and functionally equivalent on Windows.

All checks are green on the current head commit (build, ESLint, Prettier, `tsc`, Socket Security, Greptile). The earlier red run is superseded and can be ignored.

owendavidprice:
Thanks for the thorough review — you're right that the cleaner fix is to remove the second exec path rather than re-implement cross-platform behaviour on it. Pushed `12324bf6` doing that.

**What changed**

- The AI tools (`src/tools/*`) and `claude-code-stats` now call a small `runCcusage()` helper that invokes the CLI through **`execa`** (the same library `useExec` uses) with an args array, instead of `child_process.exec` with a hand-assembled command string. `cross-spawn` handles `.cmd`/`PATHEXT` resolution, so both exec paths now behave identically.
- **Dropped the `shell` option entirely.** As you noted, `exec` already ran through `cmd.exe` and `execa` resolves shims via `PATHEXT`, so forcing a shell was redundant.
- **Fixed the PATH-casing hazard.** `getExecOptions` now resolves the real key with **`path-key`** and writes back to it, instead of spreading `process.env` and adding a second `PATH` entry. This was the bug you flagged — on Windows the inherited variable is `Path`, so the old code left two entries with ambiguous precedence.
- **Reduced the `node-path-resolver` Windows branch** to just the npm global prefix (`%APPDATA%\npm`), the one location you identified as worth keeping. Removed `getWindowsNodePaths` and its `ProgramFiles\nodejs` entry (already covered by the inherited PATH).

`execa` and `path-key` were already in the tree via `@raycast/utils`; they're now direct dependencies at the same versions (`execa@5.1.1`, `path-key@3.1.1`).

**Verified on a real Windows box** (the env-casing behaviour you asked about specifically): with `execa` and no shell, `path-key` resolves the PATH key and `npx ccusage@latest ... --json` resolves `npx.cmd` and returns data correctly. `tsc`, `ray build`, ESLint, Prettier and the Jest suite (15 tests) all pass; the extension loads via `ray develop` and the views render live data.

You mentioned other code-quality points that would fall away once this was delegated to the dependency — happy to take those on if any remain.

owendavidprice:
Update after testing on a real Windows box (Raycast for Windows beta) — there's a wrinkle worth flagging.

After the execa refactor, the **views still failed** with "ccusage command failed", so I added runtime diagnostics inside Raycast's extension process to capture exactly what happens. Findings:

1. Raycast **does** inherit the full user PATH (the Node dir was present), and `npx ccusage@latest daily --json` runs fine **via execa** from that process. So the AI tools (now on execa) work, and the PATH/case handling is sufficient — no registry/PATH reconstruction needed. 👍
2. But the **views use `useExec`**, and in the `@raycast/utils` version here `useExec` spawns through **raw `child_process.spawn`**, not cross-spawn. On Windows that throws `ENOENT` for the `npx.cmd` shim unless run through a shell. Minimal repro on the same box:

   ```js
   child_process.spawn("npx", ["--version"])               // → ENOENT
   child_process.spawn("npx", ["--version"], {shell:true}) // → exit 0
   ```

So the cross-spawn delegation we were counting on for `useExec` doesn't hold in this version — the views genuinely need a shell on Windows. The latest commit (`c665895a`) therefore:

- Enables `shell` **on Windows only** in `getExecOptions` (fixes the views; execa tolerates it, and the args are fixed tokens + validated dates, so no quoting risk).
- Keeps the execa refactor for the AI tools / `claude-code-stats`.
- **Drops** the manual PATH/casing reconstruction beyond `path-key` + `%APPDATA%\npm`, since the diagnostics confirmed it wasn't needed.

If you'd prefer the `shell` flag scoped to just the `useExec` call sites rather than shared `getExecOptions` (the execa tools don't strictly need it), I'm happy to split it — let me know which you prefer. Verified end to end: the views now render live data in Raycast on Windows, and `tsc`/`ray build`/ESLint/Prettier/15 Jest tests pass.

bendrucker:
Hmm seems like this could be worth fixing upstream versus working around in the extension, even in this narrower way

pernielsentikaer:
Drafting this until we have time to look into the other PR you opened @bendrucker

PR #28450, opened 2026-05-31
Update raycast-motion-preview extension

ayarse:
## Description

This is an update to the existing **Motion Preview** extension. Motion Preview previews the animation file currently selected in Finder in a floating window, supporting Rive (`.riv`), Lottie JSON (`.json`), and dotLottie (`.lottie`).

This release focuses on turning the preview from a one-shot viewer into something you can actually navigate and control.

### Added

- **Folder carousel** — a thumbnail strip lets you browse and jump between every animation in the same folder, without going back to Finder to reselect.
- **Playback controls** — a floating control bar to play/pause and scrub through an animation's timeline.
- **Background color control** — switch the preview background, including a transparent option, to check how an animation reads on different surfaces.
- **dotLottie state machines** — state machine switching now works for dotLottie files, alongside the existing Rive state machine support.

### Changed

- The preview window is now sized to the animation's own aspect ratio instead of a fixed frame.

## Checklist


pernielsentikaer (changes requested):
Hi 👋

Thanks for the update @ayarse 🧑‍🚒

Is there a way NOT to add the binaries since we have no cotrol over what they actually do?

You can see a bit more here: https://developers.raycast.com/basics/prepare-an-extension-for-store#binary-dependencies-and-additional-configuration

---

ayarse:
Hey @pernielsentikaer, thanks for taking a look! 🙏

Just confirming I'm addressing the right thing: I assume you mean the `.wasm`/`.js` runtimes in `assets/lib/`? 

**Why they're there:** the extension renders Rive and Lotties inside a `WKWebView`, which needs the official runtimes ([`@rive-app/canvas`](https://www.npmjs.com/package/@rive-app/canvas) and [`@lottiefiles/dotlottie-web`](https://www.npmjs.com/package/@lottiefiles/dotlottie-web), each with a `.wasm` core). I used to pull these from the jsDelivr CDN at runtime, so now I vendored them locally to make the preview work fully offline instead of fetching on every open.

**Would this work as a fix**: If I pin both packages as `devDependencies` and add a small sync script that copies the runtime files out of `node_modules` into `assets/lib/`?

Happy to explore another route if you'd prefer.

pernielsentikaer:
That would be a better way yeah, so we know more about where they originate from

ayarse:
I ended up taking a different route than the devDependency + copy idea. Rather than vendoring the runtimes, I removed them from the repo and fetch them on first run instead (`src/runtimes.ts`). They're pulled from version-pinned jsDelivr URLs (`@rive-app/canvas@2.38.0`, `@lottiefiles/dotlottie-web@0.74.0`), each checked against a hardcoded SHA-256, then cached locally so it stays offline-friendly after the first open.

No binaries in the repo now, and it's clear where they come from right in the code.

pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #28456, opened 2026-05-31
Add choose-browser extension

aryankush25:
## Description

**Choose Browser** lists all browsers installed on your Mac and lets you set any one as the system default — directly inside Raycast, with no dialogs or system popups.

- Detects every browser registered as an HTTP/HTTPS handler using the `defaultbrowser` CLI
- Finds each app bundle live via Spotlight (`mdfind`) and reads its display name from `Info.plist` — no hardcoded browser list
- Shows the current default pinned at the top with a **Default** tag
- Pressing Enter on any browser sets it as default immediately with a toast confirmation
- Auto-dismisses the macOS system confirmation dialog via AppleScript

**Requirement:** [`defaultbrowser`](https://github.com/kerma/defaultbrowser) must be installed (`brew install defaultbrowser`).

## Screencast

https://github.com/user-attachments/assets/placeholder

## Checklist


0xdhrv (commented):

- [default-web-browser-manager](https://www.raycast.com/clins1994/default-web-browser-manager) by @clins1994 

PR #28459, opened 2026-05-31
Update Plexus: Windows + WSL support, detect all localhost web servers

RensHoogendam:
## Description

This updates **Plexus** with Windows support and broadens it from "Node.js servers" to **any localhost web server**.

### What's new
- **Windows support** (`platforms: ["macOS", "Windows"]`). Discovery, process info, and "Kill process" now work on Windows.
  - Discovery uses `netstat` for the port list and a single CIM query (only for servers that pass the probe) for command lines — chosen after profiling, because `powershell.exe` startup alone was ~1s.
  - Kill uses `taskkill` on Windows, `kill` on macOS.
- **WSL-hosted servers** are detected too. Node/PHP/etc. servers running inside WSL are forwarded to `localhost` but owned by `wslrelay` on the Windows side, so they're enumerated via `wsl.exe` (`ss` + `/proc`), shown with a `WSL: <distro>` tag, project names read over `\\wsl.localhost\...`, and killed via `wsl kill`.
- **Any web server, not just Node.** Detection now enumerates every listening TCP port and keeps the ones that answer an HTTP request with a 2xx/3xx + `text/html` response. This surfaces PHP/Laravel/Python/etc. dev sites and hides non-web services (Postgres, Redis, MySQL) and desktop-app IPC servers (Steam, Logitech, SignalRGB) that reply with errors or non-HTML.
- **Faster + cached.** Cross-platform discovery defers per-process lookups to the few ports that pass the probe; a short raw-socket probe replaces `fetch` (which crashes Node's HTTP parser on non-HTTP ports). The command UI uses `useCachedPromise`, so reopening is instant. End-to-end discovery went from ~3.3s to ~0.7s on Windows.

### Cross-platform behavior
- macOS path keeps using `lsof`/`ps` and is unchanged in behavior aside from the broadened detection (and now lists non-Node web servers too).
- Naming resolves from `package.json`/folder name (web roots like `public/` resolve to the parent), with the page `<title>` taking over in the UI.

### Notes / limitations
- HTTPS-only dev servers aren't detected yet (would need a TLS probe ignoring self-signed certs).
- Windows Herd `.test` vhosts on port 80 aren't listed (served by hostname, not per-port).

## Checklist
- [x] `npm run build` succeeds (macOS + Windows toolchain)
- [x] `npm run lint` passes
- [x] Updated `CHANGELOG.md` and bumped the version
- [x] Verified end-to-end on Windows (native + WSL servers) and via the macOS code path


0xdhrv (commented):
Hey @RensHoogendam 👋

Can you please look into the failing checks?

---

0xdhrv (approved):
Looks good to me, approved ✅

PR #28463, opened 2026-05-31
Add jewish-date extension

jeremieberrebi:
## Description

This extension adds a simple Raycast command to get today's Jewish date written in Hebrew.

The command opens a compact list with two explicit choices:

- **Before Shkia-Sunset**: copies the Jewish date for the current civil day.
- **After Shkia-Sunset**: copies the Jewish date for the following Jewish day, since the Jewish calendar date changes after sunset.

Selecting either option and pressing `Enter` copies the formatted date directly to the clipboard. The output is fully written in Hebrew letters, for example:

```text
ט"ו סיון תשפ"ו
```

The extension is intentionally small and focused: it does not require setup, preferences, location access, network calls, or third-party APIs. It relies on the built-in JavaScript `Intl.DateTimeFormat` Hebrew calendar support to get the Hebrew calendar components, then formats the day and year with Hebrew numerals so users get the traditional readable form.

## Use Case

This is useful for users who frequently need to paste the current Jewish date into messages, notes, documents, forms, certificates, community announcements, or religious/lifecycle-related records.

The before/after Shkia-Sunset choice avoids ambiguity in the evening, when the Jewish date has already advanced even though the civil date has not.

## Validation

- Ran `npm run lint`
- Ran `npm run build`
- Tested locally with `npm run dev`

## Notes

No additional configuration is required.


0xdhrv (commented):

We already have an extension in the Store that deals with Hebrew dates and Zmanim. Could we consider enhancing the existing extension below instead of creating another one?

* [Hebrew Date & Zmanim](https://www.raycast.com/mendel_g/hebrew-date-zmanim) by @mendelg

The Hebrew-script / gematria formatting in this PR looks like a useful differentiator, since the existing extension currently appears to output Hebrew dates in English transliteration. If that’s the main workflow you’re aiming to add, we’d love to see whether it can be integrated into the existing extension to avoid duplication and improve discoverability.

jeremieberrebi:
Hello

Sorry for the delay. If I can upgrade the extension from @mendelg, I can effectively try.

There are more information in his extension but not simple answers like what I've done.

Personally, I was using both extension because I'm using them for two different type of usage

- Jeremie Berrebi

PR #28135, opened 2026-05-19
klack: rebuild with stats and standalone support

JaceThings:
## Summary

Klack 2.1.x added a new `current stats` AppleScript command. This PR rebuilds the extension on top of the modern Klack scripting dictionary, expands its surface, and ships a live stats view. Done with Henrik's blessing — he invited me as a collaborator on his fork to land this update.

The store identity (`name: klack`, `author: henrikruscon`) is preserved so stars, install count, and the existing listing URL carry over.

## What's new

- **Klack Stats** view — live keystroke / ding / click counts and favourite-switch breakdown, refreshing while open. Per-switch tinted tag chips that adapt to light vs. dark Raycast themes. "Copy Stats as Markdown" action and a "Tracking since" line that surfaces Klack's `firstTrackedAt`.
- **Set Volume** view — 0–100 scrubber with a custom-input form.
- **Turn Klack on / Turn Klack off / Wake Klack** — explicit state commands using the v2 sdef.
- **AI extension tools** — `toggle`, `turn-on`, `turn-off`, `wake-up`, `set-switch`, `set-volume`, `get-status`. All include runtime guards against unknown inputs.
- **Super Red** switch (Keychron).
- **Per-switch icons** in the picker.
- **Redundancy-aware feedback** — every command reports when Klack is already in the requested state instead of silently no-op'ing.
- **Standalone Klack support** — the [tryklack.com](https://tryklack.com) build shares the bundle ID and sdef, so this extension drives both.

## What's fixed

- Switch search is case-insensitive.
- "Current" badge correctly identifies multi-word switch names (e.g. "Crystal Purple", "Japanese Black").
- `Select` action no longer requires the initial `current switch` fetch to have resolved.
- Set Volume to Soft / Balanced / Loud work under Klack v2 (the old `volume preset` enum no longer exists in the sdef — these now call the integer `volume` command directly).
- Error toasts distinguish "Klack is not installed" vs. "Klack needs an update" vs. "Permission required" instead of one generic message.
- Install check no longer leaks a rejected promise after a transient `getApplications()` failure.

## Performance

- Multi-field state reads collapsed into one AppleScript spawn.
- Per-process install-check memoization (and a TTL'd cache key for cross-process cold starts).
- Optimistic mutations with `shouldRevalidateAfter: false` — no post-action refetch.
- `useCachedPromise` seeded from a persistent `Cache` for instant cold-start paint across all commands.
- `closeMainWindow()` no longer blocked-on for no-view commands.
[...]

PR #28472, opened 2026-06-01
klack: polish stats header and tighten changelog

JaceThings:
## Summary

Two small follow-ups to the recent [Stats and Standalone Support](https://github.com/raycast/extensions/pull/28135) release:

- **Stats header** — drops the leading "Tracking" word so the section now reads `Total Usage · Since May 20, 2026` instead of `… · Tracking since May 20, 2026`. Henrik flagged "tracking" as sounding negative in casual UI copy and asked to simplify.
- **Changelog** — tightens the bullets on the existing `[Stats and Standalone Support] - 2026-05-20` entry so the release notes read more like Klack's earlier entries (short, user-facing) and less like an internal commit log. Dates, section headers, and all other entries are untouched.

## Test plan

- [x] `npx ray build -e dist`
- [x] `npx ray lint`
- [x] Stats view manually verified — header now reads `Total Usage · Since May 20, 2026`

### DIFF
--- a/extensions/klack/CHANGELOG.md
+++ b/extensions/klack/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Klack Changelog
 
+## [Polish] - {PR_MERGE_DATE}
+
+- Revised wording in the Stats view
+
 ## [Fix README Link] - 2026-05-20
 
 - Fixed broken contributor GitHub link in the README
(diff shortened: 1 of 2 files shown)


0xdhrv (approved):
Looks good to me, approved ✅

PR #28474, opened 2026-06-01
[Search Router] Make query argument optional to support fallback search

janosorcsik:
## Description

This PR fixes fallback command support, resolves a build-breaking script bug, and addresses a missing `await` on an async call.

## Bug Fixes

- **Fallback search support** — `query` argument changed to `required: false` so the command works when Raycast launches it as a fallback search (previously the command would not launch without a pre-filled query)
- **Fix `download-kagi-bangs` script crash** — `utils.ts` imported `@raycast/api` which Node.js cannot resolve; moved `safeOpenUrl` (the only Raycast-dependent function) into `search.tsx` where it's used, leaving `utils.ts` dependency-free
- **Fix generated export name** — script was generating `export const searchEngines` but all importers expected `builtinSearchEngines`; corrected the template and regenerated
- **Fix missing `await` on `showToast`** in `BrowseSearchEngines.setAsDefault`

## Data

- Update bang list from Kagi

## Packages

- Update dependencies

## Checklist


extensions/search-router/src/browse-search-engines.tsx
```diff
@@ -201,8 +201,8 @@ export default function BrowseSearchEngines() {
                       style={Action.Style.Destructive}
                       onAction={() => handleDeleteEngine(searchEngine.s, searchEngine.t)}
                       shortcut={{
-                        modifiers: ["ctrl"],
-                        key: "x",
+                        macOS: { modifiers: ["cmd"], key: "x" },
```
janosorcsik: Hi @0xdhrv & @raycastbot,
I changed the shortcut to Ctrl+X globally because that is the default for deleting in Raycast.

0xdhrv: Thanks, I have updated the same

0xdhrv (approved):
Looks good to me, approved ✅

PR #28476, opened 2026-06-01
Update producthunt extension

chrismessina:
## Description

Follow-up to #28445 (the v2.4 hotfix that migrated data fetching to the official Product Hunt GraphQL API v2). That PR intentionally kept the old HTML scraper in the tree as a fallback while the API path was verified in production. This PR (v3.0) completes the migration by removing the now-dead scraper and tightening credential handling.

Note that the Product Hunt API is less rich than scraping, but their new Cloudflare defenses mean this extension will keep breaking. Migrating to the API increases complexity by requiring credentials, but gives us a way to maintain essential functionality.

**Removes the HTML scraper and dead code**
- Delete the HTML scraper, its test commands, and scraper-only helpers; drop the `cheerio` / `@types/cheerio` dependencies. All data now comes from the official API or the public Atom feed — no production code fetches Product Hunt HTML.
- Remove the dead `Shoutout` type and the scraper-only "Built With" UI.
- Unify logging on a single logger (`@chrismessina/raycast-logger`) with automatic credential redaction.

**Improves the credential / basic-feed experience**
- When only one of API Key / Secret is filled, fall back to the basic feed with a clear "Missing credentials" notice instead of a blocking error.
- Add a **Reload Extension** action so updated API credentials take effect without quitting. (Raycast reads preferences once per command launch, so a plain refresh keeps using the keys from when the command opened.)
- Re-authenticate from scratch on refresh so corrected credentials are picked up instead of reusing a stale token.
- Align the basic-feed banner and the toast so both explain why the feed is showing and how to fix it.

No user-facing changes to the happy path; the AI tool (`get-latest-products`) output is unchanged and was verified against the live API.

## Screenshots

producthunt-1

producthunt-2

## Checklist

- [x] I ran `npm run build` and `npm run lint` and there are no errors
- [x] I ran `npm run lint --fix` to autofix any lint errors
- [x] I tested my changes on macOS
- [x] I checked that the extension does not collect any sensitive data
- [x] I updated the CHANGELOG using the `{PR_MERGE_DATE}` placeholder


0xdhrv (approved):
Looks good to me, approved ✅

PR #28477, opened 2026-06-01
Update visual-studio-code extension

### LINKED ISSUE #28311: [Visual Studio Code] Search recent projects
### Extension

### Raycast Version

1.104.18

### OS Version

iOS Tahoe 26.1

### Description

Search recent projects shows now project

### Steps To Reproduce

1. Open search recent projects
2. Search window is empty

Image

### Current Behaviour

_No response_

### Expected Behaviour

_No response_

ecntu:
## Description

Fixes `Search Recent Projects` returning an empty list on macOS with VS Code 1.118+.

VS Code 1.118 moved the recently-opened workspaces list out of `globalStorage/state.vscdb` and into a new shared storage database at `~/.vscode-shared/sharedStorage/state.vscdb`. The `sharedDataFolderName` (e.g. `.vscode-shared`) is a home-directory dotfolder on all platforms — it is not inside `~/Library/Application Support` on macOS or `%APPDATA%\\Roaming` on Windows.

The previous path construction had two mistakes on Mac (`Library/Application Support` prefix + a spurious `User/` segment) and one on Windows (the extra `User/` segment), so `fs.existsSync` always returned false and the extension silently fell through to the old `globalStorage` DB, which is now empty for this key on 1.118+.

Fix collapses both platform branches into a single line using `homedir()` directly, which is what VS Code itself does ([source](https://github.com/microsoft/vscode/blob/main/src/vs/platform/environment/common/environmentService.ts)). Backward compatible — on older VS Code where `~/.vscode-shared` doesn't exist, `existsSync` still falls through to the globalStorage path as before.

Fixes #28311

### DIFF
--- a/extensions/visual-studio-code-recent-projects/CHANGELOG.md
+++ b/extensions/visual-studio-code-recent-projects/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Visual Studio Code Changelog
 
-## [Fix: Search Recent Projects empty on macOS] - {PR_MERGE_DATE}
+## [Fix: Search Recent Projects empty on macOS] - 2026-06-02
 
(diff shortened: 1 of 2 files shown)


0xdhrv (approved):
Looks good to me, approved ✅

PR #28483, opened 2026-06-02
Add spark-mail extension

v-khanna:
## Description

Adds Spark Mail — browse, search, read, and triage your Spark inbox from Raycast via the official Spark CLI. Six commands: Inbox (smart views + load-more), Search Mail (keyword + semantic, with a thread reader), Calendar, Search Contacts, Browse Folders, and Compose. Write actions are auto-hidden for Read-only accounts since Spark gates triage behind Premium. Requires Spark Desktop running + the spark CLI. Not affiliated with Readdle; Spark is a trademark of its owner.

## Screencast

Screenshots are included in metadata/ (command list, Inbox, Search, Calendar, Folders, Compose).

## Checklist


v-khanna:
Thanks for the review! Addressed all three points in the latest push:

1. **`isSparkInstalled()` always returning `true`** — it now scans the directories on the CLI `PATH` for a `spark` binary (and validates the preference path when one is set) instead of assuming the bare `"spark"` fallback is installed. Users without the CLI now get the `SparkNotInstalled` setup screen rather than a raw ENOENT toast.
2. **Manual `Preferences` interface** — removed; now using the auto-generated `Preferences` type from `raycast-env.d.ts`.
3. **`splitAddresses` splitting on whitespace** — now splits on commas only, so friendly-formatted addresses like `Alice Smith <alice@example.com>` are preserved as a single recipient.

`npm run build` and `npm run lint` both pass.

v-khanna:
Addressed both issues from the latest review:

1. **Rules of Hooks** — each command (`inbox`, `search`, `calendar`, `contacts`, `folders`, `compose`) is now split into a thin guard component that performs the `isSparkInstalled()` check (no hooks) and an inner `*View` component that calls all hooks unconditionally. No hook runs after a conditional early return.
2. **`EmailList` arg serialization** — the `usePromise` key now uses `JSON.stringify`/`JSON.parse` instead of joining/splitting on a separator, so folder IDs containing spaces (e.g. `account:Sent Messages`) are preserved as a single argument. (The old separator was also an errant NUL byte rather than a space, so this removes that too.)

`npm run build` and `npm run lint` both pass.

v-khanna:
Thanks — addressed both points from the latest pass:

1. **`sparkPath` default neutralising the PATH fallback** — removed the hardcoded `"default": "/usr/local/bin/spark"` (kept it as a `placeholder` hint instead). With no default, an unset preference is empty, so `resolveSparkOnPath()` runs and scans `/usr/local/bin` **and** `/opt/homebrew/bin`. Apple Silicon users with the CLI at `/opt/homebrew/bin/spark` are now detected out of the box.
2. **Unresolved attachment-download toast** — the animated toast handle is now captured and resolved to `Success`/`Failure` via an effect once the attachment-enabled refetch settles, and the generic thread-load toast is suppressed while a download is in flight, so the two no longer stack.

`npm run build` and `npm run lint` both pass.

pernielsentikaer (changes requested):
Could you add at least one [screenshot](https://developers.raycast.com/basics/prepare-an-extension-for-store#screenshots) of the extension and add them to a folder called `metadata` - you need to use our tool to get the right padding on the images

---

PR #28524, opened 2026-06-03
Add create-md-from extension

juavazmor:
## Description

**Create Markdown from** turns content into a `.md` file without leaving Raycast. Three commands, each opening a quick form (pre-filled, focused name + destination folder) so the common path is just *invoke → save*:

- **From Selection** — saves the currently selected text.
- **From Clipboard** — smart by intent: a URL is fetched and converted to Markdown (Readability + Turndown); a rich (HTML) copy keeps its images and formatting; otherwise the plain text is saved verbatim.
- **From Current Tab** — saves the active browser tab as Markdown via the Browser Extension's reader mode. Requires the [Raycast Browser Extension](https://www.raycast.com/browser-extension).

Other details:
- Destination folder is a configurable preference, and the last folder you used is remembered for next time.
- Saving never overwrites — on a name clash it appends ` 2`, ` 3`, … (Finder-style).
- File names are derived from the page title / first line of text, sanitized and trimmed; the `.md` extension is added automatically (any typed `.md`/`.markdown` is collapsed, so you never get `name.md.md`).

No external services and no API keys required.

## Screencast

Screenshots of the three commands (the save form for each source) are included in `metadata/`.

## Checklist


juavazmor:
Thanks for the review! Addressed 2 of the 3 points:

- ✅ **Manual `Preferences` interface** — removed the hand-written interface in `src/lib/save.ts`; now using the auto-generated global `Preferences` type from `raycast-env.d.ts` (and added that file to `tsconfig.json` `include` so it resolves).
- ✅ **TOCTOU in the dedup loop** — replaced the `existsSync` + `writeFile` loop with an exclusive-create write (`writeFile(..., { flag: "wx" })`) that retries on `EEXIST`. The collision check and the write are now atomic, so a concurrently-created file can never be clobbered.
- ↩️ **Title case `from` → `From`** — keeping it lowercase. Raycast's own `ray lint` explicitly expects `"Create Markdown from"` and flags `"From"` as a title-case warning, so I'm matching the official linter rather than the generic English title-case rule here.

juavazmor:
Fixed in 89bcd9e. Confirmed empirically with the project's deps (linkedom + @mozilla/readability): `parseHTML(html)` leaves `document.baseURI` null, so relative `src`/`href` stayed relative. Injecting `<base href={url}>` into `document.head` before `Readability.parse()` makes linkedom derive `baseURI` from it, and Readability now resolves relative URLs to absolute (verified `/x.png` → `https://example.com/x.png` and `../rel.png` → `https://example.com/blog/rel.png`).

0xdhrv (commented):
Thanks for the contribution 🔥 

This seems to overlap quite closely with the existing [Paste to Markdown](https://www.raycast.com/weicodes/paste-to-markdown) extension. While this PR adds some useful differences, especially creating a `.md` file directly and supporting sources like selection, clipboard URLs, and the current browser tab, the core domain is still converting copied/web content into Markdown.

Rather than adding a separate Store extension, I think this would be better as an improvement to the existing extension. A good path forward could be to add:

* Output options, for example: paste into the active app, copy Markdown to clipboard, or save/create a Markdown file
* Additional input sources, for example: selected text, clipboard URL, rich clipboard content, and current browser tab
* Any missing source handling from this PR as key differentiators inside the existing extension

PR #28532, opened 2026-06-03
Improve GIF conversion in yafw (fix palette, add quality/fps/speed/loop)

stefan-imbesi:
## Description

This improves GIF support in **yafw** and fixes a bug where the two‑pass palette was generated but never applied.

Context: this comes out of the discussion on #26652, where I proposed a standalone `video2gif` extension. @0xdhrv suggested consolidating GIF functionality into the existing **yafw** extension instead, and [gave the go‑ahead to open this PR](https://github.com/raycast/extensions/pull/26652#issuecomment-4609750963). So this is that PR — bringing the GIF features into yafw, implemented in yafw's own style. cc @pablopunk @0xdhrv

### The bug
`src/objects/ffmpeg.gif.ts` ran `palettegen` in pass 1 to write `palette.png`, but **pass 2 never referenced it** (it only applied `scale`), then deleted the palette in the `finally` block. So the optimized palette was generated and thrown away, making conversion effectively single‑pass — visible color banding / lower quality. The frame rate was also hardcoded to 30fps (with a `@TODO`).

```
# before (pass 2 — palette ignored):
ffmpeg -i input -vf scale=W:-2 output.gif

# after (pass 2 — palette applied):
ffmpeg -i input -i palette.png -lavfi "[0:v] fps=F,scale=W:-2:flags=lanczos [x]; [x][1:v] paletteuse=dither=sierra2_4a" -loop 0 output.gif
```

### What changed
- **Fixed the two‑pass palette** — pass 2 now applies the palette via `paletteuse` (passed as a second input). This is the core fix.
- **Made GIF output configurable via `Convert To` command preferences**, keeping yafw's zero‑config ethos (sensible defaults, no new UI surface):
  - **GIF Quality** (1–100 → number of palette colors)
  - **GIF Frame Rate** (replaces the hardcoded 30fps)
  - **GIF Speed** (0.25×–2×)
  - **GIF Loop** (forever / once)
- Resizing a GIF now uses **Lanczos** scaling for sharper output. The `Resize` command continues to preserve the source frame rate (it doesn't pass an fps), and now also benefits from the palette fix.

**One default to flag:** the frame‑rate default is **15fps** (previously hardcoded to 30). 15 is the common GIF default and produces noticeably smaller files — but happy to set it back to 30 if you'd prefer to keep the prior behavior.

### Testing
- `npm run build` and `npm run lint` both pass.
- I verified the exact ffmpeg commands the code builds (including yafw's `zsh -l -c` wrapper and `sanitizeFileName`) against sample videos, with **spaces and parentheses in the paths**:
  - the palette is applied (`paletteuse` present), and quality/fps/speed/loop are all reflected in the output (confirmed frame counts, dimensions, and loop flag via `ffprobe`);
  - the `Resize → GIF` path preserves the source frame rate;
  - the temporary `palette.png` is cleaned up afterward.

### Proposed follow‑up: live size estimation
In #26652 I also had a live "estimated output size" that updated as you changed settings. That needs a small form (view command), which doesn't fit yafw's zero‑config `Convert To` flow, so I left it out here to keep this PR focused on the bug fix + controls. **If you'd welcome it, I'm happy to add a lightweight GIF form in a follow‑up PR** — I already have a working size heuristic from video2gif to port. Let me know your preference given the "zero config" philosophy.

## Checklist

### DIFF
--- a/extensions/yafw/CHANGELOG.md
+++ b/extensions/yafw/CHANGELOG.md
@@ -1,5 +1,13 @@
 # YAFW Changelog
 
+## [Improve GIF conversion] - {PR_MERGE_DATE}
+
+- Fix GIF conversion two-pass palette: the generated palette is now applied with `paletteuse`, producing noticeably better colors and less banding (previously the palette was generated and discarded, so output was effectively single-pass)
+- Add GIF quality preference (1-100) controlling the number of palette colors
+- Make GIF frame rate configurable via a preference (previously hardcoded to 30fps), defaulting to 15fps for smaller files
+- Add GIF speed and loop (forever/once) preferences
+- Use Lanczos scaling when resizing GIFs for sharper output
+
 ## [Add video trim command] - 2026-01-28
 
 - Add new "Trim" command to trim videos by specifying start time and either end time or duration
--- a/extensions/yafw/CHANGELOG.md
+++ b/extensions/yafw/CHANGELOG.md
@@ -1,6 +1,6 @@
 # YAFW Changelog
 
-## [Improve GIF conversion] - {PR_MERGE_DATE}
+## [Improve GIF conversion] - 2026-06-04
 
 - Fix GIF conversion two-pass palette: the generated palette is now applied with `paletteuse`, producing noticeably better colors and less banding (previously the palette was generated and discarded, so output was effectively single-pass)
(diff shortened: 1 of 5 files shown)


0xdhrv (approved):
Looks good to me, approved ✅

PR #28540, opened 2026-06-04
[blip-raycast] Add Windows cross-platform support

brunocbreis:
## Summary

- Adds OS detection via `process.platform` — no Raycast API exists for this, so Node's built-in is the right approach
- On **Windows**, gets the selected File Explorer item via PowerShell `Shell.Application` COM automation, since `getSelectedFinderItems()` is macOS-only with no Windows equivalent in the Raycast API
- On **Windows**, triggers Blip by invoking its registered shell context menu verb via PowerShell — the direct equivalent of how macOS Services work
- Uses `runPowerShellScript` from `@raycast/utils` for both PowerShell calls (the Raycast-idiomatic approach)
- All "Finder" UI strings are now platform-aware: "Finder" on macOS, "File Explorer" on Windows
- `package.json`: added `"Windows"` to `platforms`, updated command title and description to be generic

## Test plan

- [ ] macOS: both commands still work as before (no behaviour change on macOS)
- [ ] Windows: "Send to Blip" pre-fills with active File Explorer selection
- [ ] Windows: "Send Selected File to Blip" sends the selected Explorer item directly
- [ ] Windows: UI description text shows "File Explorer" instead of "Finder"
- [ ] Windows: clear error shown if Blip's shell extension is not installed

### DIFF
--- a/extensions/blip-raycast/CHANGELOG.md
+++ b/extensions/blip-raycast/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Blip Changelog
 
+## [Added Windows support] - 2026-06-04
+
+- Added cross-platform support for Windows via PowerShell automation
+- File Explorer selection detection on Windows using Shell.Application COM automation
+- Blip invocation on Windows via registered shell context menu verb
+- Platform-aware UI strings: "Finder" on macOS, "File Explorer" on Windows
+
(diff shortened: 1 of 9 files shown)


pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #28553, opened 2026-06-04
fix(linear): move Set Priority shortcut off the Copy Issue ID slot

### LINKED ISSUE #28523: [Linear] Same shortcut for "copy issue ID" and "set priority"
### Extension

https://www.raycast.com/linear/linear

### Raycast Version

0.62.2.0

### OS Version

_No response_

### Description

From an issue list or an issue itself, `⌘ .` is the shortcut for both "copy issue ID" and "set priority" (defaults to "set priority")

Image

Image

### Steps To Reproduce

1. open Linear extension "my issues"
2. highlight a ticket
3. hit the shortcut `⌘ .`

### Current Behaviour

opens the "set priority" dialog

### Expected Behaviour

copy the issue ID (previous extension behavior)

spokodev:
Closes #28523.

`Copy Issue ID` and `Set Priority` were both bound to `⌘ .` on macOS, because `Set Priority` used `Keyboard.Shortcut.Common.Pin` which resolves to the same shortcut as the explicit `cmd+.` on `Copy Issue ID`. Hitting `⌘ .` from an issue list opened the Set Priority submenu instead of copying the identifier, which the reporter notes was the previous behaviour.

`Set Priority` now uses `⌘ ⌥ P` on macOS and `Ctrl Alt P` on Windows. `⌥ + P` is not bound anywhere else in the Linear extension (the only other `p` shortcuts use the `Ctrl Shift` modifier in `CreateIssueForm` and `ProjectSubmenu`), and `Copy Issue ID` gets its original slot back.

### Verification

- `ray build` clean
- `ray lint` clean (ESLint + Prettier + package.json + icon + metadata validators)
- Grep across `extensions/linear/src/` confirms no other action uses `cmd+opt+p` or `ctrl+alt+p`
- Spot-checked the action panel manually: `Copy Issue ID` now responds to `⌘ .`; `Set Priority` responds to `⌘ ⌥ P`; no other shortcut is affected


0xdhrv (approved):
Looks good to me, approved ✅

PR #28555, opened 2026-06-04
[Things] Optional database reading and expanded AI tools

janosorcsik:
## Description

This PR significantly expands the AI tool surface and adds an optional mode that reads lists directly from the Things SQLite database instead of using Apple Events (JXA). It also fixes a date-setting crash and recurring-deadline handling.

## New Features

- **Expanded AI tools** — Added `get-todos` (filter by list/project/area), `search-todos` (keyword search across title and notes), `get-todo-details`, `get-todos-details` (batch fetch), `get-project-details`, `get-area-details`, and `add-json` (create projects with headings and nested to-dos in a single call).
- **Optional "Unofficial API" mode** — A new preference reads `inbox`, `today`, `anytime`, `upcoming`, `someday`, `logbook`, and `trash` directly from the Things SQLite database via `executeSQL` (`@raycast/utils`) instead of Apple Events. It requires Full Disk Access and is off by default; the standard JXA path remains the default.

## Bug Fixes

- Fix `set-todo-property` crashing Things when setting date fields (`dueDate`, `activationDate`, `completionDate`, `cancellationDate`) — values are now passed as JS `Date` objects via JXA instead of plain strings, using a local-time constructor to avoid an off-by-one in negative-offset timezones.
- When the Unofficial API is enabled, recurring tasks now show the correct upcoming deadline instead of a placeholder value, and a `dueDateIsRecurring` flag distinguishes a fixed deadline from one that shifts with each recurrence.

## Code Quality

- Split the API into a thin facade (`api.ts`) over separate SQL (`api-sql.ts`) and JXA (`api-jxa.ts`) implementations, selected by the `useUnofficialApi` preference.
- Remove the `get-list-todos` AI tool (superseded by `get-todos`); the `getListTodos` API function is retained for the UI views.
- Add flat AI-friendly types (`TodoSummary`, `TodoDetails`, `ProjectDetails`, `AreaDetails`, `ChecklistItem`).
- Update AI instructions and tool descriptions (search-first, batch-details, write-operation routing).

## Packages

- Update dependencies

## Checklist


bendrucker:
> Logbook and Trash still use JXA as a fallback.

These are generally going to be the most expensive lists to read so I was inclined to ask why until I found the code comment claiming it's required. That doesn't seem right, why would that data not be in the same DB?

> SQLite failure falls back to JXA automatically

Would like to see an explicit setting for DB file access, since it means:

* Granting full disk access
* Using an unofficial API

Given the doubling of `api.ts` it would also be nice to split up the SQL and JXA clients, and maybe a third types file if they need to share. Would help clarify at the callsite given conditional use/fallback (e.g., `sql.getTodos` vs. `jxa.getTodos`).

janosorcsik:
Hi @bendrucker,
thanks for the feedback.
I added a preference to use the unofficial API for reading; the default is to use JXA. Everything has an SQL variant now.

bendrucker:
Thanks! Will try to do a pass over the code if I have time but those were the only two functional notes I had.

extensions/things/src/components/TodoList.tsx
```diff
@@ -21,7 +21,7 @@ const ALL_TAGS = 'all';
 const NO_TAG = 'none';
 
 function parseTags(tagString: string | null | undefined): string[] {
-  return tagString?.split(', ').filter(Boolean) ?? [];
+  return tagString?.split(',').filter(Boolean) ?? [];
```
bendrucker: `parseTags` splits on `','`, but on the default Apple Events path tags come from `tagNames()`, which joins with `', '` (comma-space). I confirmed it against the live app: a to-do tagged shopping and travel returns `"shopping, travel"`. So this yields `['shopping', ' travel']`, and since the tag dropdown is built from clean `tag.name()` values, `expandedTags.has(' travel')` is `false` and the to-do drops out of the filtered view. The `No Tag` filter and the applicable-tags set break the same way. `TodoDetail.tsx:42` has the same change. The SQLite path joins with `','`, so the two need different handling. Splitting on `/,\s*/` covers both.

extensions/things/src/api.ts
```diff
@@ -214,21 +131,34 @@ export const getProjectName = (projectId: string) =>
...
+    const [y, m, d] = value.split('-').map(Number);
+    valueExpr = `new Date(${y}, ${m - 1}, ${d})`;
+  } else {
+    valueExpr = `'${value}'`;
+  }
+  return executeJxa(
     `
   const things = Application('${preferences.thingsAppIdentifier}');
-  things.toDos.byId('${todoId}').${key} = '${value}';
+  things.toDos.byId('${todoId}').${key} = ${valueExpr};
```
bendrucker: `getTodoName` and `getProjectName` use `escapeJxa`, but `setTodoProperty` interpolates `todoId`, `key`, and `value` straight into the script. A `value` with an apostrophe breaks it: `set-todo-property(key: 'notes', value: "It's done")` emits `... .notes = 'It's done';` and the write fails. Separately, the tool exposes `key` as a free-form string with no allowlist, and it lands as `.${key} =` in the source, which is an injection surface driven by model output. Escaping `value` and constraining `key` would close both. `deleteTodo` and `deleteProject` interpolate the ID unescaped too.

extensions/things/src/api-jxa.ts
```diff
@@ -0,0 +1,458 @@
...
+      areaName: areaRef ? areaRef.name() : undefined,
+      areaId: areaRef ? areaRef.id() : undefined,
+    };
+  });
+`,
+    `Query todos`,
+  );
+}
+
+export async function queryTodoDetailsJxa(appId: string, todoId: string): Promise<TodoDetails | null> {
```
bendrucker, line 241: `queryTodoDetailsJxa` (and the project/area variants) call `.properties()` on `byId(badId)`, which throws. The SQLite path returns `null`, and the tools depend on that with `if (!details) throw 'not found'`. On this path that branch is dead, and an unknown or trashed ID surfaces "Things automation interface error" instead of "To-do with ID X not found." Returning `null` on a failed lookup would align the two paths.

extensions/things/src/api-jxa.ts
```diff
@@ -0,0 +1,458 @@
...
+
+export async function searchTodosJxa(appId: string, query: string): Promise<TodoSummary[]> {
+  const escaped = query.replace(/\\/g, '\\\\').replace(/'/g, "\\'");
+  return executeJxa(
+    `
+  const things = Application('${appId}');
+  const q = '${escaped}'.toLowerCase();
+  const todos = things.toDos().filter(todo => {
+    const props = todo.properties();
+    return props.status === 'open' &&
```
bendrucker, line 319: `searchTodosJxa` filters on `status === 'open'` with no trashed check, so a to-do in the Trash (still `open`) is returned. The SQLite path excludes `trashed = 0`. The search-first flow then hands the model an ID for an item it should treat as gone.

extensions/things/src/api-jxa.ts
```diff
@@ -0,0 +1,458 @@
...
+  ${opts.projectId ? `const todos = things.projects.byId('${escapeJxa(opts.projectId)}').toDos();` : opts.areaId ? `const todos = things.areas.byId('${escapeJxa(opts.areaId)}').toDos();` : opts.listName ? `const todos = things.lists.byId('${commandListNameToListIdMapping[opts.listName as CommandListName] ?? opts.listName}').toDos();` : `const todos = things.toDos();`}
+  return todos.map(todo => {
+    const props = todo.properties();
+    const projectRef = props.project;
+    const areaRef = props.area;
+    return {
+      id: props.id,
+      name: props.name,
+      dueDate: props.dueDate ? props.dueDate.toISOString().slice(0,10) : undefined,
+      dueDateIsRecurring: false,
```
bendrucker, line 227: Every JXA query path sets `isRecurring` and `dueDateIsRecurring` to `false`. The AI instructions say "when `dueDateIsRecurring` is true, the due date shifts ... do NOT treat it as a fixed deadline," so on the default path the model treats recurring due dates as fixed. If that's an accepted limitation here, the instructions could note the field is only meaningful with the database preference on.

extensions/things/src/api-sql.ts
```diff
@@ -0,0 +1,992 @@
...
+  date.setDate(date.getDate() + days);
+  return encodeThingsDate(date.getFullYear(), date.getMonth() + 1, date.getDate());
+}
+
+/** Parse the recurrence offset (in days) from a Things plist XML recurrence rule. */
+function parseDeadlineOffset(plistXml: unknown): number | null {
+  if (!plistXml || typeof plistXml !== 'string') return null;
+  const match = plistXml.match(/<key>ts<\/key>\s*<integer>(-?\d+)<\/integer>/);
+  if (!match) return null;
+  return Math.abs(parseInt(match[1], 10));
```
bendrucker: `parseDeadlineOffset` does `Math.abs(parseInt(match[1], 10))` and feeds it to `addDaysToThingsDate(nextInstanceStartDate, offset)`. A deadline that falls before the occurrence stores a negative `ts`, and `Math.abs` flips it, so the deadline computes after the start instead of before. Worth checking against a recurring to-do whose deadline precedes its start.

extensions/things/src/api-sql.ts
```diff
@@ -0,0 +1,992 @@
...
+      areaName: string | null;
+      areaTags: string | null;
+    };
+    const projectRows = await executeSQL<ProjectRow>(
+      getThingsDBPath(),
+      `SELECT p.uuid as id, p.title as name,
+        CASE p.status WHEN 2 THEN 'canceled' WHEN 3 THEN 'completed' ELSE 'open' END as status,
+        COALESCE(p.notes, '') as notes,
+        (SELECT GROUP_CONCAT(tg.title, ',') FROM TMTaskTag tt JOIN TMTag tg ON tg.uuid = tt.tags WHERE tt.tasks = p.uuid) as tags,
+        NULL as dueDate, NULL as activationDate, -- dates not fetched in collection summary; use queryProjectDetailsSQL for full data
```
bendrucker: `getCollectionsFromDB` selects `NULL as dueDate, NULL as activationDate` for projects and area to-dos, so `get-projects` reports no dates on this path while the Apple Events version returns them. Either fetch them here or drop the columns and note that callers needing dates use `get-project-details`.

extensions/things/src/api-sql.ts
```diff
@@ -0,0 +1,992 @@
...
+  return _thingsDBPath;
+}
+
+// Things stores dates as packed Int64: (year << 16) | (month << 12) | (day << 7)
+const YEAR_SHIFT = 16;
+const MONTH_SHIFT = 12;
+const DAY_SHIFT = 7;
+const MONTH_MASK = 0xf;
+const DAY_MASK = 0x1f;
+const RECURRING_DEADLINE_PLACEHOLDER = 262213760;
```
bendrucker: The packed-date layout, these placeholder sentinels, and the status/type/start codes are undocumented Things internals, and `convertThingsDate`'s range check accepts any plausible integer, so a layout change would return a valid-looking wrong date rather than an error.

extensions/things/src/api-sql.ts
```diff
@@ -0,0 +1,992 @@
...
+  AreaDetails,
+  ChecklistItem,
+} from './types';
+
+// Things stores its data in a SQLite database. Tries known locations in order:
+//   1. ThingsData-*/Things Database.thingsdatabase/main.sqlite
+//   2. Things Database.thingsdatabase/main.sqlite
+//   3. ~/Library/Containers/com.culturedcode.ThingsMac/.../Things.sqlite3
+//   4. Things Database.thingsSQLite
+function findThingsDBPath(): string {
```
bendrucker, line 31: `findThingsDBPath` returns the legacy fallback without checking it exists, and `getThingsDBPath` caches whatever it returns for the process lifetime. If no known location matches, or Things rotates its `ThingsData-*` directory mid-session, queries then hit a missing or stale file. With no error handling on this path, that shows up as a raw SQLite/`ENOENT` error rather than the Full Disk Access guidance the Apple Events path gives. Validating existence before caching and wrapping failures in a typed error would help.

extensions/things/src/api-sql.ts
```diff
@@ -0,0 +1,992 @@
...
+  return { effectiveDeadline: null, effectiveStartDate, dueDateIsRecurring: false };
+}
+
+const TODO_JOINS = `
+  FROM TMTask t
+  LEFT JOIN TMTask p ON t.project = p.uuid
+  LEFT JOIN TMArea pa ON p.area = pa.uuid
+  LEFT JOIN TMArea a ON t.area = a.uuid`;
+
+const TODO_SELECT_SUMMARY = `
```
bendrucker, line 132: A few patterns repeat enough that a schema change touches many sites: the status-to-string mapping (8 places), the tag `GROUP_CONCAT` subquery (6) and its area variant (4), the overlap between `TODO_SELECT_SUMMARY` / `TODO_SELECT_DETAIL` / `LIST_SELECT`, the three converters (`rowToTodoSummary`, `todoToSummary`, `mapListTodoRow`), and the near-identical blocks in `getAnytimeTodosFromDB` and the two branches of `getCollectionsFromDB`. Helpers like `mapStatus`, `tagListSubquery(alias)`, and a shared select fragment would shrink the surface a Things change can break.

extensions/things/src/api-sql.ts
```diff
@@ -0,0 +1,992 @@
...
+
+export type CollectionMap = {
+  tags: string[];
+  tagsWithHierarchy: TagWithParent[];
+  projects: Project[];
+  areas: Area[];
+  lists: List[];
+};
+
+export async function getCollectionsFromDB<K extends keyof CollectionMap>(
```
bendrucker, line 789: `getCollectionsFromDB` awaits the project, project-todos, area, and area-todos queries sequentially with no data dependency between them, so `Promise.all` would collapse four round-trips into one wait. `queryTodoDetailsSQL` and `queryTodosDetailsSQL` have the same pattern with their checklist query.

extensions/things/src/api-sql.ts
```diff
@@ -0,0 +1,992 @@
...
+
+export type QuickFindData = {
+  areas: Array<{ id: string; name: string }>;
+  projects: Array<{ id: string; name: string; areaName?: string }>;
+  todos: Array<{ id: string; name: string; status: string; projectName?: string; areaName?: string }>;
+};
+
+// Read directly from Things' SQLite database — a single SQL query with JOINs
+// replaces many serialized Apple Events.
+export const getQuickFindDataFromDB = async (): Promise<QuickFindData> => {
```
bendrucker, line 920: `getQuickFindDataFromDB` builds a large `json_object(...)` in SQL, `JSON.parse`s it, then filters for nulls `json_group_array` can't emit and remaps null to undefined. `executeSQL` already returns parsed rows, so three plain queries with inline `?? undefined` would drop the JSON builder, the parse, and the dead filters.

extensions/things/src/api-jxa.ts
```diff
@@ -0,0 +1,458 @@
...
+  opts: {
+    listName?: string | null;
+    projectId?: string | null;
+    areaId?: string | null;
+  } = {},
+): Promise<TodoSummary[]> {
+  return executeJxa(
+    `
+  const things = Application('${appId}');
+  ${opts.projectId ? `const todos = things.projects.byId('${escapeJxa(opts.projectId)}').toDos();` : opts.areaId ? `const todos = things.areas.byId('${escapeJxa(opts.areaId)}').toDos();` : opts.listName ? `const todos = things.lists.byId('${commandListNameToListIdMapping[opts.listName as CommandListName] ?? opts.listName}').toDos();` : `const todos = things.toDos();`}
```
bendrucker: The unfiltered, `projectId`, `areaId`, and `listName` branches call `.toDos()` with no status or trashed filter, so completed, canceled, and trashed items come back. The SQLite equivalent applies `status = 0 AND trashed = 0`, and `get-todos` advertises open to-dos. The no-filter branch also walks the whole database including the logbook.

bendrucker (changes requested):

bendrucker:
Found a handful of bugs, reported inline. It's possible that a some of the JXA ones were preexisting.

In reviewing the SQL, I found some of the internals reverse engineering that generally concerns me about this approach, most notably the packed-date layout, the placeholder sentinels, the rt1_* columns, and the recurring-master logic. Ideas for mitigation:

1. Check the schema version and assert early that the code is working with the expected database shape. If there's no explicit version in the database, fingerprinting the schema might be a workaround option.
2. DRY out `api-sql.ts` such that reverse engineering is only done once

extensions/things/src/components/TodoList.tsx
janosorcsik: Split with comma-space, and SQL queries also use comma-space to concatenate tags. The last change also fixed how the tags were shown in the todo list.
dyn-b8a9dcc2a742ebfa8618a6bbbf3f02fb

extensions/things/src/api.ts
janosorcsik: Added escaping.

extensions/things/src/api-jxa.ts
janosorcsik: Used `whose` in JXA to check whether an item exists.

extensions/things/src/api-jxa.ts
janosorcsik: `things.todos()` already excludes items from Trash.

extensions/things/src/api-jxa.ts
janosorcsik: `things.todos()` already excludes items from Trash.

extensions/things/src/api-jxa.ts
janosorcsik: Updated instructions for AI about recurring fields.

extensions/things/src/api-sql.ts
janosorcsik: Added `Promise.all` at all three places.

extensions/things/src/api-sql.ts
janosorcsik: Removed `Math.abs` from `parseDeadlineOffset`. This case can't happen in practice, but the sign was still wrong.

pernielsentikaer (commented):
Thanks for the update @janosorcsik 

@loris do you want to check this?

extensions/things/src/api-sql.ts
janosorcsik: Added `assertPackedDateEncoding` to verify Things SQL date calculation.

extensions/things/src/api-sql.ts
janosorcsik: `getQuickFindDataFromDB` is simplified

extensions/things/src/api-sql.ts
janosorcsik: Added the missing fields.

extensions/things/src/api-sql.ts
janosorcsik: Added helpers to use them in SQL queries.

extensions/things/src/api-sql.ts
janosorcsik: Added an error message if the Things database doesn’t exist.

janosorcsik:
Thanks for the detailed review. I've fixed everything, created a `things-internals.ts` file for the reverse-engineered database, and added a schema check to verify the database.

pernielsentikaer (commented):
@bendrucker can you have another look since you flagged some issues?

bendrucker:
Hi yes haven't had a chance to look again, putting in a reminder for Sunday

extensions/things/src/api.ts
```diff
@@ -214,27 +131,40 @@ export const getProjectName = (projectId: string) =>
...
+    const [y, m, d] = value.split('-').map(Number);
+    valueExpr = `new Date(${y}, ${m - 1}, ${d})`;
+  } else {
+    valueExpr = `'${escapeJxa(value)}'`;
+  }
+  return executeJxa(
     `
   const things = Application('${preferences.thingsAppIdentifier}');
-  things.toDos.byId('${todoId}').${key} = '${value}';
+  things.toDos.byId('${escapeJxa(todoId)}').${key} = ${valueExpr};
```
bendrucker, line 161: The escaping added for `value` and `todoId` covers part of the earlier thread, but `key` is still interpolated raw here as `.${key} = ...` and arrives straight from model output via `tools/set-todo-property.ts` with no allowlist. An unsupported or malformed property name produces broken JXA, and because `key` is neither escaped nor constrained it is also an injection surface. An allowlist of writable Things properties would close it. This runs on the default path regardless of the database preference.

extensions/things/src/api-jxa.ts
```diff
@@ -0,0 +1,466 @@
...
+  return result as TodoDetails;
+}
+
+export async function queryTodosDetailsJxa(appId: string, todoIds: string[]): Promise<TodoDetails[]> {
+  if (!todoIds.length) return [];
+  const idList = todoIds.map((id) => `'${escapeJxa(id)}'`).join(', ');
+  const results: TodoDetails[] = await executeJxa(
+    `
+  const things = Application('${appId}');
+  const ids = [${idList}];
```
bendrucker, line 280: The `whose({id})` guard added to the single-item detail functions is right, but this batch path still uses `things.toDos.byId(id).properties()`, so one unknown or trashed id throws and the whole batch fails. The SQL equivalent returns the found subset. The same `whose` guard per id, or filtering before mapping, would align them.

extensions/things/src/things-internals.ts
```diff
@@ -0,0 +1,192 @@
...
+  const date = new Date(y, m - 1, d);
+  date.setDate(date.getDate() + days);
+  return encodeThingsDate(date.getFullYear(), date.getMonth() + 1, date.getDate());
+}
+
+/**
+ * Assert that the packed-date codec round-trips correctly.
+ * Call this once at startup to detect if Things has changed its date encoding.
+ */
+export function assertPackedDateEncoding(): void {
```
bendrucker, line 119: `assertPackedDateEncoding` round-trips `encodeThingsDate` through `convertThingsDate` using the same in-file constants, so it passes regardless of what Things actually stores. It cannot detect the layout change it was added to catch. A guard that compares against a known-good value read from the real database, or that asserts the `REQUIRED_SCHEMA` version, would catch drift.

extensions/things/src/api-jxa.ts
```diff
@@ -0,0 +1,466 @@
...
+  inbox: 'TMInboxListSource',
+  today: 'TMTodayListSource',
+  anytime: 'TMNextListSource',
+  upcoming: 'TMCalendarListSource',
+  someday: 'TMSomedayListSource',
+  logbook: 'TMLogbookListSource',
+  trash: 'TMTrashListSource',
+};
+
+export function escapeJxa(value: string): string {
```
bendrucker, line 96: `escapeJxa` escapes single quotes but not backslashes. A value ending in `\` (a note or a path) escapes the closing quote and breaks the script. `searchTodosJxa` already escapes backslashes first. Worth reusing that here, since `setTodoProperty` and the delete functions all depend on this helper.

extensions/things/src/api-jxa.ts
```diff
@@ -0,0 +1,466 @@
...
+  return todos.map(todo => {
+    const props = todo.properties();
+    const projectRef = props.project;
+    const areaRef = props.area;
+    const s = props.status;
+    return {
+      id: props.id,
+      name: props.name,
+      status: s !== 'open' ? s : undefined,
+      dueDate: props.dueDate ? props.dueDate.toISOString().slice(0,10) : undefined,
```
bendrucker: Things returns date-only fields at local midnight. `toISOString().slice(0,10)` converts to UTC first, so for any UTC+X user this yields the previous calendar day. This is the inverse of the write-path off-by-one fixed in `setTodoProperty` (see the local-time `Date` note in `api.ts`). The summary mappers elsewhere format the same fields as a full ISO timestamp, so the path is also inconsistent with itself.

extensions/things/src/api-jxa.ts
```diff
@@ -0,0 +1,466 @@
...
+    tags: (todo.tagNames() || '').split(', ').filter(t => t),
+    dueDate: props.dueDate ? props.dueDate.toISOString().slice(0,10) : undefined,
+    dueDateIsRecurring: false,
+    activationDate: props.activationDate ? props.activationDate.toISOString().slice(0,10) : undefined,
+    isRecurring: false,
+    projectName: projectRef ? projectRef.name() : undefined,
+    projectId: projectRef ? projectRef.id() : undefined,
+    areaName: areaRef ? areaRef.name() : undefined,
+    areaId: areaRef ? areaRef.id() : undefined,
+    checklistItems: [],
```
bendrucker: Seems like this has potential to confuse an agent, versus a way to clearly show that the checklist is unavailable. Maybe could be `null` if unavailable at the tool level.

extensions/things/src/api-jxa.ts
```diff
@@ -0,0 +1,466 @@
...
+
+    // Some calls only update data and don't return anything
+    if (!result) {
+      return;
+    }
+
+    // JXA's non-human-readable output is similar to JSON, but is actually a JSON-like representation of the JavaScript object.
+    // While values should not be `undefined`, JXA will include {"key": undefined} in its output if they are.
+    // This is not valid JSON, so we replace those values with `null` to make it valid JSON.
+    return JSON.parse(result.replace(/:\s*undefined/g, ': null'));
```
bendrucker: This replacement runs over the whole serialized payload, so a name or note containing the literal `: undefined` gets rewritten to `: null`. Building the object without `undefined` values, or narrowing the match, would avoid corrupting user text.

extensions/things/src/api-jxa.ts
```diff
@@ -0,0 +1,466 @@
+import { runAppleScript } from '@raycast/utils';
+import { CommandListName, Todo, TodoSummary, TodoDetails, ProjectDetails, AreaDetails } from './types';
+import { CollectionMap, QuickFindData, organizeLists } from './api-sql';
```
bendrucker: Seems like maybe these should be in a module shared by SQL/JXA vs JXA importing SQL. CollectionMap and QuickFindData could just move to `types.ts`, `organizeLists` maybe to helpers.

extensions/things/src/api-sql.ts
```diff
@@ -0,0 +1,990 @@
...
+    ${LIST_SELECT}
+    WHERE
+      t.type IN (0, 1)
+      AND t.trashed = 1
+    ORDER BY t.userModificationDate DESC
+  `,
+  );
+}
+
+export async function getListTodosFromDB(commandListName: CommandListName): Promise<Todo[]> {
```
bendrucker, line 767: This switch has no default branch, so a `listName` outside the enum returns `undefined` and the following `.map` throws. The JXA path handles the same input gracefully. A default branch would keep parity. (Database path only.)

bendrucker (commented):
Found a few more things, some of the JXA stuff may be pre-existing

extensions/things/src/api.ts
janosorcsik: Key is verified to only include allowed values

extensions/things/src/api-jxa.ts
janosorcsik: Replaced `byId` with `whose`

extensions/things/src/api-jxa.ts
janosorcsik: Added backslash to `escapeJxa`

extensions/things/src/api-sql.ts
janosorcsik: Default branch added with an empty array.

extensions/things/src/api-jxa.ts
janosorcsik: Now it returns null for checklist items in JXA

extensions/things/src/things-internals.ts
janosorcsik: There isn't a fix to do in the database that we can rely on for verification.
We could create one through JXA to test the date encoding, but after deleting it, the todo ends up in the Trash. And it would be an expensive step to run on every command.
How did you, @bendrucker  have the schema version check in mind?

extensions/things/src/api-jxa.ts
janosorcsik: Date calculation is updated

extensions/things/src/api-jxa.ts
janosorcsik: Replaced the call with JSON.stringify to run inside JXA

extensions/things/src/api-jxa.ts
janosorcsik: Types are moved into `types.ts`

pernielsentikaer (commented):
Do you mind having another look @bendrucker

bendrucker:
Looking now

extensions/things/src/things-internals.ts
bendrucker: Checked `user_version`. Things leaves it at 0, so that's a dead end. But there's another anchor: the `Meta` table has a `databaseVersion` key (a plist-wrapped integer). It reads 27 on my stable install and 24 on an older one, so it does track the storage format.

You could assert against that instead of the round-trip:

```sql
SELECT value FROM Meta WHERE key = 'databaseVersion';
```

Parse the int and check it against a version you've verified the date layout on. One indexed lookup on a 3-row table, no throwaway todo needed. An exact-match check would trip on every harmless migration, so a supported set (or a floor) is probably the best we could do.

Having it would be more of a defensive step: if there are future breaking changes you could have code adapt to the database version and handle before/after.

bendrucker (approved):
Everything from the last round addressed, thanks.

One open thread left on `assertPackedDateEncoding`, non-blocking. I left a note there with a real `databaseVersion` you can assert against if you want a genuine drift check.

bendrucker (approved):
_Don't mind the duplicate submits..._

bendrucker (approved):
_Don't mind the duplicate submits..._

extensions/things/src/things-internals.ts
janosorcsik: I checked my database version against the latest stable Things (`3.22.11 (32211507)`), and it's 26. Yours is 27, so this isn't an exact number.

Future database changes are still caught by `assertDatabaseSchema`, therefore, I'll skip this change.

janosorcsik:
Hi @bendrucker,
Thanks for your thorough review!
@pernielsentikaer, the PR could be merged.

pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #28559, opened 2026-06-04
Add from extension

albertolezaun-afk:
## Description

Adds **From**, an extension for [From](https://getfrom.app) — an outliner that understands what you write (it classifies notes, tasks and events, dates them and files them under the right context).

The extension brings frictionless capture to Raycast:

- **Create in From** — write a thought and it lands in today's note as a note, task or event, with an optional date.
- **Search From** — full-text search across the whole vault; open any result in the Mac app (`from://` deep link) or on the web.
- **Open Today's Note** — jump straight to the daily note.

It talks to the From server with the user's personal API token (the same long-lived token used by From's Claude/MCP integration): `POST /mcp` (JSON-RPC tools `from_create_node`, `from_get_today_note`) and `GET /search/nodes`. The token and an optional self-host server URL are set via extension preferences.

## Screencast

_Screenshots will be added shortly._

## Checklist


0xdhrv (approved):
Looks good to me, approved ✅

albertolezaun-afk:
Thanks for the thorough review! Everything flagged is now addressed (some by @0xdhrv, the rest in the latest commit):

- **search.tsx — stuck spinner**: `setLoading(false)` is now called on the empty-query branch, so clearing the search bar mid-request no longer leaves the spinner visible.
- **create.tsx — "event" type dropped**: no longer applicable. Create was rewritten to delegate to From's capture engine via deep link, so it infers note/task/event (and the date) from the text — there's no type dropdown to drop.
- **from-client.ts — manual `Prefs` interface**: replaced with the auto-generated `Preferences` type. I also added `raycast-env.d.ts` to `tsconfig.json`'s `include` so the generated global type resolves.
- **package.json — unused `@raycast/utils`**: removed from dependencies.
- **Missing CHANGELOG**: added with the `{PR_MERGE_DATE}` placeholder.
- **Deep link on Windows**: `platforms` is now `["macOS"]`, matching the Mac-only `from://` capture scheme.

`ray build` and `ray lint` both pass with no errors or warnings.

PR #28612, opened 2026-06-07
Add google-directory-search extension

vict0rsch:
## Description

This extension is meant to search one's contact AND directory. In my company I often end-up needing the email addresses from colleagues (I fill forms, manage software etc.) and having this directory available quickly is life-saving.

## Checklist

Extra:

- [x] I disclaim that this extension is vibe-coded. I have used all available online resources to build it right from the start and adversarially reviewed it to be secure.


vict0rsch:
Looking at the failed test vs the actual image it looks like the test is misleading?

0xdhrv (commented):

We already have extensions in the Store that deal with related Google Contacts / Google Workspace workflows. Could we consider whether this functionality belongs in one of the existing extensions below instead of creating another one?

- [Google Contacts](https://www.raycast.com/aic/google-contacts) by @aicayzer
- [Google Workspace](https://www.raycast.com/raycast/google-workspace) by @raycast

This PR overlaps with Google Contacts around the personal contacts workflow: both use the Google People API, require a user-supplied Google OAuth Client ID, and expose a `search-contacts` view command for browsing/searching contacts.

That said, I can see an important differentiator here: this PR also supports searching the Google Workspace organization directory, which is different from general personal contacts.

For the Workspace directory part specifically, I think it would be better to explore adding that functionality to the existing **Google Workspace** extension instead. Since this is about discovering people in a Workspace organization, it feels more naturally grouped with the broader Google Workspace integration rather than living as a separate contacts-focused extension.

If there are unique features or workflows you’re aiming to add beyond personal contact search and Workspace directory lookup, we’d love to hear them. Otherwise, consolidating this into the existing Google Contacts or Google Workspace extension would help avoid duplication and improve discoverability.

> This would help avoid duplication and keep related functionality clearly positioned for users.

aicayzer:
Agree with @0xdhrv that a separate directory-search extension probably isn't the way to go. Happy to take a PR into Google Contacts for any personal-contacts updates here. Also open to another option — adding Workspace directory search as an additional command in Google Contacts itself, disabled by default unless someone opts in. Curious what you both think.

vict0rsch:
Thanks for the feedback! To be honest I started with workspace directory search and added contacts _a posteriori_ thinking it would be useful to have all the emails I typically need at hand. But I see your points. I'll open a PR to the Google Workspace extension instead as it feels like a more natural home. And if someone wants directory search AND their personal contacts, well there's an extension for that 😏

PR #28620, opened 2026-06-08
Update quarantine-manager extension

nurkamol:
## Description

Update to the existing **Quarantine Manager** extension.

This implements a community feature request from the original submission (#25892): support for checking **apps and folders**, not just single files. macOS app bundles in particular can contain many internal quarantined files that a top-level check would miss.

### What's new

- **Directory scanning in both commands.** Selecting a folder or `.app` now scans its contents for quarantined files:
  - **Apps (`.app`)** are scanned **recursively** (bundles are self-contained), surfacing every internal quarantined file plus the bundle's own flag.
  - **Plain folders** are scanned **one level deep** (immediate contents only) so large directories like `Downloads` stay responsive.
- **Check Quarantine Status** lists each quarantined item found inside an app/folder; you can drill into any of them for details.
- **Remove Quarantine** shows an aggregate summary and clears `com.apple.quarantine` from a whole bundle/folder at once via `xattr -dr`, after a confirmation showing how many items are affected. A recursive "Remove All Attributes" (`xattr -cr`) is also available.
- **Fix:** raw binary attributes (e.g. `com.apple.macl`, `com.apple.provenance`) are now displayed as hex instead of unreadable control characters.

Directory scanning batches `xattr` calls and uses `spawnSync`/`execFileSync` with array arguments (no shell-string interpolation of paths), consistent with the existing hardening.

Resolves the request in #25892.

## Screencast

Tested locally via `npm run dev` and a distribution build (`npm run build`) in Raycast:
- **Check Quarantine Status** on an `.app` → recursive list of internal quarantined files (and the bundle's own flag).
- **Check Quarantine Status** on `~/Downloads` → immediate quarantined children only.
- **Remove Quarantine** on an app/folder → count-confirmation, then recursive clear; re-scan shows clean.
- Single-file behavior unchanged; binary attributes render as hex.

## Checklist

### DIFF
--- a/extensions/quarantine-manager/.gitignore
+++ b/extensions/quarantine-manager/.gitignore
@@ -3,6 +3,7 @@ dist/
 .DS_Store
 
 .claude/
+CLAUDE.md
 
 # Raycast specific files
 .raycast-swift-build
--- a/extensions/quarantine-manager/CHANGELOG.md
+++ b/extensions/quarantine-manager/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Quarantine Manager
 
-## [Scan apps and folders] - {PR_MERGE_DATE}
+## [Scan apps and folders] - 2026-06-08
 
 ### Added
 
--- a/extensions/quarantine-manager/CHANGELOG.md
+++ b/extensions/quarantine-manager/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Quarantine Manager
 
+## [Scan apps and folders] - {PR_MERGE_DATE}
+
+### Added
+
+- Scan directories for quarantined files: `.app` bundles are scanned recursively (they often contain many internal quarantined files), while plain folders are scanned one level deep so large trees stay responsive
+- **Check Quarantine Status** lists every quarantined item found inside an app or folder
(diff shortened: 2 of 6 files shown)


nurkamol:
Thanks for the review @greptile-apps — addressed in the latest commit:

- **Issue 1 (single-child false "clean")** — fixed. The shallow scan now appends the directory itself as a sentinel argument so `xattr` always has ≥2 path args and emits the `<path>: <attr>` prefixed format, even when a folder has exactly one child. Verified a lone quarantined child is now detected.
- **Issue 2 (ARG_MAX on large folders)** — fixed. Children are processed in chunks of 256 rather than passed in a single `xattr` invocation. Verified on a 300-file folder.
- **Issue 3 (action title case)** — kept as `(recursive)`. The `@raycast/prefer-title-case` rule in `@raycast/eslint-config` warns on `(Recursive)` and auto-fixes it to lowercase, so `ray lint` is only clean with the lowercase form. Confirmation-dialog titles and headings still use Title Case `(Recursive)`.

nurkamol:
@0xdhrv thank you for reviewing and approving the original submission (#25892)! This is a follow-up update to the same extension, adding the app/folder quarantine scanning that was requested there. If you have a moment, would you mind taking a look? All CI checks pass and the Greptile review feedback has been addressed. No rush at all — much appreciated. 🙏

0xdhrv (approved):
Looks good to me, approved ✅

PR #28648, opened 2026-06-09
[Skills] Resolve skill contents for skills in subdirectories

DaleSeo:
## Description

When we load a skill's contents, we check a couple of standard GitHub paths: `skills/<id>/SKILL.md` and `<id>/SKILL.md`. These cover single-skill repositories and skills at the top level, but they don't account for skills organized in subdirectories, which is common in multi-skill repos. For instance, in https://github.com/mattpocock/skills, each skill is placed under a category folder, so grill-me is located at `skills/productivity/grill-me/SKILL.md`. This results in every guess returning a 404 error, and the loader defaults to the repo-root README. As a result, the detail view and the "Copy Skill Contents" action end up showing the README instead of the actual skill.

This change maintains the quick flat-path guesses for the usual cases. If those guesses fail, it reads the repo's git tree to locate the real SKILL.md at any depth and fetches it using the blob API. Additionally, it removes the fallback to the repo-root README, so if a SKILL.md is genuinely missing, it will report "No Skill Contents Found" instead of mistakenly copying the wrong file.

## Checklist

### DIFF
--- a/extensions/skills/CHANGELOG.md
+++ b/extensions/skills/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Skills Changelog
 
+## [Fix Skill Contents for Nested Skills] - {PR_MERGE_DATE}
+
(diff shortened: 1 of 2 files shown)


keito4 (approved):

0xdhrv (approved):
Looks good to me, approved ✅

PR #28669, opened 2026-06-09
[Slack] Add deep link arguments to Set Status

foobarnes:
## Description

Adds two optional arguments — `statusText` and `emoji` — to the **Set Status** command so it can be driven by a deep link or [Quicklink](https://manual.raycast.com/quicklinks). When the command is launched with either argument, it sets the status directly and closes the window; launched without arguments, it behaves exactly as before (shows the status list/form).

This lets you create one-tap Quicklinks like “Lunch 🍔” or “Heads down 🎧”:

```
raycast://extensions/mommertf/slack/set-status?arguments=%7B%22statusText%22%3A%22Lunch%22%2C%22emoji%22%3A%22%3Ahamburger%3A%22%7D
```

where `arguments` decodes to `{ "statusText": "Lunch", "emoji": ":hamburger:" }`.

### Notes
- Either argument may be omitted; only the field you provide is changed (the other is preserved from your current status).
- `emoji` accepts a Slack emoji name with or without colons (`hamburger` or `:hamburger:`).
- Statuses set this way don't auto-expire; the in-app **Set New Status** form is still the way to choose an expiration.

## Screencast / Screenshots

_Behavior is identical to the existing command when launched with no arguments; the new path is the deep-link/Quicklink flow described above._

## Checklist

### DIFF
--- a/extensions/slack/CHANGELOG.md
+++ b/extensions/slack/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Slack Changelog
 
+## [Add deep link arguments to Set Status] - 2026-06-09
+
+- The **Set Status** command now accepts optional `statusText` and `emoji` arguments, so a deep link or Quicklink can set your status in one step (e.g. `raycast://extensions/mommertf/slack/set-status?arguments=%7B%22statusText%22%3A%22Lunch%22%2C%22emoji%22%3A%22%3Ahamburger%3A%22%7D`).
+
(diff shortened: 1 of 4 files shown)


RobErskine:
This is rad @foobarnes, thanks for this addition. It looks good to me

foobarnes:
Thanks @RobErskine! 🙏 Small tweak, but it'll at least have 1 immediate user (me)

0xdhrv (approved):
Looks good to me, approved ✅

PR #28671, opened 2026-06-09
Update no-more-caffeine(Fix blank caffeine amount on initial open)

atzzCokeK:
## Description

Fixes a bug where the **Caffeine Amount** field was always blank when opening the Log Caffeine command (and after tapping "Log Another").

**Root cause:** Both `drinkType` and `caffeineAmount` were initialised to `""`. The `useEffect` that auto-fills the amount from the selected preset has an early-return for `!drinkType`, so it never ran on first render — even though the dropdown visually showed "Coffee" (Raycast renders the first item by default, but React state was still `""`). The same reset-to-`""` bug existed in the "Log Another" action.

**Fix:** Initialise both states directly from `BUILT_IN_PRESETS[0]` so the correct values are present from the very first render, before any effects fire.

## Screenshots
| Before | After |
| -- | -- |
| image |  image |

## Checklist

### DIFF
--- a/extensions/no-more-caffeine/CHANGELOG.md
+++ b/extensions/no-more-caffeine/CHANGELOG.md
@@ -1,6 +1,6 @@
 # No More Caffeine Changelog
 
-## [Fix Default Caffeine Amount] - {PR_MERGE_DATE}
+## [Fix Default Caffeine Amount] - 2026-06-09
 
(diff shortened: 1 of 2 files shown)


0xdhrv (approved):
Looks good to me, approved ✅

PR #28711, opened 2026-06-11
Add call-contacts extension

AlexanderGuru:
## Description

New extension: **Call Contacts** — search your macOS contacts and call them in two keystrokes. The default action opens a `tel:` link, so the call is placed through the paired iPhone via Continuity (a regular cellular call). FaceTime Audio/Video and Messages are available as secondary actions.

Highlights:

- Search by name, organization, or any part of a phone number (including middle digits)
- Dial any typed number that isn't in the contacts
- Frequently called contacts rise to the top (frecency sorting, with a Reset Ranking action)
- Contact photos with an on-disk cache; colored initials avatars as fallback
- Fast and private: a single CNContactStore pass via the JXA/ObjC bridge — no Apple Events, the Contacts app is never launched, one macOS permission, no network calls. ~500 contacts load in about a second

## Screencast

Screenshots are included in the `metadata/` folder (contact names and numbers in them are anonymized).

## Checklist


pernielsentikaer (commented):
Hi 👋 

I just tested the search a number in the middle, and it seems to work fine, can you check if it's also working for you? It's also soon possible to search directly from the root search 🙂

PR #28728, opened 2026-06-12
[Slack] Fix Set Status deep link erroring on raw emoji

foobarnes:
## Description

Follow-up bug fix to #28669, which added optional `statusText` and `emoji` deep link arguments to the **Set Status** command.

That PR didn't account for how the **base Set Status command** renders the new `emoji` argument. Raycast attaches its built-in **emoji auto-picker** to a text argument field, so picking from it inserts a **raw Unicode emoji** (e.g. 👈) rather than the `:name:` text the command was written to expect. `normalizeEmoji` then colon-wrapped the glyph into `":👈:"`, which Slack's `users.profile.set` rejects with `profile_status_set_failed_not_valid_emoji` — so the command errored out for anyone who used the picker (the natural way to fill the field).

### Fix

`normalizeEmoji` now detects a raw emoji glyph and maps it back to its Slack `:name:` code via a reverse of `SLACK_EMOJI_CODE_MAP` (with a variation-selector fallback so e.g. `☝️` resolves). ASCII input (`rocket` / `:rocket:`) keeps the existing colon-normalizing path, so the Quicklink/deep-link examples from #28669 keep working unchanged.

### Before — raw emoji from the picker errors out

https://github.com/user-attachments/assets/6c82aaab-6dd5-491b-ad87-f6eb3429f70b

### After — raw emoji is mapped to its `:name:` and the status is set

https://github.com/user-attachments/assets/84c2960c-718d-4831-8c11-3671a41bb929

### Checklist

- [x] I checked that screenshots/recordings are up to date

### DIFF
--- a/extensions/slack/CHANGELOG.md
+++ b/extensions/slack/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Slack Changelog
 
-## [Fix Set Status deep link with raw emoji] - {PR_MERGE_DATE}
+## [Fix Set Status deep link with raw emoji] - 2026-06-13
 
 - Fix the **Set Status** command erroring with `profile_status_set_failed_not_valid_emoji` when the `emoji` argument is a raw emoji (e.g. 👈) inserted by Raycast's emoji picker. Raw emoji are now mapped back to their Slack `:name:` code.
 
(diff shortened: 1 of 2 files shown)


foobarnes:
Sorry to have missed this, @RobErskine @0xdhrv. I was hyper-focused on the one use case I was trying to implement and didn't think holistically enough.

foobarnes:
After use, I'm really interested in adding the ability to set the status expiration time in this same way. Do you think that that many fields would be cluttery? Is there a way to prevent args from automatically being added to the main command prompt?

0xdhrv (approved):
Looks good to me, approved ✅

clins1994:
@foobarnes awesome. thank you

PR #28734, opened 2026-06-12
[amphetamine] Forget "Until Time" dates

bendrucker:
## Description

Changes the "Configure New Session" command so the Until Time mode stops restoring a stale absolute time. The picker used `storeValue`, which re-filled the exact datetime from a past session on each launch. That time has already passed, so it was never a useful default.

The picker now starts empty and requires a future time. Blur and submit both validate it, showing an inline error otherwise. For Duration mode is untouched. It still remembers your last duration and interval, which is the behavior people want there.

## Screencast

Screenshot 2026-06-12 at 12 16 45

Screenshot 2026-06-12 at 12 16 49

## Checklist

### DIFF
--- a/extensions/amphetamine/CHANGELOG.md
+++ b/extensions/amphetamine/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## [Require a fresh until-time on each launch] - {PR_MERGE_DATE}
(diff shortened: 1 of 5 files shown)


0xdhrv (approved):
Looks good to me, approved ✅

PR #28748, opened 2026-06-13
Add claude-code-resume extension

izumiz-dev:
## Description

Claude Code Resume lets users browse and resume any previous Claude Code session directly from Raycast, rather than only the latest session available through `claude --continue`.

It displays each session’s title, prompts, and last reply, then resumes the selected session in its original directory. It also supports opening recent projects and works on macOS and Windows, including WSL and native PowerShell sessions.

Original repository: https://github.com/izumiz-dev/raycast-claude-code-resume

### Open Claude Code Project Command

https://github.com/user-attachments/assets/169e9e95-e0ba-495a-a967-4abe77e09ff6

### Resume Claude Code Session Command

https://github.com/user-attachments/assets/2f30ffbb-6eed-4da2-9c4f-a4c7221e30fa

### Windows

https://github.com/user-attachments/assets/298d51db-1184-4411-b89d-b45c303e3feb

## Checklist


izumiz-dev:
Thanks for the review! All four points are addressed:

1. **eslint** — now uses `defineConfig([...raycastConfig])` from `eslint/config`.
2. **Preferences** — dropped the manual `Prefs` interface for the generated `Preferences` type (added `raycast-env.d.ts` to the tsconfig `include` so it resolves).
3. **`clip`** — exported from `sessions.ts` and imported in `list-sessions.tsx`, removing the duplicate.
4. **Temp scripts** — leftover launch scripts are now swept on each launch. I used a startup sweep rather than `setTimeout`, since a Raycast command process can exit before the timer would fire.

Smoke-tested on both Windows and macOS — working as expected with no issues.

izumiz-dev:
Small fix pushed while this is still in review — not a new feature, just cleaning up a visible issue before the first release.

The macOS Ghostty path used `open -na Ghostty`, and `-n` spawns a fresh app instance every launch, so each session added another Ghostty icon to the Dock. Ghostty 1.3+ has an AppleScript dictionary, so it now opens a window in the *running* instance instead — single Dock icon. Falls back to the old path on older Ghostty or if the one-time macOS Automation consent is declined.

Verified on Ghostty 1.3.1; `ray lint`/`ray build` pass. Only the macOS Ghostty path changed. (izumiz-dev/raycast-claude-code-resume#3)

0xdhrv (commented):

We already have a few extensions in the Store that cover similar Claude Code workflows. Could we consider enhancing one of the existing extensions below instead of creating another one?

* [ClaudeCast](https://www.raycast.com/qazi0/claudecast) by @qazi0
* [Claude Sessions](https://www.raycast.com/kud/claude-sessions) by @kud
* [Claude Code Launcher](https://www.raycast.com/stephendolan/claude-code-launcher) by @stephendolan

There is significant overlap around browsing or resuming Claude Code sessions and opening Claude Code projects directly from Raycast.

If there are unique features or workflows you’re aiming to add—such as Windows, WSL, and native PowerShell support—we’d love to hear about them and see whether they could be integrated into one of these extensions to avoid duplication and improve discoverability.

>

PR #28766, opened 2026-06-14
Add BubbleUsage extension

Prajz:
## Description

Adds BubbleUsage, a Raycast extension for tracking connected AI coding agents and locally observed usage. The extension focuses on what providers expose locally or through available dashboards instead of asking users to enter manual quota limits.

## Features
- Detects connected/disconnected state for Claude, Codex, Gemini, and OpenCode Go
- Shows local 5 hour, weekly, and monthly usage windows where session logs support them
- Provides connect/install actions for disconnected agents
- Opens provider usage dashboards when available
- Supports OpenCode Go setup with local workspace/auth-cookie storage for dashboard usage


0xdhrv (commented):
.

0xdhrv (commented):
Thanks for your contribution @Prajz 🔥

We already have an extension in the Store that covers this functionality. Could we consider enhancing the existing extension below instead of creating another one?

* [Agent Usage](https://www.raycast.com/thuggyduck/agent-usage) by @thuggyduck

The existing extension already provides a consolidated view for monitoring AI coding-agent usage, including Claude, Codex, Gemini, and OpenCode Go—the same providers supported by this PR. It also detects local credentials, displays provider quota and reset windows, and supports the OpenCode Go workspace ID and authentication-cookie setup.

It currently supports several additional providers, including Amp, Copilot, Cursor, Droid, Grok, Kimi, Synthetic, Antigravity, z.ai, and MiniMax. It also includes a menu-bar command and supports both macOS and Windows.

Because the purpose, supported providers, configuration, and core usage-tracking workflow overlap very closely, the functionality introduced here would likely fit better as an enhancement to the existing extension.

PR #28776, opened 2026-06-15
[Apple Mail] Escape compose AppleScript strings

### LINKED ISSUE #28775: [Apple Mail] Compose fails when subject contains double quotes
### Extension

https://www.raycast.com/yug2005/mail

### Raycast Version

0.64.4.0 (Beta)

### OS Version

macOS 26.5.1

### Description

When composing a new email in the Raycast Apple Mail extension, sending fails if the subject contains double quotes (").
This appears to be an AppleScript escaping issue. Removing the quotes from the subject makes the message send successfully.

- The issue is reproducible only when the subject contains double quotes.
- Removing the quotes avoids the error.
- This suggests user-provided text is not being escaped correctly before being passed into AppleScript.

### Steps To Reproduce

1. Open the Apple Mail extension in Raycast
2. Start composing a new message
3. Enter a subject containing double quotes, for example: He said "hello"
4. Send the message
5. See error

### Current Behaviour

The command fails with:

```
Error: Command failed with exit code 1: osascript
356:361: syntax error: Expected “,” or “}” but found “error”. (-2741)

xre:run-applescript.js:38:21
async i$:compose-new-message.js:354:5
```

### Expected Behaviour

The email should be composed and sent successfully, with the subject line correctly escaped.

mturac:
## Summary

Fixes #28775.

Escapes the user-provided strings that are interpolated into the Apple Mail compose AppleScript before sending. This covers subjects with double quotes, and also keeps the same compose path safe for sender, recipients, message body, account/mailbox names, and attachment aliases.

## Root cause

`sendMessage` already had an AppleScript string escaping helper, but the compose script interpolated the outgoing message fields directly. A subject such as `He said "hello"` could therefore terminate the AppleScript string literal and produce a syntax error from `osascript`.

## Validation

- `npm run lint`
- `npm run build`
- `git diff --check`

### DIFF
--- a/extensions/mail/CHANGELOG.md
+++ b/extensions/mail/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Mail Changelog
 
+## [Fix] - {PR_MERGE_DATE}
+
+- Fix sending composed messages when the subject or other message fields contain double quotes.
+
 ## [Improvement] - 2026-06-10
 
 - Add option to manually set the type of a mailbox (inbox, junk, etc.), useful for non-English mail accounts
--- a/extensions/mail/CHANGELOG.md
+++ b/extensions/mail/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Mail Changelog
 
-## [Fix] - {PR_MERGE_DATE}
+## [Fix] - 2026-06-15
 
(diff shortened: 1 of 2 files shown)


pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #28780, opened 2026-06-15
Add ten-four extension

berwickgeek:
## Description

**Ten Four** is a shelf of clean, copyable snippets. You push snippets to it from your terminal (or from agent tools like Claude Code), then browse and copy them from Raycast.

### Problem it solves

When a CLI prints something you want to copy (a URL, an API key, a path, a command, a multi-line code block), selecting it straight out of the terminal gives you mangled line breaks and stray indentation, because you are selecting reflowed text off the terminal's character grid. Ten Four fixes the root cause: snippets travel as **data**, never as rendered terminal text, so you copy them out of Raycast with the exact bytes intended (no wrapping, no leading spaces).

### How it works

```
terminal / Claude Code  --(tenfour CLI)-->  ~/.ten-four.json  -->  Raycast "Ten Four"
       (writer)                              (the shelf file)        (reader)
```

### Commands

1. **Ten Four Shelf** (`view`) — a searchable list of your snippets, newest first, with a live-updating detail pane that renders multi-line content verbatim. Actions: Copy Snippet, Paste to Active App, Copy Label, Pin/Unpin, Toggle Detail, Remove, Clear Shelf. The list watches the shelf file and refreshes as new snippets arrive.
2. **Install Ten Four CLI** (`view`) — a one-click installer for the companion `tenfour` CLI (the writer). It shows install status and the target path, then copies the bundled CLI into a directory on the user's `PATH`.

### Privacy and data handling

- **No network calls, no authentication, no telemetry.** Everything is local.
- Snippets are stored as JSON in `~/.ten-four.json` (overridable with the `TENFOUR_FILE` env var). The shelf keeps the most recent 200 entries; pinned entries are never trimmed.

### About the bundled `assets/tenfour` file

`assets/tenfour` is the companion CLI, included so the extension is self-contained. It is:

- A single **zero-dependency Node script** (only Node built-ins: `fs`, `os`, `path`), ~130 lines, human-readable, no compiled/minified binary.
- Used **by the extension itself**: the *Install Ten Four CLI* command reads it from `environment.assetsPath` and copies it into the user's `PATH` (preferring `/opt/homebrew/bin` or `/usr/local/bin`, falling back to `~/.local/bin`).
- Installed **only on an explicit button press**, never automatically. The command clearly states the destination path before installing.

This is what lets a terminal or an agent run `tenfour "..."` to push a snippet onto the shelf the extension reads. Users who prefer not to install the CLI can still use the extension; the CLI is purely the optional "writer" side.

## Screencast

Ten Four demo

The clip: Claude Code answers a question and pushes the resulting SQL query onto the shelf on its own (via the bundled `tenfour` CLI); the query is then copied from Raycast and pasted, cleanly, into a SQL console.

Screenshots are included in `metadata/`:

- `metadata/ten-four-1.png` — the **Ten Four Shelf** list with the detail pane.
- `metadata/ten-four-2.png` — the **Install Ten Four CLI** command.

Companion repository (CLI source, README, install script): https://github.com/berwickgeek/ten-four

## Checklist

- [x] I checked that files in the `assets` folder are used by the extension itself (`icon.png` is the extension icon; `tenfour` is read and installed by the *Install Ten Four CLI* command)
- [x] I checked that assets used by the `README` are placed outside of the `metadata` folder (the store `README` references no local images)


berwickgeek:
Thanks for the automated review. Addressed in the latest commits on this branch:

- **Unused `@raycast/utils`** — removed from `dependencies` and regenerated `package-lock.json` to match.
- **Reveal in Finder** — switched to `Action.ShowInFinder` (idiomatic, keeps the standard ⌘⇧R shortcut).
- **ESLint config** — now uses the official `defineConfig([...raycastConfig])` v9 pattern.

Two points left as-is, intentionally:

- **`author` field** — `jaymcc` is correct: it's my Raycast account slug. The bot compared it to my GitHub login (`berwickgeek`), which is a different namespace and isn't a Raycast user, so using it would actually fail validation. `ray lint` validates `jaymcc` successfully.
- **Copy vs. symlink in *Install CLI*** — copying is deliberate. A symlink into the extension's `assetsPath` would leave a dangling `tenfour` on the user's `PATH` the moment the extension is updated or uninstalled; a copy keeps the installed CLI robust and independent of the extension lifecycle. I added a code comment noting the rationale.

`ray lint` and `ray build` both pass clean.

berwickgeek:
Followed up on the remaining Greptile points in the latest commit:

- **`persist` failure / state desync** (the one real finding) — `mutate()` now writes to disk *first* and only updates the UI state on success. If `writeFileSync` throws, it surfaces a `Toast.Style.Failure` and bails, so the shelf view can no longer drift out of sync with `~/.ten-four.json`. The "Removed" / "Shelf cleared" success toasts only fire on a confirmed write.
- **CLI blocks on interactive TTY** — the implicit stdin fallback in `assets/tenfour` is now guarded with `!process.stdin.isTTY`, so running `tenfour` with no args and no pipe fails fast with the usage message instead of hanging on `Ctrl+D`. The explicit `-` / `--stdin` path is unchanged.

Left the `scripts/generate-icon.js` / `install.sh` dev artifacts in place — they're harmless and `ray build` only bundles the entry points, but happy to drop them from the package if you'd prefer a leaner tree.

`ray lint` and `ray build` both pass clean.

extensions/ten-four/metadata/ten-four-2.png
0xdhrv, line 1: Issue: metadata image style

Reference:

0xdhrv (changes requested):
Hey @berwickgeek 👋 

I have added a comment for you to address.

---

berwickgeek:
Thanks @0xdhrv 🙏 Retook both metadata screenshots so the footer now shows only the extension icon and command name, with the local-extension (hard drive) indicator removed, matching the store appearance. Same background as before, dimensions unchanged (2000x1250). Ready for another look whenever you get a chance.

extensions/ten-four/assets/tenfour
0xdhrv, line 1: **suggestion** (non-blocking): Could we keep the CLI outside the extension assets so updates can be streamlined?

Reference:
- https://developers.raycast.com/basics/prepare-an-extension-for-store#binary-dependencies-and-additional-configuration

0xdhrv (changes requested):
Hey @berwickgeek 👋 

I have added a comment for you to address.

---

extensions/ten-four/assets/tenfour
berwickgeek: Thanks @0xdhrv 🙏 Fair point, let me explain the reasoning here.

`assets/tenfour` isn't a compiled binary dependency in the sense the docs describe. It's a tiny (~2 KB) zero-dependency Node.js text script. The *Install Ten Four CLI* command simply copies it into the first writable bin dir on the user's `PATH`, so it lives independently of the extension after that. Nothing in the extension executes it at runtime, so there's no runtime binary to stream or extract.

Because it's plain source text with no build step and no external fetch, bundling it in `assets/` keeps the extension fully local (no network access at all, which is one of its selling points) and means the installed CLI keeps working even after the extension is updated or removed. Downloading it at runtime instead would add a network dependency and a failure mode for no real gain given the size.

Happy to revisit if you'd still prefer it moved, but on that basis I've left it as-is. Marking as ready for review 🙂

0xdhrv (approved):
Looks good to me, approved ✅

PR #28784, opened 2026-06-15
ado-search: customisable work item state order

stoumann:
## Description

Adds a **"Preferred State Order"** setting to the **My Work Items** command so users can pin states to the top of the list in their own order — e.g. surfacing "Doing" / "In Progress" above everything else.

### Why
The list groups work items by state, but the order was driven by the `States to Show` `TagPicker`, and Raycast's `TagPicker` doesn't preserve the order items are selected (it returns them in the picker's own order). So there was no reliable way to control section order.

### How
- New optional **Preferred State Order** text field in Setup: comma-separated, case-insensitive (e.g. `Doing, In Progress, Code Review`).
- Listed states float to the top in that order; any remaining selected states follow in the existing order; everything else stays alphabetical.
- Backward compatible — an empty field keeps the previous default ordering. Persisted alongside the other in-app settings via `LocalStorage`.
- Updated the `States to Show` help text to point at the new field.

### Drive-by
Replaced a stray NUL byte in the rich-text cache key separator with an explicit Unicode escape — no behaviour change, but the raw byte made the file read as binary to some tooling.

## Checklist
- [x] `npm run lint` passes
- [x] Changes limited to `extensions/ado-search/`


0xdhrv (approved):
Looks good to me, approved ✅

PR #28790, opened 2026-06-15
Add read-this-later extension

shearmds:
## Checklist


shearmds:
Hi, and thanks for taking a look! 👋

This extension brings my Read This Later read-it-later service into Raycast, so saving and triaging articles never requires leaving the keyboard.

What it does

Show Unread Links — browse your unread saved articles, then open (and auto-mark-read), mark read, or copy a URL without touching the mouse.
Save Current Tab — save the frontmost browser tab to your list. It reads the exact active-tab URL via the Raycast browser extension (with an AppleScript fallback), and fetches a clean og:title when the browser only reports a domain — so saves are accurate even on paywalled news sites.
How it fits the product Read This Later is a small cross-platform read-it-later app — an iOS/iPadOS app (share sheet, Siri/Shortcuts) and a Chromium browser extension (Dia, Chrome, Arc, Brave, Edge), with everything syncing automatically across devices. Raycast is the natural fit for the desktop "save what I'm looking at / what should I read next" moment, and this extension closes that loop. More on the project: https://shearmds.github.io/ReadThisLater/

On the sync token The extension talks to the app's hosted sync service over HTTPS and authenticates with a personal token. Existing users already have this token from the iOS app / their account; it's entered once in the extension's preferences and stored in the macOS keychain (never in source). The README's Setup section documents where to get it. Happy to expand that wording if you'd like it clearer for new users.

Grateful for any feedback — glad to make changes.

shearmds:
Thanks for the thorough review — all really helpful. I've pushed fixes:

Preferences interface: Removed the manual definition; now using the auto-generated Preferences type. ✅
Debug logs: Removed — they were leftover diagnostics and shouldn't have shipped. Good catch on the privacy angle. ✅
Favicons: Switched list items to getFavicon(item.url) with a bookmark fallback. Much nicer, thanks. ✅
Sequential round-trips: Fixed setRead to a single request — the list view already holds the full item, so no pre-fetch is needed. I've intentionally kept saveItem as a read-then-write: the first request powers the "Already saved" duplicate detection that surfaces a helpful message, and saving the current tab is an infrequent, manual action. Happy to collapse it to one request if you'd prefer I drop that duplicate check.

shearmds:
Good catch — that's a real bug. An og:title containing an apostrophe (e.g. content="World's Best Pizza") was truncated at the apostrophe, and since the match was non-empty the <title> fallback never ran. Fixed by matching the closing delimiter to the opening quote (separate patterns for "…" and '…'). Pushed.

shearmds:
Its been three weeks now. is there any way to get someone to review this? thanks

0xdhrv (commented):

We already have an extension in the Store that covers this workflow. Could we consider enhancing the existing extension below instead of creating another one?

* [[Read Later](https://www.raycast.com/jimluo/later)](https://www.raycast.com/jimluo/later) by @jimluo

Both extensions let users save the current browser tab and browse or manage saved links from Raycast, so the core functionality is very similar.

If there are unique features or workflows you’re aiming to add, we’d love to hear about them and see whether they could be integrated into the existing extension to avoid duplication and improve discoverability.

shearmds:
Hi there. Thanks for responding. This extension syncs directly to an iOS app of mine that does have some overlap with other bookmark programs but also has some unique functions. One of those is that there are very few extensions that sync with iOS apps directly. Also, my app has an offline reading mode and the ability to take notes on the sites saved. If you wouldn’t mind looking at my Webb page, it might help.

https://shearmds.github.io/ReadThisLater/

I think having the Raycast extension would be helpful for the many people who love Raycast but would like to use it with my suite of apps.

Thanks again.

Michael D. Shear
White House Correspondent
The New York Times
***@***.***
202-681-0415 (mobile/signal)

shearmds:
Also, this submission was a very early version of the app, which has advanced significantly with new features since then. I was told not to try to update the raycast extension until it had been approved, but I am quite open to adding to the feature set to make it comparable to my other apps, and in fact would plan to do that.

please let me know if i should resubmit, and if doing so will require another long wait.

thanks

mike

0xdhrv (approved):
Looks good to me, approved ✅

shearmds:
Thank you. I look forward to updating and extending the extension.

PR #28810, opened 2026-06-16
World Cup: add menu bar and live score commands

brunocbreis:
## Description

Adds two commands to the existing **FIFA World Cup 2026™** extension (`isma`), contributing to it rather than shipping a separate extension:

- **World Cup Menu Bar** (`menu-bar`) — shows the live match score in the macOS menu bar with a dropdown of every fixture (live / upcoming / finished), plus a one-click toggle to mute goal alerts and shortcuts to open the ESPN scoreboard.
- **World Cup Live Score** (`no-view`, 1-min interval) — keeps the live score under the command name and posts a HUD when a goal is scored. The HUD is celebratory (⚽🎉) by default, or commiserating (⚽😭) when the goal is against the team set in the per-command **Your Country** preference.

Both commands are self-contained under `src/lib/` and reuse the extension's existing `flags` map, so the existing **Matches** command is untouched. Data comes from the same ESPN `fifa.world` scoreboard endpoint.

## Checklist

- [x] `npm run build` and `npm run lint` pass
- [x] I added a CHANGELOG entry

### DIFF
--- a/extensions/world-cup/CHANGELOG.md
+++ b/extensions/world-cup/CHANGELOG.md
@@ -1,5 +1,10 @@
 # World Cup Changelog
 
+## [Menu Bar & Live Score] - {PR_MERGE_DATE}
+
+- Added a **World Cup Menu Bar** command showing the live score in the menu bar with a dropdown of every fixture, and a quick toggle to mute goal alerts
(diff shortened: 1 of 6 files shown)


0xdhrv (approved):
Looks good to me, approved ✅

PR #28166, opened 2026-05-20
[iTerm] Fix profile parsing

mturac:
## Description

Reads the iTerm preferences plist as JSON and maps each profile object directly instead of pairing filtered `Guid` and `Name` lines from shell output. This keeps profiles visible when plist fields appear in a different order or include additional keys.

Fixes #26822

## Screencast

N/A

## Checklist

- [x] I ran `npm run lint` and `npm run build`

PR #28865, opened 2026-06-18
[iTerm] Fix profile parsing for plists with Infinity floats

### LINKED ISSUE #28863: [iTerm] "Open iTerm Profile" still broken after #28166/#28723 - plutil JSON conversion fails on Infinity floats in status bar config
### Extension

https://www.raycast.com/ron-myers/iterm

### Raycast Version

1.104.19

### OS Version

26.3 (25D125)

### Description

PR #28166 switched profile parsing from `PlistBuddy+grep` to `plutil -convert json`. PR #28723 further improved this to `plutil -extract "New Bookmarks" json` to handle plists that can't be converted in their entirety.

   However, **both approaches fail** when iTerm profiles contain IEEE 754 `Infinity` float values, which iTerm writes by default for status bar components with unlimited max width:

   ```
   New Bookmarks[*].Status Bar Layout.components[*].configuration.knobs.maxwidth = inf
   ```

   JSON has no representation for `Infinity`, so `plutil` rejects the conversion even when scoped to just the `New Bookmarks` key:

   ```
   $ plutil -extract 'New Bookmarks' json -o - ~/Library/Preferences/com.googlecode.iterm2.plist
   .../com.googlecode.iterm2.plist: Invalid object in plist for JSON format
   ```

   This affects any user who has status bar components configured in their iTerm profiles. Enabling the status bar and keeping the default "unlimited" max width is enough to trigger it.

   **Root cause:** `plutil -convert json` and `plutil -extract ... json` cannot serialize plist `<real>` values that are `infinity` or `nan`, because JSON (RFC 8259) does not support IEEE 754 special values. iTerm stores `inf` as
   the `maxwidth` knob for status bar components with no width limit.

   **Suggested fix:** Use `xml1` format instead of `json`:

   ```
   plutil -extract "New Bookmarks" xml1 -o - ~/Library/Preferences/com.googlecode.iterm2.plist
   ```

   XML plist format can represent all plist types including `<real>inf</real>`. I verified this produces all profiles correctly on my machine, while both JSON approaches fail.

### Steps To Reproduce

   1. Open iTerm → Preferences → Profiles → Session → enable "Status bar enabled"
   2. Configure status bar with components that have the default unlimited max width (this is the default for most components)
   3. Open Raycast → "Open iTerm Profile"
   4. Only some or no profiles are shown

   To verify the underlying cause:

   ```bash
   # This will fail with "Invalid object in plist for JSON format":
   plutil -extract 'New Bookmarks' json -o - ~/Library/Preferences/com.googlecode.iterm2.plist

   # This works — xml1 handles Infinity floats:
   plutil -extract 'New Bookmarks' xml1 -o - ~/Library/Preferences/com.googlecode.iterm2.plist
   ```

### Current Behaviour

_No response_

### Expected Behaviour

All saved iTerm profiles should be listed.

SimonMayerhofer:
## Description

`plutil -extract "New Bookmarks" json` (introduced in #28166, refined in #28723) fails when iTerm profiles contain IEEE 754 `Infinity` float values. iTerm stores `+infinity` as the `maxwidth` knob for status bar components with no width limit — a common default configuration.

```
$ plutil -extract 'New Bookmarks' json -o - ~/Library/Preferences/com.googlecode.iterm2.plist
Invalid object in plist for JSON format
```

This PR switches from `json` to `xml1` format for the `plutil -extract` call and uses the `plist` npm package to parse the XML output. The `xml1` format handles all plist types including `<real>+infinity</real>`.

### Changes

- `get-iterm-profiles.tsx`: Replace `plutil -extract ... json` + `JSON.parse` with `plutil -extract ... xml1` + `plist.parse`
- `package.json`: Add `plist` and `@types/plist` dependencies
- `CHANGELOG.md`: Add bug fix entry

Fixes #28863

## Screencast

Tested locally with 12 profiles, all containing status bar components with `+infinity` maxwidth values. All profiles now appear correctly.

### DIFF
--- a/extensions/iterm/CHANGELOG.md
+++ b/extensions/iterm/CHANGELOG.md
@@ -1,6 +1,6 @@
 # iTerm Changelog
 
-## [Bug Fixes] - {PR_MERGE_DATE}
+## [Bug Fixes] - 2026-06-19
 
 - Fixed `Open iTerm Profile` failing when profiles contain values not representable in JSON (e.g. Infinity floats from status bar config)
 
--- a/extensions/iterm/CHANGELOG.md
+++ b/extensions/iterm/CHANGELOG.md
@@ -1,5 +1,9 @@
 # iTerm Changelog
 
+## [Bug Fixes] - {PR_MERGE_DATE}
+
+- Fixed `Open iTerm Profile` failing when profiles contain values not representable in JSON (e.g. Infinity floats from status bar config)
+
 ## [Bug Fixes] - 2026-06-16
 
 - Fixed `Open iTerm Profile` if parts of a user's iTerm preferences plist other than profiles cannot be converted to JSON
--- a/extensions/iterm/package-lock.json
+++ b/extensions/iterm/package-lock.json
@@ -9,11 +9,13 @@
       "dependencies": {
         "@raycast/api": "^1.104.11",
         "@raycast/utils": "^2.2.3",
-        "file-url": "^4.0.0"
+        "file-url": "^4.0.0",
+        "plist": "^3.1.0"
       },
       "devDependencies": {
         "@raycast/eslint-config": "^2.1.1",
         "@types/node": "25.5.2",
+        "@types/plist": "^3.0.5",
         "@types/react": "19.2.14",
         "eslint": "^10.2.0",
         "prettier": "^3.8.1",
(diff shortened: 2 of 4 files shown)


ches:
LGTM to me in general 👍🏼

I figured after #28723 that someone eventually would run into profile features I don't use… that didn't take long 😅 Adding the plist dep for parsing XML that doesn't have these serialization gaps seems like a good middle ground versus using the scripting API that needs additional user setup effort outside of Raycast for Python (would be nice if there is a JS implementation of it that this extension can use).

I liked pushing an expected invariant down to the external tool boundary with `-expect array` better than more complicated error handling paths in code, but that's mostly subjective.

0xdhrv (approved):
Looks good to me, approved ✅

PR #28881, opened 2026-06-19
fix(todoist): guard against empty sync response arrays

HrushiYadav:
Several Sync API mutation functions access `[0]` on response arrays without checking if the array is empty. If the Todoist Sync API returns an empty array (network hiccup, server-side delay, eventual consistency), `undefined` gets merged into the cached data, corrupting the UI state and causing downstream crashes when React tries to render the `undefined` entry.

`updateTask` already has this guard (`updatedData.items.length === 0` early return) -- this PR adds the same pattern to the 6 functions that were missing it:

- `updateProject` -- `updatedData.projects[0]`
- `moveTask` -- `updatedData.items[0]`
- `addLabel` -- `addedData.labels[0]`
- `updateLabel` -- `updatedData.labels[0]`
- `updateFilter` -- `updatedData.filters[0]`
- `updateComment` -- `updatedData.notes[0]`

### DIFF
--- a/extensions/todoist/CHANGELOG.md
+++ b/extensions/todoist/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Todoist Changelog
 
+## [Guard against empty sync response arrays] - {PR_MERGE_DATE}
+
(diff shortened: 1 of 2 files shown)


0xdhrv (approved):
Looks good to me, approved ✅

PR #28882, opened 2026-06-19
fix(spotify-player): fix stale seek position in menu bar skip/back actions

HrushiYadav:
The menu bar "Skip 15 seconds" and "Back 15 seconds" actions for podcast episodes use `currentlyPlayingData.progress_ms` to compute the seek target. This value comes from the last API fetch, which could be minutes old. The longer since the last fetch, the further off the seek lands.

For example: if the episode was at 0:30 when last fetched and the user clicks "Skip 15s" a minute later, it seeks to 0:45 instead of the expected ~1:45.

The standalone `skip15.ts` and `back15.ts` commands don't have this issue because they fetch fresh data right before seeking. The menu bar uses cached React state instead.

**Fix:** When the track is playing, estimate the real position by adding `Date.now() - timestamp` (time elapsed since the Spotify API generated the response) to `progress_ms`. When paused, `progress_ms` is already accurate since playback isn't advancing.

### DIFF
--- a/extensions/spotify-player/CHANGELOG.md
+++ b/extensions/spotify-player/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Spotify Player Changelog
 
+## [Fix stale seek position in menu bar] - {PR_MERGE_DATE}
+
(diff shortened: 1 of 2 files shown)


0xdhrv (approved):
Looks good to me, approved ✅

PR #28883, opened 2026-06-19
fix(grafana): use encodeURIComponent for Explore link panes parameter

HrushiYadav:
`createExploreLink` builds an Explore URL with a `panes` query parameter containing `JSON.stringify()` output. It uses `encodeURI()` to encode the whole query string, but `encodeURI` does not encode `{`, `}`, `"`, `:`, or `&` -- all of which appear in JSON. This produces broken URLs that Grafana cannot parse.

Also fixes a `&&` typo in the query string (`schemaVersion=1&&panes=...` should be `schemaVersion=1&panes=...`).

**Before:**
```
/explore?schemaVersion=1&&panes={%22zmo%22:{%22datasource%22:%22abc%22}}
```

**After:**
```
/explore?schemaVersion=1&panes=%7B%22zmo%22%3A%7B%22datasource%22%3A%22abc%22%7D%7D
```

### DIFF
--- a/extensions/grafana/CHANGELOG.md
+++ b/extensions/grafana/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Grafana Changelog
 
-## [Fix Explore link URL encoding] - {PR_MERGE_DATE}
+## [Fix Explore link URL encoding] - 2026-06-19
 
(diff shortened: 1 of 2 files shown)


Lp-Francois (approved):
Thanks for the fix :)

0xdhrv (approved):
Looks good to me, approved ✅

PR #28884, opened 2026-06-19
Fix SQL injection in Arc history and download search

HrushiYadav:
## Description

The Arc extension's history and download search have SQL injection vulnerabilities in how user input is escaped for LIKE queries:

1. **Wrong quote escaping**: The code escaped double quotes (`""`), but SQLite string literals use single quotes. Double quotes are identifier quotes in SQLite, so the escaping was ineffective.

2. **Missing LIKE wildcard escaping**: Characters `%` and `_` are wildcards in SQL LIKE expressions. A search for `100%` or `test_data` would match unintended results. Added proper escaping with an `ESCAPE` clause.

3. **Unescaped input in `getHistory`**: The non-hook `getHistory` function passed raw `searchText` directly to `getHistoryQuery` without any escaping, unlike `useHistorySearch` which at least attempted escaping.

All three issues are fixed across both history search (`history.ts`) and download search (`sql.ts`, `search-downloads.tsx`).

## Changes

- `src/history.ts`: Fix escaping in `useHistorySearch`, add `ESCAPE` clause to LIKE in `getHistoryQuery`, add escaping for `getHistory`
- `src/sql.ts`: Add `ESCAPE` clause and single quotes to LIKE in `getDownloadQuery`
- `src/search-downloads.tsx`: Fix escaping to handle single quotes and LIKE wildcards
- `CHANGELOG.md`: Add entry


0xdhrv (commented):
Hi @HrushiYadav 👋

Thanks for the contribution 🔥

---

0xdhrv (approved):
Looks good to me, approved ✅

PR #28885, opened 2026-06-19
[codex] Support configuring target key lights

merill:
## Summary

- add a `Configure Target Lights` command so users can choose between controlling all lights or only selected discovered lights
- persist the selected target lights and apply that targeting across toggle, brightness, temperature, preset, and tool-based operations
- use Elgato accessory info to show friendlier discovered light names and make rediscovery/retry logic more resilient for transient connection resets

## Why

The extension currently applies commands to every discovered Elgato light. That breaks down for setups with lights in different rooms, where a shortcut should only affect a chosen subset of devices.

## Implementation notes

- added a small local target-selection store with `all` vs `selected` modes
- introduced a dedicated configuration UI for discovering lights and choosing the ones to control
- expanded light matching so selected lights can be re-identified by display name, Bonjour name, host, FQDN, or IP address after rediscovery
- updated toggle to retry more safely after transient network errors such as `ECONNRESET`

## Validation

- `npm run lint`
- `npm run build`
- manually tested in Raycast Beta with multiple Elgato lights, including selecting only a subset and toggling them successfully


merill:
Here's how it works.

You first use `Configure Target Lights`

CleanShot 2026-06-19 at 21 38 38@2x

You can then control which ones the commands like Toggle Lights will apply to.

CleanShot 2026-06-19 at 21 39 34@2x

Now Raycast on my study won't toggle the Elgato lights that are in my son's room. Plus he can control the lights in his room through Raycast without turning my lights on and off.

merill:
@thomaspaulmann I've fixed the Greptile review comments. Cheers.

0xdhrv (approved):
Looks good to me, approved ✅

PR #28911, opened 2026-06-20
Orion: fix Search Tabs showing no open tabs

giannisanni:
## Description

Fixes the **Search Tabs** command for the Orion extension, which currently shows **no tabs**. Reproduced on Orion 1.0.8 (147.1).

### Root cause

`Search Tabs` enumerates tabs via JXA, reading `URL`/`name` off the individual tab objects returned by `window.tabs()`:

```js
orion.windows().forEach(window => {
  window.tabs().forEach(tab => { ...tab.url()... tab.name()... })
})
```

On current Orion, calling those property getters on the **objects returned by `window.tabs()`** fails with Apple Event errors (`-1700 "Can't convert types"` / `-1728 "Can't get object"`). The first `tab.url()` throws, `executeJxa` swallows the error, and `fetchLocalTabs` falls back to `[]`, so the list is empty. `Open Tab` and `Close Tab` broke the same way — their `.find(t => t.url() === …)` callbacks throw.

The scripting dictionary is unchanged (`tab` still exposes `URL` and `name`); only per-object getter resolution on the evaluated collection is broken.

### Fix

Two access patterns still work reliably:

- **Bulk** property access on the tab collection: `window.tabs.url()` / `window.tabs.name()` (returns arrays)
- **Indexed** tab specifiers: `window.tabs[i]`

Changes:
- `useTabs` lists tabs via bulk access (also fewer Apple Events, so faster).
- `OpenTabAction` / `CloseTabAction` find the matching index from the bulk arrays, then act on `window.tabs[index]`.

Existing behavior (dedupe by URL, keep first occurrence) is preserved.

### Verification

- `ray build` and `ray lint` pass.
- Tested against live Orion 1.0.8 (147.1): listing returns every open tab across all windows (25 tabs / 7 windows here) with correct titles and URLs; the index lookup used by open/close resolves to the correct tab.

### DIFF
--- a/extensions/orion/CHANGELOG.md
+++ b/extensions/orion/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Orion Changelog
 
-## [Fix Search Tabs] - {PR_MERGE_DATE}
+## [Fix Search Tabs] - 2026-06-21
 
 - Fix "Search Tabs" showing no tabs on current Orion versions. Orion's AppleScript bridge no longer resolves `URL`/`name` getters on the individual tab objects returned by `window.tabs()`; the failure was silently swallowed, leaving the list empty. Switched to bulk property access (`window.tabs.url()` / `window.tabs.name()`) for listing tabs and indexed tab specifiers (`window.tabs[i]`) for opening and closing them.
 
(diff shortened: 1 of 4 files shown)


0xdhrv (approved):
Looks good to me, approved ✅

PR #28914, opened 2026-06-20
[Unsplash] Add Windows support for wallpaper, clipboard, and save

arinbalyan:
## Description

Adds Windows support to the Unsplash extension. macOS code paths are untouched -- early-return `process.platform === "win32"` branches were added before the original AppleScript code in each function.

- Wallpaper: PowerShell + Win32 SystemParametersInfo
- Clipboard: PowerShell System.Windows.Forms  
- Save: downloads to Desktop (macOS keeps the folder picker)
- package.json updated to include "Windows" in platforms
- CHANGELOG.md updated with {PR_MERGE_DATE}

## Screencast

N/A -- platform-conditional code, no UI changes.

## Checklist

### DIFF
--- a/extensions/unsplash/CHANGELOG.md
+++ b/extensions/unsplash/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Unsplash Changelog
 
-## [Windows Support] - {PR_MERGE_DATE}
+## [Windows Support] - 2026-06-29
 
 - Added Windows support for setting wallpaper, copying images to clipboard, and saving images
 - Wallpaper: uses PowerShell + Win32 SystemParametersInfo on Windows
--- a/extensions/unsplash/CHANGELOG.md
+++ b/extensions/unsplash/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Unsplash Changelog
 
+## [Windows Support] - {PR_MERGE_DATE}
+
+- Added Windows support for setting wallpaper, copying images to clipboard, and saving images
+- Wallpaper: uses PowerShell + Win32 SystemParametersInfo on Windows
+- Clipboard: uses PowerShell System.Windows.Forms on Windows
+- Save: downloads to Desktop on Windows (folder picker is macOS-only)
+
 ## [Modernize & OAuth Setup Guide] - 2026-05-28
 
 - Added OAuth setup guide shown before login — displays the required redirect URI and a Connect button so users can set up their Unsplash app without confusion
--- a/extensions/unsplash/package.json
+++ b/extensions/unsplash/package.json
@@ -9,7 +9,8 @@
     "Kang",
     "yug2005",
     "xmok",
-    "alexi.build"
+    "alexi.build",
+    "arinbalyan"
   ],
   "license": "MIT",
   "keywords": [
@@ -333,6 +334,7 @@
     "publish": "npx @raycast/api@latest publish"
   },
   "platforms": [
-    "macOS"
+    "macOS",
+    "Windows"
   ]
 }
--- a/extensions/unsplash/src/functions/copyFileToClipboard.ts
+++ b/extensions/unsplash/src/functions/copyFileToClipboard.ts
@@ -1,6 +1,10 @@
 import { showToast, Toast, environment } from "@raycast/api";
 import { runAppleScript } from "@raycast/utils";
+import { exec } from "child_process";
+import { promisify } from "util";
 import { existsSync } from "fs";
+
+const execP = promisify(exec);
 
 interface CopyFileToClipboardProps {
   url: string;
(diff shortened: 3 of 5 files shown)


eggsy:
AI is great!

I don't have access to a Windows machine right now but you can check Greptile suggestions and follow its directions.

So far, really cool addition! Been expecting this one for a while, thanks 🔥

0xdhrv (commented):
Hi @arinbalyan 👋

Thanks for the contribution 🔥

Could you look into the suggestions from Greptile and also the score is 3/5?

---

arinbalyan:
@0xdhrv I have resolved all the issues pointed out by the greptile bot.. also for the latest commit no more comments left by the bot.. The PR is ready for review

0xdhrv:
> @0xdhrv I have resolved all the issues pointed out by the greptile bot.. also for the latest commit no more comments left by the bot.. The PR is ready for review

Can we please share a screencast of the same in action on Windows?

arinbalyan:
https://github.com/user-attachments/assets/32ede2a3-14f7-4b8d-ad83-fecef09bdfa9

Here's a screen recording from windows @0xdhrv

0xdhrv (approved):
Looks good to me, approved ✅

0xdhrv:
Hey @eggsy 👋 

Unrelated to this PR, could you please take a look?

There are a few other Unsplash API guideline issues worth flagging:

- Name usage: Unsplash says apps cannot use the Unsplash name directly in the application name. The extension title is simply “Unsplash”.
- User-created developer app requirement: Unsplash says applications should not require users to register for a developer account to use the app. The README instructs users to create an app on the Unsplash developer page and enter their access/secret keys.
- Download endpoint tracking: Unsplash says apps must call the photo.links.download_location endpoint when users perform something download-like. The action code appears to pass direct photo.urls values into copy/download/wallpaper actions.

Do you think we can tackle these accordingly?

Reference:
- https://help.unsplash.com/en/articles/2511245-unsplash-api-guidelines

eggsy:
Hey @0xdhrv,

Thanks for bringing these up! I took a look at each one:

**Name usage:** Fair point, though I'd say we're in a gray area here. The extension is "Raycast: Unsplash" in the store, so it has platform context. But I get why they'd want to be strict about it. We could rename to "Unsplash for Raycast" if you think it's worth it.

**Developer account requirement:** This is actually intentional. Unsplash's public API limits apps to 50 requests/hour. If we bundled a shared key, all our users would be competing for those same 50 requests. By having each user register their own app, everyone gets their own quota. It's better UX overall since no one gets rate-limited by other users. The README touches on this, but we could probably explain it more clearly.

**Download endpoint tracking:** This one we should definitely fix. I checked the code and we're currently passing direct photo URLs from the API response. Unsplash provides a `photo.links.download_location` endpoint that we should be calling whenever users download, copy, or set wallpaper. 

I can put together a PR for that if you'd like.

pernielsentikaer:
We can't even use "Unsplash" according to the ToS. If we change the logo/name, then we can get a real account, and limits should be more fair @eggsy — but I agree, it's a bit weird

xmok:
Hey all:

We could probably do something like I did for the Dropbox extension. They also did not allow me to use "Dropbox" in the title so I retitled it as [The Blue Cloud](https://www.raycast.com/chyroc/the-blue-cloud) with a different logo and they approved my OAuth account increasing limits similarly to Unsplash.

So it's doable!

Best.

EDIT: Also worth adding that they were not OK with "dropbox" in the Raycast Store URL (slug) so Per helped rename that from **dropbox** to **the-blue-cloud** as well.

eggsy:
From what I can tell from the https://help.unsplash.com/en/articles/2511257-guideline-replicating-unsplash guidelines, they specifically don't allow standalone wallpaper applications. Since this is a Raycast extension, not a wallpaper app itself, I think we're probably okay. You could even argue that, because Raycast is more of a developer/productivity platform, Unsplash might be be okay with users building their own extensions on top of it.

The bigger concern is getting a production app. We can't ship the access key and secret key with the extension, so we'd need to run a proxy somewhere to keep them private. That's doable.

That said, even with a production app, we'd still only get 5,000 requests per hour. For an extension with 30k+ users, that could become a problem pretty quickly. Sure, not everyone will be using it at the same time, but the rate limit is shared across everything: searches, image details, downloads, likes, and everything else. As the extension grows, we could easily hit the limit, making both the proxy and the extension unreliable.

I'll push the changes for the download endpoint and UTM parameters. We can continue the discussion there, and maybe follow @xmok's suggestion or we could try reaching out to Unsplash to see if they can officially approve or support this use case.

Kudos @pernielsentikaer.

PR #28915, opened 2026-06-20
Orion: add Arc-style Command Bar (unified palette)

giannisanni:
## Description

Adds a **Command Bar** to the Orion extension — an Arc-style unified palette that searches your open tabs, bookmarks, reading list, history, and the web from a single command.

> **Dependency note:** This builds on #28911 (the "Search Tabs returns no open tabs" fix), because the palette reuses the fixed tab access. Until #28911 is merged its commit also appears in this PR's diff — happy to rebase so this shows only the Command Bar changes once #28911 lands.

### Command Bar

A new `Command Bar` command (the existing five commands are unchanged). It runs as a dynamic list whose order the extension controls:

1. **Top Hit** — best local match (prefers an open tab → bookmark → reading list → history)
2. **Search the Web** — `Search <engine> for "<query>"`
3. **Suggestions** — live search-engine autocomplete (debounced)
4. **Open Tabs**, **Bookmarks**, **Reading List**, **History** — per-source sections

- Empty query shows open tabs (instant tab switcher).
- Results **open in Orion**, not the system default browser.
- **Profile switcher** in the search bar (reuses the existing `ProfileDropdown`). History uses `useSQL` and needs Full Disk Access; if it's missing, the History section is omitted rather than blocking the whole palette.
- New **Search Engine** preference (DuckDuckGo default; Google, Brave, Kagi). DuckDuckGo/Google/Brave supply autocomplete via their OpenSearch endpoints; Kagi has no public endpoint, so it gets the web fallback only.

### Auto-close launcher tabs (optional, preference-gated)

If you point Orion's homepage / new-tab at the Command Bar deeplink (`raycast://…/command-bar`) so a new tab opens the palette, Orion leaves a blank tab behind. When you act on a result, the Command Bar closes those blank tabs (identified by the `raycast://` URL scheme). Controlled by the **Auto-close launcher tabs** preference (on by default); it only ever closes `raycast://` tabs and never a window's last tab.

### Verification

- `ray build` and `ray lint` pass.
- Tested live against Orion 1.0.8 (147.1): the palette lists tabs/bookmarks/history, suggestions load, results open in Orion, and the profile switcher scopes the file-based sources.

### Notes

- Also hardens `useReadingList` against a missing `reading_list.plist` (returns empty instead of throwing).


0xdhrv (approved):
Looks good to me, approved ✅

PR #28925, opened 2026-06-21
feat(google-calendar): Narrow OAuth scope for least-privilege access

Asafffff:
## Summary

Narrows the Google Calendar OAuth scope from `calendar` (full read/write access to all calendar resources) to three least-privilege scopes:

| Scope | Grants | Used by |
|-------|--------|---------|
| `calendar.events` | View and edit events | `events.list`, `events.insert`, `events.get`, `events.update`, `events.delete` |
| `calendar.calendarlist.readonly` | View subscribed calendars | `calendarList.list` |
| `calendar.freebusy` | View availability | `freebusy.query` |

## Changes

- **`src/lib/google.ts`** — Replace `calendar` scope with the three narrower scopes. Remove `getAutoAddHangouts()` (dead code — used `settings.get`, never called). Remove `useCalendar()` hook (used `calendars.get` for conference properties, which are already available from `calendarList.list`). Clean up unused `withCache` import.
- **`src/create-event.tsx`** — Derive `allowedConferenceSolutionTypes` from the already-fetched calendar list via `useMemo`, replacing the removed `useCalendar` hook. This also eliminates a redundant API call.
- **`package.json`** — Add contributor.
- **`CHANGELOG.md`** — Add changelog entry.

## ⚠️ Note

Existing users will need to **re-authorize** due to the scope change. This is a one-time prompt handled automatically by Raycast's OAuth flow.


0xdhrv (commented):
Thanks for the update here. I initially thought narrowing the scopes could be useful, especially for Google Workspace environments where admins may be stricter about approving apps with broader permissions.

That said, after thinking it through, I agree that we should probably avoid changing this unless the current scopes are actually causing a problem. Since the scopes seem to work as expected today, narrowing them could make the extension fail later when new commands need additional access. It would also force all existing users to re-authenticate, which is a pretty big impact given this is one of the most downloaded extensions.

So unless there’s a concrete issue with the current permissions, I think it’s better to keep the scopes as they are.

cc: @pernielsentikaer

asafkmeta:
> ly thought narrowing the scopes could be useful, especially for Google Workspace environments where admins may be stricter about approving apps with broader permissions.
> 
> That said, after thinking it through, I agree that we should probably avoid changing this unless the current scopes are actually causing a problem. Since the scopes seem to work as expected today, narrowing them could make the extension fail later when new commands need additional access. It would also force all existing users to re-authenticate, which is a pretty big impact given this is one of the most downloaded extensions.
> 
> So unless there’s a concrete issue with the current permissio

@0xdhrv 
Thanks for reviewing this.

I'm submitting this as a follow-up to the security assessment required for onboarding the extension to my organization's Google Workspace - exactly as mentioned. I understand that this may cause friction for existing extension users, but from a security standpoint, the principle of least privilege is essential given the extension's current state. If a future command requires additional access, we can reassess the permissions and have users review them again. Currently, the permissions allow for the potential addition of new commands (and possible security vulnerabilities) without users being aware of the access pattern

0xdhrv:
> > ly thought narrowing the scopes could be useful, especially for Google Workspace environments where admins may be stricter about approving apps with broader permissions.
> > That said, after thinking it through, I agree that we should probably avoid changing this unless the current scopes are actually causing a problem. Since the scopes seem to work as expected today, narrowing them could make the extension fail later when new commands need additional access. It would also force all existing users to re-authenticate, which is a pretty big impact given this is one of the most downloaded extensions.
> > So unless there’s a concrete issue with the current permissio
> 
> @0xdhrv Thanks for reviewing this.
> 
> I'm submitting this as a follow-up to the security assessment required for onboarding the extension to my organization's Google Workspace - exactly as mentioned. I understand that this may cause friction for existing extension users, but from a security standpoint, the principle of least privilege is essential given the extension's current state. If a future command requires additional access, we can reassess the permissions and have users review them again. Currently, the permissions allow for the potential addition of new commands (and possible security vulnerabilities) without users being aware of the access pattern

Hey @Asafffff 👋 

We tested this and it seems like it’s working as expected, without the log-out issue.

0xdhrv (approved):
Looks good to me, approved ✅

PR #28947, opened 2026-06-22
Orion: add README and credit contributor

giannisanni:
Follow-up to #28915.

- Adds a **README** for the Orion extension (it had none, so the store page only showed the one-line description). It documents the commands, the preferences, and how to use the **Command Bar** as Orion's new-tab page (Arc-style) via the deeplink.
- Adds `gianni_sanrochman` to `contributors` (authored the Command Bar in #28915).


0xdhrv (approved):
Looks good to me, approved ✅

Issue #18630, opened 2025-04-16
[Image Modification] ...


michaelrubloff:
### Extension

https://www.raycast.com/HelloImSteven/sips

### Raycast Version

1.96.0

#### Error:
```
Error: Command timed out after undefined milliseconds: osascript
Timed out

Timeout.<anonymous>:exec-utils.js:32:33

---
29:     const timeoutPromise = new Promise((_resolve, reject) => {
30:         timeoutId = setTimeout(() => {
31:             spawned.kill("SIGTERM");
32:             reject(Object.assign(new Error("Timed out"), { timedOut: true, signal: "SIGTERM" }));
33:         }, timeout);
34:     });
35:     const safeSpawnedPromise = spawnedPromise.finally(() => {
---

listOnTimeout:timers:594:17
process.processTimers:timers:529:7
```

### Steps To Reproduce

convert any image to webp or avif doesn't work

Issue #17617, opened 2025-03-06
[Image Modification] Unable to optimize PNG files


jpaulgale:
### Extension

https://www.raycast.com/HelloImSteven/sips

### Raycast Version

1.93.1

### macOS Version

15.2

### Description

Hitting below error.

### Steps To Reproduce

Attempt to optimize a PNG file using "Optimize Images"

### Current Behaviour

Provides error 

```bash
Command failed with exit code 1: osascript
343:348: execution error: Can’t make file "{{YOUR_FILE PATH}}" into type alias. (-1700)```

### Expected Behaviour

properly optimized file

PR #28959, opened 2026-06-23
[Image Modification] Fix Optimize command failing when the assets path contains spaces

Aayush9029:
## Description

The **Optimize** command never works for installed users on PNG, GIF, BMP, TGA, WebP, and AVIF images. JPEG and SVG are unaffected.

### Root cause

`optimizeOperation.ts` shells out to bundled binaries (`pngout`, `cwebp`, `avifdec`, `avifenc`) with the binary path interpolated **unquoted** into `execSync()`:

```ts
execSync(`chmod +x ${pngoutPath}`);
execSync(`${pngoutPath} -s${strategy} -y -force "${pngPath}" "${newPath}"`);
```

For installed extensions, `environment.assetsPath` resolves under `~/Library/Application Support/com.raycast.macos/...` — which contains a space. The shell splits on it, so `chmod` and the binary invocation fail and the whole command throws. It works in `ray develop` only because the dev checkout path usually has no space, which is why this slipped through.

The same unquoted pattern affected the `cwebp` chmod in `getWebPBinaryPath` (`utils.ts`), which is on the optimize-WebP path. JPEG (`NSBitmapImageRep` via AppleScript) and SVG (SVGO, in-process) don't shell out to a bundled binary, so they were never broken — matching user reports that "optimize does nothing" specifically on PNGs.

Related reports: #17617 (Unable to optimize PNG files), #18630 (webp/avif convert failing).

### Fix

Quote the bundled-binary paths everywhere the Optimize command uses them — consistent with the already-quoted `supportPath` in `stripEXIFOperation.ts`. One-line-per-call mechanical change, no behavior change beyond surviving spaces.

### Verification

Reproduced deterministically: with an `assetsPath` containing a space, the current unquoted `execSync` throws `command not found`; the quoted version runs `pngout` and writes the optimized file. `npx ray lint` and `npx ray build` (TypeScript check) both pass.

> Note: the same unquoted-binary pattern still exists in `removeBgOperation.ts` (potrace/cwebp/avif) and the `dwebp` chmod. Happy to fold those into this PR or open a follow-up — let me know your preference.

## Checklist

- [x] `npm run build` and `ray lint` pass
- [x] I added a CHANGELOG entry (`{PR_MERGE_DATE}` placeholder)
- [x] I checked that assets are properly linked

### DIFF
--- a/extensions/sips/CHANGELOG.md
+++ b/extensions/sips/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Image Modification Changelog
 
+## [Fix Optimize Command on Asset Paths Containing Spaces] - {PR_MERGE_DATE}
+
+- Fixed the Optimize command failing for PNG, GIF, BMP, TGA, WebP, and AVIF images when the extension's assets path contains a space (e.g. `~/Library/Application Support/...`). The bundled binary paths (`pngout`, `cwebp`, `avifdec`, `avifenc`) are now quoted before being passed to the shell.
+
 ## [Security Maintenance] - 2026-05-21
 
 - Updated the extension to address security advisories.
--- a/extensions/sips/CHANGELOG.md
+++ b/extensions/sips/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Image Modification Changelog
 
-## [Fix Commands on Asset Paths Containing Spaces] - {PR_MERGE_DATE}
+## [Fix Commands on Asset Paths Containing Spaces] - 2026-06-23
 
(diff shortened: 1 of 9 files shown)


0xdhrv (commented):
Hi @ 👋

Thanks for the contribution 🔥

---

Aayush9029:
addressed the greptile review 👍🏼

0xdhrv (approved):
Looks good to me, approved ✅

PR #28960, opened 2026-06-23
fix: keep easy ocr fallback text

### LINKED ISSUE #28934: [Easy OCR] "Failed to OCR the image!" when autodetect lands on a language with no traineddata installed, even on normal English screenshots
### Extension

https://www.raycast.com/Rafo94/easy-ocr

### Raycast Version

1.104.19

### OS Version

macOS Sequoia 15.7.7

### Description

**TL;DR:** With "Enable language autodetect" on, Easy OCR fails with "Failed to OCR the image!" on normal English screenshots.

I set up Easy OCR on a new Mac M4 Pro and kept getting "Failed to OCR the image!" on screenshots that have perfectly normal English text in them. The same extension worked on my old laptop. Tesseract itself is healthy (`5.5.2` from Homebrew, `arm64`), and running it by hand on the same image gives me clean text.

Image

After poking at it I found that turning OFF `"Enable language autodetect"` makes it work every time. So the failure is coming from the autodetect path.

Image

Here is what I think is happening, from reading the source:

With autodetect on, the extension runs a first OCR pass in my default language (eng), which succeeds, then runs the detected language through a second pass. The check that decides whether the detected language is usable is `isValidLanguage` in `src/utils.ts`, and it only checks the code against the hard-coded `languages` object in `src/lib/languages.ts`. It never checks whether Tesseract actually has that language file on disk.

A few codes in that list have no matching `.traineddata` file even after I install the full pack with `brew install tesseract-lang`. The pack ships Tagalog as `fil` and  Kurdish as `kmr`, and it does not include the Fraktur variants, so `tgl`, `kur`, `dan_frak`, `deu_frak`, `deu_latf` and `slk_frak` stay missing while the extension still treats them as valid. When the detector returns one of those, the second pass calls Tesseract with a language that has no file, Tesseract exits with "Error opening data file ... Failed loading language", `node-tesseract-ocr` rejects, and it lands in the single catch in `index.tsx` that shows "Failed to OCR the image!".

The two that actually bite are `tgl` and `kur`, because the bundled detector returns those for real text. `raycast-language-detector` (which wraps franc / tinyld /  languagedetect) hands back `tgl` for very ordinary short English words. "binary" and "path" both come back as Tagalog for me, which is a bit funny since those are labels on the extension's own settings screen, and that short kind of text is exactly what I was capturing. 

Worth calling out: setting the "Tesseract language" field to eng does not protect you,  since that field only drives the first pass. The second pass uses whatever the detector  guessed.

### Steps To Reproduce

  1. Install Tesseract with Homebrew. A default install (only eng/osd/snum) and a full `brew install tesseract-lang` both reproduce it.
  2. In Easy OCR settings, tick "Enable language autodetect" and leave "Tesseract language" as eng.
  3. OCR a short piece of English text that the detector reads as Tagalog. Single words like "binary" or "path" do it reliably.
  4. You get "Failed to OCR the image!". Untick autodetect, capture the same thing again, and it works.

### Current Behaviour

The OCR fails with "Failed to OCR the image!" and nothing is copied, even though the first English pass already produced correct text. It happens whenever the autodetected language has no installed traineddata file.

### Expected Behaviour

If the detected language has no installed data file, fall back to the default language (or just keep the first pass result) instead of failing the whole OCR. Checking the detected code against the real installed list from `tesseract --list-langs`, rather than the static `languages` map, would fix the root cause. Wrapping the second pass so a failure there falls back to the first result would also do it.

suzunn:
## Problem
Easy OCR can successfully read text with the configured fallback language, then fail the whole command when autodetect picks a language that Tesseract cannot load locally. That drops the first successful OCR result and shows a generic failure toast instead.

Fixes #28934.

## Solution
I added a cached installed-language check using `tesseract --list-langs` before running the autodetected second pass. If Tesseract cannot list languages, the extension keeps the previous behavior. If the second pass still fails, the command now keeps the first OCR result instead of failing the whole flow.

## Testing
- `npm ci --ignore-scripts`
- `npx eslint src\index.tsx src\utils.ts`
- `npx prettier --check src\index.tsx src\utils.ts CHANGELOG.md`
- `git diff --check`

I also tried the normal Raycast CLI paths. `npm ci` fails during `@raycast/api` postinstall on Windows because `bin/ray npm-post-install` is invoked through `cmd.exe`, and invoking `ray lint` through Git Bash downloads an x86 binary that Windows cannot execute. `npx tsc --noEmit` is blocked because the generated Raycast `Preferences` global type is not present outside the Raycast build flow.

### DIFF
--- a/extensions/easy-ocr/CHANGELOG.md
+++ b/extensions/easy-ocr/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Tesseract OCR Changelog
 
+## [Bug Fixes] - 2026-06-23
+
+- Kept the first OCR result when autodetected languages are not installed in Tesseract
+
 ## [Enhancements] - 2025-04-15
 
 - Added automatic language detection option with either Raycast AI, Franc, LanguageDetect or Tinyld
--- a/extensions/easy-ocr/package.json
+++ b/extensions/easy-ocr/package.json
@@ -1,95 +1,98 @@
 {
-    "$schema": "https://www.raycast.com/schemas/extension.json",
-    "name": "easy-ocr",
-    "title": "Easy OCR",
-    "description": "Use Tesseract OCR to extract text from screenshot",
-    "icon": "command-icon.png",
-    "author": "Rafo94",
-    "contributors": [
-        "quiknull"
-    ],
-    "categories": [
-        "Productivity"
-    ],
-    "license": "MIT",
-    "commands": [
-        {
-            "name": "index",
-            "title": "Area OCR",
-            "description": "Select screen area which you want to extract the text",
-            "mode": "no-view"
-        }
-    ],
-    "preferences": [
-        {
-            "name": "tesseract_path",
-            "type": "textfield",
-            "required": false,
-            "title": "Tesseract binary path",
-            "description": "Absolute path to your tesseract binary - check documentation if default path is not correct",
-            "placeholder": "Tesseract binary path",
-            "default": "/opt/homebrew/bin/tesseract"
-        },
-        {
-            "name": "autodetect_lang",
-            "title": "Autodetect language",
-            "label": "Enable language autodetect",
-            "description": "Enable auto detection fo language, Tesseract language will be used as a fallback language if autodetect fails or doesn't find suitable lang",
-            "type": "checkbox",
-            "required": true,
-            "default": false
-        },
-        {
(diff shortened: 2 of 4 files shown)


pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #28961, opened 2026-06-23
Update Bookface extension

chrismessina:
## Description

Update to the published **Bookface** extension, adding new commands and bringing it up to date with YC CLI 0.0.14.

- Added new dropdown filters for additional types exposed in YC CLI 0.0.14
- **Log out of YC** — new command that clears stored YC CLI credentials, with a confirmation prompt.
- **Update Required gate** — when the installed YC CLI is too old to run, the extension now detects the version-gate refusal and routes to an in-app screen that runs `yc update` for you, instead of surfacing a raw parse error.
- **Knowledge Base results** — YC Knowledge Base articles now render in Search (they were silently dropped before, as the type was unhandled).
- **Agent transparency** — Ask answers now show "What the agent did" (the searches the YC agent ran to ground its response), using the `tool_calls` the CLI returns.
- **CSV export** — with a type filter selected in Search, `⌘⇧E` exports that type's full matching set as CSV (`⌘⇧C` copies it), reaching every match rather than just the displayed page.
- **Reliability fix** — large searches (e.g. "stripe") previously returned no results because the CLI's stdout pipe was truncated at OS pipe-buffer boundaries when its process exited before draining. Output is now captured via a temp file so the full payload is read regardless of size.
- **Polish** — a "Check Again" action on signed-out screens (Raycast has no window-focus hook, so this is a one-keystroke re-check after authenticating), and an opt-in Verbose Logging preference for diagnostics (sensitive values redacted).

### Checklist

- [x] I ran `npm run lint` (and `npx tsc --noEmit`) — all pass
- [x] I checked that this update does not break existing commands or remove functionality
- [x] I updated the `CHANGELOG.md` following the [changelog conventions](https://developers.raycast.com/basics/prepare-an-extension-for-store#changelog)

### DIFF
--- a/extensions/bookface/CHANGELOG.md
+++ b/extensions/bookface/CHANGELOG.md
@@ -1,6 +1,16 @@
 # Bookface Changelog
 
-## [Initial Version] - 2026-06-08
+## [New commands and YC CLI 0.0.14 support] - {PR_MERGE_DATE}
+
+- Added a `Log out of YC` command that clears stored YC CLI credentials (with a confirmation prompt).
+- Detect when the YC CLI is too old to run and route to an in-app **Update Required** screen that runs `yc update` for you, instead of failing with a raw error.
+- Render Knowledge Base articles in search results (previously dropped silently).
+- Show what the YC agent searched ("What the agent did") beneath each Ask answer.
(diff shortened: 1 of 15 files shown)


0xdhrv (approved):
Looks good to me, approved ✅

PR #28981, opened 2026-06-24
ado-search: fix My Work Items settings not saving

stoumann:
## Description

Fixes a bug in the **My Work Items** command where the Setup form silently discarded the user's selections, so saving persisted the defaults.

### Symptom
Configuring the form (default project, states, types, default repository, default base branch), clicking Save (which shows a green "Settings saved" toast), then reopening Setup shows everything back at the defaults.

### Cause
Each field's `onChange` handler bailed out while option lists were still loading, e.g.:

```ts
const handleRepoChange = (next: string) => {
  if (!formInitialized || loadingRepos) return; // <- dropped the selection
  setDefaultRepo(next);
};
```

`getAllRepositories` scans every project, so `loadingRepos` stays true for a while. Selections made during that window were dropped, the controlled fields snapped back, and Save wrote the defaults.

### Fix
Keep only the `!formInitialized` guard (which prevents the initial mount from clobbering the loaded values — the dropdowns already render the current value as a fallback item). Allow `onChange` to update state while option lists are loading. Verified locally: the form now retains and persists selections.

## Checklist
- [x] `npm run lint` passes
- [x] Changes limited to `extensions/ado-search/`


0xdhrv (approved):
Looks good to me, approved ✅

PR #28984, opened 2026-06-24
Update wifi-password-reveal extension

0xdhrv:
## Description

Fixes intermittent `Command failed: netsh wlan show profile name="..." key=clear` errors on Windows.

**Root cause:** The previous implementation used Node.js `exec()`, which on Windows spawns `cmd.exe /d /s /c "..."`. When the command string already contains inner double quotes (e.g. `name="GAN CAFE - TANG 1"`), `cmd.exe`'s `/s` flag strips the outermost quote pair — which can cause it to mangle profile names that contain spaces, leaving `netsh` with a truncated or malformed name that it can't find. Additionally, `cmd.exe` uses the system's legacy ANSI code page, so SSIDs with Unicode characters would be decoded incorrectly.

**Fix:** Replaced both Windows `exec()` calls with `runPowerShellScript` from `@raycast/utils`. PowerShell:
- handles argument quoting correctly without the `/s /c` double-quote stripping issue
- outputs UTF-8/UTF-16 by default, so Unicode SSIDs are decoded correctly
- provides a clean `Promise`-based API, removing the nested callback pattern and making error/loading cleanup safer via `try/catch/finally`

Also hardened the SSID profile-name parser (`parseNetshWlanProfiles`) to use `indexOf` + `substring` instead of `split(": ")`, so network names that contain `": "` are no longer truncated.

## Screencast

_Bug fix — no new UI. The change is only observable on Windows when a saved network name contains spaces or special characters._

## Checklist


pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #28989, opened 2026-06-24
Add p00f-raycast extension

mcdays94:
## Description

**p00f** is a zero-knowledge, ephemeral clipboard. The sender encrypts content locally, shares a link, and the content poofs once its timer runs out or its reveal budget is used up. The hosted service at [p00f.me](https://p00f.me) only ever holds ciphertext.

This extension brings p00f into Raycast with three commands:

- **Create Poof** (view): full form. Text or one file, TTL presets + custom (1 minute to 30 days), Reveal budget presets + custom + unlimited, optional PIN (4 to 128 chars), secret-kind masking, masked URL mode, reveal-anchored TTL, viewer-delete, reveal captcha, and recipient countdown. The result detail exposes `Copy Link`, `Paste Link in Frontmost App`, `Copy Owner Token`, `Burn Now`, and `Open in Browser`.
- **Poof Selection** (no-view): turns the selected text into a Poof. When Finder is frontmost with exactly one file selected, shares the file bytes instead.
- **Poof Clipboard** (no-view): turns one clipboard file, plain text, or HTML-as-text fallback into a Poof.

### How it works

1. The extension generates a 32-byte master key on the user's Mac.
2. Content is encrypted locally via [`@p00f/core`](https://www.npmjs.com/package/@p00f/core) (AES-GCM-256).
3. Only ciphertext is uploaded to p00f.me. The Fragment Key stays in the URL fragment of the resulting Link and never reaches the server.
4. After the configured TTL or Reveal budget runs out, the Poof burns server-side.

### Privacy and integration notes

- Every Link and owner token is copied with `Clipboard.copy(..., { concealed: true })`, so neither ever enters Raycast Clipboard History.
- No Raycast or client-identifying headers are sent. The hosted service treats the extension as just another anonymous machine-path caller, with the same rate-limit floor as the `poof` CLI.
- Owner tokens are not persisted to disk by default.
- All preferences have sensible defaults (5 min TTL, 1 Reveal, no PIN, no reveal captcha). Every default is overridable per-Poof in the **Create Poof** form.

### Repository and ecosystem

- Source: [github.com/mcdays94/p00f](https://github.com/mcdays94/p00f), package at `packages/raycast/`.
- Core engine: [`@p00f/core@0.2.0`](https://www.npmjs.com/package/@p00f/core) on npm.
- Sister CLI: [`@p00f/cli`](https://www.npmjs.com/package/@p00f/cli) on npm.
- License: MIT.

This is the initial submission (`CHANGELOG.md` carries `[Initial Version]`).

## Screencast

Store screenshots live in `metadata/`. The 4 PNGs at 2000x1250 cover:

- Raycast root search showing all three commands with the brand icon.
- **Create Poof** form, top half (Text / File / TTL / Custom TTL).
- **Create Poof** form, bottom half (Reveal budget / Custom Reveals / PIN / six toggles).
- Result Detail after a real create, with the action panel open.

The Poof shown in the Result Detail screenshot was revealed-and-burned via the CLI immediately after capture, so the Link in that frame is dead by design.

## Checklist


mcdays94:
Thanks for the review, Greptile! Just pushed `0681d64` addressing #1, #2, and #3:

- **#1**: Added `$schema` to `package.json`.
- **#2**: `categories` is now `["Security", "Developer Tools"]`.
- **#3**: Removed the manual `PoofPreferences` interface; the three callers now use `getPreferenceValues<Preferences>()` against the auto-generated type in `raycast-env.d.ts`, and `createDefaultsFromPreferences` accepts `Preferences` directly.

Deliberately NOT applying **#4**. The `showCountdown` ternary matches the project-wide convention from [ADR-0014](https://github.com/mcdays94/p00f/blob/main/docs/adr/0014-countdown-and-autoclear.md). There is no server-side default to rely on; `showCountdown` is a client-side render hint folded into the encrypted metadata. The contract is: only stamp `showCountdown: false` when explicitly disabled, otherwise omit so the default-on path adds no metadata bytes.

This is locked across all three shells (web, CLI, Raycast) by [`test/core.test.ts:117`](https://github.com/mcdays94/p00f/blob/main/test/core.test.ts#L117) ("folds `showCountdown=false` into the encrypted metadata, default stays on"). The same pattern lives in:

- `src/shared/core.ts:85`: `if (input.showCountdown === false) meta.showCountdown = false;`
- `src/cli/poof.ts:165`: `const showCountdown = a.flags["no-countdown"] ? false : undefined;`
- `src/client/main.ts:731`: `if (!showCountdown) meta.showCountdown = false;`

Switching the Raycast form to `values.showCountdown ?? true` would stamp `true` into every default-on poof's encrypted metadata, diverging from web and CLI and breaking the "default needs no bytes" contract.

mcdays94:
Just pushed `d8b66f5` addressing Greptile's significant finding from the re-review: dropped the `npm run build --prefix ../core` prebuild from the `build` and `dev` scripts. The published extension now relies entirely on `@p00f/core@^0.2.0` from the npm registry (as listed in `dependencies`), and `ray build` runs standalone. Anyone cloning `raycast/extensions` and running `npm run build` from `extensions/p00f-raycast/` will succeed.

For the in-repo monorepo workflow (`mcdays94/p00f`), the README's "Local development" section now includes an explicit `npm run build --prefix packages/core` step that contributors run when they change core source.

Deferring the two minor style notes (toast title/message split in the no-view commands, fire-and-forget Promises in `result-detail.tsx`) as code polish rather than bugs. Happy to address them if you would prefer them in for the Store submission.

extensions/p00f-raycast/package.json
```diff
@@ -0,0 +1,187 @@
+{
+  "$schema": "https://www.raycast.com/schemas/extension.json",
+  "name": "p00f-raycast",
```
0xdhrv:
```suggestion
  "name": "p00f",
```

Can you please change the name to `p00f` rather than `p00f-raycast`?

extensions/p00f-raycast/package.json
```diff
@@ -0,0 +1,187 @@
...
+  "title": "P00f",
```
0xdhrv: also, for the title, something like


```suggestion
  "title": "P00f - Ephemeral Clipboard",
```

0xdhrv (changes requested):
Hey @mcdays94 👋 

I have added a comment for you to address. Also please change the folder name to `p00f` instead of `p00f-raycast`.

---

mcdays94:
Thanks for the review @0xdhrv! Pushed `c096800` with everything you asked for, plus a few fixes I wanted in before you test again.

**Your asks:**

- **Folder renamed** to `extensions/p00f/`.
- **`name`** is now `p00f`.
- **`title`** is now `P00f - Ephemeral Clipboard`.

**Also cleaned up while I was in here:**

- **Standard lint toolchain added.** The extension was missing `@raycast/eslint-config`, `eslint`, `prettier` and `@types/react`, so `ray lint` was printing `ESLint is not installed, skipping formatting check` and only running Prettier. Added those plus an `eslint.config.js` and a `fix-lint` script. The config is ESM rather than the usual CommonJS one, because this package is `"type": "module"`:

  ```js
  import { defineConfig } from "eslint/config";
  import raycast from "@raycast/eslint-config";

  export default defineConfig([...raycast]);
  ```

- **Burn Now left a dead link on screen.** Greptile flagged this on the first pass and I deferred it. After a successful burn the Detail view kept showing the live-looking link with every action still enabled, including a second Burn Now. It now switches to a burned state with the actions removed, so there is no way to hand out a link that no longer resolves.

- **Fire-and-forget action handlers.** The `onAction` callbacks returned promises that were dropped, so a failure became an unhandled rejection with no user feedback. All of them are now wrapped and surface a Failure toast.

- **Toast title/message split** in the two `no-view` commands. They were putting the raw error text in the toast title instead of the message.

**Verification:**

- Clean `npm ci` then `ray build` and `ray lint` at the new path, both green.
- 185 unit tests green in the upstream repo.
- An 18-check create -> info -> reveal -> burn round-trip against the live service, covering plain text, PNG bytes (the screenshot path), reveal-budget exhaustion, owner-token burn, wrong-PIN rejection, and reveal-anchored TTL arming. All passing.

Marking as ready for review. Happy to iterate on anything else.

extensions/p00f-raycast/package.json
mcdays94: Done in `c096800` — `name` is now `p00f`, and the folder moved to `extensions/p00f/` to match.

extensions/p00f-raycast/package.json
mcdays94: Done in `c096800` — `title` is now `P00f - Ephemeral Clipboard`. Passes `ray lint`'s Title Case check.

0xdhrv (approved):
Looks good to me, approved ✅

PR #28994, opened 2026-06-24
Add [REST Countries] extension

yusifaliyevpro:
## Description

This extension adds the REST Countries API, making it directly queryable in Raycast.

## Checklist


yusifaliyevpro:
@0xdhrv Ready for merge

0xdhrv (commented):
Thanks for your contribution  

- [country-lookup](https://www.raycast.com/pernielsentikaer/country-lookup) by @pernielsentikaer 

yusifaliyevpro:
@0xdhrv Thanks for review. Yes I have seen it. But it no longer works because v3.1 API endpoints are disabled. And this extension now has AI features. What do you think?

yusifaliyevpro:
@0xdhrv Also, I would need to change `country-lookup` extension commands, which is not good for current users of Country Lookup, there will be conflicts. So I decided to create a new extension instead.

0xdhrv:
> @0xdhrv Thanks for review. Yes I have seen it. But it no longer works because v3.1 API endpoints are disabled. And this extension now has AI features. What do you think?

You can update the existing extension with the same features since the service remains unchanged.

0xdhrv:
> @0xdhrv Also, I would need to change `country-lookup` extension commands, which is not good for current users of Country Lookup, there will be conflicts. So I decided to create a new extension instead.

As far as improvements go, this can go as update

PR #28997, opened 2026-06-24
Add sir-golf extension

thejulez:
sir.golf — live PGA Tour / LPGA / DP World Tour leaderboards, the full tour
schedule with every major badged, one-tap Add to Calendar, and season rankings.
$0, no account, no API key, no backend — fetches ESPN's free public JSON API
client-side and caches via Raycast.

- Leaderboard — live tournament (graceful fallback to most-recent; 30s auto-refresh)
  plus season rankings; per-player detail pane.
- Golf Season — schedule with majors badged, filter by tour or by where it
  streams (incl. 🇦🇹 Austria), and Add to Calendar (⌘⇧A).

By the sir.golf brand (https://sir.golf).


thejulez:
Thanks for the review! Pushed a follow-up commit (`3f612e7`) addressing all five points:

1. **Stale ICS `DTSTAMP`** — now set to the actual creation time per RFC 5545 (`new Date().toISOString()…`), instead of the fixed past constant, so calendar apps no longer risk silently dropping a re-imported event.
2. **Score color semantics** — swapped to follow Raycast's convention: under par → `Color.Green`, over par → `Color.Red` (even par stays secondary). The golf-broadcast palette was deliberate but agreed it reads as an error state in Raycast.
3. **Manual `Prefs` interface** — removed; now using the generated `Preferences` type from `raycast-env.d.ts`.
4. **Unused `flag` prop** — removed from `PlayerDetailPane` (and its call site). The country flag already renders as the row's leading icon, so the prop only produced a dangling separator.
5. **Dev-only artifacts** — removed `spike/`, `CLAUDE.md`, and `VALIDATION.md` from the submission, and cleaned up the README references to them.

`npm run build` passes clean. Ready for another look whenever you get a chance 🙏

0xdhrv (approved):
Looks good to me, approved ✅

PR #29003, opened 2026-06-24
Fix ShareX library path handling

### LINKED ISSUE #28975: [ShareX] Use exact screenshots folder path instead of appending /screenshots/
### Extension

https://www.raycast.com/the_revolution/sharex

### OS Version

Win 11

### Description

I would like the ShareX extension setting for the screenshots folder path to use the exact path provided by the user.

Currently, when I set a folder path, the extension appears to append `/screenshots/` to it automatically. This causes the extension to look in a different folder than the one I configured.

For example, if I set:
```
C:\MyScreenShotsFolder\
```

the extension should use exactly that path.

Instead, it seems to use something like:
```
C:\MyScreenShotsFolder\screenshots
```

The requested behavior is that the configured path should be treated as the final screenshots folder path, without adding any extra subfolder automatically.

This would make the extension work better with custom ShareX folder structures, especially for users who changed the default ShareX capture location.

Users who use a custom ShareX screenshots folder or organize their ShareX captures differently from the default folder structure.

This would also help users who sync screenshots with folders like OneDrive, Google Drive, Dropbox, or custom project folders, where the exact configured path matters.

### Anything else?

_No response_

mralonsocorona:
## Summary

Fixes the ShareX Browse Library command so the configured `screenshotsPath` is treated as the final screenshots folder.

Previously, the command only scanned subfolders inside the configured path. This meant files stored directly in a custom ShareX screenshots folder were ignored.

## Changes

- Add support for media files located directly inside the configured screenshots folder.
- Keep the existing behavior that scans recent subfolders, preserving support for organized ShareX folder structures.
- Reuse a shared media-file regex for direct files and subfolder files.
- Rename the resolved path variable from `basePath` to `screenshotsPath` for clarity.

## Validation

- `npm run lint`
- `npm run build`

Fixes #28975

### DIFF
--- a/extensions/sharex/CHANGELOG.md
+++ b/extensions/sharex/CHANGELOG.md
@@ -1,6 +1,6 @@
 # ShareX Changelog
 
-## [Fix] - {PR_MERGE_DATE}
+## [Fix] - 2026-06-25
 - Include media files stored directly in the configured screenshots folder.
 
 ## [Feature] - 2026-05-05
--- a/extensions/sharex/CHANGELOG.md
+++ b/extensions/sharex/CHANGELOG.md
@@ -1,5 +1,8 @@
 # ShareX Changelog
 
+## [Fix] - 2026-06-24
+- Include media files stored directly in the configured screenshots folder.
+
 ## [Feature] - 2026-05-05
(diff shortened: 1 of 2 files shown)


pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #29029, opened 2026-06-25
[Reclaim] Fix duplicate recurring sync events on calendar view

ajohnson1200:
## Description

Fixes a bug where a recurring event synced across multiple calendars shows up as duplicates in the **My Calendar** view instead of being merged down to a single entry.

`filterMultipleOutDuplicateEvents` builds a lookup `Set` of existing event ids keyed by `recurringEventId ?? eventId`, then for each sync event it decodes the `reclaim-personal-sync:<sourceId>:…` id and checks `ids.has(sourceId)` to drop the duplicate.

The problem: for a **recurring** source event, `recurringEventId` is the bare series id (e.g. `0e9aualupvvi2j0k6mst8cc8cq`), so that's what lands in the Set. But the sync copy's decoded id embeds the source's **full instance id** including the recurrence suffix (e.g. `0e9aualupvvi2j0k6mst8cc8cq_20260625T200000Z`). The lookup therefore misses and the duplicate sync event leaks onto the calendar. Non-recurring events are unaffected because they fall back to `eventId` (the full id), which is why this only reproduces on recurring events.

The fix indexes the `Set` by **both** `eventId` and `recurringEventId`, so the embedded full-instance id always matches.

## Screencast / Testing

Verified against three cases:
- recurring source + sync copy → 1 event kept (was 2 — the bug)
- non-recurring source + sync copy → 1 event kept (unchanged)
- orphan sync event with no source present → 1 event kept (not incorrectly dropped)

## Checklist

- [x] I added a `CHANGELOG.md` entry
- [x] I ran the change locally and verified the behavior

### DIFF
--- a/extensions/reclaim-ai/CHANGELOG.md
+++ b/extensions/reclaim-ai/CHANGELOG.md
@@ -1,6 +1,6 @@
 # reclaim Changelog
 
-## [Fixes] - {PR_MERGE_DATE}
+## [Fixes] - 2026-06-25
 - Fix duplicate recurring events showing on the calendar when synced across multiple calendars
 
 ## [Update] - 2026-05-27
--- a/extensions/reclaim-ai/CHANGELOG.md
+++ b/extensions/reclaim-ai/CHANGELOG.md
@@ -1,5 +1,8 @@
 # reclaim Changelog
 
+## [Fixes] - {PR_MERGE_DATE}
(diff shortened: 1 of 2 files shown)


gunrein (approved):

pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #29030, opened 2026-06-25
Fix Google Tasks create task success handling

### LINKED ISSUE #28900: [Google Tasks] "Create Task" command fails
### Extension

https://www.raycast.com/elliotdes/google-tasks

### Raycast Version

0.65.1.0

### OS Version

Windows Windows 10 Core 10.0.19045.0 amd64

### Description

The extension has access to my Google account evidenced by the fact that the "View Lists" command shows my task list and its contents. 

But the "Create Task" command doesn't work. I get the green "Task Created!" message in Raycast when I make a new task using "Create Task", but the tasks I enter never actually show up in my list.

### Steps To Reproduce

1. Use the "Create Task" command.
2. Fill out the form, hit the "Create Task" button.

### Current Behaviour

3. Raycast window shows green "Task Created!" message.
4. The task created never shows up on my task list.

### Expected Behaviour

3. Raycast window shows green "Task Created!" message.
4. The task shows up on my task list.

mralonsocorona:
## Summary

Fixes the Google Tasks Create Task command so it only shows the success toast and closes the form after the task creation request has completed successfully.

Previously, the form called the async `onCreate` handler without awaiting it, then immediately showed `Task Created!` and navigated back. If the Google Tasks API request failed, users could still see a success message even though the task was not created.

## Changes

- Make the create form `onCreate` contract async.
- Await task creation before showing the success toast.
- Keep the form open and show a failure toast if creation fails.
- Refresh the task list after successful creation from list views.
- Add a changelog entry.

## Validation

- `npm run lint`
- `npm run build`

Fixes #28900

### DIFF
--- a/extensions/google-tasks/CHANGELOG.md
+++ b/extensions/google-tasks/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Google Tasks Changelog
 
-## [Fix] - {PR_MERGE_DATE}
+## [Fix] - 2026-06-26
 
 - Await task creation before showing success and closing the create task form.
 
--- a/extensions/google-tasks/CHANGELOG.md
+++ b/extensions/google-tasks/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Google Tasks Changelog
 
+## [Fix] - {PR_MERGE_DATE}
+
+- Await task creation before showing success and closing the create task form.
+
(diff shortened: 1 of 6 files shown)


0xdhrv (approved):
Looks good to me, approved ✅

PR #29063, opened 2026-06-27
Add latex-ocr-queen extension

mejistus:
## Checklist


0xdhrv (commented):

We already have an extension in the Store that covers this functionality. Could we consider enhancing the existing extension below instead of creating another one?

* [SimpleTexOCR](https://www.raycast.com/Arthals/simpletexocr) by @zhuozhiyongde

Both extensions provide the same core workflow: capture an image containing a mathematical formula, recognize it using an OCR or vision service, convert it to LaTeX, and copy the result to the clipboard. Both also support raw LaTeX, inline math, and display/block math output formats.

This PR adds some useful flexibility, including support for multiple OpenAI-compatible vision providers and the ability to process images from Finder or the clipboard. However, these feel like natural enhancements to the existing LaTeX OCR extension rather than enough differentiation for a separate Store extension. Adding configurable providers and additional image sources to SimpleTexOCR would preserve the new workflows while keeping formula-to-LaTeX functionality consolidated and easier to discover.

PR #29066, opened 2026-06-27
Add contrast-lab extension

fracazo:
## Description
Contrast Lab is a new extension for checking color contrast. A single command takes a foreground and background color, plus optional font size and weight, and reports readability against both standards:

WCAG 2 ratio with AA/AAA pass/fail for normal and large text.
APCA (the WCAG 3 candidate model), with the Lc value and a pass/fail based on the chosen font size and weight.

It also suggests the nearest passing color: when a pair fails AA, it finds the closest foreground that passes by adjusting lightness in OKLCH (preserving hue and chroma), so the fix still looks like the chosen color rather than snapping to black or white. The result view shows a live preview swatch of text on background for both the original and the fix, with copy actions for hex and OKLCH.
Inputs accept HEX, RGB, HSL, and OKLCH.
Why this extension: the existing color tools in the store compute WCAG 2 contrast only. None support APCA, which accounts for font size and weight and is increasingly used in modern design systems. This fills that gap, and pairs it with a live preview and a one-tap fix rather than just a number.
The contrast math is a pure, framework-free library with unit tests covering both algorithms and the fix search, verified against reference values.
## Screencast

contrast-lab-1
contrast-lab-2

## Checklist


fracazo:
Thanks, all addressed in commit: ee5f858617c4f621ee5797c5a53e8c5d4700a92b:

 WCAG AA/AAA now compare the raw (unrounded) ratio; translucent colours are composited over the background before scoring; removed unused @raycast/utils and colorparsley; invalid font sizes fall back to the default; and fixed the npm test glob so the suite is reliably discovered (27 tests).

fracazo:
> Congratulations on your new Raycast extension! 🚀
> 
> We're currently experiencing a high volume of incoming requests. As a result, the initial review may take up to 15 business days.
> 
> Once the PR is approved and merged, the extension will be available on our Store.

15 business days are long past @raycastbot can you assign a human to review this PR?

0xdhrv (commented):

We already have an extension in the Store that covers color contrast calculations. Could we consider enhancing the existing extension below instead of creating another one?

* [Polished](https://www.raycast.com/lekoarts/polished) by @LekoArts

Both extensions let users enter foreground and background colors and evaluate whether the combination meets accessibility contrast requirements. The existing **Contrast Ratio** command currently calculates the WCAG contrast ratio and reports AA/AAA compliance.

This PR adds several useful capabilities beyond the existing implementation, including APCA calculations, font size and weight considerations, support for RGB, HSL, and OKLCH input, live previews, and suggestions for the nearest color that passes WCAG AA. However, these features feel like natural enhancements to the existing contrast command rather than a separate extension, especially since Polished is already positioned as a broader collection of color conversion, modification, readability, and contrast utilities.

If there are additional unique workflows you’re aiming to add, we’d love to hear them and see if they can be integrated into Polished to avoid duplication and improve discoverability.

fracazo:
Thanks for the thorough review! Fair concern about duplication. I'd like to make the case that this is different value rather than a second version of the same command, and the core of it is maintenance scope.

**Polished is a wrapper around the [`polished`](https://polished.js.org/) library.** Its only runtime dependency is `polished`, and the Contrast Ratio command exposes what that library provides: the WCAG 2 ratio and AA/AAA guidelines. Contrast Lab runs on a different engine (`apca-w3`, `culori`, `colorparsley`): signed APCA Lc with font size/weight thresholds, OKLCH parsing with sRGB gamut mapping, and a lightness search that finds the nearest passing color. None of that exists in `polished`, so merging wouldn't be enhancing the existing command, it would be grafting a parallel contrast engine onto Polished that @LekoArts would then be on the hook to maintain. That seems like the wrong outcome for a deliberately lightweight extension.

**They also answer different questions.** Polished answers "what's the ratio of these two colors" as one of several color utilities. Contrast Lab answers "will this text actually be readable, and if not, what's the closest color that fixes it". It's an accessibility evaluation tool rather than a color toolbox, which is also where the roadmap goes: clipboard auto-prefill and a palette picker are next, building toward palette-wide checking (a pass/fail matrix of every foreground/background pair in a palette across both models). That workflow would sit oddly inside a general color-utils extension.

That said, I get the discoverability concern and don't want to force a duplicate into the Store. Would it make sense to get @LekoArts' take? If they'd like to absorb this scope into Polished, I'm happy to contribute it there instead. If it's outside what they want to maintain, I'd love to keep Contrast Lab standalone as the contrast/accessibility-focused option.

extensions/contrast-lab/metadata/contrast-lab-1.png
0xdhrv, line 1: **issue**: metadata image style

You need to use **"Capture Window"** command to take a screenshot of the extension

Reference:

extensions/contrast-lab/metadata/contrast-lab-2.png
0xdhrv, line 1: **issue**: metadata image style

You need to use **"Capture Window"** command to take a screenshot of the extension

Reference:

0xdhrv (changes requested):
Hey @fracazo 👋 

---

fracazo:
@0xdhrv Both metadata images were retaken with the Window Capture command, so they now use the standard background at 2000x1250, and the local extension indicator is gone from the bottom bar. 

Thanks for the pointer to the docs!

0xdhrv (approved):
Looks good to me, approved ✅

PR #29094, opened 2026-06-29
Add flypy extension

rxbit:
## Description

Lookup Xiaohe encoding for Chinese characters

## Screencast

image

## Checklist


0xdhrv (commented):
This PR overlaps directly with two existing Store extensions that already cover Xiaohe/Flypy code lookup. ([Raycast][1])

We already have extensions in the Store that cover this functionality. Could we consider enhancing one of the existing extensions below instead of creating another one?

* [Xiaohe Query](https://www.raycast.com/dongqing/xiaohe-query) by @SysDQ
* [Flypy Code Query](https://www.raycast.com/absir1949/flypy) by @absir1949

All three extensions provide the same core workflow: enter one or more Chinese characters and look up their Xiaohe/Flypy encoding. The proposed extension also displays character decomposition, pinyin, Unicode information, and links to external dictionaries, but these feel like natural additions to the existing **Xiaohe Query** extension, which already focuses on Xiaohe codes, decomposition, and expanded character information.

The local bundled dictionary and support for querying multiple characters could also be useful improvements to the existing extension rather than grounds for another standalone extension with nearly identical functionality.

PR #29099, opened 2026-06-29
Add downdetector extension

MagnasiePro:
## Checklist


MagnasiePro (commented):
Thanks for the review — all three P1 findings are fixed in `8a018956e`. They shared one root cause: the **Region** preference (`locale`) was not part of the invalidation for the `usePromise` hooks or the persisted history URLs, so changing Region while a command stayed mounted kept serving the previous Downdetector site.

**1. `search-service.tsx` — Regionless Promise Cache**
Search hook deps are now `[query, retryCount, locale]` (with `const locale = getLocale()`), so a Region change revalidates instead of returning the previous region's result.

**2. `service-detail.tsx` — Slug-Only Detail Cache**
Detail hook is now keyed `[slug, locale]`. Metadata link, markdown link, browser action and report fallback all derive from region-aware `detail.url` / `getStatusUrl(slug)`.

**3. `use-search-history.ts` — Persistent Regional History URL**
History no longer persists the region-specific `url` (`HistoryItem` = `Pick<Service, "slug" | "name" | "status">`); the browser action recomputes it from the slug via `getStatusUrl(item.slug)` at use-time, so cross-region history always opens the current site.

Also added `getLocale()` in `api.ts` as the single source of truth for the region (base URL, locale paths, and the search/detail cache keys). Verified: `tsc --noEmit`, `npm run lint`, and `npm run build` all pass.

0xdhrv (approved):
Looks good to me, approved ✅

PR #29103, opened 2026-06-29
Update wojak-picker extension

itsMeOnli:
## Description

Update to the existing **Wojak Picker** extension. This changes how images are hosted:

- **Images now load from a public jsDelivr git CDN** (backed by a GitHub asset repo) instead of hotlinking the source site. This directly addresses the earlier review suggestion to support alternate/git-CDN hosting.
- **Pre-generated, compressed thumbnails** (320px, palette-quantized) for a faster, lighter grid.
- **Zero configuration.** The CDN base URL is a constant in source — there are no preferences, accounts, or API keys to set up. The extension works immediately on install.
- **Offline fallback.** If the CDN is unreachable, the last synced library is served from cache and an offline indicator appears in the navigation title.

No new commands or permissions; the search/copy UX is unchanged.

## Checklist


0xdhrv (commented):
Hi @itsMeOnli 👋

Thanks for the contribution 🔥

Also, since this is self-hosted, it seems a bit too complicated for nontechnical users to set up. Can you find an alternative way for the startup to work better?

---

itsMeOnli:
Hi @0xdhrv 

I'll be sure to implement the suggestions from greptile and think of ways to make it simpler. 

Shall mark it as ready once the changes are in place :)

0xdhrv:
> Hi @0xdhrv
> 
> I'll be sure to implement the suggestions from greptile and think of ways to make it simpler.
> 
> Shall mark it as ready once the changes are in place :)

Sure, mark it as ready when you are

Thanks a lot for your contributions 🔥

itsMeOnli:
Thanks for the patience @0xdhrv 🙏

Reworked this to remove the configuration burden entirely.

**Zero setup.** The jsDelivr CDN URL is now a constant in source and the `Library Base URL` preference is gone. There's nothing for a user to configure, and the old "Setup required" dead-end (which a blank preference could trigger) is gone with it.

**Dropped the self-hosting surface.** `deploy/` (nginx config + guide), `scripts/build-deploy.js`, `scripts/deploy-vps.sh` and `.env.local.example` are no longer part of the extension — that tooling belongs to the asset repo backing the CDN, not to something a user or reviewer should have to read.

**Greptile fixes.**
- Restored the full `filteredWojaks` dependency array — this was a real bug: the grid stayed empty until you typed or switched category.
- Dropped the legacy `.eslintrc.json`; `eslint.config.mjs` (flat config) was already present, so the PR was adding a duplicate.
- Split `remoteError` into `loadError` and `isOffline`. It had been carrying two unrelated meanings, so a zero-result search while offline reported "Couldn't load wojaks" even though the library had loaded from cache. Offline state now shows in the navigation title instead of hijacking the search-results message.

**Metadata images.** Removed the three new screenshots that were failing the padding check. The three already on `main` still describe the UI accurately, since this PR doesn't change the UX.

Net effect: the PR is down to three files — `src/search-wojaks.tsx`, `README.md` and `CHANGELOG.md`. `package.json`, `package-lock.json` and `.gitignore` are byte-identical to `main` again (the lockfile diff turned out to be only `"peer": true` churn from a different npm version).

Ready for another look whenever you have time.

0xdhrv (approved):
Looks good to me, approved ✅

PR #29105, opened 2026-06-29
Fix TinyPNG multiple compression command

natsustan:
## Description

Fixes the TinyPNG `Compress Images Multiple Times` command by restoring the missing `node-fetch` import used for TinyPNG API requests.

Also adds `thuggyduck` to the extension contributors list and refreshes the npm lockfile metadata generated by the current npm version.

## Screencast

Not included because this is a small bugfix to an existing no-view command.

## Checklist

### Verification

- `npm run build`
- `npm run lint`
- `npm test`

### DIFF
--- a/extensions/tinypng/CHANGELOG.md
+++ b/extensions/tinypng/CHANGELOG.md
@@ -1,6 +1,6 @@
 # TinyPNG Changelog
 
-## [New Command] - 2026-06-29
+## [New Command] - {PR_MERGE_DATE}
 
 - Added `Compress Images Multiple Times` to run TinyPNG compression multiple times in one command.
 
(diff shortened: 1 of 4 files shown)


natsustan:
Thanks for the review. I checked the distribution build output for this specifically:

```bash
npx ray build -e dist -o .tmp-raycast-build
```

In `.tmp-raycast-build/compressImagesMultipleTimes.js`:

- `require("node-fetch")` occurrences: 0
- `node-fetch` package strings: 9
- bundled fetch internals (`fetch-blob`, `formdata-polyfill`, etc.): present
- bundle size: 422,672 bytes

So Raycast/esbuild is bundling `node-fetch` into the command output rather than leaving it as an external CommonJS require. This matches the existing `compressImages` command, which already imports `node-fetch` the same way.

0xdhrv (approved):
Looks good to me, approved ✅

PR #27908, opened 2026-05-15
Add 2fas-authenticator extension (v1.1.0)

LockeAG:
## Description

**2FAS Authenticator**. Search and copy TOTP codes from [2FAS](https://2fas.com) exports directly in Raycast.

2FAS is a great mobile authenticator but has no desktop app. This extension bridges that gap by importing your 2FAS encrypted export and generating TOTP codes locally. No cloud, no network calls.

> **Note on history:** a previous PR (#26126) was auto-closed by the stale-bot before reaching review. This resubmission is **v1.1.0** with significant hardening over the original 1.0.0. See `CHANGELOG.md` for the full list. Highlights below.

### Commands

| Command | Description |
|---------|-------------|
| Search OTP | Search all services with native Raycast filtering and live TOTP countdown |
| Recent OTP | Quick access to pinned and recently used services with live codes |
| Import Vault | Import encrypted \`.2fas\` export files |
| Setup | Check vault status, lock the vault, re-import, or delete |

### Security highlights

- AES-256-GCM encrypted local vault, key in macOS Keychain
- Atomic re-import (parse + validate before touching existing vault)
- Stable service IDs (re-import preserves recents cache)
- 5 MB import cap, vault envelope version + shape validation
- 5-minute in-memory cache TTL plus explicit \`Lock Vault\` action
- Concealed clipboard copy
- HUD-based copy feedback (no OTP code shown on screen)
- Zero network calls, zero external crypto dependencies
- Full threat model in \`SECURITY.md\`

### Screencast

Screenshots in \`metadata/\` show Search OTP with live countdown, Import Vault form, Recent OTP, and Setup views.

## Checklist

- [x] I read the [extension guidelines](https://developers.raycast.com/basics/prepare-an-extension-for-store)
- [x] I read the [documentation about publishing](https://developers.raycast.com/basics/publish-an-extension)
- [x] I ran \`npm run build\` and tested this distribution build in Raycast
- [x] I checked that files in the \`assets\` folder are used by the extension itself
- [x] I checked that assets used by the \`README\` are placed outside of the \`metadata\` folder

PR #29139, opened 2026-06-30
2FAS Authenticator: correct author handle

LockeAG:
This corrects the `author` field for the 2FAS Authenticator extension.

The extension was merged in #27908 with `author: "Locke"`, but `Locke` is a different, pre-existing Raycast user. My account is `Lock` (https://www.raycast.com/Lock), so the extension is currently attributed to the wrong person on the store. Every commit in #27908 is authored by me (GitHub `LockeAG`); this only fixes the handle so the extension is credited correctly.

Changes:
- `package.json`: `author` `Locke` -> `Lock`
- `README.md`: store install badge URL updated to `/Lock/`
- `README.md`: development setup referenced the wrong package manager (`npm`), which fails because the source repo enforces pnpm via an `only-allow pnpm` preinstall hook. Switched the documented commands to pnpm and pointed the clone URL at the actual repo.

No functional or code changes.

### DIFF
--- a/extensions/2fas-authenticator/CHANGELOG.md
+++ b/extensions/2fas-authenticator/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## [Fix Author Attribution] - {PR_MERGE_DATE}
+
+- Correct the extension author handle so it is credited to the right Raycast account
+
 ## [Initial Version] - 2026-06-30
 
 ### Commands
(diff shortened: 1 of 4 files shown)


pernielsentikaer:
We don't use `pnpm` though, can you revert that to `npm` in `README` @LockeAG

0xdhrv (approved):
Looks good to me, approved ✅

PR #29142, opened 2026-06-30
Update dynamic-font-size extension

tavlean:
## Description

Fixes several cases where the command produced invalid CSS, and adds inline validation so broken values can no longer be copied.

### Bugs fixed

- Equal minimum and maximum viewport widths produced `clamp(1rem, Infinityrem + Infinityvw, 3rem)`. This is now caught with a clear message.
- An empty or non-numeric field (for example `.`) produced `NaN` in the output. This is now caught.
- A comma decimal such as `1,5` was silently parsed as `1`. Commas are now accepted as decimal separators.
- A maximum smaller than the minimum is now reported instead of producing an inverted result.

### Improvements

- Inline per-field validation with `useForm` from `@raycast/utils`. The copy action is blocked while any field is invalid.
- Updated to the latest Raycast API.

## Checklist


pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #29147, opened 2026-06-30
[asana] Fix pagination error in large workspaces

### LINKED ISSUE #24015: [Asana] Create Task fails with "result is too large" pagination error in large workspaces
### Extension

https://www.raycast.com/raycast/asana

### Raycast Version

1.104.1

### OS Version

26.1 (25B78)

### Description

When attempting to open the "Create Task" command in the Asana extension, it hits a 400 error before the form fully loads. The error occurs immediately when the extension tries to fetch projects or other workspace data.

Error message: Request failed with status code 400
The result is too large. You should use pagination (may require specifying a workspace)!

Image

Image

### Steps To Reproduce

Open Raycast
Invoke the "Create Task" command for Asana
Select a workspace with many projects/tasks (in my case: centreforeffectivealtruism.org)
Error appears before the form can be fully used

### Current Behaviour

The form hits a 400 error. The Asana API is returning this error because the extension is requesting too much data without implementing pagination.

### Expected Behaviour

The Create Task form should load without hitting an error code, with the extension properly paginating through large result sets from the Asana API.

nulltrope:
## Description

This PR fixes https://github.com/raycast/extensions/issues/24015 by introducing pagination when fetching users/tags from the Asana API.


0xdhrv (approved):
Looks good to me, approved ✅

PR #29149, opened 2026-06-30
Quit Applications: show menu bar apps

artisticmedic:
## Description

`Quit Applications` only listed apps with a Dock presence, so menu bar / accessory (`LSUIElement`) apps — Stats, Shottr, BetterMouse, Kap, Music Presence, etc. — never appeared and couldn't be quit from the command.

### Root cause
Discovery used System Events `every process whose background only is false`. System Events reports `background only = true` for accessory apps, so they were filtered out and only Dock apps survived. The existing `ps` fallback over-corrected, surfacing helper processes (Chrome Helper, …) and system daemons.

### Fix
Switched discovery to `NSWorkspace.runningApplications` (via JXA). It:
- includes accessory / menu-bar apps;
- needs no Automation permission — removing the recurring `-1743` / `-1728` / timeout failure class that's been patched repeatedly;
- filters to apps a user would actually quit: drops `prohibited` activation policy (daemons / XPC) and nested helper bundles; always keeps regular (Dock) apps; keeps accessory apps only under `/Applications` or `~/Applications`, so third-party menu-bar utilities show while Apple's UI agents (Dock, Spotlight, WindowManager…) stay hidden.

The `ps`-based fallback is retained for the unlikely error path. Quitting still uses `tell application "…" to quit` (unchanged).

### Also in this PR
- Fixed the wrong list row being removed when quitting an app while another quit was in flight (stale render-closure index applied to a positional splice).
- Fixed duplicate React keys / unstable rows when two running apps share a name — the list key and dedup now use the bundle path.
- Small cleanups: dropped a now-dead `.app` filter and a dead (case-inconsistent) exclude re-check in “Quit All”; suffix-anchored the `.app` name strip.
- Modernized dependencies to the current Raycast API (`@raycast/api` 1.104, React 19, `@raycast/utils` 2), matching the rest of the repo.

## Checklist
- [x] I ran `npm run lint` and it passed
- [x] `assets` folder unchanged; existing icon reused

### DIFF
--- a/extensions/quit-applications/.eslintrc.json
+++ b/extensions/quit-applications/.eslintrc.json
@@ -1,4 +1,0 @@
-{
-  "root": true,
-  "extends": ["@raycast"]
-}
--- a/extensions/quit-applications/CHANGELOG.md
+++ b/extensions/quit-applications/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Quit Applications Changelog
 
+## [Show menu bar apps] - {PR_MERGE_DATE}
+
+- Switched running-app discovery to NSWorkspace so menu bar / accessory (LSUIElement) apps now appear in the list, while system agents and helper processes stay hidden.
+- Fixed the wrong row being removed from the list when quitting an app while another quit was still in progress.
+- Fixed duplicate list entries when two running apps share the same name.
+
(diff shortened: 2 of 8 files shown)


0xdhrv (approved):
Looks good to me, approved ✅

PR #29152, opened 2026-07-01
[Everything Search] Add extension

moviezhou:
Add Everything Search extension — instant file search for Windows powered by Voidtools Everything engine.

## Features

⚡ Instant search — powered by Everything NTFS index, results in milliseconds
🖥️ Native Raycast UI — file icons, Action Panel support
🌏 Chinese support — full GBK/CP936 encoding for Chinese filenames
📁 File actions — Open, Show in Explorer, Copy Path, Open with...
🔧 Configurable — custom es.exe path and max results limit


0xdhrv (commented):

We already have an extension in the Store that provides the same functionality. Could we consider enhancing the existing extension below instead of creating another one?

- [Everything Search](https://www.raycast.com/anastasiy_safari/everything-search) by @anastasiy_safari

Notable differences you added that would be valuable to merge:
* Display of total match count

Because the core functionality overlaps significantly with an existing Store extension, we can’t accept this as a new extension in its current form.

If there are unique features or workflows you're aiming to add, we'd love to hear them and see if they can be integrated into that extension to avoid duplication and improve discoverability.

> As mentioned in our [extension guidelines here ↗](https://manual.raycast.com/extensions)

PR #29156, opened 2026-07-01
[Windows Terminal] Fix profiles opening in System32

artistro08:
Profiles without a `startingDirectory` were inheriting Raycast's System32 working directory. Now launches from the user's home so the default is sensible; configured `startingDirectory` values still take precedence.


0xdhrv (approved):
Looks good to me, approved ✅

PR #29158, opened 2026-07-01
Update plexus extension

RensHoogendam:
## Description

Update to the existing **Plexus – Localhost Search** extension. Three groups of changes, no new commands or permissions:

**Fix — slow dev servers were being missed.** The HTTP probe timed out before dev servers that server-render their first response (Next.js, Nuxt, Angular, …) could answer — those routinely take ~1s in dev mode. The probe now waits long enough, so they're detected reliably instead of silently skipped.

**Performance — faster, more resilient loading.** Servers now stream into the list as each one is confirmed by the probe instead of blocking on the slowest, and the last results are cached so reopening shows them instantly while a fresh scan runs in the background. A single unresponsive port can no longer fail the whole scan, and each page title is read from the same probe response (no second request per server).

**UI — a richer, friendlier list.**

- Toggleable detail panel (⌘Y) showing framework, URL, port, PID, source (incl. WSL distro), project path, and page title.
- Color-coded framework tags, with the fallback icon tinted to match, so each server is identifiable at a glance.
- Refresh action (⌘R) and a clearer empty state.

Existing behavior (open in browser, copy URL/PID, kill process, Windows + WSL support) is unchanged.

## Screencast

_Straightforward logic/UI update to an existing extension — happy to add a screen recording if the reviewers would like one._

## Checklist

### DIFF
--- a/extensions/plexus/CHANGELOG.md
+++ b/extensions/plexus/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Plexus Changelog
 
-## [Detect slow dev servers, instant loading, and a richer list] - {PR_MERGE_DATE}
+## [Detect slow dev servers, instant loading, and a richer list] - 2026-07-01
 
 ### Added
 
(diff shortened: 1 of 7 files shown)


0xdhrv (approved):
Looks good to me, approved ✅

PR #29162, opened 2026-07-01
Add ocr extension

ielyas:
## Checklist


0xdhrv (commented):

We already have an extension in the Store that covers this core functionality. Could we consider enhancing the existing extension below instead of creating another one?

* [ScreenOCR](https://www.raycast.com/huzef44/screenocr) by @neo773

Both extensions let users select an area of the screen, capture it, extract the visible text, and copy the result to the clipboard. The existing extension already supports selected-area and full-screen recognition, multiple recognition languages, configurable accuracy, language correction, custom vocabulary, and optional line-break handling.

This PR adds some useful capabilities through OpenRouter vision models, including model selection, customizable OCR instructions, Markdown-preserving output, and provider-specific configuration. However, these feel like natural enhancements or an optional AI-powered recognition mode for **ScreenOCR**, rather than enough differentiation for a separate extension with the same primary screenshot-to-text workflow.

Adding OpenRouter as an alternative recognition provider and exposing the formatting options in the existing extension would preserve these improvements while keeping screen OCR functionality consolidated and easier to discover.

PR #29178, opened 2026-07-02
Update dev-servers extension

tavlean:
### Automatic port fallback for Shopify themes

- Shopify theme dev has no next-free-port fallback: when its fixed default port 9292 is already taken — for example by the main checkout while you start a git worktree of the same theme, the CLI just crashes with EADDRINUSE (Shopify/cli#5554).

- Starting a Shopify theme now probes the default port first and, when it's taken, exports SHOPIFY_FLAG_PORT (the documented env binding for --port) with the next free one. Because it's an environment variable, the fix reaches the CLI through any wrapping: a bare theme root started as shopify theme dev and a dev script that nests it under concurrently both come up on their own port. An explicit --port in the user's own script still wins, since the CLI gives command-line flags precedence over env vars. A short-lived reservation prevents parallel multi-folder starts from picking the same port.

- Additionally, when a spawn dies on a port conflict anyway (a non-Shopify server with a hardcoded port), the failure toast now says "a port is already in use by another process" instead of the generic "not detected after 15s" detected from the portion of the startup log written by that specific attempt and its View Startup Log action opens the log of the server that actually hit the conflict.

## Checklist


0xdhrv (approved):
Looks good to me, approved ✅

PR #29179, opened 2026-07-02
Add sbb-connections extension

melvinlauber:
## Checklist


0xdhrv (commented):

We already have an extension in the Store that covers the same service and core workflow. Could we consider enhancing the existing extension below instead of creating another one?

* [Swiss ÖV](https://www.raycast.com/Smoca/swiss-ov) by @OliverMannhart

Both extensions use `transport.opendata.ch` to search Swiss public-transport connections between two stations for a specified date and time. Both display departure and arrival details, duration, transfers, platforms, transport products, and individual journey sections.

This PR improves the experience with station autocomplete, paginated results, favorite routes, a more structured search flow, copying connection details, and an action to continue on the SBB website. However, these feel like natural enhancements to **Swiss ÖV** rather than enough differentiation for a separate extension using the same API and providing the same primary connection-search command.

The existing extension also already includes a **Connection History** command, while this PR introduces favorites. Both history and favorites could be maintained together in an updated Swiss ÖV extension.

0xdhrv (commented):
.

PR #29181, opened 2026-07-02
Add browsearch extension

Bryan-0220:
## Description

**Browsearch** lets you search your Firefox browsing history directly from Raycast and open results instantly in Firefox — all without leaving the keyboard.

**How it works:**
- Queries your local Firefox `places.sqlite` database (copied to a temp file, so Firefox is never interrupted)
- Ranks results by frecency and visit count, with prefix-match boosting
- Detects URLs typed directly and opens them without searching
- Falls back to a configurable search engine (Google, DuckDuckGo, Bing, or Custom) when no history matches
- Secondary action opens any result in a new Firefox window
- 100% local — no network requests, no external APIs, no telemetry

**Requirements:** Firefox installed on Windows. No API keys needed.

**Platform:** Windows only.

## Checklist


0xdhrv (commented):

We already have extensions in the Store that cover this core functionality. Could we consider enhancing one of the existing extensions below instead of creating another Firefox history search extension?

* [Mozilla Firefox](https://www.raycast.com/crisboarna/mozilla-firefox) by @crisboarna
* [Browser History](https://www.raycast.com/crisboarna/browser-history) by @crisboarna

The strongest overlap is with **Mozilla Firefox**. Both extensions search Firefox’s local `places.sqlite` history database, display matching pages, open results in Firefox, and provide a fallback web-search workflow. The existing extension also includes commands for opening a new Firefox tab and searching Firefox bookmarks.

There is additional overlap with **Browser History**, which already supports Firefox history alongside several other browsers and includes a configurable fallback search engine when no history result is found.

This PR adds meaningful Windows-specific implementation work, including Windows Firefox profile discovery, copying the active database before querying it, frecency and visit-count ranking, direct URL detection, and explicit new-tab or new-window actions. However, these feel like natural Windows support and search-quality improvements for the existing **Mozilla Firefox** extension rather than enough differentiation for a separate extension with the same primary command.

Adding Windows support to Mozilla Firefox would preserve the focused Firefox experience, while the reusable Windows history-reading logic could also help extend Browser History across platforms.

If there are other unique features or workflows you’re aiming to add, we’d love to hear them and see if they can be integrated into the existing extensions to avoid duplication and improve discoverability.

PR #29198, opened 2026-07-02
[codex] Fix Spotify player lyrics command

matijrn:
## What changed

- Replaced the broken `genius-lyrics` runtime dependency with a small LRCLIB API client for the Find Lyrics command.
- Kept current-track resolution on the extension's existing Spotify Web API path instead of adding extra AppleScript/OAuth fallbacks.
- Removed the old Genius scraper helper and the unused dependency tree from `package-lock.json`.
- Updated the README, command description, and changelog to stop referring to Genius as the lyrics provider.

## Why

The command was failing because the runtime `genius-lyrics` integration no longer reliably returned lyrics. The first version of this PR solved that with multiple fallbacks, but that made the feature harder to reason about. This version keeps the command functional with a single lyrics provider and the existing Spotify metadata path.

## Validation

- `npm run build`
- `npm run lint`
- Direct LRCLIB smoke test for Rick Astley - Never Gonna Give You Up, Michael Jackson - Human Nature, and John Mayer - In Your Atmosphere returned lyrics.


hugodemenez:
Don't we have too many fallbacks for this extension commands? Couldn't we make a functional feature for this lyrics fetcher?

hugodemenez:
I think this PR can stay narrower. The extension already has a few fallback-heavy patterns, and this change adds more around lyrics fetching. For this PR, could we keep the lyrics lookup straightforward: get the current track from the Spotify Web API, then fetch lyrics through one Genius path?

A source selector or additional lyrics providers could be a separate draft PR, since that is more of a product/UX expansion than a fix for this command.

hugodemenez:
Thanks @matijrn for addressing the comment this fast!

matijrn:
Thanks for the feedback. I simplified the PR to keep it focused: it now uses the existing Spotify Web API path for the current track and a single LRCLIB lookup for lyrics, with no AppleScript/OAuth fallback or source selector in this change.

I also fixed the changelog entry to use `{PR_MERGE_DATE}` and re-ran `npm run build` and `npm run lint` successfully.

0xdhrv (approved):
Looks good to me, approved ✅

PR #29199, opened 2026-07-02
Add New Tab from Selection extension

artisticmedic:
## Description

**New Tab from Selection** opens whatever you have selected (or the clipboard, or the selection plus typed text) in a **new browser tab**. If the selection is a URL it opens directly; otherwise it is searched with your chosen engine — the same default behavior as typing into a new tab.

## Demo

https://github.com/user-attachments/assets/775fafbd-febb-44bf-9976-9c11ffb4cb07

### Commands
- **New Tab from Selection** — open the highlighted text in a new tab (URL opens directly, otherwise searched)
- **New Tab from Clipboard** — same, sourced from the clipboard
- **New Tab from Selection + Text** — open the selection plus extra terms you type (press Tab to focus the box)

### Preferences
- **Search engine** — DuckDuckGo (default), Google, Brave, Bing, Kagi, Startpage, or a **custom `{query}` template** for any engine
- **Browser** — where the tab opens (defaults to system default)
- **URL handling** — open the selection directly when it is a URL

### How this differs from native Quicklinks and existing extensions
- **vs Raycast Quicklinks** (which can fire a `{selection}`/`{clipboard}` placeholder at a fixed URL): this **detects whether the selection is a URL and opens it directly** instead of searching, lets you **choose the target browser**, offers a **custom `{query}` template** for any engine, and can **append typed terms** to the selection — none of which a static Quicklink does.
- **vs the *Quick Search* extension** (a hard-coded command per engine, default browser only, `@raycast/api ^1.45`): preference-driven engine + custom template, browser override, and direct-URL opening.

Uses the native `getSelectedText()` API (no clipboard clobber, no synthetic ⌘C).

## Checklist
- [x] `npm run build`, `npm run lint`, and `npm ci` pass cleanly
- [x] Unit tests included (`npm test` — 29 passing)
- [x] `package-lock.json` committed
- [x] Screenshots added to `metadata/` (2000×1250, Raycast Window Capture framing)


artisticmedic:
commenting to avoid going stale!

0xdhrv (commented):
Hey @artisticmedic, thanks for the submission and for the work you’ve put into it!

After reviewing the extension, the primary workflows can already be achieved using Raycast’s native Quicklinks:

- Selected text can be passed to a search URL using `{selection}`
- Clipboard content can be passed using `{clipboard}`
- Additional text can be collected using `{argument}` and combined with the selection
- Any search engine or custom URL template can be configured directly in the Quicklink
- A specific browser can be selected using the Quicklink’s **Open With** option
- A global hotkey can be assigned to run the Quicklink directly

For example:

https://www.google.com/search?q={selection}

or:

https://www.google.com/search?q={selection}%20{argument name="additional text"}

The automatic URL-versus-search detection is a useful convenience, but users can achieve comparable value with separate Quicklinks for opening a selected URL and searching selected text. This difference isn’t substantial enough to justify maintaining a standalone Store extension.


You can find more information about Quicklinks here:
https://manual.raycast.com/quicklinks

Thanks again for contributing!

artisticmedic:
@0xdhrv that's pretty metal, I hadn't thought to use quicklinks that way before. Thanks!

PR #29227, opened 2026-07-04
[Agent Usage] move to useCachedPromise with TTL to reduce API calls

### LINKED ISSUE #27057: [Agent Usage] Claude show 429 Too Many Requests
### Extension

https://www.raycast.com/thuggyduck/agent-usage

### Raycast Version

1.104.12

### OS Version

macOS 26.4

### Description

Image

### Steps To Reproduce

I just ran the extension.

### Current Behaviour

Showing an error

### Expected Behaviour

Showing usage

kevin-david:
## Description

- Refactors the state management for agent fetching to use `@raycast/utils` `useCachedPromise`, reducing boilerplate (deleting ~900 lines of manual `useState` and `useEffect` logic)
- Implements a configurable, global disk cache with a standard TTL limit (defaulting to 3 minutes) via Raycast`s `Cache` API.
- Minor updates to tooltip displays in the menu bar 

### Why?
Previously, having the menubar extension run in the background (or even just frequently opening the main command) would blast the APIs on every single mount. This was causing constant loading screens (spinners flashing / UI jank / `Loading...` messages), but the real pain was I kept getting hit with `429 Rate Limit Exceeded` errors from Claude because the usage endpoints were being hammered automatically. This should also fix https://github.com/raycast/extensions/issues/27057.

By introducing a global TTL cache and syncing it across all views, we completely eliminate those redundant network hits, protect against rate-limiting bans, and make the UI smoother.

### Testing 

I tested this for Codex, Copilot, Antigravity, z.ai, and Claude as the screenshots below show. Would appreciate anyone with other tool subscriptions testing those as well. 

## Screencast

image

image

image

## Checklist


extensions/agent-usage/src/agents/hooks.ts
```diff
@@ -1,154 +1,244 @@
...
+import { useCachedPromise } from "@raycast/utils";
+import { useCallback } from "react";
 import type { UsageState } from "./types";
+import type { AccountUsageState } from "../accounts/types";
+import { isOpenCodeActiveToken } from "./opencode-active";
+
+export const fetchTtlCache = new Cache({ namespace: "agent-usage-ttl" });
+export function getTtlMs(): number {
+  const prefs = getPreferenceValues<{ cacheTtl?: string }>();
+  const parsed = parseInt(prefs.cacheTtl || "180", 10);
```
kevin-david: is this the canonical way to do defaults in raycast preferences?

extensions/agent-usage/src/agents/hooks.ts
```diff
@@ -1,154 +1,244 @@
...
+    const cachedRaw = fetchTtlCache.get(ttlKey);
+    let cachedData;
+    let lastFetched = 0;
+    if (cachedRaw) {
+      if (cachedRaw.startsWith("{")) {
+        try {
+          cachedData = JSON.parse(cachedRaw);
+          lastFetched = cachedData.timestamp || 0;
+        } catch {
+          /* fallback */
```
kevin-david: this fallback seems weird, why are we testing if its json? don't we control writes, so if it's malformed that would be an exceptional case? b/c this cache didn't exist until we created it ?

extensions/agent-usage/src/agents/hooks.ts
```diff
@@ -1,154 +1,244 @@
...
-      }
-
-      await fetchData();
-    }, [enabled, fetchData]);
+    }
+    const isStale = Date.now() - lastFetched > getTtlMs();
+
+    const fetcherFn = useCallback(
+      async (agent: string, t: string) => {
+        if (!t) {
```
kevin-david: let's use a non-one-letter variable

extensions/agent-usage/src/agents/hooks.ts
```diff
@@ -1,154 +1,244 @@
...
+
+    const { data: accounts } = usePromise(options.getAccounts);
+
+    const accountsHash = accounts ? JSON.stringify(accounts.map((a) => a.token)) : "loading";
+    const ttlKey = `ttl-${options.agentName}-accounts-${accountsHash}`;
+    const cachedRaw = fetchTtlCache.get(ttlKey);
+    let cachedData;
+    let lastFetched = 0;
+    if (cachedRaw) {
+      if (cachedRaw.startsWith("[")) {
```
kevin-david: why are we doing this weird fallback/guard here too?

extensions/agent-usage/src/amp/fetcher.ts
```diff
@@ -67,11 +70,28 @@ function getExecFailureMessage(error: unknown): string {
   return stderr || stdout || execError.message;
 }
 
+export async function fetchAmpUsageFromCli(
```
kevin-david: why are we introducing this function and modifying the above around line 38-44? it seems unrelated to caching

extensions/agent-usage/src/claude/fetcher.ts
```diff
@@ -521,68 +520,68 @@ async function fetchClaudeUsage(
...
-
-    setIsLoading(true);
-    setError(null);
-
-    const { credentials, error: credentialsError } = readClaudeCredentials();
-    if (!credentials) {
-      if (requestId !== requestIdRef.current) {
-        return;
+export function useClaudeUsage(enabled = true): import("../agents/types").UsageState<ClaudeUsage, ClaudeError> {
+  // eslint-disable-next-line @typescript-eslint/no-require-imports
```
kevin-david: let's not introduce new eslint-disables, fix this properly using idiomatic imports

extensions/agent-usage/src/claude/fetcher.ts
```diff
@@ -521,68 +520,68 @@ async function fetchClaudeUsage(
...
+  const cachedRaw = fetchTtlCache.get(ttlKey);
+  let cachedData;
+  let lastFetched = 0;
+  if (cachedRaw) {
+    if (cachedRaw.startsWith("{")) {
+      try {
+        cachedData = JSON.parse(cachedRaw);
+        lastFetched = cachedData.timestamp || 0;
+      } catch {
+        /* fallback */
```
kevin-david: another weird fallback

extensions/agent-usage/src/codex/fetcher.ts
```diff
@@ -226,182 +228,94 @@ function getResetsInSeconds(window: { reset_after_seconds?: number; reset_at?: n
...
+  const cachedRaw = fetchTtlCache.get(ttlKey);
+  let cachedData;
+  let lastFetched = 0;
+  if (cachedRaw) {
+    if (cachedRaw.startsWith("{")) {
+      try {
+        cachedData = JSON.parse(cachedRaw);
+        lastFetched = cachedData.timestamp || 0;
+      } catch {
+        /* fallback */
```
kevin-david: another weird fallback

extensions/agent-usage/src/codex/fetcher.ts
```diff
@@ -226,182 +228,94 @@ function getResetsInSeconds(window: { reset_after_seconds?: number; reset_at?: n
...
-
-  useEffect(() => {
-    if (!enabled) {
-      requestIdRef.current += 1;
-      setUsage(null);
-      setError(null);
-      setIsLoading(false);
-      setHasInitialFetch(false);
-      return;
+  const ttlKey = "ttl-codex-primary";
```
kevin-david: codex has its own ttl? 

also do we have a cache of TTL values, or what's going on here?

extensions/agent-usage/src/codex/renderer.tsx
```diff
@@ -15,6 +15,7 @@ export function formatCodexUsageText(usage: CodexUsage | null, error: CodexError
   const fallback = formatErrorOrNoData("Codex", usage, error);
   if (fallback !== null) return fallback;
   const u = usage as CodexUsage;
+  if (!u.fiveHourLimit) return "Invalid Cache. Please refresh.";
```
kevin-david: "Invalid cache" seems like a case we should handle versus telling the user to do something. Under what conditions could this actually occur?

extensions/agent-usage/src/copilot/fetcher.ts
```diff
@@ -176,96 +174,97 @@ async function fetchCopilotUsage(token: string): Promise<{ usage: CopilotUsage |
...
-      preferenceToken: cleanedPreferenceToken,
-    } = await resolveCopilotAuthTokens({
-      preferenceToken,
-    });
-
-    if (!primaryToken) {
-      if (requestId !== requestIdRef.current) {
-        return;
+export function useCopilotUsage(enabled = true): import("../agents/types").UsageState<CopilotUsage, CopilotError> {
+  // eslint-disable-next-line @typescript-eslint/no-require-imports
```
kevin-david: no introduction of eslint-disables, there must be a better way

extensions/agent-usage/src/droid/fetcher.ts
```diff
@@ -70,67 +69,74 @@ function parseDroidApiResponse(data: unknown): { usage: DroidUsage | null; error
...
+  const cachedRaw = fetchTtlCache.get(ttlKey);
+  let cachedData;
+  let lastFetched = 0;
+  if (cachedRaw) {
+    if (cachedRaw.startsWith("{")) {
+      try {
+        cachedData = JSON.parse(cachedRaw);
+        lastFetched = cachedData.timestamp || 0;
+      } catch {
+        /* fallback */
```
kevin-david: another fallback

extensions/agent-usage/src/minimax/fetcher.ts
```diff
@@ -69,71 +67,59 @@ async function fetchMiniMaxUsage(token: string): Promise<{ usage: MiniMaxUsage |
...
-
-  const revalidate = useCallback(async () => {
-    if (!enabled) {
-      return;
-    }
-
-    await fetchData();
-  }, [enabled, fetchData]);
+export function useMiniMaxUsage(enabled = true): import("../agents/types").UsageState<MiniMaxUsage, MiniMaxError> {
+  // eslint-disable-next-line @typescript-eslint/no-require-imports
```
kevin-david: more eslint-disables

extensions/agent-usage/src/opencode-go/fetcher.ts
```diff
@@ -97,91 +95,77 @@ export async function fetchOpencodegoUsage(
...
+  // eslint-disable-next-line @typescript-eslint/no-require-imports
+  const { getPreferenceValues: getPrefs } = require("@raycast/api") as typeof import("@raycast/api");
+  // eslint-disable-next-line @typescript-eslint/no-require-imports
+  const { useCachedPromise } = require("@raycast/utils") as typeof import("@raycast/utils");
+
+  const preferences = getPrefs<Preferences.AgentUsage>();
+  const workspaceId = (preferences.opencodegoWorkspaceId as string)?.trim() || "";
+  const authCookie = (preferences.opencodegoAuthCookie as string)?.trim() || "";
+
+  // eslint-disable-next-line @typescript-eslint/no-require-imports
```
kevin-david: more eslint-disables

extensions/agent-usage/src/agent-usage-menubar.tsx
```diff
@@ -65,6 +68,8 @@ const codexIcon: Image.ImageLike = { source: { light: "codex-icon.svg", dark: "c
 
 export default function MenuBarCommand() {
   const prefs = getPreferenceValues<Preferences>();
+  const now = Date.now();
```
kevin-david: is `now` used?

extensions/agent-usage/src/agent-usage-menubar.tsx
```diff
@@ -45,14 +44,18 @@ interface MenuBarAgent {
...
+  revalidate: (force?: boolean) => Promise<void>;
+  lastFetchedAt?: number;
   /** True if this account's token matches the one configured in OpenCode */
   isOpenCodeActive?: boolean;
 }
 
 type Preferences = Preferences.AgentUsageMenubar;
 
 function getMenuItemTitle(name: string, value: string, isOpenCodeActive?: boolean): string {
+  if (value === "Loading...") {
```
kevin-david: switching on a string value that's supposed to be user-visible seems like the wrong pattern

extensions/agent-usage/src/agent-usage-menubar.tsx
```diff
@@ -232,13 +255,19 @@ export default function MenuBarCommand() {
       isCodexVisible
         ? codexAccounts.map((account) => ({
             id: `codex-${account.accountId}` as AgentId,
-            name: account.label === "Default" ? "Codex" : `Codex • ${account.label}`,
+            name:
+              account.accountId === "loading"
```
kevin-david: why can account ID be "loading"? that seems very strange

applies elsewhere as well

extensions/agent-usage/src/agent-usage-menubar.tsx
```diff
@@ -301,25 +348,27 @@ export default function MenuBarCommand() {
...
+    await Promise.all(visibleAgents.map((a) => a.revalidate(true)));
     await showHUD("Agent Usage Refreshed");
   };
 
+  const lastFetchedTimestamps = visibleAgents
+    .map((agent) => agent.lastFetchedAt)
+    .filter((t): t is number => typeof t === "number" && t > 0);
+  const latestFetchedAt = lastFetchedTimestamps.length > 0 ? Math.max(...lastFetchedTimestamps) : undefined;
+
+  const formatTimeAgoShort = (timestamp?: number) => {
```
kevin-david: both the menubar and the agent usage modal should use the same method to do this - confirm that they do, and no utility method exists elsewhere

extensions/agent-usage/src/agents/hooks.ts
kevin-david: **[Fable 5 high]** Mostly yes — the `default: "180"` in package.json is the canonical mechanism and covers the unset case (`getPreferenceValues` returns it). The runtime guard only matters when the user clears the textfield or types garbage. That parsing now lives in `parseTtlSeconds` in `src/agents/usage-cache.ts` (falls back to 180, clamps negatives to 0, unit-tested) instead of an inline `isNaN` dance. a63e663579

extensions/agent-usage/src/agents/hooks.ts
kevin-david: **[Fable 5 high]** You were right — we control the writes, and the dual JSON/bare-number branch only existed because two fetchers (minimax, opencode-go) wrote bare timestamps while the rest wrote JSON payloads. Fixed by making the hook factory the single writer of one canonical JSON payload in a versioned namespace (`agent-usage-ttl-v2`). A parse failure is now only possible when an older extension version wrote a different shape, and it reads as a cache miss (silent refetch) — `parseCachedPayload` in `src/agents/usage-cache.ts`, with tests. a63e663579

extensions/agent-usage/src/claude/fetcher.ts
```diff
@@ -521,68 +520,68 @@ async function fetchClaudeUsage(
...
+      if (!credentials) {
+        return {
+          usage: null,
+          error: credentialsError,
+          timestamp: Date.now(),
+        };
+      }
+      const result = await fetchClaudeUsage(credentials);
+      const newData = { ...result, timestamp: Date.now() };
+      fetchTtlCache.set(ttlKey, JSON.stringify(newData));
```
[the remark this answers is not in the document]
kevin-david: **[Fable 5 high]** Fixed factory-wide: only successful fetches are persisted, so an error result never refreshes the TTL and the next launch retries immediately. Regression-tested in `src/agents/usage-cache.test.ts` ("isPayloadFresh rejects error payloads so failures are retried"). 0318331308, a63e663579

extensions/agent-usage/src/copilot/fetcher.ts
```diff
@@ -176,96 +174,97 @@ async function fetchCopilotUsage(token: string): Promise<{ usage: CopilotUsage |
...
+          errorType: result.error?.type,
+        })
+      ) {
+        result = await fetchCopilotUsage(cleanedPreferenceToken);
+      }
 
-    await fetchData();
-  }, [enabled, fetchData]);
+      const newData = { ...result, timestamp: Date.now() };
+      fetchTtlCache.set(ttlKey, JSON.stringify(newData));
```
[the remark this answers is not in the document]
kevin-david: **[Fable 5 high]** Fixed — same success-only persistence rule as the Claude thread, applied to every provider through the shared factory: a failed primary+fallback attempt returns the error to the UI but writes nothing to the cache, so the retry happens on the next mount. a63e663579

extensions/agent-usage/src/copilot/fetcher.ts
```diff
@@ -176,96 +174,97 @@ async function fetchCopilotUsage(token: string): Promise<{ usage: CopilotUsage |
...
+  const preferences = getPreferenceValues<Preferences>();
+  const preferenceToken = preferences.copilotAuthToken?.trim() || "";
+
+  // eslint-disable-next-line @typescript-eslint/no-require-imports
+  const { useCallback } = require("react") as typeof import("react");
+
+  // eslint-disable-next-line @typescript-eslint/no-require-imports
+  const { fetchTtlCache, getTtlMs } = require("../agents/hooks") as typeof import("../agents/hooks");
+
+  const ttlKey = `ttl-copilot-${preferenceToken}`;
```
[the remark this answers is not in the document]
kevin-david: **[Fable 5 high]** Fixed — the cache no longer keys on the preference token. The fetcher hashes the *effective* auth material (resolved local GH_TOKEN/GITHUB_TOKEN + preference token) into each payload via `resolveAuthKey`, and a payload only counts as fresh when the hash matches. A local token change now bypasses the cached entry immediately. Raw tokens also no longer appear in cache keys or persisted values — only the SHA-256 hash. a63e663579

extensions/agent-usage/src/agents/hooks.ts
kevin-david: **[Fable 5 high]** Fixed — that code was replaced wholesale by the shared factory in `src/agents/hooks.ts`; no single-letter names remain. a63e663579

extensions/agent-usage/src/agents/hooks.ts
kevin-david: **[Fable 5 high]** Fixed along with the other fallback branches — one canonical JSON payload format, versioned namespace, parse failure = cache miss. The accounts variant also stopped persisting tokens: rows are cached token-less and re-joined with live tokens on load. a63e663579

extensions/agent-usage/src/amp/fetcher.ts
kevin-david: **[Fable 5 high]** It was an unused test seam (no test ever called `fetchAmpUsageFromCli`) plus an unrelated removal of the Windows `~/.local/bin/amp.exe` fallback path. Dropped the seam and restored the path. I kept the useExec→plain-async-fetcher conversion itself, because the shared cached hook needs a promise-returning fetcher — the old `useExec`-based hook cannot be TTL-cached. a63e663579

extensions/agent-usage/src/claude/fetcher.ts
kevin-david: **[Fable 5 high]** Fixed, and the fix is architectural rather than cosmetic: the `require()`s existed because `@raycast/api` has no runtime entry outside Raycast (`main: null` — the runtime injects it), so any module in a test's import graph that top-level-imports it crashes the Node test runner. All hook wiring now lives in `src/agents/provider-hooks.ts` (the one module besides `hooks.ts` that touches `@raycast/api`), and provider `fetcher`/`auth`/`parser` modules are pure with ordinary top-level imports. Zero `eslint-disable` requires remain; the rule is recorded in AGENTS.md. a63e663579

extensions/agent-usage/src/claude/fetcher.ts
kevin-david: **[Fable 5 high]** Fixed — same as the hooks.ts thread: single canonical payload format, and the Claude hook is now a `createUsageHook` wiring in `provider-hooks.ts` with no cache-parsing code of its own. a63e663579

extensions/agent-usage/src/codex/fetcher.ts
kevin-david: **[Fable 5 high]** Fixed by deletion — `useCodexUsage` turned out to be dead code (the registry entry was never invoked; both the list and the menu bar render Codex from `useCodexAccounts`), so the whole hand-rolled hook is gone along with its fallback parsing. a63e663579

extensions/agent-usage/src/codex/fetcher.ts
kevin-david: **[Fable 5 high]** It was a per-provider cache *key* (`ttl-codex-primary`) in the shared cache, not a separate TTL — the TTL duration is the single `cacheTtl` preference; the cache stores one payload (usage + error + timestamp + auth hash) per provider. That said, this hook was dead code (never invoked — Codex renders from `useCodexAccounts` in both commands) and is deleted; the remaining cache keys are uniform `<agentId>` / `<agentId>-accounts`. a63e663579

extensions/agent-usage/src/codex/renderer.tsx
kevin-david: **[Fable 5 high]** It could only occur when a payload written by an older extension version had a different shape than the current `ClaudeUsage`/`CodexUsage` type. Now handled at the read side instead of the render side: payloads live in a versioned namespace (`agent-usage-ttl-v2`, bumped whenever the shape changes) and are shape-validated on read, so an old/invalid entry reads as a cache miss and refetches automatically. All three "Invalid Cache" renderer guards are removed. a63e663579

extensions/agent-usage/src/copilot/fetcher.ts
kevin-david: **[Fable 5 high]** Fixed — same as the Claude thread: hook wiring moved to `provider-hooks.ts`, `copilot/fetcher.ts` is pure with ordinary imports, no `eslint-disable` requires remain. a63e663579

extensions/agent-usage/src/droid/fetcher.ts
kevin-david: **[Fable 5 high]** Fixed — droid is a `createUsageHook` wiring now; no provider-local cache parsing. a63e663579

extensions/agent-usage/src/minimax/fetcher.ts
kevin-david: **[Fable 5 high]** Fixed — same architectural fix (wiring in `provider-hooks.ts`, pure fetcher). Worth noting this hook also had the worst caching bug of the set: it stamped the TTL timestamp *before* fetching and never persisted the payload, so within the TTL window the other command showed "No data" with no retry. The shared factory fixes that too (persist-on-success only, after the fetch). a63e663579

extensions/agent-usage/src/opencode-go/fetcher.ts
kevin-david: **[Fable 5 high]** Fixed — same as the other require() threads, and opencode-go shared the minimax bug (timestamp stamped before the fetch, payload never persisted). Both are plain `createUsageHook` wirings now. a63e663579

extensions/agent-usage/src/agent-usage-menubar.tsx
kevin-david: **[Fable 5 high]** It was — in the `formatTimeAgoShort` closure below — but easy to miss. Gone now: the shared helper in `src/agents/format.ts` defaults to `Date.now()`, so the menu bar has no render-time `now` variable at all. The list view keeps its 5s ticker so the label stays fresh while the window is open. a63e663579

extensions/agent-usage/src/agent-usage-menubar.tsx
kevin-david: **[Fable 5 high]** Agreed — fixed by passing the state instead of sniffing the string: `getMenuItemTitle(name, value, isLoading, isOpenCodeActive)` branches on the `isLoading` flag; the "Loading..." accessory text is display-only again. a63e663579

extensions/agent-usage/src/agent-usage-menubar.tsx
kevin-david: **[Fable 5 high]** It couldn't, actually — those branches were dead: the accounts hooks never emitted an `accountId === "loading"` row, so multi-account providers silently vanished from the menu bar while loading. Fixed properly: the hooks now return `{ accounts, isLoading, revalidate }`, and the menu bar renders an explicit loading item from that flag. All four magic-id ternaries are gone. a63e663579

extensions/agent-usage/src/agent-usage-menubar.tsx
kevin-david: **[Fable 5 high]** Confirmed there was no existing utility (`src/agents/format.ts` only had reset-time/duration/percent helpers). Both commands now share `formatTimeAgoShort` + `latestTimestamp` from `format.ts`; the duplicated inline closures are gone. a63e663579

extensions/agent-usage/src/agents/provider-hooks.ts
```diff
@@ -0,0 +1,300 @@
...
+    const { accessToken } = await resolveDroidAuth();
+    if (!accessToken) {
+      return {
+        usage: null,
+        error: {
+          type: "not_configured",
+          message:
+            "Droid not configured. Run `droid` to log in (auto-detected from ~/.factory/auth.v2.* or ~/.factory/auth.*).",
+        },
+      };
```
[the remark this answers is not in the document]
kevin-david: **[Fable 5 high]** Fixed — Gemini now hashes the local refresh token into its cached payload via `readGeminiAuthKey()` (the refresh token is stable across access-token rotation but changes on re-auth), so switching accounts within the TTL invalidates the cache like the other credential-backed providers. The in-extension reauth flow was already covered because it calls `revalidate()`, which always bypasses the TTL. fc58094b74

extensions/agent-usage/src/agents/hooks.ts
```diff
@@ -1,154 +1,205 @@
...
+            error: result.error,
+            isOpenCodeActive: openCodeKey ? isOpenCodeActiveToken(account.token, openCodeKey) : false,
+          };
+        }),
+      );
+
+      const payload: Payload = { usage: rows, error: null, timestamp: Date.now(), authHash };
+      if (rows.some((row) => row.usage !== null)) {
+        usageCache.set(cacheKey, JSON.stringify({ ...payload, usage: stripAccountTokens(rows) }));
       }
```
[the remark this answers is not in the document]
kevin-david: **[Fable 5 high]** Fixed in b2e169abbd. Account payloads are now persisted only when every row succeeded (`allAccountRowsSucceeded`, a pure helper with a regression test covering the mixed success/failure case), matching the single-account success-only rule. Trade-off accepted: one persistently failing account means the healthy accounts also refetch each launch, but transient errors retrying beats pinning a failed row for the TTL.

extensions/agent-usage/src/agents/usage-cache.ts
```diff
@@ -0,0 +1,99 @@
...
+
+export function parseCachedPayload<TUsage, TError>(
+  raw: string | undefined,
+): CachedUsagePayload<TUsage, TError> | undefined {
+  if (!raw) return undefined;
+  let parsed: unknown;
+  try {
+    parsed = JSON.parse(raw);
+  } catch {
+    return undefined;
```
kevin-david, line 53: any logging functionality we can use here?

check for other try/catch blocks that could be worth logging too, if the facility exists

extensions/agent-usage/src/agents/usage-cache.ts
```diff
@@ -0,0 +1,99 @@
...
+  }
+  if (typeof parsed !== "object" || parsed === null || Array.isArray(parsed)) return undefined;
+  const candidate = parsed as Record<string, unknown>;
+  if (!("usage" in candidate) || !("error" in candidate)) return undefined;
```
kevin-david: this string-y logic feels wrong for lines 52/53. better approach? 

also `typeof parsed !== object`??

extensions/agent-usage/src/agents/provider-hooks.ts
```diff
@@ -0,0 +1,303 @@
...
+import { fetchOpencodegoUsage } from "../opencode-go/fetcher";
+import type { OpencodegoError, OpencodegoUsage } from "../opencode-go/types";
+import { fetchSyntheticUsage, SYNTHETIC_OPENCODE_KEY } from "../synthetic/fetcher";
+import type { SyntheticError, SyntheticUsage } from "../synthetic/types";
+import { resolveZaiAuthTokens } from "../zai/auth";
+import { fetchZaiUsage, ZAI_OPENCODE_KEY } from "../zai/fetcher";
+import type { ZaiError, ZaiUsage } from "../zai/types";
+
+/**
+ * All provider hook wirings live here — the one module (besides `hooks.ts`)
```
kevin-david: Why this decision vs alternatives? Also the imports for this are pretty wild

extensions/agent-usage/src/opencode-go/parser.ts
```diff
@@ -169,12 +169,12 @@ function extractSolidHydrationData(html: string): SolidHydrationData {
   // Try multiple patterns for usage data
   // Pattern 1: Original with specific indices
   let usageSection = script.match(
-    /\$R\[24\]\(\$R\[18\],\$R\[\d+\]=\{([^}]*rollingUsage[\s\S]*?)\}\);\$R\[24\]\(\$R\[20\]/,
+    /(?:\$R\[24\]\(|\)\()\$R\[18\],\$R\[\d+\]=\{([^}]*rollingUsage[\s\S]*?)\}\);\s*\$R\[24\]\(\$R\[20\]/,
```
kevin-david, line 175: Why these regex changes?

extensions/agent-usage/src/agents/usage-cache.ts
kevin-david: **[Fable 5 high]**

Replaced the `typeof/null/Array.isArray` triple and the `"usage" in candidate` sniffing with a named `isRecord` type guard + one consolidated shape check (eb509dc605). The `candidate` alias is gone — `isRecord` narrows `parsed` directly, so the field checks read as a single predicate. The `"usage"/"error" in` checks stay, since validating an unknown JSON shape always comes down to key membership, but they no longer sit next to a confusing `typeof parsed !== "object"` (which existed only because `typeof null === "object"` and arrays are objects too — exactly what `isRecord` hides). 17 cache-payload tests still pass.

extensions/agent-usage/src/agents/usage-cache.ts
kevin-david: **[Fable 5 high]**

The only facility is `console.error` — used in just two spots (`accounts/storage.ts`, `agents/opencode-auth.ts`); no Raycast `Log` and no structured logger. This file is deliberately Raycast-free so the plain-Node tests can load it, so `console.error` is the only option here regardless.

Low value at this specific catch: every payload is written by us via `JSON.stringify`, so malformed JSON means disk corruption or external tampering, and returning a cache miss (re-fetching) is the correct response either way. The catches actually worth logging are the silent auth-read ones (`codex/auth.ts`, `gemini/auth.ts`, `droid/auth.ts`) that swallow to `null`/defaults and surface to the user as "not configured" with no breadcrumb. Want me to do a `console.error` pass over those — and add one here for parity — or leave the pure helper quiet?

extensions/agent-usage/src/agents/provider-hooks.ts
kevin-david: **[Fable 5 high]**

**Decision:** every provider hook needs `@raycast/api` (`getPreferenceValues`, `LocalStorage`, `useCachedPromise`), but the fetcher/auth/parser modules must stay `@raycast/api`-free so the Node test runner can import them directly — that is the hard invariant (the file's own doc comment states only it and `hooks.ts` may touch `@raycast/api`). Centralizing the wiring here is what keeps the providers test-loadable; the alternative, hooks inside each fetcher, would drag `@raycast/api` into every fetcher and break that.

A lazy registry/map was considered: it would hide the import wall, but you still enumerate every provider's `(fetcher, auth, types)` somewhere, and you lose the per-call generic type-safety (`createUsageHook<TUsage, TError>`) the flat block gives. So the flat block is the explicit, greppable manifest.

**Imports:** the noise is real but mechanical — two alphabetized lines per provider (value + type). Options: (a) accept it as the provider manifest (my lean); (b) add an `index.ts` barrel per provider to collapse to one line each, at the cost of ~15 new files for little real win; (c) region-comment the block. Happy to do (b) if you want the collapse.

extensions/agent-usage/src/opencode-go/parser.ts
kevin-david: **[Fable 5 high]**

The opencode-go dashboard's Solid hydration can register the usage block as an IIFE — `($R[24]=(r,d)=>{…})($R[18],$R[27]={…rollingUsage…})` — so the object is preceded by `)(`, not `$R[24](`. Old Pattern 1 anchored on `\$R[24]\(` and missed that, falling through to Pattern 3, which recovers the object with a fragile `lastIndexOf("{")`/`indexOf("}")` brace scan. The new alternation `(?:\$R[24]\(|\)\()` lets Pattern 1 match both the direct-call and IIFE-call prefixes so the specific pattern hits first; the `\s*` / `[\s\S]*?` tweaks just tolerate whitespace between the rolling/weekly fields. The fixture in `parser.test.ts` is exactly the IIFE form and covers it.

Process note: this is a parser-robustness fix, not caching-related — it got bundled into the refactor commit `940863f` with no comment. Want me to add a one-line comment on Pattern 1 noting the IIFE variant?

extensions/agent-usage/src/agents/provider-hooks.ts
kevin-david: right, but that was a decision you made. don't use what was written here to back your argument - I want it from first principles

extensions/agent-usage/src/opencode-go/parser.ts
kevin-david: what's the impact if we don't make this change?

extensions/agent-usage/src/agents/usage-cache.ts
kevin-david: those would be good to address if it's a change we introduced

kevin-david: Adjusted the new cache parse path to log with `console.error("Failed to parse usage cache payload:", error)` before returning a cache miss, and added a focused test for the diagnostic. I did not expand this into Codex/Gemini/Droid auth reads here because those modules are not changed in this PR diff.

extensions/agent-usage/src/agents/provider-hooks.ts
kevin-david: Reworked the comment from first principles: this file is the Raycast adapter layer that combines preferences, hook lifecycle, and shared caching with provider modules that remain plain-Node importable for tests. The import block stays explicit/noisy because moving this wiring into providers would couple fetcher/auth/parser modules to the Raycast runtime.

extensions/agent-usage/src/opencode-go/parser.ts
kevin-david: Added the impact inline: without accepting the `($R[24]=...)(...)` IIFE hydration form, parsing falls through to the broad brace-scan fallback, which can lose weekly/monthly quota data from the nested usage object. The existing OpenCode Go fixture is that IIFE shape.

0xdhrv:
Hey @natsustan 👋 

Can you take a look at this?

0xdhrv (approved):
Looks good to me, approved ✅

0xdhrv (approved):
Looks good to me, approved ✅

PR #29239, opened 2026-07-04
Fuzzy File Search: add Cmd+Enter shortcut for Show in Finder

gilitos92:
## Summary
- Reorder actions so **Show in Finder** is the secondary action and receives **Cmd+Enter**
- **Open With** remains on **Cmd+O**

## Motivation
In Raycast List views, the 1st action gets Enter and the 2nd gets Cmd+Enter. Open With was second, so Cmd+Enter opened the app picker instead of revealing the file in Finder.

### DIFF
--- a/extensions/fuzzy-file-search/CHANGELOG.md
+++ b/extensions/fuzzy-file-search/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Fuzzy File Search Changelog
 
-## [Show in Finder Shortcut] - {PR_MERGE_DATE}
+## [Show in Finder Shortcut] - 2026-07-06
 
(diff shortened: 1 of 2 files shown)


erykksc:
Is the order of actions that is also present in the default file search? 
I think it is good ergonomics to have the same order and keyboard shortcuts present as the default search to make the experience of new users great and transition to the extension smooth.

gilitos92:
> Is the order of actions that is also present in the default file search? I think it is good ergonomics to have the same order and keyboard shortcuts present as the default search to make the experience of new users great and transition to the extension smooth.

Yes!

0xdhrv (approved):
Looks good to me, approved ✅

Issue #29171, opened 2026-07-02
[Google Contacts] Setup instructions in README are outdated and don't match current Google OAuth 2.0 requirements


megaroeny:
### Extension

https://www.raycast.com/aic/google-contacts

### Raycast Version

0.66.1 (Beta)

### OS Version

Windows Windows 11 Core 10.0.26200.0 amd64

### Description

The Google Contacts extension README.md contains outdated OAuth 2.0 setup instructions that no longer align with Google's current OAuth 2.0 configuration requirements. Users following the current README will encounter errors during setup.

### Steps To Reproduce

1. Open the Google Contacts extension README.md setup section
2. Follow the OAuth 2.0 setup steps as currently documented
3. Attempt to configure the OAuth consent screen following the instructions
4. Notice that Google Cloud Console now requires additional steps (Audience configuration) not mentioned in the README

### Current Behaviour

Users must perform additional configuration steps beyond what is documented:
- The OAuth consent screen configuration is incomplete
- The critical "Audience" setting change from Testing to Published is not mentioned
- Users may receive errors or be unable to complete setup following only the current README instructions
- Without publishing, Google will block the login because its being "tested" by the developers (yourself)

### Expected Behaviour

The README should provide complete, current OAuth 2.0 setup instructions that align with Google's latest requirements, including the step to configure the audience from Testing to Published in the Google Cloud Console

aicayzer:
Pretty sure this is the Testing → Audience → Test users step in the setup guide getting missed, not a publishing issue — see [#29276](https://github.com/raycast/extensions/pull/29276) for two FAQ entries I've added to try to head this off for the next person. (Let me know if I've misunderstood and there's something else going on — happy to amend.)

megaroeny:
> Pretty sure this is the Testing → Audience → Test users step in the setup guide getting missed, not a publishing issue — see [#29276](https://github.com/raycast/extensions/pull/29276) for two FAQ entries I've added to try to head this off for the next person. (Let me know if I've misunderstood and there's something else going on — happy to amend.)

Ah yes! I appreciate it. Some of the details of the location of things were a bit off (guessing from updates on Google's side), so I must've missed that step trying to translate things. It's working now, thanks!

Issue #29170, opened 2026-07-02
[Google Contacts] Access blocked: Google Contacts has not completed the Google verification process


megaroeny:
### Extension

https://www.raycast.com/aic/google-contacts

### Raycast Version

0.66.1.0

### OS Version

Windows Windows 11 Core 10.0.26200.0 amd64

### Description

Got this when I was trying to login to my Google account after entering my Client ID:

<img width="1126" height="606" alt="Image" src="https://github.com/user-attachments/assets/c654df6f-f0f1-4909-9d20-5247bd1753c4" />

```
Error 403: access_denied
Request details: access_type=offline scope=https://www.googleapis.com/auth/contacts response_type=code redirect_uri=com.raycast:/oauth?package_name=Extension code_challenge_method=S256 state=eyJpZCI6IjNmMWU5NWM4LTk5MmItNDZjNy1hNjZjLTNjY2RkN2E0Nzk2MiIsInByb3ZpZGVyTmFtZSI6Ikdvb2dsZSIsImZsYXZvciI6InJheWNhc3QteCIsInNjaGVtZSI6ImNvbS5yYXljYXN0In0 flowName=GeneralOAuthFlow client_id=1088146849467-3cdkl4s6pjesuubadj0oih1ksl3tqhtj.apps.googleusercontent.com code_challenge=btKrWcCC0v7n8PdBr0otIINsGsuMWnpaK6SXmpg0RNw
``` 

### Steps To Reproduce

1. Follow the Readme
2. Enter Client ID in extension settings
3. Launch a command
4. Try to Login to Google

PR #29276, opened 2026-07-06
docs(google-contacts): add FAQ entries for access_denied / Testing vs Published

aicayzer:
## Description

Adds two FAQ entries to the Google Contacts README covering the `Error 403: access_denied` OAuth error reported in #29171 (and likely the root cause of #29170 too):

1. What causes it (missing test-user setup on the OAuth consent screen)
2. Why the fix is adding a test user, not publishing the app — publishing a Contacts-scope (restricted-scope) app would trigger Google's verification review, which the bring-your-own-Client-ID setup is specifically designed to avoid.

Docs-only, no behaviour change.

Refs #29171

## Checklist


0xdhrv (approved):
Looks good to me, approved ✅

PR #29280, opened 2026-07-07
[Quick Notes] Show clear errors when opening a note externally fails

biknandy:
## Description

`Open Note Externally` (⌘O) could surface a cryptic macOS dialog — *"The application can't be opened. -50"* — because the note path was handed to `open()` without checking it exists. This happens when the auto save folder was deleted/moved since it was set, or the note's markdown file was never exported to it.

Now, before opening:

- **No auto save location set** → failure toast telling the user to set a folder in the extension settings, with a primary action that opens the extension settings directly
- **Folder set but missing on disk** → failure toast with the same settings shortcut
- **Note file missing** → failure toast suggesting saving the note or running `Sync with Folder` first
- `open()` is awaited in a try/catch so any remaining failure shows a toast instead of a macOS error dialog

The `Sync with Folder` command's existing "No Auto Save Location Set" toast gets the same message and open-settings action for consistency.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist

### DIFF
--- a/extensions/quick-notes/CHANGELOG.md
+++ b/extensions/quick-notes/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Quick Notes Changelog
 
-## [Clear Errors for Open Note Externally] - {PR_MERGE_DATE}
+## [Clear Errors for Open Note Externally] - 2026-07-07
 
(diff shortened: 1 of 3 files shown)


pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #29283, opened 2026-07-07
Fix Clean Text active app paste

MoisesCervera:
## Summary

- remove Clean Text's explicit frontmost application lookup before rendering the paste action
- use Raycast's standard `Clipboard.paste()` path for active-app paste behavior
- add a shared paste fallback that copies the cleaned text if pasting fails

## Why

Clean Text rendered a paste action by calling `getFrontmostApplication()` and using the returned app path as a file icon. When Raycast cannot resolve the active application's bundle, the command can fail with `Cannot get bundle for application`.

Other text-focused extensions generally rely on `Clipboard.paste()` directly instead of resolving the frontmost app first. This keeps Clean Text aligned with that pattern and avoids the fragile bundle lookup.

## Validation

- `npm run lint`
- `npm run build`
- `npx tsc --noEmit`

### DIFF
--- a/extensions/clean-text/CHANGELOG.md
+++ b/extensions/clean-text/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Clean Text Changelog
 
+## [Fix Active App Paste] - 2026-07-07
+
+- Fixed a crash that could occur when pasting cleaned text if Raycast couldn't resolve the active app bundle.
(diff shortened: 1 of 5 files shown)


pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #29308, opened 2026-07-08
Fix Video Downloader transcript extraction failing when Download Path contains ~

ericsr:
## Description

Transcript extraction (the `extract-transcript` AI tool) fails with:

```
ENOENT: no such file or directory, mkdir '~/Downloads/.tmp-subtitles'
```

**Root cause:** the Download Path directory preference defaults to the literal string `~/Downloads`, and Raycast stores it with the tilde. `src/transcript.ts` joins it with `.tmp-subtitles` and calls `fs.mkdirSync`, but Node's fs APIs do not expand `~`. Video downloads were unaffected because the raw path is passed to yt-dlp, which expands `~` itself — so the bug only surfaces in transcript extraction.

**Fix:** expand a leading `~` (bare `~` or `~/...`; `~user/...` paths are intentionally left alone) to `os.homedir()` once in `src/utils.ts`, so every consumer of the exported `downloadPath` receives a usable filesystem path.

Note: `src/index.tsx` reads its own `downloadPath` directly from `getPreferenceValues` and passes it to yt-dlp's `-o`, which works with the raw value. That pre-existing pattern is left untouched here to keep the diff minimal, but unifying it on the `utils.ts` export would be a reasonable follow-up.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

## How has this been tested

- `npx ray build`, `npx tsc --noEmit`, and `npx ray lint` all pass.
- Reproduced the failure with the default `~/Downloads` preference value, applied the equivalent change to the installed extension bundle, and confirmed transcript extraction succeeds afterwards (macOS, yt-dlp 2026.7.4).

## Checklist

- [x] I read the [documentation about contributing to an existing extension](https://developers.raycast.com/basics/contribute-to-an-extension)

https://claude.ai/code/session_01Ff9RFZdiz1wxCSVWY5LTgZ


0xdhrv (approved):
Looks good to me, approved ✅

PR #29316, opened 2026-07-09
Update Beszel extension (fix system detail fields, add metrics)

wyattjoh:
## Description

Beszel `2.0.1`. This update repairs the system detail panel, which regressed against recent Beszel agent versions.

Recent Beszel agents omit several `info` fields the extension previously treated as guaranteed (hostname `h`, kernel `k`, CPU cores `c`, CPU model `m`) and report network throughput via `bb` (bytes) instead of the deprecated `b` (MB). As a result the detail view rendered `undefined` for Network, Chip, Kernel, and Hostname.

Changes:

- Network reads the current `bb` field and displays a human readable rate
- Hostname falls back to the connection host when the agent does not report one
- Chip and Kernel render only when the agent provides them
- Added GPU, load average, temperature, battery, and extra filesystem metrics when reported by the agent
- Updated the `get-system` AI tool to the same field mapping
- Bundled with a dependency upgrade to latest and a migration to the ESLint flat config

## Checklist

### DIFF
--- a/extensions/beszel-extension/.eslintrc.json
+++ b/extensions/beszel-extension/.eslintrc.json
@@ -1,4 +1,0 @@
-{
-  "root": true,
-  "extends": ["@raycast"]
-}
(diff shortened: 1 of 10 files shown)


0xdhrv (approved):
Looks good to me, approved ✅

PR #29337, opened 2026-07-09
Update wishapp extension

DrGains:
## Description

Update to the existing **WishApp** extension (author `bilal_k`). No new commands, no new dependencies, no manifest changes. This is a correctness and cross-platform pass over the two existing commands.

### Windows fixes

The extension declares `"platforms": ["macOS", "Windows"]`, but several things were macOS-only in practice:

- **Shortcuts did not bind on Windows.** Every custom shortcut used the flat `{ modifiers: ["cmd"], key: "…" }` form. Per the [Keyboard docs](https://developers.raycast.com/api-reference/keyboard), `cmd` is an "ambiguous" modifier and must be declared per-platform. All shortcuts now use the `{ macOS, Windows }` form, and Refresh/Copy use `Keyboard.Shortcut.Common.Refresh` / `Common.Copy`. (Copy moves from `⌘.` to the standard `⌘⇧C`; `Ctrl+.` on Windows is Raycast's own Pin shortcut.)
- **On-screen hints showed `⌘` glyphs to Windows users.** They now render as `Ctrl+Enter` / `Ctrl+Shift+P`.
- **Image uploads silently skipped optimization for non-ASCII paths.** `runPowerShellScript` pipes the script into `powershell.exe -Command -`, and Windows PowerShell decodes stdin using the console's OEM code page rather than UTF-8, so a temp path under e.g. `C:\Users\Søren\…` arrived mangled and GDI+ threw. Paths are now base64-encoded, keeping the generated script pure ASCII.

Tested on Windows 11.

### Correctness fixes

- **Reservation privacy.** A wishlist can be set to hide reservations from its owner so gifts stay a surprise. The reservation UI respected that, but `List.Item`'s `keywords` did not — searching `reserved` filtered the list down to exactly the reserved items. The keyword is now gated on the same flag, and item previews no longer name who reserved what (matching the website and mobile app).
- **Duplicate error toasts.** `useCachedPromise` already shows a failure toast unless `onError` is passed. A separate `useEffect` was showing a second one. Now handled via `onError` + `showFailureToast`.
- **URL validation matched the server.** The scraper requires `https` via a literal `startsWith("https://")`. `new URL(x).protocol === "https:"` accepts `https:/x.com` and `HTTPS://X.COM`, which the server then rejects. The normalized `url.href` is now what gets sent and stored.
- Product images served over `http://` now load in previews; wishlist previews no longer break when the image URL already carries a query string.

### Cleanup

Deduplicated the shared empty state and the wishlist-fetching hook, collapsed the extension/MIME tables in `upload.ts` to a single source of truth, and split constants and formatting helpers out of `types.ts` / `image.ts`. Net ~100 fewer lines. No behaviour change.

## Screencast

Existing screenshots in `metadata/` are unchanged and still accurate — the UI is the same, aside from the corrected shortcut hints.

## Checklist

### DIFF
--- a/extensions/wishapp/CHANGELOG.md
+++ b/extensions/wishapp/CHANGELOG.md
@@ -1,5 +1,17 @@
 # WishApp Changelog
 
+## [Fixes and Cleanup] - {PR_MERGE_DATE}
+
+- Item previews no longer name who reserved an item, matching the website and the mobile app, which only ever show how many are reserved
+- Fixed searching a wishlist for "reserved" filtering to the reserved items even when that wishlist is set to keep reservations hidden from you
+- Product URLs must now start with `https://`, so an unsupported link is caught in the form instead of failing against the server
+- Fixed product images served over `http://` not loading in previews
+- Fixed wishlist previews rendering a broken image when the image URL carried a query string
+- Fixed keyboard shortcuts not working on Windows: every shortcut now declares its macOS and Windows binding, and Copy uses the standard cross-platform shortcut
+- Fixed on-screen shortcut hints showing macOS-only ⌘ symbols to Windows users
+- Fixed image uploads skipping optimization when the file path contained a non-ASCII character, such as a temp folder under a user name with an accent
+- Fixed network errors showing two stacked toasts instead of one
+
 ## [Initial Version] - 2026-07-09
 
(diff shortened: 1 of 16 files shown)


0xdhrv (approved):
Looks good to me, approved ✅

Issue #29320, opened 2026-07-09
[Claude Code Usage (ccusage)] Menubar Usage Limit is not refreshed


alexandrelissowski:
### Extension

https://www.raycast.com/nyatinte/ccusage

### Raycast Version

0.67.1.0

### OS Version

macOS Tahoe 26.5.2 arm64

### Description

The Raycast command shows 11% today's usage. 
The menubar shows 2%. 
The Claude settings show 12%. 

I would like that the menubar shows the same usage as the command. 

<img width="2000" height="1250" alt="Image" src="https://github.com/user-attachments/assets/79771709-b044-49d2-98f3-42ae9a90d389" />

<img width="975" height="732" alt="Image" src="https://github.com/user-attachments/assets/641468f1-69ef-486c-9ad2-f83b7cde1d34" />

### Steps To Reproduce

1. Enable Menubar ccusage
2. Launch ccusage command in Raycast
3. Compare values

### Current Behaviour

Figures doesn't match. 

### Expected Behaviour

have the same figures on menubar and command.

bendrucker:
This is not an actionable issue. Not reproducible as written.

> The Raycast command shows 11% today's usage.
> The Claude settings show 12%.

This isn't a bug regardless, `ccusage` does local computation, Claude reports from the server. A difference that is attributable to other devices/clients or trivial math differences is not something this extension can fix.

alexandrelissowski:
You are right. I'm sorry when I reported this issue, it annoyed me. But after few usage of Claude, it seems that it is solved and doesn't happen anymore. 

Thanks

bendrucker:
Took a quick look at whether there might be any obvious issue and did find some suspects:

https://github.com/raycast/extensions/pull/29338

This is _not_ going via `ccusage` as I originally thought. But it is subject to a 5m refresh.

PR #29338, opened 2026-07-09
ccusage: reconcile menu bar usage readouts

bendrucker:
## Description

The menu bar and the main command disagreed about Claude usage: the reporter saw the menu bar read 2% while the command read 11% and Claude's own settings read 12% for the same day. Both surfaces already draw from one source (`useClaudeUsageLimits` → the `usage-limits-cache` singleton → Claude's OAuth usage API), so this was framing and freshness, not two calculations of one number.

Two fixes:

- **Framing.** `showRemainingUsage` is now a single extension-level "Usage Display Mode" preference that governs the menu bar title, the progress bars, and the command's Usage Limits accessory together. It defaults to Consumed, so every rate-limit percentage matches Claude's settings out of the box, and flipping to Remaining flips all of them. Previously the preference was menu-bar-scoped, titled "Progress Bar Mode", and defaulted to Remaining, so the numeric title showed `100 − util` while the command accessory and Claude showed `util`.
- **Freshness.** `lastFetched` was only set on a successful in-process fetch and never written to the cache, so every menu bar relaunch restored it to `null` and flagged the restored blob stale, and the command's "Last Updated" read blank on a fresh process. It now persists alongside the data blob and restores on load, and staleness is derived from real blob age against the user's configured refresh interval.

### Decisions

- Rejected the narrower fix of hardcoding the title to consumed while the preference governed only the bars: that left the title and bars able to disagree within the same menu. One toggle now covers every at-a-glance readout.
- Rejected making the preference diverge from Claude by default. Consumed is the default so the out-of-the-box reading matches Claude's settings, which is what the reporter compared against.
- The command's expanded detail metadata keeps its precise "Utilization" (consumed) labels. Those rows are a consumption analytics panel, and relabeling them as remaining would be wrong.

### Migration

`showRemainingUsage` moved from the menu bar command's preferences to an extension-level preference, so an existing Remaining setting resets to the new Consumed default. Noted in the changelog.

Verified with the extension's built-in mock (`five_hour.utilization = 28`): default Consumed shows `28%` in both the menu bar title and the command accessory (matching Claude), and Remaining shows `72%` in both.

Reported in #29320.

## Screencast

Not applicable. There is no new command or view. The change is a percentage that flips framing (5-Hour utilization renders the same value the command and Claude already show) and a freshness timestamp that was previously blank on a fresh process. Behavior is verified with the extension's built-in mock, as described above.

## Checklist

### DIFF
--- a/extensions/ccusage/CHANGELOG.md
+++ b/extensions/ccusage/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Claude Code Usage (ccusage) Changelog
 
-## [Reconcile menu bar usage readouts] - {PR_MERGE_DATE}
+## [Reconcile menu bar usage readouts] - 2026-07-10
 
 ### Fixed
 
--- a/extensions/ccusage/CHANGELOG.md
+++ b/extensions/ccusage/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Claude Code Usage (ccusage) Changelog
 
+## [Reconcile menu bar usage readouts] - {PR_MERGE_DATE}
+
+### Fixed
+
+- The menu bar limit readouts (5-Hour, 7-Day, Highest Utilization) now show consumed usage, matching the main command and Claude's settings. The "Progress Bar Mode" preference continues to govern only the progress bars
+- The last-fetched time now persists across relaunches, so a fresh menu bar process reports the real age of its data in "Last Updated" and the stale-data warning instead of a blank timestamp
+
 ## [Honor server rate-limit backoff] - 2026-06-10
 
 ### Fixed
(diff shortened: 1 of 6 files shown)


bendrucker:
Addressed the `isLoading` P2. It is now initialized to `true` only when a fetch will actually run (`restoredRateLimitedUntil === null && (restoredData === null || isBlobStale(restoredLastFetched))`), so a fresh relaunch with a still-fresh blob no longer reports a spurious loading state (c6154d6).

Note that the framing approach changed since this review. Rather than hardcoding the title to consumed, `showRemainingUsage` is now a single extension-level "Usage Display Mode" preference (default Consumed) that governs the menu bar title, the progress bars, and the command's Usage Limits accessory together, so they stay consistent in either mode (7fac685).

0xdhrv (approved):
Looks good to me, approved ✅

PR #29341, opened 2026-07-10
Improve Open Chat performance with caching and contact photo hydration

IslamTayeb:
## Summary
- Speeds up Open Chat with stale-while-revalidate chat catalog caching and contact-name map persistence.
- Restores contact photos: paint immediately from the persisted photo cache, then refresh visible + catalog photos in the background (including already-cached IDs).
- Adds Swift `fetchContactPhotosForContactIds` for photo-only lookups (names still match with `loadPhotos: false`).
- Removes extension-local test scripts (`scripts/` + `test:*` npm entries); no benchmarks.

## Test plan
- [ ] Open Chat with a warm cache: contact avatars appear immediately from cache (not initials-only).
- [ ] Open Chat with a cold cache / after clearing Raycast cache: names then photos hydrate without blanking the list.
- [ ] Toggle **Load Contact Photos** off: initials/generated avatars only; groups still show group photos when available.
- [ ] Search/filter chats; photos continue to paint for visible rows.
- [ ] Confirm no `scripts/` folder or `test:*` / benchmark npm scripts ship with the extension.


IslamTayeb:
Closing to reopen a cleaned-up PR with full cache (catalog + names + photos) after stripping benchmark/test artifacts.

PR #29343, opened 2026-07-10
[Is It Alive?] Fix site ID generation

jchristianhall:
## Description

Wasn't able to get this working when I installed it and Codex suggested this. Tested it out locally and the extension works!

Fixes adding monitored sites in Raycast environments where the Web Crypto global is unavailable.

## Root cause

The extension called `crypto.randomUUID()` as a global. Importing `randomUUID` from `node:crypto` explicitly uses Raycast's managed Node runtime.

## User impact

Adding any status page currently fails before the site can be saved.

## Reproduction

1. Open **Check Status Pages**.
2. Add `https://status.claude.com`.
3. Submit the form.
4. Observe `crypto is not defined`.

**Before**

Raycast showing the crypto is not defined error

**After**

Raycast showing the Claude status page working locally

## Validation

- Reproduced the Store build failure with `https://status.claude.com`
- Verified the local development build can add the same status page
- `npm run lint`
- `npm run build`

### DIFF
--- a/extensions/is-it-alive/CHANGELOG.md
+++ b/extensions/is-it-alive/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Is It Alive? Changelog
 
-## [Fix Site Creation] - {PR_MERGE_DATE}
+## [Fix Site Creation] - 2026-07-10
 
 - Fix adding sites when the Web Crypto global is unavailable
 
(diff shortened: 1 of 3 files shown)


0xdhrv (approved):
Looks good to me, approved ✅

PR #29345, opened 2026-07-10
Add cl1p extension

meslva:
## Description

New extension: **cl1p.net** — saves text to [cl1p.net](https://cl1p.net) and copies the resulting URL to the clipboard, so it can be pasted into a chat or another machine's browser. By default a cl1p is destroyed the moment someone opens it, which makes it a convenient one-time channel for moving a snippet between devices.

The extension ships two commands:

| --- | --- | --- |
| **Save to cl1p.net** | `view` | A form with a title and a multi-line content field. On submit, saves the content and copies `https://cl1p.net/<title>` to the clipboard. |
| **Save Selected Text to cl1p.net** | `no-view` | Takes the title as a command argument, reads the text currently selected in the frontmost app, saves it, and copies the URL — no form, no window. |

Both commands show a success HUD/toast with the copied URL, and a failure toast when the title is already taken, the token is rejected, or the request fails.

### Preferences

One required preference, `API Token` (`password` type), holding the cl1p.net API access token. Users generate it from the Preferences page on cl1p.net after creating a free account; the README documents the steps. Free accounts allow up to 20 writes per day.

### Notes for the reviewer

- Only runtime dependency is `@raycast/api` — the API is a single `fetch` call, so no HTTP client was added.
- `Preferences` and `Arguments.SaveSelected` use the auto-generated types from `raycast-env.d.ts` rather than hand-written interfaces.
- The title is `encodeURIComponent`-escaped in both the request URL and the URL copied to the clipboard, so titles containing spaces or symbols produce a valid link.
- No text is stored anywhere by the extension itself — it is a straight pass-through to cl1p.net.

## Screencast

**Both commands in the root search**

Root search

**Save to cl1p.net — note name and multi-line content**

Save form

**Save Selected Text to cl1p.net — note name entered as a command argument**

Save selected text

## Checklist


0xdhrv (approved):
Looks good to me, approved ✅

PR #29351, opened 2026-07-10
Add ai-quota-usage extension

Bo-Feng-1024:
## Description

Adds **AI Quota Usage** — a single-command extension that shows your **Claude Code** and **Codex** subscription quota (how much is left, when each window resets) plus recent usage, in one list.

- **Claude Code** — live quota (5-hour, weekly, and per-model weekly) from the `oauth/usage` endpoint, read from your current Claude Code login. No setup needed. If that login is expired or lacks the required scope, the row says exactly that and asks you to re-login (the extension never refreshes tokens for you).
- **Codex** — live 5-hour and weekly quota from the `wham/usage` endpoint using your current Codex login. If the request fails, the latest local session-log snapshot is shown as a fallback, with days-old snapshots marked stale.
- **Usage / cost** — today and this week (tokens + $) for both tools, via `ccusage`.
- **Low-quota warning** — any window under a configurable threshold (default 20%) is highlighted in red.

Everything is **read-only** and fetched only when you open the command (never polled in the background); `oauth/usage` results are cached ~5 min because that endpoint rate-limits hard.

Works with zero configuration for most users. Advanced preferences — Claude/Codex data directories, an optional Claude OAuth token, and a custom `npx` path — are available as escape hatches but are not required.

## Screencast

**Claude Code**

Claude Code quota

**Codex**

Codex quota

## Checklist


0xdhrv (commented):

* [Agent Usage](https://www.raycast.com/thuggyduck/agent-usage) by @natsustan

Both extensions monitor Claude Code and Codex subscription usage by reading credentials from the users’ existing CLI logins and calling the same usage endpoints:

* Claude Code: `https://api.anthropic.com/api/oauth/usage`
* Codex: `https://chatgpt.com/backend-api/wham/usage`

Both show the remaining percentage and reset time for the short and weekly quota windows. Agent Usage also supports these providers alongside Amp, Copilot, Cursor, Gemini, Grok, Kimi, Droid, and several other coding agents, with both list and menu-bar commands.

This PR adds some useful improvements:

* Today and weekly token/cost totals through `ccusage`
* A local Codex session-log fallback when the live endpoint fails
* Configurable low-quota warnings
* More explicit expired-token, missing-scope, stale-data, and rate-limit states
* A deliberately read-only approach that does not refresh or modify CLI credentials

These would be valuable enhancements to the existing Agent Usage extension rather than enough differentiation for another extension focused on the same Claude Code and Codex quota data.

Could we consider contributing the usage-cost rows, fallback handling, warning threshold, and read-only credential behavior to Agent Usage instead?

> This would help avoid duplication and keep AI coding-agent usage monitoring consolidated in one place.

PR #29379, opened 2026-07-11
Claude Code Usage (ccusage): keep rate-limit progress bars visible during backoff

kud:
## Description

Fixes a regression where the menu bar's **Rate Limits** section (the progress bars) disappears intermittently.

Since the "Honor server rate-limit backoff" change, `fetchUsageLimits()` early-returns at the top when a backoff window is active (`rateLimitedUntil`, restored from cache). That early return happens _before_ the line that sets `isUsageLimitsAvailable = true`. Because the menu-bar command is a short-lived process that re-initialises module state on every tick, any cold start during a backoff window leaves `isUsageLimitsAvailable` at its initial `false` — and the whole section is gated on that flag in `menubar-ccusage.tsx`, so the bars vanish even though valid cached limit data is available.

**Fix:** seed `isUsageLimitsAvailable` from the restored cache (`restoredData !== null`) in the initial `cacheState`. If cached limit data exists, a token was valid on the previous run, so the feature is available and the (stale) bars should render immediately instead of waiting for a fetch the backoff guard will skip. The normal fetch path still corrects the flag either way once the window clears.

**Testing:** `npm run typecheck`, `ray lint`, and `ray build` all pass.

## Screencast

N/A — internal state fix; no visual change other than the bars remaining visible during a backoff window.

## Checklist

### DIFF
--- a/extensions/ccusage/CHANGELOG.md
+++ b/extensions/ccusage/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Claude Code Usage (ccusage) Changelog
 
-## [Keep rate-limit bars visible during backoff] - {PR_MERGE_DATE}
+## [Keep rate-limit bars visible during backoff] - 2026-07-12
 
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


0xdhrv (approved):
Looks good to me, approved ✅

PR #29388, opened 2026-07-12
Add quick-web-search extension

RamiruWanigathunga:
## Description

Adds **Quick Search** — a Raycast extension for searching the web across multiple engines from one command. Query Perplexity, Google, DuckDuckGo, Bing, or YouTube with live search suggestions as you type.

Features:

- **Quick Search** — one command for all five engines; switch engines with a dropdown that remembers your last selection
- **Live suggestions** — autocomplete suggestions fetched as you type
- **Search history** — recent searches shown when the search bar is empty (can be disabled in preferences)
- **Hotkey-friendly** — assign a hotkey (e.g. ⌘P), press it, type, hit Enter
- Supports macOS and Windows

## Screencast

r0
r2
r3
r4
r5

## Checklist


RamiruWanigathunga (commented):
Hey @0xdhrv , could you please take a look at my codebase and share your feedback? Thanks a lot! 🙌

0xdhrv (commented):

We already have an extension in the Store that covers this core functionality. Could we consider enhancing the existing extension below instead of creating another one?

* [Universal Website Search](https://www.raycast.com/pernielsentikaer/any-website-search) by @pernielsentikaer

Both extensions provide a single command for searching multiple websites or search engines, switching the active destination from a dropdown, fetching live suggestions from Google or DuckDuckGo, and opening the selected query in the browser. Both also support macOS and Windows.

The existing extension is more flexible because users can add, edit, and remove arbitrary websites rather than being limited to a predefined set of engines. This PR adds some useful workflow improvements, including recent-search history, remembering the last selected engine, fallback-command behavior, and quick **Open With** actions. However, these feel like natural enhancements to **Universal Website Search** rather than enough differentiation for a separate extension with substantially the same search experience.

Adding history, persisted site selection, fallback handling, and cross-site actions to the existing extension would preserve the improvements from this PR while also making them available for every user-configured website.

RamiruWanigathunga:
Thanks for the feedback @0xdhrv  🙏

I understand the overlap with Universal Website Search, but my intention is to keep building this extension into a more exclusive and faster way to search the web.

Beyond the current features (history, persisted engine selection, fallback handling, quick Open With), I’m planning to add support for adding/editing/removing arbitrary websites, advanced workflows, and performance-focused improvements. 
My goal is to make this extension stand out with unique capabilities rather than duplicate what already exists.

I’d like to continue developing it as a separate extension so users who want a streamlined, high‑performance search experience can benefit directly.

0xdhrv:
> Thanks for the feedback @0xdhrv 🙏
> 
> I understand the overlap with Universal Website Search, but my intention is to keep building this extension into a more exclusive and faster way to search the web.
> 
> Beyond the current features (history, persisted engine selection, fallback handling, quick Open With), I’m planning to add support for adding/editing/removing arbitrary websites, advanced workflows, and performance-focused improvements. My goal is to make this extension stand out with unique capabilities rather than duplicate what already exists.
> 
> I’d like to continue developing it as a separate extension so users who want a streamlined, high‑performance search experience can benefit directly.

Thanks for clarifying your plans 🙏

However, we need to evaluate the extension based on the functionality included in the PR today rather than features that may be added later.

At the moment, the core workflow still substantially overlaps with Universal Website Search:

* selecting a website or search engine from a dropdown
* entering a query in a single command
* showing live search suggestions
* opening the search in a browser
* supporting both macOS and Windows

Universal Website Search already supports adding, editing, and removing arbitrary websites, so the planned custom-website functionality would also overlap with functionality that is already available.

The remaining differences recent search history, remembering the selected engine, fallback behavior, and a quick “Open With” action are useful conveniences, but they do not appear to provide a sufficiently different user outcome to justify a separate extension. Performance-focused improvements would also need to be demonstrated concretely rather than used as a product distinction on their own.

Could we consider contributing these improvements to Universal Website Search instead? Search history, persisted website selection, and improved fallback-command handling could all strengthen the existing extension while keeping this workflow consolidated and easier for users to discover.

For us to reconsider it as a separate extension, the PR would need to include a materially different workflow or capability that cannot reasonably fit into Universal Website Search not only planned additions or a more streamlined implementation.

PR #29412, opened 2026-07-13
Update WinSCP extension

THEN00P:
## Description

Sessions were only ever read from `WinSCP.ini` next to the executable, which is not where WinSCP puts them by default. On a default install the extension therefore showed an empty list, or the "not configured" error, until the user manually switched WinSCP over to INI storage.

This PR reads sessions from the places WinSCP actually stores them, and fixes the session parsing and launching along the way.

**Storage**

- Read sessions from the **registry** (`HKCU\Software\Martin Prikryl\WinSCP 2\Sessions`), which is WinSCP's default storage. They are read through PowerShell's registry provider rather than `reg.exe`, which is commonly blocked by policy on managed machines. The script sticks to cmdlets so it also works under AppLocker's ConstrainedLanguage mode, and the JSON is written to a file rather than stdout so that host and user names containing non-ASCII characters do not come back mangled by the console code page.
- Read `WinSCP.ini` from **`%APPDATA%`**, the default INI location, which was previously never checked. A portable `WinSCP.ini` next to the executable still takes precedence, matching WinSCP's own lookup order.
- Detect WinSCP installed in `Program Files`, or via a package manager such as Scoop or Chocolatey, in addition to the per-user install folder. Setting the installation folder preference by hand is now only needed for unusual locations, and its description says so.

**Sessions**

- Launch sessions by their **stored identifier** rather than their display name, so sessions whose name contains a space or other encoded character (`My%20Site`) now open instead of failing.
- Show the **protocol** in the subtitle (`sftp://user@host`), and no longer render `undefined@host` for sessions saved without a user name.
- List **workspaces once** instead of once per session they contain, and mark them with their own icon.
- Launch `WinSCP.exe` directly instead of going through `start` in a shell.

**Also**

- Added a *Launch in New Instance* action (`⌘⇧↵`) and a *Refresh Sessions* action (`⌘R`).
- A registry read that fails now reports the failure with a way to work around it, rather than silently showing an empty list that reads as "you have no sessions".
- Added unit tests for the INI/registry parsing and the path lookup (`npm test`, 29 tests). Parsing and path resolution are pure functions in `src/winscp/`, so they are tested without touching the registry or the filesystem.

## Checklist

### DIFF
--- a/extensions/winscp/CHANGELOG.md
+++ b/extensions/winscp/CHANGELOG.md
@@ -1,5 +1,16 @@
 # WinSCP Changelog
 
+## [Registry Sessions, Protocols and Workspace Fixes] - {PR_MERGE_DATE}
+
+- Added support for sessions stored in the registry, which is where WinSCP keeps them by default. They are read through PowerShell's registry provider rather than `reg.exe`, which is often blocked by policy on managed machines
+- Added support for the `WinSCP.ini` in `%APPDATA%`, which previously was never read even though it is the default INI location
+- Added detection of WinSCP installed in `Program Files` or through a package manager such as Scoop or Chocolatey
+- Sessions are now launched with their stored identifier, so sessions whose name contains spaces or other encoded characters open correctly
+- Session subtitles now show the protocol (`sftp://user@host`) and no longer show `undefined@host` for sessions without a user name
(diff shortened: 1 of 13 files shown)


0xdhrv (approved):
Looks good to me, approved ✅

PR #29432, opened 2026-07-14
google-chrome-profiles: spawn detached osascript before showing HUD

smpeotn:
## Description

Fixes profile actions (Bring to Front, New Tab, New Window, Open URL) still silently failing in the store build after the detached-spawn fix (#27615): the HUD appears but no Chrome action happens.

**Root cause:** `openGoogleChrome` awaited `showHUD` *before* spawning the detached `osascript` subprocess. `showHUD` closes the main window, which starts the extension process teardown, so in the distribution build the Node process could be killed before the `spawn` call ever ran. Dev mode keeps the extension process alive, which is why this never reproduces under `npm run dev` — only with the distribution build (and thus in the store version).

**Fix:** spawn the detached subprocess first (synchronously, before any `await`), then show the HUD. `runDetachedAppleScript` now returns whether the spawn succeeded, and the HUD is skipped on failure so the failure toast stays visible.

**Verification (distribution build, no dev server):**
- Before: invoking an action left no temp script file behind — the process died before `writeFileSync`/`spawn` ran. The generated AppleScript itself runs fine (~0.9s) when executed manually.
- After: with temporary debug logging in the dist build, the temp script is written, `osascript` is spawned detached, and exits with code 0 in ~600ms; the profile actually switches. Also verified by replacing the store-installed extension's files with this build and exercising the actions through the store install path.

## Screencast

Straightforward ordering fix; happy to add one if needed.

## Checklist

### DIFF
--- a/extensions/google-chrome-profiles/CHANGELOG.md
+++ b/extensions/google-chrome-profiles/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Google Chrome Profiles Changelog
 
+## [Fix] - {PR_MERGE_DATE}
+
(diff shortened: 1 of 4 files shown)


0xdhrv (approved):
Looks good to me, approved ✅

frouo:
Thank you guys!

PR #29449, opened 2026-07-14
Add xkeen-manager extension

viktorbrs:
## Checklist


pernielsentikaer (changes requested):
Hi 👋

A few things before merge as I see it, there might be more after fixes

Bugs / UX  
1. `ProfilesList` — please destructure `error` from the data hook. On SSH failure the UI currently falls through to “No profiles yet,” which reads as an empty healthy state rather than a connection problem. Prefer a failure empty state + `Retry` / `Setup SSH`.  
2. `IpDetail` — same pattern: on failure, `metadata` can stick on “Loading…”. Surface `error` with a clear message + `Refresh` (same idea as `Health`/`Logs`).

Store packaging  
3. Please drop internal design/plan material from the published tree: `docs/superpowers/**`, `docs/CHANGELOG-v2.md`, and `create_screenshots.py`. Keep root `CHANGELOG.md` + `EN`/`RU` `READMEs`.  
4. `Metadata` screenshots look like placeholder cards from `create_screenshots.py`. Store listings work much better with real `Raycast` UI captures (e.g. main hub, `Profiles`, `Quick Add`, `Health` / menu bar) at the usual `2000×1250` `16:10` size.

Docs / setup  
5. `README` step that says `SSH` alias `keenetic` doesn’t match the preference default and wizard default (`xkeen`). Please align.  
6. `assets/setup.sh` — checked; fine as a readable setup script (not a vendored binary). Two soft notes: (a) wizard copy is Russian-only while the extension is bilingual in `docs/UI`; bilingual or `EN` prompts would help non-`RU` users; (b) it may create a passphraseless `id_ed25519` and append to `~/.ssh/config` — worth a one-line note in the `README` so users know what the wizard does locally.

Happy to re-review once those are in. Great foundation overall.

---

viktorbrs:
Thanks for the detailed review! All points are addressed:

**1. `ProfilesList` — failure state.** `error` is now destructured from the hook. On SSH failure the list shows a "Connection error" item (orange warning icon, truncated error message) with `Retry`, `Setup SSH Connection` (opens setup instructions) and `Copy Error` actions. When cached profiles exist, a "Connection error — showing cached data" banner is prepended above them, so stale data never reads as fresh. "No profiles yet" is now shown only when there really are no profiles and no error.

**2. `IpDetail` — same pattern.** `error` is surfaced: the metadata panel is hidden on failure and the error is rendered in the markdown body (same style as `Health` / `Logs`), with `Refresh` and `Copy Error`. Placeholder values no longer stick on "Loading…" — they render as `—`, and the VPN-active check no longer depends on that literal.

**3. Internal docs removed.** `docs/superpowers/**`, `docs/CHANGELOG-v2.md` and `create_screenshots.py` are gone from the tree; only root `CHANGELOG.md` + EN/RU READMEs remain. The stray files came from local scratch that wasn't in `.gitignore` — that's fixed now, so they won't reappear.

**4. Screenshots.** These are real Raycast captures (Window Capture at 2000×1250), not generated cards — `create_screenshots.py` was an unused local artifact that never produced anything in `metadata/`, which is what made them look suspect. Sorry for the confusion; the script is deleted. Happy to reshoot or add a menu bar capture if you'd still like a different set.

**5. README alias.** Both READMEs now say `xkeen`, matching the preference default and the wizard default.

**6. `assets/setup.sh`.** (a) All prompts and messages are bilingual now (EN first, RU second). (b) Both READMEs document what the wizard does locally: generates a passphraseless `~/.ssh/id_ed25519` if missing, copies the public key via `ssh-copy-id`, and appends a `Host` block with `ControlMaster` settings to `~/.ssh/config` — nothing else on the Mac is modified.

Tests (155) and lint are green, and the build was verified in Raycast. Ready for another look 🙏

pernielsentikaer (approved):
Looks good to me, approved 🔥

Issue #3414, opened 2022-11-04
[Night Light] `Change Night Shift Color Temperature` command does not work


hrvstr:
# Extension – [Night Light](https://github.com/raycast/extensions/tree/f13c0b4a5be51c905ac3f994b416897671ea3f8c/extensions/night-light/)
Author: @devmoath 

## Description

When trying to change the color temperature with this Raycast extension nothing happens. Switching and toggling Night Shift does work. Setting the temperature directly with the `nightlight` package in the terminal works, too.

## Steps To Reproduce

1. Set the color temperature to `0` manually via `nightlight temp 0` in the terminal
2. Run Raycast command `Change Night Shift Color Temperature`
3. Set the value to `100` and press `Enter` - Nothing happens
4. Set the color temperature to `100` manually via `nightlight temp 0` in the terminal - Temperature changes

## The current behavior
The color temperature does not change.

## The expected behavior
The color temperature does change.

### Raycast version
Raycast v1.42.2
Extension v1
Nightlight v0.3.0

pernielsentikaer:
@devmoath do you have time to check this?

devmoath:
@pernielsentikaer yeah sure.

Hi @hrvstr, I followed the steps you provided and the command is working as expected.

NOTE: this command `Change Night Shift Color Temperature` will just change the temp color without activate the night shift. at least this is the implementation in the mean time.

hrvstr:
Hey @devmoath,

Thanks for the reply. I just did some more testing and found out that it is a different problem. I didn't realize that I can't confirm the new temperature value by just hitting `ENTER`. I have to press `CMD` + `ENTER`. Then it is working as expected. I am not sure if this is intentional or not but it kind of confused me as it is not really a "secondary action" like in other parts of Raycast where this key combo is used.

devmoath:
@hrvstr yeah you are correct, I will see if I can change that

devmoath:
Hi @pernielsentikaer, I'm reading the docs to make the submit form uses `ENTER` instead of `CMD` + `ENTER`, but I did not find anything related to that. 

also I'm searching for example from other extensions and I find out this, The create issue command from [GitHub Beta](https://www.raycast.com/raycast/github-beta) shows one button in the store screenshots but when downloaded its still uses `CMD` + `ENTER` for the form submission (see images bellow).

My question: is this possible to do ? if so please direct me to it please.

[...]

PR #29450, opened 2026-07-14
fix(night-light): apply color temperature before dismissing view

alankyshum:
## Description

Fixes the **Change Night Shift Color Temperature** command in the Night Light extension, which did not actually change the color temperature.

The submit handler dismissed the view (`popToRoot` / `closeMainWindow`) **before** awaiting the `nightlight` CLI call. Because dismissing the view tears down the command's process, the `await nightlight(\`temp ${temperature}\`)` line never executed, so the temperature was never applied. Toggling Night Shift on/off worked because those are `no-view` commands that run the CLI first.

This was independently verified against the `nightlight` CLI as source of truth: running the extension's exact `execSync` path changes the value (`nightlight temp` reads back the new value), while the shipped ordering left it unchanged.

### Fix

Run the `nightlight` command **before** dismissing the view:

```tsx
onSubmit={async ({ temperature }) => {
  if (validate(temperature)) {
    await nightlight(`temp ${temperature}`, `Set color temperature to ${temperature}`);

    if (closeWindow) closeMainWindow();
    else await popToRoot({ clearSearchBar: true });
  }
}}
```

Refs the earlier report of this in #3414.

## Screencast

Straightforward one-line reordering; no UI change. Behavior verified via the `nightlight` CLI (`nightlight temp 10` → run command with `55` → `nightlight temp` now reports `55`; previously stayed `10`).

## Checklist


0xdhrv (approved):
Looks good to me, approved ✅

PR #29471, opened 2026-07-15
Update read-this-later extension

shearmds:
## Checklist


shearmds:
## Description

An update to my existing extension, published as `read-this-later`.

**Renamed to "Research Sync"** to match its companion browser extension and iOS app, which were rebranded a while ago — the Raycast extension was the last client still on the old name and icon. The `name` slug and both command names are deliberately **unchanged**, so the listing keeps its identity and any existing hotkeys keep working. Only display strings and the icon change.

### Added

- **Folders** — links are grouped into the folders the sync service assigns, with unsorted links last.
- **Unread / Read / All filter** — previously read links were unreachable from the extension entirely.
- **Read articles inside Raycast** (`⌘↵`). Article bodies are captured at save time and end-to-end encrypted; the extension fetches and decrypts them, then renders them as Markdown. On a link with no captured body, `⌘↵` shows its note and details instead.
- **Capture on save** — when the Raycast browser extension is connected, `Save Current Tab` also extracts the article body and uploads it. Without it, the link is still saved, just without article text.
- Note and capture-status indicators in the list, **mark as unread**, and **delete**.

### Fixed

- **Writes now round-trip fields this extension doesn't own** (`notes`, `offline`, `folder`). The sync service merges whole items last-write-wins, so any field missing from a write was erased for the browser and iOS clients too. This previously worked only by accident, and is now enforced through a single helper with a regression test.
- Publisher boilerplate ("Advertisement", "Skip Advertisement", audio-player labels) is stripped out of captured articles.

### New dependencies

- `linkedom`, `@mozilla/readability`, `dompurify` — extracting and sanitising the article from the page's live DOM.
- `turndown`, `turndown-plugin-gfm` — converting the stored HTML to Markdown for `Detail`.
- `vitest` (dev) — unit tests, run with `npm test`.

`jsdom` was the obvious choice for the first three, but it resolves `./xhr-sync-worker.js` at module load, which esbuild can't bundle: the extension builds cleanly and then throws `Cannot find module './xhr-sync-worker.js'` at runtime. `linkedom` avoids that entirely and is ~5x smaller in the bundle.

### Review feedback addressed

All four points from the automated review on this PR are fixed:

- `deleteBody` had no request timeout, unlike every other fetch, so a stalled server could hang the delete action after the tombstone had already been written. Every request now goes through one `fetchWithTimeout`, so a future fetch can't forget.
- `BASE_URL` was declared in three modules; it now lives once in `service.ts`.
- The empty state said "Nothing unread" under the All filter.
- The frozen crypto interop vector was read from a sibling checkout with a `describe.skip` fallback, so the test guarding byte-compatibility with the other clients silently skipped in CI. The vector is now vendored into `test/` and the suite always runs — breaking the HKDF salt fails three of its tests, where before it would have passed CI.

### Notes for review

- Screenshots in `metadata/` are updated to the current UI and pass `check_raycast_images.py`. They're composed rather than straight from Capture Window — see below.
- Article bodies are stored on the sync service as ciphertext only; the key is derived from the user's own sync token, which stays in Raycast's encrypted preferences.
- `npm test` runs 66 unit tests covering the sync write paths, the encryption, and the boilerplate stripping.

**A note on the screenshots, in case it's useful to the team:** Capture Window on Raycast 0.68.0.0 (beta) emits a 1500x1048 window into the 2000x1250 frame. That leaves 202px of vertical padding to share, and it's placed off-centre, so it measures ~7% at the top and fails `check_raycast_images.py`, which wants 8–17%. The window's maximum for that check is 1050px, so there's 2px of headroom. Screenshots taken with the same tool for the initial submission had a 1500x950 window and passed. These were composed manually to get the padding into range.

## Screencast

Screenshots of the three main surfaces are included in `metadata/`: the list grouped into folders with the filter dropdown, the article reader, and a link's note and details.

## Checklist

0xdhrv (approved):
Looks good to me, approved ✅

PR #29490, opened 2026-07-16
Add ssh-image-drop extension

free4416:
## Description

SSH Image Drop — send clipboard images and Finder-selected files to your
servers over SSH, and pull remote files back, without leaving Raycast.

### Commands

- **Send Clipboard Image** — sends the clipboard image over SSH and copies the remote path
- **Send File to Server** — sends files selected in Finder to a chosen server
- **Pull File from Server** — downloads the remote file whose path is in the clipboard and reveals it in Finder
- **Add Server** — registers a server; the password is stored in the macOS Keychain by default (optional SSH key install)

### Highlights

- One-click Quicklinks pinned to a server for all three transfer commands —
  deeplink hosts are re-validated against the user's known server list before
  any transfer, and file lists are never accepted via deeplink (see Threat model in README)
- Passwords never touch disk, argv, or LocalStorage — macOS Keychain + SSH_ASKPASS
- Reads hosts from `~/.ssh/config` but never modifies it (uses a separate managed include file)
- Requires macOS 13+ (SFTP-based scp)

## Screencast

Screenshots are included in the `metadata/` folder.

## Checklist


0xdhrv (commented):

* [Terminal Image Paste](https://www.raycast.com/iisweetheartii/terminal-image-paste) by @IISweetHeartII

Both extensions address the same primary workflow: making a clipboard image available inside a local or remote terminal session by saving or transferring it and providing the resulting file path.

The existing Terminal Image Paste extension already:

* Saves the clipboard image as a local file
* Syncs the image to configured SSH hosts
* Produces a path that can be used in SSH, tmux, and terminal applications
* Provides the workflow through a no-view command suitable for a hotkey

This PR expands that workflow with:

* Selecting a specific destination server instead of syncing to every configured host
* Sending Finder-selected files and folders
* Pulling remote files back to the local machine
* Server management and `~/.ssh/config` discovery
* Per-server Quicklinks
* Password storage through the operating system’s credential store
* Windows support
* More extensive path validation and transfer security

These are useful additions, but the clipboard-image-to-remote-path use case substantially overlaps with Terminal Image Paste.

Could we consider contributing target-server selection, file transfers, reverse transfers, and improved SSH credential handling to Terminal Image Paste instead?

> This would help avoid duplication and keep terminal and SSH image-pasting workflows consolidated in one extension.

free4416:
Thanks for the careful review, and for laying out the comparison so clearly.

You're right that there is overlap on one command. **Send Clipboard Image** and Terminal Image Paste's `paste-image-path` both end at "the image is on the remote host and you have a path." I don't want to pretend that away.

Where I'd push back is on scope. This isn't an image-paste extension with extras — it's an SSH file transfer extension, and the clipboard image is one of four entry points:

| Command | Terminal Image Paste equivalent |
| --- | --- |
| Send Clipboard Image | ✅ overlaps |
| Send File to Server (files **and folders**, prefilled from the Finder selection) | — |
| Pull File from Server (remote → local, reveals in Finder/Explorer) | — |
| Manage Servers (add/edit/delete, `~/.ssh/config` discovery, credential storage) | — |

The two also differ in delivery model, and I think both are legitimate. Terminal Image Paste is local-first: it saves locally, types the path, and syncs to *every* configured host in the background with a 3-second timeout, so an unreachable host never blocks the paste. This one is target-addressed: you pick a server (or bind one to a hotkey via a Quicklink), the transfer is confirmed, and the *remote* path is what lands on your clipboard. Different guarantees for different workflows.

Two practical concerns about consolidating into Terminal Image Paste:

1. **Naming and discoverability.** Folder uploads, reverse transfers, and server management don't belong under a name that describes pasting an image. Someone searching "send file to server" or "pull file from server" is unlikely to look inside "Terminal Image Paste," so the added functionality would be harder to find, not easier.

2. **It would be a replacement, not an iteration.** Terminal Image Paste is a single-file, macOS-only command. What's proposed here is 22 source files across a platform-adapter layer (macOS + **Windows**, which Terminal Image Paste doesn't target), an OS credential store path (Keychain / DPAPI via `SSH_ASKPASS`), a deeplink trust gate, and 126 unit tests covering path validation, argv construction, and the managed `ssh_config` writer. Merging that in would rewrite the extension wholesale and change what it is — which seems like a worse outcome for its users and its author than leaving it as the focused tool it is.

I'm happy to reduce the ambiguity from my side rather than leave it to users. Some options I'd gladly take:

- Lead the description and store copy with file/folder transfer rather than clipboard images, so the two are clearly distinguishable at a glance.
- Drop or rename any part of the framing you think reads as duplicative.

Would any of those address the concern? And if you'd still prefer consolidation after that, I'd like to understand how you'd want the Windows client and credential handling to fit there before I start.

0xdhrv:
> Thanks for the careful review, and for laying out the comparison so clearly.
> 
> You're right that there is overlap on one command. **Send Clipboard Image** and Terminal Image Paste's `paste-image-path` both end at "the image is on the remote host and you have a path." I don't want to pretend that away.
> 
> Where I'd push back is on scope. This isn't an image-paste extension with extras — it's an SSH file transfer extension, and the clipboard image is one of four entry points:
> 
> Command	Terminal Image Paste equivalent
> Send Clipboard Image	✅ overlaps
> Send File to Server (files **and folders**, prefilled from the Finder selection)	—
> Pull File from Server (remote → local, reveals in Finder/Explorer)	—
> Manage Servers (add/edit/delete, `~/.ssh/config` discovery, credential storage)	—
> The two also differ in delivery model, and I think both are legitimate. Terminal Image Paste is local-first: it saves locally, types the path, and syncs to _every_ configured host in the background with a 3-second timeout, so an unreachable host never blocks the paste. This one is target-addressed: you pick a server (or bind one to a hotkey via a Quicklink), the transfer is confirmed, and the _remote_ path is what lands on your clipboard. Different guarantees for different workflows.
> 
> Two practical concerns about consolidating into Terminal Image Paste:
> 
> 1. **Naming and discoverability.** Folder uploads, reverse transfers, and server management don't belong under a name that describes pasting an image. Someone searching "send file to server" or "pull file from server" is unlikely to look inside "Terminal Image Paste," so the added functionality would be harder to find, not easier.
> 2. **It would be a replacement, not an iteration.** Terminal Image Paste is a single-file, macOS-only command. What's proposed here is 22 source files across a platform-adapter layer (macOS + **Windows**, which Terminal Image Paste doesn't target), an OS credential store path (Keychain / DPAPI via `SSH_ASKPASS`), a deeplink trust gate, and 126 unit tests covering path validation, argv construction, and the managed `ssh_config` writer. Merging that in would rewrite the extension wholesale and change what it is — which seems like a worse outcome for its users and its author than leaving it as the focused tool it is.
> 
> I'm happy to reduce the ambiguity from my side rather than leave it to users. Some options I'd gladly take:
> 
> * Lead the description and store copy with file/folder transfer rather than clipboard images, so the two are clearly distinguishable at a glance.
> * Drop or rename any part of the framing you think reads as duplicative.
> 
> Would any of those address the concern? And if you'd still prefer consolidation after that, I'd like to understand how you'd want the Windows client and credential handling to fit there before I start.

> Lead the description and store copy with file/folder transfer rather than clipboard images, so the two are clearly distinguishable at a glance.

If you can do the above, that would be great

free4416:
Done — and it turned out to be in more places than the description.

`aea52b42` (Aug 7, just before your message landed) had already rewritten the README and store description to lead with file and folder transfer. Going back over it after your reply, I found three more surfaces that still led with the clipboard image, and those are fixed now in `665871e3`:

- **Command order in the manifest** — `Send Clipboard Image` was listed first, so the Store page led with it no matter what the copy said. `Send File to Server` now comes first.
- **CHANGELOG** — the first entry opened with "Send clipboard images, files, and folders". Reordered.
- **README intro line** — the one-line use case led with the screenshot example. Now it leads with pushing config files and folders.

The clipboard image is still there as one of the four commands, just no longer the thing the extension introduces itself with.

free4416:
@0xdhrv
Thanks for the fix — much appreciated.

0xdhrv (approved):
Looks good to me, approved ✅

PR #29492, opened 2026-07-16
fix(zotero): make Search Zotero concurrency-safe (follow-up to #29478)

### LINKED ISSUE #29250: [Search Zotero] memory limit: JS heap out of memory while rest of raycast works fine
### Extension

https://www.raycast.com/reckoning-dev/zotero

### Raycast Version

1.104.21

### OS Version

15.7.7 (24G720)

#### Error:
```

[kOnExit]:worker:316:26
```

### Steps To Reproduce

1. Search any entry in Zotero

### Current Behaviour

Breaks on consecutive runs even after deleting local cache. Sometimes works for a few queries after 

### Expected Behaviour

1. Entries are simply found.

isaac-tes:
## Description

Follow-up hardening to #29478 (which fixed #29250).

#29478 removed the per-open memory blowup by loading only a slim copy of the database into `sql.js`. However, a **burst of concurrent searches** — `getCollections()` plus per-keystroke `searchResources()` fired while typing on first open, before a cache exists — still has each call independently open the database and rebuild the full dataset in parallel, and each writes the cache file over the same path.

Two problems remain from that:

1. **Peak memory multiplies** with the number of concurrent searches (each builds its own multi-thousand-row array and opens its own database).
2. **The cache write is not atomic** and several writes race on the same path, so an interrupted/interleaved write can leave a **truncated or 0-byte cache**. That cache then fails to `JSON.parse`, which forces a rebuild on every subsequent search — an endless loop that reproduces the original `JS heap out of memory` crash even though the slim fix is in place.

### Fix

- **Serialize `openDb()`** so concurrent callers never each hold a database copy in memory at the same time.
- **De-duplicate in-flight `getData()` / `updateCache()`** so a burst of concurrent searches shares a single build and a single cache write.
- **Write the cache atomically** (temp file + `rename`) so an interrupted or interleaved write can never leave a truncated cache.

### Verification

Reproduced against a real **471 MB** library by firing `getCollections()` + 6 `searchResources()` concurrently on a cold cache (simulating fast typing on first open):

- **Before:** each concurrent search builds and writes independently; the concurrent writes are what produced a 0-byte cache and the recurring crash.
- **After:** exactly **one** `getData` build and **one** atomic cache write; peak RSS ~**200 MB** instead of scaling with the number of concurrent searches; the cache is always valid and no temp files are left behind.

Behavior and results are otherwise unchanged; `ray build` + `ray lint` are clean.

## Checklist

- [x] I ran `npm run build` and it passed (`ray build` + `ray lint` clean)
- [x] I tested the change locally against a large (471 MB) library
- [x] I updated the `CHANGELOG.md`

### DIFF
--- a/extensions/zotero/CHANGELOG.md
+++ b/extensions/zotero/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Zotero Changelog
 
-## [Fixes] - {PR_MERGE_DATE}
+## [Fixes] - 2026-07-16
 
 - Prevent a memory spike and cache corruption when several searches run at once (e.g. fast typing right after opening the command): database opens are serialized, the data rebuild is de-duplicated so concurrent searches share one build, and the cache is written atomically (temp file + rename) so an interrupted or interleaved write can no longer leave a truncated cache that crashes on the next launch. Follow-up to #29478 / #29250
 - Fix a memory leak in the collections dropdown: the database opened by `getCollections` is now closed after use
--- a/extensions/zotero/CHANGELOG.md
+++ b/extensions/zotero/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Zotero Changelog
 
+## [Fixes] - {PR_MERGE_DATE}
+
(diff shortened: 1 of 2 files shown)


isaac-tes:
Addressed the pre-existing note from the review: `getCollections()` now frees its statement and calls `db.close()`, so the collections dropdown no longer leaks a sql.js/WASM database instance. Included in this PR since it's a small, on-theme memory fix.

0xdhrv (approved):
Looks good to me, approved ✅

PR #29506, opened 2026-07-17
Update duckduckgo-image-search extension

anwarulislam:
## Checklist


extensions/duckduckgo-image-search/src/search-image.tsx
jag-k, line 56: I'd like to move it out of function scope and make it real constant

extensions/duckduckgo-image-search/utils/helpers.ts
jag-k, line 201: This is not a quicklook action at all. Please use [Action.ToggleQuickLook](https://developers.raycast.com/api-reference/user-interface/actions#action.togglequicklook) with [Grid.Item#quickLook](https://developers.raycast.com/api-reference/user-interface/grid#:~:text=quickLook) for it

jag-k (changes requested):

jag-k:
Technically, the part of the reason, why I didn't add the quickLook -- we need to download the image, which is not that good for storage on the machine. And also generally slows down entire process of showing the content.

PR #29514, opened 2026-07-17
Add port-watcher extension

trgnn:
## Description

One list, two halves: what is listening on localhost right now, read live from your machine — and the dev servers you declared, launchable and killable from the same place.

## Screencast

port-watcher-1
port-watcher-2
port-watcher-3
port-watcher-4
port-watcher-5

## Checklist


0xdhrv (commented):

* [Dev Servers](https://www.raycast.com/tavlean/dev-servers) by @tavlean

The overlap is substantial. Both extensions provide a live view of local development servers and let users start, inspect, open, and terminate them from Raycast.

The existing Dev Servers extension already provides:

* Live detection of running local development servers
* Project-based grouping instead of only displaying ports
* Starting servers from Finder selections, recent projects, or chosen folders
* Automatic run-script, framework, and package-manager detection
* Killing and restarting individual servers or entire projects
* Opening servers in the browser, terminal, or editor
* Startup logs and failure reporting
* Network URLs, ports, uptime, Git branches, and framework metadata
* A menu-bar command with server status and quick actions

* Displaying every listening localhost process, including undeclared servers
* Explicit user-defined server profiles with custom commands
* Matching profiles by working directory when the port changes
* Creating a profile from an already-running process
* Detecting bare listeners, system daemons, and LAN-exposed services
* Explicit SIGTERM-to-SIGKILL escalation
* Persisting profiles and launch logs outside Raycast storage
* Suggestions based on `package.json`, `manage.py`, `Cargo.toml`, Makefiles, and static HTML projects

However, these appear to be natural process-discovery and profile-management enhancements to the existing Dev Servers extension rather than enough differentiation for a separate localhost server manager.

Could we consider contributing the broader listener detection, configurable profiles, profile capture, and additional project-type suggestions to the existing Dev Servers extension instead?

> This would help avoid duplication and keep local development-server discovery and process management consolidated in one extension.

tavlean:
Antonin reached out to me directly before this review resolved, which I appreciated.

Having looked at Port Watcher properly, I see it as complementary rather than duplicative, and I would not want its core features merged into Dev Servers. The two make opposite design bets: Dev Servers is project-first and zero-config (automatic framework detection, only real dev servers shown), while Port Watcher is port-first and explicit (every listener visible, launching through user-declared profiles). Showing all listeners and declared launch profiles are deliberately out of scope for Dev Servers, so those capabilities would simply not exist in the store if this PR is closed.

I support these existing side by side, ideally with description that make the distinction obvious to users.

PR #29516, opened 2026-07-17
fix(jira-search-self-hosted): ignore stale search responses

thezaff:
## Summary

Typing longer issue numbers can start overlapping Jira requests, allowing a slower response for a partial query to overwrite the latest result. Search now ignores responses from superseded queries, so an input such as `4107` keeps the result from its exact lookup. Regression coverage also confirms that four-digit issue numbers produce the expected Jira key.

## Validation

- `npm test`
- `npm run build`
- `npm run lint`

---

[Compound Engineering](https://github.com/EveryInc/compound-engineering-plugin)
Codex


0xdhrv (approved):
Looks good to me, approved ✅

PR #29524, opened 2026-07-17
Add mouse-cursor-toggle extension

Dhaiwat10:
## Description

A simple Raycast extension that lets people toggle the visibility of their mouse cursor.

Run **Toggle Mouse Cursor** once to hide the cursor and run it again to show it. The cursor stays hidden while using other apps.

## Checklist

- [x] I read the [publishing documentation](https://developers.raycast.com/basics/publish-an-extension)
- [x] I ran `npm run build` and tested the extension in Raycast
- [x] I checked that files in the `assets` folder are used by the extension
- [x] I checked that README assets are outside the `metadata` folder


0xdhrv (commented):
Could we update this extension to follow the same native-helper structure as the [Color Picker extension](https://github.com/raycast/extensions/tree/8d430b4ab715ddb9145a2c10c03c64a2d6d97767/extensions/color-picker)?

Color Picker keeps its Swift implementation under `swift/` and imports it directly from TypeScript using Raycast’s native module support:

```ts
const { pickColor } = await import("swift:../swift/color-picker");
```

Using the same approach here would avoid committing and directly executing a precompiled binary from `assets/`, remove the custom `swiftc`/`lipo` build scripts, and ensure the native code is built through the standard Raycast toolchain from the reviewed Swift source.

Could you please restructure the cursor helper accordingly?

---

Dhaiwat10:
@0xdhrv thanks for the review! I think the PR should be good now

0xdhrv (approved):
Looks good to me, approved ✅

PR #29526, opened 2026-07-17
feat: add google-task extension

Itsharshitgoat:
## Description

Adds the Google Task extension for Raycast.

### Commands
- **View Lists** — View task lists
- **Create Task** — Quickly create a new task

### AI Tools
- **Get Task Lists** — Fetch the user's Google Tasks lists
- **Get Tasks** — Fetch tasks from a specific Google Tasks list
- **Add Task** — Create a new task in a Google Tasks list


0xdhrv (commented):
Thanks for your contribution  

- https://www.raycast.com/elliotdes/google-tasks by @elliotdes

PR #29538, opened 2026-07-18
Add fxCodex extension

maximkrouk:
## Description

Raycast integration for [fxCodex](https://github.com/CaptureContext/fxcodex), a macOS CLI for managing isolated Codex workspaces.

The extension allows users to:

- Create, focus, rename, erase, and delete multiple Codex workspaces without signing out
- Customize workspace icons
- Optionally preserve the `Codex.app` name following its rebranding to `ChatGPT.app`
- Choose between the bundled CLI and an explicitly configured external installation

### Other Notes

- The bundled executable is verified against its published SHA-256 checksum.
- Automatic updates are available only when the user opts into using an external executable.
- The bundled and external executables share the same application-support directory and configuration. However, the extension always invokes the bundled executable with `FXCODEX_DISABLE_AUTO_UPDATE=1`, so the global auto-update preference cannot cause the bundled copy to modify itself.
- Metadata images are taken using raycast window capture tool, so technically they are correct and CI failed due to low contrast with the background (https://raycastcommunity.slack.com/archives/C02HEMAF2SJ/p1784480239092589)

## Screencast

> [!Note]
> 
> _Screencast is just a bit outdated, you can check screenshots in metadata folder for up-to-date UI_

https://github.com/user-attachments/assets/0700a2d6-8012-4708-acb9-4ce5267b230e

## Checklist


extensions/fxcodex/package.json
```diff
@@ -0,0 +1,46 @@
...
+		"@raycast/api": "^1.104.20",
+		"@raycast/utils": "^2.2.7"
+	},
+	"devDependencies": {
+		"@raycast/eslint-config": "^2.2.0",
+		"@types/node": "22.19.17",
+		"@types/react": "19.0.10",
+		"eslint": "^10.5.0",
+		"prettier": "^3.8.5",
+		"typescript": "^6.0.3"
```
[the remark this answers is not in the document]
maximkrouk: I used Raycast extension template and didn't modify this code, but I do use Raycast Beta, maybe that's why this was reported, however CI took those changes fine, so I'm resolving the comment

extensions/fxcodex/assets/bin/fxcodex
0xdhrv, line 1: **issue** (blocking): Keep the binary separate from the extension package.

Would it make sense to have users install the binary from a trusted source instead? The extension could provide setup instructions and clear error handling when it’s missing, while allowing the binary to be updated independently and keeping the extension more modular.

0xdhrv (changes requested):
Hey @maximkrouk 👋 

I have added a comment for you to address.

---

extensions/fxcodex/assets/bin/fxcodex
maximkrouk: I was thinking about it, but I think there is value in keeping a local copy, currently it's only ~15.5mb and provides a reliable locked version. Example of corrupted state with external executable:
- installed via brew with autoupdate to latest enabled
- new version with some breaking changes is released (I follow semver, so it's a major version update)
- brew updates external cli
- some features stop working (I tested some scenarios to partially recover the extension from such failures, so at least diagnotics should be available)
- users would have to downgrade

In current setup:
- Users are free to use an external executable (flexibility), but always can easily switch back to a bundled one which is never updated separately from the extension (stability)
- Users can even skip the part with global cli installation and just stick to extension-only use if they don't need CLI (ease of setup)

However I'm open to keep the binary separate an case it doesn't look sufficient for you, both approaches have pros and cons, I just wanted to explicitly highlight pros of the current one

maximkrouk:
Moved to in-review to get a reply :)

Feel free to move back to draft if binary-less extension is still preferred, just wanted to give reasoning behind the current implementation and figure out if changes are still needed.

0xdhrv:
> Moved to in-review to get a reply :)
> 
> Feel free to move back to draft if binary-less extension is still preferred, just wanted to give reasoning behind the current implementation and figure out if changes are still needed.

The overall implementation is fine, I would still prefer the binary-less implementation, if you can do that, that would be amazing.

Thanks for the contribution 🔥

0xdhrv (approved):
Looks good to me, approved ✅

PR #28005, opened 2026-05-17
[Todoist] Reduce Menu Bar Tasks sync size

mturac:
## Description
Fixes #27858.

The Menu Bar Tasks command used the shared full Todoist sync, which also pulls heavier resources such as notes, reminders, locations, filters, and sections. The menu bar only needs user, projects, items, labels, and collaborators for rendering and menu actions, so this lets the sync hook accept a resource list and uses a smaller sync payload for the menu bar command. Other Todoist commands keep the existing full sync behavior.

## Screencast
N/A; background menu bar memory usage fix.

## Testing
- npm run lint --prefix extensions/todoist
- npm run build --prefix extensions/todoist
- git diff --check

PR #29552, opened 2026-07-18
[Todoist] Fix menu bar out-of-memory crash

kud:
## Description

Fixes the Todoist Menu Bar command crashing with:

> Worker terminated due to reaching memory limit: JS heap out of memory (Code: 6, Worker Error)

**Root cause:** every Todoist command (My Tasks, Menu Bar, etc.) shared a single `useCachedState` cache key `"data"` holding the _entire_ sync state — every task, plus every comment (`notes`) and every location. The Menu Bar command runs in a background worker with a tight JS heap. On each refresh it loaded that whole shared blob off disk, merged new sync data into it, and re-serialised the whole thing back to disk — on larger accounts (lots of comments) this exceeded the worker's heap limit and crashed.

A prior PR, #28005 ("[Todoist] Reduce Menu Bar Tasks sync size"), narrowed what the menu bar _fetched_ from the Todoist Sync API, but left the shared cache key in place — so the full blob (including comments/locations from other commands) was still pulled back into memory on every merge. This PR completes that intent by scoping the _cache_ too, not just the fetch.

**The fix (3 source files + CHANGELOG):**

- `src/hooks/useCachedData.ts`: now accepts an optional `cacheKey` parameter, defaulting to `"data"` — every existing caller that doesn't pass a key is unaffected.
- `src/hooks/useSyncData.ts`: threads that optional `cacheKey` through to `useCachedData`.
- `src/menu-bar.tsx`: now passes its own dedicated cache key, `"menu-bar-data"`, which holds only the 5 resource types the menu bar actually needs (`user`, `projects`, `items`, `labels`, `collaborators`) — no `notes` (comments) or locations. The background worker never touches the heavy shared blob anymore. The command also always performs the sync now (previously it conditionally skipped syncing when launched from within another command, relying on that other command having freshened the shared `"data"` cache — that assumption no longer holds now the cache is isolated, so it always syncs its own small slice).

No UI or behavioural changes for other commands — the default cache key is unchanged, so `My Tasks` and the rest keep sharing `"data"` exactly as before.

## Screencast

No visual changes — this is an internal caching fix with no UI impact.

## Checklist

### DIFF
--- a/extensions/todoist/CHANGELOG.md
+++ b/extensions/todoist/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Todoist Changelog
 
+## [Fix menu bar out-of-memory crash] - {PR_MERGE_DATE}
+
+- Fixed the Menu Bar command crashing with "Worker terminated due to reaching memory limit" on larger accounts. Every command previously shared one cache key holding the entire sync state (all tasks, comments, and locations); the menu bar's background worker now uses its own cache key holding only the small slice of data it needs (user, projects, items, labels, collaborators), completing the sync scoping started in #28005.
+
 ## [Omit paid task fields unless requested] - 2026-06-28
 
 - **Create Task tool**: The AI tool now omits empty deadline and duration payloads, and its field descriptions clarify that those paid Todoist fields should only be sent when explicitly requested. Standard task creation with ordinary due dates remains compatible with Free plan accounts.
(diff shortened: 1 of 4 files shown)


0xdhrv (approved):
Looks good to me, approved ✅

PR #29572, opened 2026-07-19
Add sidecar-display extension

chiptoma:
## Description

**Sidecar Display** connects an iPad over Sidecar and forces it to **extend**
instead of mirror — from a hotkey or the menu bar — without ever writing or
relocating the Mac's main display.

Two interchangeable engines sit behind one interface, auto-selected:

- **BetterDisplay** — drives the `betterdisplaycli` binary (needs the
  BetterDisplay app). Proven and low-maintenance.
- **Native** — no runtime dependency. A small Swift helper connects/disconnects
  via the private `SidecarCore` framework loaded at runtime with `dlopen`
  (the binary links nothing private), and controls mirror state through
  **public** CoreGraphics.

### Features
- **Connect / Disconnect / Fix Mirroring** commands, plus a **Sidecar Status**
  menu bar with a one-click Auto-Reconnect toggle.
- **Auto-Reconnect** background command that restores a link dropped on its own
  (e.g. waking from sleep), with configurable backoff and a heartbeat that never
  abandons a wanted link.
- **Fix Mirroring** clears macOS Sidecar's own mirror mode — which is invisible
  to every display API — by reconnecting the main BetterDisplay virtual screen.

### Safety
The extension never writes or power-cycles the main display; mirroring always
keeps the current main as master, and both mode writes are refused when the iPad
is itself main. These invariants are enforced on every path and proven by
headless unit tests (no hardware needed).

### Note for reviewers (private framework)
The native engine uses the private `SidecarCore` framework via runtime
`dlopen`/selectors and public CoreGraphics — **no opaque binary is committed**;
the Swift ships as auditable source, compiled by `extensions-swift-tools`.
If the private-API use is a concern, the BetterDisplay engine is a full-featured,
dependency-free-of-private-APIs fallback and can ship on its own.

Tested on macOS 26 with an iPad over Sidecar, both engines.

## Checklist


0xdhrv (commented):

We already have a Sidecar extension in the Store:

* [Sidecar](https://www.raycast.com/philip_chen/sidecar) by @lsps9150414

The existing extension already lets users connect and disconnect an iPad as a second display from Raycast.

This PR adds more Sidecar functionality, including:

* Choosing between extend and mirror modes
* Automatically reconnecting after the connection drops
* Showing the connection status in the menu bar
* Detecting whether the iPad is nearby
* Fixing mirroring issues
* Supporting multiple Sidecar devices

Could these features be added to the existing Sidecar extension instead of introducing another extension for the same workflow?

> This would keep Sidecar-related functionality in one extension while improving the existing extension with connection status, display-mode controls, and automatic reconnection.

chiptoma:
Thanks for taking a look.

The overlap is narrower than the feature list suggests — these aren't two implementations of the same thing with different polish. They're different mechanisms.

The existing extension drives the System Settings UI through AppleScript. It opens the Displays pane and clicks the device popup by accessibility-tree index:

```applescript
set popUpButton to menu button 1 of group 1 of group 3 of splitter group 1 of group 1 of window 1
on error
  set popUpButton to pop up button 1 of group 1 of group 2 of splitter group 1 of group 1 of window 1
```

Two hardcoded paths because the hierarchy moved in Tahoe. It requires Accessibility permission, opens a window on every use, and breaks whenever Apple reshuffles that pane.

This extension talks to `SidecarCore` directly and controls mirroring through public CoreGraphics. No System Settings, no AppleScript, no Accessibility permission.

That difference is why the features you listed can't simply be added to the existing extension. UI scripting can only *issue clicks* — it has no way to ask a question. Auto-reconnect needs to read whether the link is up. Presence detection needs to read whether the iPad is within reach. The menu-bar status needs both. None of those are reachable by clicking buttons in a settings pane, so they aren't features to port; they're consequences of the engine.

There's also a safety dimension. macOS decides which display becomes main when Sidecar attaches, and it can pick the iPad. This extension refuses to write or relocate the main display on every path, and that invariant is enforced by tests. An earlier iteration of mine got it wrong: it scrambled every window and caused a logout. That's documented in `docs/ARCHITECTURE.md`, along with the reasoning behind each guard.

So I don't think folding these together produces one better extension — it would mean discarding the existing implementation and replacing it wholesale with this one, which is a different proposition from what you're asking.

0xdhrv (approved):
Looks good to me, approved ✅

PR #29573, opened 2026-07-19
Remo: infinite scroll in folders and trash, clearer search

alfredmouelle:
## Description

Updates the **Remo** extension:

- **Folders and Trash** now paginate and load more notes as you scroll, instead of stopping at the first batch.
- **Search** shows your 20 most recent notes by default, then searches note titles and content as you type (with a clearer placeholder and section hint).
- **Removed the Inbox and Quick Capture shortcuts** from the Folders view: the notes API now selects by view, and these two views are no longer served. This is an intentional removal, documented in the changelog.
- Internal: note lists now go through a single paginated endpoint.

## Checklist

- [x] I checked that files in the `assets` folder use widely supported formats

### DIFF
--- a/extensions/remo-notes/CHANGELOG.md
+++ b/extensions/remo-notes/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Folders and Trash now load more notes as you scroll, instead of stopping at the first batch
 - Search shows your 20 most recent notes by default, then searches note titles and content as you type
+- Removed the Inbox and Quick Capture shortcuts from the Folders view
 
 ## [Pinned notes & improved rendering] - 2026-06-30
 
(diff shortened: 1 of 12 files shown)


alfredmouelle:
Thanks for the review!

**Inbox / Quick Capture removal** — intentional. The notes API now selects by view, and those two views are no longer served. I've documented it in the changelog and updated the PR description.

**P2, trash ordering** — no behavior change. The server serves `view=trash` via the `by_user_deleted` index (`[userId, deletedAt]`) with `.order('desc')`, so notes come back in `deletedAt`-descending order. The previous client-side sort (`deletedAt ?? updatedAt`) was equivalent, since trash notes always have `deletedAt` set and the fallback never triggered. Most-recently-deleted stays first.

0xdhrv (approved):
Looks good to me, approved ✅

PR #29587, opened 2026-07-20
Teleport: fix Login command (OTP MFA mode + reliable failure detection)

lamberttraccard:
## Description

This fixes the Teleport `Login` command, which failed silently on clusters that require a second factor (MFA).

### Problem
`assets/scripts/login.sh` drives `tsh login` through `expect` without setting an MFA mode. On clusters whose preferred local MFA is WebAuthn, a modern `tsh` (auto mode) attempts WebAuthn first, which cannot be completed from Raycast's headless subprocess, so `tsh` stalls on the MFA step and is killed by the 15s `spawnSync` timeout. Because success was inferred only from the absence of the substring "ERROR" in stdout, the command showed a green "Logged in !" while nothing actually happened.

### Changes
- `assets/scripts/login.sh`: pass `--mfa-mode=otp` (the command already takes an OTP argument), wait for the real password and OTP prompts instead of sending them blindly, propagate `tsh`'s exit code, and treat an already-valid session (tsh exits before prompting) as success.
- `src/utils.ts`: capture both stdout and stderr, throw on a non-zero exit code or spawn error, and raise the timeout to 60s.
- `src/login.tsx`: treat any thrown error as a failure and surface the real message in the toast.

### Testing
Validated end to end against a live Teleport v18 cluster: password (from preferences) + a fresh OTP, session created successfully. `npm run build` and `npm run lint` pass.


0xdhrv (approved):
Looks good to me, approved ✅

PR #29609, opened 2026-07-21
Add scoop-app-launcher extension

noidwasavailable:
## Description

Scoop apps directory is convoluted and if added to Raycast root, it pollutes search results with all the different installed versions like below.

image

There is also an issue of launching apps like this that doesn't always respect the current User Profile and may result in not using the correct user data folders.

Using this extension makes sure that only the `current` versions of apps managed by scoop are launched, and makes use of frecency sorting to make it easier to look for apps just as in Raycast root.

It should (in theory) be able to load the correct user profiles by passing current User Profile as arguments to PowerShell, which is used to launch apps.

## Known Limitations

1. Only the default `~/scoop` is recognized automatically. If you installed scoop elsewhere, you have to manually specify it in the extension preferences.

## Screencast

image
Automatically detected scoop folder and their executables.

## Checklist


0xdhrv (commented):

* [Scoop](https://www.raycast.com/zsombor_biro/scoop) by @Giton22

Both extensions center around discovering and working with applications/packages already installed through Scoop on Windows.

* Listing installed Scoop packages
* Searching and installing packages
* Updating, uninstalling, and cleaning up packages
* Managing Scoop buckets

* Launching the current executable for installed Scoop apps
* Frecency-based sorting for frequently launched apps
* Running apps as administrator and supporting a custom Scoop root

Could we consider contributing the app-launching, frecency sorting, and administrator-launch capabilities to the existing extension instead?

> This would help avoid duplication and keep Scoop management consolidated in one extension.

PR #29634, opened 2026-07-21
Add Gold Price extension

Noice-Anas:
## Description

Adds **Gold Price** — a `view` command that shows the daily gold price **per gram** for 24K / 22K / 21K / 18K, plus 1-month, 3-month, 6-month, and 1-year averages with a karat selector.

- Live spot price and daily change vs. the previous close.
- Period averages computed from a rolling ~1-year daily history.
- Selectable display currency (SAR default; AED, KWD, QAR, BHD, OMR, USD, EUR, GBP).
- Quota-friendly caching to stay within the metals.dev free tier (100 req/mo).

Data source: [metals.dev](https://metals.dev) (user provides their own free API key).

## Checklist

- [x] I ran `npm run build` and tested the extension locally
- [x] I ran `npm run lint` and it passed
- [x] I added screenshots (2000×1250) in `metadata/`
- [x] The `author` field matches my Raycast/GitHub handle


0xdhrv (commented):
Hey @Noice-Anas 👋

Thanks for your contribution!

This is great addition to the Store. I would like to suggest if we can tweak this to support for more metals, and have the same API being used for all of them. Also, rename the extension to something "Metals Price" or "Metals Prices" or something similar.

---

Noice-Anas:
Hey @0xdhrv 👋 — thanks for the review, both suggestions are in.

**Renamed to "Metals Prices"** (`gold-price` → `metals-prices`, directory renamed too), and it now covers **gold, silver, platinum and palladium**.

**On using the same API for all of them** — it turned out to be free in the strongest sense. Both metals.dev endpoints already return every metal in a single response: `/latest` carries the whole `metals` map, and each day of `/timeseries` does too. So the three new metals added **zero additional requests**, and switching metals in the UI fires nothing at all — it re-renders from data that was already loaded. The quota profile is unchanged from the version you reviewed: first open = 2 requests, steady state = 0.

A few details worth flagging:

- **Purity per metal.** Each metal shows the grades it's actually traded in — gold keeps the karat convention (24K/22K/21K/18K), the white metals use parts-per-thousand fineness (999/958/925/900, 950, 500). Metal and purity share the one `searchBarAccessory` Raycast allows, sectioned by metal, and the last selection is remembered between launches.
- **History is shared across metals.** Days are stored per-metal, and a backfill marks every metal the response carried — so loading gold's 1-year window also unlocks silver, platinum and palladium rather than charging four times for the same requests.
- **Display precision now scales with the value.** Two fixed decimals suited gold but silently broke cheaper metals in high-value currencies — 925 silver in KWD is ~0.38/g with a ~0.001 daily move, and both rendered as `0.00`. Precision is now chosen from the selected metal's price magnitude and applied consistently across the row.

**Deliberately out of scope:** `/latest` also returns industrial metals (copper, aluminum, lead, nickel, zinc), but `/timeseries` does **not** — they'd have a spot price with no daily change and no period averages, and no meaningful purity grades, so every section of this UI would only half-work. The `lbma_*` / `mcx_*` / `ibja_*` keys are these same four metals quoted by a different authority. Very happy to add either — industrial metals as a clearly spot-only section — if you'd like them in.

One honest note before you look: **the metadata screenshot is still the old gold-only capture.** I'm retaking it against the new UI and will push it shortly — flagging it so it doesn't read as an oversight. Everything else is ready, and `npm run build` + `npm run lint` both pass.

Noice-Anas:
Screenshots are in — the gold-only capture is replaced with three shots of the renamed extension:

1. **Gold** per gram by karat (24K/22K/21K/18K) with the 1-month average loaded.
2. **The metal + purity selector**, sectioned by metal — Gold 24K/22K/21K/18K, Silver 999…, and so on.
3. **An on-demand history load**, with the toast showing `2 requests used — covers every metal`, which is the quota behaviour described above.

All three are 2000×1250. That was the last outstanding item from my previous comment — this is ready for a look whenever you have time. 🙏

Noice-Anas:
Hey @0xdhrv 👋,
just a friendly nudge.
is there anything you would suggest for me to change?
or you can review and approve this now. 

I am excited to have my first raycast extension on the store 🚀.

thank you.

0xdhrv (approved):
Looks good to me, approved ✅

PR #29646, opened 2026-07-22
[Windows Terminal] Add non-quake admin action when quake preference is on

artistro08:
## Description

Follow-up to #28197. When the `Open profiles in quake window` preference is on, all admin launches route into the quake window — but elevated quake windows don't respond to Windows Terminal's global quake shortcut (Win+`), so once the drop-down loses focus there's no way to summon it back.

This adds an escape hatch: while the preference is on, an additional **Open as Administrator (Non-Quake)** action (⌃↵) is exposed alongside the quake admin action (⌃⇧↵), letting users open an elevated session in a normal window.

## Screencast / Screenshot

_The action only appears when the preference is enabled._

## Checklist

- [x] I read the [contribution guidelines](https://github.com/raycast/extensions/blob/main/CONTRIBUTING.md).
- [x] I read the [documentation about publishing](https://developers.raycast.com/basics/prepare-an-extension-for-store).
- [x] I ran `npm run build` and tested that this distribution build works as expected.
- [x] I ran `npm run lint` and it passed.
- [x] I made this pull request with the intended target branch.
- [x] My extension is compatible with all supported OSes (or I checked "compatible" with only the supported ones).
- [x] I updated the CHANGELOG file with the `{PR_MERGE_DATE}` placeholder.

### DIFF
--- a/extensions/windows-terminal/CHANGELOG.md
+++ b/extensions/windows-terminal/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Windows Terminal Changelog
 
-## [Non-Quake Admin Action] - {PR_MERGE_DATE}
+## [Non-Quake Admin Action] - 2026-07-22
 
 - Added an "Open as Administrator (Non-Quake)" action (⌃↵) that appears while the `Open profiles in quake window` preference is on, so users can still open an elevated session in a normal window — elevated quake windows don't respond to the global quake shortcut (Win+`).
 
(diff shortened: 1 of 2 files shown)


0xdhrv (commented):
Hi @artistro08 👋

Thanks for the contribution 🔥

---

artistro08:
Thank you! Should be good now

0xdhrv (approved):
Looks good to me, approved ✅

PR #29653, opened 2026-07-22
[Windows Terminal] Always open administrator in a normal window

artistro08:
## Description

Follow-up to #29646. Instead of exposing two separate admin actions when the `Open profiles in quake window` preference is on, "Open as Administrator" now always launches into a regular window — regardless of the preference. Elevated quake windows don't respond to Windows Terminal's global quake shortcut (Win+`), so routing admin through quake left users unable to summon the drop-down back once it lost focus. The separate "Open as Administrator (Non-Quake)" action from the previous release is no longer needed and has been removed.

Also skipped loading the PowerShell profile when spawning the elevated launcher (`powershell -NoProfile`), which noticeably cuts the delay between triggering the action and the UAC prompt.

## Checklist

- [x] I read the [contribution guidelines](https://github.com/raycast/extensions/blob/main/CONTRIBUTING.md).
- [x] I read the [documentation about publishing](https://developers.raycast.com/basics/prepare-an-extension-for-store).
- [x] I ran `npm run build` and tested that this distribution build works as expected.
- [x] I ran `npm run lint` and it passed.
- [x] I made this pull request with the intended target branch.
- [x] My extension is compatible with all supported OSes (or I checked "compatible" with only the supported ones).
- [x] I updated the CHANGELOG file with the `{PR_MERGE_DATE}` placeholder.

### DIFF
--- a/extensions/windows-terminal/CHANGELOG.md
+++ b/extensions/windows-terminal/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Windows Terminal Changelog
 
+## [Always Open Administrator in a Normal Window] - {PR_MERGE_DATE}
+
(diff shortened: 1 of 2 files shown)


0xdhrv (approved):
Looks good to me, approved ✅

PR #29657, opened 2026-07-22
Add claude-usage extension

dzg:
## Description

Adds `claude-usage`, a menu bar extension for tracking [Claude Code](https://claude.com/claude-code) usage limits.

Claude Code enforces a rolling 5-hour session limit alongside a weekly one, and it's easy to run into either mid-task without warning. This keeps the numbers visible: percentage used and time until reset in the menu bar, with a breakdown of session, weekly, and usage-credit consumption when you open it. Each row shows a progress icon that fills clockwise and turns red past 95%, and clicking a row copies its numbers.

## How it works

The extension reads the OAuth token Claude Code already stores in the macOS Keychain (`Claude Code-credentials`) and calls Anthropic's usage endpoint with it. It never prompts for credentials, stores none of its own, and sends nothing anywhere other than `api.anthropic.com`. If Claude Code isn't installed or signed in, the menu shows a clear error rather than going blank.

## Preferences

**Refresh Frequency** — 1 to 30 minutes, default 5.

The endpoint is rate limited and that limit is shared with Claude Code itself, so responses are cached to disk. The command wakes on Raycast's fixed one-minute interval but only makes a request once the cache has aged past the chosen frequency; otherwise it replays the cached response. `Refresh Now` (⌘R) always bypasses the cache.

## Notes for reviewers

- **macOS only** — it shells out to `/usr/bin/security` for the Keychain lookup.
- **The usage endpoint is not part of Anthropic's public API.** `/api/oauth/usage` is the same endpoint Claude Code's own `/usage` command uses, so it could change without notice. The caching above doubles as the mitigation: on any non-200 response the extension replays the last known figures instead of erroring, so a change degrades it to stale-but-visible data rather than breaking it outright.

## Checklist


0xdhrv (commented):

We reviewed the submission against existing extensions in the Store and found substantial overlap with **Claude Code Usage (ccusage)** and **Agent Usage**.

The overlapping functionality includes:

* Monitoring Claude Code usage from the menu bar
* Displaying the rolling five-hour and seven-day limits
* Showing utilization percentages and reset times
* Reading Claude Code OAuth credentials
* Calling Anthropic’s `/api/oauth/usage` endpoint
* Supporting configurable refresh and caching behavior

Your submission also introduces some useful improvements:

* Model-specific weekly usage rows
* Usage-credit details
* Custom circular progress indicators
* Last-known usage fallback when the endpoint fails

However, these additions extend the same underlying workflow and data source rather than introducing a meaningfully distinct use case.

Could we consider adding these improvements to the existing extension instead?

* [Claude Code Usage (ccusage)](https://www.raycast.com/nyatinte/ccusage) by @nyatinte
* [Agent Usage](https://www.raycast.com/thuggyduck/agent-usage) by @natsustan

Contributing them there would:

* Keep Claude Code usage functionality consolidated
* Avoid multiple extensions offering nearly identical workflows
* Improve discoverability for existing users
* Allow the current extension to benefit from your additional usage details and interface improvements

If there is a unique workflow or use case that is not covered by the existing extension, please share more details. We would be happy to reconsider whether it is sufficiently distinct or explore how it could be integrated into `ccusage`.

> This helps avoid duplication and keeps related functionality consolidated in one place, as outlined in our [extension guidelines ↗](https://manual.raycast.com/extensions)

dzg:
Neither of those does what I want, which is this:
Screenshot Raycast Beta 2026-07-23_11-15-53@2x

ccusage shows:
Screenshot CleanShot 2026-07-23_11-16-18@2x

and agent usage shows:
Screenshot CleanShot 2026-07-23_11-16-39@2x

dzg:
P.S. That's without even going into what the dropdowns show – of which 99%
in those other extensions is totally irrelevant to me.

dzg:
Just adding these screenshots which didn't come through via email response:

Neither of those does what I want to se, which is this:
Screenshot Raycast Beta 2026-07-23_11-15-53@2x

ccusage shows:
Screenshot CleanShot 2026-07-23_11-16-18@2x

agent usage shows:
Screenshot CleanShot 2026-07-23_11-16-39@2x

0xdhrv:
> Just adding these screenshots which didn't come through via email response:
> 
> Neither of those does what I want to se, which is this: Screenshot Raycast Beta 2026-07-23_11-15-53@2x
> 
> ccusage shows: Screenshot CleanShot 2026-07-23_11-16-18@2x
> 
> agent usage shows: Screenshot CleanShot 2026-07-23_11-16-39@2x

Thanks for clarifying, but this is why I mentioned that you can enhance the existing extension to add support for these features. They’re useful improvements, so it’s better to include them in the existing extension. This will also avoid duplication, which can confuse users when they try to find the correct extension.

dzg:
That's ok, I'll just keep this for my own use.
Too much work to even figure out how to minimize those other extensions to get what I want.

dzg:
Hey @0xdhrv  – I went ahead and did as you suggested, please see
https://github.com/raycast/extensions/pull/29782

0xdhrv:
Closing this. Thanks for your help and contribution!

PR #29664, opened 2026-07-23
Update r2-uploader extension

mazaoshe:
## Description

Adds a new "Browse R2 Files" command plus several improvements to the upload command:

- **Browse R2 Files** (new view command): navigate the bucket's folders like a file manager, preview files (inline image preview via a short-lived signed URL, so it works even on private buckets), copy a file's link, and delete files. Deleting a folder recursively counts every file under it (including subfolders) and shows the exact count before permanently batch-deleting them.
- **Upload Path Prefix** preference: store uploads under a folder instead of the bucket root, using the same date/name placeholders as the filename format (closes a longstanding feature request, mazaoshe/Raycast-UploadImageR2#1).
- **Folder argument**: optionally set the upload folder per invocation; it's "sticky" across uploads (persisted until changed or reset with `/` or `root`) and reflected live in the command's subtitle, so the active folder is always visible before uploading.
- **Link Format** dropdown (Plain URL / Markdown / HTML) replacing the old Generate Markdown checkbox.
- Bug fix: `{year}`/`{day}` filename placeholders used invalid dayjs format tokens and silently produced incorrect output.
- Friendlier error when Finder isn't the frontmost app (previously surfaced a raw stack trace).

## Screencast

See `metadata/screenshot-3.png` and `metadata/screenshot-4.png` for the new Browse R2 Files command (list/detail preview and the action panel with Copy URL / Copy Markdown / Delete File).

## Checklist


0xdhrv (approved):
Looks good to me, approved ✅

PR #29665, opened 2026-07-23
Add Polish Text extension

lekhanh1234:
## Description

Adds **Polish Text**, a Raycast extension that rewrites highlighted (or manually typed) text into clearer, friendlier phrasing using an AI provider.

- Bring-your-own-key (BYOK): the user supplies their own OpenAI, Anthropic, or OpenRouter API key in extension preferences. No backend, no license system, no data leaves the user's machine except directly to their chosen provider.
- Flow: select text (or type it) → review/edit in a form pre-filled from the selection → submit → result screen shows both original and polished text → Enter pastes the polished text over the original selection, or copy it instead.

## Checklist

- [x] `author` in `package.json` matches my Raycast username
- [x] `license` is `MIT`
- [x] Using latest `@raycast/api` version
- [x] `npm run build` and `npm run lint` pass locally
- [x] `package-lock.json` included
- [x] README included (extension requires an API key)
- [x] 512x512 PNG icon included


lekhanh1234:
Thanks for the review — pushed a fix for each code issue:

1. **Provider request never times out** — added a 30s timeout via `AbortController`; a stalled response now surfaces a distinct "timed out" error instead of hanging on "Polishing..." forever.
2. **Temporary errors strand the form** — the error screen now shows an "Edit Text" action for retryable failures (rate limit, network, timeout, too-long input), carrying the previously entered text back into the form so the user can fix and resubmit without closing/reopening the command.
3. **Text interpreted as Markdown** — original/polished text is now wrapped in a Markdown code fence sized past the longest backtick run present in the text, so headings/lists/quotes in the user's own text render literally.
4. **Preference shape duplicated** — dropped the hand-rolled `Preferences` interface in favor of Raycast's generated global `Preferences` type (also updated `tsconfig.json` to include `raycast-env.d.ts` so this type-checks outside the `ray` CLI too).
5. **Changelog** — added, with the `{PR_MERGE_DATE}` placeholder.
6. **Screenshots** — per the [Store prep guide](https://developers.raycast.com/basics/prepare-an-extension-for-store#screenshots), these are recommended (not required) — happy to add them if that's actually a blocker for this extension, just let me know.

All changes verified locally: `tsc --noEmit`, `npm run test` (15/15), `npm run lint`, `npm run build` all pass.

lekhanh1234:
Good catch on **Timeout Ends Before Body Read** — fixed. The abort timeout now stays active through \`response.json()\`, not just the initial \`fetch()\` call, so a provider that returns 200 headers but stalls while streaming the body is now caught too. Added a test for that case; 16/16 passing, \`tsc\`/lint/build all clean.

0xdhrv (commented):

* [Text Enhance](https://www.raycast.com/max13021302/text-enhance) by @max1302

Both extensions take selected or manually entered text, use AI to improve/rewrite it, and let the user copy or paste the generated result back into their workflow.

* Using selected text or clipboard content as input
* Multiple enhancement modes such as clarity, grammar, polishing, shortening, and expanding
* A wide range of tones and writing purposes
* Model selection and custom instructions
* Follow-up corrections and iteration
* Saved presets and enhancement history

* Bring-your-own-key support
* OpenAI, Anthropic, and OpenRouter provider selection
* Direct provider usage without relying on Raycast AI

However, these appear to be natural enhancements to the existing extension rather than enough differentiation for a separate extension with the same primary AI text-rewriting workflow.

Could we consider contributing BYOK and external provider support to the existing Text Enhance extension instead?

> This would help avoid duplication and keep AI-assisted text enhancement consolidated in one extension.

lekhanh1234:
Thank you @0xdhrv for your review!

I didn’t know that plugin existed. I’ll close this PR and create a new one when I have time to add the BYOK feature to it.

PR #29670, opened 2026-07-23
Add JWT Debugger extension

Im-Tae:
## Description
New extension: **JWT Debugger** — a token.dev-style workbench to decode, encode, edit, re-sign, and verify JSON Web Tokens, entirely on-device.

- **Bidirectional**: edit the header / payload / signing key / algorithm and the JWT re-signs live; paste a JWT and it decodes.
- **Algorithms**: HMAC (`HS256/384/512`), RSA (`RS`/`PS256/384/512`), ECDSA (`ES256/384/512`), and unsecured `none`.
- **Keys**: Base64-encoded secret toggle for HMAC; PEM private/public keys for asymmetric algorithms.
- **Verification** status shown in the title bar (🟢 valid / 🔴 invalid / 🟡 unsigned or unchecked).
- Loads a JWT from the clipboard on open. Nothing leaves your machine.

The icon is the official JWT (jwt.io) brand mark.

## Checklist
- [x] `npm run build` passes without errors
- [x] `npm run lint` passes without errors
- [x] Tested the distribution build in Raycast
- [x] Screenshots added to `metadata/`


0xdhrv (commented):

We reviewed this submission against the existing **JWT Decoder** extension and found substantial overlap in the primary JWT inspection workflow.

Both extensions allow users to:

* Load a JWT from the clipboard
* Decode its header and payload
* Inspect the decoded JSON locally
* Copy header and payload data
* Process tokens using the `jose` library
* Use the extension on macOS and Windows

The existing extension is:

* [JWT Decoder](https://www.raycast.com/gdsmith/jwt-decoder) by @gdsmith

Your submission does introduce several useful capabilities that are not currently available there:

* Editing JWT headers and payloads
* Encoding and re-signing tokens
* Verifying token signatures
* Supporting HMAC, RSA, RSA-PSS, and ECDSA algorithms
* Using Base64 secrets and PEM public/private keys
* Showing live valid, invalid, unsigned, or unchecked verification states

These are meaningful additions, so the submission is not a direct duplicate. However, they extend the same underlying workflow already provided by JWT Decoder.

Could we consider adding this functionality to the existing extension instead for example, through a new **JWT Debugger** or **Sign and Verify JWT** command?

This would:

* Keep JWT decoding and debugging functionality consolidated
* Avoid two extensions with nearly identical clipboard-based decoding entry points
* Improve discoverability for existing JWT Decoder users
* Allow the existing extension to grow from a decoder into a complete JWT workbench

If there is a workflow that cannot reasonably fit into JWT Decoder, please share more details. Otherwise, contributing these features to the existing extension would be the preferred approach.

> This helps avoid duplication and keeps related functionality consolidated in one place, as outlined in our [extension guidelines ↗](https://manual.raycast.com/extensions).

Im-Tae:
@0xdhrv 
Thank you for taking the time to review this so thoroughly — I really appreciate it! 🙏

You're right that the clipboard decode entry point overlaps with JWT Decoder. My intent was a bit different from a pure decoder: **JWT Debugger** is a single bidirectional workbench where decoding and **encoding / re-signing** happen in the same view — you edit the header, payload, key, or algorithm and the token is re-signed live (similar to token.dev), with live valid/invalid/unsigned verification states. Because of that, its main screen and interaction model are quite different from JWT Decoder's read-only decode view.

I did consider adding these features to the existing JWT Decoder, but since it's another maintainer's (@gdsmith) already-published extension with a focused decode UI, I was hesitant to change its established design or risk regressing a finished extension I don't own.

That said, I'm happy to go whichever way you prefer:

1. Would you like me to add this as a **new command** (e.g. "JWT Debugger" / "Sign and Verify JWT") inside the existing JWT Decoder extension instead? If so, should I open that PR against @gdsmith's extension?
2. If keeping it separate is fine, I can **remove the clipboard auto-decode entry point** here so the two extensions no longer overlap on that specific workflow.

Just let me know which direction you'd prefer and I'll adjust. Thanks again!

gdsmith:
FWIW I'd be happy for you to add this to the existing one

Im-Tae:
@gdsmith 

Thanks! In that case, I'll move this over to the existing JWT Decoder extension and open a PR there instead.

The plan is to add it as a separate command (e.g. JWT Debugger / Sign and Verify JWT) so the existing decode workflow remains unchanged while adding the editing, signing, and verification features.

0xdhrv:
Thanks both @Im-Tae @gdsmith ❤️

0xdhrv:
Closing this in favor of https://github.com/raycast/extensions/pull/29705

PR #29672, opened 2026-07-23
Add private-internet-access extension

vxirau:
## Description

A new extension for **Private Internet Access**, built on `piactl`, PIA's
official command-line interface. It brings the full region catalog into
Raycast with country flags, live connection status, port-forwarding support,
favorites, and hotkey-friendly commands.

**Commands**

| Command | Mode | Action |
|---|---|---|
| Open Detailed | view | Browse regions, check status, connect, disconnect, change settings |
| Toggle Connection | no-view | Connect if disconnected, disconnect if connected |
| Connect Most Recent | no-view | Reconnect to the last-used region |

### Why a separate extension rather than a contribution to `pia-controls`

The store guidelines allow a new extension when changes "would be significant
and potentially disrupt existing workflows" and the extension "works standalone
and brings something new."

This is a ground-up rewrite with a different command set and architecture.
Folding it into `pia-controls` would rename that extension's commands and break
hotkeys its users have already bound, which seemed worse for its users than
shipping alongside it. What's new here:

- All 166 regions with country flags, searchable by country, city, or region id
- Port-forwarding support tagged per region, plus the active forwarded port
  shown and copyable
- Geo-located regions flagged, so it's clear when an IP is registered in a
  country but the server is hosted elsewhere
- Favorites and recently used regions
- Settings from the action panel: port forwarding, LAN access, and protocol
- A correctness fix worth calling out: `piactl get pubip` returns the
  ISP-assigned address and does **not** change while the tunnel is up. Showing
  it as the connected IP would display the user's real home address, so this
  extension shows `vpnip` when connected and only shows the public IP when
  disconnected, labelled as unprotected.

### Notes

- **No credentials.** Sign-in stays entirely in the PIA app; the extension
  never reads or stores them.
- **No third-party requests.** Flags ship with the extension rather than
  loading from an image CDN, so browsing regions doesn't reveal to anyone which
  countries a user is looking at. The only network call is to PIA's own public
  server list, cached for 24h.
- **Flag licensing.** Rendered from [flag-icons](https://github.com/lipis/flag-icons)
  (MIT), attributed in the README with the license text in `NOTICE.md`.
- **No implicit state changes.** Nothing connects, disconnects, or alters a PIA
  setting unless the user triggers it. When PIA's daemon is inactive the
  extension explains the options rather than enabling background mode itself.
- **Command execution.** All CLI calls use `execFile` with positional
  arguments, never a shell. Region ids are validated against a slug pattern
  before being passed through, since favorites and recents come back from
  local storage.

## Screencast

**Connected — status row with region flag, protocol, and VPN IP**

Connected status

**Region list — favorites, port-forwarding and geo tags**

Region list

**Settings from the action panel**

Settings panel

## Checklist


0xdhrv (commented):

* [Private Internet Access Controls](https://www.raycast.com/yakitrak/pia-controls) by @yakitrak

Both extensions use Raycast to control Private Internet Access, including connecting and disconnecting the VPN, checking connection state, browsing available regions, and connecting to a selected region.

* Connect and disconnect commands
* Connection and region status
* A searchable list of PIA regions
* Connecting directly to a selected region
* Showing the currently connected region

* Richer region metadata with country flags, geo-location, and port-forwarding availability
* Favorites and recently used regions
* Port-forwarding status and forwarded-port access
* PIA settings such as protocol, LAN access, and port forwarding
* Toggle and reconnect-to-most-recent commands

However, these appear to be substantial enhancements to the existing PIA extension rather than a separate primary workflow.

The existing command IDs could also remain available for backward compatibility, so adding these capabilities would not necessarily require breaking users' current hotkeys or workflows.

Could we consider contributing the richer region browser, favorites/recents, port-forwarding support, and settings controls to the existing extension instead?

> This would help avoid duplication and keep Private Internet Access controls consolidated in one extension.

vxirau:
Happy to consolidate, you're right that keeping the existing command IDs removes the backward-compatibility concern I raised, so a separate extension isn't justified.

I'll rework this as a contribution to pia-controls: keeping connect, disconnect, status and connectToRegion as-is so existing hotkeys keep working, upgrading connectToRegion into the richer region browser, and adding toggle and connect-recent alongside them.

Two things worth flagging since it's a substantial change to @Yakitrak's extension:

Most of utils.ts would be rewritten, mainly because the region model has to carry country code, port-forwarding and geo flags (sourced from PIA's server catalog rather than the piactl slug list), and because connection state becomes an enum rather than a boolean so unreadable states aren't reported as "disconnected". I'd like @Yakitrak to weigh in on the approach before I push it.

It includes some correctness fixes beyond the new features, e.g. piactl read failures currently resolve to "disconnected" in status, and region display names are derived from the slug rather than PIA's catalog (srilanka gives "Srilanka" rather than "Sri Lanka").

Should I convert this PR in place, or close it and open a fresh one against extensions/pia-controls?

0xdhrv:
> Happy to consolidate, you're right that keeping the existing command IDs removes the backward-compatibility concern I raised, so a separate extension isn't justified.
> 
> I'll rework this as a contribution to pia-controls: keeping connect, disconnect, status and connectToRegion as-is so existing hotkeys keep working, upgrading connectToRegion into the richer region browser, and adding toggle and connect-recent alongside them.
> 
> Two things worth flagging since it's a substantial change to @Yakitrak's extension:
> 
> Most of utils.ts would be rewritten, mainly because the region model has to carry country code, port-forwarding and geo flags (sourced from PIA's server catalog rather than the piactl slug list), and because connection state becomes an enum rather than a boolean so unreadable states aren't reported as "disconnected". I'd like @Yakitrak to weigh in on the approach before I push it.
> 
> It includes some correctness fixes beyond the new features, e.g. piactl read failures currently resolve to "disconnected" in status, and region display names are derived from the slug rather than PIA's catalog (srilanka gives "Srilanka" rather than "Sri Lanka").
> 
> Should I convert this PR in place, or close it and open a fresh one against extensions/pia-controls?

Yeah, feel free to create a new PR for the existing extension, and I will prioritise it

PR #29707, opened 2026-07-24
Add MySQL Client extension

Im-Tae:
## Description
New extension: **MySQL Client** — a keyboard-first client for MySQL / MariaDB / Aurora MySQL. Connect via the `mysql2` driver, run queries, browse schema, and manage multiple connection profiles.

### Commands
- **Run Query** — run SQL and view the results as a table (copy as JSON / CSV).
- **Command Palette** — common commands (Show Databases, Process List, Status, Variables…).
- **Browse Schema** — drill through databases → tables → columns, view the `CREATE TABLE` statement, and see foreign-key relationships (with a "Copy as Mermaid ER diagram" action).
- **Query History** — re-run recent queries and keep favorites.
- **Manage Connections** — multiple profiles (host / port / user / password / database / SSL) with a default and a Test Connection action; saving verifies the connection first.

Also includes an AI tool (`query`) that runs read-only statements directly and confirms data- or schema-changing ones.

Connection details are stored in Raycast's local encrypted storage. Authentication is User & Password with SSL modes (off / require / require-insecure); AWS IAM auth is planned for a follow-up.

The icon is the official MySQL logo.

## Checklist
- [x] `npm run build` passes without errors
- [x] `npm run lint` passes without errors
- [x] Tested the distribution build in Raycast against a local MySQL 8.4 database
- [x] Screenshots added to `metadata/`


pernielsentikaer (changes requested):
Hi @Im-Tae 👋

- **Surface load errors in Browse Schema (and similar lists)** — destructure `error` from `usePromise` and show an `EmptyView` with the real message (plus Retry) instead of “No databases” / “No tables” when the connection fails.
- **Run Query Form defaults** — avoid mounting the Form while connections are still loading (or remount with a `key` once the default id is known). Right now the Connection dropdown can stay empty after load, and submit fails silently when no connection is selected — a Failure toast would help too.
- **Please add some `ai.evals`.** You have `ai.instructions` but no evals, so there's nothing pinning the tool's behaviour as the model or your descriptions change. A handful covering the common paths (list tables, describe a table, a plain SELECT, and one write that should hit confirmation) would be enough. `npx ray evals` runs them.
- **The preferences fallback is narrower than it looks.** `connectionFromPreferences` hardcodes port `3306` and `ssl: "off"`, so anyone whose server isn't on the default port or needs TLS gets a connection that can't work, with no hint why. Either add port and SSL preferences, or say in the preference descriptions that the fallback is for a plain local server and Manage Connections is the real path.

One non-blocking note: every call opens its own connection and closes it in a `finally`, so drilling databases → tables → columns is a fresh handshake each step, and `listForeignKeys` makes two. Correct, just slower than it needs to be over a remote host. Fine to leave for a follow-up.

Nice-to-haves in the same pass if easy: cap rows returned by the AI tool (e.g. 100 + `truncated`), and renaming **Command Palette** → **Common Commands** to avoid overlap with Raycast’s root UI.

---

Im-Tae:
Thanks for the thorough testing @pernielsentikaer  🙏 Addressed everything:

> - Browse Schema errors: lists and detail views now surface the real error message with a Retry action instead of "No …".
> - Run Query: the form waits for connections to load before mounting (dropdown gets its default), and submitting with no connection / empty SQL now shows a Failure toast.
> - ai.evals: added evals for listing tables, describing a table, a plain SELECT, and a write (DELETE) that hits confirmation.
> - Preferences fallback: added Port and SSL preferences so it's no longer stuck on 3306/off, and clarified in the descriptions that the fallback is for a plain local server (Manage Connections is the real path).
> - AI rows: capped at 100 with a truncated flag.
> - Renamed Command Palette → Common Commands.
> 
Left the per-call connection reuse as a follow-up as you suggested. Ready for another look 🔥

pernielsentikaer (approved):
Hi @Im-Tae 👋

Thanks for addressing the feedback so thoroughly 💪

I've re-checked HEAD: Browse Schema now surfaces real errors with Retry, Run Query waits for connections and fails loudly on empty connection/SQL, ai.evals cover the common paths, Port/SSL preferences are in place, AI rows are capped at 100, and the command is renamed to Common Commands.

Looks good to me, approved 🔥

PR #29710, opened 2026-07-24
Add Webhound to MCP Registry

mfkhalil:
## Description

Adds Webhound to the official built-in MCP registry.

Webhound runs private, budgeted reports and datasets. A user gives Hound a research question and a dollar budget; the budget controls how much research effort the investigation receives. Hound is Webhound's research harness built with DeepSeek V4 Pro and GPT-5.4 across planning, execution, verification, and assembly.

The entry uses `mcp-remote` to bridge Raycast's registry format to Webhound's hosted Streamable HTTP endpoint. Webhound's OAuth flow authenticates every connection to that user's own account; no publisher or shared credential is included.

## Verification

- `npm run lint`
- `npm run build`
- Confirmed Webhound publishes usable OAuth protected-resource and authorization-server metadata.
- Confirmed an unauthenticated request to the hosted endpoint returns `401`, rather than exposing a shared connection.
- Ran the exact `mcp-remote` bridge against production with a clean local auth
  directory. OAuth completed through Webhound, Streamable HTTP connected, and
  the client requested the tool list without any key in the Raycast entry.


0xdhrv (approved):
Looks good to me, approved ✅

PR #29712, opened 2026-07-24
Simple Reminder: multi-language input, inline add, menu bar date format

builtbybengt:
## Description

Three additions to the Simple Reminder extension:

- **Multi-language input**: reminders can now be phrased in English, German, French, Spanish, Dutch, Portuguese, Japanese, Chinese or Russian (e.g. `erinnere mich morgen um 15 Uhr an das Meeting`). The language is auto-detected per input using chrono-node's locale parsers.
- **Inline add**: the "Add Reminder" command now accepts an optional argument, so a reminder can be typed and submitted directly from Raycast's root search without opening the extension first.
- **Menu bar date format preference**: a new "Menu bar date format" preference (Relative / Short / Long / ISO 8601) controls how reminder dates are displayed in the menu bar, in addition to the existing relative-time default.

## Screencast

N/A - text/preference changes, verified via `npm run build`, `npm run lint`, and manual testing of the natural-language parsing across the supported languages.

## Checklist


pernielsentikaer (commented):
Thanks for this, @builtbybengt — nice set of improvements to Simple Reminder 🙌

Multi-language parsing, optional inline add from root search, and the menu bar date-format preference are all welcome. Changelog/manifest look good, CI is green, and this is clearly an improvement to the existing extension (not a store duplicate of Apple Reminders / etc.).

A couple of things before we merge:

1. **Guard the inline auto-submit** in `src/index.tsx`  
   The `useEffect` that creates a reminder when `!isLoading && props.arguments.reminder` can run more than once (e.g. effect re-runs / Strict Mode), which would create duplicates. A `useRef` “already submitted for this launch” flag (or equivalent) would make this safe.

2. **Remove unused `natural`**  
   The rewrite of `extractTopicAndDateFromInputText` drops `WordPunctTokenizer`, and that was the only usage in this extension. Please remove `natural` from `package.json` / lockfile so we don’t ship a dead dependency.

Nits (non-blocking):
- `parsedDate?.date as Date` is a bit sharp — an explicit null check / throw would match the toast path more clearly.
- Worth a quick DE smoke test that topic text doesn’t keep particles like “an das …” after stripping the date/prefix.
- Greptile’s note on handwritten preference types is fair long-term, but this extension already used `SimpleReminderPreferences` the same way — fine to leave as-is for this PR.

Happy to approve once (1) and (2) are in. Thanks again!

PR #29719, opened 2026-07-25
Add Windows support to Browser Tabs

artistro08:
## Description

Adds Windows support.

macOS reads tabs through AppleScript. Windows has no equivalent, so tabs are read from the tab strips of running browsers through UI Automation, in a small native helper under `rust/` built with [`raycast/extensions-rust-tools`](https://github.com/raycast/extensions-rust-tools). Nothing extra needs installing, and Firefox and Zen Browser work on Windows even though macOS cannot support them.

**macOS is untouched.** `applescript-utils.ts` is byte-identical to `main`, and a six-line dispatcher picks the implementation by `process.platform`. On a non-Windows build, `ray build` skips Rust compilation and emits the shim with `sideEffects: false`, so the binary is never called off Windows.

Windows behaviour, all documented in the README:

- Windows does not expose a tab's URL. The tab being displayed reports its own; background tabs are matched by title against browser history. Tabs with no match get no URL, and the actions needing one are hidden.
- Site icons are read from the browser's own icon store instead of being fetched, so listing needs no network access.
- Minimized windows are not listed; browsers stop exposing their tab strip while minimized.
- For browsers with workspaces, such as Zen Browser's spaces, only the current one is listed.

Nothing is bundled or downloaded: SQLite comes from `winsqlite3.dll`, which ships with Windows. Only the helper's source is committed, and `Cargo.lock` pins the tooling to a commit.

## Screencast

Existing screenshots cover macOS, which is unchanged. Happy to add a Windows screencast if useful.

## Checklist


artistro08:
Sorry for all the chatter here, but I realized the first implementation that I did was over‑engineered and very bad. So, I restarted and rebuilt everything in Rust.

pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #29720, opened 2026-07-25
Readwise Reader: Add sorting and random shuffle to List Documents

cwouyang:
## Description

This PR adds **sorting** and a seeded **Random** shuffle to the **List Documents** command.

- New **Sort Documents By…** action (`⌘⇧S` / `Ctrl⇧S`): Date Moved, Date Saved, Date Published, Date Last Opened, Author, Category, Length, Progress, Title, and **Random**.
- **Toggle Sort Direction** (`⌘⇧O` / `Ctrl⇧O`) for field sorts; **Reshuffle** (`⌘R` / `Ctrl R`) for Random. Random uses a stable seed so ordinary re-renders don't reshuffle.
- New **Default sort** preference for List Documents.
- Sorting is applied to the whole loaded list (previously each page was sorted independently).

Missing values (e.g. an empty author) always sort last regardless of direction, and equal values fall back to a stable tie-break so the order doesn't jump around.

### Note for reviewers — current display-side scope

Sorting and Random apply to the documents **currently loaded** into the Raycast list (loaded page-by-page via `usePromise` pagination), **not the entire Reader location**. The Reader Document List API has no server-side sort parameter, so a true whole-library sort would require a local index/cache. I've intentionally scoped that out as a follow-up (and say so in the README) to keep this PR focused. Happy to discuss if you'd prefer a different approach here.

### Pre-existing fixes included

While testing I hit two pre-existing issues in List Documents that this feature makes easier to trigger, so I fixed them here:

1. **Crash on a document with an empty author / website / category.** `Metadata.Label`'s `text` accepts `string | { value: string }`, and `typeof null === "object"`, so passing `null` made Raycast read `null.value` and throw. Guarded the three labels with `?? undefined`. (Sorting by Author groups such documents together, making the crash easy to reach.)
2. **Duplicate rows near the bottom of short lists.** Raycast's `pagination` re-fetches and appends the initial page on scroll-to-bottom even when `hasMore` is `false`, producing duplicate document ids (and duplicate React keys). Fixed by de-duplicating accumulated documents by `id` before rendering. (The new global sort just made the duplicates adjacent and obvious.)

Both were reproducible on `main` before this change.

## Screencast

https://github.com/user-attachments/assets/2954836d-2de7-4e98-a08b-b74447b6089b

## Checklist

### DIFF
--- a/extensions/readwise-reader/CHANGELOG.md
+++ b/extensions/readwise-reader/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Readwise Reader Changelog
 
-## [Add sorting and random shuffle to List Documents] - {PR_MERGE_DATE}
+## [Add sorting and random shuffle to List Documents] - 2026-08-14
 
 - Add a `Sort Documents By…` action to `List Documents` supporting Date Moved, Date Saved, Date Published, Date Last Opened, Author, Category, Length, Progress, Title, and Random
 - Add `Toggle Sort Direction` and, for Random, `Reshuffle`
--- a/extensions/readwise-reader/CHANGELOG.md
+++ b/extensions/readwise-reader/CHANGELOG.md
@@ -6,6 +6,7 @@
 - Add `Toggle Sort Direction` and, for Random, `Reshuffle`
 - Add a `Default sort` preference
 - Sort the full loaded list instead of sorting each page independently
+- Fix a crash when viewing a document whose author, website, or category is empty
 
 ## [Add Windows support] - 2026-07-07
 
--- a/extensions/readwise-reader/CHANGELOG.md
+++ b/extensions/readwise-reader/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Readwise Reader Changelog
 
+## [Add sorting and random shuffle to List Documents] - {PR_MERGE_DATE}
+
+- Add a `Sort Documents By…` action to `List Documents` supporting Date Moved, Date Saved, Date Published, Date Last Opened, Author, Category, Length, Progress, Title, and Random
+- Add `Toggle Sort Direction` and, for Random, `Reshuffle`
+- Add a `Default sort` preference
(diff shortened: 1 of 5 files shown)


cwouyang:
Addressed the P2 (manually duplicated preference type): replaced the hand-written `Preference` type with Raycast's generated `Preferences.ListDocuments` in `getPreferenceValues` (55fdab9). The generated type already types `defaultSortBy` as the exact value union from the manifest, so it stays in sync with `package.json` and needs no cast.

cwouyang:
Friendly follow-up — this has been green and ready for about a week. @zach-fuller (or any of the maintainers) if you have a moment to take a look, I'd be glad to address any feedback. 🙏

pernielsentikaer (changes requested):
Hi @cwouyang 👋

1. **Empty state while loading with a category filter** in `src/list-documents.tsx`  
   The condition changed from `data?.length === 0` to `displayData.length === 0`. Because `displayData` is built from `data ?? []`, when `data` is still `undefined` (e.g. after changing category and `usePromise` reloads) this can briefly show “No documents found” during loading. Please gate on loading/data, e.g. `!isLoading && displayData.length === 0 && category !== undefined` (or keep an explicit `data !== undefined` / `data?.length === 0` style check).

---

cwouyang:
Good catch — fixed in `436c9f9`. The empty state now renders only once documents have actually loaded (guarded on `data?.length === 0`), so it no longer flashes "No documents found" while a category reloads, and it also stays hidden on a fetch error (when `data` is still `undefined`). Ready for another look whenever you have a moment 😊

pernielsentikaer (approved):
Hi @cwouyang 👋

Thanks for the follow-up 💪

Re-reviewed HEAD — the empty state is gated on `data?.length === 0` again, so “No documents found” no longer flashes while a category reloads, and it stays hidden on a fetch error.

Looks good to me, approved 🔥

PR #29727, opened 2026-07-25
Update Wrap Unwrap extension

chrismessina:
# Update Wrap Unwrap extension

## Description

Correctness and performance pass over the wrap/unwrap reflow engine. Seven rounds of
adversarial review plus property fuzzing surfaced 30 defects; every one was reproduced
with a concrete input before being fixed, and each is covered by a regression test.

Two of them are the reason for this submission:

**1. Large pastes could hang Raycast.** Four quadratic code paths meant an input inside
the extension's own supported 1 MB limit could take minutes with no progress shown.
Measured before → after: 188 KB of indented code 10,299 ms → 14 ms; an 833 KB single
paragraph 18,031 ms → 22 ms; 1 MB of prose 32,056 ms → 39 ms. The causes were a per-line
rescan of all prior records in the classifier, an anchored regex tested against the whole
accumulated paragraph, a `slice()` on that accumulator forcing V8 to flatten the rope on
every join, and a per-break scan of all remaining tokens.

**2. `Strip Soft Hyphens` destroyed real compounds.** `well-` + `known` unwrapped to
`wellknown`. An ASCII `-` at a line break cannot be distinguished from a genuine compound
by pattern, so the preference now strips only a true Unicode soft hyphen (U+00AD), which
exists solely to mark an optional break. An ASCII hyphen is always preserved and the two
halves join tight. The preference description and README both promised
`inter-` + `esting` → `interesting`; both now describe the actual behavior, since the old
text would have shipped a false claim in the settings UI.

## Other fixes

**Data loss / corruption**

- A trailing `>` wrapped onto its own line re-read as an empty blockquote and the
  character was **deleted**. `---`, `===`, `***`, `___`, and `|---|---|` similarly became
  horizontal rules, setext underlines, or table separators. A line is now judged by what
  it would parse as, so multi-token constructs (`_ ___`) are caught too.
- A literal hyphen inside inline code or a link URL was corrupted when the span broke
  across lines (`` `foo-` `` + `` `bar` `` → `` `foobar` ``).
- Text containing Unicode private-use characters was silently rewritten or deleted by the
  internal inline-token placeholder.

**Structure**

- Blockquote indentation was discarded, hoisting a quote out of its containing list item.
- A list-nested blockquote merged with a following root-level one, and a single quote was
  split when its markers carried different but legal indentation (CommonMark allows 0–3
  spaces before `>`).
- Continuation lines under a `>` marker written without a following space never rejoined.
- A fenced block was closed early by a line carrying an info string, exposing its code to
  reflow.
- Tab-indented code was reflowed as prose. Indent width is now measured in columns, so
  any tab/space mix reaching column 4 is recognized.
- A link label with nested brackets, and a URL containing balanced parentheses, were split
  mid-link.

**Reflow quality**

- Hyphen rejoining failed for non-ASCII text (accented Latin, Cyrillic, Greek, CJK, and
  astral characters kept a dangling space).
- Wrapping overran the requested column on tab-indented lines; a tab counts as four
  columns, not one character.
- Tab-indented and task-item (`- [ ]`) continuation lines were misaligned.
- An even-numbered run of trailing backslashes was treated as a hard break instead of
  escaped literal backslashes, and backslash-escaped inline delimiters were read as real
  ones.
- An invalid `Wrap Column` value with a numeric prefix (`12px`) was accepted as `12`
  instead of falling back to 80.
- `String(error)` could throw on a null-prototype value, escaping the failure toast.

## Checklist

## Verification

```
npm test          → 147 pass, 0 fail   (was 102)
npx tsc --noEmit  → exit 0
npx ray lint      → exit 0
npx ray build     → exit 0
```

Beyond the unit tests, three property harnesses ship as permanent regression guards:

- **List-shape round trip** — 10,080 combinations (7 indents × 10 markers × 3 gaps ×
  4 task states × 4 quote forms × 3 widths): 0 failures. This is what found the
  no-space-`>` bug that four review rounds had missed.
- **Inline-state consistency** — 21,632 probes asserting no join loses or fabricates
  characters: 0 failures.
- **Trailing-construct round trip** — every construct shape the classifier recognizes as
  a final token: 62/62.

Plus explicit performance guards, so a reintroduced rope-flattening `slice()` or
full-accumulator regex fails a test instead of silently hanging in the field.

No new commands, preferences, or dependencies. No user-visible behavior changes beyond
the fixes above and the corrected `Strip Soft Hyphens` description.

### DIFF
--- a/extensions/wrap-unwrap/.prettierrc
+++ b/extensions/wrap-unwrap/.prettierrc
@@ -1,0 +1,4 @@
+{
+  "printWidth": 120,
+  "singleQuote": false
+}
--- a/extensions/wrap-unwrap/CHANGELOG.md
+++ b/extensions/wrap-unwrap/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Wrap Unwrap Changelog
 
-## [Reflow Correctness and Large-Paste Performance] - {PR_MERGE_DATE}
+## [Reflow Correctness and Large-Paste Performance] - 2026-07-26
 
 - Fix **Unwrap Text** joining a hyphen-broken word with a dangling space — `well-` + `known` now rejoins as `well-known`, not `well- known`. Applies to compounds, hyphen chains (`state-of-the-art`), and numeric ranges (`5-10`).
 - Fix **Strip Soft Hyphens** destroying real compounds. The preference now removes only a true Unicode soft hyphen (U+00AD), which exists solely to mark a line break; an ASCII `-` is always preserved, since it cannot be distinguished from a genuine compound.
--- a/extensions/wrap-unwrap/CHANGELOG.md
+++ b/extensions/wrap-unwrap/CHANGELOG.md
@@ -1,5 +1,32 @@
 # Wrap Unwrap Changelog
 
+## [Reflow Correctness and Large-Paste Performance] - {PR_MERGE_DATE}
+
+- Fix **Unwrap Text** joining a hyphen-broken word with a dangling space — `well-` + `known` now rejoins as `well-known`, not `well- known`. Applies to compounds, hyphen chains (`state-of-the-art`), and numeric ranges (`5-10`).
+- Fix **Strip Soft Hyphens** destroying real compounds. The preference now removes only a true Unicode soft hyphen (U+00AD), which exists solely to mark a line break; an ASCII `-` is always preserved, since it cannot be distinguished from a genuine compound.
+- Fix hyphen rejoining for non-ASCII text — accented Latin, Cyrillic, Greek, CJK, and astral characters now rejoin correctly instead of keeping a dangling space.
+- Fix a literal hyphen inside inline code or a link URL being corrupted when the span broke across lines (`` `foo-` `` + `` `bar` `` no longer becomes `` `foobar` ``). Link destinations with balanced parentheses are handled correctly.
+- Fix blockquote indentation being discarded, which hoisted a quote out of its containing list item.
+- Fix **Unwrap Text** failing to rejoin lines inside a blockquote written without a space after the marker (`>text` rather than `> text`) — the wrapped continuation lines were left as separate lines instead of reflowing. `>text` and `> text` are the same quote, so they now group together.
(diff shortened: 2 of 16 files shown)


chrismessina:
Pushed a fix for both review findings. Both reproduced first, both have regression tests.

**1. `classify.ts` — list context went stale.** Confirmed: the state deciding whether an indented blockquote is nested inside a list item was cleared only on a blank line, but a line back at the margin closes a list too.

```
1. item
root prose closes the list
   > alpha beta
> gamma delta
```

The 3-space quote was still measured against the list item's content column and marked list-nested, so it split from the root quote below it instead of reflowing as one paragraph. A `closesOpenList` predicate now clears the state on any unindented, unquoted, non-list line. Indented continuations are excluded (they belong to the open item) and so are quoted lines (they are the records the flag exists to classify).

**2. `wrap.ts` — the probe was O(width) per rejected break.** Also confirmed, and the more serious of the two. Gathering tokens up to `contBudget` meant each rejected break re-scanned up to a full line's worth of characters, and `width` has no upper bound — the preference is uncapped and a cross-extension `launchContext` supplies it directly, bypassing `parseWidth`.

| input | before | after |
| --- | --- | --- |
| 20k unsafe tokens @ width 20000 | 4048ms | 22ms |
| 40k unsafe tokens @ width 20000 | 9564ms | 49ms |
| width 80 → 20000 (same input) | 46ms → 4048ms | 28ms → 22ms |

The probe is now capped at 8 tokens as well as the budget. Whether a line *starts* a block is decided by its opening token or two — a list marker, a fence, an ATX run, a setext underline — so a handful is always sufficient. Runtime is now flat in width and linear in run length.

You were right that the width-80 perf test did not cover this path; there is now a guard at width 20000, paired with a detection assertion so the bound cannot be "fixed" by weakening it.

Verified the cap does not weaken block detection: the 62-construct trailing-token round trip, a 10,080-case list-shape fuzz, and the multi-token synthesis cases (`_ ___`, a trailing `>`) all still pass, including at width 5000 where the cap binds.

Tests 147 → 151. `tsc --noEmit`, `ray lint`, and `ray build` all clean on the built tree.

0xdhrv (approved):
Looks good to me, approved ✅

PR #29728, opened 2026-07-25
Add music-recognizer extension

erenalpars:
## Description

**Identify the song playing on your PC with Shazam.**

It records a few seconds of system loopback audio ("what you hear"), so it works with any source: Spotify, YouTube, a game, a livestream; nothing needs to be playing in a supported app. The match comes back with cover art, album and release year, plus one-keystroke links to Spotify, YouTube Music, Apple Music and Shazam. Every match is saved locally and searchable later.

**Two commands:** Recognize Song and Recognition History.
**Preferences:** Recording duration (3–12 s) and a primary music service, which becomes the Enter action on a match.

**Windows-only.** Capture is WASAPI loopback through a bundled PowerShell/C# recorder that writes 16 kHz mono PCM directly; no ffmpeg, no DLLs, no drivers.

**How it works:** Shazam has no public API, so the extension uses the same endpoint the official clients do; requests go straight from the user's machine and only the fingerprint ever leaves it, never the audio. Fingerprinting happens locally via `shazamio-core` (MIT, bundled as WebAssembly; its origin and SHA-256 are documented in the README). The dependency chain is plain MIT; the GPL-2.0 alternatives were avoided. The README states plainly that this is an unofficial integration, not affiliated with Shazam or Apple.

## Screencast

music-recognizer-1
music-recognizer-2
music-recognizer-3

## Checklist


erenalpars:
Both Greptile findings are addressed: the captured WAV is now removed on every path (guarded `finally`, so a locked file can't block retries), and preferences use the generated `Preferences` types.

Shazam has no public API, so the extension uses the same endpoint the official clients do; requests come straight from the user's machine and only the fingerprint leaves it, never the audio. The dependency chain is plain MIT (GPL-2.0 alternatives avoided); the bundled wasm's origin and SHA-256 are in the README.

The neutral "Socket Security: Pull Request Alerts" check failed with an error Socket attributes to their own API/bot; their "Project Report" check passed.

erenalpars:
comment to remove stalled status

0xdhrv (commented):

The current implementation in `src/lib/shazam.ts` (`requestRecognition`) POSTs fingerprints to `https://amp.shazam.com/discovery/v5/en/US/iphone/-/tag/...` with headers that impersonate the official Shazam iPhone client (`User-Agent: Shazam/14.1.0 (iPhone; iOS 16.0; Scale/3.00)`, `X-Shazam-Platform: IPHONE`, `X-Shazam-AppVersion: 14.1.0`). The README also states that Shazam has no public API and that this is an unofficial integration.

We can't accept that as the Store integration. Submissions need to follow the terms of the services they use:

> Please check the terms of service of third-party services that your extension uses.

and

> Violates the Terms of Service of the service provided. Example: scraping a website without permission.

Apple's supported path is [ShazamKit](https://developer.apple.com/shazamkit/), which isn't available to a Windows Raycast extension this way. Could we switch to a music-recognition service that publishes an API for third-party apps, instead of the undocumented client endpoint?

The Store screenshots (`metadata/music-recognizer-1.png` through `music-recognizer-5.png`) still show the local-development indicator on the command icon. Could you recapture them with Window Capture (Save to Metadata ticked) so that icon isn't in the published images?

Separately, [Music Recognition](https://www.raycast.com/lemikeone/music-recognition) by @lemikeone already covers **Identify Song** on macOS through the official Shortcuts **Recognize Music (Shazam)** action. This PR's Windows loopback capture is a real platform difference, so a Windows listing isn't automatically a duplicate. If there's a permitted API, we can look at uniqueness again.

---

PR #29733, opened 2026-07-26
Add claude-artifacts extension

chrismessina:
## Description

Search your Claude Code artifacts from Raycast and open the right one in one keystroke.

Artifacts are easy to *make* and weirdly hard to *find again* — a Claude Code agent finishes an overnight build, hands you an artifact to review, and two days later it's buried in forty browser tabs. This extension makes them searchable from where you already start every task, sorted most-recent-first because the one you want is almost always the one you just made.

**One command:** Search Artifacts — a list with relative dates ("2d ago"), the publishing project as a subtitle, and a project filter.

## Screenshot

screenshot-1

## Checklist

## How it works — and why there's no API call

**This extension makes no network requests.** There is no artifacts API available to individual users, so it doesn't call one. Instead:

1. A Claude Code `PostToolUse` hook (`scripts/record-artifact.sh`) appends each artifact to a local index at `~/.claude/artifacts.json` as it's published.
2. The Raycast command reads that file — instantly, offline, no credentials.

This was researched before settling on it. The alternatives, and why each is out:

| Approach | Why not |
|---|---|
| Anthropic Messages API | Stateless; no access to account artifacts. |
| Compliance API | The right endpoint, but needs an **Enterprise** Compliance Access Key and is org-scoped, not personal. Unavailable on Pro/Max. |
| Headless `claude -p` | The artifact tool is interactive-only; verified not exposed to headless mode. |
| claude.ai internal endpoints | Requires a session cookie — prohibited by Anthropic's [Consumer Terms](https://www.anthropic.com/legal/consumer-terms) §3 (automated access). **Not implemented, and never will be.** |
| Reading local app data | The desktop app stores no artifact index — evictable browser cache only. |
| An MCP server | Doesn't exist; the [feature request](https://github.com/anthropics/claude-code/issues/12858) was closed as not planned. |

**Reviewer note on the two shell scripts:** they're user-installed Claude Code hooks, not code the extension executes. They live in the repo rather than the bundle deliberately, so a user can read exactly what will run on their machine before installing it. Neither makes a network call. `record-artifact.sh` writes only the index and its own log; `probe-artifact-hook.sh` is a read-only diagnostic that appends to a temp file. The extension itself never spawns them.

## Deliberate design decisions

- **Separate from the existing [Claude extension](https://www.raycast.com/florisdobber/claude).** That one authenticates with an Anthropic API key; artifacts live behind the claude.ai *account* session and are unreachable with an API key. Bundling them would prompt for a credential that plays no part in the feature working — so the first time the list broke, you'd debug the wrong one.
- **macOS-only.** The command is portable (`os.homedir()` resolves on Windows), but the recording hook is `bash` + `jq`. Claiming Windows support would mean a user installs it, the command loads, and the list stays permanently empty with no explanation. The README documents exactly what a PowerShell port would need.
- **No "Refresh" command.** The index can drift (renames and deletions don't propagate). That's real but slow, and the problem being solved is recency, not archival completeness. The schema carries a `version` field to reserve the seam. The README states the limitation plainly.
- **Uses [`@chrismessina/raycast-kit`](https://www.npmjs.com/package/@chrismessina/raycast-kit)** for the failure toast, so it carries a Copy Error action and redacts secrets from clipboard text. Zero runtime deps, `@raycast/api` as a peer.

## Testing

Beyond `tsc --noEmit` / `ray build` / `ray lint`, the index reader was exercised against 13 fixture states — missing file, empty file, malformed JSON, truncated JSON, non-array `artifacts`, bare array, single item, duplicate ids, junk rows (`null`, numbers, strings, id-less objects), very long titles, and the full index. Each renders a specific state rather than a blank list; a bad row is skipped rather than blanking everything.

The hook was tested against five payload shapes and at 40-way concurrency (40/40 rows, no duplicates, no corruption). It uses a kernel-backed `flock(2)` lock — an earlier lockfile-plus-age-based-reaper design silently lost rows under burst publishing, because a lockfile's mtime can't distinguish a slow writer from a dead one. Notes in `docs/hook-payload.md`.


0xdhrv (approved):
Looks good to me, approved ✅

PR #29735, opened 2026-07-26
Add rich-emoji-search extension

maxludden:
## Checklist


0xdhrv (commented):
Thanks for the contribution!

There seems to be quite a bit of overlap here with both Raycast's native **Emoji & Symbols** feature and the existing [Emoji Search](https://www.raycast.com/FezVrasta/emoji) extension, which already provides emoji searching and inserting/copying.

I can see that this extension adds some functionality specifically for Python Rich, such as searching Rich's own emoji aliases and copying/pasting the exact `:name:` markup. However, I'm not sure that warrants another standalone emoji-search extension when we already have an established extension covering the same core workflow.

Would it make sense to contribute the Rich-specific functionality to the existing Emoji Search extension instead? For example, Rich markup/name actions or additional aliases could potentially enhance the existing extension without introducing another extension with largely overlapping functionality.

As mentioned in our [Extensions Guidelines](https://manual.raycast.com/extensions-guidelines#before-acceptance), we generally try to avoid extensions that provide comparable functionality to an existing Raycast feature or Store extension unless there is a sufficiently distinct use case.

PR #29736, opened 2026-07-26
[Safari] Restore local tabs in Search Tabs

gumadeiras:
## Description

`Search Tabs` could show only iCloud tabs while local Safari tabs were still loading. With more than 500 open tabs, loading local tabs took approximately 15 seconds.

### Cause

The Swift bridge requested each tab's title and URL separately through ScriptingBridge. This produced two Apple events per tab.

The command also checked `!devices` for its loading state. Because `devices` starts as an empty array, the loading indicator was never shown.

### Fix

- Fetch titles and URLs in one batched request per Safari window.
- Fall back to individual reads only when Safari omits empty values from a batch. This preserves tab positions and empty titles or URLs.
- Pass the actual local-tab loading state to both commands that use `useDevices`.
- Do not query the iCloud tabs database when iCloud tabs are disabled.
- Resolve `blob:` favicons through their origin, with `Icon.Link` as the fallback.

There are no command, preference, schema, dependency, or tab-identifier changes.

## Validation

Tested on macOS 27 and Safari 27 with Raycast v1 1.104.23 and Raycast v2 Beta 0.70.0.0.

- Installed Store extension with iCloud disabled: no local results.
- Patched extension with iCloud disabled: 518 local tabs.
- Patched extension with iCloud enabled: local and iCloud tabs.
- Swift and JXA returned the same 517 tabs in the comparison test.
- Local tab loading improved from approximately 14.66 seconds to 0.48 seconds.
- A blank tab retained its empty URL and correct index.
- Real `blob:` tabs rendered without invalid URL errors.
- `npm ci`
- `npm run build`
- `npm run lint`
- Distribution build with Node 22.22.2
- `git diff --check`

Related: #29296 touches the same Swift bridge and hook for multi-Safari support, but does not address this loading regression.

## Screencast

Not included because the test profile contains private tab titles.

## Checklist

- [x] I read the publishing documentation
- [x] I built the distribution and tested the extension in Raycast
- [x] No asset changes are required

### DIFF
--- a/extensions/safari/CHANGELOG.md
+++ b/extensions/safari/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Safari Changelog
 
-## [Fix] - {PR_MERGE_DATE}
+## [Fix] - 2026-08-12
 
 - Restore fast, complete local tab loading while preserving tabs with empty titles or URLs.
 - Avoid querying iCloud tabs when they are disabled.
--- a/extensions/safari/CHANGELOG.md
+++ b/extensions/safari/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Safari Changelog
 
+## [Fix] - {PR_MERGE_DATE}
+
+- Restore fast, complete local tab loading while preserving tabs with empty titles or URLs.
+- Avoid querying iCloud tabs when they are disabled.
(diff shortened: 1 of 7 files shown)


pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #29739, opened 2026-07-26
Update get-app-icon extension

chrismessina:
## Description

An update to **Get App Icon** with one visible headline change and a batch of correctness fixes underneath.

### Grid icons are now sharp

The grid rendered `{ fileIcon: app.path }`. `NSWorkspace.icon(forFile:)` returns an image whose **nominal size is 32pt** even though it carries representations up to 2048×2048 — so a ~128pt grid tile was upscaling a 32pt bitmap ~4×. List rows looked fine because they render close to that nominal size; the grid did not.

get-app-icon-grid

The extension now extracts real 256px PNGs and points grid tiles at those, falling back to `fileIcon` until an app's entry exists.

The naive approach — one `xcrun swift` per icon — costs **~1.08s each**, almost entirely Swift compilation, which is unusable for a few hundred apps. Batching every app through a **single** process amortises that away: **~1.6s for 262 apps** (~6ms of real work per icon). Results are cached in `environment.supportPath`, invalidated by bundle mtime, and pruned when an app is uninstalled. A progress toast counts icons as they land, so the first run never looks stalled.

### Simpler preferences

The eight per-size checkboxes are gone, replaced by a single **Default Size** dropdown. Per-size export moved into a new **Export Icon Size…** submenu (mirroring the existing **Copy Icon Size…**), so a one-off size no longer requires changing preferences first. Format toggles and their descriptions were rewritten — they previously restated the option name back at the user.

ICNS is deliberately excluded from the size submenu: an `.icns` file always contains every size, so honouring a single-size request there would be a lie. It falls back to PNG when ICNS is the only enabled format.

### Shortcuts

Migrated to `Keyboard.Shortcut.Common` where the semantics genuinely match:

| Action | Before | After |
| --- | --- | --- |
| Copy Icon | `⌘⇧C` hand-rolled | `Common.Copy` (same chord) |
| Copy App Path | `⌘.` | `Common.CopyPath` (`⌘⌃C`) |
| Copy App Name | `⌘⇧.` | `Common.CopyName` (`⌘⌥C`) |

`⌘.` is Raycast's reserved **Pin** chord, so "Copy App Path" was squatting on it.

**Show in Finder** no longer binds `⌘↩` — Raycast reserves that for a panel's secondary action and was already assigning it to "Export All Sizes". Note `@raycast/eslint-plugin`'s `no-reserved-shortcut` does **not** catch this: it compares `key` literally and never normalises `"return"` to `"enter"`, so a green lint was not evidence the panel was conflict-free.

`⌘E` on "Export Icons" is left as a custom shortcut. It collides with `Common.Edit`'s chord, but exporting is not editing, and a wrong `Common` is worse than an honest custom binding.

### Error handling

Every failure path now carries a **Copy Error** action (previously zero of four did), via [`@chrismessina/raycast-kit`](https://www.npmjs.com/package/@chrismessina/raycast-kit) — zero runtime dependencies, `@raycast/api` as a peer. Errors that aren't `Error` objects render a readable message instead of `[object Object]`.

**Show Export Folder in Finder** used to dead-end with "Export folder not found" for any app you hadn't exported yet. It now opens your output folder and says why.

### Cache hardening

From an adversarial review pass:

- Extractor job fields are **base64-encoded**, so an app path containing a tab or newline can't forge an extra record and write outside the cache directory.
- Cache files are keyed by `sha256(fullPath)`. The previous scheme replaced non-alphanumerics with `_`, collapsing `A-B.app` and `A_B.app` onto one entry — two apps sharing a cache file, one showing the other's icon.
- Writes go to a temp file and are renamed into place, and per-icon success is reported honestly rather than assumed.
- Leaving the grid aborts the extractor process instead of letting it run on, and the progress toast is always dismissed.
- Exporting an app invalidates its cached tile — exports read the live bundle, so that's the natural moment to self-heal a stale icon without probing every app's version on launch.

### Housekeeping

- README images moved to `media/` (out of `metadata/`, and deliberately not into `assets/`, which ships inside the extension).
- Added a third screenshot; all three are 2000×1250.
- Removed `TODO.md` and `.windsurf/rules/` — local dev scratch, not Store content.
- `tsconfig.json` gained a `types` field so editors resolve `@types/node` (CLI builds were unaffected).

## Screencast

The updated screenshots in `metadata/` show the change directly — the grid tiles are now rendered from extracted 256px icons rather than the upscaled 32pt system image.

## Checklist


chrismessina:
@greptile could you take another look, especially looking at the PR body which addresses your concerns?

0xdhrv (approved):
Looks good to me, approved ✅

PR #29740, opened 2026-07-26
[Safari] Handle large Safari bookmark files

gumadeiras:
## Description

Prevents Safari bookmark loading from failing with `maxObjectCount exceeded` for users with large `Bookmarks.plist` files.

Safari now raises the binary plist parser object limit before reading the local Safari bookmark file, while leaving bookmark traversal and command routing unchanged.

`bplist-parser` was already installed through `simple-plist`. Safari now declares it as a direct dependency because it imports the parser directly to configure the instance used in the production Raycast bundle. npm deduplicates both references to `bplist-parser@0.3.2`, so this does not add another parser copy.

This applies the same fix and limit as the merged [Browser Bookmarks PR #28062](https://github.com/raycast/extensions/pull/28062).

Validation:

- `npm ci`
- `npm run lint`
- `npm run build`
- `ray build --environment dist`
- `git diff --check`
- Tested the production distribution in Raycast 1.104.23
- Tested Search Bookmarks with an 86,319-object Safari plist
- Tested Search Reading List
- Tested Search Tabs, Bookmarks and History

## Screencast

Not included; this is a parser-limit fix with no UI change.

## Checklist

- [x] I checked that files in the `assets` folder are used by the extension itself (this PR does not add or change assets)
- [x] I checked that assets used in the `README` are located outside the metadata folder if they were not generated with the [metadata tool](https://developers.raycast.com/basics/prepare-an-extension-for-store#how-to-use-it) (the Safari extension has no README)


pernielsentikaer (commented):
Hi @gumadeiras, thanks for this — nicely investigated and the write-up made it easy to verify.

I checked the mechanism rather than taking it on trust: `bplist-parser@0.3.2` reads `exports.maxObjectCount` at parse time, so mutating the imported module object does raise the limit for the reads that go through `simple-plist`. 

The lockfile keeps a single top-level `bplist-parser@0.3.2` (and `simple-plist` pins that exact version), so there's no second parser copy. And since `useBookmarks.ts` is the only plist read path and every command plus both AI tools import it, the module-level setting covers everything.

One suggestion: the merged Browser Bookmarks fix (#28062) reached the same parser via `import { bplistParser } from "simple-plist"`, which simple-plist re-exports. This extension already has a hand-written `src/simple-plist.d.ts`, so adding `bplistParser` to that declaration would get you the same result without the new direct dependency or the extra `src/bplist-parser.d.ts`. It also keeps the two Safari bookmark fixes consistent.

Minor, related: `src/bplist-parser.d.ts` types the module as only `{ maxObjectCount: number }`, so anyone later importing `parseBuffer` from it would hit a confusing type error.

Happy to merge either way — if you'd rather leave it as is, say so and I'll approve.

gumadeiras:
> Hi @gumadeiras, thanks for this — nicely investigated and the write-up made it easy to verify.

thanks! updated the PR with your suggestions: now using `bplistParser` through the existing `simple-plist` import; removed the direct `bplist-parser` dependency and its declaration file, and added the re-export to the existing `simple-plist` declaration. This also removes the incomplete module type

pernielsentikaer (approved):
Thanks for switching to the simple-plist re-export — looks good. Approving.

PR #29758, opened 2026-07-27
Kesha Voice Kit: setup guidance and faster failure feedback

drakulavich:
## Description

Update to the existing **Kesha Voice Kit** extension. It reworks what happens when the underlying CLI isn't installed, isn't finished setting up, or the microphone never delivers audio — previously each of those produced a session that silently could not produce a transcript.

- **Not-found guidance rewritten** as numbered setup steps, leading with Homebrew (`brew install drakulavich/tap/kesha-voice-kit`) rather than assuming the user knows bun, and including the required `kesha install` step. The probed-path listing is demoted to a trailing troubleshooting line.
- **Error views gained an action panel**: copy the error text, open extension preferences, or open the setup guide.
- **Preflight before recording**: the resolved CLI is probed for version and engine availability, so an unfinished setup renders a "finish setup" view naming the exact remaining command instead of starting a recording. The probes run in parallel and fail open, so a probe that cannot run never blocks dictation.
- **Early no-signal warning**: if the level meter reports no sample within ~8 s, a warning points at the macOS Microphone permission. It is deliberately non-blocking — a broken meter must not abort a session that may still be capturing audio through the CLI.
- **Idle auto-stop now also advances while the meter is unavailable**, so a session with a dead meter stops at the silence deadline instead of running to the full recording cap.

No new dependencies, no new assets, no changes to `package.json`, commands, or preferences.

## Screencast

No UI-flow change to screencast: the recording view is unchanged. The changes are in error/setup states and in the timing of an existing message. Existing store screenshots remain accurate.

## Checklist

- [x] I ran `npm run build` (see note below on the in-Raycast check)

Verified on this branch: `npm run build`, `npm run lint`, and the extension's vitest suite (75 tests) all pass. The manual walkthrough of the distribution build inside Raycast is being done separately by the author and I will confirm it in a comment before this needs review — flagging it explicitly rather than ticking the box for a check that hasn't happened yet.


drakulavich:
@0xdhrv sorry to ping you directly — you reviewed the initial version of this extension in #29681, so you have the most context on it. Whenever you have a spare moment, would you mind taking a look at this follow-up? No rush at all, and I completely understand if the queue is long.

It is an update to the existing extension, no new commands or dependencies. Short version: a first-run user without the CLI, or with the CLI but without the engine models, previously got a recording session that could not produce a transcript; they now get actionable setup guidance before recording starts, with actions on the error view. A microphone that delivers no audio now surfaces guidance within a few seconds instead of leaving the session silent.

The review bot found two real bugs in the recording lifecycle during this PR, both fixed and covered by regression tests I verified fail against the old code. The diff also ended up smaller than when I opened it, since I removed a redundant probe and some duplication along the way.

Happy to split anything out or adjust if you would prefer smaller pieces. Thanks for your time either way.

0xdhrv (approved):
Looks good to me, approved ✅

PR #29762, opened 2026-07-27
Add media-switcher extension

muhammadrizo-y:
## Description

The Media Switcher extension provides a centralized, keyboard-driven way to manage active media sessions directly from Raycast. It streamlines audio and video workflows by allowing users to toggle playback, adjust volume, and switch between competing media sources without needing to hunt through open applications.

*   Switch seamlessly between active media sessions, with the ability to pause others while activating your selection.
*   Control playback state, track navigation, and volume levels using standardized shortcuts for both macOS and Windows.
*   Bring target media applications to the foreground instantly via the "Reveal Application" action.
*   Configure volume sensitivity through a dedicated `Volume Step` preference.

## Checklist


muhammadrizo-y:
@greptileai

The implementation is intentionally best-effort rather than perfect because the Windows media session API does not provide a stable session ID. Session identity is therefore approximated using ordinals plus a title prefix, which is inherently hacky and can only be made so reliable.

You should not keep chasing rare edge cases or minor inconveniences with increasingly complex recovery logic. In particular, if a switch fails, paused competitors should not be automatically restored just to avoid a small extra step for the user. The user can replay or unpause manually.

These are conscious tradeoffs and acceptable limitations of the Windows API, not bugs in the implementation.

PR #29763, opened 2026-07-27
Add relay extension

genna-cervantes:
## Description

Adds **Relay**, a new extension for moving files from your Mac to a remote machine over SSH.

The extension provides a single command, **Upload Clipboard File**. It reads your last few
clipboard entries, shows any files among them in a thumbnail grid, and uploads the one you
pick to a configured SSH host with `scp`. The remote path is copied to your clipboard on
success so you can paste it straight into a terminal, editor, or chat.

Details:

- Two preferences: **SSH Host** (a `user@host` destination or a `~/.ssh/config` alias) and
  **Destination Directory** (absolute or `~/`-relative, created with `mkdir -p` if missing).
- Uploads are renamed to `clipboard-<timestamp>.<ext>` to avoid collisions while keeping
  the original extension.
- Authentication is key-based only — Relay shells out to non-interactive `ssh`/`scp`, so
  common failures (bad key, unknown host, unwritable directory, unreachable host) are mapped
  to specific, actionable toast messages instead of raw stderr.
- The Raycast window closes with a HUD confirmation once the upload finishes.

## Screencast

https://github.com/user-attachments/assets/982a12ff-457a-41a1-a5de-47692cfcd7e5

## Checklist


pernielsentikaer (changes requested):
Hi @genna-cervantes 👋

Thanks for contributing Relay!

We already have [SSH Image Drop](https://www.raycast.com/wjsong/ssh-image-drop) in the Store by @free4416, which supports uploading files over SSH and copying their remote paths.

Your recent-clipboard-file grid and timestamp naming add a different workflow, but we’d prefer bringing compatible improvements into the existing extension. Would you be open to contributing those there instead? If there’s a reason combining them wouldn’t work, could you explain it here? 🙂

---

PR #29781, opened 2026-07-27
p00f: fix "crypto is not defined" crash on every command

mcdays94:
## Description

Hotfix for `p00f`, which I submitted in #28989 and which merged earlier today. **Every command in the published extension fails**, so the extension is currently unusable in the Store.

Running any of the three commands throws:

```
ReferenceError: crypto is not defined
```

Reproduced on two separate machines, using the Store build (not a dev build).

## Root cause

p00f encrypts on-device before anything is uploaded. The crypto lives in `@p00f/core`, which is deliberately written against the **ambient Web Crypto global** so one implementation runs unchanged in browsers, in `workerd`, and under Node:

```js
crypto.getRandomValues(iv);
await crypto.subtle.importKey("raw", ikm, "HKDF", false, ["deriveKey"]);
```

Raycast's extension host runs each command in a worker scope that has **no binding for `crypto`**. So the very first `getRandomValues` throws, before any request leaves the machine.

I missed this because every environment I tested in resolves the global fine. A plain `node` process, a plain `worker_threads` Worker, and Raycast's own bundled Node runtime (`v22.22.2`) all expose `globalThis.crypto`. Only the actual extension-host scope does not, so `ray build` and `ray lint` both stayed green and the bug reached the Store.

## Fix

Fixed in the shell, not in core. Core keeps its browser and `workerd` compatibility, and importing `node:crypto` there would break both.

New `src/lib/webcrypto.ts` installs Node's `webcrypto` on `globalThis` when the host provides nothing usable, and is imported first in all three entry points. Two deliberate details:

- **It no-ops when a working global already exists**, so a genuine host implementation is never shadowed by the Node one. It also treats a *partial* global (present but missing `subtle`) as needing the polyfill, since that is just as unusable for HKDF and AES-GCM.
- **`defineProperty` rather than assignment.** Where `crypto` is a getter-only accessor, a plain assignment silently no-ops and leaves the broken value in place.

Core reads `crypto` lazily inside its functions and never at module scope, so importing the shim first is sufficient.

## Verification

Confirmed the mechanism rather than guessing at it. Deleting `globalThis.crypto` to emulate the host scope reproduces the reported error exactly, and the shim repairs it:

```
precondition: crypto global present initially = true | after delete = undefined
  [REPRO] ReferenceError: crypto is not defined
  [SHIM] applied. typeof crypto = object | subtle = object
  [VERIFY] encrypt+upload OK: https://p00f.me/c/XsuQD0UG5T0DWJEQ...
  [VERIFY] reveal ok: true | plaintext matches: true
```

Also verified:

- The shim **survives bundling** in all three built commands, and is ordered before the first `crypto` use (byte offset 497 vs 913).
- Clean `npm ci` then `ray build` and `ray lint`, all green.
- 7 new regression tests upstream, 192 green overall.

## Note on user data

No user content was ever at risk. The failure is local and happens *before* anything is transmitted, so no Poof was created, uploaded, or partially encrypted. The zero-knowledge model is unchanged: the server still only ever receives ciphertext.

## Checklist

### DIFF
--- a/extensions/p00f/CHANGELOG.md
+++ b/extensions/p00f/CHANGELOG.md
@@ -1,5 +1,9 @@
 # p00f Changelog
 
+## [Fix Web Crypto crash on every create] - {PR_MERGE_DATE}
+
+- Fixed every command failing with `ReferenceError: crypto is not defined`. p00f encrypts on your device using the Web Crypto API, and Raycast's extension host runs commands in a scope with no `crypto` global, so encryption threw before any Poof could be created. The extension now installs Node's Web Crypto implementation itself. No Poof content was ever at risk: the failure happened before anything left the machine.
+
 ## [Initial Version] - 2026-07-27
 
 - **Create Poof**: full form for text or one file with TTL, Reveal budget, PIN, secret kind, masked URL, reveal-anchored TTL, viewer-delete, reveal captcha, and countdown options.
--- a/extensions/p00f/CHANGELOG.md
+++ b/extensions/p00f/CHANGELOG.md
@@ -1,6 +1,6 @@
 # p00f Changelog
 
-## [Fix Web Crypto crash on every create] - {PR_MERGE_DATE}
+## [Fix Web Crypto crash on every create] - 2026-07-27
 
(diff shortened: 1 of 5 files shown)


0xdhrv (approved):
Looks good to me, approved ✅

PR #29785, opened 2026-07-27
Update quarantine-manager extension

nurkamol:
## Description

Fixes an out-of-memory crash reported through Raycast's error reporting for the `manage-quarantine` command (`Worker terminated due to reaching memory limit: JS heap out of memory`), plus two data-correctness bugs found while tracing it.

**The crash.** An app installed from a DMG or zip carries `com.apple.quarantine` on *every* file inside the bundle, not just the bundle root — `Final Cut Pro.app` on my machine has 38,114 of them. The scan then did two unbounded things: it read each match back with its own `xattr -p` child process (~2ms each, so ~75s of blocking `spawnSync` calls), and it rendered one `List.Item` per file, each with an 11-action panel and an eagerly-constructed `Detail` subtree. That is what exhausted the worker heap.

- Paths and values now come from a single `xattr -p com.apple.quarantine -r` pass instead of a listing pass plus one process per match — **2.5s and one process**, down from ~75s and 38k processes. Line parsing is anchored on the quarantine value's shape so paths containing `": "` still parse correctly.
- Listings are capped at 500 items (per directory and across the whole selection). The true count is reported in a notice row, and a new **Remove Quarantine Recursively** action (`⌘⇧R`) clears everything including the unlisted items — without it, per-item removal would silently leave them quarantined.
- Quarantine data is parsed once per target instead of inside the sort comparator (which built a `Date` and a locale-formatted string per comparison) and again per render.
- `spawnSync` failures are now surfaced. `ENOBUFS`/`ETIMEDOUT` produce a visible "scan did not complete" state; previously the truncated output was reported as "No quarantine found", i.e. a failed scan looked like a clean file. Exit status is deliberately still ignored, since `xattr -p` exits non-zero whenever any path simply lacks the attribute.
- The tree is only walked for a scope count when nothing was found, since that count is displayed solely in the empty state.

**Download dates were 31 years in the future.** The quarantine timestamp is hex seconds since the Unix epoch, but it was being read as Mac absolute time (adding the 2001-01-01 offset). Verified against real records: Free Download Manager's stamp on Final Cut Pro reads `2026-06-28` correctly and `2057-06-28` under the old code. Sort order was unaffected — the offset was uniform — so only the displayed dates change.

**Escaped source names.** Some downloaders write `\xNN` escapes into the quarantine record (`Free\x20Download\x20Manager`). These are now decoded for display. Decoding happens per field *after* splitting on `;`, so a decoded `\x3b` cannot fabricate an extra field, and non-printable results are left escaped.

### Measured on `/Applications/Final Cut Pro.app` (38,114 quarantined files)

| | before | after |
|---|---|---|
| scan | ~75s, 38,114 child processes | 2.5s, 1 process |
| rows rendered | 38,114 | 500 + notice row |
| result | JS heap out of memory | 17 MB heap |

## Checklist

### DIFF
--- a/extensions/quarantine-manager/.gitignore
+++ b/extensions/quarantine-manager/.gitignore
@@ -3,6 +3,9 @@
 # dependencies
 /node_modules
 
+# build output
+dist/
+
 # Raycast specific files
 raycast-env.d.ts
 .raycast-swift-build
@@ -12,17 +15,5 @@ compiled_raycast_rust
 
 # misc
 .DS_Store
-
-node_modules/
-dist/
-.DS_Store
-
 .claude/
 CLAUDE.md
-
-# Raycast specific files
-.raycast-swift-build
-.swiftpm
-compiled_raycast_swift
-compiled_raycast_rust
-raycast-env.d.ts
--- a/extensions/quarantine-manager/CHANGELOG.md
+++ b/extensions/quarantine-manager/CHANGELOG.md
@@ -1,5 +1,15 @@
 # Quarantine Manager
 
+## [Faster scans and correct download dates] - {PR_MERGE_DATE}
+
+### Fixed
+
+- Scanning an app installed from a DMG or zip no longer runs out of memory. Such apps carry the quarantine flag on **every** file inside them (Final Cut Pro: over 38,000), which the scan previously read one file at a time. It now reads them in a single pass — about 2.5 seconds instead of well over a minute.
+- Large scans list the first 500 quarantined items and report the full count, with **Remove Quarantine Recursively** (`⌘⇧R`) to clear everything — including the items too numerous to list individually.
+- Download dates were shown 31 years in the future (2057 instead of 2026); the quarantine timestamp was being read as Mac absolute time rather than Unix epoch seconds.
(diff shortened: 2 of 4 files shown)


nurkamol:
@0xdhrv when you get a chance, would you mind taking a look at this one?

It fixes an out-of-memory crash in `manage-quarantine` that surfaced through Raycast's error reporting (`Worker terminated due to reaching memory limit`). Apps installed from a DMG or zip carry `com.apple.quarantine` on every file inside the bundle — `Final Cut Pro.app` has 38,114 — and the scan was spawning one `xattr` process per match and rendering one row per file. It now reads paths and values in a single pass (2.5s instead of ~75s) and caps the list at 500 rows, with a recursive removal action covering everything beyond the cap.

It also corrects two data bugs found along the way: the quarantine timestamp was read as Mac absolute time rather than Unix epoch seconds, which put every download date 31 years in the future, and escaped source names (`Free\x20Download\x20Manager`) are now decoded for display.

Greptile's review is addressed — it flagged a cap-bypass on directly selected roots and files, both fixed, now 5/5 with no open comments.

Still marked as a draft pending a final pass over the distribution build locally; I'll flip it to ready for review once that's done.

nurkamol:
Checklist is complete — this is now marked ready for review.

0xdhrv (approved):
Looks good to me, approved ✅

PR #29786, opened 2026-07-27
Thock: add Preview Soundpack command

simply-sunny:
## Description

Adds a **Preview Soundpack** command to the Thock extension: browse the Thock soundpack registry and hear any pack played back as a few keystrokes, without installing it.

- Fetches the registry manifest live, so the list stays current instead of being hardcoded.
- Sectioned by category (Keyboard / Mouse), searchable by name, brand, or author.
- Packs already on disk are tagged `Installed` and get a secondary **Set as Active** action.
- **Preview** shells out to `thock-cli preview <uuid>`, with an animated toast while it plays.

## Dependency — please don't merge yet

The `preview` subcommand this calls is not in a released `thock-cli` yet. It's proposed upstream in kamillobinski/thock#135 (issue kamillobinski/thock#134). Keeping this as a draft until that lands and ships; happy to rebase or adjust naming if the upstream command changes.

## Checklist

`npx ray build`, `npx ray lint`, and `npx tsc --noEmit` are all clean.

### DIFF
--- a/extensions/thock/CHANGELOG.md
+++ b/extensions/thock/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Thock Changelog
 
+## [Preview Soundpack] - {PR_MERGE_DATE}
+
+- Added a `Preview Soundpack` command that lists the soundpack registry and plays a few keystrokes from any pack without installing it
+- Installed packs are tagged and can be set as the active soundpack from the same list
+
 ## [Chore: Updated the README] - 2025-05-12
 
 ## [Initial Version] - 2025-05-02
--- a/extensions/thock/README.md
+++ b/extensions/thock/README.md
@@ -2,6 +2,12 @@
 
 Control Thock with Raycast
 
+## Commands
+
+- **Toggle Thock** — enable or disable Thock.
+- **Set Switch Set** — pick the active switch set.
+- **Preview Soundpack** — browse the soundpack registry and hear any pack before installing it. Packs you already have are tagged `Installed` and can be set as active from the same list.
+
 Note: 
 thock-cli is bundled automatically when you install Thock via Homebrew:
 `brew install thock`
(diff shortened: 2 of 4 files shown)


SonicKarnati:
Thanks — two of the three were real and are fixed in 5398733; the third is a false positive.

**1. Remote ids reaching the shell — valid, fixed.** Genuine bug, and my fault: quoting an untrusted value inside a command string isn't protection. Switched to `execFile` with an argv array, so no shell is involved at all.

I checked that this was actually exploitable rather than theoretical, using an id of `x"; touch pwned.txt; echo "`:

```
after exec()      -> pwned.txt exists: true
after execFile()  -> pwned.txt exists: false
```

**3. CLI lookup ignores PATH — valid, fixed.** Now tries the Apple Silicon and Intel Homebrew prefixes and falls back to `thock-cli` on `$PATH`, with the Homebrew prefixes appended to `PATH` for that lookup. Worth noting the other two commands in this extension still hardcode `/opt/homebrew/bin` — I left them alone to keep this PR scoped, but they have the same Intel-Mac problem and could use the same treatment.

**2. Installed packs use wrong path — false positive, no change.** `SOUNDPACKS_DIR` is already Thock's application-support directory. The review doesn't say which other base it expected, and all three sources agree with what I used:

- app, `CustomSoundpackHelper.getCustomSoundpackDirectory()` → `Library/Application Support/Thock/Soundpacks`
- app, `SoundpackRegistryService.install()` → unpacks to that directory, named `<uuid>`
- `thock-cli` → `SOUNDPACKS_DIR="$HOME/Library/Application Support/Thock/Soundpacks"`

And on a machine with packs installed:

```
/Users/…/Library/Application Support/Thock/Soundpacks/186CA33E-…/config.json
/Users/…/Library/Application Support/Thock/Soundpacks/23D8CF7B-…/config.json
```

which is exactly the `path.join(SOUNDPACKS_DIR, id, "config.json")` this checks. Happy to revisit if there's a specific alternate path in mind.

`npx ray build`, `npx ray lint`, and `npx tsc --noEmit` are still clean.

PR #29798, opened 2026-07-28
Update Claude extension

chrismessina:
This is the second half of the split you asked for. [#30287](https://github.com/raycast/extensions/pull/30287) — the API-compatibility fixes — merged on 2026-08-19; everything here is the structural work that was deliberately held back.

**Users get:** one Recents command instead of three overlapping ones, presets they can export and edit as YAML, and a way to get their data out.

## Answering your three questions

**`History` and `Saved Answers` — removed, not kept as thin wrappers.** I went with your second option rather than your first, deliberately. `history` opening Recents pre-filtered would have been nearly free, but `saved` had no filter to open Recents *with*: the Status dropdown is Active/Archived/All, and pinning is not a status — so keeping `saved` meant inventing a fourth filter dimension to preserve a command whose concept had genuinely dissolved into pinning. Keeping one and dropping the other seemed worse than being consistent. The removal is now the **first line of the changelog**, in user-facing language, and it names the concrete consequence (a bound hotkey or Quicklink stops working). If you'd still rather have the wrappers, say so and I'll add both.

**Agent export — dropped, importer kept.** `exportPresetsAsAgents`, `modelToAgent`, and `toAgentModelId` are gone. `importAgentsAsPresets` stays, and the module header now says plainly that only the import direction ships and why. You were right that unreachable unverified code invites someone to wire it up later.

**`assets/icon@dark.png` — intentional.** The extension icon was redrawn in this PR and reads correctly on both backgrounds, so the separate dark variant is no longer needed. `package.json` declares `icon.png` alone.

## How this was rebased onto #30287 — worth knowing before you read the diff

This branch predates #30287's review, so a plain rebase would have **reverted** every fix that came out of it. Twelve files were touched by both. Four take the merged version wholesale; the rest were merged hunk by hunk, and `src/utils/models.ts` was composed by hand because it was added independently on both sides and has no merge base at all.

Two of PR 1's hunks appear here as *rejected*: this branch still subtracted `max_tokens` from `max_input_tokens` and still estimated tokens by character count, both of which were corrected in review. A check enforces that neither comes back, alongside the non-streaming retry, the byte-based estimate, the request-generation claim, the mounted guard, and the stream abort listener.

## Commands: five → three

**Ask Question · Recents · Presets**

`Conversations`, `History`, and `Saved Answers` were three views over one dataset. They're now **Recents**, with a Status filter and per-conversation Pin, Archive, Rename, Delete. Existing data merges automatically on first launch; the migration is idempotent, verifies its write by re-read before deleting anything, and copies aside anything it cannot parse rather than dropping it.

01-ask
02-recents
03-regenerate

`Models` → **Presets** — the command `name` is deliberately unchanged, so existing hotkeys keep working; only the title moved. `Save Answer` → **Pin Conversation**, matching Raycast's Clipboard History convention (`Icon.Tack`, ⌘.).

## Storage

All collections moved to one store where mutations are read-modify-write against freshly read storage, closing the cross-command races. Two limits are documented rather than papered over: `LocalStorage` has no compare-and-swap, so genuinely simultaneous in-flight writes remain last-write-wins, and the migration has a one-await window where a concurrent delete can lose. Both are stated in the source at the point they apply.

The legacy keys are migrated and then deleted. That is the only automatic destructive step, and it is gated on a branded `VerifiedPayload` that can only be minted by re-reading the key and finding what was written — so "verify, then delete" is checked by the compiler rather than by a comment.

## Added

- **Export History to JSON** (⌘⇧E in Recents) — all conversations, regardless of filter.
- **Export Stored Data to JSON**, offered from the empty state only. The history export reads the in-memory list, so it writes nothing in exactly the case a user needs it; this one writes raw storage, including anything an upgrade set aside unparsed.
- **Preset import/export as YAML** (⌘⇧E / ⌘⇧I in Presets), with name-collision handling (skip or replace, reported as a tally).
- **Regenerate an answer** with a different preset or model, appended rather than replaced.
- Each answer shows which model or preset produced it.

## Dependencies

`@raycast/utils` 1 → 2, ESLint 9 flat config, TypeScript 5.9, Prettier 3, React 19 types. Dropped unused `raycast` and `cross-fetch`. `@types/react` is pinned to exactly `19.0.10` to match `@raycast/api`'s own dependency — `^19.0.10` resolves a second copy of the types and `tsc` then fails on `ReactNode`.

## Breaking

- **`History` and `Saved Answers` are removed** as commands. Content lives in Recents; bound hotkeys need re-pointing. First line of the changelog.
- **Delete is now complete.** It previously removed the conversation but left the same answers in History and Saved Answers — invisible, still on disk. The export actions are the counterpart.

## Also in this PR

`AGENTS.md` and `CONCEPTS.md` at the extension root. Not universal practice, and I know that — this one is the Claude extension, most of the recent work was done with an agent, and the migration has invariants that are easy to remove without anything failing. `AGENTS.md` is written for whoever has to fix an upgrade that went wrong: what the storage keys are, what retirement does and does not guarantee, and what to do when Recents comes up empty.


pernielsentikaer (changes requested):
Hi @chrismessina 👋

**I'd like to take you up on splitting it.** Right now this mixes live API fixes, a 5→3 command consolidation, a new storage layer with a 774-line migration, and four new features. If the `temperature` rejection is a hard 400 on current models and the non-streaming path fails for everyone with streaming disabled, that's breakage in the Store today and shouldn't wait behind a structural review.

So: PR 1 with the API-compatibility fixes only (shared request builder, `max_tokens` from the models API, model list pagination, the 10-minute clamp, and the context-window bounding). We can move on that quickly. Then the consolidation, storage layer and new features as a second PR, reviewed properly.

**On removing `History` and `Saved Answers`.** This extension has ~31,900 installs, and both are published commands, so anyone with a hotkey or Quicklink bound to either loses it with no signal beyond the changelog. I noticed you kept `name: model` while retitling it to Presets, which avoids exactly this for that command — so I suspect you've already weighed it.

Two options I'd be happy with: keep `history` and `saved` as thin commands that open Recents with the right filter preselected for a release or two before removing them, or go ahead with the removal and make it the first line of the changelog in language a user will understand. I lean towards the first. Your call as long as it's deliberate.

**Two small things:**

- `agentInterop.ts` exports `exportPresetsAsAgents` / `modelToAgent`, which you note are deliberately unwired because the mapping came from ray.so samples rather than a verified export. I'd drop the export half until it can be checked against a real one, and keep the importer. Shipping unreachable code invites someone to wire it up later without knowing it was unverified.
- `assets/icon@dark.png` is deleted and the description doesn't mention it. Intentional?

Everything else checks out: metadata swapped cleanly, README images all referenced, and you added yourself to `contributors` rather than changing `author`, which is right.

---

pernielsentikaer (commented):
Is this ready for another look, or still in draft? 🙂

PR #29801, opened 2026-07-28
Update gh-pr-tracker extension

chrismessina:
## Description

Bug fixes and error handling for **GitHub Pull Requests**, plus an opt-in faster fetching mode.

### Fixed — the Unread PR Alert badge could show a stale count

A background-launched menu bar command only commits its render once `isLoading` settles to `false`, and it gets a short execution window to do so. The command previously seeded itself from an **async** `LocalStorage` read, which could resolve after that window had closed — leaving the previously committed badge in place.

The badge payload now lives in a synchronous `Cache` that the command reads in a `useState` initializer, so its first render is already correct and already settled. The view command writes that cache *before* launching the menu bar, so the payload survives even when `launchCommand` throws (which it does when the menu bar command has not yet been activated for background refresh).

### Fixed — several states rendered as a blank window

The empty view was gated on `!isLoading && !demoMode`. "Mark All as Caught Up" triggers a revalidate, so `isLoading` flipped true, the empty view unmounted, and the command rendered an empty `List` with nothing in it. Same for demo mode once every demo PR was marked read.

Related: hiding every event type previously showed "All caught up!", which was untrue — the activity existed, it was filtered out. It now says so and offers Event Filters. Refresh is now the primary action on the empty state; previously the only action offered was Demo Mode.

### Fixed — error handling

- **Rate limit exhaustion now says what happened.** GitHub returns `403` (not `429`) when the hourly quota is used up, which read as a permissions problem and sent users to check token scopes. The message now names the quota, roughly when it resets, and which preferences reduce per-refresh cost.
- **Failure toasts carry a Copy Error action.**
- **Corrupt local data is detected and reset** instead of crashing on the next launch. Cached pull requests, seen state, and event filters are all shape-checked; previously a structurally valid but malformed payload passed and threw downstream.
- **Silent failures now log.** Seven `catch` blocks discarded their errors, including the one that made the badge bug undiagnosable.

### Added — Faster Fetching (experimental, off by default)

Fetches activity over the GraphQL API in two passes: cheap metadata (`number` + `updatedAt`) for many pull requests, then full activity only for those that may contain unread items.

Measured against a repository with 312 open pull requests:

| | REST | GraphQL |
| --- | --- | --- |
| Rate-limit cost | ~84 points | **7 points** |
| Metadata pass | — | 200 PRs / 1,002ms / 2 points |

It falls back to REST automatically on any error, including on GitHub Enterprise servers whose version predates the required fields. Retries transient gateway failures (500/502/503/504) with jittered backoff; never retries auth or rate-limit errors.

### Added — Verbose Logging (off by default)

Structured logging for troubleshooting, with automatic redaction of tokens and credentials — including those embedded in URLs, which the underlying logger does not scrub on its own.

### Changed — one-time effect on read state

Label changes and force pushes you had already read may reappear as unread once after updating. Their identity is now derived from the event itself rather than an API-assigned id, so read state stays stable across future updates and across both fetching modes. Reviews, comments, and commits keep their read state.

## Checklist

---

<sub>Review feedback addressed in c2fd783: a truncated GraphQL result could record a "fully seen" watermark from incomplete data, and the menu bar could leave a stale badge after the count reached zero.</sub>


chrismessina:

extensions/gh-pr-tracker/src/api-graphql.ts
```diff
@@ -0,0 +1,397 @@
...
+}
+
+function couldContainUnreadActivity(updatedAt: string, seen: SeenState | undefined): boolean {
+  // `lastSeen` is intentionally NOT used: a single-item action advances it while other activity
+  // remains unread. Only the new, full-PR watermark has the required semantics.
+  if (!seen?.fullySeenAt) return true;
+  const updatedAtMs = Date.parse(updatedAt);
+  const fullySeenAtMs = Date.parse(seen.fullySeenAt);
+  if (!Number.isFinite(updatedAtMs) || !Number.isFinite(fullySeenAtMs)) return true;
+  return updatedAtMs > fullySeenAtMs - UPDATED_AT_SAFETY_MARGIN_MS;
```
MiguelOlsen: Backfilled watermarks set `fullySeenAt = pr.updated_at` (`seen.ts applyFullySeenWatermarks`).
`updated_at` is monotonic, so on the next scan of an unchanged PR `updatedAt == fullySeenAt`, and `W > W − 60000` is always true → always re-fetched. The skip is unreachable for every `updated_at` -valued watermark.

That's exactly the population the backfill targets (seen-state predating the feature) so `skippedByWatermark` stays 0, 150/150 still fetched.

I would drop the margin
```suggestion
  return updatedAtMs > fullySeenAtMs;
```

extensions/gh-pr-tracker/CHANGELOG.md
```diff
@@ -4,6 +4,27 @@ All notable changes to the **GitHub Pull Requests** Raycast extension are docume
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
+## [1.1.2] - {PR_MERGE_DATE}
```
MiguelOlsen: nitpick: This are considerable changes with new significant features, I would argue this should be a minor and not a patch (1.2.0)

MiguelOlsen (commented):
This changes make a deliberate change from my original vision of the menu bar, I had it disappear when no new activity was unread, so the user would only see it in the menu bar when it is relevant (hence the notification bell icon chosen). I liked that idea but I would also not be opposed to keeping it on constantly, though that would require an icon change to have a clearer intent to the user, since an empty notification bell feels to invasive for a single extension in this scope


Overall great changes, thank you for your contribution 😁

chrismessina:
> This changes make a deliberate change from my original vision of the menu bar, I had it disappear when no new activity was unread, so the user would only see it in the menu bar when it is relevant (hence the notification bell icon chosen). I liked that idea but I would also not be opposed to keeping it on constantly, though that would require an icon change to have a clearer intent to the user, since an empty notification bell feels to invasive for a single extension in this scope
> 
> 
> Overall great changes, thank you for your contribution 😁

Appreciate that context! You're right, and that was an unintended change on my part — I was fixing a stale badge and didn't notice I'd also made the bell permanent. [Reverted](https://github.com/raycast/extensions/pull/29801/commits/0af437d4cba68e29d5fd93bfc79a65d127a50933) to your original behavior as the default.

One wrinkle worth flagging: plain `null` doesn't reliably clear a count Raycast has already committed, since it restores menu bar items from its database rather than re-executing the command. That's what produced the stale badge in the first place. So the transition from a non-zero count to zero now commits one empty render, and the *next* launch hides. The count clears, _then_ the icon disappears — rather than a "5" persisting after you've caught up.

Added a **Menu Bar Icon** preference for anyone who prefers it always visible, off by default. Agreed on your point about the icon — if always-on ever became the default it'd want a different glyph, but as an opt-in the bell seems fine.

`9368942`.

extensions/gh-pr-tracker/src/api-graphql.ts
chrismessina: Good catch — you're right, and it explains a field symptom I couldn't account for: `skippedByWatermark: 0` with 150/150 still fetched, run after run.

I didn't take the suggestion literally though, because there turn out to be two kinds of watermark and they need opposite treatment:

- **Backfilled** (`fullySeenAt = pr.updated_at`, set when a fetch finds nothing unseen) — this is on GitHub's clock, the *same* clock it's compared against. Equality means "unchanged since we last looked", so there's no lag to absorb and the margin makes the skip unreachable. Exactly your point.
- **Mark-as-read** (`fullySeenAt = Date.now()`, from `markPRSeen`/`markAllSeen`) — wall clock. I measured a standalone inline review comment preceding `updated_at` by 6–10s, so a PR marked read inside that window can have activity GitHub hasn't reflected yet. Dropping the margin here would silently hide it.

So watermarks now record which clock they came from, and the margin applies only to wall-clock ones. Entries written before the field existed default to wall-clock, keeping the margin for them.

Fixed in `89274c7`. Thanks — the unchanged-PR case now actually skips.

MiguelOlsen (approved):
LGTM, Great work!

chrismessina:
@0xdhrv want to take a look?

0xdhrv (approved):
Looks good to me, approved ✅

PR #29802, opened 2026-07-28
Update model-context-protocol-registry: UseMyContext description

usemycontext:
## Description

Removes the hard-coded tool count from the UseMyContext registry entry description (`8 tools for profile, ...` -> `Tools for profile, ...`).

The server's tool surface has grown since the original submission (13 tools today), so the listed `8` already understates it - and any number in the description goes stale again with the next addition. The capability summary (profile, file search and reads, cited answers from documents, exact table queries) stays as is.

One word changed in `entries.ts`, plus the changelog entry.

## Type of change

- [x] Existing extension update

## Checklist

- [x] The change is limited to the `model-context-protocol-registry` extension
- [x] CHANGELOG.md entry added with `{PR_MERGE_DATE}`

### DIFF
--- a/extensions/model-context-protocol-registry/CHANGELOG.md
+++ b/extensions/model-context-protocol-registry/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Model Context Protocol Registry Changelog
 
+## [Update UseMyContext description] - {PR_MERGE_DATE}
+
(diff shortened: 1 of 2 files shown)


0xdhrv (approved):
Looks good to me, approved ✅

PR #29808, opened 2026-07-28
Twenty: fix metadata API parsing for current Twenty versions

pmaluf:
## Description

The Twenty extension currently fails to load on up-to-date Twenty workspaces. Opening **Create Object Record** shows:

```
[ { "expected": "object", "code": "invalid_type", "path": [ "data" ],
    "message": "Invalid input: expected object, received array" } ]
```

Twenty changed the shape of its REST metadata API (the new envelope is gated behind the `IS_REST_METADATA_API_NEW_FORMAT_DIRECT` feature flag, which is now on for workspaces), and the extension's Zod schemas only accepted the old shape:

| | Extension expected | Current Twenty |
|---|---|---|
| `GET /metadata/objects` | `{ data: { objects: [...] } }` | `{ data: [...], pageInfo, totalCount }` |
| `GET /metadata/objects/:id` | `{ data: { object: {...} } }` | the object, with no envelope |
| `isCustom` on objects/fields | required | dropped in 2.12 |
| `description`, `icon`, `isActive`, `isSystem`, `isNullable` | required | optional, and omitted from the JSON when unset |
| pagination | none | id cursor, 60 per page by default |

Both envelopes are still in the wild, so this PR makes the extension accept either rather than switching to the new one — self-hosted instances on older versions keep working.

### Changes

- **`dataModelSchema.ts` / `recordFieldSchema.ts`** — the response envelopes are now unions over both shapes, normalised by `extractDataModels()` / `extractDataModelWithFields()`. Fields the API may omit are `nullish`.
- **`TwentySDK.ts`** — pages through `/metadata/objects` with `limit=200` and `starting_after` until `hasNextPage` is false (capped at 25 pages), so workspaces with more than 60 objects are listed in full. Filters now use `isSystem !== true` / `isActive !== false` so records aren't dropped when those flags are absent.
- **`TwentySDK.ts`** — failures now surface Twenty's own error message (e.g. `Missing authentication token`) instead of `statusText`, and a schema mismatch reports the failing path in prose rather than dumping the raw validation array into the UI, which is what produced the error above.
- **`create-object-record.tsx`** — with `isCustom` gone there is no way to tell standard and custom objects apart, so they render as a single "Objects" section; on servers that still send `isCustom` the original Standard/Custom split is unchanged. List keys switched from `randomUUID()` to the object id, which was remounting the whole list on every render.

No changes to `package.json`, commands, preferences, or assets.

## Screencast

No screenshot: this is a bug fix with no intentional UI change — before the fix the command renders only the error above, after it the object list renders as it always did (with the Standard/Custom sections merged when the server no longer reports `isCustom`).

Verified by:

- `npm run build` and `npx ray lint` both clean
- running the extension against a live Twenty workspace on the new API format — the object list loads and record creation works again
- parsing checks over both response envelopes, servers that send no `pageInfo`, payloads with the optional flags absent, and the exact `{ data: [...] }` payload from the report

## Checklist

### DIFF
--- a/extensions/twenty/CHANGELOG.md
+++ b/extensions/twenty/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Twenty Changelog
 
+## [Fix] - {PR_MERGE_DATE}
+
+- Fixed "Invalid input: expected object, received array" on newer Twenty versions, which changed the metadata API response shape
+- Objects beyond the first page are no longer silently dropped — the metadata API now paginates at 60 objects by default
+- Objects and fields no longer disappear when optional metadata flags are absent
+- API errors now show the message returned by Twenty instead of a raw validation dump
+
 ## [Fix Bug] - 2025-11-27
 
 - Fixed issue when adding API Key, it would not work
--- a/extensions/twenty/CHANGELOG.md
+++ b/extensions/twenty/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Twenty Changelog
 
-## [Fix] - {PR_MERGE_DATE}
+## [Fix] - 2026-07-28
 
 - Fixed "Invalid input: expected object, received array" on newer Twenty versions, which changed the metadata API response shape
 - Objects beyond the first page are no longer silently dropped — the metadata API now paginates at 60 objects by default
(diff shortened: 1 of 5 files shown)


0xdhrv (approved):
Looks good to me, approved ✅

PR #29810, opened 2026-07-28
Teleport: window behavior preference and per-connection environment

BBonnet22:
# Teleport: window behavior preference and per-connection color for databases

## Description

Two small quality-of-life improvements for database connections, both driven by parameters of the database client URL scheme, see the [TablePlus docs](https://docs.tableplus.com/gui-tools/manage-connections).

cc @lamberttraccard (extension author) for sign-off.

### Window Behavior preference

When opening a database connection, the extension launches the database client such as TablePlus via a `client://…` URL. Because the URL did not set the client's `windowMode` parameter, every connection opened in a **new standalone window** — annoying when a client window is already open, since you get a new window instead of a tab.

This adds a **Window Behavior** preference, a checkbox labelled "Open connections in a new window":

- **checked**, the default → new standalone window via `windowMode=isolated`, unchanged from today
- **unchecked** → tab in the current window via `windowMode=tabbed`

### Per-connection environment

The connection URL previously hard-coded `environment=production` and `statusColor=6D0000` for every database, so every connection showed the **Production** tag and the same red status bar — even a staging database.

This adds a **Setup Environment** action (`⌘⇧K`) to tag a database with one of TablePlus's built-in environments — Local, Testing, Development, Staging, Production. The chosen value is sent as the connection's `environment` parameter (so the correct tag shows in the status bar) along with a matching `statusColor`, and is stored per connection like the default-database setting. The environment also shows as a colored tag in the list.

Each environment's status-bar color is configurable through preferences (one hex field per environment, with the default shown as a placeholder). Connections with no environment set open with no tag, exactly as before this change.

## Type of change

- [x] New feature (non-breaking change which adds functionality)

## Checklist

- [x] `npm run lint` passes
- [x] Updated `CHANGELOG.md`

### DIFF
--- a/extensions/teleport/CHANGELOG.md
+++ b/extensions/teleport/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Teleport Changelog
 
+## [Window Behavior and Environment Colors] - {PR_MERGE_DATE}
+
+- Add a "Window Behavior" preference to open database connections in a tab of the current window instead of a new window; opening in a new window stays the default
+- Add a "Setup Environment" action to tag a database with one of TablePlus's environments (Local, Testing, Development, Staging, Production); the tag drives the connection's `environment` and status-bar color instead of the previously hard-coded Production, and each environment's color is configurable in preferences. Untagged connections open with no environment tag
+
 ## [Fix Login] - 2026-07-21
 
 - Force OTP MFA mode so login works on clusters that prefer WebAuthn
--- a/extensions/teleport/CHANGELOG.md
+++ b/extensions/teleport/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## [Window Behavior and Environment Colors] - {PR_MERGE_DATE}
 
 - Add a "Window Behavior" preference to open database connections in a tab of the current window instead of a new window; opening in a new window stays the default
-- Add a "Setup Environment" action to tag a database with one of TablePlus's environments (Local, Testing, Development, Staging, Production); the tag drives the connection's `environment` and status-bar color instead of the previously hard-coded Production, and each environment's color is configurable in preferences. Untagged connections open with no environment tag
+- Add a "Setup Environment" action to tag a database with one of TablePlus's environments (Local, Testing, Development, Staging, Production); the tag drives the connection's `environment` and status-bar color instead of the previously hard-coded Production, and each environment's color is configurable in preferences. Untagged connections open with no environment tag and a red status bar by default
 
 ## [Fix Login] - 2026-07-21
 
--- a/extensions/teleport/CHANGELOG.md
+++ b/extensions/teleport/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Teleport Changelog
 
-## [Window Behavior and Environment Colors] - {PR_MERGE_DATE}
+## [Window Behavior and Environment Colors] - 2026-07-30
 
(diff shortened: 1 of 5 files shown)


lamberttraccard (approved):
Extension author sign-off: this looks good to me, approving.

Nice work: the `isolated` default keeps the current window behavior untouched, `hexColor()` validates and falls back instead of producing a broken URL, and `SetupEnvironmentForm` follows the exact conventions of the existing `SetupDefaultDatabaseForm`. Build and lint are green.

Three non-blocking notes, none of which need to hold the merge:

1. `connect-database.sh` sends `environment=…`, but the [TablePlus URL scheme docs](https://docs.tableplus.com/gui-tools/manage-connections) only document `env=…`. The `environment` key does work in practice (the previously hard-coded `environment=production` did show the Production tag), but it is an undocumented alias that could stop working in a future TablePlus release. Sending both keys is cheap insurance:

   ```sh
   QUERY="${QUERY}&env=${ENVIRONMENT}&environment=${ENVIRONMENT}"
   ```

2. `hexColor()` accepts 3-digit hex, so `#ABC` passes validation and is forwarded as `statusColor=ABC`, which TablePlus most likely ignores: the user gets no color and no explanation. Either expand the short form to 6 digits or restrict the regex to `[0-9A-F]{6}`.

3. Behavior change worth being explicit about: today every connection shows the red Production status bar, and after this change connections show no tag until they are tagged one by one. That is the right call (a staging database labelled "production" was worse), and the CHANGELOG covers it, so no change requested.

On the Greptile comment about environment state not being scoped to the active Teleport cluster: that is the same keying the pre-existing `database-defaults` store uses, so it is not a regression from this PR. Worth fixing for both stores in a follow-up rather than here.

Last small thing: `contributors` takes Raycast store handles, so please double-check that `bbonnet` matches your Raycast account and not just your GitHub username.

0xdhrv (commented):
Hey @BBonnet22 👋

Thanks for your contribution!

Can you please check the feedback provided by @lamberttraccard ?

Also, thanks for the review @lamberttraccard !

BBonnet22:
Thanks @0xdhrv, and thanks a lot @lamberttraccard for the detailed sign-off and notes 🙏 Addressed the feedback in the latest commit:

`env` vs `environment`:  I verified that `env=` alone shows the tag in TablePlus, so I switched to the documented `env=` parameter and dropped the undocumented `environment=` alias entirely (rather than sending both).

3-digit hex: hexColor() now restricts to 6-digit hex (/^[0-9A-F]{6}$/); anything shorter/invalid falls back to the environment's default color instead of forwarding a value TablePlus ignores.

Behavior change (no tag by default) : kept as intended and documented in the CHANGELOG. One small refinement since your review: an untagged connection now opens with a red status bar by default (still no tag label), so it stays visually flagged rather than showing the client's default color.

Cluster scoping: the new database-environments store is now keyed by active cluster + database name (${cluster}::${name}), so same-named databases across clusters no longer share a tag. As you noted, the pre-existing database-defaults store keeps its name-only keying,  it has live user data, so it's better handled with a migration in a dedicated follow-up.

For the contributors handle, yes bbonnet is my Raycast store account, not just my GitHub username.

Also updated the default environment colors with (Local 006633, Testing 666666, Development 003366, Staging 996633, Production 660000); each is overridable in preferences.

0xdhrv (approved):
Looks good to me, approved ✅

PR #29815, opened 2026-07-28
Add expiration-reminder extension

joseacta:
## Checklist


joseacta:
Thanks for the review. Pushed f41f030 addressing all three findings.

**1. "Confidential client breaks PKCE" -> false positive, no code change**

The OAuth client is correct as written. The `client_id` it ships is registered as a **public** client on our side, and the secretless `authorization_code` exchange succeeds in production — verified against a live tenant with an instrumented build, which logged a single token exchange sending `code_verifier` and no `client_secret`, returning HTTP 200.

The "confidential client" claim came from a stale internal QA note that I had accidentally included in the submission. That note predated the server-side change making the client public and was never updated. It was wrong, and it was never part of the extension.

Root cause fixed: **the internal docs are removed from this PR entirely.** A `docs/` folder with our PRD, QA plan and support playbook, plus a local `scripts/make-icon.js`, were included by mistake, none of it belongs in this repo. That's 775 lines gone.

**2. Preference type duplicates manifest -> valid, fixed**

Dropped the hand-written `ExtensionPreferences` interface for the generated `Preferences` global.

This also needed `raycast-env.d.ts` added to `tsconfig.json`'s `include`, without it the generated types aren't in the program at all, which is why the interface existed in the first place. That now matches the convention used elsewhere in this repo (e.g. `extensions/github`, `extensions/brew`). The generated type is a strict improvement: literal unions (`"7" | "30" | "60" | "90"`) rather than bare `string`.

**3. ESLint config skips `defineConfig` -> valid, fixed**

Applied as suggested, verified against eslint 9.39.5.

`ray lint` and `ray build` are green, and `tsc --noEmit` passes against the regenerated `raycast-env.d.ts`.

---

One note for whoever tests this: every command requires an Expiration Reminder account, so the extension can't be exercised without a login. Happy to provide test-tenant credentials, let me know the preferred private channel and I'll send them over.

joseacta:
@raycastbot do not close. Waiting on @0xdhrv for approval

0xdhrv (commented):
Hey @joseacta 👋

- Store extensions cannot include [external analytics](https://developers.raycast.com/basics/prepare-an-extension-for-store#analytics). `src/lib/telemetry.ts` is a no-network debug seam today, but the listing still ships a Telemetry preference, a changelog line, and comments about pointing `send` at PostHog. Could you remove the preference, the `track()` call sites, and that changelog entry?

- `metadata/expiration-reminder-image-1.png` is Raycast root search, including an unrelated Raycast AI command. Store screenshots should show the extension itself. Could you recapture that slot as a real command (Show About-To-Expire Items would fill the gap), and recapture `expiration-reminder-image-4.png` without records like `Postman Test Item2`? Please also drop `metadata/README.md`; it still says the captures are not committed.

- Command titles should avoid articles, per the [naming guidelines](https://developers.raycast.com/basics/prepare-an-extension-for-store#extensions-and-commands-naming). Could you rename `Create an Expiration` → `Create Expiration`, `Create a Contact` → `Create Contact`, and `Show a Contact's Expirations` → `Show Contact Expirations`? Recapture any screenshot that still shows the old titles.

- `platforms` is unset, so this would install on Windows. The source is HTTP + OAuth with no OS-specific APIs. Could you set `"platforms": ["macOS", "Windows"]` if you have verified it on Windows, or `"platforms": ["macOS"]` until that is tested?

- In `src/oauth/client.ts`, `runRefresh()` catches every refresh failure and calls `authorize()`. An offline or 5xx refresh currently sends the user back to the browser. Could you keep the stored tokens on transient failures and only start a new consent flow when the refresh token is actually rejected?

- The Sign Out confirm in `src/components/actions.tsx` says the server-side session stays valid, but `signOut()` POSTs `/oauth/revoke`. Could you update that copy to match the revoke (and Title Case the action as `Sign Out`)?

- `ExpirationDetail` binds both Copy Name and Copy Expiration Date to ⌘C (`Keyboard.Shortcut.Common.Copy`). The same clash is on Copy Email / Copy Phone in `ContactSearchList` when a phone number is present. Could you give the secondary copy actions a different shortcut?

- Could you change the changelog heading to `## [Initial Release] - {PR_MERGE_DATE}`, and align `.gitignore` with the current Raycast template (`/node_modules`, `raycast-env.d.ts`, `.swiftpm`, `compiled_raycast_swift`, `compiled_raycast_rust`)?

---

0xdhrv:
Hey @joseacta 👋

One more on the Store screenshots: the command footer still shows the local development icon (the green badge next to the command name on `metadata/expiration-reminder-image-2.png`, `expiration-reminder-image-3.png`, and `expiration-reminder-image-4.png`). Could you recapture those with Raycast Window Capture and Save to Metadata so that badge is gone?

On `platforms`: if it is left unset, the listing is published as macOS. As a suggestion, could you still set `"platforms": ["macOS"]` (or `["macOS", "Windows"]` if you have verified Windows) so the listing is explicit?

PR #29816, opened 2026-07-28
Update tmux-sessioner extension

mattiacolombomc:
## Description

This PR extends tmux-sessioner with the ability to open a tmux session in a **new terminal tab or window**, project-bootstrap support in Create New Session, and session-management quality-of-life features.

> [!IMPORTANT]
> **Why these additions — the driving workflow**: pairing each AI coding-agent session (Claude Code, in my case) with a dedicated tmux session and project folder. With the new preferences set, a single Raycast command (`Create New Session myproject` → `⏎`) creates the folder, starts a persistent tmux session inside it, launches the agent via the startup command, and opens everything in a new terminal tab. Because tmux sessions survive terminal restarts (and reboots, with resurrect/continuum setups), the *Manage Tmux Session* list then doubles as a browsable, re-attachable history of past project sessions — which is also why the startup command is typed into the shell rather than run as the session command: the session must outlive the agent process. Everything is generic and preference-driven, so the same mechanism works for dev servers, editors, or any other per-project command.

**The problem**: switching to a session currently relies on `tmux switch-client`, which requires an already-attached tmux client. To open a session in a fresh tab you had to open a tab manually, run `tmux` (creating a throwaway numeric session), and then switch — polluting the session list with junk sessions like `34`, `35`, …

**The solution**: new *Open in New Tab* / *Open in New Window* actions that launch the configured terminal with `tmux attach-session -t '=<name>'` via per-terminal mechanisms:

| Terminal | New Tab | New Window | Mechanism |
|---|---|---|---|
| iTerm2 | ✅ | ✅ | AppleScript (`create tab/window with default profile` + `write text`) |
| Ghostty ≥ 1.3 | ✅ | ✅ | AppleScript (`new tab/window with configuration {initial input:…}`) |
| WezTerm | ✅ | ✅ | `wezterm cli spawn` (+ `--new-window`), `open -na` fallback |
| kitty | ✅ (remote control enabled) | ✅ | `kitten @ launch --type=tab` via socket / `open -na kitty` |
| Terminal.app | — | ✅ | AppleScript `do script` |
| Alacritty | — (no tabs) | ✅ | `open -na Alacritty --args -e` |
| Warp | — | — | Not scriptable; hint toast to use Switch |

Actions are hidden when the configured terminal doesn't support them; runtime failures (e.g. missing Automation permission, kitty remote control off) show explanatory failure toasts.

## Changes

- **Open in New Tab (`⌘T`) / Open in New Window (`⇧⌘T`)** actions on each session (new `src/utils/terminalLaunchUtils.ts` dispatch module)
- **Create and Open in New Tab/Window** secondary submit action in *Create New Session*, and an *Open in Terminal After Create* preference that makes it the default submit action
- **Project bootstrap in Create New Session**: optional *Create New Folder* mode that creates `<default directory>/<name>` for the session (with a *Create New Folder by Default* preference), an optional *Startup Command* typed into the fresh session via `tmux send-keys` (prefilled from the new *New Session Startup Command* preference — e.g. `npm run dev` or an AI coding agent CLI), and a session-name argument so `Create New Session myproject` from root search prefills the form. Typing the command into the shell (instead of making it the session command) keeps the session alive after the command exits.
- Session list now shows **window count, attached indicator, and last activity**, plus a proper empty state when no tmux server is running
- **Kill Multiple Sessions (`⇧⌘X`)**: checklist view with toggle (`⏎`), range selection (`⇧⏎`), Select All / Select All Numeric / Deselect All, and kill with confirmation (`⌘⏎`)
- **Kill All Numeric Sessions (`⌥⇧⌘X`)** to clean up throwaway sessions in one action; *Delete This Session* now asks for confirmation
- **Fix**: session names, directories and window names are now shell-quoted everywhere (names with spaces/quotes previously broke commands; the old `replace(" ", "\\ ")` only escaped the first space) and tmux targets use exact matching (`=name`) to avoid prefix-match accidents
- **Fix**: opening a session in the terminal right after creating it now awaits the full open chain before `popToRoot`, so the command process isn't torn down while the AppleScript is still pending
- Bumped `@raycast/api` to latest (with `react` 19 to match its types)

## Notes for reviewers

- macOS-only (unchanged)
- No new dependencies or commands; three new optional preferences (startup command, create-folder default, open-after-create) and one optional command argument — all defaults preserve existing behavior exactly
- AppleScript runs via `execFile("osascript", ["-e", …])` (no shell layer); all shell interpolations go through a single-quote escaping helper (`src/utils/shellUtils.ts`)
- Attaching an already-attached session intentionally creates a mirrored client (plain `attach-session`, no `-d`)
- Tested end-to-end on iTerm2 in Raycast (tab + window, no-window case, names with spaces/quotes, deleted-session race) and the full bootstrap flow (folder creation, startup command execution in the right directory, session survival after command exit, open-after-create); other terminals follow their documented CLIs/dictionaries (Ghostty commands verified against `Ghostty.sdef`) and degrade to explanatory toasts on failure

## Screenshots

Existing Store metadata is unchanged; happy to add updated screenshots of the new actions/accessories if preferred.

## Checklist


pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #29818, opened 2026-07-28
Add bluetooth extension

shivajreddy:
## Description

A Windows-only Bluetooth extension with two commands:

- **Bluetooth Devices** — lists paired devices grouped into Connected and Available, with connect and disconnect actions
- **Toggle Bluetooth** — turns the radio on or off

How it works:

- **Listing** goes through WinRT (`Windows.Devices.Enumeration.DeviceInformation`, plus `BluetoothDevice`/`BluetoothLEDevice.ConnectionStatus`) so it covers Classic and LE devices alike. A dual-mode device that appears on both transports is merged into a single row.
- **The radio toggle** uses `Windows.Devices.Radios.Radio`.
- **Connect/disconnect** uses the Win32 `BluetoothSetServiceState` API. That API is Classic-only, so LE peripherals such as mice and keyboards are listed with a _Managed by Windows_ label and deliberately offer no connect action, rather than an action that would fail. Disabling a service also uninstalls it, so connecting re-enables a known profile list instead of the (by then empty) installed set. Each service change blocks roughly three seconds inside the driver, so they are issued in parallel and the wait does not grow with the number of profiles.

Everything runs through the bundled `assets/bluetooth.ps1`, invoked with `execFile("powershell.exe", ...)` using `-NoProfile -NonInteractive -ExecutionPolicy Bypass` and `windowsHide`. The script writes a single JSON object to stdout and reports failures as `{ "error": "..." }`. No administrator rights, no network access, and no dependencies beyond `@raycast/api` and `@raycast/utils`.

Scanning, pairing and removing devices are intentionally out of scope for this first version.

## Screencast

Screenshots are in `metadata/`. Happy to add a screencast if that would help the review.

## Checklist


0xdhrv (commented):
Hi @shivajreddy 👋

Thanks for the contribution 🔥

---

PR #29819, opened 2026-07-29
Update raycast-store-updates extension

chrismessina:
## Description

Fixes a crash reported on Windows, substantially reduces this extension's GitHub API usage, and improves the menu-bar command added in the last release.

### The Windows crash

`TypeError: e is not iterable` (`view-store-updates`, Raycast 0.70.0.0, Windows 10.0.26200).

The extension reads merged PRs from the GitHub API to detect updates and removals. The response was cast to an array and iterated without a runtime check — but GitHub does not always answer an array-returning endpoint with an array. On rate limits and auth failures it returns `{message, documentation_url}`, and on Windows requests traverse a system proxy that can return an HTML error page or an empty body **with a 200 status**.

Two guards were needed, because the failure happens in two different places:

- `response.json()` throws `SyntaxError` on an HTML or empty body *before* any shape check can run, so the parse itself is now wrapped.
- A successfully-parsed non-array payload is rejected explicitly rather than cast.

Every fetch that feeds an iteration (PR list, PR files, PR labels, the store feed) now routes through the same guard.

### GitHub API usage

The extension is usable with no token (60 requests/hour, per IP, shared across both commands). Resolving *which extension a merged PR belongs to* previously cost one `/pulls/{n}/files` request per unresolved PR — measured at up to **29 billed requests in a single scan**, so two scans could exhaust the hourly budget and lock the user out.

- Extension slugs now resolve from the `extension:` label already present in the PR list response, at no extra request.
- The `/files` fallback is capped per scan (5 unauthenticated, 50 with a token). A PR that exceeds the cap keeps its title-derived slug — the previous behaviour.
- A slug is only *adopted* when the PR's own label asserts it. Previously a branch name that happened to match another extension could produce that extension's store URL and changelog.

Related rate-limit fixes:

- Only treat a `403` as a rate limit when GitHub reports `X-RateLimit-Remaining: 0`. A 403 from a proxy or VPN carries no rate-limit headers and was starting an hour-long cooldown GitHub never asked for.
- Without a reset header the cooldown is the ordinary 5 minutes rather than a fabricated hour, and a stored cooldown beyond GitHub's one-hour window self-heals.

### Menu bar

menubar

- Shows each extension's own icon, with rounded corners in both the menu bar and the main list.
- **Hold ⌥** on an item to open that extension's changelog *inside Raycast* instead of its store page (via `MenuBarExtra.Item`'s `alternate`, macOS Sonoma+; falls back to opening the changelog on GitHub if the view command is disabled).
- New preference for whether the menu bar counts all updates or only updates for extensions you have installed.
- On first run the badge now reflects the last 24h of activity. Previously it seeded its watermark to *now*, so a fresh install showed an empty badge until the next extension happened to ship — correct, but indistinguishable from broken.
- Fixed "Mark All as Seen" advancing the watermark past items that were never shown.

### Other fixes

- Only report an extension as removed on a definitive 404, so a transient server error no longer marks a live extension as deleted.
- Fixed duplicate entries when two removal PRs deleted the same extension (the dedup check released across an `await` inside a concurrent map).
- Fixed the macOS platform icon being nearly invisible in dark mode (was tinted 80% black; now theme-aware).
- Category tags now carry an icon, matching the Store's own category styling.
- Keep previously loaded updates on screen when a refresh fails, instead of clearing the list.
- Failure toasts carry a "Copy Error" action.

### Experimental: GraphQL transport

Adds an **opt-in, off-by-default** `useGraphQL` preference. It requires a GitHub token — GraphQL has no unauthenticated tier — and falls back to REST without one, or on any error. Both commands share a single entry point so exactly one transport ever runs.

Measured against the live API: 1 point of the 5,000/hr GraphQL budget for 50 PRs, and `states: MERGED` filters server-side so nothing is fetched only to be discarded.

REST remains the default and fully supported path.

## Screencast

The screenshots have been updated to reflect the changes above (rounded extension icons, category icons, the menu bar).

hero

## Notes for reviewers

- **Screenshot count went 7 → 6.** The 7th was a menu-bar shot that predates these changes and no longer reflects the UI; the remaining six are all newly captured.
- **`docs/api-cost.md`** documents the API budget — which endpoints are billed, why the `/files` fallback is capped, and why GraphQL is opt-in.
- `media/` holds the two README images, kept outside `metadata/` per the guidelines.

## Checklist


0xdhrv (approved):
Looks good to me, approved ✅

PR #29821, opened 2026-07-29
Add raycast-dev-project-launcher extension

amine2233:
## Description

**Dev Project Launcher** recursively scans your development directories and turns every
subfolder into a launchable project entry, so you can jump into the right tool without
leaving Raycast.

- **Recursive discovery** — scans whichever of `~/Developer`, `~/Development`, `~/Projects`,
  `~/Code`, `~/dev` exist (or any roots you configure), up to a configurable depth, skipping
  `node_modules`, `Pods`, `DerivedData` and any folder starting with `.` or `_`.
- **Project-type detection** via marker files — Xcode, Swift Package, Kotlin/Gradle,
  Android/Gradle, Node, TypeScript, Python, Rust, Go, Java/Maven, Flutter, Ruby, or a bare
  `.git` repo as fallback. The scanner stops descending once a folder is classified, so a
  monorepo's internal packages never appear as separate projects.
- **Opens with the right tool per type** — `Enter` uses the app mapped as *preferred* for the
  detected type (Xcode projects and Swift packages open in Xcode via `xed`, Gradle in Android
  Studio, everything else in VS Code). VS Code (`Cmd+E`), WebStorm (`Cmd+W`) and iTerm
  (`Cmd+T`) remain available as explicit actions, alongside Reveal in Finder, Copy Path,
  Open With… and Rescan.
- **Manage App Paths** — a second command that stores the per-project-type app mapping in
  `LocalStorage`, so you can edit the app used for any type, or register a brand-new type
  (e.g. `deno`) with its own marker files and mapping, without touching source or
  reinstalling. Static extension preferences are a fixed build-time schema and can't grow new
  keys at runtime, which is why this lives in `LocalStorage` rather than in `package.json`.

Each app field accepts a bare CLI command (`xed`, `code`, `idea`), an absolute binary path, an
absolute `.app` path, or a bundle identifier such as `com.apple.dt.Xcode` — the last of these
so that side-by-side installs like `/Applications/Xcode-26.6.0.app` still resolve.

## Review feedback addressed

- **Default bypasses root detection** — removed the `~/Developer` default from
  `developmentPath`. A non-empty default meant `detectDefaultRoots()` never ran, so a user
  whose projects live in `~/Projects` saw "No projects found". The preference now defaults to
  empty and auto-detection does its job.
- **Custom types never match projects** — user-registered types can now declare marker files
  (new **Marker Files** field in *Manage App Paths*). `detectProjectType()` checks them before
  the builtin rules, so declaring `deno.json` for a `deno` type wins over the `package.json`
  that would otherwise classify the folder as Node. Previously a custom mapping could never be
  reached, since the scanner only ever emitted builtin types or `generic`.
- **View commands lack Store metadata** — screenshot moved from `assets/` to `metadata/` and
  resized to the required 2000×1250.
- **Prettier width diverges from standard** — `printWidth` raised from 100 to 120 and the
  source reformatted.

## Checklist


amine2233:
why the PR is removed ?

PR #29822, opened 2026-07-29
Add argocd extension

erecarte-twilio:
## Description

Adds an argocd extension which lets you access ArgoCD resources quicker.

## Checklist


pernielsentikaer (commented):
Hi @erecarte-twilio 👋

The Search Applications and Search Projects lists show a helpful EmptyView when the API call fails, but not when the cluster is empty or the search matches nothing. In those cases Raycast just shows a blank “No results” list, which feels broken. Please add `List.EmptyView` for both the empty-data and no-match cases (and keep the existing error view).

Favorites can still lose a star that is toggled while the first LocalStorage read is in flight. `useFavorites` already returns `ready`, but the list commands never use it, so the initial `setFavorites(new Set(list))` can overwrite a newer toggle. Please sequence that first read against toggles (or ignore the load if a toggle already happened) so a favorite cannot disappear after you just added it.

Search Applications and Search Projects have no command subtitle, and the login command is titled just “Login”. In the root search that makes the extension hard to recognise. Please add a subtitle like `Argo CD` on every command, and rename Login to something like `Log In` / `Login to Argo CD`.

The official product name is **Argo CD** (with a space). The manifest title, README, and CHANGELOG currently say `ArgoCD`. Please align those with the official spelling. The slug `argocd` can stay.

On the resources Grid, `tintedIconForKind` computes a health tint and then returns the raw SVG path, so the kind tiles stay gray instead of reflecting Degraded / Progressing / Healthy. Please apply the tint to the SVG (or use a Raycast icon with `tintColor` as the fallback you already have).

If the `argocd` CLI is not on `PATH`, Login surfaces a raw `spawn argocd ENOENT` error. A short message pointing at `brew install argocd` (and the API Token preference as an alternative) would help a lot.

---

PR #29823, opened 2026-07-29
Add tidyports extension

dan-fetch-studio:
## Checklist


dan-fetch-studio:
Fixed the P1 from the automated review — it was a real bug.

Only `TidyPortsMissingError` was handled, so a CLI timeout, non-zero exit, or malformed JSON left `data` undefined and fell through to the "Nothing listening" empty state. That reports a quiet machine when the lookup actually failed, which is the exact misdiagnosis this extension exists to prevent. It now shows the error with a retry action.

PR #29825, opened 2026-07-29
skills: fix runtime detection and root-level SKILL.md lookup

keito4:
## Description

Fixes the four unresolved issues reported for the Skills extension.

**1. `Unable to find a working bunx or npx command` (list-installed-skills)**

`node-path-resolver.ts` only searched version managers (nvm/fnm/n/volta), Homebrew, `/usr/bin` and `/bin`. Raycast launches with the GUI PATH rather than the login shell's, so runtimes installed anywhere else were invisible. Added: `~/.bun/bin` (Bun's official installer, where `bunx` lives), Nix / nix-darwin profiles (`/etc/profiles/per-user/<user>/bin`, `~/.nix-profile/bin`, `/run/current-system/sw/bin`, `/nix/var/nix/profiles/default/bin`), mise and asdf shims, and `/usr/local/bin`.

**2. `Command failed: bunx --silent skills@latest list -g --json` (list-installed-skills)**

`bunx --silent` suppresses bun's own diagnostics, so the rejection carried no stderr and the run aborted without falling back. A bunx failure with no diagnostic output now retries through `npx`, which either succeeds or reports an actionable reason. Failures where bunx did explain itself still abort immediately, so no redundant second run.

**3 & 4. `Could not fetch SKILL.md ... may use a non-standard repo layout` (read-skill)**

Two causes:

- The repo-tree lookup filtered on `path.endsWith("/SKILL.md")`, which excludes single-skill repos that keep `SKILL.md` at the repository root. Root-level `SKILL.md` is now accepted, but only when it is the repo's sole skill, so a multi-skill repo's top-level file cannot shadow a real match.
- The `read-skill` AI tool had its own resolver that guessed at just two flat paths, instead of reusing `fetchSkillContent()` behind the detail view. It now delegates, so nested, owner-prefixed and root-level layouts all resolve.

## Verification

- `npm run lint`, `npx tsc --noEmit` and `npm run build` pass
- Against `alexi-build/raycast-extensions-skill` (the repo in the reported error): the two old candidate URLs return 404, the new root-level fallback resolves the 5,716-char `SKILL.md`
- On a Nix-managed Mac with a simulated GUI PATH (`/usr/bin:/bin:/usr/sbin:/sbin`): the old resolver fails with `ENOENT`, the new one resolves `npx 11.6.2` via `/etc/profiles/per-user/<user>/bin`
- Ran in Raycast on that same Nix-managed Mac, which gave a direct before/after on one machine: the current Store build fails "Manage Skills" with `Unable to find a working bunx or npx command`, while this build lists all 14 installed skills and their agents. `bunx` is genuinely absent there, so the read-only fallback to `npx` is what carries the command

## Checklist

### DIFF
--- a/extensions/skills/CHANGELOG.md
+++ b/extensions/skills/CHANGELOG.md
@@ -3,7 +3,8 @@
 ## [Fix Runtime Detection and Skill Lookup] - {PR_MERGE_DATE}
 
 - Detect `bun`/`node` installed outside a version manager or Homebrew, so "Unable to find a working bunx or npx command" no longer appears for Bun's official installer (`~/.bun/bin`), Nix / nix-darwin profiles, mise, and asdf
-- Fall back to `npx` when `bunx` fails without any diagnostic output, so a silent `bunx` failure no longer surfaces as "Command failed: bunx --silent skills@latest …" with no reason and no retry
+- Report what the `skills` CLI printed when it exits non-zero, instead of only "Command failed: bunx --silent skills@latest …" with no reason — the CLI writes its errors to stdout, which was being discarded
+- Fall back to `npx` when `bunx` dies without printing anything at all, limited to read-only commands so a mutating `add`/`remove`/`update` is never run twice
 - Resolve skills whose `SKILL.md` sits at the repository root (single-skill repos), which previously failed to load in the detail view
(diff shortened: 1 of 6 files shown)


0xdhrv (approved):
Looks good to me, approved ✅

PR #29830, opened 2026-07-29
Update Otty extension icon

nightsu:
## Summary

- Replace the Otty extension icon with the flat Otty face artwork.
- Keep the bundled icon at 512×512 PNG (20 KB).

## Why

The previous icon did not match the intended Otty branding.

## Validation

- Confirmed the PNG is 512×512 with RGBA transparency.
- Confirmed the file is below the 50 KB target.


0xdhrv (approved):
Looks good to me, approved ✅

PR #29845, opened 2026-07-29
Update get-app-icon extension

chrismessina:
## Description

Exporting an app's icons twice wrote both exports to the same folder, so re-exporting after an app update silently replaced the previous version's icons with no warning. Export folders now include the app's version — `Bleep 3.4.0 App Icons` — so each version keeps its own. This also adds an **Export Icons As…** submenu for a one-off export in a single format, and fixes several grid-icon caching bugs found while verifying the change.

### Changes

**Added**

- **Versioned export folders.** Folder names now carry `CFBundleShortVersionString`. Re-exporting the *same* version still overwrites, which is the repair path for a partial or interrupted export.
- **Export Icons As…** — pick one format (PNG, JPEG, or ICNS) for a single export without changing your format preferences. For keeping PNG as the default but occasionally wanting only the ICNS.

**Fixed**

- **Grid icons now update when an app changes its icon.** Roughly half of installed apps name their icon file something other than the conventional `AppIcon.icns` — Visual Studio Code's is `Code.icns` — and a changed icon in one of those apps was invisible to the cache, so the grid could show a stale tile indefinitely.
- **Every export offers "Reveal in Finder" (`⌘O`) on its success toast.** The action was attached but carried no keyboard shortcut, so nothing on screen indicated it existed.
- **Failed exports no longer leave empty folders behind.** Format directories are created only once there is something to write, and the app folder is removed only when the export itself created it — a pre-existing folder is left alone even when empty.
- **Show Export Folder in Finder** finds exports made before folders were versioned, instead of reporting none.
- The grid no longer re-extracts icons it has already cached when part of an app bundle can't be read, and no longer keeps showing an out-of-date icon once it can be read again.
- Folder names are clamped to the filesystem's 255-byte limit by shortening the app name rather than the version, so two releases can never collapse onto one folder.

### Note on the icon cache

The grid's icon cache was reworked during this PR. Cache entries are now named for the source state they were drawn from — a hash of the app path plus the observed timestamps of the files that determine its icon — so checking freshness is a single "does this file exist?" rather than a comparison between several separately-read values. Entries for superseded states are collected on the next grid visit.

The practical effect is the stale-tile fix above, plus the two blind-source behaviours. Users upgrading will see one normal icon extraction on their first grid visit (about a second for a few hundred apps, behind the existing progress indicator) as the cache repopulates under the new naming.

## Checklist

- [ ] I ran `npm run build` and tested this distribution build in Raycast


chrismessina:

chrismessina:
@greptile This finding is stale — it describes a revision that no longer exists on this branch. Please re-check against head `5ad757b6dc77`.

**`src/icon-cache.ts:198–214`** — failed stats are no longer converted uniformly to zero. Only `ENOENT`/`ENOTDIR` (genuinely absent) contribute 0; every other errno sets `unverifiable`:

```ts
async function iconSourceStamp(appPath: string): Promise<IconSourceStamp{
  const results = await Promise.all(
    iconStampPaths(appPath).map(async (target) ={
      try {
        return { mtime: (await stat(target)).mtimeMs, unverifiable: false };
      } catch (error) {
        const code = (error as NodeJS.ErrnoException).code;
        const absent = code === "ENOENT" || code === "ENOTDIR";   // line 205
        return { mtime: 0, unverifiable: !absent };                // line 206
      }
    }),
  );
  return {
    mtime: Math.max(...results.map((result) =result.mtime)),
    unverifiable: results.some((result) =result.unverifiable),
  };
}
```

**`src/icon-cache.ts:226–240`** — an unverifiable source short-circuits to stale **before** any mtime comparison, so an unchanged root timestamp is never reached:

```ts
const [cachedStat, source] = await Promise.all([stat(cached), iconSourceStamp(appPath)]);
if (source.unverifiable) return appPath;                          // line 232
return cachedStat.mtimeMs >= source.mtime ? null : appPath;
```

Trace for the scenario in the comment: root `100`, cached PNG `150`, children `EACCES` → `{ mtime: 100, unverifiable: true }` → **stale, re-extracted**. The pre-update icon is not accepted.

On the second clause: wall-clock stamping is gated at **line 286** by `usable = !source.unverifiable && …`, so a missing optional leaf alone can't trigger it. When it does apply, the timestamp is accurate rather than optimistic — extraction draws through `NSWorkspace`, which resolves the icon even when a subpath can't be `stat`'d, so those pixels are post-update.

Verified empirically at this head: bundle root `chmod 000` → `Contents`/`Info.plist`/`Resources` all `EACCES` → re-extracts; after access recovers it settles rather than re-extracting on every grid visit.

0xdhrv (approved):
Looks good to me, approved ✅

PR #29852, opened 2026-07-30
Update github extension

### LINKED ISSUE #29778: [GitHub] Add “Show README” action to repository commands
### Extension

https://www.raycast.com/raycast/github

### Description

After finding and selecting a repository through any of these commands:

- **Search Repositories**
- **My Latest Repositories**
- **My Starred Repositories**

There should be a **Show README** action in the repository’s action menu. Selecting it should display the repository’s README inside Raycast instead of opening a browser.

Expected behavior:

- Render the README clearly in a Raycast Detail view.
- Display relative links and images correctly.
- Show a clear message when the repository has no README.
- Keep the existing **Open in Browser** action available.

### Who will benefit?

Anyone who wants to quickly understand a repository without leaving Raycast.

### Anything else?

Similar functionality was requested and implemented for the GitLab extension:

- [GitLab feature request #21377](https://github.com/raycast/extensions/issues/21377)
- [GitLab implementation PR #23378](https://github.com/raycast/extensions/pull/23378)

The GitHub Trending extension also has relevant README rendering improvements that may be useful as a reference:

- [GitHub Trending PR #27985](https://github.com/raycast/extensions/pull/27985)

ridemountainpig:
## Description
- Added a "Show Readme" action to the Search Repositories, My Latest Repositories, and My Starred Repositories commands that renders the repository's README inside Raycast, with relative links and images resolved to absolute URLs. Close #29778.

## Screencast
CleanShot 2026-07-30 at 14 17 29@2x
CleanShot 2026-07-30 at 14 25 15@2x

## Checklist

### DIFF
--- a/extensions/github/src/components/RepositoryReadme.tsx
+++ b/extensions/github/src/components/RepositoryReadme.tsx
@@ -1,0 +1,27 @@
+import { Action, ActionPanel, Detail } from "@raycast/api";
+
+import { ExtendedRepositoryFieldsFragment } from "../generated/graphql";
+import { useReadme } from "../hooks/useRepositories";
+
+export default function RepositoryReadme({ repository }: { repository: ExtendedRepositoryFieldsFragment }) {
+  const { data, isLoading } = useReadme(repository);
+
+  const markdown =
+    !isLoading && data && !data.found
+      ? `# ${repository.nameWithOwner}\n\n> This repository doesn't have a README.`
+      : (data?.markdown ?? "");
+
+  return (
+    <Detail
+      isLoading={isLoading}
+      navigationTitle={`${repository.nameWithOwner} · README`}
+      markdown={markdown}
+      actions={
+        <ActionPanel title={repository.nameWithOwner}>
+          <Action.OpenInBrowser url={repository.url} />
+          <Action.CopyToClipboard content={repository.url} title="Copy Repository URL" />
+        </ActionPanel>
+      }
+    />
+  );
+}
--- a/extensions/github/src/components/RepositoryReadme.tsx
+++ b/extensions/github/src/components/RepositoryReadme.tsx
@@ -7,7 +7,7 @@ export default function RepositoryReadme({ repository }: { repository: ExtendedR
   const { data, isLoading } = useReadme(repository);
 
   const markdown =
-    !isLoading && data && !data.found
+    data && !data.found
       ? `# ${repository.nameWithOwner}\n\n> This repository doesn't have a README.`
       : (data?.markdown ?? "");
 
--- a/extensions/github/src/helpers/repository.ts
+++ b/extensions/github/src/helpers/repository.ts
@@ -215,10 +215,13 @@ export function rewriteReadmeUrls(markdown: string, rawDownloadUrl: string): str
     return markdown;
   }
 
+  // Insert `/blob/` after `owner/repo` instead of guessing the ref segment, so
+  // subdirectory READMEs (`.github/`, `docs/`) and slash-containing refs stay intact.
   const rawBase = rawDownloadUrl.slice(0, rawDownloadUrl.lastIndexOf("/") + 1);
-  const blobBase = rawBase
-    .replace("https://raw.githubusercontent.com/", "https://github.com/")
-    .replace(/\/([^/]+)\/$/, "/blob/$1/");
+  const blobBase = rawBase.replace(
+    /^https:\/\/raw\.githubusercontent\.com\/([^/]+)\/([^/]+)\//,
+    "https://github.com/$1/$2/blob/",
+  );
 
   const isAbsolute = (url: string) => /^[a-z][a-z0-9+.-]*:/i.test(url) || url.startsWith("//") || url.startsWith("#");
 
@@ -234,12 +237,15 @@ export function rewriteReadmeUrls(markdown: string, rawDownloadUrl: string): str
     }
   };
 
+  // Allow one level of balanced parentheses so filenames like `image(1).png`
+  // aren't truncated; the trailing group captures an optional title.
+  const destination = "((?:\\([^)]*\\)|[^()\\s])+)([^)]*)";
+  const imagePattern = new RegExp(`!\\[([^\\]]*)\\]\\(${destination}\\)`, "g");
+  const linkPattern = new RegExp(`(?<!!)\\[([^\\]]+)\\]\\(${destination}\\)`, "g");
+
   return markdown
-    .replace(/!\[([^\]]*)\]\(([^)\s]+)([^)]*)\)/g, (_m, alt, url, tail) => `![${alt}](${resolve(url, rawBase)}${tail})`)
-    .replace(
-      /(?<!!)\[([^\]]+)\]\(([^)\s]+)([^)]*)\)/g,
-      (_m, text, url, tail) => `[${text}](${resolve(url, blobBase)}${tail})`,
-    )
+    .replace(imagePattern, (_m, alt, url, tail) => `![${alt}](${resolve(url, rawBase)}${tail})`)
+    .replace(linkPattern, (_m, text, url, tail) => `[${text}](${resolve(url, blobBase)}${tail})`)
     .replace(/(<img[^>]+src=["'])([^"']+)(["'])/gi, (_m, pre, url, post) => `${pre}${resolve(url, rawBase)}${post}`)
     .replace(/(<a[^>]+href=["'])([^"']+)(["'])/gi, (_m, pre, url, post) => `${pre}${resolve(url, blobBase)}${post}`);
 }
--- a/extensions/github/src/helpers/repository.ts
+++ b/extensions/github/src/helpers/repository.ts
@@ -208,30 +208,38 @@ const formatRepositoryUrl = (repoNameWithOwner: string, protocol: "https" | "ssh
  *
  * @param markdown {string} Raw README markdown.
  * @param rawDownloadUrl {string} The README's raw download URL from the GitHub API.
+ * @param readmePath {string} The README's path within the repository (e.g. `.github/README.md`),
+ *   used to resolve root-relative paths (starting with `/`) against the repository root.
  * @returns {string} Markdown with relative URLs rewritten to absolute ones.
  */
-export function rewriteReadmeUrls(markdown: string, rawDownloadUrl: string): string {
+export function rewriteReadmeUrls(markdown: string, rawDownloadUrl: string, readmePath = ""): string {
   if (!rawDownloadUrl) {
     return markdown;
   }
 
   // Insert `/blob/` after `owner/repo` instead of guessing the ref segment, so
   // subdirectory READMEs (`.github/`, `docs/`) and slash-containing refs stay intact.
-  const rawBase = rawDownloadUrl.slice(0, rawDownloadUrl.lastIndexOf("/") + 1);
-  const blobBase = rawBase.replace(
-    /^https:\/\/raw\.githubusercontent\.com\/([^/]+)\/([^/]+)\//,
-    "https://github.com/$1/$2/blob/",
-  );
+  const toBlob = (raw: string) =>
+    raw.replace(/^https:\/\/raw\.githubusercontent\.com\/([^/]+)\/([^/]+)\//, "https://github.com/$1/$2/blob/");
+
+  // `*Dir` is the README's own directory (for normal relative paths); `*Root` is the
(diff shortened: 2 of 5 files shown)


extensions/github/src/helpers/repository.ts
```diff
@@ -194,6 +194,62 @@ type AdditionalCloneFormatOptions = {
...
+    const trimmed = url.trim();
+    if (!trimmed || isAbsolute(trimmed)) {
+      return url;
+    }
+    try {
+      return new URL(trimmed.replace(/^\/+/, ""), base).toString();
+    } catch {
+      return url;
+    }
+  };
```
gfazioli: **Blocking: angle-bracket destinations break already-absolute links**

CommonMark allows wrapping a link destination in `<>` so it can contain spaces or parentheses. `isAbsolute()` is tested on the raw string, which starts with `<`, so `/^[a-z][a-z0-9+.-]*:/` doesn't match — the destination is treated as relative and rebased onto `blobBase`.

Repro, straight out of `octokit/octokit.js`'s README:

```
IN : Node 18 ships [the `fetch` method](<https://nodejs.org/en/blog/announcements/v18-release-announce#fetch-(experimental)>).
OUT: Node 18 ships [the `fetch` method](https://github.com/octokit/octokit.js/blob/main/%3Chttps://nodejs.org/en/blog/announcements/v18-release-announce#fetch-(experimental)%3E).
```

The link renders fine but sends the user to a github.com 404 instead of nodejs.org. Note the balanced-parens handling already works here — it's only the `<>` wrapper that trips the absolute check.

Unwrapping before the check and re-wrapping on output fixes it:

```suggestion
  const resolve = (url: string, base: string) => {
    const trimmed = url.trim();
    // CommonMark allows wrapping the destination in <> so it can contain spaces or parentheses.
    const wrapped = /^<(.*)>$/.exec(trimmed);
    const inner = wrapped ? wrapped[1] : trimmed;
    if (!inner || isAbsolute(inner)) {
      return url;
    }
    try {
      const resolved = new URL(inner.replace(/^\/+/, ""), base).toString();
      return wrapped ? `<${resolved}>` : resolved;
    } catch {
      return url;
    }
  };
```

extensions/github/src/helpers/repository.ts
```diff
@@ -194,6 +194,62 @@ type AdditionalCloneFormatOptions = {
...
+ * @returns {string} Markdown with relative URLs rewritten to absolute ones.
+ */
+export function rewriteReadmeUrls(markdown: string, rawDownloadUrl: string): string {
+  if (!rawDownloadUrl) {
+    return markdown;
+  }
+
+  // Insert `/blob/` after `owner/repo` instead of guessing the ref segment, so
+  // subdirectory READMEs (`.github/`, `docs/`) and slash-containing refs stay intact.
+  const rawBase = rawDownloadUrl.slice(0, rawDownloadUrl.lastIndexOf("/") + 1);
```
gfazioli: **Root-relative paths should resolve against the repository root, not the README's directory**

GitHub's docs are explicit: *"Links starting with `/` will be relative to the repository root."* The `trimmed.replace(/^\/+/, "")` below collapses `/foo` and `foo` into the same resolution, which is only equivalent when the README itself sits at the repository root. For a `.github/README.md`:

```
IN : ![logo](/assets/logo.png)
OUT: https://raw.githubusercontent.com/o/r/main/.github/assets/logo.png   <- broken
GH  : https://raw.githubusercontent.com/o/r/main/assets/logo.png
```

The API response already carries what's needed to fix this exactly — pass `data.path` into the helper and derive a repository-root base by stripping the README's own path, which stays correct even for slash-containing refs:

```ts
const rootBase = rawDownloadUrl.slice(0, rawDownloadUrl.length - readmePath.length);
```

then pick `rootBase` (dropping the leading `/`) when the destination starts with a slash, and `rawBase`/`blobBase` otherwise.

Low urgency: 0 of the 15 READMEs I sampled use root-relative paths, and it only misbehaves for the `.github/` and `docs/` README locations.

extensions/github/src/hooks/useRepositories.ts
```diff
@@ -44,3 +45,24 @@ export function useReleases(repository: ExtendedRepositoryFieldsFragment) {
...
+export function useReadme(repository: ExtendedRepositoryFieldsFragment) {
+  const { octokit } = getGitHubClient();
+
+  const [owner, name] = repository.nameWithOwner.split("/");
+  return useCachedPromise(
+    async (owner: string, name: string) => {
+      try {
+        const { data } = await octokit.rest.repos.getReadme({ owner, repo: name });
+        const content = Buffer.from(data.content, "base64").toString("utf-8");
+        return { markdown: rewriteReadmeUrls(content, data.download_url ?? ""), found: true };
```
gfazioli: **Two edge cases around the decoded content**

**Non-Markdown READMEs.** `getReadme` resolves whatever file GitHub picked, and that isn't always Markdown:

| Repository | `path` |
|---|---|
| `python/cpython`, `scipy/scipy` | `README.rst` |
| `elastic/elasticsearch` | `README.asciidoc` |
| `torvalds/linux` | `README` |

reStructuredText `.. image::` directives and `====` underlines render as mangled Markdown in the `Detail` — no badges, no images, broken headings. Gating on something like `/\.(md|markdown|mdown|mkdn?)$/i.test(data.name)` and wrapping non-Markdown content in a fenced block would degrade gracefully instead.

**READMEs above 1 MB.** The Contents API returns `content: ""` with `encoding: "none"` past that size, so this path yields `{ found: true, markdown: "" }` — an empty `Detail` with no explanation, indistinguishable from a bug. A guard on `data.encoding !== "base64" || !data.content` could show a short "too large to preview, open in browser" message.

Minor, same line: the extension's other call sites use `octokit.<namespace>.<method>` (`octokit.actions.…` in `tools/list-workflow-runs.ts`, `octokit.activity.…` in `unread-notifications.tsx`) rather than the `.rest.` alias — worth matching for consistency.

extensions/github/src/components/RepositoryReadme.tsx
```diff
@@ -0,0 +1,27 @@
...
+import { ExtendedRepositoryFieldsFragment } from "../generated/graphql";
+import { useReadme } from "../hooks/useRepositories";
+
+export default function RepositoryReadme({ repository }: { repository: ExtendedRepositoryFieldsFragment }) {
+  const { data, isLoading } = useReadme(repository);
+
+  const markdown =
+    !isLoading && data && !data.found
+      ? `# ${repository.nameWithOwner}\n\n> This repository doesn't have a README.`
+      : (data?.markdown ?? "");
```
gfazioli, line 12: **The `!isLoading` guard suppresses the empty state on repeat visits**

`useCachedPromise` is stale-while-revalidate: on a second visit to a README-less repository it returns the cached `{ found: false, markdown: "" }` *with* `isLoading: true`. The condition then fails and the expression falls through to `data?.markdown ?? ""`, so the `Detail` renders blank until revalidation finishes, at which point the message pops in.

The guard is also redundant — `data` is `undefined` before the first resolution, so `data && !data.found` already covers the initial load on its own:

```suggestion
  const markdown =
    data && !data.found
      ? `# ${repository.nameWithOwner}\n\n> This repository doesn't have a README.`
      : (data?.markdown ?? "");
```

extensions/github/src/helpers/repository.ts
```diff
@@ -194,6 +194,62 @@ type AdditionalCloneFormatOptions = {
...
+  // aren't truncated; the trailing group captures an optional title.
+  const destination = "((?:\\([^)]*\\)|[^()\\s])+)([^)]*)";
+  const imagePattern = new RegExp(`!\\[([^\\]]*)\\]\\(${destination}\\)`, "g");
+  const linkPattern = new RegExp(`(?<!!)\\[([^\\]]+)\\]\\(${destination}\\)`, "g");
+
+  return markdown
+    .replace(imagePattern, (_m, alt, url, tail) => `![${alt}](${resolve(url, rawBase)}${tail})`)
+    .replace(linkPattern, (_m, text, url, tail) => `[${text}](${resolve(url, blobBase)}${tail})`)
+    .replace(/(<img[^>]+src=["'])([^"']+)(["'])/gi, (_m, pre, url, post) => `${pre}${resolve(url, rawBase)}${post}`)
+    .replace(/(<a[^>]+href=["'])([^"']+)(["'])/gi, (_m, pre, url, post) => `${pre}${resolve(url, blobBase)}${post}`);
```
gfazioli: **Minor: the replaces are context-free, so code blocks get rewritten too**

A README that documents Markdown or HTML snippets will have the URLs inside its examples resolved — `[example](./foo.md)` written inside a fenced block comes out as `[example](https://github.com/owner/repo/blob/main/foo.md)`, and `` in an HTML example gets absolutized as well.

I masked fenced blocks and inline code spans across 15 real READMEs (114 fence delimiters in total) and got 0 hits, so this is genuinely low priority — flagging it only in case you'd rather skip code regions.

Related and equally low priority: reference-style destinations aren't handled at all, so `![logo][l]` with `[l]: assets/logo.png` at the bottom of the file stays relative and the image doesn't load.

gfazioli (changes requested):
## Summary

Nice addition — an in-Raycast README viewer is a natural fit for the repository commands, and it's well integrated: `⌥⌘R` is free extension-wide and matches the section's `⌥⌘<letter>` pattern (`⌥⌘I`, `⌥⌘P`), `onPush` mirrors the sibling push actions, the empty-state copy follows the existing `ProjectDetail.tsx` pattern, and the CHANGELOG correctly uses `{PR_MERGE_DATE}`.

Deriving the base URL from the README's `download_url` instead of guessing the ref is the right design decision. I verified it holds for the two cases a naive implementation breaks on:

- subdirectory README: `.github/README.md` + `../docs/x.md` → `https://github.com/o/r/blob/main/docs/x.md` ✅
- slash-containing ref: `release/v2` → `https://github.com/o/r/blob/release/v2/docs/x.md` ✅

Both greptile P1s are addressed at `bc58bc3` — verified, they can stay resolved.

## Requesting changes for one issue

`rewriteReadmeUrls()` turns **already-absolute** external links into broken `github.com` URLs when the destination uses CommonMark's angle-bracket form `[text](<url>)`. Reproduced against the README of `octokit/octokit.js` — the library this PR calls:

```
IN : [the `fetch` method](<https://nodejs.org/en/blog/announcements/v18-release-announce#fetch-(experimental)>)
OUT: [the `fetch` method](https://github.com/octokit/octokit.js/blob/main/%3Chttps://nodejs.org/en/blog/announcements/v18-release-announce#fetch-(experimental)%3E)
```

The user clicks a Node.js docs link and lands on a github.com 404. Details and a suggested fix are inline.

## Non-blocking findings

I tested the helper against a corpus of 15 real READMEs; the frequencies below are from that sample.

1. **Non-Markdown READMEs render as garbage** — `getReadme` also resolves `README.rst` (`python/cpython`, `scipy/scipy`), `README.asciidoc` (`elastic/elasticsearch`) and extension-less `README` (`torvalds/linux`). Inline.
2. **Root-relative paths (`/…`) resolve against the README's directory, not the repository root** — GitHub's docs: *"Links starting with `/` will be relative to the repository root."* Only wrong for subdirectory READMEs. Inline.
3. **The empty state is suppressed on repeat visits** because of `useCachedPromise`'s stale-while-revalidate semantics. Inline.
4. **READMEs above 1 MB render blank** — the Contents API returns `content: ""` with `encoding: "none"`. Inline.
5. **Code fences and inline code are rewritten too** — 0/15 in my corpus, so low priority. Inline.
6. **Reference-style links and images are not rewritten** — `[l]: assets/logo.png` with `![logo][l]` stays relative and the image breaks. Also 0/15.
7. **Quadratic backtracking** on an unclosed `](` followed by a long non-whitespace run: 8k chars ≈ 63 ms, 32k ≈ ~1 s. Pathological input only, noting for completeness.
8. **Any 404 is reported as "doesn't have a README"**, including a repository that was renamed or became inaccessible.
9. **Octokit call style** — every other call site in the extension uses `octokit.<namespace>.<method>` (`octokit.actions.…`, `octokit.activity.…`); this one uses the `.rest.` alias.
10. **Naming** — the CHANGELOG writes "README" but the action title reads "Show Readme".
11. **`Detail` polish** — `Action.OpenInBrowser url={repository.url}` opens the repository root; `data.html_url` would open the README itself. A `Detail.Metadata` block would match `IssueDetail.tsx`.

Everything outside the first section is optional from my side — happy to approve once the angle-bracket case is handled.

pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #29853, opened 2026-07-30
Google Chrome: speed up Search Tabs

qeesung:
## Description

`Search Tabs` re-ran the tab-listing AppleScript on **every keystroke**, so typing in the search bar repeatedly shelled out to Chrome instead of filtering the list it already had.

Two changes:

1. **Fetch the tab list once per command run.** `query` was a dependency of the `usePromise` call in `useTabSearch`, so every character typed re-triggered `getOpenTabs()`. The list is now fetched once and filtered in memory via `useMemo`. Lowercasing is hoisted out of the per-keystroke path too.

2. **Batch the Apple Events.** The script read `title` and `URL` one tab at a time (`repeat with t in tabs of w`), costing two Apple Events per tab. Reading `title of tabs of w` / `URL of tabs of w` fetches a whole window in one event each. The favicon still needs one event per tab, so it is only evaluated when the `useOriginalFavicon` preference is on.

## Measurements

Measured on macOS with Chrome, comparing the old and new AppleScript directly:

| | Before | After |
|---|---|---|
| Listing 42 tabs across 2 windows | 1570 ms | 250 ms |
| Listing 18 tabs across 3 windows | 1080 ms | 315 ms |
| Opening Search Tabs and typing a 3-char query | ~6120 ms (4 script runs) | 278 ms (1 run) |

## Correctness

The new AppleScript's output was verified byte-for-byte identical to the previous implementation, including window IDs and tab indices — `setActiveTab` / `closeActiveTab` / `reloadTab` locate tabs by those values, so a mismatch there would activate the wrong tab.

The favicon branch (`useOriginalFavicon`) changes `execute t javascript` to `execute tab i of w javascript` since the loop no longer binds `t`; that branch compiles and behaves the same as before.

I also tried rendering the previously cached tabs on open via `useCachedPromise` to remove the initial wait entirely, but dropped it: showing a closed tab for even a few hundred milliseconds means the user can hit Enter on a tab that no longer exists. Correctness of the list matters more here than shaving the last 300 ms.

## Checklist

### DIFF
--- a/extensions/google-chrome/CHANGELOG.md
+++ b/extensions/google-chrome/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Read the open tabs from Chrome only once per command run instead of on every keystroke; filtering now happens in memory.
 - Fetch tab titles and URLs per window in a single Apple Event instead of two per tab.
+- Address tabs by Chrome's tab id instead of their position, so activating, reloading or closing a tab is unaffected by tabs being opened, closed or reordered in the meantime.
 
 ## [Add Search Windows Command] - 2026-08-16
 
--- a/extensions/google-chrome/CHANGELOG.md
+++ b/extensions/google-chrome/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Google Chrome Changelog
 
+## [Speed up Tab Search] - {PR_MERGE_DATE}
+
+- Read the open tabs from Chrome only once per command run instead of on every keystroke; filtering now happens in memory.
+- Fetch tab titles and URLs per window in a single Apple Event instead of two per tab.
+
 ## [Add Search Windows Command] - 2026-08-16
 
 - Add a new Search Windows command to search and select open Google Chrome windows.
--- a/extensions/google-chrome/src/actions/index.tsx
+++ b/extensions/google-chrome/src/actions/index.tsx
@@ -36,7 +36,7 @@ export async function getOpenTabs(useOriginalFavicon: boolean): Promise<Tab[]> {
           repeat with i from 1 to count of _props
             set _p to item i of _props
             ${faviconStatement}
-            set _output to (_output & (title of _p) & _field_sep & (URL of _p) & _field_sep & _favicon & _field_sep & _w_id & _field_sep & i & _field_sep & (id of _p) & _rec_sep)
+            set _output to (_output & (id of _p) & _field_sep & _w_id & _field_sep & i & _field_sep & (URL of _p) & _field_sep & _favicon & _field_sep & (title of _p) & _rec_sep)
           end repeat
         end repeat
       end tell
--- a/extensions/google-chrome/src/actions/index.tsx
+++ b/extensions/google-chrome/src/actions/index.tsx
@@ -5,9 +5,12 @@ import { NOT_INSTALLED_MESSAGE } from "../constants";
 import { runAppleScript as runAppleScriptRaycast, showFailureToast } from "@raycast/utils";
 
 export async function getOpenTabs(useOriginalFavicon: boolean): Promise<Tab[]> {
-  // Reading `title of tabs` / `URL of tabs` fetches a whole window in one Apple Event,
-  // instead of two round-trips per tab. The favicon still needs one event per tab,
-  // so it is only evaluated when the preference is on.
+  // `properties of tabs` returns every tab's title, URL and id for a whole window in
+  // a single Apple Event, instead of two round-trips per tab. Reading the three
+  // attributes as separate lists would be a similar win, but they would be separate
+  // snapshots: a tab opened or closed in between shifts one list against the others
+  // and a row would end up carrying another tab's id. The favicon still needs one
+  // event per tab, so it is only evaluated when the preference is on.
   const faviconFormula = useOriginalFavicon
     ? `execute tab i of w javascript ¬
         "document.head.querySelector('link[rel~=icon]') ? document.head.querySelector('link[rel~=icon]').href : '';"`
@@ -21,12 +24,11 @@ export async function getOpenTabs(useOriginalFavicon: boolean): Promise<Tab[]> {
       tell application "Google Chrome"
         repeat with w in windows
           set _w_id to get id of w as inches as string
-          set _titles to title of tabs of w
-          set _urls to URL of tabs of w
-          set _tab_ids to id of tabs of w
-          repeat with i from 1 to count of _titles
+          set _props to properties of tabs of w
+          repeat with i from 1 to count of _props
+            set _p to item i of _props
             set _favicon to ${faviconFormula}
-            set _output to (_output & item i of _titles & "${Tab.TAB_CONTENTS_SEPARATOR}" & item i of _urls & "${Tab.TAB_CONTENTS_SEPARATOR}" & _favicon & "${Tab.TAB_CONTENTS_SEPARATOR}" & _w_id & "${Tab.TAB_CONTENTS_SEPARATOR}" & i & "${Tab.TAB_CONTENTS_SEPARATOR}" & item i of _tab_ids & "\\n")
+            set _output to (_output & (title of _p) & "${Tab.TAB_CONTENTS_SEPARATOR}" & (URL of _p) & "${Tab.TAB_CONTENTS_SEPARATOR}" & _favicon & "${Tab.TAB_CONTENTS_SEPARATOR}" & _w_id & "${Tab.TAB_CONTENTS_SEPARATOR}" & i & "${Tab.TAB_CONTENTS_SEPARATOR}" & (id of _p) & "\\n")
           end repeat
         end repeat
       end tell
@@ -158,6 +160,9 @@ export async function setActiveTab(tab: Tab): Promise<void> {
           if (item i of _tab_ids as text) is "${tab.tabId}" then
             set index of w to 1
             set active tab index of w to i
+            -- Confirm the index still resolved to the intended tab; if the tab moved
+            -- in between, fail loudly rather than leaving another tab activated.
+            if (id of active tab of w as text) is not "${tab.tabId}" then error "Tab moved while activating"
             return true
           end if
         end repeat
--- a/extensions/google-chrome/src/actions/index.tsx
+++ b/extensions/google-chrome/src/actions/index.tsx
@@ -121,24 +121,37 @@ export async function openNewTab({
   }
 }
 
-// The three actions below address tabs by Chrome's tab id rather than by position.
-// A listed tab's position is only valid for the instant it was read: if tabs are
-// opened, closed or reordered afterwards, a positional index points at a different
-// live tab — which for `closeActiveTab` means closing the wrong one.
+// The actions below address tabs by Chrome's tab id rather than by position, and
+// search every window rather than only the one the tab was listed in. A listed
+// tab's position is only valid for the instant it was read: tabs opened, closed,
+// reordered or dragged to another window afterwards would otherwise make an action
+// hit a different live tab — or, for `closeActiveTab`, close the wrong one.
+// Chrome reports tab ids as text, so they must be compared as text — an unquoted
+// numeric literal never matches and the loop would silently fall through.
+const locateTabById = (tabId: string) => `
+      set _found_window_id to missing value
+      set _found_index to 0
+      repeat with w in windows
+        set _tab_ids to id of tabs of w
+        repeat with i from 1 to count of _tab_ids
+          if item i of _tab_ids is "${tabId}" then
+            set _found_window_id to id of w
+            set _found_index to i
+            exit repeat
+          end if
+        end repeat
+        if _found_window_id is not missing value then exit repeat
+      end repeat
+      if _found_window_id is missing value then error "Tab not found"
+      set _wnd to first window whose id is _found_window_id
+      set index of _wnd to 1
+      set active tab index of _wnd to _found_index`;
 
 export async function setActiveTab(tab: Tab): Promise<void> {
   await runAppleScript(`
     tell application "Google Chrome"
       activate
-      set _wnd to first window where id is ${tab.windowsId}
-      set index of _wnd to 1
-      set _tab_ids to id of tabs of _wnd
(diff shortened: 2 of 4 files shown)


0xdhrv (commented):
Hi @qeesung 👋

Thanks for the contribution 🔥

---

qeesung:
Thanks both — Greptile's catch is correct, and it's worse than cosmetic staleness: `tabIndex` is positional, so restoring cached lines after tabs are opened/closed/reordered would make activate, close and reload act on a *different live tab*, not just show a dead one.

That commit is already gone. A tester hit the same root cause in practice (a closed tab still listed after reopening the command), so I dropped the caching change and force-pushed before this review landed — the branch you're looking at no longer contains it.

What remains are the two optimisations that don't trade away correctness:

- **Fetch the tab list once per command run.** `query` was in the `usePromise` deps, so every keystroke re-ran the AppleScript; filtering now happens in memory via `useMemo`.
- **Batch the Apple Events.** `title of tabs of w` / `URL of tabs of w` reads a whole window per event instead of two events per tab.

`src/search-all.tsx` is no longer touched either — that change only existed to stop the loading row from hiding cached rows, which is moot now.

New AppleScript output was diffed byte-for-byte against the old one, including window IDs and tab indices, since `setActiveTab`/`closeActiveTab`/`reloadTab` locate tabs by those values.

Current diff: 3 files, +46/-30. Ready for another look when you have time.

0xdhrv (commented):

---

qeesung:
Rebased onto the latest `main` — the conflict is resolved.

It was confined to `CHANGELOG.md` (my entry vs. the new `Add Search Windows Command` entry). Both are kept; mine sits on top since `{PR_MERGE_DATE}` resolves to a date later than 2026-08-16.

`src/actions/index.tsx` merged cleanly without intervention — my change is inside `getOpenTabs`, while the new `getOpenWindows` / `setActiveWindow` were appended at the end of the file, so the two edits never overlap.

Verified against the new code: `tsc --noEmit`, `ray lint` and `ray build` all pass, and the new Search Windows command builds fine alongside this change. `useWindowSearch` doesn't consume `useTabSearch`, so the hook's internal restructuring doesn't affect it.

Diff is unchanged at 3 files, +46/-30. Marking as ready for review again — thanks for your patience.

qeesung:
Good catch — fixed at the root rather than worked around.

The underlying issue is that tabs were addressed by **position**. A listed tab's index is only valid for the instant it was read, so anything that opens, closes or reorders tabs afterwards makes that index point at a different live tab — and for `closeActiveTab` that means closing the wrong one. This predates the PR (the old code had the same flaw, with a shorter window since the list was re-read on every keystroke), but since this PR widens that window it belongs here.

`getOpenTabs` now also reads `id of tabs of w`, and `setActiveTab` / `closeActiveTab` / `reloadTab` locate their target by that id. `Tab.key()` uses it as well, so React keys stay stable when tabs are reordered.

Verified against a real Chrome window, reproducing exactly the scenario described:

- created tab A at index 12
- inserted another tab before it → A moves to index 13, index 12 is now the inserted tab
- closing A **by id** closed A and left the inserted tab intact

The previous index-based script would have closed the inserted tab in that state.

Cost is one extra Apple Event per window: 0.45s → 0.55s for 66 tabs across 6 windows, against 1.5s+ for the original per-tab implementation. `tsc --noEmit`, `ray lint` and `ray build` all pass.

qeesung:
Addressed — plus a real bug this review surfaced.

**Cross-window lookup.** The lookup now scans every window instead of only the one the tab was listed in, and raises `Tab not found` when the id is gone. Raycast surfaces that as an error, so it no longer silently does nothing.

One caveat worth recording, though: in the exact scenario described — moving a tab to *another window* — Chrome assigns it a **new id**. Verified on a live Chrome: a tab created with id `4272277`, then moved to another window, came back as `4271954`, with the old id gone entirely. So that particular case cannot be tracked by id at all, and the honest behaviour is a clear "Tab not found" rather than a silent no-op or acting on whatever now occupies that index. Cross-window search is still worth keeping for cases where a window changes identity, so it stays.

**The bug this uncovered.** Chrome reports tab ids as *text*, not numbers. My previous commit interpolated the id as a bare numeric literal, so `item i of _tab_ids is 4272205` never matched — `setActiveTab` fell through its loop and did nothing at all. `closeActiveTab` and `reloadTab` were unaffected only because their `whose id is ...` clause is resolved by Chrome itself, which coerces the comparison. `Tab.tabId` is now a `string`, compared as a quoted literal.

Verified against a live Chrome:

- activating now jumps to the requested tab (it was a no-op in the previous commit)
- closing tab A with a tab B appended after it removes A and leaves B intact

`tsc --noEmit`, `ray lint` and `ray build` all pass.

qeesung:
Both addressed.

**`as text` cast** — taken as suggested. Chrome does report ids as text on the build I tested against, but the cast makes the comparison hold either way.

**Stale index** — fair catch, and it defeated the point of the id lookup: the scan produced an index that a *later* Apple Event then consumed, reopening exactly the race the id was meant to close. `close` and `reload` no longer involve an index at any point — they act on the tab reference Chrome returns for `every tab of w whose id is ...`.

`setActiveTab` can't fully escape it: `active tab index` is the only activation Chrome exposes. `set active tab of w to <tab reference>` is rejected outright ("can't set active tab of window..."), so there is no reference-based path for activation. It now consumes the index inside the same loop iteration that produced it, rather than carrying it across statements, which narrows the window to a single iteration.

Verified against a live Chrome:

- activating jumps to the requested tab
- closing tab A, with a tab B appended after it, removes A and leaves B intact

`tsc --noEmit`, `ray lint` and `ray build` all pass.

qeesung:
Both points addressed — and the first one had a fix that's strictly better than what I had.

**Parallel lists losing identity.** Correct, and it was self-inflicted: three separate `title` / `URL` / `id` reads are three separate Apple Events, so a tab opening or closing in between shifts one list against the others and a row ends up carrying another tab's id.

`properties of tabs` returns all three for a whole window in a *single* event, so each row is one consistent snapshot. It also turns out to be faster than the three-list form — 0.35s against 0.58s for 61 tabs across 6 windows — so this removes the race and improves the original optimisation at the same time.

**Activation not revalidating.** `setActiveTab` now confirms that the index it set actually landed on the intended tab (`id of active tab`), and errors out instead of leaving a different tab activated. It still has to go through `active tab index`, since Chrome exposes no reference-based activation — `set active tab of w to <tab reference>` is rejected — but a mismatch is now loud rather than silent.

`close` and `reload` remain free of positional indices entirely; they act on the reference from `every tab of w whose id is ...`.

Verified against a live Chrome, all three actions:

- activating jumps to the requested tab and the post-check passes
- reloading targets the requested tab
- closing tab A, with tab B appended after it, removes A and leaves B intact

`tsc --noEmit`, `ray lint` and `ray build` all pass.

qeesung:
Both addressed.

**Favicon branch.** Correct — it indexed live tabs (`tab i of w`) *after* the property snapshot had been taken, so a tab opening or closing in between could attach one tab's icon to another tab's row, and a single unreachable tab aborted the entire listing. It now addresses the tab by id (`first tab of w whose id is (id of _p)`) and is wrapped in a `try`, so a failure leaves that one row's favicon empty rather than dropping every tab.

Worth noting this also fixes a pre-existing failure mode: with the favicon preference on but "Allow JavaScript from Apple Events" not enabled, the listing previously returned nothing. It now returns all tabs with empty favicons — verified on a live Chrome, all 66 tabs come back.

**Activation not retrying.** Fair — reporting the mismatch after the fact still left the wrong tab focused. `setActiveTab` now rescans and retries (up to three times) before giving up, so a shifted list is corrected rather than reported.

Verified against a live Chrome:

- listing returns every tab with the favicon preference enabled and Apple Events JavaScript unauthorised
- activation lands on the intended tab, post-check passes
- reload targets the intended tab
- closing tab A with tab B appended after it removes A and leaves B intact
- an unknown id fails with `Tab not found` rather than acting on something else

`tsc --noEmit`, `ray lint` and `ray build` all pass.

qeesung:
Good catch, and it was a genuine injection — thanks for pushing on it.

Reproduced before fixing: a page title containing `~~~` shifts every later field, so `tabId` ends up holding page-controlled text, and `tabId` is interpolated into the action scripts. A title ending in `" & (do shell script "…") & "` produced `whose id is "" & (do shell script "…") & ""`. A page fully controls its own title, so this was reachable by visiting a crafted page.

Fixed at the source rather than by escaping:

- **Delimiters are now the unit/record separator control characters** (`\x1F` / `\x1E`), matching `ChromeWindow` in this same file. Neither can appear in a title or URL, so fields cannot be shifted. This also closes the same hole via newlines, which previously separated records and are equally page-controlled.
- **`tabId` is validated to be digits before interpolation**, as defence in depth — it is the only field that reaches a script.

One bug this surfaced while testing: records were filtered by `line.length !== 0`, which was correct when splitting on `\n` but not when splitting on `\x1E` — the trailing newline osascript appends is *not* empty and parsed into a tab with no url and no id. Records are now filtered by containing a field separator.

Verified against a live Chrome:

- 67 records for 67 open tabs, every record with six fields and a numeric id
- a title containing both `~~~` and a newline no longer shifts any field
- a forged non-numeric id is rejected before reaching AppleScript
- activation, reload and close all still target the intended tab

`tsc --noEmit`, `ray lint` and `ray build` pass.

qeesung:
Addressed — the close/reload lookup now matches either form.

It first tries the quoted text comparison, and if that matches nothing, retries with an unquoted numeric one, so it resolves regardless of which type a given Chrome build reports. Interpolating unquoted is safe here because `tabIdForScript` has already validated the id to be digits.

For the record, on the Chrome build tested against, `whose id is "4272205"` and `whose id is 4272205` both resolve — Chrome coerces inside the `whose` clause, which is why close and reload kept working while the explicit `item i of _tab_ids is 4272205` comparison in the activation loop did not. The retry makes that independent of Chrome's behaviour rather than relying on it.

Verified on a live Chrome: reload targets the intended tab, close removes tab A and leaves an appended tab B intact, and an unknown id still fails with `Tab not found`.

`tsc --noEmit`, `ray lint` and `ray build` pass.

qeesung:
Fixed — a failed activation no longer changes what the user is looking at.

You're right that this was a real side effect rather than a theoretical one: activation must assign a positional index and then verify what it landed on, so when all three attempts failed, the last incorrect selection stayed active and the error was thrown on top of it.

The active window and tab are now captured before the first attempt and restored before the error is raised, so a failed jump leaves the browser exactly as it was found.

Verified against a live Chrome:

- with an id that matches nothing, the call fails with `Tab not found` and the previously active window and tab are unchanged
- with a valid id, activation still lands on the intended tab
- close and reload still target the intended tab

`tsc --noEmit`, `ray lint` and `ray build` pass.

qeesung:
Addressed structurally, and I measured the underlying assumption rather than asserting it.

**What Chrome actually does.** I tested whether a page can get a control character into its own title, both ways:

- `<title>A&#x1F;B</title>` via a `data:` URL
- `document.title = String.fromCharCode(65, 31, 66, 30, 67)`

In both cases Chrome hands the title back through AppleScript with those characters replaced by **spaces** (character id 32, verified per-character). So a page cannot embed a delimiter today, and the forged-id path is not reachable on current Chrome.

**Why the code no longer depends on that.** Relying on a browser's sanitising behaviour is a weak guarantee, so the record format no longer needs it: Chrome-controlled identifiers are serialised first and page-controlled strings last, with the title reclaiming any trailing fields.

    tabId ⟂ windowId ⟂ tabIndex ⟂ url ⟂ favicon ⟂ title…

If a delimiter ever did reach a title, it could only corrupt the trailing fields — it cannot shift `tabId`, which is the one value interpolated into AppleScript and used to target close and reload. Combined with the existing digit validation, forging an id would require Chrome to both stop sanitising titles *and* the id to survive a `^\d+$` check while sitting in field 0.

Verified against a live Chrome: every record parses with a numeric tab id; a title containing delimiters leaves the leading fields intact and is reassembled for display; activation, reload and close still target the intended tab.

`tsc --noEmit`, `ray lint` and `ray build` pass.

qeesung:
@0xdhrv this is ready for another look.

The merge conflict is resolved (rebased on current `main`), every review point is addressed, all checks are green, and Greptile's latest pass is **5/5 with no blocking findings** — it lists the stale-index, moved-window, id-type, snapshot-consistency, favicon-race and delimiter-injection issues as resolved.

Summary of what's in here:

**Performance** — the original intent

- The tab list is read once per command run instead of on every keystroke; filtering happens in memory.
- `properties of tabs` reads a whole window in one Apple Event instead of two per tab.

Measured with 66 tabs across 6 windows: listing went from ~1080 ms to ~350 ms, and typing a 3-character query went from roughly 6.1 s of AppleScript work down to a single call.

**Correctness** — surfaced during this review, all pre-existing except where noted

- Tabs are addressed by Chrome's tab id rather than by position, so an action can no longer hit a different tab after a reorder. This was reachable before: `closeActiveTab` could close the wrong tab.
- Lookups scan every window, and a failed activation restores the previously active tab instead of leaving another one selected.
- The favicon read addresses its tab by id and is wrapped in `try`, so it can neither mismatch icons nor abort the whole listing — the listing previously returned nothing when the preference was on without Apple Events JavaScript authorised.
- Record fields use control-character delimiters and are ordered so page-controlled text cannot reach `tabId`, with digit validation before interpolation. The injection this closes was introduced by this PR's own id handling and is fixed here.

Every behavioural claim above was verified against a live Chrome rather than reasoned about; details are in the comments on each commit. Happy to split any part out if you'd prefer a narrower PR.

0xdhrv (commented):
Could we avoid relying on `\x1F` / `\x1E` never appearing in page-controlled values such as the tab title?

The current parsing assumes Chrome sanitizes these control characters, but if one reaches the serialized output it can shift the fields and potentially associate a result with the wrong `tabId`. The numeric validation prevents arbitrary AppleScript injection, but it doesn't fully protect against targeting another valid numeric tab ID.

Could we use a serialization format that preserves field boundaries regardless of the page contents (for example, encoding the values or using JSON) instead?

---

PR #29870, opened 2026-07-30
Update convex extension

cn0ss:
## Description

Update to the **Convex** extension. The most important part: **the currently published version is broken** — Convex removed the internal dashboard endpoint the extension used for listing deployments (`/api/dashboard/projects/{id}/instances` now returns 404), so "Manage Projects" fails for every user. Additionally, newer Convex deployments live on region-scoped domains (`name.eu-west-1.convex.cloud`) where the `https://<name>.convex.cloud` URL the extension constructs no longer resolves, breaking Data / Logs / Run Function for those deployments.

### Fixes

- **Deployments list 404**: fetch deployments from the public Convex Management API (`/v1/projects/{id}/list_deployments`) instead of the removed internal endpoint
- **Region-scoped deployment URLs**: store the actual `deploymentUrl` reported by the API with the selected context and use it for all deployment API calls (data, logs, functions) instead of deriving it from the deployment name
- **Deploy key mode**: the "Configure Deploy Key" command saved keys to LocalStorage but the auth hook only read preferences, so keys configured through the command were never used; also adds deploy key support to "Open Dashboard", "Copy Deployment URL", and "Run Function", and derives the deployment type from the key prefix (`dev:`/`prod:`) instead of assuming prod
- **Stale context**: switching teams/projects now clears the downstream deployment selection in storage instead of silently keeping the previous team's deployment
- Profile fetching degrades gracefully when the endpoint rejects CLI-style tokens (was breaking "Manage Projects")

### New

- **Show Deployment Health** command: dashboard-style live charts (rendered as inline SVG) for function call volume, failure rate, cache hit rate, scheduler lag, and function concurrency — same `app_metrics` endpoints, window, and top-k defaults as the Convex dashboard's health page, auto-refreshing every 10s — plus cron jobs, scheduled functions, deployment state, and document count
- **Deployment Status in Menu Bar** command (disabled by default, 5m background refresh): call volume and failure alerting in the menu bar

### Maintenance

- `@raycast/utils` 1.x → 2.x, `@raycast/api` → 1.104.x, ESLint 8 (EOL) → 9
- Type checking re-enabled in the build script (was `--skip-types`) with the surfaced type errors fixed
- Removed four unwired legacy command files and unused components (~2,500 lines of dead code), stray screenshots in `metadata/`, a reserved `⌘,` shortcut, and non-Title-Case action titles

## Screencast

The existing store screenshots in `metadata/` still represent the extension. Happy to add a screencast of the new Deployment Health command if wanted.

## Checklist


cn0ss:
Both Greptile findings were valid — fixed and pushed (amended into the single commit):

1. **Metric failures report healthy status**: each health metric now resolves independently (`null` = unavailable, distinct from `0` = healthy). The menu bar shows a gray question-mark state with "Failure data unavailable" instead of a green check when data can't be fetched, and the health view marks affected metrics as "unavailable" with a retry note. If every metric fails (unreachable deployment), the whole refresh errors visibly.

2. **Logout retains deployment host**: `logout` now persists the cleared selection (including the stored deployment URL), so a later sign-in cannot inherit the previous account's deployment context.

pernielsentikaer (approved):
Looks good to me, approved

PR #29874, opened 2026-07-31
Update karakeep extension

chrismessina:
## Description

When a self-hosted Karakeep instance is unreachable, the extension previously failed with an opaque `fetch failed` error and an empty list — with no indication of what was wrong or how to fix it. This adds first-class handling for that state across every command.

For the common case of a local Docker deployment, it goes further: it detects that the container is stopped, offers to start it, waits for the API to come back, and reloads. Remote and hosted instances are unaffected — the Docker checks only run for a local address, and the recovery action is hidden when there's no local container to start.

karakeep-isnt-running

### Changes

- **Offline recovery across all 11 commands.** The seven list/detail views show a recovery screen with Start, Try Again, Open Docker, Extension Settings, and Copy Error actions instead of an empty list. The three create forms show an inline notice and promote "Start Karakeep" to the primary action. Quick Bookmark recovers inline through its toast.
- **Docker-aware start.** Finds the container by matching the configured port against declared port bindings, then starts the whole Compose project — Karakeep's stock deployment is `web` + `meilisearch` + `chrome`, and starting only `web` yields an instance that answers HTTP but can't search or crawl.
- **Typed input is protected.** Create and edit forms check reachability *before* writing, so a stopped instance can't cost you what you typed; the input is copied to the clipboard as a backstop.
- **Better error reporting.** Connection failures now surface the underlying cause (e.g. `ECONNREFUSED`) instead of `fetch failed`, and no longer stack a generic "Failed to fetch latest data" toast on top of the recovery screen.
- **Fixed stale-cache display.** A cached list could be shown as though it were current while the server was unreachable, including after switching to a different list or tag.
- Browser-extension links moved into an "Add to Browser" submenu.
- Updated the extension icon; refreshed screenshots (trimmed 9 → 6 for the Store limit).
- New strings are localized in both `en` and `zh`.

## Checklist


0xdhrv (approved):
Looks good to me, approved ✅

PR #29885, opened 2026-07-31
Update soundboard extension

muhammadrizo-y:
## Description

- Add support for Windows platform using Rust
- Add Stop action for currently playing sounds
- Add icons for sound entries

## Screencast

https://github.com/user-attachments/assets/94c1a42c-983c-4b5a-a0ff-036531b3afad

## Checklist


pernielsentikaer (changes requested):
Hi @muhammadrizo-y 👋

- **Please add a screencast.** The screencast section of the description is still the empty template, and Windows playback is the part none of the reviewers here can verify — we're all on macOS. A short recording of play, stop, and the new grid layout on Windows would let this move.
- **Windows Stop should explicitly tear down `MediaPlayer`** — in `rust/src/main.rs`, when the stop event fires (or playback ends), please `Pause`, clear the source, and call `Close()` on the player (a small RAII guard like your `StopEventGuard` is perfect so every return path cleans up). Relying on Drop/Release alone can leave audio playing after Stop. After the change, play → Stop mid-clip on Windows should go silent immediately.
- **macOS play failures** — when `afplay` errors (missing/unreadable file), show a Failure toast the same way the Windows path already does.

Two small notes, neither blocking:

- `build-rust` in `scripts` isn't needed — `ray build` compiles the Rust for you, and no other Rust extension in the repo carries an equivalent script. Worth dropping so the scripts block stays standard.
- In `play_windows`, playing the same file while it's already playing gets a handle to the existing named event and resets it, which would swallow a stop aimed at the first player. Genuinely an edge case; fine to leave, just flagging that I saw it.

Also worth knowing for later: the `stalled_ticks > 25` guard ends playback after ~5s without the position moving. Sensible safety net, but it's the first place I'd look if anyone reports a long sound cutting off early.

Nice-to-have: a small “playing” accessory on List/Grid items while a sound is active, so Stop is obvious without opening the Action Panel.

---

muhammadrizo-y:
Hi @pernielsentikaer 👋 
I've added a screencast and addressed the issues

pernielsentikaer (approved):
Hi @muhammadrizo-y 👋

The screencast, the Windows Close-on-stop path, and the macOS failure toast are all in. Per-instance stop events are the right fix for the same-file play race.

PR #29891, opened 2026-08-01
fix(brew): stop Show Installed rendering empty from a warm cache

TurboCoder13:
### 🤖 Disclosure

## Description

This PR fixes **Show Installed rendering an empty list on every open after the first** — a permanent
failure, not a transient one, because the broken value is what gets cached and re-served. Search is
affected by the same cause. No new commands, no new dependencies, no user-visible surface changes.

### Bug fixes

- **Show Installed lists nothing once the cache is warm.** The installed-packages hook resolves to a
  value holding two `Map`s (formulae and casks, keyed by name/token). `useCachedPromise` persists that
  value through a JSON cache, and `JSON.stringify(new Map())` is `{}` — so every entry is dropped on
  the way in. On the next open the value rehydrates as `{ formulae: {}, casks: {} }`, fails the
  consumer's `instanceof Map` guard, and both lists render empty while the user has hundreds of
  packages installed. It never recovers, because the poisoned entry keeps being re-served. Switching
  the filter dropdown doesn't help — it re-renders from the same broken value.

  Confirmed from disk rather than inferred. The serialised cache entry read, in full:

  ```json
  {"formulae":{},"casks":{}}
  ```

  Every other cached entry for the extension holds real data, because those values are plain arrays
  and survive serialisation.

- **Search stops marking installed packages** once that entry is poisoned — it consumes the same hook
  for installed-state detection on every row.

### Technical Highlights

- The hook now caches the serialisable `InstallableResults` (plain arrays) instead of the mapped
  `InstalledMap`, and rebuilds the name-keyed `Map`s in a `useMemo` on read. Lookup stays keyed
  access, not a linear scan.
- `asInstallableResults` rejects any cached value whose `formulae`/`casks` are not arrays, so entries
  written by earlier versions of the extension are discarded rather than rendered as an empty list.
- The hook's **returned shape is unchanged**, so no consumer needed editing.
- The `instanceof Map` guards in `showInstalledPackages` and `isInstalled` are deliberately left in
  place. They are redundant once the cause is fixed, but removing them belongs with the tests that
  would prove the new path — and they are what converts this failure into silent emptiness rather
  than a crash.
- Measured against real `brew info --json=v2 --installed` output (207 formulae, 48 casks): before,
  the cache entry serialised to `{"formulae":{},"casks":{}}` and a warm open rendered 0 formulae /
  0 casks; after, it renders 207 / 48.
- **Cache footprint grows from 26 bytes to ~670 KB** for that install set, since the entry now holds
  the real payload instead of nothing. This is the size the code always intended to store and it sits
  well inside Raycast's cache capacity, but it is a genuine change worth stating.
- No new dependencies, runtime or dev. No unrelated version bumps.

## Screencast

No new screenshots. This changes no visible surface — it makes an existing one work — so the current
`metadata/` captures remain accurate.

## How to Test

No special hardware or credentials needed. The bug is invisible on a first open, which is presumably
how it shipped, so **the second open is the case that matters**.

1. `npm ci && npm run dev` inside `extensions/brew`
2. Run **Brew: Clear Cache** to start cold.
3. Open **Show Installed**. The list populates. *(This works on `main` too.)*
4. Close the command and open **Show Installed** again. **On `main` this renders "No Results"
   permanently; with this change it populates.**
5. With the list populated, switch the filter to Formulae, then Casks, then back to All — each renders
   the correct subset.
6. Open **Search**, type a package you have installed, and confirm it is still marked as installed.
7. To see the old failure mode directly, inspect the extension's cached value between steps 3 and 4 —
   on `main` it reads `{"formulae":{},"casks":{}}`.
8. `npm run lint`, `npm run build`, `npx tsc --noEmit` all pass.

## Checklist

- [x] `npm run lint` passes

### DIFF
--- a/extensions/brew/CHANGELOG.md
+++ b/extensions/brew/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Brew Changelog
 
-## [Bug fix] - {PR_MERGE_DATE}
+## [Bug fix] - 2026-08-03
 
 - Fixed "Show Installed" listing no packages on every open after the first. The installed-package lookups are `Map`s, which serialise to `{}`, so the cached value was emptied on write and then re-served empty forever. The serialisable form is cached now and the lookups are rebuilt on read; cache entries written by earlier versions are discarded rather than trusted.
 - Fixed Search intermittently failing to mark packages as installed, which had the same cause.
--- a/extensions/brew/CHANGELOG.md
+++ b/extensions/brew/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Brew Changelog
 
+## [Bug fix] - {PR_MERGE_DATE}
+
+- Fixed "Show Installed" listing no packages on every open after the first. The installed-package lookups are `Map`s, which serialise to `{}`, so the cached value was emptied on write and then re-served empty forever. The serialisable form is cached now and the lookups are rebuilt on read; cache entries written by earlier versions are discarded rather than trusted.
+- Fixed Search intermittently failing to mark packages as installed, which had the same cause.
+
 ## [Bug fix] - 2026-07-10
 
 - Search now works instantly against the existing package index while it refreshes in the background, instead of blocking until the refresh completes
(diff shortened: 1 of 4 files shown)


pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #29892, opened 2026-08-01
[1Password] Stop re-authenticating while the desktop app session is usable

say8425:
## Description

Rebased onto `main` after #30195 landed, which rewrote most of `src/v8/utils.ts` and already covers the first review point. What is left is the macOS half of the same bug, plus the two error-surfacing gaps.

### The sign-in check still trusts `op whoami` on macOS

#30195 added an `op account get` fallback for `getSignInStatus()`, but returned early on every other platform:

```ts
} catch {
  if (!isWindows) {
    return false;
  }
```

So macOS still reports a signed-out state whenever the desktop app's delegated session is the only one available, and the extension shows **Authentication Required** even though the CLI can read items. This PR deletes those three lines; the reasoning in the comment right below them already applies to both platforms.

Measured on macOS 26.4.1 with `op` 2.34.0, app integration enabled, immediately after `op signout --all`:

| command | exit | time |
| --- | --- | --- |
| `op whoami` | 1 — `account is not signed in` | 104 ms |
| `op account get` | 0 — establishes the delegated session | 6147 ms |

Following the extension's own call order from that state, the whole chain now succeeds:

```
getSignInStatus()   op whoami                        exit 1    104ms
                    op account get                   exit 0   6147ms   -> true
useAccount()        op whoami --format=json          exit 0    132ms
usePasswords2()     op --account <id> items list     exit 0   1347ms   -> 639 items
```

### Errors during loading fall through to "No items found"

Only `ConnectionError` had a view of its own. Every other failure — including anything `handleErrors` maps to a generic `ExtensionError` — fell through to the empty-success state, which reads as an empty vault rather than a command that failed. Both list commands now render a shared `LoadError` view with **Retry** and **Copy Error Details**, so "No items found" means what it says.

The title falls back to a generic one unless the error carries a curated title, because `handleErrors` ends in `new ExtensionError(stderr)` and `ExtensionError` copies the title from the message when none is given — so a generic error would otherwise put raw `stderr` in the heading.

**My Vaults** also stopped resolving the account. It only passed it to `VaultActionPanel`, which never destructured it, and `vault list` does not need it — so an account lookup failing would otherwise have hidden a vault list that loaded fine. The unused `account` prop is gone too.

### Authentication errors lose everything after the first line

`errorRegex` captured a single log line, so the steps `op` prints for fixing the error were dropped. `extractOpErrorMessage` keeps everything from the first log line onwards. Because `List.EmptyView`'s description clamps to about two lines, the rest is reachable through **Copy Error Details**.

Checked against real `op` output — a multi-line error goes from 1 line to 3 including the remediation steps, and input with no `op` prefix (`spawn op ENOENT`) and empty input both pass through unchanged.

### README

The README listed four features and one FAQ entry, and did not mention the desktop app integration that 1Password 8 needs, the **Auto Renew Authorization** command, or any preference. Rewritten to cover requirements, commands, preferences, and troubleshooting.

`closeWindowAfterCopying` is documented as having no effect: it appears only in `package.json`, is never read in `src/`, and the copy/paste actions call `showHUD`, which dismisses Raycast either way.

## Review feedback

> **`useAccount()` still uses `op whoami`** … align account resolution with the delegated session

Resolved by #30195, which added the `whoami` → `account get` fallback with field mapping. Re-verified end-to-end above: from a signed-out state the list loads 639 items rather than falling through to "No items found".

> **Error EmptyViews** — when `accountError` / `itemsError` is set, show a failure `EmptyView`

Done in `LoadError.tsx`, used by both `Items.tsx` and `Vaults.tsx`.

One caveat on how the above was checked: the sign-in and account-resolution chain was verified by driving `op` directly in the states described, and the extension was run through `ray develop`, but I did not force a load failure to look at the new `LoadError` view in the UI. Worth a look when you test it.

## Trade-off worth flagging

`getSignInStatus()` runs synchronously while the view renders, and the `account get` fallback took 6.1 s when it had to establish the session. That cost only applies when no session exists — the state that previously showed **Authentication Required** and required a manual **Reload** — and `whoami` returns in ~130 ms once a session is live. It is the same behaviour #30195 already ships on Windows, so this PR does not change the shape of it, but it is worth a second opinion.

I tried guarding the probe with `OP_NO_AUTO_SIGNIN=1` to avoid the blocking. Measurements showed that makes the probe equivalent to `whoami` in every state I could produce — the variable disables exactly the session establishment that makes `account get` useful here — so I dropped it. What is left is a 30 s timeout, matching the one `signIn` already uses: not a budget for the probe, just a stop against the call hanging on a prompt nobody answers.

That hang matters most in **Auto Renew Authorization**, which runs unattended every 9 minutes. Before this PR the command was a no-op on macOS for delegated-session users, because its `getSignInStatus()` guard always failed for them — so this is new behaviour there, and if the app is locked the renewal can surface an authorization prompt. I have not been able to observe the locked-app path directly; if that is a concern, I am happy to gate the fallback out of the background command.

## Checklist

### DIFF
--- a/extensions/1password/CHANGELOG.md
+++ b/extensions/1password/CHANGELOG.md
@@ -1,5 +1,13 @@
 # 1Password Changelog
 
+## [Fix Repeated Re-Authentication] - {PR_MERGE_DATE}
+
+- Fixed the "Authentication Required" screen appearing on macOS even though the 1Password desktop-app integration works. The `op account get` fallback added in the previous release only ran on Windows, so macOS still trusted `op whoami`, which reports "account is not signed in" without an `op signin` session.
+- Because the same check gates **Auto Renew Authorization**, that command starts renewing on macOS for people who sign in through the desktop app. Its guard always failed for them before, so it never renewed anything.
+- Failures while loading accounts or items now show the error with **Retry** and **Copy Error Details** instead of the "No items found" empty state.
+- Authentication errors from the 1Password CLI keep their full output, including the steps the CLI suggests, instead of only the first line.
+- Documented the desktop app integration, the available commands and preferences, and how to enable Auto Renew Authorization in the README.
+
 ## [Windows Desktop-App Integration Fixes] - 2026-08-13
 
 - Fixed constant "cannot connect to 1Password app" failures on Windows. The 1Password app briefly has no CLI pipe listener right after a previous `op` process disconnects; back-to-back `op` invocations are now serialized and connection failures are retried instead of surfacing immediately.
--- a/extensions/1password/CHANGELOG.md
+++ b/extensions/1password/CHANGELOG.md
@@ -1,6 +1,6 @@
 # 1Password Changelog
 
-## [Fix Repeated Re-Authentication] - {PR_MERGE_DATE}
+## [Fix Repeated Re-Authentication] - 2026-08-14
 
 - Fixed the "Authentication Required" screen appearing on macOS even though the 1Password desktop-app integration works. The `op account get` fallback added in the previous release only ran on Windows, so macOS still trusted `op whoami`, which reports "account is not signed in" without an `op signin` session.
 - Because the same check gates **Auto Renew Authorization**, that command starts renewing on macOS for people who sign in through the desktop app. Its guard always failed for them before, so it never renewed anything.
--- a/extensions/1password/README.md
+++ b/extensions/1password/README.md
@@ -1,14 +1,91 @@
 # 1Password
 
-## Features
+Browse your 1Password items from Raycast, copy or paste credentials and one-time
+passwords, open items in the app or the browser, and generate new passwords.
 
-- Browse and copy passwords
-- Browse and manage vault list
-- Generate passwords
-- Support 1Password 7 and 1Password 8
+Works with 1Password 8 and 1Password 7. Pick which one you use under
+**1Password App Version** in the extension preferences.
 
-## FAQ
+## Requirements
 
-### I got the Authorization Requires prompt while using this extension
+### 1Password 8
 
-The usage of 1Password 8 requires [1Password CLI](https://developer.1password.com/docs/cli/get-started/).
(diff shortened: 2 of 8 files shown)


pernielsentikaer (changes requested):
Hi @say8425 👋

- **`useAccount()` still uses `op whoami`** (`src/v8/utils.ts`) while **My Passwords** gates `items list` on that succeeding. Auth now correctly treats a live delegated desktop session as signed in via `op account get`, but in the exact state you measured (`whoami` exits 1, `account get` exits 0) the list can still fail and fall through to **“No items found”** instead of loading items. Please re-verify that moment end-to-end in Raycast, and align account resolution with the delegated session (delegation-aware command + field mapping, whoami→account get fallback, or list without `--account` when safe).
- **Error EmptyViews** — when `accountError` / `itemsError` is set, show a failure `EmptyView` (message + Retry / Copy Error Details) instead of the empty-success “No items found” state.

---

say8425:
@pernielsentikaer Thanks for your feedback!

* The account-resolution half is already handled by #30195, which added the whoami => account get fallback with field mapping to `useAccount()`. What was still missing was the other half: #30195 gated the same fallback in `getSignInStatus()` to Windows, so on macOS whoami failing still meant a signed-out verdict and the Authentication Required screen — the list never got a chance to load. This PR removes that platform gate.
* Done. Both list commands now render a shared LoadError view — message, Retry, Copy Error Details — for any accountError / itemsError, instead of only ConnectionError.

pernielsentikaer (approved):
Hi @say8425 👋

Thanks for the follow-up 💪

Re-reviewed HEAD against the notes from last time — both landed:

- **Delegated session on macOS** — `getSignInStatus()` now falls back to `op account get` on every platform, so a live desktop session is no longer treated as signed out when `whoami` exits 1. Combined with the `useAccount()` fallback from #30195, My Passwords can load items in that state.
- **Error EmptyViews** — account/item failures use the shared LoadError view (message, Retry, Copy Error Details) instead of “No items found.”

Looks good to me, approved 🔥

PR #29893, opened 2026-08-01
fix(renaming): prevent silent data loss from case-only renames, hard links, and dangling symlinks

TurboCoder13:
### 🤖 Disclosure

## Description

This PR fixes one guard: the "target already exists" check that protects every rename. It
answered "are these two names the same file?" from the platform (`darwin` ⇒ lowercase both
strings) instead of from the filesystem, which fails in both directions — it lets a rename
destroy a real file on a case-sensitive volume, and it clears renames the filesystem will
silently refuse to perform. The guard now asks the filesystem directly (inode identity +
directory-entry counting), so it is correct on any volume without probing or special-casing.
No new commands, no UI changes, no new dependencies.

### Bug fixes

- **A case-only rename can silently destroy a file on a case-sensitive volume.** macOS is
  case-insensitive by default, not by requirement — APFS can be formatted case-sensitive, and
  commonly is on developer machines and external drives. There, `FOO.txt` and `foo.txt` are two
  files, but the guard lowercased both names, concluded "same file", and skipped the overwrite
  check — so renaming one onto the other replaced the target with no warning and no record.
  Batch pre-flight missed the same conflict for the same reason. The guard now compares
  `dev`+`ino`, so the answer is a property of the volume, not the platform.
- **Renaming a file onto a hard link of itself reported success while doing nothing.** POSIX
  `rename()` between two directory entries of one inode is a silent no-op, so inode identity
  alone is not enough to skip the guard — the extension reported a successful rename while both
  entries remained on disk. This held even for differently-cased or differently-normalized
  spellings of the link's name (`BAR.txt` for `bar.txt`, NFD for NFC), which a case-insensitive
  volume resolves onto the existing entry; verified against real `fs.rename` behaviour on APFS.
  Such renames are now blocked as "already exists".
- **A rename could silently destroy a symlink whose target is missing.** The existence check
  used `access()`, which follows symlinks — a dangling symlink at the target path was reported
  as "nothing there" and overwritten. `rename` operates on directory entries, so the check now
  uses `lstat`. The same change lets a dangling symlink itself be renamed (previously the
  source was reported missing).

### Technical Highlights

- `isSamePath` (string comparison) is replaced by two explicit predicates: `isSameFile`
  (`dev`+`ino` via `lstat`) and the stricter `isSameEntry`, the only check allowed to skip the
  overwrite guard. `isSameEntry` never compares names — a case-insensitive volume folds case
  *and* Unicode normalization, and replicating its exact rules is a losing game. Instead it
  counts directory entries for the inode (`nlink === 1` fast path; directories are always their
  own sole entry), and blocks conservatively when the inode has two entries in the directory.
- Every failure path answers conservatively: an unreadable path or directory means "treat the
  target as a distinct existing file and refuse to overwrite".
- Batch conflict detection keeps its case-collapsed keys, but only as a cheap prefilter — every
  candidate is confirmed with `isSameEntry` before a conflict is raised or suppressed, so a
  case-sensitive volume no longer treats a differently-cased source as vacating the target.
- 26 new tests (109 total) run against the real filesystem — hard links, dangling symlinks,
  Unicode normalization — plus mocked-identity tests so both volume behaviours are exercised
  regardless of where the suite runs, and a conditional test that only runs on a genuinely
  case-sensitive volume.

## Screencast

No new captures — this PR changes guard logic only; commands, UI and preferences are unchanged.

## How to Test

No special hardware or credentials needed; the headline bug needs a case-sensitive volume,
which a throwaway disk image provides.

1. `npm ci && npm run dev` inside `extensions/renaming`
2. **Everyday path (unchanged behaviour):** rename any file to a name differing only in case —
   it succeeds, as it always has.
3. **Hard-link false success:** `touch /tmp/foo.txt && ln /tmp/foo.txt /tmp/bar.txt`, then use
   Rename File(s) to rename `foo.txt` to `bar.txt` (or `BAR.txt`). The store version reports
   success while nothing happens; this version reports "already exists".
4. **The headline data-loss bug:** create a case-sensitive volume with
   `hdiutil create -size 10m -fs "Case-sensitive APFS" -volname CS /tmp/cs.dmg && hdiutil attach /tmp/cs.dmg`,
   create `/Volumes/CS/FOO.txt` and `/Volumes/CS/foo.txt` with different contents, then rename
   `FOO.txt` to `foo.txt`. The store version silently replaces `foo.txt`; this version blocks
   with "already exists".
5. `npm test` — 109 tests.

## Checklist

- [x] `npm run lint` passes

### DIFF
--- a/extensions/renaming/CHANGELOG.md
+++ b/extensions/renaming/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Rename Changelog
 
+## [Fix] - {PR_MERGE_DATE}
+
+- Fix a case-only rename silently overwriting a different file on case-sensitive volumes
+- Decide whether a rename target is the same file by inode identity rather than by lowercasing the path, so the overwrite guard and batch conflict detection are correct on any filesystem
+
 ## [Security Fix] - 2026-04-06
 
 - Replace AppleScript-based renaming with native Node.js `fs.rename()` to fix shell injection vulnerability
--- a/extensions/renaming/CHANGELOG.md
+++ b/extensions/renaming/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 - Fix a case-only rename silently overwriting a different file on case-sensitive volumes
 - Decide whether a rename target is the same file by inode identity rather than by lowercasing the path, so the overwrite guard and batch conflict detection are correct on any filesystem
+- Fix a rename silently destroying a symlink at the target path when the symlink's own target was missing
+- Allow renaming a symlink whose target is missing, instead of reporting the source as gone
 
 ## [Security Fix] - 2026-04-06
 
--- a/extensions/renaming/CHANGELOG.md
+++ b/extensions/renaming/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Rename Changelog
 
-## [Fix] - {PR_MERGE_DATE}
+## [Fix] - 2026-08-03
 
 - Fix a case-only rename silently overwriting a different file on case-sensitive volumes
 - Decide whether a rename target is the same file by inode identity rather than by lowercasing the path, so the overwrite guard and batch conflict detection are correct on any filesystem
--- a/extensions/renaming/package-lock.json
+++ b/extensions/renaming/package-lock.json
@@ -259,7 +259,6 @@
         }
       ],
       "license": "MIT",
-      "peer": true,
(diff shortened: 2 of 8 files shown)


0xdhrv (approved):
Looks good to me, approved ✅

PR #29898, opened 2026-08-01
Kesha Voice Kit: detect setup state from structured CLI output

drakulavich:
## Description

Update to the existing **Kesha Voice Kit** extension.

Before starting a dictation session the extension checks that the `kesha` CLI and its speech engine are actually usable. That check decided whether the engine was installed by matching the string `"not installed"` in `kesha status` output — coupling the extension to human-readable wording the CLI could reword at any time, which would have shipped a broken setup check to Store users.

It now reads `kesha status --json`, a machine-readable payload added in kesha CLI 1.25.0.

Compatibility matters here because the CLI is installed by the user from npm/Homebrew and is versioned independently of this extension, so a given machine can have an older one. A CLI too old for the flag prints its usual human output instead (the CLI's arg parser ignores unknown flags), which is the signal to fall back to the previous text check. The fallback is selected by the *kind* of output rather than by whether parsing succeeded — output that is JSON but does not satisfy the contract is reported as a CLI/extension version mismatch instead of being passed to the text match, which would have missed its marker and reported a missing engine as healthy.

Two behaviour improvements come with it:

- An engine that is installed but cannot report its capabilities (corrupt or incompatible binary) is now refused before recording starts, instead of capturing audio that could never be transcribed.
- The finish-setup view names the remedy that matches the actual problem — completing a first install, repairing a broken one, or updating a version-skewed CLI — rather than one message for all three.

`tests/helpers/async.ts` picks up a `waitFor` helper that the updated controller test depends on.

## Checklist

- [x] I ran `npm run build` — it completes cleanly

## Verification

Run in `extensions/kesha-voice-kit`:

- `npm test` — 91 passed
- `npm run lint` — clean
- `npm run build` — built successfully

The probe's behaviour is covered case by case in `tests/kesha-bin-status-json.test.ts`: structured payload for an installed / missing / unusable engine, contract-breaking JSON, an older CLI's text output with and without the marker, and empty or malformed output, which stays permissive so the CLI's own error reporting takes over.

One caveat I would rather state than imply: this change was verified through the automated suite and the build, **not** by driving a dictation session in the Raycast app against all four setup states — reproducing a corrupt-but-present engine and a version-skewed CLI by hand is awkward. The affected code paths are the preflight decision and the copy shown in the setup view; the recording, transcription, and clipboard paths are untouched. Happy to do a manual pass on anything specific a reviewer wants covered.

### DIFF
--- a/extensions/kesha-voice-kit/CHANGELOG.md
+++ b/extensions/kesha-voice-kit/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Kesha Voice Kit Changelog
 
+## [Sturdier setup detection] - {PR_MERGE_DATE}
+
+- Detect whether the engine is installed from the CLI's machine-readable status output instead of matching its wording, so a reworded CLI message can no longer be mistaken for a broken install. Older CLIs keep working through the previous check.
+- Refuse to start a session when the engine is installed but cannot run, instead of recording audio that could never be transcribed, and say how to repair it.
+- Tell a CLI/extension version mismatch apart from a broken engine, so the suggested fix matches the actual problem.
+
 ## [Setup guidance and faster failure feedback] - 2026-07-27
 
 - Rewrite the "kesha CLI not found" message as numbered setup steps, leading with Homebrew so the guidance works without knowing about bun, and include the required `kesha install` step.
--- a/extensions/kesha-voice-kit/CHANGELOG.md
+++ b/extensions/kesha-voice-kit/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Kesha Voice Kit Changelog
 
-## [Sturdier setup detection] - {PR_MERGE_DATE}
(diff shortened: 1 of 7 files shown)


drakulavich:
Pushed `f1c3e49` addressing the Greptile finding.

The classifier returned the same "not structured" signal for valid JSON of the wrong type as for plain human text, so an array or a scalar fell through to the legacy text check, missed its marker, and was reported as a healthy engine. Since an older CLI prints human text and never JSON, a non-object cannot be a version fallback — it is a broken contract and now fails closed like the other contract failures.

Test coverage extended to arrays, numbers, strings, `null` and booleans. `npm test` 91 passed, `npm run lint` clean, `npm run build` succeeds.

0xdhrv (approved):
Looks good to me, approved ✅

PR #29899, opened 2026-08-01
fix(mutedeck): accept `call: "inactive"` so commands work outside a meeting

chadrwalters:
## Description

Every MuteDeck command has been failing with **"Invalid API response format"** whenever no meeting is active.

`isValidMuteDeckStatus` treats `call` as *optional, and if present, only ever `"active"`*:

```ts
export type CallStatus = "active";
...
if (status.call !== undefined && status.call !== "active") return false;
```

MuteDeck 4.x reports `call: "inactive"` when it detects no meeting, so the validator rejects the response, `getStatus()` throws, and — because all four commands call `getStatus()` before doing anything — every command fails. It only breaks *outside* a meeting, which is what made it look intermittent.

Reproduced against a live MuteDeck 4.9.2 instance on `http://localhost:3491/v1/status`:

```json
{"call":"inactive","control":"system","mute":"inactive","record":"disabled",
 "share":"disabled","status":200,"teams_api":"disabled","version":"4.9.2","video":"disabled"}
```

Running the extension's own validator against that payload returns `false` on the `call` check. Delete the `call` key, or set it to `"active"`, and it passes — confirming that single line as the cause.

### Changes

1. **`CallStatus` now admits `"inactive"`**, and the check delegates to the existing `isValidActiveStatus` helper.
2. **`control` is relaxed to warn-and-continue.** It was validated against a closed set (`system | zoom | webex | teams | google-meet`), so the next conference app MuteDeck supports would break every command in exactly the same way. Unknown platforms are now logged and allowed through. A non-string `control` is still rejected.

Nothing else changed — no dependency bumps, no refactoring.

### Verification

A 21-assertion harness covering the live payload and the full status matrix:

- Accepts: live MuteDeck 4.9.2 payload, in-meeting (zoom / teams presenting + recording), legacy payloads with no `call` key, and an unknown future platform.
- Still rejects: garbage `call` values, non-string `control`, missing `mute`, non-numeric `status`, and `null`.

`tsc --noEmit` is clean and `prettier --check` passes against this extension's `.prettierrc`.

## Screencast

Straightforward validation fix with no UI change. Before: every command showed "Failed to Get Status — Invalid API response format" outside a meeting. After: Show Status correctly reports MuteDeck as Running / Not in Meeting, and the toggle commands work.

## Checklist

### DIFF
--- a/extensions/mutedeck/CHANGELOG.md
+++ b/extensions/mutedeck/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## [Fix commands failing when not in a meeting] - {PR_MERGE_DATE}
+
+- Fixed all four commands failing with "Invalid API response format" whenever no meeting was active. MuteDeck 4.x reports `call: "inactive"`, but the status validator only accepted `call` being absent or `"active"`, so `getStatus()` threw before any command could run.
+- Unrecognized `control` platforms are now logged and allowed through instead of failing validation, so a future MuteDeck release adding a conference app cannot break every command the same way. A non-string `control` is still rejected.
+
 ## [Initial Version] - 2025-03-02
 
 - Initial version of the MuteDeck extension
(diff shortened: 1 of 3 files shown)


0xdhrv (approved):
Looks good to me, approved ✅

Issue #16653, opened 2025-01-27
[Slack] Error: Worker terminated due to reaching memory limit


alankritjoshi:
### Extension

https://www.raycast.com/mommertf/slack

### Raycast Version

1.89.1

### macOS Version

14.7

#### Error:
```
Error: Worker terminated due to reaching memory limit: JS heap out of memory

[kOnExit]:worker:313:26
Worker.<computed>.onexit:worker:229:20
```

### Steps To Reproduce

1. Install slack extension
2. Use personal slack token
3. Update token in Slack extension
4. Run any of the commands as part of extension

### Current Behaviour

<img width="746" alt="Image" src="https://github.com/user-attachments/assets/7b1bbc07-1c72-48bd-90f9-2794c4da18d4" />

### Expected Behaviour

I should be able to see the results from the command

PR #29907, opened 2026-08-02
[Slack] Fix Search Emojis memory crash in large workspaces

Naoya-Otani:
## Description

Fixes a crash in the `Search Emojis` command in workspaces with a very large number of custom emojis (tens of thousands):

```
```

**Cause**: The command rendered every custom emoji in the workspace as a `Grid.Item` at once. In workspaces with tens of thousands of custom emojis, building and serializing the React tree exceeds the worker's heap limit. This is the same crash as reported in #16653.

**Changes**:

- Render the emoji grid in slices of 1,000, with a trailing "Show More" item that reveals the next slice on demand.
- Reset the visible slice when the search text changes.
- Search (including AI search) still runs over the full emoji list — only rendering is limited.
- Memoize `Object.entries(data)` and the `allEmojiNames` prompt string, which were recreated on every render, and fix missing dependencies (`emojis`, `executeAISearch`) in the existing `useMemo` hooks.

**Why not the built-in `pagination` API?**

An earlier version of this PR used the `Grid` `pagination` prop. Testing in an affected workspace showed that Raycast's pagination memory safeguard refuses to load further pages:

```
Refusing to paginate further as it could cause the extension to run out of memory. Currently using: 78.077MB.
```

The safeguard estimates the next page's heap growth from the largest recent heap increase, so the initial load of a huge emoji list trips it immediately — and once tripped, pagination stays disabled until the command is reopened. In other words, the built-in pagination gets disabled in exactly the workspaces this PR targets, so the command uses an explicit "Show More" item instead.

**Verified**: In a workspace with tens of thousands of custom emojis, the command now displays, searches, and reveals more emojis via "Show More" without crashing.

**Follow-up**: The same "load and render everything" pattern crashes other commands in large workspaces (e.g. `Search Messages` embeds an `ActionPanel.Submenu` listing every workspace member into every result item). I plan to address that in a separate PR to keep this one focused.

### AsIs

https://github.com/user-attachments/assets/43815d7e-9912-4a70-a771-75083df1723b

### ToBe

https://github.com/user-attachments/assets/321bc2fa-3a77-41d0-8045-59bf3ccfbde2

## Checklist


Naoya-Otani:
Could you please let me know if the "Show More Items" display format is okay?
Also, this is my first OSS pull request!

Naoya-Otani:
SCR-20260802-qpmg

pernielsentikaer (changes requested):
Hi 👋

It looks like a nice enhancement, maybe it's possible to use [Pagination](https://developers.raycast.com/api-reference/user-interface/grid#pagination) directly on the Grid, did you try that?

---

Naoya-Otani:
@pernielsentikaer 

Thanks for taking a look! 

Yes, the first version of this PR did use the `Grid` `pagination` prop
(commit d7195e0, `pageSize: 200`). 

I had to replace it because Raycast disables it in exactly the
workspaces this PR targets.

In an affected workspace, dev mode logs:

    Refusing to paginate further as it could cause the extension to run out of memory. Currently using: 78.077MB.

Looking at the runtime (`Raycast.app/.../api/node_modules/@raycast/api/index.js`), the safeguard is:

    paginationOk = (heapUsed + maxDelta) / 1048576 < resourceLimits.maxOldGenerationSizeMb

where `maxDelta` is the *largest* heap increase over the last 20 render commits, and the check
short-circuits on `if (!paginationOk) return false` — so it never re-enables for the lifetime of
the command. Loading the emoji map is a single large heap jump, which becomes the predicted cost
of every subsequent page.

Two consequences that made me drop the prop:

1. `Grid` renders `pagination={ok ? props.pagination : undefined}`, so once tripped, `onLoadMore`
   never fires again — the docs confirm this happens "regardless of the `hasMore` value".
2. The warning is gated on `NODE_ENV !== "production"`, so in a store build affected users would
   silently see the first page only, with no way to load more and no error.

Server-side paging isn't an option either: `emoji.list` takes only `token` and `include_categories`
— there's no `cursor`/`limit`, so the full map always arrives in one response.

The explicit "Show More" item doesn't touch the `pagination` prop, so it isn't affected by the
safeguard. Happy to switch back if you'd prefer the built-in pagination and think the heuristic
should be addressed on the Raycast side instead — this looks like it affects other extensions too
(_e.g. the `zoo` extension's README tracks the same message as an open issue_).

pernielsentikaer:
Thanks for flagging this issue. I created an internal ticket for us to look into this in the near future. Until there is a fix, I'll see this as the best solution. I will get back to you as soon as there is a resolution.

pernielsentikaer (approved):
Looks good to me, approved 🔥

Thanks for the thorough write-up on why Grid pagination doesn't work here — Show More is the right approach for this fix.

Naoya-Otani:
Thank you so much for the review and for merging this!

And thank you for creating an internal ticket to look into the pagination heuristic — I really appreciate you taking the time to dig into the root cause on the Raycast side. Happy to help test or provide more details from an affected workspace whenever that work picks up.

As mentioned above, I'll open a separate PR for the related `Search Messages` crash to keep things focused.

PR #29914, opened 2026-08-02
Add flip-scroll extension

oh-roman:
## Description

Adds Flip Scroll — a single no-view command that toggles macOS natural scrolling,
for switching between trackpad and mouse without opening System Settings.

### On the approach

The command calls `setSwipeScrollDirection` from the private `PreferencePanesSupport`
framework via JXA. Since that is a private API, here is why:

- `defaults write -g com.apple.swipescrolldirection` alone does not take effect until
  the next login. It appears to work only while System Settings is open, because that
  app applies the change for you.
- `activateSettings -u` does apply it, but it re-applies the entire global preference
  domain including the keyboard context. That re-registers the system symbolic hotkeys
  and can hand ⌘Space back to Spotlight on every toggle.
- `setSwipeScrollDirection` is the function the Trackpad pane itself calls and changes
  only this one setting. No window opens, no Accessibility permission is required.

Verified on macOS 26 (Tahoe).

## Screencast

The command has no UI — it runs and shows a HUD. Screen recording below shows a page
scrolled in one direction, the hotkey pressed, and the same gesture scrolling the
opposite way.

## Checklist


0xdhrv (commented):

- [Change Scroll Direction](https://www.raycast.com/gregoralbrecht/change-scroll-direction) by @gregoralbrecht

Both are a single no-view command that toggles macOS natural scrolling so you can switch between trackpad and mouse without digging through System Settings.

- **Change Scroll Direction**, a no-view toggle between "natural" and "mouse-friendly"
- Immediate apply via System Settings UI automation, including a Tahoe-specific script

- Applies the same preference through `setSwipeScrollDirection` instead of opening System Settings
- No Accessibility permission, and no full-domain `activateSettings` refresh that can steal ⌘Space
- A HUD that reports the new state (trackpad vs mouse)

Could we consider contributing the live PreferencePanesSupport toggle to the existing extension instead?

> This would help avoid duplication and keep scroll-direction toggling consolidated in one extension.  

oh-roman:
Thanks for the detailed review — agreed, the primary workflow does overlap, so I'm happy
to contribute the improvements to Change Scroll Direction instead.

What I'd bring over:

- Replace the System Settings UI automation with a direct `setSwipeScrollDirection` call
  from PreferencePanesSupport — the same function the Trackpad pane itself calls. It
  writes the preference and applies it live in one step.
- That removes `scripts.ts` with its per-version AppleScripts (Tahoe / Ventura / older)
  and the `run-applescript` dependency, so there's no UI hierarchy to re-fix on each
  macOS release, and no Accessibility permission needed.
- Read current state via `defaults read -g com.apple.swipescrolldirection` (treating a
  missing key as "on", which is how macOS behaves on a fresh account) so the HUD can
  report the resulting state — "Natural scrolling ON (trackpad)" / "OFF (mouse)" —
  instead of "Scroll direction changed".
- The existing jest suite covers `fetchAppleScript`, which goes away with the scripts;
  I'll rework the tests accordingly.

Heads up that this replaces the command's implementation end to end, so expect a rewrite
of `changeScrollDirection.ts` rather than a small patch. I'd add myself to `contributors`.

One separate, entirely optional question for @gregoralbrecht: I drew an icon for this
while working on my own version, and I'd be glad to offer it if you like it — happy to
post it here first so you can see it before deciding. Completely your call, and a "no"
changes nothing about the code above; I'll send that either way.

Would you prefer I close this PR and open a fresh one against
`extensions/change-scroll-direction`, or repurpose this branch?

0xdhrv:
> Thanks for the detailed review — agreed, the primary workflow does overlap, so I'm happy to contribute the improvements to Change Scroll Direction instead.
> 
> What I'd bring over:
> 
> * Replace the System Settings UI automation with a direct `setSwipeScrollDirection` call
>   from PreferencePanesSupport — the same function the Trackpad pane itself calls. It
>   writes the preference and applies it live in one step.
> * That removes `scripts.ts` with its per-version AppleScripts (Tahoe / Ventura / older)
>   and the `run-applescript` dependency, so there's no UI hierarchy to re-fix on each
>   macOS release, and no Accessibility permission needed.
> * Read current state via `defaults read -g com.apple.swipescrolldirection` (treating a
>   missing key as "on", which is how macOS behaves on a fresh account) so the HUD can
>   report the resulting state — "Natural scrolling ON (trackpad)" / "OFF (mouse)" —
>   instead of "Scroll direction changed".
> * The existing jest suite covers `fetchAppleScript`, which goes away with the scripts;
>   I'll rework the tests accordingly.
> 
> Heads up that this replaces the command's implementation end to end, so expect a rewrite of `changeScrollDirection.ts` rather than a small patch. I'd add myself to `contributors`.
> 
> One separate, entirely optional question for @gregoralbrecht: I drew an icon for this while working on my own version, and I'd be glad to offer it if you like it — happy to post it here first so you can see it before deciding. Completely your call, and a "no" changes nothing about the code above; I'll send that either way.
> 
> Would you prefer I close this PR and open a fresh one against `extensions/change-scroll-direction`, or repurpose this branch?

Thanks for understanding ❤️ 

You can create a new PR for the existing extension, and I advise to backup this branch if you want to

PR #29915, opened 2026-08-02
Add shortcut-vault extension

syedhy:
## Checklist


syedhy:
@greptile review

syedhy:
@greptile review

syedhy:
@greptile review

pernielsentikaer (commented):
Hi 👋

I'm not sure what this extension is for that `Search Menu Bar Items` doesn't already do. We already have similar extensions in the store, and it might be too niche to add that much value

syedhy:
This extension is especially useful for Raycast power users who have accumulated dozens—or even hundreds—of shortcuts across different apps, websites, and commands. Instead of trying to remember what each shortcut does or where it belongs, you can keep everything organized in one central, searchable place.

You can quickly look up a shortcut, see what it’s for, which app or website it belongs to, and access the information you need without digging through scattered notes or trying to remember it.

For anyone who relies heavily on shortcuts, this provides a fast, simple way to build a personal shortcut library and keep everything organized and easy to find.

syedhy:
This extension is especially useful for Raycast power users who have
accumulated dozens—or even hundreds—of shortcuts across different apps,
websites, and commands. Instead of trying to remember what each shortcut
does or where it belongs, you can keep everything organized in one central,
searchable place.

You can quickly look up a shortcut, see what it’s for, which app or website
it belongs to, and access the information you need without digging through
scattered notes or trying to remember it.

For anyone who relies heavily on shortcuts, this provides a fast, simple
way to build a personal shortcut library and keep everything organized and
easy to find.

pernielsentikaer:
Hi @syedhy 👋

Thanks for the extra context — a personal shortcut library (what a shortcut does, which app it belongs to) is a different job from Search Menu Bar Items, which invokes the current app’s live menu.

A few things still need to change before this can go to the Store:

- `src/lib/storage.ts` can create `storage.lock` and then throw on the post-loop timeout, before the `finally` release. Later commands then fail to save. Only throw when the lock was never acquired, and always release if you own it.
- Export treats a failed read as an empty vault (“Nothing to export”). Show the error and a Retry action when `getCustomShortcuts()` throws.
- Recapture the Export screenshot without the `/Users/hyder/...` path.
- Seven commands overlap. Search Shortcuts already filters by source, scope, and owner — Search Default, Search Custom, and Manage Custom can be filters/actions on one list.
- Keep `CHANGELOG.md` user-facing (drop the lint/typecheck/“100% clean verification” notes).
- `navigationTitle` on the root Add Shortcut form duplicates the command title. Set it only on the nested Edit screen.
- Validate bundled `sourceUrl` values with the same http/https check you already use for imports, before `OpenInBrowser`.

---

syedhy:
I’ve tried to make the changes you requested in the past few days and i hope this is enough for the approval of the extension , if you have any more suggestions , please comment

syedhy:
@pernielsentikaer Just a reminder for the review of the extension after the recommended changes

pernielsentikaer (changes requested):
Hi @syedhy 👋

Thanks for the updates and for explaining the personal-library workflow.

[Shortcut Library](https://www.raycast.com/rhrabun/shortcut-library) by @rhrabun already supports searching and managing personal shortcuts with JSON import/export. Your bundled reference collection and modifier-aware search add useful differences, but we’d prefer bringing compatible improvements into the existing extension.

Would you be open to contributing those features there? If combining them would be impractical, could you explain why? 🙂

---

syedhy:
Hi Per 👋

Thanks for taking the time to review the PR and for pointing this out.

I understand that Shortcut Library already provides some overlapping
functionality, particularly around personal shortcuts and JSON
import/export. However, my intention with this project is for it to remain
a standalone extension rather than becoming a set of features within
Shortcut Library.

The goal is to make this a more complete shortcut-management/reference tool
with its own workflow. Some of the differences I’m aiming for include:

   - *Scope-aware shortcuts* — distinguishing between global,
   application-specific, and web-app shortcuts.
   - *Multiple tags and categorization* — allowing shortcuts to be
   organized and filtered in more flexible ways.
   - *Modifier-aware search* — making it easier to find shortcuts based on
   their actual key combinations.
   - *A bundled reference library* — providing a curated collection of
   shortcuts out of the box rather than relying entirely on user-added
   shortcuts.
   - *Custom shortcut management* — allowing users to add and organize
   their own shortcuts alongside the reference collection.
   - *Import/export* — making personal shortcut collections portable and
   allowing users to back up or transfer them.
   - *A different overall UX and purpose* — the intention is for this to
   function as a standalone shortcut reference and management tool rather than
   simply extending an existing personal-shortcut workflow.

Because of the combination of these features and the overall direction of
the project, I’d prefer to keep it as a standalone extension rather than
contributing the functionality directly to Shortcut Library.

That said, I completely understand the concern about overlapping
functionality. I’d be happy to clarify the differences further or make
changes to ensure the distinction between the two extensions is clear.

Thanks again for taking the time to review it! 😊

rhrabun:
Hi guys,
If I may throw my 2 cents to this discussion:
Seems like @syedhy has more plans and features in mind for this kind of extension that I do for mine - so I could propose that my Shortcuts Library gets replaced or taken over with this one if that's a possibility.
I've developed mine just because I needed the json import and ability to fuzzy search shortcuts, so, if this one support (and also extends) that functionality, I'm happy to make room for that :)

syedhy:
Yes, actually I am developing an app that would integrate with this
extension so the users can interact with the app and create shortcuts or
Use the multiple features that I’m adding to the app and the shortcuts will
be synced with the extension so it’s a cohesive experience so that’s why
it’s not possible for me to just merge my extension with another or build
upon another architecture , thanks for the idea @rharbun

pernielsentikaer (changes requested):
So, you’re building a real app or service around the extension. If so, it makes sense to keep everything as it is now. Please confirm 🙂

---

syedhy:
> So, you’re building a real app or service around the extension. If so, it makes sense to keep everything as it is now. Please confirm 🙂
> 

Yes i am , but i would like the extension to be on store before my app goes into production if that is possible

pernielsentikaer (changes requested):
Hi @syedhy 👋

Thanks for confirming the companion-app plans and addressing the earlier feedback.

I found a few remaining issues:

- In `cross-process-mutex.ts`, heartbeat writes can briefly leave `pid.txt` empty, which another command immediately treats as an abandoned lock. I reproduced a second process entering while the first still held its critical section. Please publish heartbeat updates atomically and make recovery safe against concurrent contenders.
- If a process exits between creating `storage.lock` and writing `pid.txt`, later commands repeatedly time out. Please add safe recovery for an aged, incomplete lock.
- A vault containing 1,001 shortcuts exports successfully, but its backup cannot be imported because imports stop at 1,000. Could you ensure every supported export can be restored, including the file-size boundary? A round-trip regression should cover:
  ```ts
  const shortcuts = Array.from({ length: 1001 }, (_, i) => ({
    ...EXAMPLE_EXPORT.shortcuts[0],
    id: String(i),
  }));
  assert.doesNotThrow(() => validateExportFile(createExportFile(shortcuts)));
  ```
- Literal `+` searches currently become an empty query. Please distinguish the plus key from combination separators and add a regression test.

---

syedhy:
Hi @pernielsentikaer 👋

Thanks again for the review! All remaining feedback and bot requirements have been resolved and verified:

- ✅ **Atomic Mutex Heartbeat & Contender-Safe Recovery**: Heartbeat updates now write to a unique temporary file inside the lock directory and use atomic `fs.renameSync` to publish `pid.txt`, preventing empty file reads. Stale lock recovery uses atomic directory renaming so concurrent contenders never collide. Added safe recovery for aged, incomplete lock directories (`storage.lock` without `pid.txt` older than 15s).
- ✅ **1,001+ Shortcut Export/Import Round-Trip**: Raised capacity to 10,000 shortcuts across export and import formats, added the requested 1,001-shortcut round-trip regression test, and enforced symmetric 6 MB byte-size limits across both file exports and clipboard exports (`serializeExportFile`).
- ✅ **Literal `+` Searches**: Distinguish combination separators from the literal plus key. Standalone `+`, `cmd +`, `cmd + +`, and `zoom +` now properly match shortcuts with the plus key (e.g. Zoom In) instead of collapsing into an empty query, with regression test coverage.
- ✅ **ESLint v9 Config**: Wrapped `eslint.config.mjs` with the repository-required `defineConfig` helper from `eslint/config`.

The PR is marked ready for your review! 🔥

@greptile review

pernielsentikaer (changes requested):
Hi @syedhy 👋

Thanks for the fixes! The atomic heartbeat update is in place, but the [stale-lock concurrency issue](https://github.com/raycast/extensions/pull/29915#discussion_r3699670063) still has a remaining race.

In `tryReclaimStaleLockDir()`, a contender can rename and delete a newly acquired lock using an earlier stale-lock check. I reproduced this with a deterministic filesystem-interleaving probe that inserts another contender’s recovery and acquisition between the check and rename.

Could you make stale recovery safe against this replacement race? Atomic rename alone doesn’t ensure the directory is still the stale one you inspected. Regression coverage for this interleaving would be helpful too 🙂

---

syedhy:
Hi @pernielsentikaer 👋

Thanks for catching that subtle replacement race! We have addressed it along with the suggestions flagged by Greptile, backed by deterministic regression coverage:

### 1. Stale-Lock Replacement Race Safety (`src/lib/cross-process-mutex.ts`)
- **Snapshot Identity Verification**: When inspecting a lock, `CrossProcessMutex` now captures a `StaleLockSnapshot` containing the directory inode (`dirIno`), file inode (`fileIno`), and exact `pid.txt` content.
- **Pre-Check**: Before initiating a rename, `tryReclaimStaleLockDir(snapshot)` verifies that `this.lockDir` still matches the inspected snapshot. If another contender recovered and acquired the lock in the interim, reclaim immediately aborts without touching the directory.
- **Post-Rename Defense-in-Depth**: If an atomic rename succeeds but the resulting directory fails the snapshot match, it is immediately restored (`fs.renameSync(staleDir, this.lockDir)`) and aborted.
- Applied the same snapshot verification to `releaseIfOwned()` and initial write-failure cleanups.

### 2. Deterministic Interleaving Regression Coverage (`scripts/test-storage-mutex.ts`)
- Added an `onBeforeReclaimForTesting` test hook to deterministically probe filesystem interleavings between the stale-lock inspection and reclaim attempt:
  - **Test 9**: Simulates an interleaved contender recovering the stale lock and actively entering its critical section; verifies that the inspecting contender refuses to rename or delete the new active lock, waits for the critical section to complete, and only then acquires cleanly.
  - **Test 10**: Validates the same replacement-race safety for aged incomplete locks without `pid.txt`.

### 3. Greptile Review Suggestions
- **Clipboard Backup Payload Boundary**: Unified file and clipboard export paths through `serializeExportFile`, enforcing the symmetric 6 MB file-size limit on clipboard backups so an export copied to the clipboard cannot exceed the import payload ceiling.
- **ESLint Config**: Wrapped the flat config in `eslint.config.mjs` with the standard ESLint v9 `defineConfig` helper from `eslint/config`.

All checks (`npm run verify`: data validation, typecheck, unit tests, ESLint, Prettier, and Raycast build) pass cleanly with 0 errors and 0 warnings.

pernielsentikaer (changes requested):
Hi @syedhy 👋

Thanks for the fixes! Two issues remain:

- The [stale-lock race](https://github.com/raycast/extensions/pull/29915#discussion_r3699670063) still reproduces when replacement happens **after** `isSnapshotMatch()` but before `renameSync()` in `src/lib/cross-process-mutex.ts`. Moving the live lock allows another command to acquire it before restoration. Could you make acquisition and recovery safe against this interleaving? The new tests currently inject replacement before the snapshot check.
- Duplicating a valid 512-character command name creates a 517-character name. Export succeeds, but that backup cannot be imported. Please keep generated names within the shared limit in `src/lib/storage.ts`, for example:
  ```ts
  commandName: `${existing.commandName.slice(0, 512 - " Copy".length)} Copy`,
  ```

---

syedhy:
> Hi @syedhy 👋
> 
> Thanks for the fixes! Two issues remain:
> 
> * The [stale-lock race](https://github.com/raycast/extensions/pull/29915#discussion_r3699670063) still reproduces when replacement happens **after** `isSnapshotMatch()` but before `renameSync()` in `src/lib/cross-process-mutex.ts`. Moving the live lock allows another command to acquire it before restoration. Could you make acquisition and recovery safe against this interleaving? The new tests currently inject replacement before the snapshot check.
> * Duplicating a valid 512-character command name creates a 517-character name. Export succeeds, but that backup cannot be imported. Please keep generated names within the shared limit in `src/lib/storage.ts`, for example:
>   ```ts
>   commandName: `${existing.commandName.slice(0, 512 - " Copy".length)} Copy`,
>   ```
> 
> I’ll convert this PR into a draft after submitting this review. Please press Ready for review when it’s ready and we’ll have a look 😊

Hi @pernielsentikaer 👋

Thanks for the follow-up review! Both issues are resolved and verified in the latest update:

1. **Recovery & Acquisition Synchronization (`src/lib/cross-process-mutex.ts`)**:
   - **Reclaim Barrier**: Added an exclusive reclaim barrier (`${this.lockDir}.reclaim`) acquired prior to inspecting and moving stale lock candidates.
   - **Acquisition Synchronization**: New command acquisitions in `runExclusive` check `isReclaimInProgress()` and back off while any recovery or restoration is taking place.
   - **Guaranteed Safe Restoration**: If a live lock replacement occurs right before `renameSync()`, the destination `this.lockDir` remains protected by the barrier so another command cannot acquire it in the interim, ensuring the live lock is restored cleanly and without interference.
   - **Interleaving Test Coverage (`scripts/test-storage-mutex.ts`)**: Added an `onBeforeRenameForTesting` hook (firing right between `isSnapshotMatch()` and `renameSync()`) and updated Test 9 to prove that concurrent commands are blocked during transition and the replaced live lock is preserved intact.

2. **Duplicated Command Name Length Limit (`src/lib/storage.ts`)**:
   - Duplicating a command now truncates the base name to keep the result within the 512-character ceiling:
     ```ts
     commandName: `${existing.commandName.slice(0, 512 - " Copy".length)} Copy`,
     ```
   - Added regression coverage in `scripts/test-import-export-format.ts` ensuring duplicated 512-character names pass import/export validation.

All checks (`npm run verify`: data validation, typecheck, tests, ESLint, Prettier, and Raycast build) are passing cleanly with 0 errors and 0 warnings.

PR #29924, opened 2026-08-03
Update dev-servers extension

tavlean:
## Menu bar restart fixed, plus icon and row polish

- **Restarting from the menu bar brings the server back.** Restart used to kill the server and stop there: Raycast unloads a menu bar command shortly after its menu closes, and the respawn was still in flight when that happened. Restart now hands the whole job to the dashboard, the same way menu bar starts already work, so you also get what starts get: a live "Restarting…" row, and a diagnosed failure row if the server never comes back, instead of silence.
- **The Kill action's icon now reads as a trash can.** The old outlined glyph with its floating handle looked like a jar with a cork once tinted red; the solid-body variant from the same icon family keeps the trash silhouette at menu size.
- **Every menu icon now draws completely.** Raycast's SVG renderer skips `line`, `polyline`, and `polygon` elements, so Copy Port and Open in Terminal had no icon at all, the trash was missing its lid, and the Restart arrows lost their heads. All glyphs are now built from paths, which render fully.
- **Projects without a usable favicon get their own color in the menu bar.** When a project's favicon only exists as an SVG (which the menu bar can't render as an image), the fallback glyph is now tinted with the favicon's dominant color instead of the framework color, so a row is recognizable per project rather than per framework. The framework tint remains only for projects with no favicon at all.
- **Killing from the menu bar confirms itself.** A small system HUD says "Killed <project>" the instant you click (and "Killed both/all N servers" for the per-project kill). The kill really is instant; the SIGKILL cannot be refused. The menu bar count, which used to be the only signal and lagged by several seconds, now updates a few seconds sooner as well.
- **Branch tags appear only when they mean something.** A project running a single worktree no longer repeats "· main" on every row. The branch shows, main included, when two worktrees of the same project run side by side, or when two entries in the Start section share a project name.
- **Fixed: restarting one of two servers running from the same folder could kill the wrong one.** The menu bar's Restart now names the exact server that was clicked, so a sibling server sharing the project folder is left alone.
- **Fixed: a dev server on a deliberately chosen high port (49152 and up) could vanish from every surface.** The phantom-server cleanup now hides only processes it knows to be helpers (`workerd`), instead of anything that looked abandoned. A server this extension starts is handed to the system the moment the command unloads, so a legitimate server looked exactly like a leftover.
- **Fixed: a project starting on a port another project used earlier could inherit that project's favicon, permanently.** Favicons are now cached per project as well as per port.
- **Favicon lookups can no longer be redirected off the dev server.** Responses served from elsewhere were already refused; now the request itself stops before a redirect leaves the server's origin, so no other host is ever contacted.

## Checklist


0xdhrv (approved):
Looks good to me, approved ✅

PR #29930, opened 2026-08-03
fix(espanso): resolve the binary by probing absolute paths

kud:
## Description

Fixes all Espanso commands failing with `Failed to run 'espanso path'` when the **Espanso Binary Path** preference is left empty — which is the default.

`getEspansoCmd()` fell back to a bare `"espanso"` and relied on `PATH` to resolve it. That cannot work inside Raycast: Raycast is launched by launchd, so its Node subprocesses inherit launchd's `PATH`, not the user's shell `PATH`. Verified on an affected machine by reading the live process environment:

```
$ ps eww -p $(pgrep -x Raycast) | tr ' ' '\n' | grep '^PATH='
PATH=/usr/bin:/bin:/usr/sbin:/sbin
```

Neither `/opt/homebrew/bin` nor `/usr/local/bin` is present, so a bare `espanso` is unresolvable regardless of how the user's shell is configured. `execSync` goes through `/bin/sh -c`, which reads no shell rc files, so nothing restores the missing entries.

The binary is now located by probing absolute candidates:

```
/opt/homebrew/bin/espanso                          (Homebrew, Apple Silicon)
/usr/local/bin/espanso                             (Homebrew, Intel)
/Applications/Espanso.app/Contents/MacOS/espanso   (app bundle)
```

The app bundle is last but is the one that always exists — the Homebrew entries are symlinks into it — so this also covers installs without Homebrew. Homebrew paths are probed first so that a deliberately brew-installed espanso wins over a stale `.app`.

Setting the preference still overrides the probe entirely, and surrounding whitespace in it is now trimmed. The bare-name fallback is kept as a last resort so that a genuinely missing install produces a clear error rather than a silently wrong path.

### Verification

- `npm test` — 18 passed (11 existing, 7 new covering preference precedence, whitespace trimming, both Homebrew prefixes, the app-bundle fallback, candidate ordering, and the no-install case)
- `npm run typecheck` — clean
- `npm run lint` — clean
- `npm run build` — succeeds
- Runtime check under a stripped `PATH=/usr/bin:/bin:/usr/sbin:/sbin`: the old bare-name call fails with `Command failed: espanso path`, while the probed path returns the real `Config:` / `Packages:` / `Runtime:` output that `getEspansoConfig()` parses.

## Screencast

Not included — this is a behavioural fix with no UI change. The observable difference is that the extension's commands work on a default install instead of erroring, without the user first having to discover and fill in the binary-path preference.

## Checklist

### DIFF
--- a/extensions/espanso/CHANGELOG.md
+++ b/extensions/espanso/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Espanso Changelog
 
+## [Fix Espanso binary not being found] - {PR_MERGE_DATE}
+
+### Bug Fixes
+
+- Fixed all commands failing with `Failed to run 'espanso path'` when the Espanso Binary Path preference was left empty. Raycast is launched by launchd and so inherits `PATH=/usr/bin:/bin:/usr/sbin:/sbin`, which contains neither Homebrew prefix — a bare `espanso` could never resolve there. The binary is now located by probing `/opt/homebrew/bin/espanso`, `/usr/local/bin/espanso`, and the `Espanso.app` bundle, so the extension works out of the box on both Apple Silicon and Intel, and with a Homebrew-less install. Setting the preference still overrides the probe, and surrounding whitespace in it is now ignored.
+
 ## [Fix Search Matches memory usage] - 2026-05-28
 
 - Reduced Search Matches memory usage by loading replacement previews only for the selected match.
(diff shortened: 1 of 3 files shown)


0xdhrv (approved):
Looks good to me, approved ✅

PR #29936, opened 2026-08-03
Kesha Voice Kit: make the silence auto-stop actually fire

drakulavich:
## Description

Update to the existing **Kesha Voice Kit** extension. Follow-up to #29898.

The idle auto-stop never fired. Whether anyone was speaking was decided by comparing the microphone's peak against `0.0001` — that is **−80 dBFS**, a test for *digital* silence, not for speech. Every live microphone clears it, so the meter state was permanently "signal", the idle timer was reset on every sample, and each dictation session ran to its full maximum duration unless the user pressed Stop.

Classification now compares rms against `0.01`, which sits between a quiet room and speech. The signal percentage also moves off `sqrt(peak)`, which inflated the bottom of the scale and drew a silent room at about 30%.

## Measured, not assumed

Recorded from a real microphone using the extension's own meter.

**Empty room, 101 samples over 12 s:**

| threshold | samples counted as speech | longest unbroken silence |
|---|---|---|
| old, `peak > 0.0001` | **101/101 (100%)** | 0 s — timer never advances |
| new, `rms > 0.01` | 7/101 | 11.2 s and counting |

**A person speaking normally, 284 samples over 30 s** — about 20 s of speech with natural pauses, then 10 s of deliberate silence:

| period | longest unbroken silence | auto-stop (needs 45 s) |
|---|---|---|
| while speaking | **4.0 s** | does not fire |
| after speaking | 10.0 s and counting | fires as intended |

Levels: speech p50 0.0020 / p90 0.0139 / max 0.0361; silence p50 0.0007 / max 0.0056. The threshold sits in the gap — `silence max 0.0056` < `0.0100` < `speech max 0.0361`.

Worth noting the mechanism works because speech is *uneven*, not because it is loud: a live voice's median actually sits below the threshold, and it is the louder syllables crossing it that keep resetting the timer.

## Trade-off

In a room with a steady noise source above the threshold, everything reads as speech and the auto-stop will not fire there. That is the same as the current released behaviour, so no regression, and it is the safe direction to be wrong in — the alternative would be cutting a speaker off mid-sentence. An adaptive noise floor was prototyped and dropped: it could do that, and distinguishing steady speech from steady noise by level alone is not possible.

## Checklist

- [x] I ran `npm run build` — it completes cleanly

## Verification

In `extensions/kesha-voice-kit`: `npm test` 93 passed, `npm run lint` clean, `npm run build` succeeds.

As with #29898, this was verified through the automated suite, the build, and the microphone measurements above rather than by driving the full recording flow in the Raycast app for every state. The change is confined to how meter samples are classified; recording, transcription and clipboard paths are untouched.

### DIFF
--- a/extensions/kesha-voice-kit/CHANGELOG.md
+++ b/extensions/kesha-voice-kit/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Kesha Voice Kit Changelog
 
-## [Silence auto-stop now works] - {PR_MERGE_DATE}
+## [Silence auto-stop now works] - 2026-08-08
 
 - Stop a dictation session after the idle period as documented. The check for whether anyone is speaking used a level 800x below any real room's noise floor, so every microphone counted as speech and the session always ran to the full maximum duration unless stopped by hand.
 - Show a signal level that matches what you hear: a silent room read about 30% before.
--- a/extensions/kesha-voice-kit/CHANGELOG.md
+++ b/extensions/kesha-voice-kit/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Stop a dictation session after the idle period as documented. The check for whether anyone is speaking used a level 800x below any real room's noise floor, so every microphone counted as speech and the session always ran to the full maximum duration unless stopped by hand.
 - Show a signal level that matches what you hear: a silent room read about 30% before.
+- Measure the level from the loudest input channel rather than the average of all of them, so a microphone plugged into one input of a multi-channel audio interface is no longer read as 30% quieter than it is and cut off mid-sentence.
 
 ## [Sturdier setup detection] - 2026-08-03
 
--- a/extensions/kesha-voice-kit/CHANGELOG.md
+++ b/extensions/kesha-voice-kit/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Kesha Voice Kit Changelog
 
+## [Silence auto-stop now works] - {PR_MERGE_DATE}
+
+- Stop a dictation session after the idle period as documented. The check for whether anyone is speaking used a level 800x below any real room's noise floor, so every microphone counted as speech and the session always ran to the full maximum duration unless stopped by hand.
+- Show a signal level that matches what you hear: a silent room read about 30% before.
+
 ## [Sturdier setup detection] - 2026-08-03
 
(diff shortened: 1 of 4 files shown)


0xdhrv (commented):
Hi @drakulavich 👋

Thanks for the contribution 🔥

---

drakulavich:
Thanks @0xdhrv — Greptile's finding was valid and is fixed in 6908e88.

**The bug.** The Swift tap summed squared samples across every channel and divided by `channelCount * frameCount`, so the rms it reported was pooled across channels. A microphone on one input of a multi-channel device is therefore attenuated by `sqrt(channelCount)` before the threshold sees it:

| channels | factor | measured speech p90 `0.0139` → | threshold |
|---|---|---|---|
| 1 | 1.000 | 0.0139 ✅ | 0.01 |
| 2 | 0.707 | **0.0098** ❌ | 0.01 |
| 4 | 0.500 | **0.0069** ❌ | 0.01 |

Greptile's arithmetic checks out exactly (`0.012 → 0.0085`). It does not stop a session instantly — 45 s of continuous sub-threshold is still required — but the headroom the original measurements relied on is gone, and on a 4-channel aggregate device there is none left.

**The fix.** The tap now reports one rms per channel and the reduction moved into TypeScript, next to the threshold it feeds:

```swift
channelRms.append(sqrt(sum / Float(frameCount)))   // per channel, not pooled
```
```ts
export function loudestChannelRms(levels: unknown): number { ... }
```

That placement is deliberate — it puts the decision under unit test, which answers the second half of Greptile's note ("does not cover multichannel attenuation"). `peak` was already the maximum across channels, so `percentFromPeak` and the displayed level are untouched, and mono inputs behave exactly as before.

**Verification.**

- `npm test` 95 passed (was 93; +2 covering multichannel attenuation and non-numeric channel data), `npm run lint` clean, `npm run build` succeeds.
- The new test is not decorative: restoring the pooled average makes it fail with `expected 0.00848528137423857 to be close to 0.012` — the precise value Greptile predicted.
- The Swift script `swiftc -typecheck`s clean and was run against a real microphone for 20 s: 183 samples, valid `{"channelRms":[...],"peak":...}` on every line, quiet room classified as speech 1/183. Consistent with the 7/101 in the PR description.

I could not test a physical multi-channel interface — I only have a mono built-in mic — so the multichannel path is verified by the arithmetic and the unit tests rather than by hardware. Flagging that explicitly rather than implying otherwise.

Marking ready for review.

0xdhrv (approved):
Looks good to me, approved ✅

PR #29941, opened 2026-08-03
Redmine: Open Issues and Created by Me commands, issue editing, project filter, AI Extension

julesvbertolino:
## Description

This supersedes #29458. There I had submitted my work as a separate forked extension (`redmine-tickets`); @pernielsentikaer asked me to enhance the existing [Redmine](https://www.raycast.com/jwickers/redmine) extension by @jwickers instead, so this PR ports everything into `extensions/redmine`. I'll close #29458 once this is up.

The extension keeps its identity — same `name`, `title`, `author`, icon and preferences. I only added myself to `contributors`.

### Commands

- **Open Issues** (new) — every open issue across projects, most recently updated first.
- **Created by Me** (new) — the open issues you reported.
- **My Issues** — the existing command, now scoped to *open* issues. Its title changed from "Show My Issues", but the command `name` is still `issue`, so existing hotkeys and aliases keep working.

All three share the same search bar: typing now runs Redmine's own full-text search (`/search.json`, subject + description) rather than only filtering the already-loaded page, and a project dropdown narrows the list to a single project (remembered via `storeValue`).

### Issue detail view

Pressing ⏎ on an issue opens a `Detail` with its description, metadata and comments, plus keyboard actions to add a comment (`⌘N`) and change status (`⌘S`), assignee (`⌘A`) or priority (`⌘P`) via `PUT /issues/:id.json`. Status options come from the issue's `allowed_statuses` when the instance provides them, falling back to `/issue_statuses.json`.

### AI Extension

A read-only `Search Issues` tool, so `@redmine` can full-text search issues (open or closed) and filter by status, assignee or creator. Writes are deliberately not exposed to the AI — editing stays in the commands above.

### Fixes and maintenance

- `orangeIssues` and `blueIssues` were only applied inside an `if (prefs.redIssues)` block, so both preferences silently did nothing when "Red Issues Priorities" was left empty. The three mappings are now built independently.
- `getRedmineUrl` ignored its `domain` argument in the fallback branch and read the pref directly.
- Replaced `node-fetch` with the built-in `fetch`, and moved from a hand-written preferences interface to `getPreferenceValues<Preferences>()`.
- Moved a `showToast` call out of render into a `useEffect`.
- Bumped `@raycast/api` and the dev toolchain; `dist` was missing from `.gitignore`.

### Screenshots

I added one screenshot of the list view with the project filter. `Screenshot-01.png` is @jwickers' original and I left it in place. Happy to add shots of the detail view and the AI Extension in a follow-up commit here if you'd like more.

## Checklist

- [x] I ran `npm run lint` and fixed the issues


julesvbertolino:
Thanks for the review — all three were valid, fixed in 4e5a149d0:

- **Stale search results** — overlapping requests while typing or switching project could let an earlier response overwrite the current results, error or loading state. The effect now marks its response outdated on cleanup and ignores it.
- **Empty `allowed_statuses`** — an empty array means the workflow permits no transition, so falling back to every status on the instance offered updates Redmine would reject with 422. The fallback now only triggers when the field is absent (Redmine < 5.0).
- **Assignee picker** — member groups were discarded, and there was no way to clear an assignee. Groups are now included as candidates and the submenu has an "Unassigned" entry (`assigned_to_id: ""`).

`npm run build` and `npm run lint` both pass.

julesvbertolino:
Good catch, this one's real too — fixed in 1b5531d0d.

**Assignee list stops at 100** — `getProjectMembers` only requested `limit=100` with no pagination, so a project with more than 100 combined user/group memberships silently truncated the Change Assignee picker. It now loops through pages via `offset`, stopping once a short page comes back or `total_count` is reached.

`npm run build` and `npm run lint` both pass.

jwickers (approved):
Looks good to me, extension built and all options ran successfully

0xdhrv (approved):
Looks good to me, approved ✅

PR #29947, opened 2026-08-03
Add spongecase extension

evnm:
## Description

This extension provides a single "Copy Spongecased Text to Clipboard" command which

1. prompts the user to enter a string,
2. when enter is hit, converts the string to spongecase (i.e. aLtErNaTiNg cApS), and
3. copies the result to the clipboard.

I'm aware that the Change Case extension exists. I just find that it doesn't match my typical use case. Rather than operating on selected or already-copied text and requiring me to navigate a menu within the command, I want a oneshot command to quickly convert text to e.g. paste into a chat message.

## Screencast

https://github.com/user-attachments/assets/b209c830-c714-4108-93df-2c5f1dc288dc

## Checklist


0xdhrv (commented):

* [Change Case](https://www.raycast.com/erics118/change-case) by @erics118

The existing extension already includes a dedicated **Convert to aLtErNaTiNg cAsE** command, with the transformation explicitly exposed as **Sponge Case**.

It already provides:

* Sponge/alternating-case conversion
* A dedicated no-view command for that transformation
* Selected-text and clipboard input
* Configurable copy or paste output behavior
* Many additional case transformations alongside Sponge Case

This PR mainly changes the input flow by accepting text directly as a command argument and immediately copying the Spongecased result.

That seems like a useful enhancement to the existing Sponge Case command rather than enough differentiation for a separate extension.

Could we consider adding optional direct text input to the existing **Convert to aLtErNaTiNg cAsE** command instead?

> This would preserve the one-shot workflow from this PR while avoiding duplication and keeping text case transformations consolidated in one extension.

erics118:
BTW, @evnm the extension does already support fallback commands and quicklinks which can replicate the intended features. You do need to go to settings and enable the relevant command though


https://github.com/user-attachments/assets/bf54f095-8d99-40bd-9fa0-4957a51df0f5

erics118:
oops, nvm i see an issue w/ the commands

evnm:
Thanks for your comment, @0xdhrv. I'd called out why I'd decided to create a separate extension in my PR description:

> I'm aware that the Change Case extension exists. I just find that it doesn't match my typical use case. Rather than operating on selected or already-copied text and requiring me to navigate a menu within the command, I want a oneshot command to quickly convert text to e.g. paste into a chat message.

But I grant that this may be more a personal preference than reasonable justification for a special-cased extension. If y'all would prefer to consolidate usage on Change Case, then I'm not going to argue. I'll just continue to use my extension in dev mode.

erics118:
There actually is a bug with the change case extension.

The proper use case would be to set sponge case as a fallback command, and then you can:
- open raycast
- type out a string
- select the sponge case transformation

Then, the transformation would use the text inputted in from the normal input field, and it would also copy/paste the transformation in.

Unfortunately, fallback text doesn't seem to be supported atm.

PR #29953, opened 2026-08-04
refactor(zshrc-manager): unify parsing on one registry and bring the write path under test

TurboCoder13:
### 🤖 Disclosure

## Description

The extension carried two independently written parser systems: ten standalone regex extractors feeding the type views, and a pattern-registry parser feeding sections and statistics. Any construct matched by one and not the other made the counts users see disagree with the lists behind them — most visibly, multi-line `plugins=(…)` arrays (Oh My Zsh's default layout) were counted by one surface and missed by the other. This PR unifies all parsing on a single registry, brings the file-writing code — the highest-risk, least-tested code in the extension — under component tests, and fixes an undo bug the new tests exposed. It also carries three small resolved-facts fixes that missed the previous PR's merge window.

### Bug fixes

- **Undo now actually reverts adds and edits.** History entries for add/edit operations recorded the post-change file as their restore point, so undoing them silently restored the state you were trying to revert. Every write now records the pre-change snapshot, and only after the write is verified, so a failed write can no longer leave a restore point for a change that never happened.
- **Counts match lists everywhere.** Statistics, section counts, and the type views all derive from one parser; multi-line `plugins=(…)`/`path+=(…)`/`fpath=(…)` arrays parse on every surface; PATH/FPATH declarations are recognized in all supported forms (`export`/`typeset -x`/`declare -x`, `+=` append, array forms, `$PATH`-relative, plain assignment); keybinding counts no longer include bare mode lines (`bindkey -e`) the Keybindings view has nothing to show for.
- **Array parsing follows zsh quoting.** Quoted elements may contain whitespace, parentheses, and escaped quotes; `$(…)` command substitutions stay one element with their text kept verbatim (nested and quoted forms included); inline comments inside array bodies are ignored. And a comment inside a multi-line array that happens to match a section-header format (`## group`, `# Section: X`) no longer splits the array — previously that skewed per-section counts, and on the write side could have let "add to section" insert a line into the middle of the array.
- Sources with quoted operands or trailing inline comments resolve correctly in the `Source Exists` fact; plugin install checks honor `ZSH_CUSTOM`/`ZSH` declared as plain (unexported) assignments; `$HOMEBREW_PREFIX`-style variables are no longer mis-expanded as `$HOME`.

### Technical Highlights

- `pattern-registry.ts` is now the single parser: `extractEntries()` returns typed entries for all 14 construct types, each carrying its line position. The legacy `utils/parsers.ts` functions and `parseZshrc` are thin projections over it. Where the two old parsers disagreed, each construct's behavior was decided explicitly and tested — never the silent union.
- Three invariant test layers: a property test (every view entry appears in registry output exactly once, over a corpus exercising every decision), a golden-count test (statistics == view counts == section counts), and occurrence-targeting tests (duplicate definitions in same-labeled sections are addressed individually; same-section duplicates are refused unchanged).
- Array bodies are tokenized by a small quote-aware scanner shared between element extraction, array-termination detection, and section detection (single/double quotes, backslash escapes, `$(…)` with its own quote context) — one semantics, so no two surfaces can disagree about what a quote covers. Deliberately not a shell interpreter: parameter expansion, `$'…'` quoting, and multi-line quoted elements are out of scope; unrecognized lines degrade to the "Other" count rather than being guessed at.
- The 701-line `edit-item-form.tsx` is split along form/preview/submit boundaries: pure content computation lives in `edit-item-write.ts`, shared by save, delete, and the diff preview — so the preview provably shows exactly what a save writes.
- 90 new tests (1,240 total): array-tokenization ground truth (quoting, escapes, command substitution, section-like comments in arrays), every form flow (create, edit, move, cancel, confirm-guard, fail-closed refusals, write-verification failure, delete ordering), the shared list controller, backup restore/delete and history undo/clear including cancelled paths, and the collections write path. Raycast components are mocked; assertions target the extension's own behavior — the exact content written, writes blocked on cancel, refusals leaving the file untouched.
- Coverage gate ratcheted: lines/statements 56 → 69 (actual ~70), functions 70 → 80. The remaining gap to 75 is the ten type views and collections browser, which the upcoming redesign replaces — deliberately not gold-plated.
- No new dependencies, no new commands, no UI changes.

### Actual Count

actual

### Before Fix

before

### After Fix

after

## Screencast

No new captures — this PR is internal (parsing, tests, write-path correctness) with no visual changes; the existing `metadata/` set remains accurate.

## How to Test

No special hardware or credentials needed — a zsh setup with a few aliases suffices.

1. `npm ci && npm test` — 1,240 tests including the new parser-unification and write-path suites
2. `npm run dev`, open the extension: pick any section in the view switcher and compare its counts against the corresponding type view — they now always agree. If your `plugins=(…)` array spans multiple lines (the Oh My Zsh default), that's the case that used to disagree.
3. Add or edit an alias, then use Undo (⌘Z from the main view, or the History view): the change is actually reverted — previously undo of an add/edit was a silent no-op.
4. Edit an alias and press ⌘⇧P (Preview Changes): the diff shown is computed by the same code path as the save.

## Checklist

- [x] `npm run lint` passes

### DIFF
--- a/extensions/zshrc-manager/CHANGELOG.md
+++ b/extensions/zshrc-manager/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## [One parser for every surface] - {PR_MERGE_DATE}
+## [One parser for every surface] - 2026-08-04
 
 ### Fixed
 
--- a/extensions/zshrc-manager/CHANGELOG.md
+++ b/extensions/zshrc-manager/CHANGELOG.md
@@ -1,5 +1,19 @@
 # Changelog
 
+## [One parser for every surface] - {PR_MERGE_DATE}
+
+### Fixed
+
+- Statistics, section counts, and the type views now agree — all of them derive from a single parser (the pattern registry), so the count you see on a section can no longer differ from the list you drill into. Previously two independently written parser systems could disagree about what is in the file
+- Multi-line `plugins=(…)`, `path+=(…)` and `fpath=(…)` array declarations — the standard Oh My Zsh layout — are now parsed everywhere; sections and search previously missed them
+- PATH and FPATH declarations in all supported forms (`export`/`typeset -x`/`declare -x`, `+=` string append, array append/set, `$PATH:`-relative, and plain assignment) are recognized consistently across every surface
+- Keybinding counts no longer include bare mode lines like `bindkey -e` that the Keybindings view has nothing to show for
+- Sources with quoted operands or trailing inline comments resolve correctly in the detail pane's `Source Exists` fact; plugin install checks honor `ZSH_CUSTOM`/`ZSH` set as plain (unexported) assignments; `$HOMEBREW_PREFIX`-style variables are no longer mis-expanded as `$HOME`
+
+### Changed
+
+- Every parsed entry now carries the line it was defined on internally, laying the groundwork for addressing duplicate definitions directly
+
 ## [Resolved facts: shadowed commands, missing sources, missing plugins] - 2026-08-03
 
 ### Added
--- a/extensions/zshrc-manager/CHANGELOG.md
+++ b/extensions/zshrc-manager/CHANGELOG.md
@@ -8,6 +8,8 @@
 - Multi-line `plugins=(…)`, `path+=(…)` and `fpath=(…)` array declarations — the standard Oh My Zsh layout — are now parsed everywhere; sections and search previously missed them
 - PATH and FPATH declarations in all supported forms (`export`/`typeset -x`/`declare -x`, `+=` string append, array append/set, `$PATH:`-relative, and plain assignment) are recognized consistently across every surface
 - Keybinding counts no longer include bare mode lines like `bindkey -e` that the Keybindings view has nothing to show for
+- Array parsing follows zsh quoting: quoted elements may contain whitespace, parentheses and escaped quotes; `$(…)` command substitutions stay one element with their text kept verbatim; inline comments inside array bodies are ignored
+- A comment inside a multi-line array that happens to match a section-header format (`## group`, `# Section: X`) no longer splits the array — per-section counts stay consistent with the views, and adding an item can no longer insert a line into the middle of an array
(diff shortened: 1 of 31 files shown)


TurboCoder13:
## File Breakdown

| Category | Count | Description |
| --- | ---: | --- |
| Implementation | 15 | Registry extractor + quote-aware array tokenizer, parser projections, section detection shielded from array interiors, form split, write pipeline, history snapshots, carried resolve fixes |
| Tests | 14 | Parser-unification invariants, array-tokenization ground truth, write path, controller, backup/history views, carried resolve/parser cases, API mock additions |
| Docs | 1 | CHANGELOG entry |
| Config | 1 | Coverage thresholds ratchet |

### Distribution

```
Implementation  ████████████░░░░░░░░░░░░░  48%
Tests           ███████████░░░░░░░░░░░░░░  45%
Docs            █░░░░░░░░░░░░░░░░░░░░░░░░   3%
Config          █░░░░░░░░░░░░░░░░░░░░░░░░   3%
```

### Implementation layout

| Area | Files | Purpose |
| --- | --- | --- |
| `src/lib` | 13 | `pattern-registry.ts` (single parser: line positions, quote-aware array tokenization/termination), `parse-zshrc.ts` (projection), `section-detector.ts` + `scoped-replace.ts` (section scans skip array interiors, read/write in lockstep), `edit-item-write.ts` (pure save/delete computation), `edit-item-preview.tsx`, `edit-item-form.tsx` (orchestration only), `delete-item.ts` + `history.ts` + `import-export.ts` + `section-writer.ts` + `toggle-item.ts` (verified-write undo snapshots), `resolve.ts` (carried fixes) |
| `src/utils` | 1 | `parsers.ts` — ten extractors as registry projections; `statistics.ts` untouched by design (agrees by construction) |
| `src` | 1 | `constants.ts` — shared source-operand pattern; dead per-construct regexes removed |

31 files, +2,818/−1,128.

0xdhrv (approved):
Looks good to me, approved ✅

PR #29955, opened 2026-08-04
fix(downdetector): handle 403 from search URL by falling through to status page

### LINKED ISSUE #29864: downdetector is using wrong search URL
### Extension

https://www.raycast.com/camille_maurel/downdetector

### Raycast Version

v0.71.0.0

### OS Version

MacOS 26.6

### Description

Looks like the old search URL (downdetector.com/search/?q=<search term> ) doesn't work and results in a 403.

Using downdetector.com/status/<search term> works as expected.

I'm configured with the Global downdetector site.

### Steps To Reproduce

1. Launch Check Service Status
2. Any text entered will result in the 403

### Current Behaviour

Doesn't work

### Expected Behaviour

Should properly show the status as shown at https://downdetector.com/status/

stareezy-1:
## Description

Fixes the downdetector extension failing with a 403 when searching for services.

Fixes #29864

## Problem

The `/search/?q=` endpoint on downdetector.com now returns HTTP 403 (Cloudflare block). The search fallback logic only treated 404 as "try next path" — a 403 caused an immediate throw, preventing the extension from reaching the direct `/status/{slug}` lookup which still works fine.

## Fix

One-character change: treat 403 the same as 404 in the search path fallback loop.

```diff
- if (res.status !== 404) throw new Error(`HTTP ${res.status} — ${url}`);
+ if (res.status !== 404 && res.status !== 403) throw new Error(`HTTP ${res.status} — ${url}`);
```

This lets the extension fall through the search path alternatives and eventually resolve via the direct status page URL.

## Test Plan

- Search for any service (e.g., "github") → no longer throws 403, resolves via `/status/github/`
- Invalid service names still produce the expected "Service introuvable" error
- Services that DO work via search still work (no behavior change for 200/404)

### DIFF
--- a/extensions/downdetector/CHANGELOG.md
+++ b/extensions/downdetector/CHANGELOG.md
@@ -1,3 +1,9 @@
+## [Fixes] - 2026-08-04
+
+### Fixed
+
+- Search no longer fails with HTTP 403 when Downdetector's `/search/` endpoint is blocked by Cloudflare; the extension now falls through to the direct `/status/{slug}` lookup.
+
 ## [Initial Release] - 2026-07-28
 
 ### Added
--- a/extensions/downdetector/CHANGELOG.md
+++ b/extensions/downdetector/CHANGELOG.md
@@ -1,4 +1,4 @@
-## [Fixes] - 2026-08-04
+## [Fixes] - {PR_MERGE_DATE}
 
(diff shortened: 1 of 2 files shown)


pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #29956, opened 2026-08-04
Add spellbook extension

netbull:
## Checklist


0xdhrv (commented):

Raycast already has [Script Commands](https://manual.raycast.com/script-commands) for this. **Create Script Command** turns a frequently used script into a Root Search command, with arguments (including dropdowns), optional confirmation, output modes, and hotkeys.

**Search Commands** / **Save Command** are the same job: keep a personal library of shell commands and run them from Raycast.

- Inline `{{param=default}}` placeholders with last-used values, so Enter can run without prompting
- A single JSON library at `~/.config/spellbook/commands.json`
- Streaming inline output and a Terminal.app / iTerm2 handoff
- A confirmation dialog for destructive patterns

However, these appear to be natural enhancements to Script Commands rather than enough differentiation for a separate Store listing. Script Commands already live in Root Search, support arguments and confirmation, and can be git-synced as files.

Could we keep this library as Script Commands instead?

> There is already a feature to accomplish this directly in Raycast, which offers comparable value.
> As mentioned in our [extension guidelines here ↗](https://manual.raycast.com/extensions-guidelines#reasons-to-be-rejected)

The Store screenshots (`metadata/spellbook-1.png` through `spellbook-4.png`) still show the local-development indicator on the command icon. Could you recapture them with Window Capture (Save to Metadata ticked) so that icon isn't in the published images?

`package.json` also omits `platforms`. The terminal path uses AppleScript for Terminal.app / iTerm2, so this would currently install on Windows. Could you set `"platforms": ["macOS"]` if this listing continues?

---

PR #29978, opened 2026-08-04
Add Wi-Fi Inspector extension

Tripchipp:
## Description

Adds **Wi-Fi Inspector**, a Raycast frontend for [macwifi-cli](https://github.com/jaisonerick/macwifi-cli) (MIT).

- **Scan Nearby Networks** — SSID, BSSID, RSSI, channel, band, security
- **Current Connection** — details for the active Wi-Fi network
- Automatic Cloudflare speed test (download / upload / latency) when viewing details on the connected network
- Copy Wi-Fi password from Keychain for saved networks
- Auto-downloads the notarized `macwifi-cli` binary from GitHub Releases (SHA-256 verified)

Requires Apple Silicon + macOS 13+. Location Services approval for WifiScanner on first scan.

## Checklist

- [x] I tested previously reported issues
- [x] Resolved as much as possible lint / TypeScript warnings and errors

## Screencast / Screenshots

Store metadata screenshots included under `extensions/wifi-inspector/metadata/`.

Made with [Cursor](https://cursor.com)


pernielsentikaer (changes requested):
Thanks for the submission, and nice work on the scanning UI! 🔥

There is a overlap with existing extensions. Nearby-network scanning is genuinely new, but three of the four features already exist in the store:
- Copy Wi-Fi password from Keychain → WiFi Password Reveal
- Speed test → Speedtest and Network Speed
- Current connection details → Network Diagnostics

We generally prefer extensions that add something the store doesn't
already cover, rather than re-bundling it. Would you be open to
focusing this on the scanner and current-connection details (the part
that's actually new) and dropping the password and speed-test
commands?

Binary download. The extension downloads macwifi-cli from GitHub
Releases at runtime. Fetching and executing an external binary after
install isn't something we can review or guarantee, so we'd need
another approach here. Could you say more about why the CLI is needed
versus calling the system APIs directly?

Happy to take another look once that's settled.

---

Tripchipp:
Thanks for the review, happy to narrow the scope.

Overlap:
I'll drop Copy Wi-Fi Password and the Cloudflare speed test (those are
actions/auto-run inside the two existing commands, not separate
commands).
Then keep Scan Nearby Networks and Current Connection. Current
Connection stays as RF-layer details (SSID, BSSID, RSSI, channel,
band, security) next to the scanner. So im not trying to replace
Network Diagnostics (IP/DNS/ping).

Why a CLI vs system APIs. Nearby scan and real BSSIDs are no longer
available from shell tools:
- airport was removed in macOS 14.4
- networksetup / wdutil / system_profiler don't return nearby networks
or real BSSIDs without Location Services
- CoreWLAN scanForNetworks only returns real BSSIDs to a Developer
ID-signed, notarized helper that has Location permission
  (https://developer.apple.com/forums/thread/718331)


The extension can't call CoreWLAN itself, and a helper id ship would
still need to be developer ID-signed and notarized.
Macwifi-cli (https://github.com/jaisonerick/macwifi-cli) already wraps
that helper, which is the only reason i use it.


A couple of alternatives i can think of where i can stop downloading the
binary at runtime:

A) Require homebrew (`brew install jaisonerick/tap/macwifi-cli`), optional
custom path preference, no download.
B) Ship the notarized darwin/arm64 binary in the extension assets instead
of downloading it (Apple silicon only).
I can add it in the PR if thats ok, or follow whatever process you use for
native binaries.
Happy to go with whatever you prefer here, including something else if you
have a better approach in mind.

PR #29979, opened 2026-08-04
Add capd extension

jamiedavenport:
## Description

Adds **Capd**, an extension for [Capd](https://capd.jxd.dev) — an open-source, local-first macOS app for capturing web pages, selected text, links, notes, and images and finding them again with full-text search.

Two commands:

- **Search** — full-text search across titles, extracted page text, selections, notes, and OCR'd image text. The search bar takes Capd's own filters (`site:`, `tag:`, `after:`, `before:`); an empty search bar lists recent captures. Results can be opened, copied as a URL/title/Markdown, or deleted.
- **Capture** — saves a link or text from the command argument, the current selection, or the clipboard. Works well as a fallback command. Re-capturing something already saved updates the existing entry rather than duplicating it, and the confirmation says so.

### How it talks to Capd

The extension shells out to the `capd` CLI that ships inside the user's own `capd.app` bundle, using its documented JSON output and exit codes. Worth calling out up front, since the guidelines ask about binaries:

- **Nothing is bundled or downloaded.** No binary is committed, fetched, or executed from a remote source. The extension only invokes a binary the user has already installed themselves.
- **It resolves the app, not a path.** `getApplications()` locates the `dev.jxd.capd` bundle, so it works wherever the user keeps `capd.app` and needs nothing on `PATH`. A preference exists as an escape hatch.
- **`execFile`, never a shell**, so search text can't be interpreted as a command.
- **No credentials, no Keychain, no network requests, no analytics.** The extension has no account concept at all — everything stays on the user's Mac.
- Capd is open source (MIT) and distributed as a notarized `.dmg` and a Homebrew cask: [github.com/jamiedavenport/capd](https://github.com/jamiedavenport/capd).

### Testing it

Capd requires macOS 26 or later. To try the extension:

```sh
brew install jamiedavenport/tap/capd
capd add https://react.dev
```

Then run **Capd: Search** and type `site:react.dev`. If Capd isn't installed, both commands show an actionable empty state linking to the install guide rather than failing silently.

Docs: [capd.jxd.dev](https://capd.jxd.dev) · [CLI and Raycast reference](https://capd.jxd.dev/cli) · [Privacy](https://capd.jxd.dev/privacy)

## Screencast

Search results, with favicons, snippets, tags, hosts, and capture dates:

Search

The same search narrowed with a `site:` filter:

Filtered search

Actions available on a result:

Actions

## Checklist


0xdhrv (approved):
Looks good to me, approved ✅

PR #29981, opened 2026-08-04
fix(google-meet): accept Brave under its real app name

### LINKED ISSUE #29968: [Google Meet] Brave Browser incorrectly rejected as unsupported
### Extension

https://www.raycast.com/ThatNerd/google-meet

### Description

The Google Meet extension stopped working with Brave after the latest update. It worked normally until yesterday, but now running `Create Meet` shows:

```
"Brave Browser" isn't a supported browser.
Choose a supported browser in the extension preferences, or set a preferred browser.
```

Brave appears to still be intended as a supported browser, so this may be a name-matching issue: macOS/Raycast reports the application name as `Brave Browser`, while the extension may be checking for `Brave`. Normalizing both names to the same browser identifier should likely resolve it.

### Steps To Reproduce

1. Set Brave Browser as the default browser on macOS.
2. Open Raycast.
3. Run the Google Meet extension's `Create Meet` command.
4. The command fails with the unsupported-browser message.

### Current Behaviour

The extension rejects Brave as unsupported and does not create the meeting.

### Expected Behaviour

A new Google Meet should be created and opened in Brave, as it did before the latest extension update.

### Environment

- OS: macOS
- Browser: Brave Browser
- Extension: Google Meet
- Command: Create Meet

darkdi:
## Description

Fixes the Google Meet extension rejecting Brave as an unsupported browser.

Fixes #29968

## Problem

macOS installs Brave as `Brave Browser.app`, so both `getDefaultApplication` and the
frontmost-process lookup report the name `Brave Browser`. `supportedBrowsers` listed it as
`Brave`, and `resolveCreationBrowser` matches that list exactly:

```ts
function isSupportedBrowserName(name: string): name is SupportedBrowsers {
  return (supportedBrowsers as readonly string[]).includes(name);
}
```

The real name is not in the list, so Create Meet stops with
`"Brave Browser" isn't a supported browser.` — the message quoted in the issue.

This surfaced in 3.x. Before that, `getOpenedBrowser` cast the detected name straight to
`SupportedBrowsers` without checking it, so `Brave Browser` reached
`tell application "Brave Browser"` unchanged and worked. The reliability overhaul replaced
that unchecked cast with a real membership check, which the entry `Brave` fails.

## Fix

Use the actual application name in the list and in the Chromium adapter switch. That name is
also what AppleScript needs in `tell application`, which is why the earlier unchecked path
worked with it.

```diff
-  "Brave",
+  "Brave Browser",
```

`Brave` was never a name macOS reports for this app, so nothing else resolves to it. The
exhaustive `never` check in `getAdapterForBrowser` covers the switch, so the rename is
compile-checked.

Other Chromium browsers in the list are unaffected — their app names already match their
entries.

### DIFF
--- a/extensions/google-meet/CHANGELOG.md
+++ b/extensions/google-meet/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Google Meet Changelog
 
+## [Fix Brave Support] - {PR_MERGE_DATE}
+
+- Fix Brave being rejected with "isn't a supported browser". macOS installs Brave as `Brave Browser.app` and reports it under that name, but the supported-browser list held `Brave`, so the exact-match check added in the previous release failed for it.
+
 ## [Reliability Overhaul & PWA Support] - 2026-08-03
 
 - Replace the fixed-delay, unbounded-recursion URL lookup with a single shared `createMeeting` pipeline used by all four commands: it polls for the generated link at a fixed interval with a hard deadline instead of guessing a one-shot sleep, and never casts a missing URL to `string`.
--- a/extensions/google-meet/CHANGELOG.md
+++ b/extensions/google-meet/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Google Meet Changelog
 
-## [Fix Brave Support] - {PR_MERGE_DATE}
+## [Fix Brave Support] - 2026-08-05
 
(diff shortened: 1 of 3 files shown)


pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #29983, opened 2026-08-04
feat(bouncer-email-verifier): add domain checks and Raycast AI tools

ekamran:
## Description

An update to **Bouncer Email Verifier**, renamed in the store to **Bouncer Deliverability** because it no longer only checks addresses. The `name` slug is unchanged, so the store URL and existing installs are unaffected.

The starting point was a bug: Bouncer returns `domain.free`, `domain.disposable`, `domain.acceptAll` and the whole `account` object as `"yes"`/`"no"`/`"unknown"` strings, but the extension typed them as booleans. Every signal on every result therefore rendered as "Not reported". Correcting the types brought the rest of the payload into scope.

**Fixes**

- Mailbox and domain signals now display. They were unreachable.
- Added the `risky` verdict. Catch-all and full-mailbox addresses were previously collapsed into "Unknown", and the verdict switch compared `reason` values against `status`, so three of its five branches could never match.
- Requests carry a timeout and abort in flight; a hung check previously left the view spinning forever, and pressing Enter twice raced two responses.
- Invalid input is rejected locally instead of spending a credit on it.
- `401`, `402` and `429` now produce distinct, actionable errors rather than a generic failure.

**New**

- **Domain checks** via `GET /v1.1/domain`, in the same command. Input decides which endpoint runs — an address contains an `@` and a domain does not — so there is nothing for the user to pick. Pasting `Name <addr@example.com>`, a signature block, or a URL all resolve correctly.
- **Three Raycast AI tools**: verify an address, check a domain, read the credit balance. `ai.instructions` tells the model that a signal of `"Unknown"` must never be reported as `"No"`, and that role/free are heuristics rather than defects.
- **A send / hold / suppress recommendation** per result, derived from Bouncer's [integration guidelines](https://docs.usebouncer.com/integration-guidelines). The source marks which rules are theirs and which are our inference, so the distinction survives future edits.
- **Local history** of the last 50 checks, filterable by verdict, ranked by frecency, and switchable off in preferences. Stored rows are validated and migrated on read so an older row cannot break the command.
- Account signals, toxicity, DNS record, `retryAfter`, and `didYouMean` suggestions that can be verified in one keystroke.
- New icon, built from solid mass rather than strokes so it holds up at the size the command list actually draws it.

**Notes for review**

- Signals are coloured by what they mean *for that verdict*: a catch-all domain is the reason a `risky` result is risky, so it is highlighted there, but on a confirmed `deliverable` it is shown as plain information. Only signals worth acting on are rendered.
- `role` and `free` never affect the recommendation. Both are heuristics over the local part and domain, and both misfire — a personal `mail@` address is commonly reported as a shared role mailbox — so they are reported as neutral facts.
- The README's privacy section now names both egress paths, including that the AI tools hand addresses to Raycast AI and therefore to the model provider behind it.
- `platforms` keeps `Windows`, which the published version declares. Restoring it surfaced three keyboard shortcuts that could not work there; those now use `Keyboard.Shortcut.Common` or declare per-platform variants. I have no Windows machine to verify on.
- `assets/icon-source.svg` is the editable source for the icon rather than a runtime asset. It shipped in the initial release and I have kept it for future edits — happy to remove it if you would rather `assets/` contained only what the extension loads.
- The `401` / `402` / `429` / timeout paths are written but were not triggered against the live API.

## Screencast

Five screenshots in `metadata/`, all captured against [Bouncer's sandbox addresses](https://docs.usebouncer.com/sandbox) so they contain no real recipients:

1. The list — input row, credit balance, and history with mixed address and domain results
2. `accept-all@sandbox.usebouncer.com` — **Risky**, Send with Caution
3. `deliverable@sandbox.usebouncer.com` — **Deliverable**, Safe to Send
4. `usebouncer.com` — the domain check, Mail Configured
5. `undeliverable@sandbox.usebouncer.com` — **Undeliverable**, Remove

## Checklist


0xdhrv (commented):
Hi @ekamran 👋

Thanks for the contribution 🔥

---

ekamran:
Thanks @0xdhrv, and thanks to Greptile — all three were real. Fixed in `9b1e02a`.

**History mutations overwrite each other** — valid, and worse than it reads. Each mutation was a read-modify-write against a single storage key, so two overlapping ones both read the same snapshot and the second discarded the first. A test writing five entries concurrently kept **one** before the fix and keeps all five after it. Mutations now run through a queue, so each read sees the previous write, and `clear` is queued alongside them so a verification landing mid-clear cannot resurrect a row.

**Implicit MX routing is rejected** — valid. RFC 5321 §5.1 falls back to the A/AAAA record when a domain publishes no MX, so an address-only domain can still receive mail, and calling that a delivery failure was wrong. A/AAAA now reports as a misconfiguration worth knowing about (orange) rather than undeliverable. Record types mail genuinely cannot be routed to, such as CNAME, keep the hard verdict, and a domain with no usable records at all is now titled "No Mail Records" rather than "No MX Record", which is what it actually means.

**Unknown catch-all becomes definitive** — valid. `acceptAll` can come back `"unknown"`, and treating that as "not catch-all" asserted server behaviour the API never established. The "rejects addresses that do not exist" claim is now made only when `acceptAll` is explicitly `"no"`; otherwise the verdict says Bouncer could not determine it.

Two things from the description still worth your eye, since neither is something I can settle here:

- `platforms` keeps `Windows` because the published version declares it and dropping it would break existing users. Restoring it surfaced three shortcuts that could not work there, now using `Keyboard.Shortcut.Common` or per-platform variants — but I have no Windows machine to verify on.
- `assets/icon-source.svg` is the editable source for the icon rather than a runtime asset. It shipped in the initial release; happy to remove it if you would rather `assets/` held only what the extension loads.

Ready for another look.

ekamran:
One more, self-reported: the same overclaim Greptile raised for domain verdicts also existed in the email path, where it was not flagged.

`status: "unknown"` is Bouncer saying it could not determine the result, but a `dns_error` reason was being reported as *"the domain has broken or missing DNS records, so mail cannot be routed"*. DNS failures are usually a dead domain, but they can be transient, and either way that is a stronger claim than an `unknown` status supports. The removal recommendation stands — Bouncer's keep list for `unknown` covers `unknown`, `timeout` and `unsupported`, and deliberately excludes this reason — but the wording now says what was actually observed and notes it can be temporary.

Also softened the full-mailbox copy from a bounce guarantee to a likelihood, for the same reason.

Pushed as `237ef1a`. I went through every remaining string that asserts an outcome and checked it against what the API establishes; the rest either restate Bouncer's own verdict or its documented terminology.

0xdhrv (approved):
Looks good to me, approved ✅

PR #29999, opened 2026-08-05
Escape user input in PyPI search query and URL

benjaminkifle:
## Description

The search term was interpolated directly into both a SQL string literal and a
URL query parameter, with no escaping for either context.

**Broken SQL.** A typed apostrophe closed the string literal early. Searching
`it's` produced:

    SELECT * FROM `projects` WHERE name LIKE 'it's%' LIMIT 50;

which the API rejects with `syntax error at position 47 near 's'`. More
seriously, the same gap allows arbitrary SQL to be grafted onto the query —
a search term of `' OR '1'='1` produced valid, executing SQL rather than a
search string.

**Broken URLs.** Characters like `&` and `#` terminated the query parameter
early, so the server received truncated SQL with no `LIMIT` clause.

### Fix

The query is now built first and passed through `encodeURIComponent` as a
single unit, with `'` and `\` escaped for the SQL string context beforehand.
The original hand-encoded the *static* parts of the URL (`%60`, `%27`, `%3B`)
but had no way to encode the interpolated term, which is the root of both
problems.

`%` and `_` are deliberately **not** escaped, so existing wildcard behavior
is unchanged.

The "View PyPI Search Results" action had the same unescaped interpolation
and is now encoded too.

This does not change how results are returned — it only fixes queries that
were previously malformed.

### Testing

| Search | Result |
| --- | --- |
| `fastapi` | 50 results — unchanged |
| `100%` | 17 results — unchanged |
| `typing_ext` | unchanged |
| `it's` | valid SQL instead of a server-side syntax error |

`npm run lint` and `npm run build` both pass.

## Screencast

No visible UI change — this fixes malformed queries, and successful searches
look identical before and after.

## Checklist

### DIFF
--- a/extensions/python/CHANGELOG.md
+++ b/extensions/python/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Python Changelog
 
-## [Fix Search Query Escaping] - {PR_MERGE_DATE}
+## [Fix Search Query Escaping] - 2026-08-05
 
 - Escape user input before interpolating it into the search query and URL.
 
--- a/extensions/python/CHANGELOG.md
+++ b/extensions/python/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Python Changelog
 
+## [Fix Search Query Escaping] - {PR_MERGE_DATE}
+
+- Escape user input before interpolating it into the search query and URL.
+
 ## [Maintenance] - 2026-01-15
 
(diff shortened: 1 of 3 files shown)


0xdhrv (approved):
Looks good to me, approved ✅

PR #30012, opened 2026-08-05
Add Touch ID authentication to Flush DNS

### LINKED ISSUE #13104: Flush DNS asks for password instead of Touch ID on macOS Sonoma
### Extension

https://www.raycast.com/rasmusbe/flush-dns

### Raycast Version

1.77.1

### macOS Version

14.5

### Description

# Description
When activating the 'Flush DNS' command with this extension, a dialog for 'osascript' is presented ("osascript wants to make changes. Enter your password to allow this." (screenshot below). There is no option to use Touch ID, even if the Touch ID PAM module (`pam_tid.so`) is configured in `/etc/pam.d/sudo`.

This greatly reduces the convenience of the extension, in my opinion. For instance, I previously achieved the same result by opening a terminal and manually pasting the command to flush the cache and restart mDNSResponder (or having it in history or whatever), leveraging my configuration that enables Touch ID for `sudo` commands. I had hoped this extension would make this even easier by it being able to launch it from Raycast.

# Cause
The command the extension is attempting to run leverages `osascript` with the `with administrator privileges` instruction: https://github.com/raycast/extensions/blob/96f33396c6ceae40f078ff897e148ab04b4a3ae0/extensions/flush-dns/src/index.ts#L48

Based on similar reports relating to `osascript` I came across while Googling, I _think_ this might be a change/regression in macOS Sonoma, specifically (but I have not used Raycast and this extension in older versions of macOS so cannot be sure), ie in that previously, passing the `with administrator privileges` instruction to `osascript` _would_ present a Touch ID-enabled dialog.

# Solution
A solution might be to drop the `with administrator privileges` component, since the shell commands also include `sudo`; in that situation, if PAM is configured to use Touch ID for `sudo`, then a Touch ID-enabled dialog _is_ presented (example screenshot below)

## Screenshot of password dialog for `osascript`
image

## Example screenshot of Touch ID dialog
With `pam_tid.so` enabled in my PAM config for `sudo`, running the following shell command (ie omitting the `with administrator privileges` part for `osascript`) produces a Touch ID-enabled dialog (for `sudo` instead of `osascript`) as follows:
```
osascript -e 'do shell script "sudo /usr/bin/dscacheutil -flushcache; sudo /usr/bin/killall -HUP mDNSResponder"'
```
image

### Steps To Reproduce

1. With the Flush DNS extension installed in Raycast on macOS Sonoma (14.5)...
2. with or without `pam_tid.so` enabled for `sudo`...
3. Run the 'Flush DNS' command
4. Observe a password-only dialog (for `osascript`) is presented, with no option to use Touch ID

### Current Behaviour

Touch ID is not usable for the Flush DNS extension

### Expected Behaviour

Using Touch ID to authorise the commands to flush the DNS cache would be much more convenient

TyceHerrman:
## Description

Adds optional Touch ID authorization for privileged DNS flushing on macOS.

- Detects an active `pam_tid.so` rule in `/etc/pam.d/sudo_local`, falling back to `/etc/pam.d/sudo`.
- Runs the selected cache commands sequentially through `/usr/bin/sudo` using exact executable and argument arrays when Touch ID is configured.
- Stops after Touch ID cancellation or a `sudo` failure without opening a surprise password prompt.
- Retains the existing one-dialog AppleScript administrator-password fallback when Touch ID is unavailable, with a 60-second timeout.
- Preserves the existing Windows and legacy macOS command selection while documenting the separate Directory Service and `mDNSResponder` cache layers.

Closes #13104.

## Screencast

Not included. Manually verified in Raycast Beta on macOS 27: the local development command presented Touch ID and completed both modern macOS cache commands.

## Validation

- `npm test` — 11 tests passed
- `npm run lint`
- `npm run build`
- `git diff --check`
- Raycast Beta Touch ID happy-path test

## Checklist

### DIFF
--- a/extensions/flush-dns/CHANGELOG.md
+++ b/extensions/flush-dns/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Flush DNS Changelog
 
-## [Touch ID Authentication] - {PR_MERGE_DATE}
+## [Touch ID Authentication] - 2026-08-12
 
 - Added Touch ID authentication for macOS DNS flushing while retaining the administrator password fallback ([#13104](https://github.com/raycast/extensions/issues/13104)).
 
--- a/extensions/flush-dns/CHANGELOG.md
+++ b/extensions/flush-dns/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Flush DNS Changelog
 
+## [Touch ID Authentication] - {PR_MERGE_DATE}
+
+- Added Touch ID authentication for macOS DNS flushing while retaining the administrator password fallback ([#13104](https://github.com/raycast/extensions/issues/13104)).
+
 ## [macOS 26+ Support] - 2026-01-04
 
 - Added support for macOS Tahoe (26) and later versions.
--- a/extensions/flush-dns/README.md
+++ b/extensions/flush-dns/README.md
@@ -3,3 +3,37 @@
 Flush the DNS cache
 
 Works for macOS 10.6+ and Windows
+
+## macOS Cache Coverage
+
+On modern macOS, Flush DNS clears both OS-level host-resolution caches:
+
+- `/usr/bin/dscacheutil -flushcache` clears the Directory Service/libinfo cache.
+- `/usr/bin/killall -HUP mDNSResponder` purges the separate `mDNSResponder` DNS-reply cache.
+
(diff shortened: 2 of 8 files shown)


pernielsentikaer (approved):
Thanks for the careful Touch ID / sudo path — looks good. Approving.

PR #30016, opened 2026-08-06
Update YouTrack extension: add live search and autocomplete

TyceHerrman:
## Description

Adds live server-side issue search and YouTrack query autocomplete to the Browse Issues command.

- replaces local filtering of the configured issue list with controlled, server-backed search
- displays YouTrack search-assist suggestions above matching issues and applies the API-provided completion ranges
- preserves the configured query as the default when the search field is empty
- keeps autocomplete usable for incomplete queries such as `State: ` without showing the issue endpoint's expected HTTP 400 as a failure toast
- continues surfacing configured-query, authentication, authorization, network, and server failures

Previously, Browse Issues always fetched the query saved in preferences, so users could only filter that fixed result set locally. This implements the global search workflow requested in #3156.

The incomplete-query toast occurred because applying a field suggestion immediately sent the partial query to the issues endpoint. YouTrack correctly rejected that intermediate grammar while the independent search-assist endpoint still returned value suggestions. Interactive issue-query 400s are now treated as an editable empty-result state; other errors are rethrown.

## Validation

- 18 Vitest tests, including interactive 400 suppression and non-400/configured-query propagation
- TypeScript `--noEmit`
- ESLint
- Prettier on all changed files
- prescribed `npm run build` distribution build
- authenticated manual testing of the final Raycast distribution: field suggestions, the incomplete `state: ` flow without a 400 toast, value suggestions, applied state values, and live matching issues
- `git diff --check`

## Checklist

### DIFF
--- a/extensions/youtrack/CHANGELOG.md
+++ b/extensions/youtrack/CHANGELOG.md
@@ -1,6 +1,6 @@
 # YouTrack Changelog
 
-## [Update] - {PR_MERGE_DATE}
+## [Update] - 2026-08-14
 
 - Add server-side issue search with YouTrack query suggestions
 
--- a/extensions/youtrack/CHANGELOG.md
+++ b/extensions/youtrack/CHANGELOG.md
@@ -1,5 +1,9 @@
 # YouTrack Changelog
 
+## [Update] - {PR_MERGE_DATE}
+
+- Add server-side issue search with YouTrack query suggestions
+
 ## [Update] - 2025-08-08
 
 - Bugfixes and improvements
--- a/extensions/youtrack/package-lock.json
+++ b/extensions/youtrack/package-lock.json
@@ -8,6 +8,7 @@
       "license": "MIT",
       "dependencies": {
         "@raycast/api": "^1.101.1",
+        "@raycast/utils": "^2.2.0",
         "beautiful-react-hooks": "^5.0.3",
         "lodash": "^4.17.21",
         "youtrack-client": "^0.5.3"
@@ -15,7 +16,6 @@
       "devDependencies": {
         "@eslint/eslintrc": "^3.3.1",
         "@eslint/js": "^9.31.0",
-        "@raycast/utils": "^2.2.0",
         "@types/lodash": "^4.17.20",
         "@types/node": "22.13.14",
         "@types/react": "19.1.8",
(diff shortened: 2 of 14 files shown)


pernielsentikaer (changes requested):
Hi @TyceHerrman 👋

- **Surface `useCachedPromise` errors with EmptyView** — destructure `error` from the issues (and ideally suggestions) hooks. On failure, show something like “Failed to load issues” + `error.message` (and Retry) instead of “No Issues Found”, which currently appears whenever the list is empty after a failed load. Keep “No Issues Found” for successful empty results only.

Nice-to-haves in the same pass if easy: move `@raycast/utils` into `dependencies` (it's imported at runtime from `browse.tsx` but still only under `devDependencies`), thread `AbortSignal` through the live-search fetch wrappers so fast typing cancels in-flight calls, and add `keepPreviousData` on the suggestions promise to reduce flicker while typing.

---

TyceHerrman:
@pernielsentikaer, I implemented:

- Issue-loading failures now show a persistent error view with the error message and a Retry action instead of “No Issues Found.” Successful empty responses still use “No Issues Found.”
- Suggestion-loading failures show their own Retry view when the list otherwise has no results. When issues are available, the suggestion failure remains nonfatal: the issues stay visible and the existing failure toast reports the suggestion error.
- `@raycast/utils` now lives in `dependencies`, with the lockfile updated accordingly.

I’m leaving `AbortSignal` propagation for a focused follow-up PR. `useCachedPromise` supports abort controllers, but the `youtrack-client` resource methods used here do not currently expose a per-call signal, so this needs a transport-level change rather than adding an unused controller.

I did not add `keepPreviousData` to the suggestions promise. YouTrack suggestions contain completion ranges tied to the query that produced them, so retaining them while `searchText` changes can apply stale offsets to a newer query. That should be handled as part of a redesigned, query-synchronized autocomplete PR that keeps each suggestion set paired with its source query.

pernielsentikaer (approved):
Hi @TyceHerrman 👋

Thanks for the follow-up 💪

Re-reviewed HEAD against the notes from last time — the error EmptyView landed:

- Issue-load failures now show “Failed to Load Issues” with `error.message` and Retry, and “No Issues Found” stays for successful empty results only.
- Suggestion failures have their own Retry view when the list is empty, and stay nonfatal when issues are already showing.
- `@raycast/utils` is in `dependencies`.

Fine to leave `AbortSignal` and `keepPreviousData` for a follow-up — both reasons make sense.

Looks good to me, approved 🔥

PR #30018, opened 2026-08-06
fix(google-meet): launch the resolved application instead of matching by name

artisticmedic:
## Description

Meetings can open in the wrong application. On my machine every command opened Obsidian — launching it even when it was quit — instead of Dia, my default browser.

The cause is that the creation URL is opened by passing the browser's display name to `open()`:

```ts
const browserName = await resolveCreationBrowser(); // "Dia"
await open(creationUrl, browserName);
```

`open()` has to match that name back to an installed application, and short browser names are substrings of unrelated app names — `Dia` is a substring of `Obsi**dia**n`. Setting "Set preferred browser" to Dia explicitly made no difference, because the same string is what gets resolved either way. Switching the preference to Google Chrome worked, since that name collides with nothing.

This is only reachable since #29481, which introduced `getDefaultApplication()` + name-based resolution. Before that the app argument was left `undefined` when no preferred browser was set, so Launch Services routed the URL by its own default and returned the right browser.

### The fix

`resolveCreationBrowser()` already has the full `Application` from both of its sources — the `appPicker` preference and `getDefaultApplication()` — and was discarding everything but `.name`. It now returns both identities, because they do two different jobs:

- **`application`** launches the browser. Passing the resolved `Application` targets the exact bundle, so no name matching happens.
- **`name`** scripts it. AppleScript addresses an app by the name macOS installs it under (`tell application "Brave Browser"`), and `getAdapterForBrowser()` selects the browser family from that same name.

This leaves the `supportedBrowsers` name matching from #29981 fully intact — that fix is about the scripting identity, which still needs the exact installed name. The change only stops the *launch* target from being re-derived from a display name.

It also brings the browser path in line with the PWA path, which already passes an `Application` to `open()` (`browser-adapters/pwa.ts`).

## Screencast

Straightforward two-file change with no UI surface, so no screencast. Verified manually with Dia set as the default browser, in both branches of `resolveCreationBrowser()`:

- **"Set preferred browser" set to Dia** — before: Obsidian launched, no meeting, no error toast. After: Dia opens `meet.google.com/new`, the link is detected and copied.
- **"Set preferred browser" left empty** (falls through to `getDefaultApplication()`) — same result. This is the path that originally broke.

`npm run build` and `npx ray lint` both pass.

## Checklist

### DIFF
--- a/extensions/google-meet/CHANGELOG.md
+++ b/extensions/google-meet/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Google Meet Changelog
 
+## [Fix Browser Launch] - {PR_MERGE_DATE}
+
+- Fix meetings opening in the wrong application when the chosen browser's name is a substring of another installed app's name. The creation URL was opened by passing the browser's display name to `open()`, which has to match that name back to an application — and `Dia` matches `Obsidian`, so meetings opened in Obsidian, launching it even when it wasn't running. The resolved `Application` is now passed instead, so the exact bundle is opened. The browser's name is still what selects the adapter and addresses the app in AppleScript, which needs the name macOS installs it under.
+
 ## [Fix Brave Support] - 2026-08-05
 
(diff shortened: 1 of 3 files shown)


0xdhrv (approved):
Looks good to me, approved ✅

PR #30041, opened 2026-08-06
fix(google-chrome-profiles): don't open a window for the previously used profile on cold start

radekcabaj:
## Description

Opening a profile while Chrome is **closed** produces two windows: the profile you asked for, plus one for whichever profile was used last.

I hit this daily — quit Chrome on profile X, ask the extension for profile Y, and get a stray X window every time alongside the Y window I wanted.

### Cause

The Bring to Front / New Tab / Open URL path starts with:

```applescript
tell application "Google Chrome" to activate
```

`activate` launches Chrome with **no arguments**. A flagless Chrome restores every profile listed in `profile.last_active_profiles` (in `Local State`), so the previously used profile's window is already on screen before System Events ever clicks the Profiles menu. The menu click then adds the requested profile as a second window.

The Profiles menu is only reachable once Chrome is up, so this path can't avoid cold-starting it — and cold-starting it via `activate` is exactly what goes wrong.

### Fix

Branch on whether the browser is already running:

- **Running** — unchanged. The Profiles-menu click still gives the existing focus / add-a-tab / find-an-existing-tab semantics.
- **Not running** — a single cold start straight into the requested profile:

```
/usr/bin/open -n -a <bundle> --args --profile-directory=<dir> --new-window [url]
```

`--profile-directory` is the profile a cold Chrome boots into, which is what suppresses the session restore. With no window to focus or add a tab to, all three actions collapse to this same cold start; `openUrl` passes the URL as the trailing argument.

`is running` is used because it's the one way to ask about an application without launching it as a side effect.

### Also fixed

The **New Window** action previously exec'd Chrome's inner Mach-O binary directly, so `do shell script` never returned and the detached `osascript` stayed alive for the browser's entire session. It now shares the `/usr/bin/open` helper, which returns as soon as Launch Services accepts the request. `BrowserConfig.binaryPath` becomes `appPath` accordingly, since `open -a` expects the `.app` bundle rather than the executable inside it.

## Testing

Two profiles (`Default`, `Profile 1`), with `last_active_profiles` containing both. For each case Chrome was fully quit first, then the profile that was **not** last used was requested. Profiles actually loaded were confirmed via `lsof` on the browser process rather than by eyeballing windows:

| | Chrome argv | profiles loaded | windows |
|---|---|---|---|
| Before (`activate`) | `…/Google Chrome` (no flags) | `Default` + `Profile 1` | 2 |
| After | `…/Google Chrome --profile-directory=Profile 1 --new-window` | `Profile 1` only | 1 |

Also re-checked with Chrome already running: Bring to Front raises the existing window, New Tab adds a tab rather than a window, and Open URL still jumps to a matching tab when one exists.

`ray lint` and `ray build` both pass.

### Note

One case this cannot address: if a profile has *"Continue where you left off"* set, Chrome restores it on any cold start regardless of `--profile-directory`. That's Chrome's startup behaviour and is out of scope here.

### DIFF
--- a/extensions/google-chrome-profiles/CHANGELOG.md
+++ b/extensions/google-chrome-profiles/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Google Chrome Profiles Changelog
 
-## [Fix] - {PR_MERGE_DATE}
+## [Fix] - 2026-08-12
 
 - Fix opening a profile while Chrome is closed producing two windows: the requested profile, plus one for whichever profile was used last. The Bring to Front / New Tab / Open URL path began with `tell application "Google Chrome" to activate`, which launches Chrome with no arguments; a flagless Chrome restores every profile listed in `profile.last_active_profiles` (Local State), so the previous profile's window appeared before System Events ever clicked the Profiles menu. Branch on whether the browser is already running: when it is, the Profiles-menu click is unchanged and keeps the focus / add-a-tab / find-an-existing-tab semantics; when it is not, do a single cold start into the requested profile with `/usr/bin/open -n -a <bundle> --args --profile-directory=<dir> --new-window [url]`. `--profile-directory` is the profile a cold Chrome boots into, which is what suppresses the session restore.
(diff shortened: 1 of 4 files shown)


pernielsentikaer (approved):
Looks good to me, approved 🔥

frouo:
Thank you @radekcabaj for your contribution 👍

PR #30053, opened 2026-08-07
Update try extension

chrismessina:
## Description

Two fixes to the `try` extension, both about the UI telling the truth about what it just did.

A success toast shown when creating a directory could outlive the command that created it — after the Raycast window closed it detached into a floating overlay and sat on the desktop. The toast handle returned by `showToast` was discarded, so nothing could ever dismiss it deterministically.

Separately, the delete action claimed "This cannot be undone", but it has always called `trash()` — the directory goes to the system Trash and can be restored. The action is now the built-in `Action.Trash`, named for what it actually does.

### Changes

- **Create:** hold the `Toast` handle and dismiss it explicitly after a short read window, so the "Created" confirmation can't follow the user around the desktop.
- **Create:** narrowed the `try`/`catch` to the one call that can fail (`createTryDirectory`), so a failure to dismiss the toast can no longer report a directory that *was* created as a failure.
- **Delete → Move to Trash:** replaced the hand-rolled `confirmAlert` + `trash()` + toast handler with the built-in `Action.Trash`, and renamed the action. Removed the incorrect "This cannot be undone" warning.
  - Note the two behavior changes this carries: `Action.Trash` has no confirmation hook, so the action no longer prompts; and it closes the main window on completion, so confirmation is via `showHUD` rather than a toast (a toast fired as the window closes is exactly the detached-overlay bug above).
- **Shortcuts:** adopted `Keyboard.Shortcut.Common` by semantics.
  - `Common.Remove` for the trash action — identical binding (`⌃X`), no user-visible change.
  - `Common.CopyPath` for Copy Path — **this changes the binding** from `⌘⇧C` (which is `Common.Copy`) to Raycast's standard path-copy shortcut `⌘⇧,`. Called out in the changelog.
- Updated `README.md` to match the renamed action, and refreshed `package-lock.json` (`@raycast/api` 1.104.19 → 1.104.24).

## Checklist

### DIFF
--- a/extensions/try/CHANGELOG.md
+++ b/extensions/try/CHANGELOG.md
@@ -1,6 +1,6 @@
 # try Changelog
 
-## [Fixes] - {PR_MERGE_DATE}
+## [Fixes] - 2026-08-07
 
 - The "Created" confirmation now clears itself instead of staying on screen after you leave Raycast
 - Renamed "Delete" to "Move to Trash", which is what it always did — directories go to the system Trash and can be restored from there
--- a/extensions/try/CHANGELOG.md
+++ b/extensions/try/CHANGELOG.md
@@ -1,5 +1,12 @@
 # try Changelog
 
+## [Fixes] - {PR_MERGE_DATE}
+
+- The "Created" confirmation now clears itself instead of staying on screen after you leave Raycast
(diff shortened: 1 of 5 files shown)


0xdhrv (approved):
Looks good to me, approved ✅

PR #30062, opened 2026-08-07
Update voiceink extension

pasqualesalza:
## Description

Two fixes to the Search Transcriptions command.

**Empty history on current VoiceInk versions.** VoiceInk renamed `powerModeName`/`powerModeEmoji` to `modeName`/`modeEmoji` via `@Attribute(originalName:)` in [Transcription.swift](https://github.com/Beingpax/VoiceInk/blob/main/VoiceInk/Models/Transcription.swift), back in May. SwiftData applies that as a column rename, so any store opened by a build since then has `ZMODENAME`/`ZMODEEMOJI`. `buildQuery` still asked for `ZPOWERMODENAME` whenever the source was `official`, so the statement failed to prepare with `no such column: ZPOWERMODENAME` and the list came back empty. Reproduced against VoiceInk 2.1.

The column name does not depend on which VoiceInk you run, but on which version last opened that particular store — an official store that has not been reopened since the rename still has the old names. So rather than swap one hardcoded name for another, the command now reads `pragma_table_info('ZTRANSCRIPTION')` first and builds the query from the columns the database actually has, taking the first known name for each field. Migrated stores, unmigrated stores and VoiceInk CE all work, and a future rename degrades to a missing badge instead of an empty list. This also removes the heuristic that inferred the schema from the bundle id in a custom path, which is where the wrong assumption lived.

**Copy and paste stopped confirming.** Until #29928 these were `Action.CopyToClipboard` and `Action.Paste`, which show a HUD and close the window for free. That PR moved them to plain `Action` callbacks to keep transcription text out of the serialized render tree — worth keeping — but the confirmation went with it: copying looked like it did nothing, and pasting fired with the Raycast window still frontmost. The helpers in `src/lib/clipboard.ts` restore both behaviours explicitly, so the text still travels through a callback rather than a prop. The comment there records why, so it does not get undone again.

Tested against a real VoiceInk 2.1 store and a synthetic pre-rename store: both return rows with the mode badge populated. A column set without the mode columns still returns rows, with an empty badge. Multi-word search and `LIKE` escaping are unchanged. `npm run build` and `ray lint` are clean, and the build was loaded in Raycast to confirm the list populates and both actions confirm.

## Screencast

N/A

## Checklist

### DIFF
--- a/extensions/voiceink/CHANGELOG.md
+++ b/extensions/voiceink/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## [Fix Empty History and Silent Copy Actions] - {PR_MERGE_DATE}
+
+- Read the transcription table's columns before querying it, instead of assuming them from the selected database source
+- Fix an empty history on current VoiceInk versions, which renamed the `powerModeName` and `powerModeEmoji` columns to `modeName` and `modeEmoji`
+- Confirm the copy and paste actions with a HUD, while still keeping the transcription text out of the render tree
+
 ## [Render Stability] - 2026-08-03
 
 - Prevent large transcription histories from overflowing Raycast's render payload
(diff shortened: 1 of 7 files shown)


pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #30068, opened 2026-08-07
Update karakeep extension

### LINKED ISSUE #30021: [Karakeep] Bookmarks list duplicates every item on reconnect (stale cache + live fetch appended, not replaced)
### Extension

https://www.raycast.com/luolei/karakeep

### Raycast Version

1.104.24

### OS Version

26.5.2 (25F84)

### Description

Every bookmark appears twice in the "Bookmarks" list — once as a row with an empty/placeholder
title, and once with the correct title. This is not a server-side issue: the bookmark count in
Karakeep itself is correct, and the extension's own Karakeep instance shows no duplicates via
the web UI or the API directly.

### Steps To Reproduce

1. Open the Bookmarks command with a working connection at least once, so the list has cached
   locally.
2. Disconnect from the network (or block the Karakeep host).
3. Reopen the Bookmarks command — the list renders once, correctly, from local cache. This
   confirms the cached data itself is not duplicated.
4. Reconnect to the network.
5. The live revalidation fetch completes and every bookmark now appears twice: the previously
   cached row, plus a second row for the same bookmark from the fresh fetch.

### Current Behaviour

Bookmarks list doubles in length after the first successful revalidation following an
offline/online transition, with one stale and one fresh row per bookmark. Half of the rows have an empty title.

### Expected Behaviour

On revalidation, the fresh page(s) fetched from `/api/v1/bookmarks` should replace/dedupe against
the cached data by bookmark `id`, not be appended alongside it.

chrismessina:
## Description

Fixes #30021 — after going offline and reconnecting, every bookmark rendered twice: once from the cached row (with an empty title) and once from the fresh fetch.

**Cause.** `useCachedPromise` *replaces* results on page 0 and *concatenates* on page > 0, and `onLoadMore` increments the page counter immediately — before the request resolves. `useEnsureScrollablePagination` had no error awareness, so opening the command while the server was unreachable still fired a prefetch: the counter advanced to 1, the request failed, and the next successful fetch landed as page 1 and was concatenated onto the rows restored from the on-disk cache.

**Fix.** The prefetch now requires that a request has actually succeeded. Checking `error` alone is not enough — on a cold offline start there is no error yet, only restored cache — so it also gates on whether the rows on screen came from a live request.

Not a regression from 2.4.0: `usePrefetchPagination.ts` is byte-identical to 2.3.2 and the caching options on the bookmarks hook are unchanged, so this reproduces on 2.3.2 as well.

### Changes

- Pagination prefetch no longer runs against an unreachable server, which is what duplicated the list on reconnect (#30021).
- `@chrismessina/raycast-logger` → 1.3.0 (credential-redaction hardening, including credentials embedded in logged URLs). No API change.
- `@raycast/api` → 1.104.24.

### A note on verification

I could not reproduce the original report on my own machine. My Karakeep runs in local Docker, so stopping it triggers the extension's container-recovery screen, which returns before the bookmark list mounts — the duplication path is unreachable in that setup. It needs a server that becomes unreachable *while the command stays open*, which is the reporter's network-loss scenario.

@erireu — if you're able to check this build against your setup, that would be a big help. The fix is derived from the library's pagination internals and reviewed against the source, but your environment is the one that actually exercises it.

## Checklist

- [X] I ran `npm run build` and tested this distribution build in Raycast
- [X] I ran `npm run lint` and `npx tsc --noEmit` — both clean
- [ ] I checked that this change does not break existing commands or remove functionality
- [X] I updated `CHANGELOG.md` per the changelog conventions

### DIFF
--- a/extensions/karakeep/CHANGELOG.md
+++ b/extensions/karakeep/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Karakeep Changelog
 
-## [2.4.1] - {PR_MERGE_DATE}
+## [2.4.1] - 2026-08-08
 
 - Fixed every bookmark appearing twice after going offline and reconnecting. The pagination prefetch ran even when the server was unreachable, which advanced the page counter while the request failed; the next successful fetch was then appended to the cached rows instead of replacing them. The prefetch now waits for a request that has actually succeeded.
 - Updated `@chrismessina/raycast-logger` to 1.3.0, which hardens credential redaction in logs — including credentials embedded in URLs, and values that previously escaped masking by their runtime type.
--- a/extensions/karakeep/CHANGELOG.md
+++ b/extensions/karakeep/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Karakeep Changelog
 
+## [2.4.1] - {PR_MERGE_DATE}
+
+- Fixed every bookmark appearing twice after going offline and reconnecting. The pagination prefetch ran even when the server was unreachable, which advanced the page counter while the request failed; the next successful fetch was then appended to the cached rows instead of replacing them. The prefetch now waits for a request that has actually succeeded.
+- Updated `@chrismessina/raycast-logger` to 1.3.0, which hardens credential redaction in logs — including credentials embedded in URLs, and values that previously escaped masking by their runtime type.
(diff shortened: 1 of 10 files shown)


0xdhrv (approved):
Looks good to me, approved ✅

PR #30069, opened 2026-08-07
[Slack] Add expiration argument to Set Status

foobarnes:
## Description

#28669 made **Set Status** launchable from a deep link with `statusText` and `emoji`, but the status it sets never expires — you still had to open the form to pick a clear time.

This adds a third optional argument, `expiration`, accepting **any number of minutes**, or a keyword:

| Value | Clears |
| --- | --- |
| _omitted_ or `0` | never _(previous behavior)_ |
| `90` | in 90 minutes — any positive integer works |
| `today` | end of today |
| `week` | end of this week |

Example — "Lunch 🍔", clearing in 90 minutes:

```
raycast://extensions/mommertf/slack/set-status?arguments=%7B%22statusText%22%3A%22Lunch%22%2C%22emoji%22%3A%22%3Ahamburger%3A%22%2C%22expiration%22%3A%2290%22%7D
```

No new expiration logic: the value goes to the existing `getExpirationTimestamp`, whose `default` branch already parsed a minute count, and the success toast reports the clear time via the existing `getTextForExpiration`, which already formats arbitrary durations. Deep links and the in-app form resolve expirations through one code path.

### Why a text argument rather than a dropdown

A dropdown would cap this at a handful of presets, but arbitrary expirations are already a first-class concept here — `getDurationOptionFromTimestamp` maps any non-preset timestamp to `custom` for the form's DatePicker. Constraining the deep link to presets would make it strictly less capable than the UI it's meant to automate.

Unparseable input is reported rather than silently accepted. `getExpirationTimestamp` maps `NaN` to `0`, so without a check `expiration=1h` would quietly set a status that never clears. `normalizeExpiration` rejects it up front and shows a failure toast, matching how #28728 handles an unrecognized emoji. This also covers `custom`, which `getExpirationTimestamp` throws on.

### Note

`expiration` on its own (no text, no emoji) doesn't trigger the launch path — it falls through to the normal list rather than expiring an empty status.

## Screencast

No screencast — the change adds one optional argument and is invisible unless it's passed. Launched with no arguments, or with only `statusText`/`emoji`, behavior is unchanged.

## Checklist

- [ ] I ran `npm run build` and [tested this distribution build in Raycast](https://developers.raycast.com/basics/prepare-an-extension-for-store#metadata-and-configuration) — build passes; interactive test still to do, hence draft

### DIFF
--- a/extensions/slack/CHANGELOG.md
+++ b/extensions/slack/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Slack Changelog
 
-## [Add expiration argument to Set Status] - {PR_MERGE_DATE}
+## [Add expiration argument to Set Status] - 2026-08-12
 
 - The **Set Status** command now accepts an optional `expiration` argument — any number of minutes, or `today` / `week` — so a deep link or Quicklink can set a self-clearing status in one step (e.g. `"expiration":"90"` for an hour and a half).
 
--- a/extensions/slack/CHANGELOG.md
+++ b/extensions/slack/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Slack Changelog
 
+## [Add expiration argument to Set Status] - {PR_MERGE_DATE}
+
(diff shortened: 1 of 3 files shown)


0xdhrv (approved):
Looks good to me, approved ✅

PR #30070, opened 2026-08-07
Vikunja: add Quick Add command with Quick Add Magic parsing

onlymykazari:
## Description

Adds a **Quick Add** command to the Vikunja extension that creates a task from a
single line, porting Vikunja's own Quick Add Magic parser so dates, labels,
projects, priorities, assignees and repeat intervals can be expressed inline.

```
Buy milk tomorrow *shopping +Inbox !3 monthly
```

The command shows a live preview of the parsed result while typing, so it is
clear what was detected before anything is created. `Enter` opens a prefilled
form to correct the parse, `Cmd+Enter` creates the task straight away.

The existing `Create Task` command is untouched.

### Why a new command instead of extending Create Task

A single-line DSL and a structured form have different mental models. Mixing
them meant the form's fields and the magic tokens in the title could disagree
about which value wins, and the command arguments shown in the root search bar
(title + project ID) did not fit a magic-input flow. A separate command keeps
both predictable.

## Changes

- Port the upstream parser (`*label`, `+project`, `!priority`, natural dates,
  repeats), including its boundary checks, so strings like `The 9/11 Report` are
  not mistaken for dates
- New **Quick Add Magic** preference to choose the prefix syntax (Vikunja,
  Todoist, or disabled); the placeholder and hints follow the selection
- New **Default Reminder** preference, relative to the due date, skipped when a
  task has no due date to anchor to
- Labels are resolved against existing ones; missing labels are only created on
  submit, so an abandoned flow leaves no stray labels behind
- Assignees are matched against project members, strictly when the search
  returns a single candidate so a partial name cannot bind the wrong person.
  Unmatched names are reported and skipped rather than blocking creation
- Repeat intervals are converted to seconds and can be edited in the form. A one
  month interval also sets `repeat_mode`, matching the web client
- Recent inputs are kept in `LocalStorage` and offered when the search bar is
  empty
- **Fix:** the project and label endpoints were only reading the first page of
  results. On instances with more than 50 labels an existing label looked
  missing and would be created again as a duplicate

## Testing

Set up `vitest` and added 291 tests covering the parser and helpers, including
the date false-positive cases from upstream's own suite.

```
npm test      # 291 passed
npm run lint  # clean
npm run build # distribution build succeeds
```

Also verified manually against a live Vikunja instance: dates, labels, project
resolution, priorities, repeats, reminders and assignees all land correctly, and
the reminder shows as "1 day before due date" in the web UI.

## Note for the maintainer

@flexorflex this is a significant addition rather than a small fix, so per the
[store guidelines](https://developers.raycast.com/basics/prepare-an-extension-for-store#contributing-to-existing-extensions-vs-creating-a-new-one)
it needs your sign-off before it can be merged. Happy to adjust the scope,
split it up, or move it to a separate extension if you would prefer.

## Checklist


pernielsentikaer (changes requested):
Hi @onlymykazari 👋

- Changing project in the confirmation form leaves the previous assignees in state while the new lookup runs, and Create Task stays enabled. Members of project A can be submitted to project B. Please reset assignees when the project changes, and disable submit (or include the lookup in `isLoading`) until it finishes.
- When the member search returns more than one user, `@alex` silently binds the first `includes()` hit in server order. That is especially easy to miss on Create Now, which never shows who was matched. Please accept a partial match only when it is unique among the contains-hits, otherwise leave the name unmatched and surface it on the toast.
- The listing now has 7 metadata screenshots (store max is 6). The two new Quick Add shots are also ~2.7 MB each versus ~650 KB for 1–5. Please keep at most 6, and compress the new PNGs.
- Preference dropdown titles should be Title Case (`No Reminder`, `At the Due Date`, `2 Hours Before`).
- If projects/labels fail to load, the command toasts and then looks empty. Please show an error EmptyView with Retry / Open Preferences.

This is a significant new command, so we also need a sign-off from @flexorflex before merge.

---

Please convert this PR to a draft while working on the requested changes. Once ready, mark it as ready for review again and we will take another look. Thanks!

PR #30071, opened 2026-08-07
[GitHub] Show organization projects in My Projects

### LINKED ISSUE #20343: [GitHub] Cannot see projects from the Org level
### Extension

https://www.raycast.com/raycast/github

### Raycast Version

1.100.3

### macOS Version

15.5

### Description

When creating an issue, I cannot select the project I need, which belongs to the organization and not the repository.

### Steps To Reproduce

1. Create issue
2. In project, observe only the projects from the repo appear, but not the org's.

### Current Behaviour

_No response_

### Expected Behaviour

I can see all projects, including those from the org, if I have the permissions.

mcmillion:
## Description

Fixes #20343.

The "My Projects" command only queries `viewer.projectsV2`, which returns projects owned by your personal account. If your projects all live under an organization — which is how most teams use Projects — the command comes up empty. That's how I found this: my company's boards are all org-owned, so "My Projects" showed me nothing at all.

What this PR does:

- Adds a `getMyProjects` query that fetches projects from your account *and* from each of your organizations, ordered by most recently updated
- Merges the two in `useMyProjects` and sorts the combined list by last update; the open/closed dropdown works the same as before
- Drops the now-unused `projectsV2` field from `getViewer`, so every other command that loads the viewer gets a slightly lighter query

One thing worth flagging for review: I didn't commit a full codegen run for `src/generated/graphql.ts`. The codegen fetches the live schema from GitHub's API, and it has drifted since that file was last generated — a fresh regen produced ~4,600 lines of unrelated churn and flipped the custom scalars from `any` to `unknown`, which breaks `my-stats-menu.tsx`. To keep this diff reviewable I added the types for the new query by hand, matching the existing file's style. Happy to do the full regen (plus the fixes it needs) as a follow-up if you'd rather get the schema current.

## Screencast

No visual changes — the list just has data in it now. Before: empty "My Projects" for anyone whose projects are org-owned. After: org projects show up alongside personal ones, newest first. Verified in `ray develop` against a real account where every project is org-owned, and the raw query against the live GraphQL API.

## Checklist

### DIFF
--- a/extensions/github/CHANGELOG.md
+++ b/extensions/github/CHANGELOG.md
@@ -1,5 +1,9 @@
 # GitHub Changelog
 
+## [Show organization projects in My Projects] - {PR_MERGE_DATE}
+
+- The "My Projects" command now lists projects owned by your organizations alongside projects owned by your user account, sorted by last update. Previously only user-owned projects were shown, so the command was empty for people whose projects all live under an organization.
+
 ## [View Pull Request Diffs] - 2026-08-06
 
 - Added a "View Diff" action to pull requests, showing changed files with per-file patches that load progressively as you scroll.
--- a/extensions/github/src/api/project.graphql
+++ b/extensions/github/src/api/project.graphql
@@ -34,3 +34,22 @@ query projectDetails($nodeId: ID!) {
     ...ProjectFields
   }
 }
+
+query getMyProjects {
+  viewer {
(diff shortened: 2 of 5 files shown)


0xdhrv (commented):
Hi @mcmillion 👋

Thanks for the contribution 🔥

---

PR #30075, opened 2026-08-08
Add portreaper extension

fanhefeng:
## Description

**Portreaper** finds orphaned dev-server processes squatting on your ports and lets you reap them from Raycast.

You kill a terminal, but the `vite` / `node` / `cargo run` it launched keeps running — reparented to the OS, still holding port 3000. Next time you `npm run dev` the port is "already in use" and you have no idea which ghost to kill.

This is not a generic port viewer. Its job is to **decide which listeners are orphaned dev-server zombies**, so you act on a verdict instead of a raw process list:

- **Every TCP listener**, plus **orphaned dev processes that hold no port at all** — a leftover `electron-vite` main process, adopted by launchd after its parent `node` died, is invisible to a port scan but is exactly the kind of residue worth clearing.
- **A verdict with its evidence.** Suspects are tiered `confirmed` / `likely` / `possible`, and each row shows the signals behind the call (reparented to PID 1, launcher chain ends at a dead shell, dead terminal session, dev-server command line, duplicate instance of the same project).
- **Exemptions applied automatically** — anything managed by `launchd`, `brew services`, or `pm2`, and anything installed in a standard location, is never flagged.
- **Star to exempt permanently.** A daemon you detached on purpose is behaviorally identical to an accidental zombie; star it once and it stops being flagged.

Terminating always asks for confirmation, and the engine **re-checks the process creation time immediately before killing** — if it moved, the kill is refused. That closes the window where a PID gets recycled between looking at the list and pressing Enter.

The classification engine is shared with the [Portreaper macOS menubar app](https://github.com/fanhefeng/portreaper); both frontends read the same whitelist file, so a star added here shows up in the desktop app's next scan and vice versa. The desktop app is optional — this extension works standalone.

### About the binary dependency (please read)

The classification logic lives in a small Rust CLI, `portreaper-cli`. On first run the extension downloads it and **verifies it against the project's published SHA-256 checksum**; a binary that fails verification is deleted, never executed. Nothing is installed system-wide — it lives in the extension's own support directory, and the UI states plainly what it is doing while downloading.

This follows the ["Binary Dependencies" guidance](https://developers.raycast.com/basics/prepare-an-extension-for-store) — downloaded from a trusted location with hash verification, rather than bundling an opaque binary:

- **Source**: the project's own GitHub Releases (`fanhefeng/portreaper`), built by a public GitHub Actions workflow — traceable build sources, not an opaque blob.
- **Integrity**: SHA-256 checked against the release's published `portreaper-cli-SHA256SUMS`; mismatch ⇒ file removed and an error surfaced.
- **Self-healing**: if the cached copy later fails to execute (truncated download, arch change, lost exec bit) it is removed and re-fetched once.
- **No user action required** — the guidance explicitly asks extensions to avoid making users perform additional downloads.

Same pattern as recently approved extensions: [glean-search #28995](https://github.com/raycast/extensions/pull/28995), [lumen #28909](https://github.com/raycast/extensions/pull/28909), and `speedtest`.

`platforms` is `["macOS"]` only: the CLI's Windows leg has no manual QA yet, so the extension does not claim support it cannot back.

## Screencast

Screenshots are in `metadata/` (light theme, 2000×1250):

1. Filtered list — `Suspects` / `Healthy` grouping with confidence tags
2. Detail panel — the signals behind a verdict, PID, category, ports, uptime
3. Action panel — Terminate / Force Kill styled as destructive, plus Star
4. Confirmation dialog — names the process, PID and port before terminating

The suspects shown are three deliberately orphaned `node` dev servers created for the capture; the list is filtered to `node` so the screenshots do not expose the unrelated applications running on the author's machine.

## Checklist

Additionally verified on a real machine: first-run download + SHA-256 verification end to end (all cached and local copies removed to simulate a fresh install — the re-downloaded binary's digest matched the published manifest byte for byte), scan and grouping, detail panel, action panel, terminate confirmation, search filtering, and the no-port orphan case.


fanhefeng:
Thanks for the review — all three findings were valid and are fixed in `22a742f0`.

**1. Schema-incompatible cached CLI could not be replaced.** Correct, and this one was the real bug. `verifyCli()` only runs `--version`, so a stale-but-runnable managed copy passes; `scan()` then rejects the schema and Retry keeps selecting the same binary — a dead end from the user's side. Now a `SchemaMismatchError` fetches a fresh copy, reusing the existing `replace an unusable copy` path. Two deliberate limits: it only replaces **our own downloaded copy** (a user-specified path is never deleted), and it retries **once** — if the fresh download still mismatches, the extension and the published CLI genuinely are different generations, and that must be reported rather than looped on.

**2. Terminate/Force Kill shown for entries without an identity token.** Correct. The maintainer notes already said those rows should not offer a kill entry point, but the code never implemented it — the actions rendered identically to working ones and could only produce a failure toast. The whole section is now gated on `start_unix != null`. The check inside `doKill()` stays: this gate is about not *presenting* the action, that one is about never *proceeding* even if some future call site presents it.

**3. Preference title casing.** Fixed: `Portreaper CLI Path`. The README and maintainer notes referenced the old label by name, so those were updated too — docs pointing at a label that no longer exists in the UI are worse than no docs.

Verified with `tsc --noEmit`, `ray lint` and `ray build`.

Unrelated but worth noting for reviewers: the engine this extension downloads was released as v0.9.0 today, which fixes a cross-frontend bug where a star added here stayed invisible to the desktop app — the starred process kept counting as a suspect there and stayed in the one-click sweep target set. The extension pulls from `releases/latest`, so no change was needed on this side.

pernielsentikaer:

- **Port Manager** (`ports`, 1,285 installs by @diegoleteliers10) — Monitor and manage open ports on your system with the ability to kill processes using specific ports.
  <sub>similarity 0.56</sub>

fanhefeng:
Thanks — the README now has a **"How this differs from a port viewer"** section that names Port Manager directly and links to its Store page.

Two things worth putting in front of the human reviewer:

**The two extensions never meet.** Port Manager declares `"platforms": ["Windows"]`; Portreaper declares `"platforms": ["macOS"]`. A macOS user does not see Port Manager in the Store, and vice versa. That is not an argument for accepting this one — just a fact that bounds the overlap.

**The overlap that does exist is the list, not the product.** Both render listening TCP ports and offer to kill what is behind one. Everything this extension is for sits after that point:

- **A verdict, not a list.** Each row is classified `confirmed` / `likely` / `possible` from orphan signals — reparented to PID 1, launcher chain terminating at a dead shell, dead tty session, duplicate instance of the same project — with hard exemptions for anything `launchd`, `brew services`, or `pm2` already manages. A dev server with a live terminal behind it is never flagged. That classification, not the port list, is the deliverable.
- **Processes with no port.** It also lists orphaned dev processes holding no port at all — an `electron-vite` main process adopted by launchd after its parent `node` died, for example — which a port scan cannot reach by construction.
- **Guarded termination.** The engine captures each process's creation time during the scan and re-reads it in the instant before signalling; a drift means the PID was recycled and the kill is refused. After the signal it re-scans until the process is actually gone (~2.5s deadline) and reports "still running" honestly rather than painting a checkmark over a survivor. A suspended process (Ctrl-Z) also gets a SIGCONT right after SIGTERM — a caught SIGTERM is never processed while stopped, which is the single most common "I killed it and nothing happened" case.
- **Shared engine.** The classification lives in a Rust binary that the Portreaper desktop app drives as well, over one shared whitelist file, so a star added here is honoured in the desktop app's next scan.

The README section says all of this in user terms, including the line that matters most to someone choosing between the two: if you only want to free port 3000, a plain port viewer is the simpler tool.

fanhefeng:
Updated again — a correction and a rewrite of the section from my last comment, plus a new icon.

**The Port Manager link in that section was broken, and I should flag that directly.** I built the URL from the GitHub handle in the duplicate-check comment (`diegoleteliers10`), but Store pages use the `author` field — the correct one is `raycast.com/dleteliers_/ports`. So when I said the README "links to its Store page," anyone who clicked it got a 404. Fixed and verified.

**The section itself is rewritten.** The first version conceded twice — at the top ("if that is your whole problem, a plain port viewer is the simpler tool") and again at the bottom ("if what you want is a list of ports, this is a heavy way to get one") — with a feature comparison in between. That reads as *"yes, this overlaps, but it also does these extra things,"* which does not actually answer the duplicate question.

It now frames the two as answering **different questions**:

> A port viewer answers *"what is on port 3000?"* — you already know something is in your way, and you need to find it and free it.
>
> Portreaper answers *"what is still running that nobody is responsible for?"* — you are not looking up a port you already care about; you are finding out what the last three days of development left behind.

That framing does more work than the feature list did. The clearest example: Portreaper lists orphaned dev processes **holding no port at all** (an `electron-vite` main process adopted by launchd, listening on nothing). Under "free this port" that is out of scope by definition; under "what did I leave behind" it is the whole point. The concession is now half a sentence, and the section closes on the distinction rather than on self-deprecation.

**Also in this push: a new extension icon**, matching a redesign of the desktop app's icon.

extensions/portreaper/metadata/portreaper-1.png
0xdhrv, line 1: **issue:** check the metadata screenshot size and padding

Can you make sure the metadata screenshot is the correct size and has the proper padding? Right now it looks cropped, and the corner is cut off weirdly.

Right now it looks cropped and has a side border. Please use the "Capture Window" command in Raycast.

Reference:

extensions/portreaper/metadata/portreaper-2.png
0xdhrv, line 1: **issue:** check the metadata screenshot size and padding

Can you make sure the metadata screenshot is the correct size and has the proper padding? Right now it looks cropped, and the corner is cut off weirdly.

Right now it looks cropped and has a side border. Please use the "Capture Window" command in Raycast.

Reference:

extensions/portreaper/metadata/portreaper-3.png
0xdhrv, line 1: **issue:** check the metadata screenshot size and padding

Can you make sure the metadata screenshot is the correct size and has the proper padding? Right now it looks cropped, and the corner is cut off weirdly.

Right now it looks cropped and has a side border. Please use the "Capture Window" command in Raycast.

Reference:

extensions/portreaper/metadata/portreaper-4.png
0xdhrv, line 1: **issue:** check the metadata screenshot size and padding

Can you make sure the metadata screenshot is the correct size and has the proper padding? Right now it looks cropped, and the corner is cut off weirdly.

Right now it looks cropped and has a side border. Please use the "Capture Window" command in Raycast.

Reference:

0xdhrv (changes requested):
Hey @fanhefeng 👋 

---

extensions/portreaper/src/search-ports.tsx
```diff
@@ -0,0 +1,1064 @@
...
+        // 用户视角是死循环。取一份最新的再试 —— 与上面「换掉不可用副本」同一套路。
+        //
+        // 只换**我们自己下载的那份**（用户显式指定的路径不擅自删），且**只换一次**：
+        // 新下的仍对不上，说明扩展与已发布的 CLI 确实不同代，那是真错误，
+        // 必须如实报给用户，绝不无限重下。
+        if (!(e instanceof SchemaMismatchError)) throw e;
+        if (cliPath !== installedCliPath(supportPath)) throw e;
+        if (!stale()) setState({ kind: "installing", step: "Updating the engine…" });
+        await rm(cliPath, { force: true });
+        cliPath = await install();
```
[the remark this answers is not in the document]
fanhefeng: Valid — fixed in `5eda9278`.

The request id only ever guarded React state; the two `rm(cliPath)` calls in the recovery branches were unguarded filesystem mutations, so a stale `load()` could delete the copy a newer one had just installed, and any scan inside that window failed with `ENOENT`.

Two changes:

1. **The `rm()` before each reinstall is gone.** `installCli()` already replaces the binary atomically (`writeFile` to a temp name → `chmod` → `rename` over the destination), so deleting first bought nothing and was the only thing that created a window with no CLI on disk.
2. **Downloads are single-flight.** `installCliOnce()` shares one in-flight promise between concurrent callers, so overlapping recoveries perform exactly one download and one `rename`; later callers just await the same result.

The "replace only our own downloaded copy, never a user-specified path" and "retry once, then report" rules are unchanged.

extensions/portreaper/metadata/portreaper-1.png
fanhefeng: Retaken with Raycast's Capture Window command in `4343d7b0` (extension loaded in development mode, `Save to Metadata`). The old image was a composited window grab, which is where the side border and clipped corners came from.

extensions/portreaper/metadata/portreaper-2.png
fanhefeng: Retaken with Raycast's Capture Window command in `4343d7b0` (extension loaded in development mode, `Save to Metadata`). The old image was a composited window grab, which is where the side border and clipped corners came from.

extensions/portreaper/metadata/portreaper-3.png
fanhefeng: Retaken with Raycast's Capture Window command in `4343d7b0` (extension loaded in development mode, `Save to Metadata`). The old image was a composited window grab, which is where the side border and clipped corners came from.

extensions/portreaper/metadata/portreaper-4.png
fanhefeng: Retaken with Raycast's Capture Window command in `4343d7b0` (extension loaded in development mode, `Save to Metadata`). The old image was a composited window grab, which is where the side border and clipped corners came from.

fanhefeng:
Thanks @0xdhrv — all four screenshots are retaken and the PR is ready for another look.

**Screenshots (`4343d7b0`).** You were right that the previous images were not from Window Capture: they were window grabs composited onto a gradient by a local script, which is what produced the side border and the clipped corners. All four `metadata/portreaper-{1..4}.png` are now taken with Raycast's own **Capture Window** command (2000 × 1250, `Save to Metadata`), with the extension loaded in development mode. Same content as before so the sequence still reads top to bottom:

1. Filtered list, details collapsed — `no port` / `stopped` / `dup of <pid>` / `confirmed` badges in one frame
2. List + detail panel on the suspended row — command, the "stopped" explanation, verdict, reasons, `State: TN · stopped`
3. ⌘K action panel — Danger Zone (Terminate / Force Kill) and Inspect
4. The terminate confirmation — `PID … · port :4321 · It is suspended; terminating resumes it so it can shut down.`

As before, the only processes on screen are throwaway demo dev servers created for the capture; the list is filtered to `dev-server` so nothing else from the machine appears. The compositing script is deleted from the project so this cannot regress.

**Also in this push:**

- greptile's remaining P1 (concurrent recovery could delete the freshly installed CLI) — fixed in `5eda9278`, details in [that thread](https://github.com/raycast/extensions/pull/30075#discussion_r3878247840): the pre-reinstall `rm()` is gone (`installCli` already replaces atomically via `rename`) and downloads are single-flight.
- The checksum is now verified in memory before anything touches disk, and a mismatch is retried once so a release landing between the two fetches is not reported to the user as a security event. A CLI scan failure now gets its own error state instead of leaking raw stderr.

`npm run build`, `npm run lint` (including the metadata validation) and `tsc --noEmit` all pass.

fanhefeng:
@0xdhrv gentle ping — the four screenshots were retaken with Raycast's Capture Window command in `4343d7b0` (2000 × 1250, no border or clipped corners), each thread has a reply, CI is green, and the PR is back in **Ready for review**. Whenever you have a moment for another look — thanks!

0xdhrv (approved):
Looks good to me, approved ✅

PR #30077, opened 2026-08-08
feat(renaming): add undo history for rename operations

TurboCoder13:
### 🤖 Disclosure

## Description

This PR makes renaming reversible: a new **Rename History** command records every rename batch and
can undo it, restoring the original filenames. Undo works at three levels — roll back to a point in
time, undo one whole operation, or undo a single file — with a conflict preview before anything is
reverted and per-file status tracking with retry. No new runtime dependencies, no changes to how the
existing commands rename.

### Features

- **Rename History command** — browse the 25 most recent rename operations with timestamps, stored
  locally in Raycast (`LocalStorage`); history never leaves the machine
- **Every command records successful renames** — Rename File(s), Replace File(s) Characters and
  Advanced Batch Rename each save exactly the files that moved (a partial failure records only the
  successes, so an undo can never try to reverse a rename that didn't happen), then open Rename
  History so the completed batch is immediately reviewable and undoable
- **Undo to a point in time** — pick a state to return to and every operation back to it is
  reversed, instead of undoing one batch at a time
- **Undo one operation or one file** — the detail view shows an operation's full old → new mapping
  grouped by status (renamed, undone, could not undo) and can undo the whole entry or a single file,
  leaving the rest renamed
- **Conflict preview before undoing** — the confirmation states how many files can be restored and
  why the rest cannot (moved or deleted since, replaced by a different file, or the original name
  taken again); conflicted files are skipped rather than failing the batch, marked with the reason,
  and can be retried once resolved
- **File previews in the detail panel** — images render inline, text files show their first lines,
  plus status, directory, size and modified date

### Technical Highlights

- All rename/undo state machinery lives in tested `src/lib/` modules (`history.ts`,
  `file-preview.ts`); vitest harness grows to **181 tests**
- History entries are identified by timestamp rather than list index, so an undo can never target
  the wrong entry after a newer rename is recorded or history is trimmed
- Undo verifies filesystem identity: each recorded rename stores the file's `dev:ino` (via `lstat`,
  so a symlink is identified by the link itself, dangling links included), and undo refuses to move
  anything else that later appears at the recorded path
- Undo mirrors the batch renamer in reverse: parent directories are restored before their children,
  and each child's recorded path is remapped through completed restores so files carried along by a
  parent's restore are found where they actually sit — and a case-only rename (`foo.txt` →
  `Foo.txt`) is recognized as its own file on case-insensitive volumes, not an occupying conflict
- Undo results are merged into the freshest stored history at write time, so a rename recorded by
  another command instance mid-undo survives instead of being overwritten by a stale snapshot;
  every history write is verified after the fact and re-merged if a concurrent writer interleaved,
  and a write that cannot be confirmed stored is reported as unsaved rather than acknowledged
- Text previews are wrapped in a code fence computed to exceed any backtick run in the content, so
  file content can never escape the fence and render as markdown
- Failure honesty throughout: a history-write failure never misreports a completed rename as failed,
  and a storage failure after files were already restored on disk says exactly that instead of
  claiming the undo failed
- File stats and previews are only read while the detail panel is visible, and memoized so list
  re-renders don't hit the filesystem
- No curated file-type lists: text-ness is sniffed from content (NUL/UTF-8 validity checks), list
  rows use the system's own file icons via `fileIcon`, and the one extension set that remains is
  the formats Raycast's markdown renderer can display inline — a renderer bound, not a file-type
  taxonomy
- The vitest coverage config now measures all of `src/**` with honest, enforced thresholds — a
  separate, higher floor protects the `src/lib/` layer where the correctness logic lives — and
  `npm run test:coverage` exits 0 and is usable as a gate
- No new runtime dependencies; uses `LocalStorage` from `@raycast/api` and `fs`/`fs/promises` only

## Screencast

Fresh screenshots in `metadata/` — the two existing captures retaken, plus the Rename History list
and an operation's detail view with inline preview.

## How to Test

No special hardware or credentials needed — everything runs on local files.

1. `npm install && npm run dev` in `extensions/renaming`
2. Create sample files: `mkdir -p ~/Downloads/samples && cd ~/Downloads/samples && touch IMG_{001..010}.jpg notes.txt`
3. Select the files in Finder, run **Rename File(s)**, enter a new base name and rename — Rename
   History opens with the recorded batch
4. Press <kbd>⏎</kbd> on the entry to view per-file details (old → new mapping, previews), then undo a
   single file with <kbd>⌘Z</kbd> — only that file is restored
5. Back in the list, press <kbd>⌘Z</kbd> on the entry — the confirmation states how many files will be
   restored; confirm and check the original names are back in Finder
6. Conflict case: rename a batch, then manually create a file with one of the original names and
   undo — that file is skipped, marked "Could not undo" with the reason, and can be retried after
   deleting the conflicting file
7. Run several batches, then <kbd>⌘Z</kbd> an older entry in the list — every operation back to that
   point is reversed after one confirmation

## Checklist

- [x] `npm run lint` passes


TurboCoder13:
## File Breakdown

| Category | Count | Description |
|----------|------:|-------------|
| **Implementation** | 12 | History core (`lib/history.ts`), record-and-navigate flow (`lib/history-nav.ts`), file previews (`lib/file-preview.ts`), the Rename History command + detail views, and recording hooks in the three existing commands |
| **Tests** | 5 | History suite, history-nav suite, file-preview suite, API/jsx-runtime test mocks |
| **Images** | 4 | `metadata/` screenshots — 2 retaken, 2 new (history list, detail view) |
| **Config** | 2 | `package.json` (register the command), vitest coverage include + enforced thresholds |
| **Docs** | 2 | README (documents all four commands), CHANGELOG |
| **Total** | **25** | |

### Distribution

```
Implementation ███████████████████         48.0%
Tests          ████████                    20.0%
Images         ██████                      16.0%
Config         ███                          8.0%
Docs           ███                          8.0%
```

### Implementation layout

| Area | Files | Purpose |
|------|------:|---------|
| `src/lib/` | 5 | `history.ts` (storage, identity-verified three-level undo, conflict preview, concurrent-write merge), `history-nav.ts` (record batch + open history), `file-preview.ts` (content-sniffed previews, stats), `constants.ts`, `logger.ts` |
| `src/` | 4 | `history.tsx` (the new command); `index.tsx`, `replace.tsx`, `advanced-rename.tsx` (record successful renames, open history after a clean batch) |
| `src/components/` | 2 | `HistoryDetailView.tsx` (status-grouped per-entry view), `HistoryFileItem.tsx` (per-file row + detail panel, system file icons) |
| `src/types/` | 1 | `RenameHistoryEntry`, `HistoryOperation` |

25 files, +2,577/−49.

pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #30078, opened 2026-08-08
Update file-tidy extension

fanhefeng:
## Description

Maintenance update for **File Tidy** — internal cleanup, **no user-facing behaviour change**. No command, preference, form or list rendering was touched.

- **Collision naming now has a single definition.** Planning and execution each carried their own copy of the `" (n)"` suffix loop. They agreed, but nothing *kept* them agreeing — and the plan preview promises the exact name a file will land under, so an edit to one side would have silently broken that promise. Both now resolve through one `firstFreeName` helper in `plan.js`.
- **The `.tidy` bookkeeping directory is referenced through a `TIDY_DIR` constant** where it used to be spelled out in logic (the scan's skip list, the plan's segment validation). User-facing copy keeps the literal spelling.
- **Every core module now carries its type declaration.** `date.d.ts` was missing entirely, and `TIDY_DIR` / `tidyPath` were undeclared in `config.d.ts`.
- **`expandTilde` moved out of core** into the CLI that was its only caller — the extension's form and file pickers always hand core absolute paths, so core no longer exports a helper for one adapter.

The remainder of the diff is the test suite: every `mkdtemp` directory is now registered and removed in an `after()` hook (leftovers had been piling up in the OS tmpdir), and the test script globs `*.test.js` so a newly added test file cannot be silently skipped.

## Screencast

No new screenshots — nothing visible changed. The three images under `metadata/` were reshot for the previous release and still match exactly what the extension renders.

## Checklist

### Verification

`ray build -e dist` compiles clean, `ray lint` passes, and the 54-case suite covering archiving, dedup, undo and the detection passes is green — including the existing tests that assert the ` (n)` collision suffix, which is what pins the refactor above to identical behaviour.

### DIFF
--- a/extensions/file-tidy/CHANGELOG.md
+++ b/extensions/file-tidy/CHANGELOG.md
@@ -1,6 +1,6 @@
 # File Tidy Changelog
 
-## [Maintenance] - {PR_MERGE_DATE}
+## [Maintenance] - 2026-08-08
 
 - Internal cleanup with no change to how the commands behave. The ` (n)` suffix used for a name collision is now produced in a single place, so the name shown in the plan and the name written to disk cannot drift apart as the code changes.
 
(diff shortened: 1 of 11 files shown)


0xdhrv (approved):
Looks good to me, approved ✅

PR #30084, opened 2026-08-08
Add Thesaurus Rex extension

invm:
## Description

Thesaurus Rex — an offline thesaurus and dictionary. Definitions, examples, synonyms, antonyms and pronunciation for any word, with copy and paste-to-frontmost-app actions.

Everything runs on the Mac. There is no API, no key and no account: on first use the extension downloads [Open English WordNet 2024](https://en-word.net/) (12.9 MB, CC BY 4.0) from a dedicated **Manage Dictionaries** command, verifies it against a pinned SHA-256 before parsing a byte of it, and transforms it on-device into a SQLite database at `environment.supportPath`. No lookup ever leaves the machine, and the data is deletable at any time from the same command.

**Commands**

- **Look Up Word** — definitions, then synonyms, then antonyms, 3 rows per section with `Show N more`
- **Search Synonyms** / **Search Antonyms** / **Define Word** — one kind each, 25 rows
- **Manage Dictionaries** — download or delete the data, with progress and installed entry counts

**Details worth flagging for review**

- **The 12.9 MB runtime download** is deliberate and opt-in: bundling the dataset would bloat the extension for everyone, and the alternative is a network call per lookup. Until it is downloaded the empty view explains why and offers the download on ⏎.
- **`node:sqlite`** is used for storage. It is an experimental builtin, but unflagged on Raycast's Node runtime; `ray build` leaves `require("node:sqlite")` external and it works in the built extension. The `ExperimentalWarning` appears in the `ray develop` console only and never reaches the UI.
- Synonyms are ranked by how many senses they share with the query word rather than by file order, which is what keeps `down` from suggesting `down feather` first.
- Attribution is per row (`Open English WordNet`), with the CC BY 4.0 licence named in the README and in Manage Dictionaries.

## Screencast

Look Up Word

Search Synonyms

Search Antonyms

## Checklist


invm:
Thanks — both review points are addressed in d008338.

**Concurrent installs (valid, fixed).** Each command runs in its own process against the one database, so Manage Dictionaries and a search command's empty view could both start an install and collide over SQLite locks and the fixed staging table names. An install now takes a lock row in `meta` inside an `IMMEDIATE` transaction; the second caller is told *"A dictionary download is already running"* instead of failing confusingly. A lock left behind by a crash expires after 15 minutes, which is comfortably past the 5 minute download timeout plus the ~10s build. Covered by a test for the held, released and stale cases.

**Command title case (valid, fixed).** `Look up Word` → `Look Up Word`, in `package.json` and the README.

Also replaced the screenshots: the previous ones failed `metadata-images` because they were desktop captures with the window off-centre. These are Window Capture output, and the listing now leads with Look Up Word rather than a single-section command.

pernielsentikaer:

- **Synonyms** (`synonyms`, 846 installs by @remorses) — Find synonyms for words and translate if necessary
  <sub>similarity 0.61</sub>
- **OneLook Thesaurus** (`onelook-thesaurus`, 886 installs by @FelipeCortez) — A powerful English thesaurus and brainstorming tool that lets you describe what you're looking for in plain terms.
  <sub>similarity 0.60</sub>
- **Word Search** (`word-search`, 24,582 installs by @RishabSwift) — Search for synonyms, antonyms, spellings, rhymes and more
  <sub>similarity 0.60</sub>
- **Thesaurus** (`thesaurus`, 7,955 installs by @ABukSwienty) — Search Merrian-Webster's Thesaurus
  <sub>similarity 0.58</sub>

invm:
All four of these are online. Thesaurus Rex is offline only. It downloads Open English WordNet 2024 once (12.9 MB) into a local SQLite database, and after that every lookup is a local query. Nothing hits the network at search time and no word you look up leaves your machine.

Synonyms is AI generated plus translation, which is a different thing to a fixed dataset. No API key, no latency, same answer every time.

OneLook Thesaurus is a remote service for going from a concept to a word. Needs a connection, depends on their uptime and rate limits, and your queries go to a third party.

Word Search covers more ground (rhymes, spelling) but has the same online dependency. Mine is narrower and deeper on the thesaurus side: synonyms ranked by how many senses they share with your word, inflection fallback so running finds run, and definitions, examples and IPA in the same view.

Thesaurus uses Merriam-Webster's API, which currently 403s on every request, so it doesn't work. Thats was my thesaurus, thats the reason I made an offline one.

Thanks for your commet.

0xdhrv (approved):
Looks good to me, approved ✅

PR #30085, opened 2026-08-08
Update file-tidy extension

fanhefeng:
## Description

**File Tidy** already flagged near-duplicates and visually similar images without moving them anywhere special — both passes are heuristics rather than proof, so the files are archived normally and the user decides what to delete. That stays exactly as it was. What changes is that the grouping now outlives the run.

Until now those flags existed only while the plan was on screen. Once the run finished, the flagged files sat in their ordinary archive folders with nothing on disk connecting them, so acting on a flag meant redoing the comparison by hand. Byte-identical duplicates never had this problem — they land in `ft_Duplicates/` with a manifest beside them.

A run that flags anything now appends a report to `.tidy/similar.md` in the destination, listing each group and where its files ended up, and the success toast offers to open it. The report sits beside the run records rather than in the archive tree: a folder of its own would imply a verdict neither pass is confident enough to make.

Two implementation details worth calling out:

- **Peers are recorded before anything moves**, so they are rewritten to their post-move locations before being written out — otherwise every path in the report would name a file that is no longer there. This is the same rewrite `keeperPath` already needed for the duplicates manifest; both now go through one helper.
- **Paths are written relative to the destination.** The report lives inside the archive, and repeating an absolute prefix on every line buried the file names the reader is actually scanning for. The formatter receives `destDir` for exactly that, so the copy stays in the adapter and core keeps returning data only.

Where files go is unchanged, deliberately: the perceptual threshold is an empirical value, and letting it move files would turn a misjudgement from one extra line of output into a photo that is no longer where the user left it.

## Screencast

No new screenshots. The list and form layouts are unchanged — the additions are a secondary action on the success toast, shown only when a run flagged something, and a sentence in the "Smart checks" info tooltip. The three images under `metadata/` still match what the extension renders.

## Checklist

### Verification

`ray lint` and `ray build -e dist` both pass. The suite is green at 57 cases, three of them new and covering this feature: the report lands at `.tidy/similar.md` and names post-move locations rather than the source paths it was built from, a peer already sitting in the destination keeps its own path while a second run appends instead of overwriting, and a run that flags nothing writes no report at all.

The pipeline was also exercised end to end against real fixtures — one image re-exported at three sizes and qualities, plus a versioned filename pair — confirming that both the perceptual and the filename-heuristic passes reach the report and that the paths in it resolve to the archived files.

### DIFF
--- a/extensions/file-tidy/CHANGELOG.md
+++ b/extensions/file-tidy/CHANGELOG.md
@@ -1,6 +1,6 @@
 # File Tidy Changelog
 
-## [Look-Alike Report] - {PR_MERGE_DATE}
+## [Look-Alike Report] - 2026-08-09
 
 - Near-duplicates and similar images are still only flagged and still archived normally — but the grouping no longer disappears when the plan closes. A run that flags anything now writes `.tidy/similar.md` in the destination, listing each group and where its files ended up, so you can work through them later instead of having to act while the plan is open. The success toast offers to open it.
 
--- a/extensions/file-tidy/CHANGELOG.md
+++ b/extensions/file-tidy/CHANGELOG.md
@@ -1,5 +1,9 @@
 # File Tidy Changelog
 
+## [Look-Alike Report] - {PR_MERGE_DATE}
+
+- Near-duplicates and similar images are still only flagged and still archived normally — but the grouping no longer disappears when the plan closes. A run that flags anything now writes `.tidy/similar.md` in the destination, listing each group and where its files ended up, so you can work through them later instead of having to act while the plan is open. The success toast offers to open it.
+
 ## [Maintenance] - 2026-08-08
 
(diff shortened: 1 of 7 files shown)


fanhefeng:
Fixed in 2e5324f — thanks, this was a real one.

Both post-move appends (the duplicate manifest and the look-alike report) now return their failure as a `REPORT_WRITE` entry in `reportErrors` instead of throwing. Every file has already moved by that point, so the run stays a success: the toast keeps its “Done” title and its undo hint, and names the record that could not be written. `similarReportPath` stays `null` on failure, so the toast never offers to open a report whose block was never appended. A test covers the case the review flagged as untested — a run where both records fail after every move has completed.

0xdhrv (approved):
Looks good to me, approved ✅

PR #30098, opened 2026-08-09
Update system-information extension

muhammadrizo-y:
## Description

- Add support for Windows platform
- Convert main command to TypeScript
- Remove Swift dependency in favor of `systeminformation`
- Show installed memory and physical disk info on Windows
- Add Windows build number and kernel version

## Checklist


pernielsentikaer (changes requested):
Hi @muhammadrizo-y 👋

- Please keep the Swift storage helper on macOS. `volumeAvailableCapacityForImportantUsage` was added in March 2025 so the Storage row matches System Settings on APFS (purgeable space). `si.fsSize()` is `statfs`/`df` and will look wrong again for existing Mac users. Windows cannot use Raycast Swift helpers, so the right split is: `swift:../swift` on macOS, `systeminformation` only on Windows. Restore `swift/`, `swift.d.ts`, and the import in both the view and `src/tools/check-storage.ts`.

- Please restore (or downscale) the four wallpaper PNGs. `macos_monterey.png` / `macos_sequoia.png` / `macos_sonoma.png` / `macos_ventura.png` went from ~366 KB combined to ~2.96 MB, all 1024×1024. They are only used as the macOS OS-row list icon (`releaseImage()`); Windows uses `Icon.Monitor` and never reads them. 512×512 or the previous files is enough — please do not ship wallpaper-resolution assets as list icons.

Smaller polish, happy to take in the same pass: the Windows storage title still uses `diskLayout()[0]` (e.g. “512 GB SSD”) while the numbers sum every `fsSize()` volume, so a second disk or a mapped drive makes the title lie; drop the internal changelog bullets (TypeScript conversion, Swift removal, dependency bump) and only keep the Windows-facing ones; if Swift stays, also keep the root `Package.resolved` — if it does not, delete that leftover SPM lockfile.

Appreciate the two Greptile follow-ups — the Windows total/free scope mismatch is resolved on HEAD.

---

Please convert this PR to a draft while working on the requested changes. Once ready, mark it as ready for review again and we will take another look. Thanks!

pernielsentikaer:
Thanks for putting Swift and the original macOS version icons back.

The view path is correct again: macOS uses `swift:../swift`, Windows stays on `systeminformation`. One piece from last time is still missing.

1. `src/tools/check-storage.ts` never came back to Swift. On macOS it still does `si.fsSize()`, picks `/`, and reports `statfs`/`df`. That will disagree with System Settings and with the Storage row you just fixed. Please restore `import { getStorageInfo } from "swift:../../swift"` on the macOS branch (same as `main`) and keep `systeminformation` only for Windows.

Happy to approve once that tool path matches the view.

Small leftover if you are already in there: the Windows storage title still takes `type` from `diskLayout()[0]` while the GB number now sums every `fsSize()` volume, so a second disk or a mapped drive can still make the title lie.

---

pernielsentikaer (changes requested):
Hi @muhammadrizo-y 👋

Thanks for restoring Swift in the storage tool 💪

The Swift helper already returns GB, but `formatStorage()` divides those values by `1e9` again. This makes the macOS AI tool report `0.00 GB` for total, used, and available storage.

Could you normalize the Swift values to bytes before the shared formatting in `src/tools/check-storage.ts`?

```ts
totalBytes = info.total * 1e9;
usedBytes = info.used * 1e9;
freeBytes = info.free * 1e9;
```

---

pernielsentikaer:
Hi @muhammadrizo-y 👋 Thanks for fixing the storage conversion 💪 Optional polish: when `fsSize()` returns no volumes, the storage tool reports `NaN%`. Could you handle zero total capacity as unavailable?

pernielsentikaer (approved):

PR #30116, opened 2026-08-10
[Antinote] major update with Antinote v2.* (beta) support

sc0:
## Description

A major update to Antinote extension. It includes 3 new commands:
- Search slotted notes (beta-specific)
- Execute timer commands (beta-specific)
- Create new note with selection (available for both beta and current stable)

The update is backwards compatible (for v1 supported commands) and provides a setting to select if you'd rather have it target the beta or the stable version if both are present.

All the commands' specifics are based on the original Antinote behavior that would be triggered within the app. Especially timer commands, the label parser is closely modeled after the original. This may lead to some potentially unintuitive suggestions, like any command starting with `s` will stop a timer. This is exactly how the original app behaves and this extension's intention is to respect that.

## Checklist


0xdhrv (approved):
Looks good to me, approved ✅

PR #30125, opened 2026-08-10
Update proxmox extension - add multiple servers support

### LINKED ISSUE #27260: [Proxmox] Allow multiple hosts
### Extension

https://www.raycast.com/CzBiX/proxmox

### Description

While the current extension allows to add one host (which shows all VMs/storage in the cluster), it would be nice to have the ability to add other non-clustered PVE nodes

Everyone with multiple non-clustered proxmox nodes

### Anything else?

_No response_

kzaoaai:
## Description

Fixes #27260

Adds support for multiple Proxmox servers:

- New `Manage Servers` command to add, edit or remove servers. Additional servers are stored in `LocalStorage`; the connection is verified on save (with the option to save anyway).
- The existing preferences keep working unchanged as the first server. They are now optional instead of required, so nothing changes for current users and no migration is needed.
- `Manage VMs` and `Manage Storage` fetch all configured servers and group the results into one section per server (single-server setups look exactly as before). An unreachable server shows a per-server error row instead of hiding the other servers' results.
- VM actions and `Open Dashboard` use the server of the row they are triggered on.

Two small fixes that came out of this:

- Fetch errors now surface `error.cause` (e.g. TLS or DNS problems) instead of Node's generic "fetch failed".
- `pveFetch` now checks `response.ok` — before, a failed VM action (e.g. missing token permission) showed a success toast.

Note for review: token secrets of additional servers live in `LocalStorage` (the preference-based server keeps using the encrypted `password` preference). Happy to change this if there is a preferred pattern for multi-instance credentials.

Tested against three Proxmox VE 9 servers (one cluster + two standalone nodes) on macOS.

## Screencast

Mock data, two servers configured:

| Manage VMs | Manage Storage |
|---|---|
| Manage VMs with two servers | Manage Storage with two servers |

## Checklist

### DIFF
--- a/extensions/proxmox/src/hooks/use-multi-pve-fetch.ts
+++ b/extensions/proxmox/src/hooks/use-multi-pve-fetch.ts
@@ -1,4 +1,4 @@
-import { useEffect } from "react";
+import { useEffect, useRef } from "react";
 import { usePromise } from "@raycast/utils";
 import type { PveServer, PveServerResult } from "@/types";
 import { pveFetch } from "@/api";
@@ -19,26 +19,29 @@ type MultiPveFetchOptions = {
 export const useMultiPveFetch = <T>(url: string, options?: MultiPveFetchOptions) => {
   const { timerInterval = 1000, execute = true } = options ?? {};
   const { servers, isLoading: isLoadingServers } = useServers();
-
-  // Serialized so the promise re-executes exactly when the server list changes
-  const serversKey = JSON.stringify(servers);
+  const abortable = useRef<AbortController>(null);
 
   const result = usePromise(
-    async (key: string): Promise<PveServerResult<T>[]> => {
-      const keyServers = JSON.parse(key) as PveServer[];
-      return Promise.all(
-        keyServers.map(async (server) => {
+    async (servers: PveServer[]): Promise<PveServerResult<T>[]> =>
+      Promise.all(
+        servers.map(async (server) => {
           try {
-            const response = await pveFetch<T>(server, url);
+            const response = await pveFetch<T>(server, url, { signal: abortable.current?.signal });
             return { server, data: response.data };
           } catch (error) {
+            // Let an aborted request reject the whole call, usePromise ignores it.
+            // Keeping it would show a stale request as a connection error.
+            if (error instanceof Error && error.name === "AbortError") {
+              throw error;
+            }
+
             return { server, error: describeFetchError(error) };
           }
         }),
-      );
-    },
-    [serversKey],
+      ),
+    [servers],
     {
+      abortable,
       execute: execute && !isLoadingServers,
     },
   );
--- a/extensions/proxmox/src/hooks/use-multi-pve-fetch.ts
+++ b/extensions/proxmox/src/hooks/use-multi-pve-fetch.ts
@@ -1,0 +1,65 @@
+import { useEffect } from "react";
+import { usePromise } from "@raycast/utils";
+import type { PveServer, PveServerResult } from "@/types";
+import { pveFetch } from "@/api";
+import { describeFetchError } from "@/utils/errors";
+import { useServers } from "@/utils/servers";
+
+type MultiPveFetchOptions = {
+  timerInterval?: number | null;
+  execute?: boolean;
+};
+
+/**
+ * Fetch the same resource from every configured server.
+ *
+ * Failures are captured per server, so one unreachable server
+ * doesn't prevent showing the results of the others.
+ */
+export const useMultiPveFetch = <T>(url: string, options?: MultiPveFetchOptions) => {
+  const { timerInterval = 1000, execute = true } = options ?? {};
+  const { servers, isLoading: isLoadingServers } = useServers();
+
+  // Serialized so the promise re-executes exactly when the server list changes
+  const serversKey = JSON.stringify(servers);
+
+  const result = usePromise(
+    async (key: string): Promise<PveServerResult<T>[]> => {
+      const keyServers = JSON.parse(key) as PveServer[];
+      return Promise.all(
+        keyServers.map(async (server) => {
+          try {
+            const response = await pveFetch<T>(server, url);
+            return { server, data: response.data };
+          } catch (error) {
+            return { server, error: describeFetchError(error) };
+          }
+        }),
+      );
+    },
+    [serversKey],
+    {
+      execute: execute && !isLoadingServers,
+    },
+  );
+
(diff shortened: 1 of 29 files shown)


extensions/proxmox/src/hooks/use-multi-pve-fetch.ts
```diff
@@ -0,0 +1,65 @@
...
+ *
+ * Failures are captured per server, so one unreachable server
+ * doesn't prevent showing the results of the others.
+ */
+export const useMultiPveFetch = <T>(url: string, options?: MultiPveFetchOptions) => {
+  const { timerInterval = 1000, execute = true } = options ?? {};
+  const { servers, isLoading: isLoadingServers } = useServers();
+
+  // Serialized so the promise re-executes exactly when the server list changes
+  const serversKey = JSON.stringify(servers);
```
CzBiX: usePromise already deep-compares its args array, no re-serialize needed

extensions/proxmox/src/hooks/use-multi-pve-fetch.ts
```diff
@@ -0,0 +1,65 @@
...
+
+  const result = usePromise(
+    async (key: string): Promise<PveServerResult<T>[]> => {
```
CzBiX: should passthrough abortable signal

CzBiX (changes requested):

extensions/proxmox/src/hooks/use-multi-pve-fetch.ts
kzaoaai: you're right. The servers array is now passed straight through.

extensions/proxmox/src/hooks/use-multi-pve-fetch.ts
kzaoaai: abortable is now passed to usePromise and its signal forwarded to every request. One detail worth flagging: the per-server catch rethrows AbortError instead of turning it into a result, otherwise a superseded revalidation renders as a "Connection Failed" row for that server.

kzaoaai:
Thanks for the review, both points are addressed:

- **Args deep-compare** — you're right, `usePromise` already compares its args with `dequal`, so the serialized key was doing work the hook does itself. The servers array is now passed straight through.
- **Abort signal** — `abortable` is now passed to `usePromise` and its signal forwarded to every request. One detail worth flagging: the per-server `catch` rethrows `AbortError` instead of turning it into a result, otherwise a superseded revalidation renders as a "Connection Failed" row for that server.

I also changed how servers are stored while I was in there: each server now lives under its own `LocalStorage` key instead of all of them in one array, so add/edit/remove no longer rewrite each other's entries (this was the lost-update path the review bot flagged). Existing configurations are migrated on the first read, keeping their original order.

Tested against three PVE 9 servers (one cluster plus two standalone nodes): list, per-server sections, storage contents, and the VM actions.

0xdhrv (commented):
Hey @CzBiX 👋 

Thanks for reviewing, can you please take a look at this again when you get a chance?

extensions/proxmox/src/hooks/use-multi-pve-fetch.ts
```diff
@@ -0,0 +1,68 @@
...
+            // Keeping it would show a stale request as a connection error.
+            if (error instanceof Error && error.name === "AbortError") {
+              throw error;
+            }
+
+            return { server, error: describeFetchError(error) };
+          }
+        }),
+      ),
+    [servers],
```
CzBiX: url not in `usePromise` args, its a trap in the future

extensions/proxmox/src/hooks/use-multi-pve-fetch.ts
```diff
@@ -0,0 +1,68 @@
...
+  execute?: boolean;
+};
+
+/**
+ * Fetch the same resource from every configured server.
+ *
+ * Failures are captured per server, so one unreachable server
+ * doesn't prevent showing the results of the others.
+ */
+export const useMultiPveFetch = <T>(url: string, options?: MultiPveFetchOptions) => {
```
CzBiX, line 19: the servers never loaded for me. @kzaoaai can you confirm it works?

CzBiX (changes requested):

extensions/proxmox/src/hooks/use-multi-pve-fetch.ts
kzaoaai: Fixed in 01bc3b4 — url is passed through the usePromise args now instead of being captured in the closure.

extensions/proxmox/src/hooks/use-multi-pve-fetch.ts
kzaoaai: Reproduced, and it was a real bug, thanks for pushing on it.

There was a leftover single-server special case: when exactly one server was configured and its request failed, the whole list was replaced by the old "Something went wrong, check your preferences" screen. So the actual error was hidden, and it pointed at the preferences even for a server added through Manage Servers. With two or more servers the per-server error row rendered correctly, which is why I did not hit it on my own setup.

Fixed in 7c4c236: the failing server is always rendered as a list item with the underlying error (TLS, DNS, HTTP status), whatever the number of configured servers.

Checked on a fresh install with a single unreachable server (self-signed cert): before the fix the generic error screen, after it a "Connection Failed" row showing unable to verify the first certificate (UNABLE_TO_VERIFY_LEAF_SIGNATURE). The healthy path with three servers still lists normally.

If your server was reachable and you still saw nothing, let me know and I will keep digging.

kzaoaai:
@0xdhrv the issue @CzBiX reported is fixed — it was a leftover single-server case that replaced the list with the generic error screen instead of showing which server failed and why (7c4c236), plus the url arg change he asked for (01bc3b4). Both review comments are answered. Should this go back out of draft for another look?

0xdhrv (commented):
Hi @kzaoaai 👋

@CzBiX do you want to check this?

CzBiX (approved):
@kzaoaai could you bump `@raycast/utils` as well to resolve raycast/utils/issues/64. I'm OK with other changes.

kzaoaai:
Pushed 327df8e: a failed migration write no longer deletes the legacy server list — the broad catch was swallowing write failures as well as unreadable values, so any server whose write didn't land was dropped permanently. Verified with a seeded legacy value and a forced write failure: the servers come back on the next read. Taking this out of draft.

0xdhrv (approved):
Looks good to me, approved ✅

PR #30127, opened 2026-08-10
Update Digger extension

chrismessina:
## Description

Digger is a website-intelligence extension. This update fixes a set of failure-path bugs — most importantly, **a dig that failed outright used to leave the loading spinner running forever with no error message** — brings all keyboard shortcuts into line with Raycast's conventions on both macOS and Windows, and refreshes dependencies.

The extension declares `platforms: ["macOS", "Windows"]`, but every custom shortcut used a bare ⌘ binding that does not exist on Windows, leaving those actions unreachable there. That is fixed throughout.

### Changes

**Failure handling**

- A failed dig (unreachable host, mistyped domain) now reports the error immediately instead of hanging on a spinner. The hook was aborting its own controller to cancel pending work, then mistaking that for user cancellation and skipping both the error state and the loading reset.
- Errors are classified from the full `cause` chain. Node's `fetch` reports transport failures as an opaque `TypeError: fetch failed` with the real reason (`getaddrinfo ENOTFOUND …`) one level down, so failures previously fell through to a generic message. A DNS failure now reads "Connection Failed" with connection-specific guidance.
- The failure screen is a centred `List.EmptyView` rather than a list row with a detail pane.
- Digging a new URL before the previous one finishes no longer lets the abandoned request overwrite the new results, blank its panels, or fill its progress bars.
- DNS & Certificates no longer displays the previous site's certificate when the current site's TLS lookup fails.
- "Open in Wayback Machine" / "Save to Wayback Machine" now resolve the correct app under Raycast beta (`RAYCAST_SCHEME`).

**Keyboard shortcuts**

- All custom shortcuts are now `Keyboard.Shortcut.Common` members or explicit `{ macOS, Windows }` pairs.
- Fixed a collision in the HTTP Headers panel where "Copy Header Name" and "Copy Header Value" were both bound to ⌘⇧C, making one unreachable. "Copy Header Name" now uses `Common.CopyName`.
- "Save to Wayback Machine" moved ⌘⇧S → ⌘⇧Y and "Copy Canonical URL" ⌘⌥C → ⌘⌥U, so they no longer shadow `Common.Duplicate` / `Common.CopyName`.

**Error reporting**

- Copy-Error actions on every user-visible failure surface (the toast, the error screen, the partial-failure banner), carrying the URL, the underlying cause, and suggestions.
- URLs written to the debug log are stripped of their query strings.

**Other**

- Screenshots reduced from 11 to 6 to meet the Store maximum. All six are unchanged files, renamed.
- README keyboard table corrected — it documented ⌘C for an action bound to ⌘⇧C — and given a Windows column.
- Removed a command subtitle that duplicated the command title.
- Dependencies: `@raycast/api` 1.104.24, `@raycast/utils` 2.2.7, plus dev dependencies. Two high-severity advisories cleared via the lockfile.

## Checklist

- [X] I ran `npm run build` and tested this distribution build in Raycast
- [x] I ran `npm run lint` and `npx tsc --noEmit` — both exit 0


0xdhrv (approved):
Looks good to me, approved ✅

PR #30140, opened 2026-08-11
Add Layout Hotkeys extension

pavel-kh:
## Description

Switch macOS keyboard layouts directly with a hotkey instead of cycling through them, and rewrite text that was typed with the wrong layout active.

macOS only offers "select next input source", so reaching a specific layout out of four means pressing the switch key three times. This gives four commands you can bind independent global hotkeys to, each landing directly on the layout assigned to it.

## Commands

- **Switch to Layout 1–4** — switch straight to the input source assigned to that slot
- **Convert Selection to Layout 1–4** — rewrite the selected text as if it had been typed with that slot's layout
- **Convert Selection** — preview the selection rewritten into each layout and pick one
- **Configure Layout Slots** — choose which input source each slot selects
- **Cycle Input Source**, **Show Current Input Source**, **Switch Input Source**

## Notes for review

**No hardcoded layout tables.** Text conversion derives its mapping at runtime from each layout's own key tables via `UCKeyTranslate` and `kTISPropertyUnicodeKeyLayoutData`. For each character it resolves which physical key and modifiers produced it under the detected layout, then asks the target layout what that key produces. Any pair of enabled layouts therefore works, including custom `.keylayout` bundles built with Ukelele, with nothing to maintain per language. Round-trips (A → B → A) return the original exactly, including Option-key diacritics such as Polish `ą ć ż`.

**Permissions.** The switching commands need none — no Accessibility, no Input Monitoring, no synthesized keystrokes. The Convert Selection commands read the frontmost app's selection, which requires Raycast's own Accessibility permission.

**Whole-field fallback.** With nothing selected, the convert commands read the focused field through the Accessibility API rather than synthesizing ⌘A, since a synthetic keystroke would land on Raycast while it briefly holds focus. Because "the focused field" can be an entire document, anything over 200 characters asks for confirmation first, and a `When Nothing Is Selected` preference disables the fallback entirely.

**Focus handling.** A Raycast command runs with Raycast briefly frontmost. With macOS's "Automatically switch to a document's input source" enabled, switching then would be undone once focus returns, so the Swift side waits up to 400 ms for Raycast to stop being frontmost before calling `TISSelectInputSource`. The hotkey commands deliberately do not dismiss the Raycast window, so they also work while typing in Raycast's own search field.

Swift is bundled as source and compiled by `extensions-swift-tools`; no prebuilt binary is committed.


pernielsentikaer:

- **Keyboard Layout Switcher** (`input-switcher`, 4,530 installs by @lucaschultz) — Extension to switch keyboard layouts
  <sub>similarity 0.66</sub>

pavel-kh:
Thanks for the pointer — I've updated the README to name Keyboard Layout Switcher and spell out the difference, in 52e38fc.

For the record on the overlap itself: Keyboard Layout Switcher has two commands, a picker listing your layouts and "Toggle Next Layout". Neither of the things this extension is for is among them.

- **A hotkey that lands on one specific layout.** Its only command that switches without further interaction is the toggle, which advances to the next layout in sequence; reaching a particular one means opening the picker and choosing from the list. Here, *Switch to Layout 1–4* are four separate commands, so one hotkey goes straight to Hebrew and another straight to Russian.
- **Repairing text typed with the wrong layout active.** The *Convert Selection* commands turn `ghbdtn` into `привет` by mapping each character through the physical key that produced it, derived from the layouts themselves rather than per-language tables. Keyboard Layout Switcher has no equivalent.

There's also no external dependency: it drives the separately installed `keyboardSwitcher` binary, while this calls Carbon's `TISSelectInputSource` from bundled Swift.

The genuine overlap is *Switch Input Source* and *Cycle Input Source*, which do roughly what its two commands do. Those are here so the four hotkeys aren't the only way in, and I'm happy to drop either if you'd rather keep the Store tidy.

pernielsentikaer:
Can't we enhance the current one by @lucaschultz

0xdhrv (commented):

- [Keyboard Layout Switcher](https://www.raycast.com/lucaschultz/input-switcher) by @lucaschultz

Both switch the macOS input source from Raycast. **Switch to Layout 1–4** is the same job **Switch Layouts** / **Toggle Next Layout** already cover, with bound slots instead of a picker or cycle.

- **Switch Layouts** to browse enabled layouts and switch
- **Toggle Next Layout** to cycle

- Four bindable slots that land on a specific layout
- **Convert Selection** via runtime `UCKeyTranslate` (any enabled pair, not just EN↔Cyrillic)
- Show current source and a configure-slots command

Could we consider contributing the slot hotkeys and convert flow to Keyboard Layout Switcher instead?

> This would help avoid duplication and keep keyboard-layout switching consolidated in one extension.  

PR #30141, opened 2026-08-11
Add TOTP extension

imbriandev:
## Summary

- Add a local-first TOTP authenticator with saved and quick OTP commands.
- Support encrypted AES-256-GCM backup export and import.

## Test plan

- `npm run lint`
- `npm run build`
- `npm run test`


pernielsentikaer:

- **One Time Password** (`one-time-password`, 2,068 installs by @lacherogwu) — Generate time-based one-time passwords
  <sub>similarity 0.78</sub>
- **QOTP** (`qotp`, 62 installs by @ariesly15) — QOTP is a simple tool to generate OTP code
  <sub>similarity 0.71</sub>
- **Proton Authenticator** (`proton-authenticator`, 825 installs by @FareedFarag) — Quick access to TOTP entries from Proton Authenticator
  <sub>similarity 0.68</sub>
- **2FAS Authenticator** (`2fas-authenticator`, 15 installs by @LockeAG) — Search and copy TOTP codes from 2FAS exports with a Keychain-encrypted local vault
  <sub>similarity 0.64</sub>
- **Two-Factor Authentication Code Generator** (`two-factor-authentication-code-generator`, 5,765 installs by @cjdenio) — Quickly access your two-factor authentication codes.
  <sub>similarity 0.62</sub>

0xdhrv (commented):

- [Two-Factor Authentication Code Generator](https://www.raycast.com/cjdenio/two-factor-authentication-code-generator) by @cjdenio

Both keep TOTP secrets locally, show the current code, and copy or paste it. **TOTP** is the same saved-authenticator workflow that **Generate 2FA Code** already provides.

- Live codes with a period countdown
- Add an account by name + secret or `otpauth://` URL (digits, period, algorithm)
- Copy or paste the code into the focused app
- **Backup 2FA Codes** and **Restore 2FA Codes**

- **Quick OTP** for a one-off secret without saving it
- AES-256-GCM encrypted export/import, plus optional automatic backups
- Conceal copied codes from Raycast clipboard history

Could we consider contributing encrypted backups and Quick OTP to the existing extension instead?

> This would help avoid duplication and keep local TOTP authenticators consolidated in one extension.  

imbriandev:
@0xdhrv  Thanks for the thoughtful feedback. We agree that the primary workflow overlaps with the existing TOTP extension, and we are open to contributing the strongest parts of this work to the existing extension rather than maintaining a separate extension with the same core purpose.

The main improvements in our implementation are:

### Quick OTP

- Generates an OTP directly from a pasted Base32 secret or `otpauth://` URI.
- Does not save the secret unless the user explicitly chooses **Add Account**.
- Shows the current OTP and expiry countdown while typing.
- Supports the URI parameters we already support for saved accounts, including issuer, digits, period, and SHA-1/SHA-256/SHA-512.
- Supports adding the current Quick OTP input to the saved account list without re-entering the secret.
- Keeps secrets concealed when copied.

This is useful for one-off login flows, migrating accounts, or using a secret temporarily without increasing the saved-account vault.

### Encrypted backups

- AES-256-GCM encrypted export/import rather than plaintext JSON.
- Passphrase-based key derivation with a random salt and authenticated encryption.
- Validates the authentication tag and backup contents before importing.
- Merges backups without creating duplicate accounts, including duplicates within the same backup file.
- Uses atomic replacement when writing a backup file, so a partial file is not left behind.
- Portable backup format that is independent of a particular provider or authenticator app.

### Automatic backups

- Users can configure an automatic backup directory and passphrase in extension preferences.
- After every account add, edit, remove, or backup import, the extension writes an encrypted `totp-backup.json` to that directory.
- The automatic backup is updated atomically and is never uploaded by the extension.
- If automatic backup is enabled but misconfigured or fails, the account mutation is rejected instead of silently leaving the backup stale.

### Account management and compatibility improvements

- Accepts standard Base32 secrets as well as `otpauth://` URIs.
- Explicit issuer support for manually entered secrets, so accounts can be grouped and searched by provider.
- Edit accounts while preserving their stable account ID.
- Search by account name, email, or issuer.
- Copy OTP, paste OTP, and copy concealed secret actions.
- No provider-specific integration, network dependency, or companion app requirement.

We would be happy to contribute these features incrementally to the existing extension, likely starting with Quick OTP and the encrypted backup format, then adding automatic backups if that fits the project’s security and UX guidelines. We can also adapt the implementation to the existing extension’s storage model and conventions rather than asking users to migrate to a separate extension.

0xdhrv:
> @0xdhrv Thanks for the thoughtful feedback. We agree that the primary workflow overlaps with the existing TOTP extension, and we are open to contributing the strongest parts of this work to the existing extension rather than maintaining a separate extension with the same core purpose.
> 
> The main improvements in our implementation are:
> 
> ### Quick OTP
> * Generates an OTP directly from a pasted Base32 secret or `otpauth://` URI.
> * Does not save the secret unless the user explicitly chooses **Add Account**.
> * Shows the current OTP and expiry countdown while typing.
> * Supports the URI parameters we already support for saved accounts, including issuer, digits, period, and SHA-1/SHA-256/SHA-512.
> * Supports adding the current Quick OTP input to the saved account list without re-entering the secret.
> * Keeps secrets concealed when copied.
> 
> This is useful for one-off login flows, migrating accounts, or using a secret temporarily without increasing the saved-account vault.
> 
> ### Encrypted backups
> * AES-256-GCM encrypted export/import rather than plaintext JSON.
> * Passphrase-based key derivation with a random salt and authenticated encryption.
> * Validates the authentication tag and backup contents before importing.
> * Merges backups without creating duplicate accounts, including duplicates within the same backup file.
> * Uses atomic replacement when writing a backup file, so a partial file is not left behind.
> * Portable backup format that is independent of a particular provider or authenticator app.
> 
> ### Automatic backups
> * Users can configure an automatic backup directory and passphrase in extension preferences.
> * After every account add, edit, remove, or backup import, the extension writes an encrypted `totp-backup.json` to that directory.
> * The automatic backup is updated atomically and is never uploaded by the extension.
> * If automatic backup is enabled but misconfigured or fails, the account mutation is rejected instead of silently leaving the backup stale.
> 
> ### Account management and compatibility improvements
> * Accepts standard Base32 secrets as well as `otpauth://` URIs.
> * Explicit issuer support for manually entered secrets, so accounts can be grouped and searched by provider.
> * Edit accounts while preserving their stable account ID.
> * Search by account name, email, or issuer.
> * Copy OTP, paste OTP, and copy concealed secret actions.
> * No provider-specific integration, network dependency, or companion app requirement.
> 
> We would be happy to contribute these features incrementally to the existing extension, likely starting with Quick OTP and the encrypted backup format, then adding automatic backups if that fits the project’s security and UX guidelines. We can also adapt the implementation to the existing extension’s storage model and conventions rather than asking users to migrate to a separate extension.

Thanks for the detailed breakdown and for being open to consolidating the work. I really appreciate the effort that has gone into these improvements.

The **Quick OTP**, encrypted backup support, and the account management improvements all sound like valuable additions. Since we agree that the primary workflow overlaps with the existing [[Two-Factor Authentication Code Generator](https://www.raycast.com/cjdenio/two-factor-authentication-code-generator), I think contributing these improvements there would be the best path forward.

Please feel free to open a PR updating the existing extension, adapting the implementation to its current storage model and conventions as you suggested. Starting incrementally with Quick OTP and encrypted backups sounds reasonable, and the remaining improvements can follow where they fit.

This way we can keep the TOTP experience consolidated in a single extension while still bringing the useful work from this PR to existing users.

Thanks again for the contribution and for working with us on this!

PR #30151, opened 2026-08-12
Update karakeep extension

chrismessina:
## Description

Adds an **Add to List** action for bookmarks, replaces the list-icon field with a searchable emoji picker, and adds an optional **Update Karakeep** command for self-hosted Docker instances. Along the way it fixes an HTTP 400 that made list creation fail outright, removes a request-per-list fan-out on every command with a list dropdown, and makes API errors say what actually went wrong instead of showing a bare status code.

### Changes

- **Add to List** — a submenu on the Bookmarks rows and Bookmark Detail (⌘⇧L) that filters by typing. Lists are fetched once per view and passed down, so it does not add a request per visible bookmark.
- **List icon picker** — the field previously told users to type `:` for an emoji picker that has no Raycast equivalent, and the API rejects an empty icon, so leaving it blank failed with `HTTP 400`. The field takes any single emoji again and ⌘I opens a searchable grid of the full emoji set. Keycap emoji (1️⃣, #️⃣, *️⃣) are now accepted — the old validator stripped U+20E3 and then tested them as plain `1`, `#`, `*`.

image

image

- **Update Karakeep** (new command, macOS only) — runs `docker compose up --pull always -d` against a local instance with live progress. A failed update names the cause (registry unreachable, not authorized, out of disk, port in use) and reports whether the instance is still running. ⌘Y opens the release notes for the running version.

image

- **Fixed a request-per-list fan-out** — a bookmark count was fetched for every list on every command with a list dropdown (45 requests on a 45-list library) and the number was wrong anyway: it was capped at the page size, so any list with more than 10 bookmarks showed "(10)". The count is gone; the API exposes no count field and no stats endpoint.
- **Duplicate list names are now distinguishable** — Karakeep allows them, so same-named lists show their parent path in the Lists view, the Add to List menu, and the parent-list pickers.
- **API errors report the real message** — Karakeep encodes validation failures as a JSON string nested inside `error.message`, and the tRPC endpoints (Search, Summarize) answer with a top-level array, so both were surfacing only `HTTP 400` / `HTTP 500`.
- **Preview image failures are logged** with the underlying cause rather than failing silently into a placeholder.
- **Settings labels rewritten** for the current preferences UI, where the large row text comes from `label` and `title` is the group header. Preference keys are unchanged.
- Emoji data is generated by `tools/build-emoji-data.mjs` and committed (~100 KB) rather than adding `@emoji-mart/data`, which unpacks to 28 MB.

## Checklist


0xdhrv (approved):
Looks good to me, approved ✅

PR #30165, opened 2026-08-12
Add layout-fixer extension

ramysami:
## Description

Adds **Layout Fixer**, a single no-view command that fixes text typed with the wrong keyboard layout.

You meant to type `مرحبا`, but the keyboard was still on English, so the screen says `lnpfh`. Run the command and it becomes `مرحبا`. It works the other way too — `اثممخ` typed on a stuck Arabic layout becomes `hello`.

**Behaviour**

| Situation | What happens |
| --- | --- |
| Text is selected | The selection is converted in place |
| Nothing selected, cursor in a text field | The whole field is selected and converted |
| Neither | A HUD says there's nothing to fix; nothing is touched |

Direction is detected automatically by counting script, so there's no "to Arabic" / "to English" choice to make, and running the command twice returns the original text.

## Screencast

ezgif-3ba6ebce7ff168f5

## Checklist


pernielsentikaer:

- **Fix Language** (`fix-language`, 610 installs by @hiiamtin) — Switch the language input source of the text when you type and forget to change the language.
  <sub>similarity 0.57</sub>

ramysami:
> ### Similar extensions already in the Store
> Automated first pass. Overlap is not a blocker on its own, but the README should make the difference clear.
> 
> * **Fix Language** (`fix-language`, 610 installs by @hiiamtin) — Switch the language input source of the text when you type and forget to change the language.
>   similarity 0.57

Appreciating your reply.
`fix-language` extension is only for Thai and also does not work if there are no selection to select the whole text input area. 

Also could you make it more clear to me what should I change in the README. 

Best regards.

pernielsentikaer:
What about enhancing the other extension instead, so we don't end up with multiple extensions serving the same purpose?

ramysami:
Sounds good, I will. Thank you for your time.

PR #30166, opened 2026-08-12
Update raycast-language-tool extension

pasqualesalza:
## Description

Premium accounts cannot be used at all: filling in *Username/Email* and *API Key* makes every check fail.

The credentials are sent to `https://api.languagetool.org/v2/check`, the free endpoint, which rejects any authenticated request:

```
$ curl -d "language=en-US" --data-urlencode "text=This is a test." \
       -d "username=…" -d "apiKey=…" https://api.languagetool.org/v2/check
HTTP 400
Error: Credentials provided, but server isn't configured to support this.
```

The same request against `https://api.languagetoolplus.com/v2/check` is accepted and authenticated instead — and that is the host the bundled `src/languagetool-swagger.json` already documents (`"host": "api.languagetoolplus.com"`).

So `checkTextWithAPI` now picks the endpoint to match the request: only requests carrying credentials go to the Premium host. Without credentials nothing changes — same host, same behaviour, no regression risk for free users.

Verified with a real Premium account: before, every check failed with the 400 above; after, checks return results normally. Also checked that the free path still works with the credential fields left empty.

One note on the diff: `raycast-env.d.ts` shows up as deleted. It is generated and listed in this extension's `.gitignore`, so the publish CLI did not carry it over. Happy to restore it if you would rather keep it tracked.

## Screencast

N/A

## Checklist


pernielsentikaer (approved):
Looks good — Premium credentials now hit the Premium host. Thanks!

lucastaonline:
Sorry for the delay, but it looks good for me too!

PR #30170, opened 2026-08-12
Fix Doorstopper menu bar status toast

### LINKED ISSUE #30133: [Doorstopper] "Failed to launch command statusmenu" toast on every enable/disable when the menu-bar command is disabled
### Extension

https://www.raycast.com/roele/doorstopper

### Description

Every time I run **Enable Doorstopper** or **Disable Doorstopper**, a red toast appears: `Failed to launch command statusmenu`.

The core action succeeds. `pmset` is set correctly (`pmset -g` shows `SleepDisabled 1` after enable). The error comes only from the menu-bar refresh step.

Looking at `src/util.ts`, the enable/disable flow always calls `launchCommand({ name: "statusmenu", type: LaunchType.Background, ... })` to refresh the menu-bar item, regardless of whether the user has enabled that optional command. Raycast's `launchCommand` throws when the target command is not enabled ("If the command does not exist, or if it's not enabled, an error will be thrown"). The `catch` block then surfaces it via `showFailureToast`, turning an expected condition (an optional menu-bar command left off, which is the default) into a scary error.

The `statusmenu` command is optional and disabled by default, so this toast fires for every user who has not explicitly turned the menu bar on.

### Steps To Reproduce

1. Leave the **Doorstopper Status Menu Bar** (menu-bar) command disabled — this is the default.
2. Run **Enable Doorstopper** (or Disable, or Toggle).
3. `SleepDisabled` flips to `1` correctly, but a `Failed to launch command statusmenu` toast appears.

### Current Behaviour

A red `Failed to launch command statusmenu` failure toast is shown on every enable/disable, even though the sleep-prevention action itself succeeded, purely because the optional menu-bar command is disabled.

### Expected Behaviour

Refreshing the optional menu-bar command should fail silently. The failure toast should be reserved for real errors (for example the privileged `pmset` call failing), not for the menu-bar command simply being disabled. Since Raycast does not expose a public API to check whether another command is enabled before launching, the fix is on the catch side: swallow the "not enabled" `launchCommand` failure for `statusmenu` rather than surfacing it.

### Raycast version

Version: 1.104.24

cyphercodes:
## Description

Fixes #30133

Suppresses the optional `statusmenu` background refresh failure after Doorstopper is toggled. The main status command still reports launch failures, but a disabled menu bar command no longer shows a misleading failure toast after the toggle succeeds.

## Screencast

N/A — small background-command error-handling fix.

## Checklist

### DIFF
--- a/extensions/doorstopper/CHANGELOG.md
+++ b/extensions/doorstopper/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Doorstopper Changelog
 
+## [Fix Menu Bar Toast] - 2026-08-12
+
+- Fixed a misleading failure toast after toggling Doorstopper when the optional menu bar command is disabled.
+
 ## [Touch ID Authentication] - 2026-07-23
 
 - Added Touch ID authentication for changing Doorstopper settings when Touch ID is enabled for `sudo`
--- a/extensions/doorstopper/CHANGELOG.md
+++ b/extensions/doorstopper/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Doorstopper Changelog
 
-## [Fix Menu Bar Toast] - {PR_MERGE_DATE}
+## [Fix Menu Bar Toast] - 2026-08-12
 
 - Fixed a misleading failure toast after toggling Doorstopper when the optional menu bar command is disabled.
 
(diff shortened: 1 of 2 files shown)


0xdhrv (approved):
Looks good to me, approved ✅

PR #30184, opened 2026-08-13
[Herdr] Expand tilde in binary path

### LINKED ISSUE #29992: [Herdr] Herdr Binary preference does not expand ~
### Extension

https://raycast.com/vlades/herdr

### Raycast Version

1.104.24

### OS Version

macOS 26.4.1

### Description

Herdr version: 0.8.0

I put `~/.local/share/mise/installs/herdr/latest/herdr` into the Herdr Binary preference and got:

```
The configured Herdr binary is not executable
```

The same path works with `/Users/me/...` spelled out. In `lib/herdr.ts`, `resolveHerdrBinary()` passes the preference straight to `access()`, which doesn't expand `~` since that's a shell thing. The placeholder in the preference is `~/.local/bin/herdr`, so the field suggests a form it then rejects.

### Steps To Reproduce

1. Set Herdr Binary to a path starting with `~`, e.g. `~/.local/bin/herdr`
2. Run any Herdr command
3. See "The configured Herdr binary is not executable"

### Current Behaviour

A `~` path is treated as a literal directory name and fails.

### Expected Behaviour

`~` should expand to the home directory, or the placeholder shouldn't show a form that doesn't work.

szupzj18:
## Description

The Herdr Binary preference advertises paths such as `~/.local/bin/herdr`, but Node filesystem APIs treat `~` literally.

This expands a leading `~/` before checking executability and adds a regression test.

## Validation

- `npm run check`
- 39 tests passed

Fixes #29992

### DIFF
--- a/extensions/herdr/CHANGELOG.md
+++ b/extensions/herdr/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## [Fix Tilde Paths] - {PR_MERGE_DATE}
+## [Fix Tilde Paths] - 2026-08-13
 
 - Expand `~` in the configured Herdr binary path.
 
--- a/extensions/herdr/CHANGELOG.md
+++ b/extensions/herdr/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## [Fix Tilde Paths] - {PR_MERGE_DATE}
+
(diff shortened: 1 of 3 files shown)


pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #30188, opened 2026-08-13
[Agent Usage] Fix Antigravity detection on 2.7+ (suffix-less language_server binary)

ljgoldenboy:
## Description

Antigravity 2.7 renamed its language server binary from `language_server_macos` to plain `language_server`. The process filter in `src/antigravity/probe.ts` only matched the suffixed names, so the running process was dropped at the very first gate and the Antigravity row stayed stuck on **"Not Running"** while the app was up. (The standalone "Antigravity Quota" extension keeps working because it filters on the app marker rather than on the binary name.)

This PR matches the common `language_server` prefix in both the macOS detection and the Windows pre-filter. Selection is otherwise unchanged: the Antigravity-specific gate (`--app_data_dir antigravity` / `antigravity` in the command line) still rejects Windsurf and Devin language servers, which ship a binary with the same prefix — I have a Devin language server running on the same machine (`language_server_macos_arm`) and it is still rejected.

Verified on macOS 26.5.1 with Antigravity 2.8.0: the probe now returns both quota groups with percentages and reset times, and the row renders the quota it already knew how to render.

Added unit tests cover three cases: the suffix-less binary, the legacy `_macos` / `_windows` names, and the rejection of a foreign (Windsurf/Devin) language server.

`npm test` (211 tests), `npm run typecheck`, `npm run lint` and `npm run build` all pass.

## Screencast

Detection-only fix with no UI change — the row simply stops reporting "Not Running" and shows the quota groups instead.

## Checklist

### DIFF
--- a/extensions/agent-usage/CHANGELOG.md
+++ b/extensions/agent-usage/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Agent Usage Changelog
 
-## [Fix Antigravity detection on 2.7+] - {PR_MERGE_DATE}
+## [Fix Antigravity detection on 2.7+] - 2026-08-13
 
 ### Bug Fixes
 
--- a/extensions/agent-usage/CHANGELOG.md
+++ b/extensions/agent-usage/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Agent Usage Changelog
 
+## [Fix Antigravity detection on 2.7+] - {PR_MERGE_DATE}
+
+### Bug Fixes
+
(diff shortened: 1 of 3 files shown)


pernielsentikaer (approved):
Hi @ljgoldenboy 👋

This is a focused fix: Antigravity 2.7+ ships a suffix-less `language_server`, and matching the common prefix (with the existing Antigravity path/flag checks still in place) is the right change. The new tests cover the 2.7+ binary, the Windows `.exe`, and a foreign Windsurf language server staying rejected.

PR #30192, opened 2026-08-13
Update twos extension

SanKlein:
Search Things now opens results in the NewTwos desktop app when it is installed, instead of always opening the browser. Opening a thing lands on its list scrolled to that row, and the browser stays on the second action.

Adds an "Open Results In" preference (Automatic / NewTwos Desktop App / Browser, defaulting to Automatic). Detection matches the app's bundle id among installed applications and pins `Action.Open` to it, degrading to browser-primary when the app is not found.


pernielsentikaer (approved):
Looks good — desktop-app open with browser fallback. Thanks!

PR #30194, opened 2026-08-13
[Karakeep] Add optional title when creating bookmarks

### LINKED ISSUE #30168: [Karakeep] Set the bookmark title when creating a bookmark
### Extension

https://www.raycast.com/luolei/karakeep

### OS Version

_No response_

### Description

`Create Bookmark` currently takes a URL (optionally prefilled from the active browser tab) and lets Karakeep derive the title from the page itself. There's no title field on the form, so saving a bookmark under a name of my own choosing is a two-step job: create it, then go back into `Bookmarks`, find it again, and rename it through Edit.

I'd like an optional **Title** field on the Create Bookmark form:

- Left empty — behaviour is exactly as it is today, and Karakeep uses the page title.
- Filled in — that value is used as the bookmark's title.

Nothing changes for anyone who doesn't type in the field, so this shouldn't cost anything in speed for the common case.

The extension already supports editing a bookmark's title after it's saved, so the field is clearly settable through the API. This is about exposing it one step earlier — at the moment I'm creating the bookmark, when I still have the context in my head.

Anyone saving pages whose `<title>` isn't a useful name on its own:

- Documentation sites where every page carries the same product name, so a dozen bookmarks all read alike and none of them says which page it actually is.
- Single-page apps and dashboards that report a generic or empty title.
- PDFs and files opened in the browser, which often come through as a filename or as "Untitled".
- Pages in a language I don't read, where I want a note-to-self title instead.

More generally: the title is what makes a bookmark findable later. Being able to write it at save time is the difference between a searchable library and a pile of links I have to tidy up afterwards — which, in practice, doesn't happen.

### Anything else?

Renaming later is not just an extra step, it's an extra step at the wrong moment. When I save something I know exactly what I want to call it; when I come back to a list of bookmarks a week later, that's gone and I'm reconstructing it from the URL.

One implementation note: I don't think the field needs to be prefilled with the fetched page title. Karakeep resolves that server-side after the bookmark is created, so it probably isn't available while the form is open. An empty field with a placeholder along the lines of "Leave empty to use the page title" would make the behaviour obvious without waiting on anything.

I've left `Quick Bookmark` out of this request on purpose — that command exists to be a single keystroke, and a form field would defeat the point of it.

mitsimi:
## Description

Resolves #30168

Adds an optional title field to the **Create Bookmark** form.

When left empty, the title is omitted from the request and Karakeep continues to derive it from the bookmarked page. When filled in, the trimmed value is saved as the bookmarks custom title.

> [!NOTE]
> I do not speak Chinese, so the Chinese translations were created with DeepL. I would appreciate it if a native or proficient Chinese speaker could verify them.

## Checklist


pernielsentikaer (approved):
Hi @mitsimi 👋

Optional title on Create Bookmark is a clean add, and using the 201 vs 200 status so a custom title still lands on an existing URL is the right fix. Changelog uses `{PR_MERGE_DATE}`, Greptile is clear.

PR #30195, opened 2026-08-13
Fix 1Password desktop-app integration failures on Windows

yuyoyuppe:
## Description

On Windows the 1Password app briefly has no pipe listener right after an `op` process disconnects, so the extension's back-to-back `op` calls fail instantly with "cannot connect to 1Password app" even though the app is running. On top of that, `op whoami` reports "account is not signed in" for app-integration setups without an `op signin` session, which forced a bogus "Authentication Required" screen whose auto-run `op signin` hit the same race.

- Route `op` invocations through a shared exec helper; on Windows, serialize them and retry desktop-app connection failures briefly (6 x 350ms). macOS keeps the fail-fast, concurrent behavior.
- Fall back to `op account get` (delegated sessions) for the Windows sign-in check and for account resolution when `whoami` fails.
- Convert useExec-based hooks to useCachedPromise over the shared helper, preserving cached results between launches.

## Checklist


pernielsentikaer (approved):
Hi @yuyoyuppe 👋

The Windows desktop-app race is handled the right way: serialize `op` on Windows, retry the no-listener window, leave macOS concurrent. Falling back from `whoami` to `account get` for delegated sessions is the missing piece, and routing copy/share/open/generate through the same helpers means those actions get the same treatment.

PR #30206, opened 2026-08-14
2FAS Authenticator: fix import crash on missing algorithm field

LockeAG:
## Description

Some 2FAS exports omit the `otp.algorithm` field on a service entry. The type declared it as a required `string`, but `mapService()` called `normalizeAlgorithm()` on it unguarded, outside the parser's error handling. A single algorithm-less service threw a `TypeError` and aborted the entire import, instead of the service landing in the existing "Skipped" section like other malformed entries.

This makes `algorithm` optional and falls through to the SHA1 default, which matches the RFC 6238 default and real-world 2FAS export behavior.

Originally reported and fixed by @Leander250 against the standalone dev repo (LockeAG/raycast-2fas-authenticator#1); ported here for the Store release.

## Checklist

- [x] I ran `npm run build` and confirmed that the build was successful
- [x] I ran `npm run lint` and confirmed that there were no errors or warnings

### DIFF
--- a/extensions/2fas-authenticator/CHANGELOG.md
+++ b/extensions/2fas-authenticator/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## [Fix Import Crash] - {PR_MERGE_DATE}
+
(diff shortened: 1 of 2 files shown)


0xdhrv (approved):
Looks good to me, approved ✅

PR #30207, opened 2026-08-14
Add openqr extension

sam-moreton:
## Description

Adds **OpenQR**, an extension for generating and managing QR codes without leaving Raycast.

It talks to the [OpenQR](https://openqr.uk) REST API through the official [`@open-qr/sdk`](https://www.npmjs.com/package/@open-qr/sdk) client. An API key is required and is created free at [openqr.uk/api](https://openqr.uk/api); it is stored via the Raycast `preferences` API as a `password` field, and nothing else is configured.

Three commands:

| Command | What it does |
| --- | --- |
| **Generate Static QR Code** | Encodes text or a URL as a permanent code. Choose PNG or SVG, pixel size, and optional foreground/background colors. Copy the image, save it, or open it. |
| **Generate Dynamic QR Code** | Creates an editable `oqr.to` short link and returns **the QR image for it**, so the printed artefact survives a later re-point. Copy, save or open the image; the short URL is copied to the clipboard automatically. |
| **Manage Dynamic QR Codes** | Lists your dynamic codes with the QR rendered in a detail pane. Copy or save the image, re-point the destination, view scan analytics, and delete. |

Two deliberate design points, in case they read oddly in review:

- The dynamic commands encode the **short URL**, never the destination. Encoding the destination would produce an image that breaks the moment the code is re-pointed, which defeats the feature.
- QR rendering is cached on disk and only the selected row in the list renders one. Every `List.Item` mounts, so a naive per-item render would fire one API request per code in the account each time the command opens.

The extension is MIT licensed. OpenQR itself is AGPL-3.0 and open source at [open-qr/openqr](https://github.com/open-qr/openqr).

## Screencast

Three screenshots are included in `metadata/` rather than a screencast, since every command resolves to a single screen and there is no multi-step flow to show:

- `generate-static-qr.png` — a rendered code, not the empty form
- `generate-dynamic-qr.png` — the result view with the code, its short URL and destination
- `manage-dynamic-qr-codes.png` — the list plus detail pane, an active code selected

All were captured against the real extension with a live API key, so the manage screenshot shows genuine data rather than a mock.

## Checklist


pernielsentikaer:

- **QR Code Generator** (`qrcode-generator`, 31,742 installs by @Melvynx) — Generate QR codes from text or URLs.
  <sub>similarity 0.61</sub>
- **QRCP** (`qrcp`, 427 installs by @yohann84L) — Transfer files over Wi-Fi from/to your computer to/from your mobile device by scanning a QR code without leaving Raycast.
  <sub>similarity 0.58</sub>
- **QR Code Scanner** (`qr-code-scanner`, 15,312 installs by @StevenRCE0) — Simple on-screen QR code scanner. ("Screen recording" permission is required)
  <sub>similarity 0.57</sub>

sam-moreton:
Thanks — README updated in e2015f7 with an explicit comparison rather than a claim of novelty.

For context on what I checked: there are four QR extensions in the repo (`qrcode-generator`, `qrcp`, `qr-code-scanner`, `vietqr-transfer`) plus the link shorteners, and only one of the three flagged actually overlaps:

- **QR Code Generator** renders locally via the `qrcode` package, offline and account-free. That genuinely overlaps OpenQR's static command, so the README now says outright: *"If you only want a static image made offline, use QR Code Generator."* We're not trying to replace it.
- **QRCP** uses a QR as the transport for a LAN file transfer, and **QR Code Scanner** decodes rather than encodes. No overlap either way.
- **Link shorteners** (Bitly, etc.) can re-point a destination but expose no QR command and no scan analytics in Raycast.

What isn't in the Store today is the dynamic code: create an `oqr.to` code, print the image, then change where it points without reprinting, plus per-code scan analytics and a browsable list of the account's codes with the QR rendered for the selected row.

pernielsentikaer (changes requested):
Hi 👋 

- Could you add at least one [screenshot](https://developers.raycast.com/basics/prepare-an-extension-for-store#screenshots) of the extension and add them to a folder called `metadata` - you need to use our tool to get the right padding on the images

sam-moreton:
@pernielsentikaer Thanks for the review. I've added the screenshots to the `metadata` folder.

pernielsentikaer:
Hi @sam-moreton 👋

- The two Store screenshots are now Window Captures of Generate Static QR Code and Manage Dynamic QR Codes, which is what I asked for, but they are still dark. Please recapture them in light mode.
- When Manage Dynamic QR Codes fails to load, it toasts and then shows "No dynamic codes yet". Please use the error on that list and show a real failure empty state so a bad API key or network error is not mistaken for an empty account.

---

sam-moreton:
Both points addressed, thanks @pernielsentikaer:

**Light mode.** Both Store screenshots are recaptured as Window Captures with the system in light appearance, same states as before (static: a generated code for `https://openqr.uk`; manage: "Event poster" selected over "Spring menu") and the same custom wallpaper, so the set stays visually consistent. Both are 2000x1250, and both QRs were re-verified with zbarimg after the reshoot: the static one decodes to `https://openqr.uk` and the manage one to the selected code's short URL.

**Failure state.** When `listDynamicCodes` fails (bad API key, network error), the command used to toast the error and then render the "No dynamic codes yet" empty view, which read as an empty account. The list now checks the hook's `error` and, while it is set, shows a dedicated empty view carrying the actual message — including the specific invalid-API-key guidance — with a Retry action right there. The "No dynamic codes yet" view only renders once a load has succeeded and returned zero codes.

Also added the `version` field (1.1.0) and a changelog entry for the update flow. The demo codes used for the manage screenshot were deleted afterwards (both slugs confirmed 404), so the account is back to zero.

pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #30212, opened 2026-08-14
Soundboard: fix "ReferenceError: crypto is not defined" when saving a sound

eric-ostholm:
## Description

Saving a sound in the Soundboard extension currently fails with:

```
ReferenceError: crypto is not defined
    at Q (index.js:11:4)
    at Object.onSubmit (index.js:1:6486)
```

**Root cause:** `nanoid` was bumped to `^6.0.1` in the 2026-08-13 release ("Bump all dependencies to the latest"). nanoid v6 dropped its Node-specific entry point, which previously used `require('crypto').randomFillSync`, and now generates random bytes via the WebCrypto **global** `crypto.getRandomValues`. Raycast's extension runtime does not expose `crypto` as a global, so the bundled nanoid throws as soon as an ID is generated - which happens on form submit, i.e. every time a sound is added.

`src/soundform.tsx` was the only consumer of nanoid, so this replaces `nanoid()` with `randomUUID()` from `node:crypto` and drops the dependency. `src/utils.ts` already imports `randomUUID` from `node:crypto`, so this keeps ID generation consistent with the rest of the extension rather than adding a new pattern.

IDs are only used as opaque React keys and lookup handles for items in `LocalStorage`, so the change in format (nanoid's 21-char ID vs. a UUID v4) has no functional impact. Existing stored items keep their current IDs.

## Checklist

- [x] I ran `npm run lint` (`ray lint`) and it passed with no errors or warnings
- [x] I checked that the extension still builds and behaves correctly
- [x] I updated the `CHANGELOG.md` with a `{PR_MERGE_DATE}` entry

## Verification

- `ray lint` passes clean (package.json, icons, metadata, ESLint, Prettier).
- `tsc --noEmit` reports no errors in `soundform.tsx`. (Three pre-existing errors remain in `utils.ts`, unrelated to this change: the `rust:../rust` import is resolved by `ray build`'s Rust plugin rather than `tsc`, plus two implicit-`any` catch parameters.)
- Bundled `src/index.tsx` with esbuild and asserted on the output: no bare global `crypto.getRandomValues` remains, nanoid's alphabet (`useandom-…`) is gone, and `randomUUID` is reachable from `SoundForm`.

One note on scope: I could not run a full `ray build` end to end, because the Rust package added for Windows support cross-compiles to `x86_64-pc-windows-msvc` and I don't have that target installed locally. This change touches no Rust code, so CI should cover it - but flagging it rather than claiming a green full build.

### DIFF
--- a/extensions/soundboard/CHANGELOG.md
+++ b/extensions/soundboard/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Soundboard Changelog
 
+## [Fix] - {PR_MERGE_DATE}
+
+- Fixed "ReferenceError: crypto is not defined" when saving a sound by replacing `nanoid` with `randomUUID` from `node:crypto`
+
 ## [Enhancement] - 2026-08-13
 
 - Add support for Windows platform using Rust
--- a/extensions/soundboard/CHANGELOG.md
+++ b/extensions/soundboard/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Soundboard Changelog
 
-## [Fix] - {PR_MERGE_DATE}
+## [Fix] - 2026-08-14
 
 - Fixed "ReferenceError: crypto is not defined" when saving a sound by replacing `nanoid` with `randomUUID` from `node:crypto`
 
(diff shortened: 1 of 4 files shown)


0xdhrv (approved):
Looks good to me, approved ✅

PR #30225, opened 2026-08-15
Add rime-manager extension

HanryYu:
## Description

Rime Manager is a local-first macOS extension for common Rime maintenance:

- Automatically discovers the Rime user data directory and installed schemas without depending on a specific configuration distribution.
- Deploys configuration changes, syncs user data, creates backups, and opens common files and logs.
- Pins, blocks, or demotes candidates. Demotion keeps the candidate available and only moves it out of the top three.
- Searches installed applications and manages per-app input mode, punctuation, preedit, and Vim mode.
- Keeps candidate text and input codes hidden by default. Revealing them requires local macOS authentication.
- Uses English Store metadata by default and includes Chinese documentation and in-extension language switching.

All configuration writes are backed up, original schema files are never modified, and no configuration data leaves the device.

## Screenshot

Rime Manager

Three Store screenshots are included under `metadata/` and show the management overview, privacy-protected candidate rules, and the installed-app picker.

## Validation

- `npm test`: 14 tests passed
- `npm run lint`: passed
- `npm run typecheck`: passed
- `npm run build`: passed and tested in Raycast

## Native Helper Provenance

`assets/rime-manager-auth` is a universal arm64 and x86_64 macOS helper built from `native/rime-manager-auth.swift` with `scripts/build-auth-helper.sh`. It only invokes LocalAuthentication and writes a short-lived one-time grant before returning to Raycast. It never receives or stores passwords, biometric data, candidate text, or input codes.

Reproducible SHA-256: `edd5bcd864a60c4f3038f964f1c4345441c7308087103e575fd9f9656f2d2c46`

Please let me know if the Store workflow requires the Raycast team to rebuild or add the binary.

## Checklist

- [x] I checked that assets used by the README are placed outside of the `metadata` folder


extensions/rime-manager/metadata/rime-manager-1.png
0xdhrv, line 1: **issue**: metadata image style

You need to use **"Capture Window"** command to take a screenshot of the extension

Reference:

extensions/rime-manager/metadata/rime-manager-2.png
0xdhrv, line 1: **issue**: metadata image style

You need to use **"Capture Window"** command to take a screenshot of the extension

Reference:

extensions/rime-manager/metadata/rime-manager-3.png
0xdhrv, line 1: **issue**: metadata image style

You need to use **"Capture Window"** command to take a screenshot of the extension

Reference:

extensions/rime-manager/assets/rime-manager-auth
0xdhrv, line 1: issue: precompiled binary

This PR adds a precompiled executable under assets/rime-manager-auth. Even though the source and build script are included, we can't verify from the PR diff that the submitted binary was produced from that source.

Could you remove the compiled binary from the PR and keep only the source/build instructions?

extensions/rime-manager/src/pin-candidates.tsx
```diff
@@ -0,0 +1,207 @@
...
+  return (
+    <List
+      isLoading={isInspecting || isLoadingRules}
+      searchBarPlaceholder="Search input codes or candidates…"
+      actions={addAction ? <ActionPanel>{addAction}</ActionPanel> : undefined}
+    >
+      {sections.map(([schemaName, schemaRules]) => (
+        <List.Section key={schemaName} title={schemaName}>
+          {schemaRules.map((rule) => (
+            <List.Item
```
0xdhrv, line 166: **issue: candidate data bypasses authentication**

The PR description says candidate text and input codes are hidden by default and require local authentication to reveal, but this view renders both `rule.candidates` and `rule.code` immediately after loading persisted rules.

Could we apply the same reveal/authentication flow here as well? Otherwise opening **Pin Candidates** exposes the saved candidate data without authentication.

extensions/rime-manager/src/lib/files.ts
```diff
@@ -0,0 +1,64 @@
...
+    await mkdir(dirname(backupPath), { recursive: true });
+    await copyFile(path, backupPath);
+  }
+
+  const temporaryPath = `${path}.raycast-rime-manager.tmp`;
+  await writeFile(temporaryPath, ensureTrailingNewline(content), "utf8");
+  await rename(temporaryPath, path);
+}
+
+export async function createFullBackup(userDataDir: string, backupRoot: string): Promise<string> {
```
0xdhrv, line 43: **issue: backup directory can recurse into itself**

If the user selects the Rime user data directory itself as `backupRoot`, `resolvedBackupRoot === sourceRoot`, so `nestedBackupSegment` stays undefined.

That means the destination is created inside the directory currently being copied, which can cause the backup operation to recursively copy its own output.

Could we explicitly reject `resolvedBackupRoot === sourceRoot`, or otherwise exclude the generated destination from the copy?

extensions/rime-manager/native/rime-manager-auth.swift
```diff
@@ -0,0 +1,89 @@
...
+        guard let executableURL = Bundle.main.executableURL, let deeplink else { return }
+        let process = Process()
+        process.executableURL = executableURL
+        process.arguments = ["--open-deeplink", deeplink]
+        process.standardOutput = FileHandle.nullDevice
+        process.standardError = FileHandle.nullDevice
+        try? process.run()
+    }
+
+    private static func writeRevealGrant(at path: String?) throws {
```
0xdhrv, line 31: **suggestion: validate the grant path before writing**

The helper accepts `--grant-path` directly from its arguments and writes to that path after authentication.

Since this is a native executable with filesystem access, could we constrain the grant location to the expected Raycast support directory instead of accepting an arbitrary path? That would keep the helper's write capability scoped to its intended purpose.

extensions/rime-manager/src/lib/local-authentication.ts
```diff
@@ -0,0 +1,49 @@
...
+const execFileAsync = promisify(execFile);
+const REVEAL_GRANT_NAME = "candidate-rules-reveal-grant";
+const REVEAL_GRANT_TTL_MS = 60_000;
+
+type AuthenticationProcessError = Error & {
+  code?: string | number;
+  stderr?: string;
+};
+
+export async function authenticateToRevealRules(): Promise<boolean> {
```
0xdhrv, line 18: **suggestion: avoid relying only on helper stdout for success**

Authentication currently returns `true` when the helper prints `AUTHENTICATED`, while the actual reveal authorization is represented by the short-lived grant file.

Would it be safer to treat successful grant creation/consumption as the source of truth, rather than stdout? That would keep the authorization flow tied to the one-time grant instead of the subprocess output.

0xdhrv (changes requested):
Hey @HanryYu 👋 

---

PR #30229, opened 2026-08-15
Add Ollama Translate extension

scorpion7slayer:
## Description

Adds **Ollama Translate**, a macOS Raycast extension for private, meaning-first translation with models installed in the user's local Ollama library.

- Translates while typing or on demand.
- Lets users choose and remember both the source and target languages, with optional automatic source-language detection.
- Preserves meaning, tone, idioms, register, formatting, and terminology through a dedicated translation prompt and an optional second-pass review.
- Includes a searchable local-model picker, local translation history, language swapping, copying, and selected-text translation.
- Sends requests only to a loopback Ollama server; remote hosts, analytics, accounts, and API keys are not used.

Ollama must be installed and running, with at least one local multilingual chat model.

## Screenshots

Start translating with Ollama

Meaning-first local translation

Choose a local Ollama model

Choose source and target languages

## Validation

- `npm run lint`
- `npm run typecheck`
- `npm run build`
- Manually exercised in Raycast with a local Ollama model, including translation, source/target selection, history, and model selection.

## Checklist


pernielsentikaer:

- **Ollama AI** (`raycast-ollama`, 36,881 installs by @MassimilianoPasquini97) — Perform Local Inference with Ollama
  AI tools: ollama-web-fetch, ollama-web-search
  <sub>similarity 0.67</sub>

pernielsentikaer:
Thanks for this, it is a well built extension and the privacy posture (loopback-only enforcement, no analytics, prompt-injection guard in the system prompts) is nice to see.

Before we go further, one thing worth weighing. **Ollama AI** (`raycast-ollama`) already ships a `Translate` command backed by local Ollama, so the two overlap on the core idea. Overlap alone is not a blocker for us. But your version is clearly stronger at translation specifically: a dedicated view with translate-while-typing, a remembered from/to language pair, local history with reuse and swap, a second-pass meaning review, and offline detection used as a hint to the model.

So there are two reasonable paths:

1. **Fold it into Ollama AI.** Its Translate command is one of 21 and is fairly basic, so your work would be a real upgrade there, and 36k+ existing users would get it without installing anything new. This needs @MassimilianoPasquini97 on board, so open an issue on the extension first rather than a large PR.
2. **Keep it standalone.** Also fine. In that case please add a short "How this differs from Ollama AI" section to the README so users can pick between them, and reviewers can see the distinction at a glance.

Let us know which you prefer and we will continue from there. Separately, I have a few code and README notes that apply either way.

scorpion7slayer:
Thanks! I prefer option 2 and would like to keep Ollama Translate standalone as a focused translation workflow. I added a short "How this differs from Ollama AI" section to the README in commit 8bee17a3. Happy to address the additional code and README notes as well.

pernielsentikaer (changes requested):
Hi @scorpion7slayer 👋

Thanks for adding the comparison section 💪

I verified two Greptile findings that still need addressing:

- [Loading state](https://github.com/raycast/extensions/pull/30229#discussion_r3789377920): starting **Translate Again** during **Double-Check Meaning**, or vice versa, leaves the spinner active. Please reset the superseded operation’s loading state when replacing its controller. At translation start, use:
  ```ts
  setIsRefining(false);
  setIsTranslating(true);
  ```
  At refinement start, use the inverse.
- [Response parsing](https://github.com/raycast/extensions/pull/30229#discussion_r3789377922): `{"translation":"","semanticBrief":"analysis"}` becomes a successful translation and history entry. Please reject invalid structured responses rather than falling through to the plain-text fallback.

scorpion7slayer:
@pernielsentikaer  Thanks for checking these points! Both issues are fixed in 8457dda9. Replacing an operation now resets its loading state, and invalid structured responses are rejected instead of falling back to raw text.

PR #30279, opened 2026-08-16
Update Twos extension - find things by the text in their photos

SanKlein:
## Description

**Search Things** now matches text read out of a thing's photos, so a receipt, whiteboard, or screenshot is findable by what it says. The Twos API exposes this as a new read-only `photo_text` field on a thing.

A thing can be *just* a photo, with empty `text`. Those results already came back from search but rendered as "(empty)", which looked like a bug. They now show the extracted text, or a "Photo" label when the image has no readable text, and carry an image icon. **Copy Text** copies what the row displays instead of an empty string.

## Screencast

No UI structure changed — same list, same actions. Only the title/icon/accessory of a photo-backed result differ.

## Checklist

### DIFF
--- a/extensions/twos/CHANGELOG.md
+++ b/extensions/twos/CHANGELOG.md
@@ -1,7 +1,12 @@
 # Twos Changelog
 
-## [Open results in the desktop app] - 2026-08-15
+## [Photo results and opening in the desktop app] - {PR_MERGE_DATE}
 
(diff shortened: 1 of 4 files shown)


pernielsentikaer (approved):
Looks good — photo rows show the OCR text instead of (empty). Thanks!

PR #29535, opened 2026-07-17
feat: add youtube-transcript extension

Itsharshitgoat:
## Description

Adds the **Youtube Transcript** extension for Raycast.

Fetches and saves the transcript of a YouTube video in your downloads folder. Requires `yt-dlp` to be installed.

### Commands
- **Youtube Transcript** — Fetches and saves the transcript of a YouTube video (save to file or copy to clipboard)

### AI Tools
- **Youtube Transcript** — Fetches the transcript of a YouTube video given its URL so AI can read it

PR #30282, opened 2026-08-16
feat(fetch-youtube-transcript): add AI tool for reading transcripts

apoorv12:
## Description

Adds an **AI tool** to the existing [Fetch YouTube Transcript](https://www.raycast.com/apoorv_khandelwal/fetch-youtube-transcript) extension, so Raycast AI can read a video's transcript and summarise it, translate it, or turn it into notes.

This implements the suggestion from #29535. In that PR a separate `youtube-transcript` extension was proposed; @0xdhrv noted it substantially duplicated this extension and asked whether the AI tool could be added here instead. Thanks to @Itsharshitgoat for raising the use case.

### The AI tool

`src/tools/fetch-transcript.ts` reuses the extension's existing `extractVideoId` / `getVideoTranscript` helpers, so there is no duplicated fetching logic.

- Optional `language` override, defaulting to the existing `defaultLanguage` preference
- Optional `maxLength` / `offset`, so long videos can be read in chunks
- The response always reports the true `totalCharacters` plus a `truncated` flag, so a partial transcript is never presented to the model as a whole video. `offset` lets it page through the rest rather than silently summarising the first portion.
- Read-only, so no confirmation is required
- Three evals included

### The existing command is unchanged

Transcripts saved to the download folder are still saved **in full**, in the configured language, for videos of any length. Truncation and per-request language overrides live only in the tool file, which the command never imports, so they cannot affect it. The only behaviour change to the command is the download-folder fix below.

### Fixes on shared code paths

- `extractVideoId` rejected several valid URL forms, including the links produced by YouTube's own share button (`watch?feature=shared&v=...`), Shorts, and livestream replays. The change is purely widening — every URL accepted before is still accepted.
- Transcript downloads now use a private `mkdtemp` directory. The previous `Date.now()`-based name could collide between concurrent runs, letting one overwrite or delete another's file. This becomes far more likely once AI can invoke the tool.
- A configured download folder that has been deleted or is unwritable no longer discards an already-fetched transcript; it is recreated, or the transcript falls back to Downloads with the user informed.

### Housekeeping

- Dependency updates; removed a stale `overrides` block whose targets were no longer in the tree
- Scoped ESLint so generated files and the CommonJS config file no longer report errors

## Screencast

A new Store screenshot is included at `metadata/fetch-youtube-transcript-4.png`, captured with Raycast's Window Capture tool. It shows a real query in Raycast AI — the `Fetch Transcript` tool being called and the resulting summary.

## Checklist


0xdhrv (approved):
Looks good to me, approved ✅

Issue #30218, opened 2026-08-14
[Spotify Player] Your Library Command


AmirHomayounfard:
### Extension

https://www.raycast.com/mattisssa/spotify-player

### Raycast Version

0.71.5.0

### OS Version

macOS Tahoe 26.6.1 arm64

### Description

Error: Command terminated after reaching the extension memory limit (100 MB JS heap).

t.toRayError:index.mjs:1202:11072

---

JavaScript Stack
t.toRayError:index.mjs:1202:11072
Worker.<anonymous>:index.mjs:1202:8315
Worker.emit:node:events:531:35
[kOnExit]:worker:316:12

### Steps To Reproduce

Error: Command terminated after reaching the extension memory limit (100 MB JS heap).

t.toRayError:index.mjs:1202:11072

---

JavaScript Stack
t.toRayError:index.mjs:1202:11072
Worker.<anonymous>:index.mjs:1202:8315
Worker.emit:node:events:531:35
[kOnExit]:worker:316:12

raulkolaric:
Hey @AmirHomayounfard, I think I found a minor tweak that can solve this issue (Merged PR https://github.com/raycast/extensions/pull/30283)

Can you try to replicate it? It is already released. https://www.raycast.com/mattisssa/spotify-player

AmirHomayounfard:
Hi @raulkolaric 
Thank you very much, I updated the extension and it works, wonderful, thanks for your prompt response, appreciate it. 
Have a nice rest of the week.

raulkolaric:
Thank you! Have a great week too.

Issue #27607, opened 2026-05-02
[Spotify Player] Crash when "Your Library" run


arctic9:
### Extension

https://www.raycast.com/mattisssa/spotify-player

### Raycast Version

0.56.1.0

### OS Version

Windows 11 Education 10.0.26200.0 amd64

### Description

Error: Command terminated after reaching the extension memory limit (100 MB JS heap).

t.toRayError:index.mjs:623:7494

---

JavaScript Stack
t.toRayError:index.mjs:623:7494
Worker.<anonymous>:index.mjs:623:4740
Worker.emit:node:events:531:35
[kOnExit]:worker:316:12

### Steps To Reproduce

1. Open raycast
2. Run the "your library" command

### Current Behaviour

Extension crashes after half a second or so once the command is run.

arctic9:
**Update**:
Other commands such as select device work as normal. I think I tried restarting Raycast but that didn't fix the issue.

Went into extension diagnostics and pressed terminated all. From what I could see, Raycast then quit and after reopening the app this command works as normal. I don't know what caused this problem in the first place.

I have limited experience with github, should I now close this report?

arctic9:
This problem has reoccurred but this last workaround still works

hugodemenez:
Hello @arctic9, do you know the steps to reproduce the error? 
Unfortunately, I am not on windows and it might be a Os related issue that I won't be able to debug. But let's see.
Thanks ahead

arctic9:
I don't know much more than what I put in the original post unfortunately. 
I can offer this screen recording?

https://github.com/user-attachments/assets/867489f7-3085-4f4e-9999-245e9de92f91

Issue #23405, opened 2025-12-04
[Spotify Player] ...JS heap out of memory error


MoZiSaan:
### Extension

https://www.raycast.com/mattisssa/spotify-player

### Raycast Version

0.39.0.0

### macOS Version

Windows 11

### Description

When using "Your library" command or "Search" command and entries or results start to load this error always happen 

Error: Worker terminated due to reaching memory limit: JS heap out of memory

[kOnExit]:worker:313:26

---
---

JavaScript Stack
[kOnExit]:worker:313:26
Worker.<computed>.onexit:worker:229:20
Worker.callbackTrampoline:async_hooks:130:17

<img width="750" height="475" alt="Image" src="https://github.com/user-attachments/assets/e0ba171e-b2a5-45c4-afd4-99ee169d5d8e" /> 

<img width="750" height="475" alt="Image" src="https://github.com/user-attachments/assets/88c0b81d-529c-4d61-985b-b58235831d45" />

### Steps To Reproduce

1. Open "Your library" command or "Search" command
2. Wait a short while as result or entries  load
3. Extension crashes with JS heap out of memory error

### Current Behaviour

Extension crashes with JS heap out of memory error

<img width="750" height="475" alt="Image" src="https://github.com/user-attachments/assets/c370577b-d0f0-48b4-b846-f5741cb0bfa3" />

### Expected Behaviour

To load results and 

<img width="750" height="475" alt="Image" src="https://github.com/user-attachments/assets/ed940f3a-8390-4f1f-b750-f0079f788b77" />

without crash

PR #30283, opened 2026-08-16
fix(spotify-player): limit liked songs library fetch

raulkolaric:
## Summary

I noticed that the liked songs entry calls `useYourLibrary()` without selecting a category. Since the hook defaults to `all`, opening the Your Library command can fetch unrelated library categories again even though this component only needs tracks.

I changed the component to request the `tracks` category explicitly. This keeps the visible behavior the same while avoiding unnecessary library data and reducing the chance of contributing to the extension's 100 MB heap failures.

## Validation

- `npm run lint --prefix extensions/spotify-player`
- `npm run build --prefix extensions/spotify-player`

I have not been able to reproduce the memory-limit crash locally, so I kept this change intentionally small and limited to the redundant library fetch.

Related issue: #30218

Related reports: #27607, #23405

### DIFF
--- a/extensions/spotify-player/CHANGELOG.md
+++ b/extensions/spotify-player/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Spotify Player Changelog
 
+## [Fix Your Library Memory Usage] - {PR_MERGE_DATE}
+
+- Load only liked songs when rendering the liked songs entry instead of fetching every library category.
+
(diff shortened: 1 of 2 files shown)


pernielsentikaer (approved):
Looks good. The liked-songs row only needs tracks, so this avoids the full library fetch.

PR #30287, opened 2026-08-16
[Claude] Fix API compatibility with current models

chrismessina:
Hi @pernielsentikaer 👋 — this is the first of the two PRs, the API-compatibility fixes only. #29798 stays in draft and will carry the consolidation, storage layer and features once this lands.

## What's broken today

Asking a question on a current model fails. Sampling parameters were removed on Claude Opus 4.7 and later, and the extension sends `temperature` on every request, so Opus 4.7, Sonnet 5 and Opus 5 all return a 400. Both the streaming and non-streaming paths built their own request body, which is what let them drift.

With **Stream Responses** off there's a second failure: the non-streaming path never clamped `max_tokens`. The SDK applies two independent gates — a 10-minute runtime projection, and a per-model cap of 8192 for the Opus 4 and 4.1 ids — and both throw before a request is sent. Both throw *synchronously*, before any promise exists, so the extension's `.catch()` never attached: the toast spun on "Getting your answer…" indefinitely with no error and no way out.

Three quieter ones. `max_tokens` came from a heuristic matching id substrings like `-4-` and `claude-opus-4`, so anything outside those shapes — `claude-opus-5` included — fell through to the legacy 4096 and silently capped presets on current models. The models endpoint is cursor-paginated and defaults to 20 per page, so a single unpaginated request truncated the list. And a long conversation sent its whole transcript every time, so once one outgrew the context window every later request failed identically and the conversation was permanently unusable with no way out from the UI.

## The changes

- **Shared request builder** — `buildModelRequestParams` in `src/utils/models.ts`. Both paths construct the request through it, so they cannot drift again. `temperature` is included only for models that accept it; `max_tokens` is clamped for non-streaming and left alone for streaming.
- **`max_tokens` from the models API** — read per model from `/v1/models`, with the name-based tiers kept only as a fallback for responses that omit the field. Replaces the heuristic in `src/views/model/form.tsx`.
- **Model list pagination** — follows `has_more`/`last_id` and requests the maximum page size.
- **Non-streaming clamp** — `MAX_NONSTREAMING_TOKENS = 8192`, the lowest cap the SDK enforces for any model, so it clears both gates without depending on `@anthropic-ai/sdk/internal/constants` (which the package export map doesn't expose, so a local copy of the per-model map would drift). Clamped rather than thrown: a shorter answer beats no answer. The call is also invoked so that a synchronous throw still reaches the error handler instead of abandoning the UI.
- **Context-window bounding** — `src/utils/contextWindow.ts` bounds the request to the model's input budget, measured with the API's own `countTokens()` and found by binary search (a long conversation shouldn't cost a linear scan of network calls). The budget is the advertised `max_input_tokens`; output is *not* reserved out of it, since the API documents the two as separate budgets. The full transcript is untouched for display, and the user is told when older turns were left out — and told when the reason was a failed token count rather than context length.
- **Stream ownership** — aborting a stream emits `"abort"`, not `"error"`, and the SDK raises an unhandled rejection when nothing listens; an aborted stream also still emits `"end"`, so a superseded request used to commit its partial answer and clear loading state under the live one. Both are handled, guarded by stream identity.

The `@anthropic-ai/sdk` bump is required — `countTokens()` does not exist on 0.18. The SDK surface this extension uses is small (constructor, `messages.stream`, `messages.create`, `countTokens`, the `MessageStream` type) and one response shape changed with it: `content` is now a union including thinking blocks, so text is filtered rather than read from `content[0]`.

## Not in this PR

The 5→3 command consolidation, the storage layer and its migration, and the new features — all deliberately held for the second PR. No commands are added or removed here, and `assets/icon@dark.png` is untouched.

## Verification

`npx tsc --noEmit`, `npx ray build`, `npx ray lint` all exit 0.

The request-building logic was checked against four deliberately reintroduced regressions — unconditional `temperature`, a missing clamp, the legacy-4096 fallback, and the 21333 clamp that Opus 4.1 still rejects — confirming each check actually fails when the behavior breaks. The non-streaming parameters are asserted against the SDK's own gate for five model ids rather than against our own constant, which is what surfaced the per-model 8192 cap in the first place.

### DIFF
--- a/extensions/claude/CHANGELOG.md
+++ b/extensions/claude/CHANGELOG.md
@@ -1,5 +1,16 @@
 # Claude Changelog
 
+## [API compatibility fixes] - {PR_MERGE_DATE}
+
+- Fix: Asking a question on Claude Opus 4.7 or newer (including Opus 4.7, Sonnet 5, and Opus 5) failed with a 400 error. Sampling parameters were removed on those models, and the extension was still sending `temperature` on every request.
+- Fix: With "Stream Responses" turned off, requests could fail outright instead of answering — and on Opus 4 and 4.1 the failure left the loading indicator spinning with no error shown. Non-streaming requests are now capped at the limit the API accepts, and if a particular model rejects even that, the request is retried at a smaller size rather than failing — so a long answer comes back shorter rather than not at all, and nothing is left hanging.
+- Fix: Presets on newer models were capped at 4,096 output tokens. The maximum is now read from the API for each model instead of guessed from its name.
+- Fix: Only the first 20 models were listed. The full model list is now loaded.
+- Fix: Long conversations stopped working permanently once they outgrew the model's context window. Older turns are now left out of the request when needed — the conversation stays readable in full, and a note says when this happened and why.
+- Fix: Asking a new question while an answer was still streaming could let the abandoned request overwrite the new one's result and leave the newer answer unstoppable.
+- Fix: Asking a second question before the first answered could let the earlier request report success and stop the loading indicator while the newer answer was still being written.
+- Improvement: Updated the Anthropic SDK, which the token-counting used by the fix above requires.
+
 ## [Fix memory leak] - 2026-02-01
 
 - Fix: Resolved JS heap out of memory error by throttling UI updates during streaming responses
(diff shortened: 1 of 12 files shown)


pernielsentikaer (approved):
Looks good — this is a careful fix for the current-model 400s, the non-streaming hang, and conversations that died once they outgrew the window.

Shared request params, models pagination, the 21333 clamp + 8192 retry, and the stream / request-generation guards all look correct. Thanks for the write-up and for landing the Greptile follow-ups @chrismessina.

PR #30294, opened 2026-08-17
Update file-tidy extension

fanhefeng:
## Description

Two changes for File Tidy:

- **New extension icon**, with all three Store screenshots re-shot to match (same three compositions as before; `ray lint` metadata checks pass).
- **Fix — previews no longer touch the destination.** The perceptual-hash cache used by similar-image detection was saved during analysis, before the user confirms anything, which could create the destination folder (and a `.tidy` folder inside it) just for previewing a plan — and silently skipped the "create destination?" consent prompt. The cache is now written only after a run completes, keyed by each image's final archived path so later runs still get cache hits. Covered by two new regression tests in the shared core (60/60 passing).

## Screencast

Updated screenshots are included in `metadata/` (form, plan top, plan bottom).

## Checklist


fanhefeng:
Both cache-related review comments verified and addressed in be8ac13:

1. **Cross-volume mtime**: confirmed — the EXDEV path's `copyFileSync` gave the copy a fresh mtime, so the re-keyed cache entry failed its size+mtime check on the next analysis and every image moved across volumes was decoded again. `moveFile` now restores the source's timestamps with `utimes` after copy verification (best-effort — a verified move doesn't fail over metadata), so both move paths have the same semantics and the entry stays valid.

2. **Snapshot overwrite**: confirmed — `saveHashCache` replaced the file with this run's analysis-time snapshot, dropping entries written by a run that completed in between. Worth noting this window predates this PR (the analysis-time save had the same read-modify-write race); deferring the write stretched it across the confirmation step. Saving now merges over the on-disk state: this run's entries win on shared keys, disk-only entries survive while their file still exists, and vanished paths are still pruned.

Both are covered by new tests (cross-volume move keeps mtime and the entry still hits; merge keeps another run's entries while dead paths drop).

0xdhrv (approved):
Looks good to me, approved ✅

PR #30302, opened 2026-08-17
Add ai-to-pdf extension

guuskaashoek:
## Description

**AI to PDF** converts Adobe Illustrator `.ai` files to print-ready PDF from Raycast, with the bleed either read from the file, set by hand, or switched off.

It does not parse the `.ai` itself — it drives Illustrator's own *Save as PDF* over Apple events, so PDF presets, colour management, fonts and transparency behave exactly as they do in the export dialog.

**Commands**

| --- | --- | --- |
| Convert AI to PDF | view | File picker (pre-filled from the Finder selection), bleed choice, PDF preset and output folder |
| Convert Finder Selection with Its Own Bleed | no-view | Converts the Finder selection, each file with the bleed its own document defines |
| Convert Finder Selection with Custom Bleed | no-view | Same, with the bleed from the preferences |
| Convert Finder Selection Without Bleed | no-view | Same, at trim size |

**Notes for review**

- *Reading a document's bleed.* Illustrator does not expose it to scripting — it is on neither `Document` nor its reflection, and only `DocumentPreset` can set one. A PDF-compatible `.ai` does record it, as the gap between the TrimBox and BleedBox of its PDF part, so `src/lib/ai-file.ts` reads it from the file. Verified against 22 real production files, reading back as 0, 1, 3 and 10 mm. Files saved without PDF compatibility stop with a message rather than a guess.
- *Bleed rounding.* Illustrator truncates a scripted bleed to whole points, so a document's 3 mm (8.504 pt) would export as 8 pt — 2.82 mm, less than asked and enough to fail a 3 mm preflight check. Bleeds are rounded up instead; the fraction gained sits at the outer bleed edge and is cut away.
- *`Use Document Bleed Settings`.* A preset saved with that option always uses the document's bleed and writes it out unrounded — the only route to a bleed that is not a whole number of points. There is no scripting flag for it, so the preset's `.joboptions` is read: such presets are used for the "From the file" mode and refused for the other two, where they would silently override the requested bleed.
- Printer's marks are explicitly switched off after the preset is applied, so a preset that enables them cannot slip them into the PDF.
- The source `.ai` is never modified. A file already open in Illustrator is converted from disk and left open; if it has unsaved changes the conversion stops instead of exporting stale artwork.
- No network access, no credentials, no external services. The only dependencies are `@raycast/api` and `@raycast/utils`.

## Screencast

Three screenshots are in `metadata/`: the conversion form, the bleed menu with its three choices, and the PDF preset list.

## Checklist


0xdhrv (approved):
Looks good to me, approved ✅

PR #30303, opened 2026-08-17
Add Jotaid extension

leeronzhang:
## Description

Adds an extension for [Jotaid](https://jotaid.com), a Markdown notes app for macOS with backlinks, a knowledge graph and a co-occurrence matrix.

**Search Notes** — matches titles and, optionally, note bodies. Because a body match often sits a screen or two down a note whose title says nothing, the row shows the surrounding words and the preview quotes the match at the top and marks every hit. `↵` opens the note in Jotaid.

**Create Note** — a place to type; whatever you write is filed in the Inbox.

**Quick Capture** — sends the text selected in whatever app you are in (or, where macOS will not hand a selection over, the clipboard) to the Inbox without stealing focus. Captures from a browser record the page's title and link rather than just the browser's name.

## Notes for the reviewer

- **Reading** uses `useSQL` against the app's own Core Data store, **read-only** — the extension never writes to it. The `permissionView` is returned when Full Disk Access is missing, and `permissionPriming` explains why it is needed.
- **Saving** goes through the app's `jotaid://` URL scheme, the same contract its PopClip extension already uses. Notes are opened in the background (`open -g`) so the capture never pulls the app to the front.
- **Nothing leaves the machine** and there is no analytics of any kind.
- Requires Jotaid 1.1.3 or later; the extension degrades to a friendly empty state when the app has never run.
- Only two screenshots: `Quick Capture` is a no-view command, so the two view commands are the whole of the UI.

## Checklist


pernielsentikaer:

- **SnapJot** (`snap-jot`, 382 installs by @k41531) — Create a memo with a timestamp
  <sub>similarity 0.69</sub>
- **Search Joplin Notes** (`search-joplin-notes`, 702 installs by @wisdom-plus) — Search and view in your joplin notes
  <sub>similarity 0.64</sub>
- **iA Writer** (`raycast-ia-writer`, 801 installs by @xvvhang) — Create or search notes with Raycast
  <sub>similarity 0.64</sub>
- **PromptNote** (`promptnote`, 130 installs by @dimpurr) — Manage your AI prompts with versioning and cloud sync
  <sub>similarity 0.64</sub>
- **Stik** (`stik`, 8 installs by @0xMassi) — Quick capture and browse your Stik notes. Create, search, organize, and manage markdown notes stored locally on your machine.
  <sub>similarity 0.63</sub>

leeronzhang:
Thanks — the README now states this outright.

Each of the flagged extensions targets a different notes app, so the overlap is in the wording rather than in what they do. I've added a paragraph saying plainly that this one works only with Jotaid (it reads Jotaid's own library and files notes through Jotaid's URL scheme, so it can't be pointed elsewhere), plus what follows from being app-specific: search knows about Jotaid's structure, and captures are filed by Jotaid itself and inherit its own handling — the Inbox, the `clipping` tag, and the title its AI fills in.

pernielsentikaer (changes requested):
Hi @leeronzhang 👋

- **Greptile P1 (open)** in `src/lib/browser.ts` — when splitting the AppleScript `url\ttitle` output, keep the rest of the title if it contains tabs, e.g. `const [url, ...titleParts] = output.split("\t"); const title = titleParts.join("\t");`. After that lands, Greptile should be able to re-score to 5/5 (currently 4/5).
- In **Search Notes**, please destructure `error` from `useSQL` and show a helpful `List.EmptyView` when the query fails (so a DB/read error doesn’t look like an empty library).

---

leeronzhang:
Hi @pernielsentikaer 👋 Thanks for testing it — both points are addressed:

- **Greptile P1 in `src/lib/browser.ts`** — the AppleScript output is now split on the first tab only (`const [url, ...titleParts] = output.split("\t")`, then `titleParts.join("\t")`), so a page title containing tabs is kept whole instead of being truncated at the first one.
- **Search Notes** — `error` is now destructured from `useSQL` and the empty view branches on it: a failed read shows a red warning icon, "Could Not Read Your Jotaid Library", and the driver's own message, so a database error no longer reads as an empty library. The search bar stays in place so the query can be retried.

`ray lint` and `ray build` are both clean. Ready for another look 🔥

pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #30311, opened 2026-08-18
Update context7 extension

chrismessina:
## Description

This adds offline documentation to Context7 and exposes the extension to Raycast AI.

Save a library and its documentation is kept on disk, so opening it is instant and works without a network connection. **Search Documentation** now searches every saved library at once by default, narrowing to one library via the dropdown — which is also what re-enables Context7's semantic search, since that endpoint takes a single library ID. Individual snippets can be saved and found again across libraries.

Two API details shaped the implementation and are worth flagging for review:

- **`/api/v2/libs/search` is capped at 5 results** with no parameter to raise it, which is a poor experience for a human scanning a list. Search calls the undocumented `/api/v1/search` (30 results, identical response schema) and **falls back to v2** on any error or unrecognised payload, so it degrades to the documented endpoint rather than breaking.
- **v2 cannot return a library without a query**, so browsing and the offline copy use `/api/v1/{libraryId}?type=json&tokens=N`. There is no v2 equivalent.

The API key is now **optional** — both v2 endpoints answer anonymously at a lower monthly quota, so the extension works before a key is entered, and a rate-limited request explains that adding one raises the limit.

`@chrismessina/raycast-logger` is added for structured request/response logging behind a `verboseLogging` preference (off by default). This extension is entirely API-driven, and the logging is aimed at whoever debugs it next rather than at me specifically. Happy to drop it if you would rather not take the dependency.

`author` is unchanged.

### Changes

- **My Libraries** — save a library and keep its documentation on disk for instant, offline search; each row shows when it was last captured, with Refresh and Refresh All
- **My Snippets** — save individual snippets and search them across every library; snippets are content snapshots, so a library refresh cannot disturb them
- **Search Documentation** — no longer requires a library ID to launch; searches all saved libraries by default, and narrowing to one adds Context7's semantic search
- **AI tools** — four tools (`search-libraries`, `get-documentation`, `search-my-libraries`, `add-library`) with instructions and evals in `ai.yaml`
- Library search returns up to 30 results instead of 5, with client-side sorting by relevance, popularity, recency, trust score, or snippet count
- Snippet rows carry a one-line description, so a library that reuses a title like "Basic example" is no longer a wall of identical rows
- Fixed keyboard shortcuts that were silently broken on Windows — several actions used `cmd`-only bindings despite the manifest declaring Windows support
- Added a "Copy Error" action to every failure toast
- Made the Context7 API key optional
- Replaced the deprecated `Icon.TextDocument`

## Screencast

Six updated screenshots are included in `metadata/`, covering all four commands.

## Checklist


chrismessina:
@greptile on the remaining stale-lock reclamation note — leaving it as is, deliberately.

`withFileLock` uses exclusive create (`open(path, "wx")`) plus an ownership token, a 1s heartbeat, and an age re-check immediately before reclaiming. Deleting a successor's lock now requires a process to stall past the 10s stale threshold *between two adjacent syscalls* — by which point its own heartbeat has stopped and it is the hung process being reclaimed.

Closing it entirely isn't possible with portable Node `fs`, which has no conditional unlink. I tried three constructions and measured each:

- `rename`-to-claim — moves the window; a reclaimer that grabs a live successor's lock has to restore it, which is worse.
- `link`-based — identical check-then-act gap.
- Generation-numbered lock files (each acquisition owns a unique path, so a reclaimer physically cannot delete a successor's) — closes this race but opens a structural one: when the directory empties the numbering resets, so a stale observation can claim a higher generation while another process claims a lower one from an empty directory. Two holders. Measured 5/6 writes surviving, against 8/8 for the current implementation.

`proper-lockfile` carries the same caveat and resolves it the same way (heartbeat + accepted residual). Fully closing it needs an OS-level advisory lock, which isn't something I'd add to an extension I don't own.

Scope of the residual: a lost update needs a crashed holder, two simultaneous reclaimers, and a multi-second stall mid-syscall-pair. Writes are atomic (temp file + rename), so the failure mode is one lost add/remove, never a corrupted store. The rationale is documented in `atomic-file.ts` so it isn't re-litigated or "simplified" away — the token, heartbeat, and age check each close a different reachable interleaving.

The genuinely reachable defects from earlier review rounds are fixed: decide-before-lock in toggles, a mutator bypassing the lock, unconditional lock deletion, a live holder judged stale, payload cleanup outside the lock, and a prune keep-set captured outside the lock.

0xdhrv (approved):
Looks good to me, approved ✅

PR #30313, opened 2026-08-18
[Herdr] Fix agent naming and session selection

bendrucker:
## Description

Update to the Herdr extension fixing bugs across agent naming, session selection, and command actions, found by auditing the extension against a live herdr 0.8.0 server and the herdr CLI source.

The visible bug: the Prompt Agent dropdown and dashboard rendered unnamed agents as a box glyph. The extension fell back to `display_agent`, a Nerd Font icon herdr renders in its own status line, which Raycast cannot display as text. Unnamed agents now show their working directory basename, falling back to the agent kind, and prompt history moves to a new storage key so saved glyph entries do not resurface.

The deepest bug: the extension selected the configured session by setting the `HERDR_SESSION` environment variable, which the CLI reads only as a fallback after the `--session` flag and socket overrides. Every CLI call and terminal launch now passes `--session` explicitly, including for the default session, because launched terminal clients inherit the Raycast process environment where a leaked `HERDR_SESSION` would otherwise retarget them. Focus detection normalizes the session it parses from a client's argv the same way, so the extension's own clients still match.

Also fixed:

- A started agent or delivered prompt no longer reports as failed when a follow-up focus or history write fails. That failure invited a re-submit that duplicated the workspace or prompt.
- Newly started agents are prompted and focused by pane id, which herdr resolves unambiguously, and a defaulted agent name gets a numeric suffix instead of failing when a live agent of the same kind exists.
- The Prompt Agent target derives from the live agents list at render, so a background refresh cannot reroute a prompt.
- Splits pass the snapshotted pane explicitly instead of splitting whatever pane the server has focused at execution time.
- Manage Worktrees lists only linked worktrees. It previously offered Remove on main checkouts.
- Renamed panes read the `label` field the CLI sends, environment values keep commas, and integration status lines carrying both a version and a path parse correctly.
- The Ghostty focus marker clears through `herdr terminal title clear`, including when the set call times out client-side.
- Menu bar actions surface errors as toasts. Custom terminal launcher templates expand `{command}` inside larger words and reject embedded `{args}`.

CLI behaviors were verified against a running herdr server (error envelopes on stderr, `label` on renamed panes, `--session default` resolving to the same running server) and against the CLI source where observation was not enough (session resolution precedence, agent target resolution, duplicate name rejection). Added unit tests covering the name fallback, session flag construction, terminal launches, launcher template expansion, and the parsers.

## Screencast

No new UI. The fixes change the behavior of existing commands.

## Checklist

### DIFF
--- a/extensions/herdr/CHANGELOG.md
+++ b/extensions/herdr/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## [Bug Fixes] - {PR_MERGE_DATE}
+## [Bug Fixes] - 2026-08-18
 
 - Show a readable name for unnamed agents instead of the status-line glyph.
 - Select the configured session with the `--session` flag on every CLI call and terminal launch, including the default session.
--- a/extensions/herdr/CHANGELOG.md
+++ b/extensions/herdr/CHANGELOG.md
@@ -1,5 +1,19 @@
 # Changelog
 
+## [Bug Fixes] - {PR_MERGE_DATE}
+
+- Show a readable name for unnamed agents instead of the status-line glyph.
+- Select the configured session with the `--session` flag on every CLI call and terminal launch, including the default session.
+- Report a started agent or delivered prompt as successful even when a follow-up focus or history step fails.
+- Prompt and focus a newly started agent by pane id. Default names get a numeric suffix when the kind collides with a live agent.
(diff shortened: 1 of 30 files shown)


0xdhrv (approved):
Looks good to me, approved ✅

PR #30327, opened 2026-08-18
Add explicit OBS output commands

chris-cald:
## Summary

Adds explicit Raycast commands for OBS output controls, alongside the existing toggle commands:

- Start/Stop Recording
- Pause/Resume Recording
- Start/Stop Stream
- Start/Stop Virtual Camera
- Start/Stop Replay Buffer

The new commands check the current OBS output state before sending start/stop/pause/resume requests, so they are safe to bind to automation workflows and keyboard shortcuts where an idempotent command is preferable to a toggle.

`Set Scene` now also accepts an optional `sceneName` argument. When supplied, the command switches directly to the matching scene. If no scene name is supplied and OBS only has one scene, it selects that scene automatically. Otherwise, the existing scene picker opens as before.

This also updates OBS connection handling to launch OBS Studio automatically when it is installed but not already reachable over WebSocket, then retries the connection once.

## Validation

- `npm run lint`
- `npm run build`

### DIFF
--- a/extensions/obs-control/CHANGELOG.md
+++ b/extensions/obs-control/CHANGELOG.md
@@ -1,5 +1,12 @@
 # OBS Control Changelog
 
+## [New Commands] - 2026-08-18
+
+- Added explicit start and stop commands for recording, streaming, virtual camera, and replay buffer
+- Added explicit pause and resume recording commands
+- Added an optional scene name argument to Set Scene for automation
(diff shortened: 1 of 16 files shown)


chris-cald:
Continued in #31371. The branch has been rebased onto current main, the fixed three-second OBS startup delay has been replaced with bounded WebSocket readiness polling, and the replacement PR is ready for review.

PR #30333, opened 2026-08-19
Update apple-reminders extension

### LINKED ISSUE #28467: [Apple Reminders] Date field returns "No Results" for relative times ("1 hour", "3 hours", "1 year")
### Extension

https://www.raycast.com/raycast/apple-reminders

### Raycast Version

0.62.0.0

### OS Version

Tahoe 26.3.1 arm64

### Description

In the Create Reminder command, the Date field's natural-language input no longer recognizes relative time expressions. 

Typing a duration such as "30 minutes", "1 hour", "3 hours", "3 days", or "1 year" returns "No Results" in the typeahead instead of offering a parsed due date to select. **As a result, no relative due date can be set from this screen and the reminder is saved with "No Date".**

This is a regression — relative scheduling worked reliably on a previous version. It's the core of the quick-capture flow and the main reason I create reminders from Raycast rather than from the Reminders app.

### Steps To Reproduce

1. Open the Create Reminder command.
2. Click the Date field and type a relative time, e.g. "1 hour".
3. Observe the typeahead suggestions. 

### Current Behaviour

Image
Image
Image

### Expected Behaviour

A parsed option appears for the typed duration, and selecting it sets the due date accordingly (e.g. 1 hour from now).

### LINKED ISSUE #28511: [Apple Reminders] Quick add reminder errors
### Extension

https://www.raycast.com/raycast/apple-reminders

### Raycast Version

1.104.19

### OS Version

26.3.1

### Description

"Quick Add Reminder" does not work at all, whether using the optional argument Note List or not. It just returns an error of "Unable to add reminder". Other extension functionality working normally is "Create Reminder" action and the "My Reminders" action. Just the "Quick add reminder" is showing this bug so I don't think it's permission based or something like this. 

### Steps To Reproduce

Attempt to create a quick reminder. 

### Current Behaviour

Unable to create reminder

### Expected Behaviour

Reminder should be created

### LINKED ISSUE #29065: [Apple Reminders] no 'natural language' to set reminder in v2
### Extension

https://www.raycast.com/raycast/apple-reminders

### Raycast Version

v0.66.1.0 (Beta)

### OS Version

26.5.1

### Description

in the classic raycast i was able to set a reminder with 'natural language' like 'in 10 minutes' or something. 
Image

in the raycast v2 this doesn't work anymore. 
Image

### Steps To Reproduce

1. open classic raycast
2. open 'create reminder' command
3. in the date field you can write in minutes and the 'date' is suggested.
4. open raycast v2
5. open 'create reminder' command
6. 3. in the date field you can write in minutes and recognize that it is not possible, the 'date' is not suggested anymore

### Current Behaviour

_No response_

### Expected Behaviour

_No response_

### LINKED ISSUE #29284: [Apple Reminders] due-date parsing not working correctly
### Extension

https://www.raycast.com/raycast/apple-reminders

### Raycast Version

0.67.1.0

### OS Version

macOS Tahoe 26.5.2 arm64

### Description

it is not possible anymore, to set a reminder to a specific time other than to the whole hour. see the screenshots attached. 
raycast v2:
Image

classic raycast (where it works as expected):
Image

(probably plays into the issue #29065 )

### Steps To Reproduce

1. launch 'create reminder' command
2. try setting the due date to something other than the whole hour, for example 3:45 pm.
3. see that it doesn't work:(

### Current Behaviour

_No response_

### Expected Behaviour

it should suggest the date and time i input like it did in the classic raycast.

### LINKED ISSUE #30159: [Apple Reminders] - "hours" shortcut doesn't work anymore
### Extension

https://www.raycast.com/raycast/apple-reminders

### Raycast Version

0.71.4.0

### OS Version

macOS Tahoe 26.4.1 arm64

### Description

Image

Previously, I was able to type just the letter "h" after any number and it would suggest "x hours". Now it suggests months instead of hours when I type H (and also when i type "h"), and now I have to type out the entire word "hour" to make it work. 

### Steps To Reproduce

create a reminder via the raycast extension and type any number followed by "h" i n the date field.

### Current Behaviour

_No response_

### Expected Behaviour

_No response_

0xdhrv:
## Description

- Fixes: https://github.com/raycast/extensions/issues/30159
- Fixes: https://github.com/raycast/extensions/issues/29284
- Fixes: https://github.com/raycast/extensions/issues/29065
- Fixes: https://github.com/raycast/extensions/issues/28511
- Fixes: https://github.com/raycast/extensions/issues/28467

## Screencast

N/A

## Checklist


pernielsentikaer:
Hi @0xdhrv

Thanks for picking this up — restoring `1h` / relative times in Create Reminder and giving Quick Add a real local fallback is the right fix for #30159, #28467, #29065, and #28511. The shared duration parser and fenced-JSON extraction look solid.

A few date-path bugs I would like tightened before we merge:

1. **Stale natural-language due date** in `extensions/apple-reminders/src/create-reminder.tsx:236-240`. `handleDueDateTextChange` only writes `dueDate` / `nlpParseRef` when parse succeeds. If someone types `tomorrow` and then replaces it with text that does not parse, the calendar and submit still use the previous date. After a successful create, `dateText` and `nlpParseRef` are also left as-is (`:177-182`), so the next reminder from the same form inherits that date. Please clear both on a failed parse (debounced is fine so mid-keystroke does not flicker) and reset them after a successful submit.

2. **Local times get shifted twice** in `extensions/apple-reminders/src/quick-add-reminder.tsx:136-138`. When AI omits `dueDate`, `resolveQuickAddReminder` fills it via `formatDueDate` → `toISOString()` (already a correct UTC instant). This branch then runs `applyAiLocalTimezone` anyway, so outside UTC the reminder lands hours off (e.g. “3:45pm” in CEST becomes 1:45pm). Only apply the offset when the datetime came from AI, e.g. `if (newReminder.dueDate && resolvedReminder.dueDate?.includes("T"))`.

3. **Fallback can create a second reminder** in `extensions/apple-reminders/src/quick-add-reminder.tsx:147-150`. The inner `catch` wraps `createReminder` + `runStoredPostCreateActions` + the success toast, then always calls `addReminderFromText`. If create succeeds and a post-create action (or bad LocalStorage JSON) throws, the user gets two reminders. Please catch only around `askAI` / resolve, or skip the fallback once `createReminder` has committed.

4. **Location fields dropped** in `extensions/apple-reminders/src/quick-add-reminder.tsx:205-218`. The old path passed the AI object straight into `createReminder`. `toNewReminder` copies title / list / dueDate / notes / priority / recurrence but not `address`, `proximity`, or `radius`, so “remind me when I get home” style Quick Adds silently lose the location. Please thread those through (and onto `ParsedQuickAddReminder`).

Smaller, non-blocking: Sort By in `ReminderActions.tsx:325` moved from ⌘⇧S to `Keyboard.Shortcut.Common.Duplicate` (⌘D), and Copy Reminder Title (`:371`) moved from ⌘. to `Common.Pin` (⌘⇧P). That will surprise existing users and is not in the changelog — worth keeping the old chords unless the remap is intentional.

Happy to take another look once those date paths are safe.

---

pernielsentikaer (approved):
Thanks @0xdhrv — the date-path fixes look good. Nice work restoring relative times in Create Reminder and tightening Quick Add.

Issue #30240, opened 2026-08-16
[Radarr] Windows Support


gabriel-vanca:
### Extension

https://www.raycast.com/xjo_nd/radarr

### Description

Would you please consider adding Windows support for this extension?

@JohanDevl

JohanDevl:
Hi @gabriel-vanca 👋🏻

Windows support is now in — PR is up: #30352  😉

Just waiting on the Raycast team's review before it can be merged 😄

JohanDevl:
@raycastbot close this issue

gabriel-vanca:
Cheers, @JohanDevl !

PR #30352, opened 2026-08-19
Update radarr extension

JohanDevl:
## Description

Adds Windows support to the Radarr extension, requested in #30240, and updates the dependencies.

### Windows support

- `platforms` now declares both `macOS` and `Windows`.
- The extension had exactly one platform-bound code path: 14 call sites opened the extension preferences through a `raycast://extensions/preferences` deeplink built from the `RAYCAST_SCHEME` env var. These are replaced by `openExtensionPreferences()`, wrapped in a small shared `OpenPreferencesAction` component.

Everything else was already portable, and I checked rather than assumed: no AppleScript or shell exec, no `process.platform`, no local filesystem paths or `path.join`, and **no declared keyboard shortcuts anywhere**, so there is no cmd-vs-ctrl modifier conflict to resolve. The only external-launch actions are `Action.OpenInBrowser` with `http(s)` URLs. The `Copy File Path` action copies a path returned by the Radarr *server*, so its separators are intentionally not the client's concern.

### Dependencies

- `@raycast/api` → v2. This also pulls in an esbuild release that clears GHSA-g7r4-m6w7-qqqr, which is a Windows-specific advisory, so it is relevant to this PR rather than incidental.
- Updated `@raycast/utils`, `@raycast/eslint-config`, `eslint`, `prettier`.
- `@types/node` and `@types/react` are pinned to the exact versions `@raycast/api` itself ships, so the type surface matches the runtime instead of drifting ahead of it. This also collapses a duplicate copy of `@types/node` in the tree.
- `brace-expansion` overrides refreshed to patched releases. `npm audit` is clean.

`typescript` deliberately stays on 5.9.x: `@raycast/eslint-config@2.2.0` declares `typescript >=4.8.4 <6.1.0`, so TS 7 is out of range.

### Docs

README prerequisites now mention macOS and Windows. I also removed the stale `System Status` command (deleted in November 2025) and rewrote the project structure section, which still described the pre-`src/lib` layout.

## Testing

`npm run lint`, `npm run build` and `tsc --noEmit` all pass, and `npm audit` reports 0 vulnerabilities.

**I have not been able to run this on Windows** — I only have a macOS machine. The Windows evidence in this PR is static: the deeplink was the sole platform-bound path, no macOS-only API remains, and the lockfile resolves the `@esbuild/win32-*` binaries so `npm ci` works there. If you would rather have a real Windows run before this is released, I am happy to wait for someone with a Windows machine to try `ray develop` against this branch.

## Checklist


0xdhrv (approved):
Looks good to me, approved ✅

PR #30367, opened 2026-08-20
feat(vibelet-search): cross-list sessions, cmd+K filter, incremental cache & content index

quanru:
## What

Performance and UX overhaul of the **vibelet-search** extension, borrowing the data-layer philosophy of a local cache + incremental rebuild:

### Performance
- **Incremental metadata cache** (`mtime + size` fingerprints per source). Re-opening the extension is near-instant after the first scan; only changed/new session files are re-read (their head, streamed, stopping as soon as the title is found).
- **Merged content index.** Full-text search now runs ripgrep over a single `messages.txt` (built incrementally from per-session segments) instead of every `.jsonl` file on every keystroke.
- **Message-level hits (seq contract).** A content hit resolves to the exact message; the detail view jumps straight to that message and highlights it, instead of scanning the file again.
- **Debounced content search** (150 ms) so fast typing doesn't spawn a ripgrep per keystroke.

### UX
- Sessions are now **cross-listed in a single timeline** (no agent grouping), with a per-row source label.
- New **`⌘K` filter dropdown**: All / Agent (Claude, Codex) / Source / Project.

### Code structure
- `parsers.ts` (719 lines) split into `adapters/`, `scanners/`, `index/`, and `load-messages.ts`, mirroring the adapter/scanner/db layering — adding a new agent is now mostly one new adapter + scanner.

### Fix
- Streamed JSONL head reading could throw `ERR_STREAM_PREMATURE_CLOSE` when stopping early, which wiped session titles; now closes the stream cleanly and keeps parsed lines.

## Verification

- `npm run lint` / `npm run build` pass.
- 104 unit tests pass (vitest), all against synthetic fixtures — including a **seq contract** test asserting content-index line numbers equal `loadSessionMessages` indices.
- On real data (152 sessions): first scan ~50 ms, incremental re-open ~5 ms, content search ~7 ms after the one-time index build (~1.4 s).


0xdhrv (approved):
Looks good to me, approved ✅

Issue #21857, opened 2025-09-29
[Apple Reminders] Subtask support


isaquepereira:
### Extension

https://www.raycast.com/raycast/apple-reminders

### Description

Would be nice having the ability to:
- View subtasks while showing all or filtered reminders;
- Create subtasks while creating a new reminder;
- Asking AI to indent or outdent tasks.

### Who will benefit from this feature?

All users which has a more granulated approach on tasks

blancpain:
Opened #30369, which implements the viewing half of this: subtasks now display nested under their parent reminder in the My Reminders command. Creating subtasks unfortunately isn't possible via any public API (EventKit, AppleScript, and Shortcuts all lack a parent/subtask concept) — details in the PR description.

PR #30369, opened 2026-08-20
[Apple Reminders] Display subtasks nested under their parent reminder

blancpain:
## Description

Subtasks are currently displayed as regular, flat reminders in the "My Reminders" command. This PR displays them nested under their parent reminder, similar to the Reminders app (related to #21857).

**How it works:** Apple's EventKit framework doesn't expose parent/subtask relationships, so the parent–child map is read (read-only) from the local Reminders SQLite stores in `~/Library/Group Containers/group.com.apple.reminders/Container_v1/Stores/`, joined to EventKit reminders via `calendarItemIdentifier` = `ZCKIDENTIFIER`, using `executeSQL` from `@raycast/utils`. Within each section, subtasks are moved directly beneath their parent and rendered with an indented circle glyph. A subtask whose parent isn't visible in the same section stays where it is, so it never disappears.

**Graceful fallback:** reading the store requires the user to grant Raycast Full Disk Access. If the store can't be read (no FDA, or Apple changes the private schema), the extension silently falls back to the existing flat list — no errors, no behavior change. The README documents the FDA requirement.

**Scope note:** this is display-only. Creating subtasks isn't included because no public API allows it — EventKit, AppleScript, and Shortcuts all lack a parent/subtask concept, and writing to the CloudKit-synced SQLite store directly would be unsafe. Search/filtering is unaffected (keywords keep the raw title).

## Screencast

_Screenshot to be added in a comment below._

## Checklist

### DIFF
--- a/extensions/apple-reminders/CHANGELOG.md
+++ b/extensions/apple-reminders/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Apple Reminders Changelog
 
+## [Display subtasks nested under their parent reminder] - {PR_MERGE_DATE}
+
+- Subtasks are now displayed indented under their parent reminder in the "My Reminders" command. Since EventKit doesn't expose subtask relationships, they are read from the local Reminders database, which requires granting Raycast Full Disk Access. Without it, reminders are displayed as a flat list, as before.
+
 ## [Fix natural-language due dates] - 2026-08-19
 
 - Restore relative due-date parsing in Create Reminder for `1h`, `1 hour`, `3 hours`, `in 10 minutes`, `3:45 pm`, `3 days`, and `1 year`.
--- a/extensions/apple-reminders/README.md
+++ b/extensions/apple-reminders/README.md
@@ -90,6 +90,10 @@ The AI will set:
 
 You can disable this behavior by toggling on the "Don't use the AI" preference for the "Quick Add Reminder" command. The intelligent date and time parsing will still work, but not the other fields.
 
+## Subtasks
+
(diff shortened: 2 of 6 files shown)


blancpain:
Good catch on the dropped descendants — fixed in 85851c0. `nestSubtasks` now emits each top-level reminder's descendants depth-first, so deeper nesting levels are preserved, with a guard so malformed parent data (e.g. a cycle) can never make a reminder disappear from the list. Verified against parent/child, grandchild-chain, cycle, and orphaned-parent cases.

0xdhrv (commented):
Thanks for tackling subtasks, nesting them in the list would be useful.

However, we can’t rely on the local Reminders database for this. Its schema and storage are undocumented macOS implementation details, so the hierarchy can break without any change on our side. It also requires granting Raycast Full Disk Access solely to enrich a list view, which is not an acceptable dependency for this command.
The fallback to a flat list prevents a hard failure, but the feature would still behave inconsistently and cannot be supported reliably across macOS releases or Reminders sync states.

Since EventKit does not provide a public, supported API for reading reminder parent–child relationships, I think we should keep the command flat for now and avoid this implementation path.

cc: @pernielsentikaer

blancpain:
Thanks for taking the time to review this. The undocumented schema and the Full Disk Access requirement were the two things I was unsure about when opening this, so I understand not wanting to take on that dependency in an official extension.

Agreed that flat is the right call until Apple exposes the parent–child relationship through EventKit.  I'll leave this open for now in case there's more to discuss, but no hard feelings if you'd rather close it.

Thanks again 🙏

0xdhrv:
> Thanks for taking the time to review this. The undocumented schema and the Full Disk Access requirement were the two things I was unsure about when opening this, so I understand not wanting to take on that dependency in an official extension.
> 
> Agreed that flat is the right call until Apple exposes the parent–child relationship through EventKit. I'll leave this open for now in case there's more to discuss, but no hard feelings if you'd rather close it.
> 
> Thanks again 🙏

Yeah, thanks for understanding, I completely understand the concern, we can revisit this when Apple exposes the same. 

Closing this based on the above.

PR #30391, opened 2026-08-21
google-chrome-profiles: add profile deletion

imbriandev:
## Summary

- Add a destructive **Delete Profile** action to the profile list.
- Require confirmation and use a dedicated `⌘ ⇧ Backspace` shortcut.
- Delete the selected profile directory and update Chrome `Local State` atomically.
- Allow deleting inactive profiles while Chrome is running; block deletion when the selected profile is active.
- Prevent path traversal, deleting non-directories, and deleting the last profile.

## Testing

- `npm run lint`
- `npm run build`
- Tested deletion logic with a temporary two-profile fixture: the selected directory was removed, the sibling directory was preserved, and `Local State` was updated.

### DIFF
--- a/extensions/google-chrome-profiles/src/util/util.ts
+++ b/extensions/google-chrome-profiles/src/util/util.ts
@@ -69,19 +69,19 @@ export const deleteChromeProfile = async (profile: Profile, browser: BrowserConf
   try {
     profileStats = await lstat(profilePath);
   } catch (error) {
-    if (error instanceof Error && (error as NodeJS.ErrnoException).code === "ENOENT") {
-      throw new Error("Chrome profile directory not found");
-    }
-    throw error;
+    if (!(error instanceof Error) || (error as NodeJS.ErrnoException).code !== "ENOENT") throw error;
   }
-  if (!profileStats.isDirectory()) throw new Error("Chrome profile path is not a directory");
-  if (await isProfileOpen(profilePath)) throw new Error(`Close the ${profile.name} profile before deleting it`);
+  if (profileStats && !profileStats.isDirectory()) throw new Error("Chrome profile path is not a directory");
 
-  const deletedProfilePath = join(dataDirectory, `.raycast-delete-${randomUUID()}`);
-  await rename(profilePath, deletedProfilePath);
+  let deletedProfilePath: string | undefined;
+  if (profileStats) {
+    if (await isProfileOpen(profilePath)) throw new Error(`Close the ${profile.name} profile before deleting it`);
+    deletedProfilePath = join(dataDirectory, `.raycast-delete-${randomUUID()}`);
+    await rename(profilePath, deletedProfilePath);
+  }
 
   try {
-    if (await isProfileOpen(deletedProfilePath)) {
+    if (deletedProfilePath && (await isProfileOpen(deletedProfilePath))) {
       throw new Error(`Close the ${profile.name} profile before deleting it`);
     }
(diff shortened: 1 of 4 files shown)


extensions/google-chrome-profiles/src/util/util.ts
```diff
@@ -1,11 +1,28 @@
...
+import { spawn, execFile } from "child_process";
+import { promisify } from "util";
 import { randomUUID } from "crypto";
 import { showToast, Toast } from "@raycast/api";
-import { BrowserConfig } from "./types";
+import { BrowserConfig, GoogleChromeLocalState, Profile } from "./types";
+
+const execFileAsync = promisify(execFile);
+
+const profileIsOpen = async (profilePath: string) => {
```
frouo:
```suggestion
const isProfileOpen = async (profilePath: string) => {
```

extensions/google-chrome-profiles/src/util/util.ts
```diff
@@ -20,6 +37,68 @@ export const ChromeAction = {
   openUrl: (url: string): ChromeTarget => ({ action: "openUrl", url }),
 };
 
+export const deleteChromeProfile = async (profile: Profile, browser: BrowserConfig) => {
+  const dataDirectory = resolve(homedir(), browser.dataPath);
+  const profilePath = resolve(dataDirectory, profile.directory);
+  if (dirname(profilePath) !== dataDirectory) throw new Error("Invalid Chrome profile directory");
+
+  const localStatePath = join(dataDirectory, "Local State");
+  const localState = JSON.parse(await readFile(localStatePath, "utf8")) as GoogleChromeLocalState;
```
frouo: We already have these kind of code in src/index.tsx:

```
        const path = join(homedir(), browser.dataPath, "Local State");
        const localStateFileBuffer = await readFile(path);
        const localStateFileText = localStateFileBuffer.toString("utf-8");
```

Is it possible to DRY it in a utility function?

frouo (changes requested):

imbriandev:
@frouo I addressed the requested changes:

- Renamed `profileIsOpen` to `isProfileOpen`.
- Centralized Chrome `Local State` reading.
- Added rollback for metadata and staged-directory cleanup failures.

All CI checks are passing. Could you please re-review when convenient?

frouo (commented):
Ok thank you.

Did you consider `greptile-apps` review feedback if relevant?

imbriandev:
Yes, I considered the relevant Greptile feedback and addressed it:

- Replaced the stale last_active_profiles check with lsof checks before and after staging the profile directory.
- Stage deletion via atomic rename, update Local State atomically, and roll back both the directory and metadata if either metadata writing or final cleanup fails.
- Changed the changelog date to {PR_MERGE_DATE}.
- Also centralized Local State reading as requested.

pernielsentikaer:
Hi @imbriandev — thanks for the careful delete path.

@frouo you still have changes requested here. The DRY / `isProfileOpen` bits and the Greptile notes look addressed in the latest revision. OK to clear that review?

One leftover from Greptile I agree with: if the final `rm` fails partway, rolling the half-deleted directory back into the live profile path can resurrect a broken profile. Safer to leave the `.raycast-delete-*` folder and tell the user.

frouo (approved):
Look good to me. Yet, I haven't the time to test it on device I am sorry.

pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #30393, opened 2026-08-21
[Agent Usage] Add AIHubMix usage provider

jalenzz:
## Summary
- Add AIHubMix balance to the Agent Usage list and menu bar via the official get-self API
- Auth uses the Access Key from extension preferences or `AIHUBMIX_ACCESS_KEY` (preference wins)
- Remaining/used USD is `quota` / `used_quota` divided by 500000

## Test plan
- [x] Enable Show AIHubMix, paste an Access Key from https://console.aihubmix.com/setting, confirm remaining/used match the console
- [x] Clear the preference and set `AIHUBMIX_ACCESS_KEY` in the login shell; confirm the list and menu bar still load
- [x] With no key configured, confirm the not-configured message
- [x] Confirm the menu bar row shows the blue badge with a light glyph (not a punched-out hole)


0xdhrv (commented):
Hi @jalenzz 👋

Thanks for the contribution 🔥

---

jalenzz:
Hi @0xdhrv, thanks for taking a look.

I went through the Greptile notes:

Manual preference types — this extension already uses SharedPrefs / AgentVisibilityPreferences for every provider (DeepSeek, MiniMax, etc.). Switching to the generated Preferences type would be a repo-wide change, so I left it as-is in this PR.

Credential resolution twice — same pattern as DeepSeek: resolveAuthKey and the fetcher both call the resolver. Cache hits only run resolveAuthKey. A cache miss when the key exists only in the login shell can spawn two shells (up to 3s each). Happy to tighten that here if you want it in this PR; otherwise I would keep it consistent with the other providers.

Please let me know which you prefer, then I will mark the PR ready for review.

0xdhrv (approved):
Looks good to me, approved ✅

PR #30394, opened 2026-08-21
feat: add MinimaxCN provider

skyblue:
Adds the MinimaxCN (MiniMax 国内版) provider alongside the existing MiniMax one, hosted at api.minimaxi.com.

- New `src/minimaxcn/` module mirrors the MiniMax structure (fetcher, parser, auth, renderer, types)
- Token resolution prefers `MINIMAX_CN_API_KEY` with fallback to `MINIMAX_API_KEY` for users sharing a key across providers
- Visible in both the list view and menu bar, gated by a new "Show MinimaxCN" preference
- Reads coding-plan model remains: 5h interval + weekly window, with `current_interval_remaining_percent` fallback when counts are 0


skyblue:
Fixed all 3 Greptile review findings (P1×2, P2×1):

1. **`src/minimaxcn/fetcher.ts`** — Switched endpoint to `api.minimaxi.com` per the provider's documented host.
2. **`src/minimaxcn/auth.ts`** — Reversed credential precedence so the documented preferred key (`MINIMAX_CN_API_KEY`) wins over the OpenCode fallback.
3. **`CHANGELOG.md`** — Replaced hard-coded date with the `{PR_MERGE_DATE}` placeholder so the merge step substitutes correctly.

All three fixes are squashed into the same commit (`10b20b63`). `npm run lint`, `npm run typecheck`, and `npm run build` all pass locally.

skyblue:
Pushed a follow-up: **removed the OpenCode credential fallback entirely** from `src/minimaxcn/auth.ts`.

Rationale: OpenCode's `minimax-coding-plan` provider id maps to the international endpoint (`api.minimax.io`), not the CN endpoint (`api.minimaxi.com`), so any token read from OpenCode would be invalid against the CN API. Even with env var taking precedence (the previous fix), keeping OpenCode in the fallback chain meant a user who only had an OpenCode token and no env var would still hit 401/403 against the CN endpoint — better to surface the "not configured" state than to silently send an international token to the CN API.

Updated:

- `src/minimaxcn/auth.ts` — dropped `readOpencodeAuthToken` call + import
- `README.md` — MinimaxCN row "OpenCode" column changed from ✓ to —
- `CHANGELOG.md` — added a bullet explaining the intentional omission, with a code comment in `auth.ts` pointing future readers at the same reason

Local `npm run lint`, `npm run typecheck`, `npm run build` all pass. New commit: `1ebcd2e7`.

skyblue:
Follow-up v3: simplified env-var resolution further by also dropping the `MINIMAX_API_KEY` fallback (in addition to the OpenCode removal in v2).

Rationale: `MINIMAX_API_KEY` targets the international endpoint (`api.minimax.io`) just like OpenCode's `minimax-coding-plan` token. The earlier "two env vars" fallback assumed users might share one key across both endpoints; in practice the international and CN accounts are independent services with separate credentials.

MinimaxCN now only reads:
1. the explicit preference token, then
2. `MINIMAX_CN_API_KEY` (env var or login shell)

Everything else (`MINIMAX_API_KEY`, OpenCode) is intentionally not read.

Updated:

- `src/minimaxcn/auth.ts` — replaced `MINIMAX_CN_ENV_KEYS` array with single `MINIMAX_CN_ENV_KEY` constant; `readEnvToken` simplified to direct read + single shell lookup
- `package.json` — preference description no longer mentions `MINIMAX_API_KEY`
- `README.md` — Supported Agents table env-var column simplified
- `CHANGELOG.md` — bullet rewritten to state `MINIMAX_API_KEY` is intentionally not read

Local `npm run lint`, `npm run typecheck`, `npm run build` all pass. New commit: `d2fa5513`.

extensions/agent-usage/src/agents/provider-hooks.ts
```diff
@@ -201,6 +205,30 @@ export const useMiniMaxUsage = createUsageHook<MiniMaxUsage, MiniMaxError>({
   },
 });
 
+// MinimaxCN — 国内版 MiniMax (api endpoint minimaxi.com vs minimax.io)
+// token 走 MINIMAX_CN_API_KEY env 或 minimaxcnApiToken preference
```
EDM115: please try to keep all comments in English, this will ease review for other contributors ❤️

extensions/agent-usage/CHANGELOG.md
```diff
@@ -1,5 +1,14 @@
 # Agent Usage Changelog
 
+## [Add MinimaxCN support] - {PR_MERGE_DATE}
+
+### New Features
+
+- Add MinimaxCN (MiniMax 国内版) provider to the main list and menu bar
+- Read `MINIMAX_CN_API_KEY` env var only; `MINIMAX_API_KEY` is intentionally not read because it targets the international endpoint and would 401 against the CN API
+- MinimaxCN intentionally does not read OpenCode credentials, since OpenCode's `minimax-coding-plan` provider id maps to the international endpoint and would yield 401/403 against the CN endpoint
```
EDM115: the changelog is mainly user-facing, and although I believe most Raycast users are developers too, internal decisions that happened during the PR review cycles shouldn't be  there

EDM115 (changes requested):

skyblue:
Both review threads addressed in this push:

**@EDM115 line comments**:
1. `src/agents/provider-hooks.ts:208` — Replaced the Chinese-language comment above `useMinimaxCNUsage` with English (and corrected the `api.minimaxi.com` vs `api.minimax.io` phrasing).
2. `CHANGELOG.md` — Removed the two internal-review-decision bullets (MINIMAX_API_KEY exclusion + OpenCode exclusion rationale). Kept only user-facing behavior: "Chinese-region MiniMax" + the coding-plan quota window shape.

**Greptile 07:02 P1+security ("International Credential Crosses CN Boundary")**: raised against v1 where `MINIMAX_CN_ENV_KEYS` still included `MINIMAX_API_KEY`. v3 (commit `d2fa5513`) removed that fallback entirely, and v4 doesn't reintroduce it — the credential is no longer sent to `api.minimaxi.com` because it is no longer read at all. No code change needed; flagging so Greptile can re-verify against the latest commit.

Local `npm run typecheck` and `npm run build` pass (lint skipped this round because the raycast.com API check timed out again — unrelated to this diff). New commit: `19fb5bae`.

@EDM115 @greptile-apps please re-review when convenient.

EDM115 (approved):

0xdhrv (approved):
Looks good to me, approved ✅

PR #30409, opened 2026-08-21
Add passgen extension

Chazzz287:
## Checklist


0xdhrv (commented):

- [Password Generator](https://www.raycast.com/joshuaiz/password-generator) by @joshuaiz

Both generate a random password from length and character-set options, then copy or paste it. **Generate Password** is the same job **Generate Random Password** already covers.

- **Generate Random Password** with length, case, numbers, and special characters
- **Copy Random Password** for a one-shot clipboard copy
- **Generate Memorable Password** for word-based formats

- `crypto.randomInt()` CSPRNG instead of weaker randomness
- Guaranteed coverage of each selected character set
- Concealed clipboard copy and a configurable symbol pool

Could we consider contributing the CSPRNG and guaranteed-charset behavior to Password Generator instead?

> This would help avoid duplication and keep password generation consolidated in one extension.  

PR #30415, opened 2026-08-21
[Espanso] Fix trigger search and overflowing labels

kud:
## Description

Two fixes to **Search Matches**.

### Matches with a label could not be found by their trigger

`MatchItem` built its title as `label ?? triggers.join(", ")`, so a match with a label showed the label *instead of* its triggers. Raycast's built-in list filter reads only `title`, `subtitle` and `keywords`, so the triggers of every labeled match fell out of the search index. Such a match was findable only when the search term happened to appear in the label prose.

Searching `email` against these three returned the first two and not the third, since they are the only ones whose label says "email":

```yaml
- trigger: ";;email.short"
  label: "Short email signature"
- trigger: ";;email.full"
  label: "Full email signature"
- trigger: ";;email.ooo"
  label: "Out of office reply"
```

Triggers now go into `keywords`, alongside the replacement text that was already there, so label and triggers are both searchable. Triggers are also split on non-alphanumeric characters, because keyword matching does not reliably reach a term sitting mid-string behind punctuation: `;;email.ooo` indexes as `[";;email.ooo", "email", "ooo"]`. Single-character fragments are dropped, so a one-letter namespace prefix does not become a keyword on every match.

### A long label overflowed the detail panel

The `Label` tag in the detail metadata neither wraps nor truncates, so a long label ran past the right edge of the panel. It is now cut to 50 characters with an ellipsis. Nothing is lost: a match with a label already shows it in full as the list item title.

Both changes have unit tests in `src/lib/utils.test.ts`, covering `buildSearchKeywords` and `truncate`. `npm run build`, `npm run lint`, `npm run typecheck` and `npm test` all pass.

## Screencast

Before this change, searching `email` returned two of the three matches above; `;;email.ooo` was missing. It now returns all three, and a long label sits inside the detail panel instead of running off the edge.

## Checklist

### DIFF
--- a/extensions/espanso/CHANGELOG.md
+++ b/extensions/espanso/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Espanso Changelog
 
-## [Fix trigger search and overflowing labels] - {PR_MERGE_DATE}
+## [Fix trigger search and overflowing labels] - 2026-08-21
 
 ### Bug Fixes
 
--- a/extensions/espanso/CHANGELOG.md
+++ b/extensions/espanso/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Espanso Changelog
 
+## [Fix trigger search and overflowing labels] - {PR_MERGE_DATE}
+
+### Bug Fixes
+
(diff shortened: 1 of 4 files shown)


pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #30424, opened 2026-08-22
Add courtlistener-search extension

cro64:
## Description

Adds **CourtListener Search**, a new extension for searching US case law from Raycast.

[CourtListener](https://www.courtlistener.com) is Free Law Project's open database of court opinions. This extension puts it behind a single command — search, read enough to know whether a case is the one you want, then open, save, or copy its citation without opening a browser.

**One command, `Search Case Law`:**

- **Search opinions** by keyword. Results show the court, decision date and citation, and page as you scroll.
- **Look up a citation.** Type or paste `410 U.S. 113`, `125 F.4th 23`, or a whole `Roe v. Wade, 410 U.S. 113 (1973)` and the top row offers a citation lookup instead of a text search — one case rather than several thousand text matches. This uses CourtListener's `/citation-lookup/` endpoint, which is throttled in its own scope, so checking a citation never spends a search.
- **Search by meaning** (`⌘⇧M`) via CourtListener's semantic engine. Useful for a question — *can police search a car after arresting the driver* returns vehicle-search cases, where a keyword search returns cases named *State v. Police* and *Turner v. Driver*. The option is hidden for citations and for advanced query syntax, which the semantic engine embeds as prose rather than parsing.
- **Filter** by court (Supreme Court, any federal circuit, state courts of last resort) and by how far back to look. Filters reset when the command closes.
- **Save cases** and keep a searchable history of recent queries, both in LocalStorage.

**Rate limits.** CourtListener's free tier is tight, so the extension is careful: searches run on `↵` rather than as you type, results are cached for 30 minutes (`⌘R` forces a fresh one), opening the command fires no request, and failed requests are never cached. When a throttle is hit, it reads the token's real limits from `/api-usage/` and counts down rather than showing an empty list.

**Setup.** Requires a free CourtListener API token — register at [courtlistener.com](https://www.courtlistener.com/register/), copy it from [the API token page](https://www.courtlistener.com/profile/api-token/), and paste it into the extension's preferences on first run. No payment or approval step. The README covers this.

## Screencast

Searching case law, with the detail pane showing the matching passage and case metadata:

Searching case law

Pasting a citation looks the case up directly rather than searching for the text:

Citation lookup

## Checklist


pernielsentikaer (approved):
Looks good to me, approved 🔥

Issue #30239, opened 2026-08-16
[Sonarr] Windows Support


gabriel-vanca:
### Extension

https://www.raycast.com/Aayush9029/sonarr

### Description

Any chance support for this extension will come to Raycast for Windows too?

@Aayush9029

JohanDevl:
@raycastbot close this issue

PR #30425, opened 2026-08-22
Update sonarr extension

JohanDevl:
## Description

Adds Windows support to the Sonarr extension, requested in #30239, and updates the dependencies.

### Windows support

- `platforms` now declares both `macOS` and `Windows`.
- The platform-bound surface here is the **keyboard shortcuts**: 19 call sites declared a shortcut with a `cmd`-only modifier, which Raycast ignores entirely on Windows. Those actions would have been reachable only through the action panel.

They are now declared for both platforms:

- 9 sites move to `Keyboard.Shortcut.Common` (`Refresh`, `Open`). I only used `Common` where its macOS binding is **identical** to the one it replaces (⌘R, ⌘O), so no macOS shortcut changes in this PR. I deliberately did *not* use `Common.Remove` or `Common.Duplicate`, whose macOS bindings (⌃X, ⌘D) would have silently rebound existing actions.
- The remaining 8 use the nested `{ macOS, Windows }` syntax, collected in one `src/lib/utils/shortcuts.ts` map so the bindings are reviewable in a single place rather than scattered across 8 command files.

One binding is not a literal translation, and it is deliberate: **Remove from Queue** is ⌘⌫ on macOS but `ctrl`+`D` on Windows, because `ctrl`+`Backspace` deletes the previous word in the always-focused search bar there. `ctrl`+`D` / `ctrl`+`shift`+`D` is what Raycast itself ships for `Common.Remove` / `RemoveAll` on Windows.

Everything else was already portable, and I checked rather than assumed: no AppleScript or shell exec, no `process.platform`, no local filesystem paths, and no `raycast://` deeplinks — this extension already opened preferences through `openExtensionPreferences()`. The only external-launch actions are `Action.OpenInBrowser` with `http(s)` URLs. The `Copy Path` action copies a path returned by the Sonarr *server*, so its separators are intentionally not the client's concern.

### Dependencies

- `@raycast/api` → v2. This also pulls in an esbuild release that clears GHSA-g7r4-m6w7-qqqr, which is a Windows-specific advisory, so it is relevant to this PR rather than incidental.
- Updated `@raycast/utils`, `date-fns`, `zod`, `@raycast/eslint-config`, `eslint`, `prettier`.
- `@types/node` and `@types/react` are pinned to the exact versions `@raycast/api` itself ships, so the type surface matches the runtime instead of drifting ahead of it.
- Overrides: `brace-expansion` is now scoped per major. The previous unscoped `"brace-expansion": "5.0.9"` forced `filelist`, which declares `^2.0.1`, onto a 5.x major. Natural resolution already lands on patched releases in both ranges. The `js-yaml@4` override is dropped — `js-yaml` is no longer anywhere in the tree. `npm audit` is clean.

`typescript` deliberately stays on 5.9.x: `@raycast/eslint-config@2.2.0` declares `typescript >=4.8.4 <6.1.0`, so TS 7 is out of range.

### Fixes

While preparing this I found the instance URL was built twice, by two implementations that disagreed. `getApiConfig()` learned to accept a `Host` written as a full URL or `host:port`; `getSonarrUrl()`, which every "Open in Sonarr" action uses, still did naive concatenation. With `Host` = `http://192.168.2.141:8989` the API worked but every browser link across all 7 view commands opened `http://http://192.168.2.141:8989:8989/...`.

Both now resolve through a single `getSonarrBaseUrl()` in `src/lib/utils/connection.ts`, so they cannot diverge again. The same change fixes a second case: a `Host` written as an explicit URL with no port (`https://sonarr.example.com`, the common reverse-proxy setup) had the `Port` preference appended to it, producing a refused connection. An explicit URL now means the scheme default.

The `Host` / `Port` preference descriptions are reworded to describe one value each, steering new setups toward a bare hostname with the port in its own field. Full-URL and `host:port` values are still normalized, so existing configurations keep working.

One deliberate behaviour change: a `Host` of `http://host` (explicit URL, no port) that relied on the `Port` field to supply `8989` now resolves to the scheme default instead. That combination only worked by accident, while the reverse-proxy case it unblocks is common and currently broken. It is called out in the changelog.

## Testing

`npm run lint`, `npm run build`, `tsc --noEmit` and `npm run format-check` all pass, and `npm audit` reports 0 vulnerabilities.

I exercised the URL resolution against 10 real-world configurations (bare host + port, `host:port`, full URL with and without port, HTTPS reverse proxy, URL base as a preference and as a path suffix, trailing slash, non-default port); all resolve as expected.

**I have not been able to run this on Windows** — I only have a macOS machine. The Windows evidence in this PR is static: the `cmd`-only shortcuts were the sole platform-bound surface, no macOS-only API remains, and the lockfile resolves the `@esbuild/win32-*` binaries so `npm ci` works there. The macOS side is behaviour-preserving by construction, since every macOS binding is byte-identical to the one it replaces. If you would rather have a real Windows run before this is released, I am happy to wait for someone with a Windows machine to try `ray develop`.

## Checklist


0xdhrv (approved):
Looks good to me, approved ✅

PR #30430, opened 2026-08-22
Add Search Session Output command to tmux-sessioner

mattiacolombomc:
## Description

Adds a **Search Session Output** command to tmux-sessioner: full-text search across the scrollback of every pane in every tmux session — both the commands you ran and the output they printed. This complements the existing session/window management commands: instead of remembering which session held that error or command, you search for it and jump straight there.

## How it works

- On open, the command captures the scrollback of every pane (`tmux capture-pane -p -J -S -` across `tmux list-panes -a`) once, then filters in memory as you type (from 2 characters). A `⌘R` action re-captures.
- Results are grouped by session, one entry per distinct matching line (newest occurrence first, repeated prompts/commands collapsed with the real count in the section subtitle). The detail pane shows the surrounding lines with the match marked.
- `⏎` switches to the matching session and window; `⌘T` / `⇧⌘T` open it in a new terminal tab/window (reusing the existing per-terminal launchers); `⇧⌘C` copies the line.
- Searching never requires a configured terminal — only the switch/open actions do, and they prompt for terminal setup on first use.

## Notes for reviewers

- macOS-only (unchanged); no new dependencies or preferences, one new `view` command
- Pane metadata is parsed with a space delimiter (pane id and window index are space-free leading fields, session name last) rather than a control-character separator, because Raycast's `execFile` stdout strips control bytes — a plain space is the only reliable delimiter for this data
- Capture is resilient: a pane closing between listing and capture is skipped rather than failing the whole search
- Scrollback depth is bounded by tmux's `history-limit`; the README notes raising it to search further back
- Tested end-to-end in Raycast: search across ~100 panes, session-grouped results with context, switch/open actions, and the empty/no-server states

## Checklist

### DIFF
--- a/extensions/tmux-sessioner/CHANGELOG.md
+++ b/extensions/tmux-sessioner/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Tmux Sessioner Changelog
 
+## [Search Session Output] - {PR_MERGE_DATE}
+
+### Added
+
+- **Search Session Output** command: full-text search across the scrollback of every pane in every tmux session — commands you ran and the output they printed. Results are grouped by session (one entry per distinct line, newest first) with the surrounding lines shown in the detail pane; `⏎` switches to the matching session and window, `⌘T` / `⇧⌘T` open it in a new terminal tab/window, `⇧⌘C` copies the line, `⌘R` reloads the scrollback.
+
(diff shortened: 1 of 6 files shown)


0xdhrv (approved):
Looks good to me, approved ✅

PR #30440, opened 2026-08-22
Link Commands: surface routers for services with a desktop app

kud:
## Description

Two changes to **Link Commands**, both about services that exist as both a native app and a website.

### Surface routers

A service like Notion, Todoist or WhatsApp has a desktop app *and* a web version. Until now the only way to express that was a command per surface, and the split falls on the wrong axis: which one you want is a fact about the machine you are sitting at, not about what you typed. A command that opens `Notion.app` is simply broken on a machine that has no `Notion.app`.

**Create Link Command** gains a **Desktop App** field. Pick the native app for a target and the generated command becomes a surface router:

```zsh
url="https://www.notion.so/"
app="/Applications/Notion.app"

if [[ -d $app ]]; then
  open "$app"
else
  open "$url"
fi
```

The app where the app is installed, the URL where it is not — so the same command, synced across machines, does the right thing on each without being edited.

The path comes from `getApplications()` rather than being assembled from the name, and the script opens that path directly. A basename is not an identity: two bundles can share one, and a renamed or unregistered app resolves to something else or to nothing, so `open -a` could miss the very bundle the presence check just confirmed.

`Open With` still composes — set both, and the web branch opens in your chosen browser while the app branch stays native.

The field appears only for `http(s)` targets with no `{query}` in them. `open -a` takes no query, so a search target has nothing an app could stand in for, and a folder has no web surface to fall back to. `create-link-command.tsx` hides the field on the same condition the generator enforces, so the form never offers a router it would silently discard.

> [!NOTE]
> **No dropdown, deliberately.** An earlier revision of this PR put an optional `App`/`Web` argument on the router so either surface could be forced. That turned out to cost the command its hotkey: Raycast has to raise the launcher to render an argument field, and `optional` only permits an empty answer — it does not skip the prompt. A router's whole value is firing without ceremony, so forcing a surface belongs on a second, argument-free command rather than on this one.

### Dropdown arguments in the detail pane

**Search Link Commands** rendered every argument as its placeholder. For a text argument that is right — it is what Raycast prompts you with. For a dropdown it is not: a placeholder names the axis but says nothing about what you can pick along it, and the choices were already parsed and sitting unused on the model.

The detail pane now lists a dropdown's choices instead. Text and password arguments fall back to the placeholder exactly as before. This half is independent of the router above and applies to any dropdown argument, including hand-written ones.

## Screencast

Not yet attached — see the checklist note below.

## Checklist

**Draft, deliberately.** `npm run build`, `ray lint` and `tsc --noEmit` all pass, and the generator was exercised directly across the router case, the router-plus-`Open With` case, a plain link, and both guard cases. What has *not* happened is a click-through of the distribution build in Raycast, so that box stays unticked rather than being ticked ahead of the fact. The screencast follows once it has.

### DIFF
--- a/extensions/link-commands/CHANGELOG.md
+++ b/extensions/link-commands/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Link Commands Changelog
 
-## [Surface Routers] - {PR_MERGE_DATE}
+## [Surface Routers] - 2026-08-30
 
 ### Added
 
--- a/extensions/link-commands/CHANGELOG.md
+++ b/extensions/link-commands/CHANGELOG.md
@@ -1,5 +1,15 @@
 # Link Commands Changelog
 
+## [Surface Routers] - {PR_MERGE_DATE}
+
+### Added
+
+- **Desktop App** on the create form — pick the native app for a service and the command becomes a surface router: it opens the app where the app is installed, falls back to the URL where it is not, and gains an optional App/Web dropdown so either surface can be forced. One row per service instead of a launcher and a separate web pin competing for the same query. Offered only for URL targets without a `{query}`, since a search command has already spent its first argument.
+
+### Changed
+
+- The detail pane lists a dropdown argument's choices rather than its placeholder. A dropdown's choices are what it actually tells you — "Surface" names the axis but not what you can pick along it.
+
 ## [Initial Version] - 2026-08-21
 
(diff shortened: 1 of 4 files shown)


0xdhrv (approved):
Looks good to me, approved ✅

PR #30441, opened 2026-08-22
Add apexconnect extension

AdvaitT17:
## Description

Apex Connect brings ApexOS smart-home control into Raycast. Search, view, and control every entity — lights, covers, climate, media players, cameras, switches, automations, and more — plus menu bar widgets for at-a-glance status (lights, covers, media players, batteries, weather, notifications). State updates are pushed live over a WebSocket connection to your ApexOS instance, and the extension automatically switches between an external (cloud) and internal (home network) URL based on WiFi SSID or reachability.

### Screenshots

| | |
|---|---|
| All Entities | Lights |
| Mediaplayers | Covers |

## Checklist


0xdhrv (commented):

[Home Assistant](https://www.raycast.com/tonka3000/homeassistant) already covers searching and controlling smart-home entities (lights, covers, climate, media players, cameras, automations, Assist, and the menu-bar widgets) against an instance URL and a long-lived access token.

This PR is a rebranded copy of that workflow for ApexOS. The client in `src/lib/haapi.ts` still calls Home Assistant’s `/api/states`, `/api/services/{domain}/{service}`, `camera_proxy`, and `mobile_app/registrations` endpoints, with the same command IDs and preferences (`instance`, `instanceInternal`, `homeSSIDs`, `token`). Pointing Home Assistant at your Apex Connect URL already gives ApexOS users that control. The `apexconnect://` companion deeplink would fit as an improvement there.

Could you contribute any Apex-specific bits (companion scheme, default cloud URL) to Home Assistant instead? That follows our [extension guidelines](https://manual.raycast.com/extensions-guidelines#reasons-to-be-rejected) and brings those improvements to existing users.

---

AdvaitT17:
@0xdhrv Fair point on the API, no argument there.

But I don't think that means it should be the same extension. ApexOS runs on HA's core, sure, but it ships as its own thing: own OS, own release channels, own companion apps, own cloud. Most of our users have no idea HA is under the hood and don't need to. They're going to search "apex" in the store, not "home assistant," so folding this into the existing extension just means they won't find it.

Also not sure upstreaming the ApexOS bits is the right call either. A deep link scheme and default URL that only apply to one downstream platform would just be dead weight for everyone else using that extension.

PR #30463, opened 2026-08-23
Fix Notion Create Database Page not saving properties

### LINKED ISSUE #30460: [Notion] Create Database Page does not set properties after data-source migration
### Extension

https://www.raycast.com/notion/notion

### Raycast Version

latest store build (post #25443)

### OS Version

macOS

### Description

`Create Database Page` creates the page, but the properties filled in the form are not written to Notion (except sometimes the title).

This is not covered by an existing open issue. #23460 is a Quick Capture feature request. #25545 / #25393 were create-page `object_not_found` failures; the “fix” in #25443 is what introduced this regression.

There are two bugs in `createDatabasePage()` / `formValueToPropertyValue()`.

#### 1. Page is created against the parent database, not the data source

After the Notion API `2025-09-03` split, databases are containers and rows live on a data source. The extension already lists **data sources** and loads the schema from the data source:

```ts
// fetchDatabases()
filter: { property: "object", value: "data_source" }

// fetchDatabaseProperties()
const dataSource = await notion.dataSources.retrieve({ data_source_id: dataSourceId });
```

On submit, #25443 added `resolveParentDatabaseId()` and creates the page with the **parent database id**:

```ts
const parentDatabaseId = await resolveParentDatabaseId(notion, database);
const arg = {
  parent: { database_id: parentDatabaseId },
  properties: {},
};
```

Current Notion docs require:

```ts
parent: { type: "data_source_id", data_source_id }
```

and the `properties` keys must match that data source’s schema. Creating against the database container can succeed while dropping the property payload.

`resolveDataSourceId()` already exists and is the correct direction. `resolveParentDatabaseId()` should not be used for page create.

#### 2. Property values are sent without the Notion type wrapper

`formValueToPropertyValue()` returns unwrapped values:

```ts
case "title":
case "rich_text": return markdownToRichText(value); // raw array
case "number":    return parseFloat(value);         // raw number
case "select":
case "status":    return { id: value };             // missing type key
case "date":      return { start: time };           // missing { date: ... }
```

The Create Page API expects:

```ts
{ title: [...] }
{ rich_text: [...] }
{ number: 42 }
{ select: { id } }
{ status: { id } }
{ date: { start } }
{ checkbox: true }
```

`ActionEditPageProperty` already wraps correctly (`{ select: { id } }`, `{ checkbox: bool }`). Create does not, so even with the right parent most properties are ignored. Title often still works because its id is literally `"title"` and the value is already a rich-text array.

#### 3. Falsy values are skipped

```ts
if (!propId || !value) return;
```

This drops checkbox `false` and number `0`.

### Steps To Reproduce

1. Open **Create Database Page**.
2. Pick a shared database / data source that has extra properties (select, status, date, number, checkbox, relation, etc.).
3. Fill the title **and** those properties.
4. Submit **Create Page**.
5. Open the new page in Notion.

### Current Behaviour

The page is created. Title may be set. Other properties stay empty / default. No error toast if Notion accepts the page and silently ignores the malformed / mismatched property payload.

### Expected Behaviour

Every filled form field is written to the new database page.

### Suggested fix

```ts
const dataSourceId = await resolveDataSourceId(notion, database);
const arg = {
  parent: { data_source_id: dataSourceId },
  properties: {},
};

// formValueToPropertyValue
case "title":     return { title: markdownToRichText(value) };
case "rich_text": return { rich_text: markdownToRichText(value) };
case "number":    return { number: parseFloat(value) };
case "select":    return { select: { id: value } };
case "status":    return { status: { id: value } };
case "date":      return { date: { start: time, ... } };
case "checkbox":  return { checkbox: value };
```

Also change the skip to `value == null || value === ""` so `false` / `0` survive.

Related: #25393, #25545, #25443

szupzj18:
## Description

Fixes #30460.

`Create Database Page` created the page but dropped the properties filled in the form. Three causes, all in `createDatabasePage()` / `formValueToPropertyValue()`:

1. **Wrong parent.** After Notion's `2025-09-03` database/data-source split, the extension lists data sources and loads the schema from the data source, but #25443 made page create resolve the *parent database* id and send `parent: { database_id }`. Creating against the database container can succeed while dropping the property payload. The page is now created with `parent: { data_source_id }` via the existing `resolveDataSourceId()` — the same direction every other database function in this file already uses. `resolveParentDatabaseId()` is removed (it only existed for this call and resolved the wrong way).

2. **Unwrapped property values.** `formValueToPropertyValue()` returned bare values (`42`, `true`, `{ id }`, a raw rich-text array) instead of the wrapped shapes the Create Page API expects (`{ number: 42 }`, `{ checkbox: true }`, `{ select: { id } }`, `{ title: [...] }`). The working `ActionEditPageProperty` path already wraps correctly. Every branch now returns the wrapped shape, including explicit `checkbox`/`url`/`email`/`phone_number` cases that previously fell through to a bare-value default.

3. **Falsy skip.** `if (!propId || !value) return;` dropped checkbox `false` and number `0`. It now skips only `null`/`undefined`/`""`.

## Verification

- `tsc --noEmit` clean on the changed files (the only remaining errors are pre-existing `Arguments`/`Preferences` global-namespace errors from running tsc outside the `ray` CLI, unrelated to this change)
- `eslint` clean on the changed files
- Shapes verified against the `@notionhq/client` v5.9.0 `CreatePageParameters` contract; `parent: { data_source_id }` is a supported parent type

I can't exercise the Raycast UI + Notion OAuth flow from here, so runtime verification would be appreciated by anyone with a Notion workspace handy.

## Checklist

- [x] I checked that [extension builds](https://developers.raycast.com/basics/debug-an-extension) without errors
- [x] I ran `npm run lint` and fixed all issues
- [x] I ran `npx tsc --noEmit` and fixed all issues
- [x] I've reviewed the [PR guidelines](https://github.com/raycast/extensions/blob/main/CONTRIBUTING.md#pull-request-basics)
- [x] I've followed the [pull request template](https://github.com/raycast/extensions/blob/main/.github/PULL_REQUEST_TEMPLATE.md)
- [x] I've updated the extension's [CHANGELOG.md](https://github.com/raycast/extensions/blob/main/.github/PULL_REQUEST_TEMPLATE.md#changelog) file

### DIFF
--- a/extensions/notion/CHANGELOG.md
+++ b/extensions/notion/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Notion Changelog
 
+## [Fix Create Database Page not saving properties] - 2026-08-24
+
+- Fix `Create Database Page` creating the page but dropping the filled-in properties after Notion's database/data-source split: the page is now created against the data source (`data_source_id`) instead of the parent database container ([#30460](https://github.com/raycast/extensions/issues/30460))
+- Fix property values being sent without their Notion type wrapper (e.g. `{ checkbox: true }`, `{ number: 42 }`, `{ select: { id } }`), which made the API silently ignore most of them
+- Fix checkbox `false` and number `0` being dropped by the form-value falsy check
+
 ## [Add Note Command] - 2026-08-19
 
 - Add a new `Add Note` command that appends a note to a page titled with the current date, nested inside a notes page (`NOTES` by default)
--- a/extensions/notion/CHANGELOG.md
+++ b/extensions/notion/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Notion Changelog
 
-## [Fix Create Database Page not saving properties] - {PR_MERGE_DATE}
+## [Fix Create Database Page not saving properties] - 2026-08-23
 
 - Fix `Create Database Page` creating the page but dropping the filled-in properties after Notion's database/data-source split: the page is now created against the data source (`data_source_id`) instead of the parent database container ([#30460](https://github.com/raycast/extensions/issues/30460))
 - Fix property values being sent without their Notion type wrapper (e.g. `{ checkbox: true }`, `{ number: 42 }`, `{ select: { id } }`), which made the API silently ignore most of them
--- a/extensions/notion/CHANGELOG.md
+++ b/extensions/notion/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Notion Changelog
 
-## [Fix Create Database Page not saving properties] - 2026-08-24
+## [Fix Create Database Page not saving properties] - {PR_MERGE_DATE}
 
 - Fix `Create Database Page` creating the page but dropping the filled-in properties after Notion's database/data-source split: the page is now created against the data source (`data_source_id`) instead of the parent database container ([#30460](https://github.com/raycast/extensions/issues/30460))
 - Fix property values being sent without their Notion type wrapper (e.g. `{ checkbox: true }`, `{ number: 42 }`, `{ select: { id } }`), which made the API silently ignore most of them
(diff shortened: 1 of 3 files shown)


pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #30464, opened 2026-08-23
fix(keepassxc): auto-lock database after inactivity period

### LINKED ISSUE #30458: [KeePassXC] Ask For Password after X minutes
### Extension

https://www.raycast.com/pabroux/keepassxc

### OS Version

Tahoe 26.5.2 (25F84)

### Description

Currently, after loggin in to my kdbx , never asks for my password again. should Ask after X time , (configurable)

thank you for the extension !

_No response_

### Anything else?

_No response_

gustavocaiano:
> **Note:** This PR is a suggestion and is fully open to changes. I'm not a maintainer of this extension — happy to adjust the approach, naming, scope, or split the work differently based on your feedback. The goal is to get #30458 resolved; the implementation below is one way to do it.

## Summary

Fixes #30458 — the `lockAfterInactivity` preference existed but never actually locked the database after the configured inactivity period.

### Root cause

`InactivityTimer.launchInactivityTimer()` ran a `setInterval` that wrote `Date.now()` to LocalStorage every 5 seconds forever (never cleared). Because the interval unconditionally wrote the current time, `lastActivity` was always ≈ now, so `hasRecentActivity()` always returned `true` and the database never locked. Two additional gaps: the staleness check ran only on mount (no in-session check while the view stayed open), and any watcher needed to start when the user unlocked via the form mid-session — not only on mount.

### Changes

**`src/utils/inactivity-timer.ts`** — rewritten:
- Activity is recorded only on genuine user interaction (unlock, search, selection), debounced to ≤1 write/sec via a module-level guard. No blind timer writes.
- `startLockWatcher(onLock)` runs a **read-only** interval (every 10s) that checks `now - lastActivity` and fires `onLock` when exceeded — it never writes `Date.now()`, so inactivity is detected correctly. Returns a cleanup function.
- `hasRecentActivity()` and the watcher share a single `readLastActivityMs()` helper that returns `null` for missing/empty/NaN timestamps, so both paths treat a corrupt/absent value as "locked" (closes a latent `NaN > threshold` footgun).
- `LOCK_AFTER_MS` is exported as the single source of truth for the threshold.
- Removed the broken `launchInactivityTimer()` entirely.

**`src/search.tsx`** — rewired:
- Mount effect loads the credential cache and calls `hasRecentActivity()`; recent → unlock, stale → delete cache + clear in-memory creds and stay locked. Cached-unlock also records fresh activity so reopening near the threshold edge doesn't surprise-lock.
- New `useEffect` keyed on `[isUnlocked]` starts the lock watcher when unlocked (covering both mount-unlock and form-unlock) and tears it down on lock/unmount.
- Uses the exported `LOCK_AFTER_MS` instead of a duplicated preference read.

**`src/utils/keepass-loader.ts`**:
- Added `clearCredentials()` to wipe the in-memory password/key file on lock (security hygiene so secrets don't linger while the unlock form is shown).
- `deleteCredentialsCache()` now also removes the `entries` cache (the full decrypted CSV of passwords/usernames/TOTPs), so secrets are not left readable in LocalStorage after auto-lock.

**`src/components/unlock-database.tsx`** — records activity on successful unlock so the timeout clock starts fresh.

**`src/components/search-database.tsx`** — records activity on search text change and list selection change.

**`CHANGELOG.md`** — added `[1.10.1]` entry.

### How it works across scenarios

| Scenario | Behavior |
|---|---|
| View open, user goes idle | Read-only watcher fires `onLock` once `now - lastActivity` exceeds the threshold → cache + in-memory creds cleared → unlock form shown. |
| Raycast closed, reopened after X min | On mount, `hasRecentActivity()` reads the persisted timestamp; stale → cache deleted, stays locked. |
| Raycast closed, reopened < X min | Recent → credentials restored, unlocked, fresh activity recorded. |
| User unlocks via form, then goes idle | `setIsUnlocked(true)` triggers the `[isUnlocked]` effect → watcher starts → locks after threshold. |
| Active user (typing/navigating) | `recordActivity` keeps `lastActivity` fresh → stays unlocked. |
| "Never" (`lockAfterInactivity = -1`) | All timer methods no-op; no watcher, never locks. |

### Validation

- `npx tsc --noEmit` → exit 0
- `npx eslint src --max-warnings=0` → exit 0
- `ray build -e dist` → exit 0

### Manual test matrix

Runtime auto-lock can't be automated here (requires a `.kdbx` database and `keepassxc-cli`), so please verify manually:

1. Set **Lock after inactivity = 2 minutes**, unlock, leave the Raycast window open and idle for >2 min → should re-lock to the unlock form.
2. Unlock, close the Raycast window, wait >2 min, reopen → should show the unlock form.
3. Unlock, close, reopen <2 min → should stay unlocked.
4. Set **Never** → should never lock and run no watcher.
5. Unlock, then type/arrow-navigate/copy within the window → should stay unlocked past 2 min while active.

### Notes

- The `entries` LocalStorage cache (decrypted CSV) is now cleared on lock. The tradeoff is that the first paint after re-unlock waits for `refreshEntriesCache` instead of using the cache — acceptable for a security feature.
- Lock can fire up to ~10s late due to the poll interval granularity.

Closes #30458.

### DIFF
--- a/extensions/keepassxc/CHANGELOG.md
+++ b/extensions/keepassxc/CHANGELOG.md
@@ -1,5 +1,13 @@
 # KeePassXC Extension Changelog
 
+## [1.10.1]
+
+### Fixed
+
+- Auto-lock now actually re-prompts for the database password after the configured inactivity period (#30458). Previously the inactivity timer wrote the current time to local storage every few seconds forever, so the database was always considered "active" and never locked. Activity is now recorded only on genuine interaction, and a read-only watcher locks the database when the period elapses — whether the view stays open or Raycast was closed and reopened.
+
+- Clear the cached database entries (decrypted passwords, usernames, and TOTP codes) from local storage when the database locks, so secrets are not left readable after auto-lock.
+
 ## [1.10.0] - 2026-03-30
 
 ### Added
--- a/extensions/keepassxc/src/components/search-database.tsx
+++ b/extensions/keepassxc/src/components/search-database.tsx
@@ -5,6 +5,7 @@ import { List } from "@raycast/api";
 import Entry from "./entry";
 import FolderFilterDropdown from "./folder-filter-dropdown";
 import { KeePassLoader, showToastKeepassxcCliErrors } from "../utils/keepass-loader";
+import { InactivityTimer } from "../utils/inactivity-timer";
 import { PinLoader } from "../utils/pin-loader";
 import { getEntryId, getFolders } from "../utils/entry-helper";
 
@@ -73,6 +74,8 @@ export default function SearchDatabase({
         folders.length > 0 ? <FolderFilterDropdown folders={folders} onFolderChange={setEntriesFolder} /> : undefined
       }
       throttle={true}
+      onSearchTextChange={() => void InactivityTimer.recordActivity()}
+      onSelectionChange={() => void InactivityTimer.recordActivity()}
     >
       {pinnedEntries.length > 0 && (
         <List.Section title="Favorites">
--- a/extensions/keepassxc/src/components/unlock-database.tsx
+++ b/extensions/keepassxc/src/components/unlock-database.tsx
@@ -3,6 +3,7 @@ import { JSX } from "react/jsx-runtime";
 import { useForm } from "@raycast/utils";
 
 import { KeePassLoader, showToastKeepassxcCliErrors } from "../utils/keepass-loader";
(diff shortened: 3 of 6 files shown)


pabroux:
Already implemented since January 2025 and still working. Not approved.

0xdhrv:
Closing as already fixed

Reference:
- https://github.com/raycast/extensions/issues/30458#issuecomment-5388515600

Issue #30447, opened 2026-08-23
[Google Translate]... Response Fetching error


Migueaali12:
### Extension

https://raycast.com/gebeto/translate

### Raycast Version

1.104.25

### OS Version

26.5.1

#### Error:
```
SyntaxError: Unexpected token '<', "<html><hea"... is not valid JSON
    at UT (/Users/mikel/.config/raycast/extensions/0e43920f-bbf6-4582-bc3f-5f63092c915f/quick-translate.js:75:11606)
    at process.processTicksAndRejections (node:internal/process/task_queues:103:5)
    at async Promise.all (index 1)
    at async gb (/Users/mikel/.config/raycast/extensions/0e43920f-bbf6-4582-bc3f-5f63092c915f/quick-translate.js:75:11709)
    at async /Users/mikel/.config/raycast/extensions/0e43920f-bbf6-4582-bc3f-5f63092c915f/quick-translate.js:82:1063
```

### Steps To Reproduce

Try to use translate tool

### Current Behaviour

it shows a syntax error

### Expected Behaviour

Show the translated text

gebeto:
Looks like it is working now according to this Fix: https://github.com/raycast/extensions/pull/30466

Issue #30371, opened 2026-08-20
[Google Translate]...


Tobiarz90:
### Extension

https://raycast.com/gebeto/translate

### Raycast Version

1.104.25

#### Error:
```
SyntaxError: Unexpected token '<', "<HTML><HEA"... is not valid JSON
    at UT (/Users/bartosztobiasz/.config/raycast/extensions/0e43920f-bbf6-4582-bc3f-5f63092c915f/quick-translate.js:75:11606)
    at process.processTicksAndRejections (node:internal/process/task_queues:103:5)
    at async Promise.all (index 1)
    at async gb (/Users/bartosztobiasz/.config/raycast/extensions/0e43920f-bbf6-4582-bc3f-5f63092c915f/quick-translate.js:75:11709)
    at async /Users/bartosztobiasz/.config/raycast/extensions/0e43920f-bbf6-4582-bc3f-5f63092c915f/quick-translate.js:82:1063
```

### Steps To Reproduce

1. Select text, e.g.: "Gemini API Usage
2. Open Raycast
3. Run Quick Translate command

PR #30466, opened 2026-08-23
Fix Google Translate extension failing with JSON parse error (429 HTML block page)

### LINKED ISSUE #30465: [Google Translate]...
### Extension

https://raycast.com/gebeto/translate

### Raycast Version

2

### OS Version

_No response_

#### Error:
```
SyntaxError: Unexpected token '<', "<html><hea"... is not valid JSON
    at UT (/Users/mts/.config/raycast/extensions/0e43920f-bbf6-4582-bc3f-5f63092c915f/quick-translate.js:75:11606)
    at process.processTicksAndRejections (node:internal/process/task_queues:103:5)
    at async Promise.all (index 0)
    at async gb (/Users/mts/.config/raycast/extensions/0e43920f-bbf6-4582-bc3f-5f63092c915f/quick-translate.js:75:11709)
    at async /Users/mts/.config/raycast/extensions/0e43920f-bbf6-4582-bc3f-5f63092c915f/quick-translate.js:82:1063
```

### Steps To Reproduce

see error

### Current Behaviour

_No response_

### Expected Behaviour

_No response_

szupzj18:
## Root cause

The vendored `@iamtraction/translate` library calls the Google Translate endpoint with `client=gtx`. Google now serves a 429 HTML block page to that client when the request originates from a non-browser TLS fingerprint (Node/undici), so `response.body.json()` throws `SyntaxError: Unexpected token '<', "<html><hea"... is not valid JSON` — the error reported in #30465, #30447, and #30371.

I confirmed this from the same machine at the same time:

| Client | Endpoint | Result |
|--------|----------|--------|
| curl (any UA) | `translate.google.com/.../single?client=gtx` | 200 JSON |
| undici / Node `https` (any UA) | `translate.google.com/.../single?client=gtx` | 429 HTML |
| undici | `translate.google.com/.../single?client=dict-chrome-ex` | 200 JSON |

The block is at the TLS-fingerprint layer, so header changes (User-Agent, Accept, etc.) do not resolve it.

## Change

Switch the API client identifier from `gtx` to `dict-chrome-ex` (the client Chrome itself uses for dictionary/translate lookups). The endpoint, query parameters, and response structure are identical — I diffed the full JSON responses for both clients and they match on every index the library parses (`body[0]` translation segments, `body[2]` / `body[8]` detected language, `body[7]` spell correction).

## Validation

- End-to-end: ran the actual vendored `translate()` function via `tsx` — `"hello"` (auto → es) returns `text: "Hola"`, `from.language.iso: "en"`, with the full rich `raw` body intact.
- Long-text POST path (>2048-char URL) also returns a correct translation.
- `ray lint` passes cleanly.

Fixes #30465.

### DIFF
--- a/extensions/google-translate/CHANGELOG.md
+++ b/extensions/google-translate/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Google Translate Changelog
 
+## [Fix translations failing with JSON parse error] - 2026-08-24
+
+- Switched the Google Translate API client identifier from `gtx` to `dict-chrome-ex` to avoid Google's 429 HTML block page for non-browser clients
+
 ## [Update] - 2026-08-10
 
 - Added a "Prioritize cross-language translations" preference to optionally move same-language results to the bottom
--- a/extensions/google-translate/CHANGELOG.md
+++ b/extensions/google-translate/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Google Translate Changelog
 
-## [Fix translations failing with JSON parse error] - {PR_MERGE_DATE}
+## [Fix translations failing with JSON parse error] - 2026-08-24
 
 - Switched the Google Translate API client identifier from `gtx` to `dict-chrome-ex` to avoid Google's 429 HTML block page for non-browser clients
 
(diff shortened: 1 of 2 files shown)


0xdhrv (approved):
Looks good to me, approved ✅

PR #30476, opened 2026-08-24
Add ejection-seat extension

chrismessina:
## Description

**macOS refuses to eject a disk and tells you "one or more programs may be using it" — without naming a single one. Ejection Seat names them.** 

ejection-seat

Pick a mounted volume and it shows every process holding a filesystem reference on it, ranked by how likely each one is the actual blocker, with an action to activate, quit, or resolve it.

- Find the processes and open files that may prevent a volume from ejecting
- Scan every mount under `/Volumes` up front, with a per-volume blocker count
- Group references into Likely Blockers, Other References, and System Services, ranked by how strongly each reference holds the volume
- Detail pane showing the owning application, PID, user, and every referenced path, collapsible when you only want the list
- Service-specific advice for Quick Look, Spotlight indexing, `fseventsd`, `revisiond`, and Time Machine
- Activate or politely quit the app holding a file, reveal a referenced path in Finder, or copy the paths
- Eject the volume with `diskutil eject` — never a forced unmount
- Surface the process named by Disk Arbitration when an eject is refused, with a shortcut to activate it
- State plainly that root-owned open files are invisible to an unprivileged scan, so an empty result is never read as "this volume is clear"

## Checklist


pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #30482, opened 2026-08-24
Update media-flow extension

egor-xyz:
## Description

Bugfix update for the **MediaFlow** extension.

Fixes a `Request frontend:main.extensionHost.renderEmpty timed out after 5000ms` crash in the **Now Playing** menu-bar command. When nothing was playing and the "Show when stopped" preference was disabled, the command returned `null` (empty render), but the live-refresh `media-control stream` subprocess (and its poll-timer fallback) kept the extension runtime alive past Raycast's 5s empty-render budget, throwing the timeout.

The live-update effect is now gated on whether a menu-bar item is actually rendered, so on the empty state the stream/timer tears down and the runtime goes idle.

Crash report: https://www.raycast.com/extension-issues/egorxyz/media-flow/7688456156

## Checklist

- [x] `npm run lint` passes
- [x] Tests pass (`npm test`, 101 passing)
- [x] Updated `CHANGELOG.md`

### DIFF
--- a/extensions/media-flow/.releaserc.json
+++ b/extensions/media-flow/.releaserc.json
@@ -1,0 +1,13 @@
+{
+  "branches": ["main"],
+  "tagFormat": "v${version}",
+  "plugins": [
+    ["@semantic-release/commit-analyzer", {
+      "preset": "conventionalcommits"
+    }],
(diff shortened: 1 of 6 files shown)


0xdhrv (approved):
Looks good to me, approved ✅

PR #30488, opened 2026-08-24
Add seo-audit extension

nurkamol:
## Description

Most SEO tools grade one URL. This crawls the whole sitemap and checks every
page, which is where the problems that cost traffic actually live.

I wrote it after three commercial graders all reported a client's site as
healthy while the language switcher on every translated article linked to a
404 — a bug none of them could see, because it was only wrong on pages they
never opened.

**Three commands, and the order matters.**

| | |
|---|---|
| **Preview Site** | How big is this, and is it the right one — ~1s, 3 requests, no page fetched |
| **Audit Site** | Crawl it and list what to change, worst first |
| **Recent Reports** | Runs the companion macOS app has already finished |

`Preview` is first on purpose. A full crawl takes minutes and a launcher is
built for the second you spend in it — running a seven-minute job behind a
keystroke is the obvious idea and the wrong one, because you close the window
and it is gone. So `Preview` answers the question you have *before* spending
those minutes: it settles the host, reads robots.txt and the sitemap, and
stops. You find out you pointed at the wrong host, or that the sitemap is
stale, in a second rather than in seven minutes.

`Audit` is capped by preference (25 pages by default). A thousand-page site
belongs in the terminal, and the empty state says so rather than leaving
somebody watching a spinner.

**Around ninety checks** — indexability, internal 404s and redirect chains,
metadata, JSON-LD that validates and still does nothing because it is not the
type Google reads for that page, Open Graph including whether `og:image`
actually loads, TLS and redirect behaviour, thin and near-duplicate content
found by fingerprint rather than by URL. Grouped into causes rather than listed
as complaints: one broken template on forty pages is one row saying forty.

**Nothing leaves the machine.** No account, no telemetry, no server in the
middle. The single exception is opt-in and named: the performance preference
asks Google's PageSpeed Insights about URLs you choose, because a `fetch` loop
cannot see rendering and a plausible wrong number is worse than no number.

**It reimplements nothing.** `import { preview } from "@nurkamol/seo-audit"` —
the same engine as the CLI, the GitHub Action, the Cloudflare Worker and the
macOS app. Levels, thresholds, grouping and ordering all arrive from the
package; this arranges them into rows. Published to npm specifically so this
folder could stand on its own here.

**Export** (`⌘E`) writes HTML, Markdown, CSV, JSON or a corrected sitemap to
`~/Downloads` using the engine's own writers. When the crawl did not see the
whole site the engine refuses to build a sitemap and the refusal is shown,
rather than a file that would delete real pages from somebody's index.

**Recent Reports** reads the same folder the macOS app writes, so a crawl run
there is here a second later with nothing synchronised. Read-only on purpose —
deleting somebody's seven minutes behind a single Return is not a good trade,
so that stays in the app where the confirmation lives.

One note for the reviewer: `ray lint` warns that the title should be
"Seo Audit". SEO is an acronym, and the Apple Style Guide that the guidelines
page cites keeps acronyms capitalised, so I have left it as **SEO Audit**.
Happy to change it if you would rather be consistent with the linter.

## Screencast

The five screenshots in `metadata/` are real captures of the running
extension — preview, results, preferences, the shared library, and an export.

## Checklist


nurkamol:
@0xdhrv if you have a moment for this one, I'd appreciate a look whenever it fits your queue.

It's a new extension. All checks are green, and the two things Greptile raised on the first pass are addressed — the hand-written preference type is now the manifest-generated `ExtensionPreferences`, and `Arguments.Audit` replaces the other copied shape. It's back to 5/5.

One thing worth flagging so it isn't a surprise: `ray lint` warns that the title should be "Seo Audit". I've kept **SEO Audit**, since SEO is an acronym and the Apple Style Guide the guidelines cite keeps acronyms capitalised — but happy to change it to match the linter if you'd rather be consistent.

No rush at all if the queue is long.

nurkamol:
Hi @0xdhrv — I hope you're doing well, and thank you for all the reviewing you do here.

I wanted to kindly ask whether this one might get a look when you have a moment. It's been open since 24 August and hasn't picked up an assignee yet, so I wondered whether it had quietly slipped past rather than waiting on anything from me.

All checks are green and the Greptile points from the first pass are addressed. Since submitting I've also updated the extension's engine dependency to the current release, so what's under review is up to date.

No rush at all — I'm happy to make any changes you'd like, and thank you either way.

pernielsentikaer (changes requested):
Hi @nurkamol 👋

Thanks for the contribution! We already have [Web Audit](https://www.raycast.com/flornkm/web-audit) by @flornkm, which covers the core SEO-checking workflow.

Your sitemap crawling, grouped findings, exports, and companion reports look like meaningful additions. We’d prefer bringing compatible improvements into the existing extension—would you be open to contributing there? If combining these workflows would be impractical, could you explain why? 🙂

---

nurkamol:
Hi @pernielsentikaer — thanks for looking, and for pointing me at Web Audit. I read through it properly before answering.

I think these are different shapes rather than overlapping ones. Web Audit analyses **one URL** — it fetches the page and its robots.txt and shows what that page says. That's genuinely useful, and it's instant.

seo-audit crawls the sitemap and checks every page, and most of what it reports doesn't exist for a single page: duplicate titles, descriptions and bodies *across* pages; pages that are linked but missing from the sitemap; orphans; click depth from the home page; internal links that 404 elsewhere on the site; hreflang sets that don't point back. Those need the whole crawl and its link graph in memory, not one document.

The other half is that the logic isn't in the extension. It imports `@nurkamol/seo-audit` from npm — the same engine the CLI, the GitHub Action, a macOS app and a Cloudflare Worker use, unmodified. The extension is ~1,400 lines of Raycast UI over a ~9,600-line engine. Folding those checks into Web Audit would mean reimplementing the engine inside an extension, and a report from Raycast would then differ from one from `seo-audit --json` — which is the thing I've been most careful to avoid.

It would also change Web Audit for the people using it today: it has no dependencies and costs about one request. This is minutes and hundreds of requests, with a kept-report library it shares with the macOS app.

That said — if there are per-page checks of mine that would fit Web Audit as it stands, I'm happy to open PRs there. That seems like the honest version of "compatible improvements", and I'd rather do that than argue for a second listing.

pernielsentikaer:
Hi @nurkamol 👋

Thanks for explaining the cross-page checks and shared-engine workflow. Those are meaningful differences from [Web Audit](https://www.raycast.com/flornkm/web-audit) by @flornkm.

Could an optional site-wide command there reuse your published engine without reimplementing it, while keeping the existing single-page command lightweight? Is there a concrete constraint that would make that impractical? 🙂

nurkamol:
Hi @pernielsentikaer — fair question, and you're right about the part I got wrong: it wouldn't mean reimplementing anything. A site-wide command in Web Audit could `import { audit } from "@nurkamol/seo-audit"` exactly as this extension does, and Web Audit's existing commands wouldn't need to change. There's no technical constraint that stops it.

The constraints are about who maintains it:

- **It moves with the engine.** In the 18 days since I opened this PR the extension has changed six times, mostly alongside engine releases: the score, Search Console, the subdomain inventory, a page ceiling after commands were killed at the 100 MB heap, re-runnable skipped checks, and Windows support. Inside Web Audit each of those becomes a PR against an extension @flornkm and @ezzylan own, so they'd be reviewing crawl changes they never asked for, every release.
- **The tests that keep it honest live with the engine.** 37 tests in the engine's repository read the extension folder directly: every declared preference is read, nothing imports outside the folder, no shortcut is macOS-only. They can't reach a folder inside another extension, so the drift they catch would come back.
- **Recent Reports belongs to my other front ends.** It reads the report library the macOS and Windows apps write, which makes sense beside them and would be odd in Web Audit.
- **Support would go to the wrong people.** Issues about the crawl would land with Web Audit's owners rather than with me.

So my preference is still a separate listing, with the maintenance sitting with whoever maintains the engine. But if you'd rather keep one SEO extension in the Store, I'll do it that way: if @flornkm is happy with it, I'll open a PR adding the site-wide commands to Web Audit, with their settings scoped to those commands so the existing ones stay exactly as they are. Just let me know which you'd prefer.

PR #30501, opened 2026-08-24
Update context7 extension

chrismessina:
## Description

Dependency-only follow-up to #30311. No behaviour changes, no new commands.

**`@raycast/api` 1.104.24 → 2.0.5.** Worth flagging for review: as far as I can tell this would be the **first 2.x extension in this repo**, and `developers.raycast.com/migration/v2` still 404s, so I verified the migration by diffing the shipped type definitions rather than following a guide. It is a soft major for this extension's surface — `engines` and `peerDependencies` are byte-identical to 1.104, all 17 `Keyboard.Shortcut.Common` members survive, and `canAccess`, `AI.ask`, `Tool.Confirmation`, `launchCommand`, `LocalStorage` and `environment.supportPath` are unchanged. The renames (`environment.commandName` → `entryPointName`, and friends) keep working as deprecated aliases and this extension does not use them. `tsc`, `ray build` and `ray lint` all pass with **zero source changes**.

If you would rather not carry the first 2.x extension in the monorepo, I am happy to drop that half and keep only the logger bump — say the word and I will push a revision.

**`@chrismessina/raycast-logger` 1.3.0 → 1.4.0.** Required for the above: 1.3.0 declared `peerDependencies: { "@raycast/api": "^1.0.0" }`, which is a hard `ERESOLVE` against 2.x, not a warning. 1.4.0 widens it to `^1.0.0 || ^2.0.0`.

1.4.0 also fixes a redaction bug I hit in #30311. The logger's encoded-secret heuristic counted `/` as a base64 character, so a library path like `/api/v1/tailwindlabs/tailwindcss` was printed as `***`. The workaround was to log the path as an array of segments; that is now reverted to the plain string, which is the only source change here.

### Changes

- `@raycast/api` `^1.104.24` → `^2.0.5`
- `@chrismessina/raycast-logger` `^1.3.0` → `^1.4.0`
- Reverted the path-segment logging workaround now that redaction is fixed upstream

## Checklist

### DIFF
--- a/extensions/context7/CHANGELOG.md
+++ b/extensions/context7/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Context7
 
-## [Raycast API 2.0] - {PR_MERGE_DATE}
+## [Raycast API 2.0] - 2026-08-25
 
 - Updated to Raycast API 2.0
 - Updated dependencies
--- a/extensions/context7/CHANGELOG.md
+++ b/extensions/context7/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Context7
 
+## [Raycast API 2.0] - {PR_MERGE_DATE}
+
+- Updated to Raycast API 2.0
+- Updated dependencies
+
 ## [My Libraries, My Snippets, and Wider Search] - 2026-08-18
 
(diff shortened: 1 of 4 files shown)


pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #30509, opened 2026-08-25
Update Image-wallet extension

diemonic1:
## Description

Windows port

1. Declared Windows support — `"platforms": ["macOS", "Windows"]` in the manifest; `@raycast/api` bumped to `^1.104.0`, required for the per-platform shortcut syntax.
2. Cross-platform shortcuts — every custom shortcut moved to the nested `{ macOS, Windows }` form (`⌘` → `Ctrl`), with Refresh and Quick Look switched to `Keyboard.Shortcut.Common`. Raycast maps `cmd` to the Windows key, so the Ctrl variants have to be declared explicitly.
3. Paths built with Node's path utilities — hardcoded POSIX separators removed; preview cache filenames are sanitised for backslashes and the drive colon (`C:`).
4. Image previews on Windows — file paths are converted to `file://` URLs via `pathToFileURL()`. Raycast does not resolve a raw `C:\...` path as an image source, which left the grid blank; this also percent-encodes spaces and non-Latin characters correctly.
5. Video previews on Windows — generated with `ffmpeg` (first frame → PNG) when it is on the `PATH`; otherwise the Card falls back to its file-type icon and a single warning is shown. The existing macOS JXA/AVFoundation path is untouched.

New features

6. PDF previews — the first page is rendered with PDFium compiled to WebAssembly (`assets/pdfium.wasm` + `@hyzyla/pdfium`, PNG encoding via `pngjs`). No native dependencies, so macOS and Windows behave identically with no external tooling. Cached in `supportPath` keyed by file path and `mtime`, so an edited PDF automatically gets a fresh preview. Rendering is asynchronous and progressive: the grid paints immediately and previews fill in as they finish, so a slow PDFium pass never blocks anything.
On size. The `pdfium.wasm` binary is 3.8 MiB (3,988,829 bytes) and ships in full, both into the repository and into the build — the extension bundle grows to roughly 3.6 MiB, up from a few tens of kilobytes. This is a deliberate trade-off: PDFium in WebAssembly is the only way to render PDFs identically on both platforms without a native module (Raycast does not build those) and without requiring the user to install external software. If the size turns out to be a blocker in review, this is the one piece that can be dropped — no other feature depends on it.
7. Card sorting — 10 modes: name, date added, date modified, file size, plus Recently Used and Most Used (usage counters recorded on paste and copy, stale paths pruned). The choice persists across launches and lives in a Sort Cards By submenu (`⌘⇧S` / `Ctrl+Shift+S`); the search bar dropdown stays the Pocket filter, since `Grid` allows only one `searchBarAccessory`.
8. `Thumbnail Layout` preference — three ways to lay out a thumbnail in its cell: `inset` (centred with padding, the previous look, default), `contain` (fills the cell with the whole image visible), `fill` (cropped edge to edge).
9. `Cards per Row` preference — column count from 3 to 8, defaulting to 5.
10. Card actions — Paste remains the first (default) card action; Copy to Clipboard is second. There is no `primaryAction` manifest preference.
11. Nested folders (recursive traversal) — scanning is now recursive: any folder at any depth that holds files becomes a Pocket, named by its path relative to the Wallet root (`memes/reactions`). Such Pockets are first-class: they appear in the Pocket filter, render as grid sections, and participate in sorting.
This neither breaks nor changes the existing folder display logic. The change is purely additive:
   * top-level folders remain Pockets, with the same names and in the same places;
   * files at the Wallet root still land in `Unsorted`;
   * no already-visible Card moved or was renamed.
The only difference is that the contents of nested folders, previously silently ignored (the old README explicitly marked them `NOT VISIBLE IN WALLET`), now show up as additional Pockets. So a user with nested folders will gain new Pockets, while nothing that already worked behaves differently. Symlink cycles are guarded with a set of resolved `realpath`s — without it, recursing through a linked directory could loop forever.
12. `.avif` support.
13. Configurable Card tooltip — the tooltip shown on hover is built from a user-chosen, reorderable list of fields (date created, date modified, file size, times used, pixel dimensions), always led by the Card's name; defaults to name only. Configured through a Configure Card Tooltip popup (`⌘⇧T` / `Ctrl+Shift+T`) next to Sort Cards By, with checkmark toggles and Move Up/Down actions, persisted across launches. Pixel dimensions are read from just the file's header (or, for videos, from the already-generated preview frame, which is captured at the source resolution) so the cost stays negligible even when the field is enabled.

Fixes

14. Error screens instead of silent failure — distinct states for "no directory selected", "directory not found" and "directory unreadable", each offering Change Wallet Directory and Retry. Wallet status and path are re-read on every fetch/revalidate rather than once at module load, so changing the directory recovers without relaunching. `walletDirectory` is no longer `required`, which previously made Raycast block the command with its own form so the directory-picker screen could never appear.
15. The existing `Suppress Read Errors` preference is preserved. Recursive scans honor it so opted-in users are not flooded with per-file toasts; when not suppressed, the toast includes a Change Wallet Directory action.
16. Deterministic ordering — `cardArr.sort()` without a comparator sorted objects as `[object Object]`, effectively preserving `readdir` order, which differs on Windows. Replaced with a `localeCompare` sort by name.
17. Symlinks — targets are resolved with `fs.statSync(path)` instead of `lstatSync(readlinkSync(...))`, which broke on relative targets and Windows junction points.
18. One directory scan instead of two — the Pocket dropdown used to read the directory independently of the grid (a module-level `savedPockets` cache plus a second `usePromise`); both now consume a single `fetchFiles()` result, and changing the filter no longer re-scans the disk.

Notes

* Changes were tested on Windows. The macOS-specific paths (JXA/AVFoundation video preview generation, raw path as image source, the `:` → `/` substitution in names) were left untouched precisely because they could not be verified on a Mac.
* Item 11 changes documented behaviour: the previous README explicitly declared nested folders invisible. The README has been updated.

## Screencast

25 08 2026 - 10ч27м45с
25 08 2026 - 10ч27м54с
25 08 2026 - 10ч28м27с

## Checklist


0xdhrv (approved):
Looks good to me, approved ✅

PR #30511, opened 2026-08-25
[Sonarr] Add a second instance with runtime switching

### LINKED ISSUE #28428: [Sonarr] Multiple Instances
### Extension

https://www.raycast.com/Aayush9029/sonarr

### Description

The raycast Radarr extension supports multiple instances. Can the Sonarr extension also allow support for multiple instances (like three)? Many people run a 1080p sonarr and a 4k sonarr. I also run an anime sonarr as well. it would be great if this extension supported that.

People who use multiple instances of Sonarr would benefit from this feature.

### Anything else?

_No response_

JohanDevl:
## Description

Closes #28428

Adds support for a second Sonarr instance, with switching between the two from any command — the setup people running separate instances (1080p / 4K / anime) asked for in that issue.

### Configuring the second instance

The second instance is described by the same preferences as the first one — host, port, URL base, connection type and API key — so both accept exactly the same setups (bare host, `host:port`, a full URL, a subpath behind a reverse proxy). Both resolve their URL through a single helper, so neither can end up with URL handling the other lacks.

Every second-instance preference is optional and gated behind an `Enable second Sonarr instance` checkbox: existing configurations keep working untouched, and nothing new has to be filled in. No existing preference was renamed or retyped, so no one has to reconfigure anything.

### Switching

The `Instance` section of every command's action panel switches instance (`⌘⇧I` / `Ctrl+Shift+I`). The selection is stored rather than kept in component state, so switching in one command carries over to the others and survives closing Raycast. The `Active Instance` preference seeds it, and every change to that preference wins over an earlier manual switch.

`Instance Status` was reworked to report every configured instance side by side, each with its own connection test, version and health checks. It also flags a second instance that is enabled but still missing its host or API key, instead of silently leaving it out of the list.

### Also in this PR

- **Fixes a crash on Windows.** Raycast returns `undefined` rather than an empty string for an optional preference that was never filled in, so reading `URL Base` threw `TypeError: Cannot read properties of undefined (reading 'trim')` and every command using it failed to open. Every preference is now read through a helper that tolerates it.
- **`API Key` becomes a password field**, so the key is masked in the preferences instead of being displayed in clear text. Raycast stores password preferences separately from plain ones, so the key may need to be entered once more after updating — this is called out in the changelog.

## Checklist

### DIFF
--- a/extensions/sonarr/.gitignore
+++ b/extensions/sonarr/.gitignore
@@ -17,6 +17,7 @@ compiled_raycast_rust
 # Claude Code
 .claude/
 CLAUDE.md
+.agents/
 
 # misc
 .DS_Store
--- a/extensions/sonarr/CHANGELOG.md
+++ b/extensions/sonarr/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Sonarr Changelog
 
-## [Second Instance and Windows Connection Fix] - {PR_MERGE_DATE}
+## [Second Instance and Windows Connection Fix] - 2026-08-25
 
 ### Added
 
--- a/extensions/sonarr/CHANGELOG.md
+++ b/extensions/sonarr/CHANGELOG.md
@@ -1,5 +1,23 @@
 # Sonarr Changelog
 
+## [Second Instance and Windows Connection Fix] - {PR_MERGE_DATE}
+
+### Added
+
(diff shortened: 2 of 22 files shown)


0xdhrv (approved):
Looks good to me, approved ✅

PR #30514, opened 2026-08-25
[Premier League] Fix stale fixtures, runaway scrolling and matchweek landing

kylemh:
## Description

- Matches sometimes opened on fixtures from 2006. It now waits until it knows which season and matchweek to show before loading anything.
- Scrolling a matchweek kept loading more games past the (usually) ten that exist, and eventually showed some fixtures twice as a result of fetch-scrolling. A matchweek now loads once, in full, and scrolling just scrolls. Pagination is explicitly and exclusively done via quick options.
- The **Matches** command would open on a matchweek that had already finished. Now, it opens on the next one to be played.
- When you _first_ open Matches, the previous matchweek sits above the upcoming one so last weekend's results are still there, and it drops away once the new matchweek kicks off. Focus **always** lands on the first upcoming fixture.
- The matchweek actions name where they take you, so "Next Matchweek (3)" and "Previous Matchweek (1)" instead of a "Matchweek (n)". "Next" comes before "previous" in the list, and they're on `]` and `[`. Both commands now show up at the top of the list.

## Screencast

N/A

## Checklist

### DIFF
--- a/extensions/premier-league/CHANGELOG.md
+++ b/extensions/premier-league/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Premier League Changelog
 
+## [Matchweek Navigation] - {PR_MERGE_DATE}
+
+- Fix Matches loading fixtures from the earliest seasons on record while the current season was still resolving.
+- Land on the upcoming matchweek instead of the one the gameweek config reports, and list the previous matchweek above it until the upcoming one kicks off.
+- Fetch a whole matchweek in one request. Scrolling no longer pulls in more matches or repeats fixtures already on screen.
+- Rename the matchweek actions to "Next Matchweek (n)" and "Previous Matchweek (n)", and bind them to `]` and `[`.
+
 ## [Fix Season Rollover] - 2026-08-21
 
 - Fix Matches, Table, Clubs and Squad showing the previous season once a new one kicks off. The season list the extension reads can trail the season it describes by months, so from July the active season is now read from the API and merged into that list.
--- a/extensions/premier-league/CHANGELOG.md
+++ b/extensions/premier-league/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Premier League Changelog
 
-## [Matchweek Navigation] - {PR_MERGE_DATE}
+## [Matchweek Navigation] - 2026-08-26
 
 - Fix Matches loading fixtures from the earliest seasons on record while the current season was still resolving.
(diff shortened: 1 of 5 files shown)


kylemh:
@anhthang made a few changes. Lemme know what you think!

anhthang:
On the scrolling fix: I'm not sure it's really a pagination issue. Back when v2 was in beta I checked and the runaway loading only happened on v2 — v1 worked fine. But I'm happy with the fixes.

On the landing behavior: I'd prefer to keep the Matches command showing the current gameweek exactly as the API returns it, rather than folding in the next matchweek's fixtures. Anyone who wants upcoming matches can use the Previous/Next Matchweek actions.

On the action labels: Could we switch to "Next: Matchweek 3" / "Previous: Matchweek 1"?

kylemh:
> On the action labels: Could we switch to "Next: Matchweek 3" / "Previous: Matchweek 1"?

Fine by me!

> On the landing behavior: I'd prefer to keep the Matches command showing the current gameweek exactly as the API returns it, rather than folding in the next matchweek's fixtures. Anyone who wants upcoming matches can use the Previous/Next Matchweek actions.

Fair enough. Could I suggest a new tab? Fixtures (Upcoming Matches) - a tab dedicated to viewing not-yet-played matches.

If one tries to use the extension as a means of determining the next match's kick-off time as-is, they're always forced into an extra fetch/action, but this new tab helps them avoid that.

anhthang:
> Fair enough. Could I suggest a new tab? Fixtures (Upcoming Matches) - a tab dedicated to viewing not-yet-played matches.

Sounds good to me. Fun fact: these used to be 2 separate commands. When I migrated to the new API version I merged them into one, mirroring how the EPL website does it — but I'm happy to bring back a dedicated Fixtures command for that flow.

One idea: with dedicated Fixtures and Results tabs, scrolling down could naturally load the next matches on Fixtures and the previous matches on Results. If we go that route, the Next/Previous Matchweek actions might not be needed anymore.

0xdhrv (approved):
Looks good to me, approved ✅

kylemh:
@anhthang i'll make a new PR asap. @0xdhrv we werent quite ready. I'll ping you when I think the PR is ready and @anhthang is happy in the future!

0xdhrv:
> @anhthang i'll make a new PR asap. @0xdhrv we werent quite ready. I'll ping you when I think the PR is ready and @anhthang is happy in the future!

You can keep it in the draft state that would help

kylemh:
I always avoid draft PRs since sometimes certain actions wont run.

PR #30522, opened 2026-08-25
Add region browser, favorites, port forwarding and settings to PIA Controls

vxirau:
## Description

Follow-up to #29672, which proposed this as a separate extension. Per @0xdhrv's review, it is instead contributed to the existing **PIA Controls** extension so PIA control stays consolidated in one place.

**All four existing commands keep their ids**, so hotkeys and workflows continue to work unchanged:

| Command | Status |
|---|---|
| `connect` | Kept. Still connects using PIA's currently selected region, without changing it. |
| `disconnect` | Kept. |
| `status` | Kept. Now also reports region and VPN IP. |
| `connectToRegion` | Kept. Upgraded into the richer region browser. |
| `toggle` | New. Connects when disconnected, disconnects when connected. |
| `connect-recent` | New. Reconnects to the region used last. |

### Added

- Country flags in the region list, searchable by country, city or region id
- Regions tagged when they support port forwarding, and when they are geo-located (IP registered in one country, server hosted elsewhere)
- An **Automatic** entry that lets PIA pick the fastest region
- Favorites and recently used regions
- Settings in the action panel: port forwarding (`⌘⇧P`), LAN access (`⌘⇧L`), and protocol
- Forwarded port surfaced and copyable once assigned
- Setup guidance when PIA, its CLI helper, or its daemon is unavailable

### Changed

- Region names now come from PIA's server catalog rather than being derived from the region id, so they read correctly. `srilanka` previously rendered as "Srilanka" and now shows "Sri Lanka". The catalog is cached for a day, with the cached copy served when offline.
- Commands no longer launch the PIA GUI through AppleScript on every invocation
- `piactl` is located from several install paths rather than assuming `/usr/local/bin`
- Shell execution replaced with `execFile` and argument arrays; region ids are validated before being passed through
- Toolchain updated to React 19 types and `@raycast/eslint-config` 2.x. This was necessary rather than cosmetic: the pinned React 18 types make every Raycast component fail to type-check as JSX against the current `@raycast/api`.

### Fixed

- A failed `piactl` read no longer resolves to `Disconnected`. Connection state is now an enum including `Unknown`, so a transient read failure cannot report an active VPN as off, cannot invert a toggle, and cannot let a wait loop confirm a state it never observed.
- Connecting to a region reports success only once the tunnel is observed reaching that state
- A failed connection attempt no longer replaces the last successfully used region

### Notes

- No credentials are read or stored; authentication stays in the PIA app.
- Flags ship with the extension rather than loading from an image CDN, so browsing regions makes no third-party requests. They are rendered from [flag-icons](https://github.com/lipis/flag-icons) (MIT), attributed in the README with the license text in `NOTICE.md`.
- Nothing connects, disconnects or changes a PIA setting unless the user triggers it. When PIA's daemon is inactive the extension explains the options instead of enabling background mode on the user's behalf.
- `piactl get pubip` returns the ISP-assigned address and does not change while the tunnel is up, so the VPN IP is shown when connected and the public IP only when disconnected, labelled as unprotected.

@Yakitrak, this replaces most of `src/utils.ts`, mainly because the region model has to carry country code, port-forwarding and geo flags sourced from PIA's catalog, and because connection state becomes an enum rather than a boolean. Happy to adjust the scope if you would prefer a smaller change.

## Screencast

**Connected status, with region flag, protocol and VPN IP**

Connected status

**Region list, with flags and the port-forwarding and geo tags**

Region list

**Settings from the action panel**

Settings panel

## Checklist


pernielsentikaer:
@Yakitrak  can you have a look at this one?

pernielsentikaer (changes requested):
Hi @vxirau 👋

Thanks for bringing these improvements into the existing extension 💪

Could you address these two issues?

- In `src/lib/regions.ts`, use `piactl get regions` as the source of accepted IDs and enrich those with catalog metadata instead of slugifying display names. PIA’s own metadata fixtures include names that differ from the catalog, causing valid-looking selections to fail. For example, the fixture maps `US New York` to `US New York City`, producing `us-new-york-city` rather than `us-new-york`. I verified this against [PIA’s CLI naming logic](https://github.com/pia-foss/desktop/blob/3d76d75c7e5693812aa1b72108cb44df7397dd32/cli/src/getcommand.cpp#L70-L147) and [metadata fixture](https://github.com/pia-foss/desktop/blob/3d76d75c7e5693812aa1b72108cb44df7397dd32/tests/res/metadata-v2.json), not a live PIA installation.
- In `ConnectionStatus.tsx`, the “Connect” action disconnects when PIA is connecting or reconnecting. Please align the label with `isActive` and use explicit connect/disconnect callbacks so a state change cannot reverse the action’s intent. For example, with `isActive` imported from `../lib/pia` and the callbacks from `../lib/actions`:

```tsx
const disconnects = isActive(status.state);

<Action
  title={disconnects ? "Disconnect" : "Connect"}
  onAction={disconnects ? disconnectVpn : connectCurrent}
/>
```

---

pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #30529, opened 2026-08-25
Update claude-artifacts extension

chrismessina:
## Description

The extension reads a local index that a Claude Code `PostToolUse` hook writes to. If that hook ever stops running, the index simply stops growing — and a list of real artifacts that is quietly frozen looks exactly like a list that is up to date. There was no way to tell the two apart from inside Raycast.

This adds that signal. When no artifact-recording hook is registered, an **Artifact Tracking Is Off** row appears above the results, leading to a setup screen that carries a prompt you can paste into Claude Code to install and register the hook for you.

image

The detection deliberately requires the hook's *command* to be an artifact recorder, not merely a matcher that covers the `Artifact` tool — unrelated integrations register catch-all `PostToolUse` hooks that also fire on `Artifact`, and matching on the matcher alone would report the index as healthy while nothing was writing to it. A settings file that cannot be parsed yields "unknown" rather than "missing", so an unreadable file never produces a warning that cannot be substantiated.

No new commands, no new preferences, no network calls.

image

### Changes

- Added an **Artifact Tracking Is Off** row at the top of the list when the Claude Code hook is not registered
- Added a **Set Up Artifact Tracking** screen with a prompt that can be pasted into Claude Code, replacing a bare JSON fragment that gave no indication of where it was meant to go
- The same screen now backs the first-run empty state, so setup guidance no longer differs depending on how you reached it
- Renamed the Finder actions from **Reveal** to **Show**, matching the verb used elsewhere in Raycast
- Refreshed dependencies within their current majors and cleared a high-severity `brace-expansion` advisory

## Checklist


0xdhrv (approved):
Looks good to me, approved ✅

PR #30530, opened 2026-08-26
Update app-store-connect extension

chrismessina:
## Description

Adds support for Apple's **individual API keys** and fixes a set of bugs in the existing commands, several of which silently corrupted what you saw on screen — a removal that blanked an entire list, a rejected key that deleted the working one, and mutations that reported success without sending a request.

The data layer is rebuilt on `useCachedPromise` cursor pagination. The previous hand-rolled hook could not refetch at all: a `previousPath` ref made re-requesting an already-loaded path a permanent no-op, so every mutation hand-spliced local state instead — which is where several of the bugs below originated.

Removes more than it adds, and drops an unused dependency.

### Individual API keys

Apple's individual keys carry no Issuer ID: they sign `sub: "user"` and omit `iss` entirely, while the JWT header is unchanged ([Apple docs](https://developer.apple.com/documentation/appstoreconnectapi/generating-tokens-for-api-requests)). Issuer ID is now optional end to end, and the credential form asks which key type you're adding.

### Fixes

- **Removing a team member or revoking an invitation cleared the whole list.** A callback parameter shadowed the outer one, making `user.id !== user.id` always false, so `.filter()` returned empty. The request itself targeted the right record — only the UI lied.
- **Adding a rejected key deleted the previously working one.** The rollback read a render-time `currentTeam`, which is the *previously* selected key, not the one just added.
- **A network blip, a permission error, or an unreadable key file during sign-in deleted stored credentials.** Rollback is now limited to a 401 (a 403 means the key is valid but lacks permission), and the private key is parsed before anything is persisted.
- **Adding or removing builds from a beta group failed from the Manage Builds screen.** Three URLs carried a trailing space that encoded as `%20`; the same calls elsewhere were correct, which is why it worked from one screen and not another.
- **Testers were created with a missing invite type and state** — both were read from the wrong level of the response, and two schema validations were computed then discarded.
- **Mutations silently did nothing while reporting success** when credentials were missing.
- **Lists appended duplicate rows when paginating**, and showed the previous app's rows while switching.
- **A failed removal left the row missing** from the list.

### Also

- App icons resolve in batched requests instead of one per row (previously one `/builds` call per visible app, against a rate-limited API).
- Failure toasts carry a **Copy Error** action, and App Store Connect's errors are rewritten into plain language instead of repeating the same sentence twice.
- Builds that can't be added to a group say so *before* submission rather than after Apple rejects them.
- Actions are grouped into sections headed by the item they act on; Builds has a real empty state; emoji in status labels are replaced with themed icons that respect the user's theme.

### Dependencies

`@raycast/api` 1.77 → 2.0.6, `@raycast/utils` 1.16 → 2.3, React 19 types, Node 22. Dropped `node-app-store-connect-api` (no imports anywhere) and `@types/node-fetch`; `jose` was reachable only transitively through the former and is now a declared dependency.

## Checklist


extensions/app-store-connect/src/Hooks/useAppStoreConnect.tsx
```diff
@@ -19,178 +28,193 @@ export class ATCError extends Error {
...
+      // `links.next`.
+      const isCollection = Array.isArray(mapped);
+
+      return {
+        data: boxPagedResult(mapped),
+        hasMore: isCollection && cursor !== undefined,
+        cursor: isCollection ? cursor : undefined,
+      };
+    },
+    [path ?? ""],
```
bangerang: **Include the selected team in the cache key**

The cache key only uses the API path. After switching teams, a command can briefly show data cached for the previous team. If the new request fails, that old data may stay visible.

Can we include the Key ID and Issuer ID/key type in the cache key, or clear the cache when switching teams?

extensions/app-store-connect/src/Components/SignIn.tsx
```diff
@@ -31,49 +44,63 @@ export default function SignIn({ children, didSignIn }: SignInProps) {
...
+      if (!values.apiKey || (!isIndividualKey && !values.issuerID)) {
         return;
       }
 
       setIsCheckConnection(true);
+      // Declared out here so the catch rolls back the exact key it added; stays undefined
+      // if we failed before persisting anything.
+      let addedTeam: Team | undefined;
 
       const privateKeyContent = fs.readFileSync(file, "utf8");
```
bangerang: **Handle errors when reading the key file**

`readFileSync` runs outside the `try`, after loading starts. If reading fails, no error is shown and the form stays loading.

Can we move the file read and credential creation inside the `try`?

extensions/app-store-connect/src/Hooks/useAppStoreConnect.tsx
```diff
@@ -19,178 +28,193 @@ export class ATCError extends Error {
...
-    if (loadAll) {
-      if (pagination) {
-        if (pagination.hasMore) {
-          pagination.onLoadMore();
-        }
-      }
+    // ponytail: hard page ceiling — this drives an unattended loop against a
+    // rate-limited API, so a runaway must terminate even if `hasMore` is wrong.
+    if (autoLoadState.current.pages >= MAX_AUTO_PAGES) {
+      autoLoadState.current.truncated = true;
```
bangerang: **Update the UI when pagination is truncated**

Changing this ref does not trigger a render. As a result, `isTruncated` may stay `false` in the UI even after the page limit is reached.

Can we store this value in state so consumers update immediately?

bangerang (changes requested):
Thank you for the contribution! The overall direction looks good.
Before merging, please address the cache issue so data from one App Store Connect team cannot be shown after switching teams. I also left two smaller comments about key-file error handling and the pagination limit.

pernielsentikaer:
Hi @chrismessina 👋

Could you look into the feedback from @bangerang 🙂

---

chrismessina:
Thanks — all three addressed.

**Cache scoped to the credential.** The key was the API path alone, identical for every team. It's now `[path, keyId:issuerId]`, and one credential snapshot both answers the cache-identity check and signs the request, so a response can't be filed under another account's key.

**Key file read moved inside `try`.** It was worse than it looked — the throw escaped past `finally`, so the form span forever with no error. Both credential forms also returned silently on an unreadable path; that's fixed too.

**`isTruncated` is state now.** You were right that the ref change didn't re-render — the "showing first N" notice had never actually appeared.

The diff also grew from hands-on testing: **"Manage Teams" is now "Manage API Keys"** (display title only — the command `name` is unchanged so existing hotkeys and deeplinks keep working), with key renaming, Individual/Team sections, and clearer removal copy, since "Delete Team" read as deleting the team at Apple when it only ever removed a locally stored key. 

The icon now matches Apple's current App Store Connect app icon, which also adopts the Liquid Glass treatment ahead of macOS 27.

bangerang (approved):

pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #30533, opened 2026-08-26
Google Chrome Profiles: fix profile switch opening the wrong profile

tos-jdkwon:
## Description

Fixes a bug where switching to a profile that's signed into a Google account brings a *different* profile to front instead.

**Root cause:** Chrome's Profiles menu bar item doesn't label a signed-in profile with its own name — it shows `${gaia_given_name} (${name})` (e.g. a profile named "Work" signed in with a Google account whose given name is "Alex" shows as "Alex (Work)" in the menu). The extension's exact-name match against the raw profile name therefore never matched a signed-in profile, and always fell through to a substring search across every menu item. That fallback could — and did — silently click an unrelated profile whose name happened to be a substring of the search text (e.g. searching "Work" could match "Work admin" or "old work" depending on menu order), with no visible error.

**Fix:** look up the profile's `gaia_given_name` from `Local State` (keyed by profile directory, which every caller already has — including the Quicklink/deeplink command, which never carried Google-account info across the deeplink and had the same latent bug) to construct the label Chrome actually displays, then try it and the raw profile name as exact-match candidates only. The substring fallback is removed entirely — a clear "Profile not found in menu" failure is safer than guessing and opening the wrong profile's session.

## Screencast

Straightforward logic fix scoped to the AppleScript profile-menu matching in `src/util/util.ts`; no UI change. Verified via `ray develop` against a real multi-profile Chrome instance with several Google-account-signed-in profiles that reproduced the bug, confirming `⌘+Return` now brings the correct profile to front.

## Checklist


extensions/google-chrome-profiles/src/util/util.ts
```diff
@@ -331,7 +371,14 @@ export const openGoogleChrome = async (
     return;
   }
 
-  const escapedProfileName = escapeAppleScriptString(profile.name);
+  const googleAccountLabel = await getGoogleAccountMenuLabel(browser, profile);
```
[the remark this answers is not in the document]
frouo: Is it possible to bring `gaia_given_name` to the openGooglChrome's profile in param, so it's not needed to read the Local State file.

frouo (commented):
Thank you for your contribution

pernielsentikaer (commented):
Hi @tos-jdkwon 👋

Can you resolve the merge conflicts in this PR?

extensions/google-chrome-profiles/src/index.tsx
frouo: If I remember well, `gaia_given_name` will be available among `gaia_name`, `user_name` and `last_downloaded_gaia_picture_url_with_size` (= the chrome user authenticate with a google account).

So you could add `profile.gaia_given_name` in the list of `&&`, and:

```
ga: {
  name: profile.gaia_name,
  givenName: profile.gaia_given_name,
  ...
```

extensions/google-chrome-profiles/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Google Chrome Profiles Changelog
 
+## [Fix] - 2026-08-25
```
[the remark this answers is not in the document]
tos-jdkwon: Fixed — confirmed against the actual convention in `docs/basics/prepare-an-extension-for-store.md:177`, thanks.

extensions/google-chrome-profiles/src/util/util.ts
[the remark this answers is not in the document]
tos-jdkwon: Good call — done. `Profile` now carries a `givenName` field: the main list reads it straight from `Local State` alongside the rest of the profile (no extra I/O), and a freshly created Quicklink carries it across the deeplink too. `openGoogleChrome` uses that directly.

The only place it still reads `Local State` is a fallback in `resolveGivenName`, and only when `profile.givenName` is absent — which happens for a Quicklink created before this field existed (its deeplink payload only has `directory`/`name`). Without that fallback such a stale Quicklink would silently fail for a signed-in profile (exact match against just the raw name never matches, and the AppleScript error is swallowed by the detached process). So it's a compat path for existing users' Quicklinks, not the common case.

extensions/google-chrome-profiles/src/util/util.ts
```diff
@@ -349,24 +396,19 @@ export const openGoogleChrome = async (
...
-
-          if foundMatch is false then
-            error "Profile not found in menu"
+        set candidateNames to {${escapedCandidates}}
+
+        set foundMatch to false
+        repeat with candidateName in candidateNames
+          if candidateName is in menuItems then
+            click menu item candidateName of profileMenu
+            set foundMatch to true
```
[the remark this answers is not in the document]
tos-jdkwon: Verified this empirically rather than assuming — Chrome's real Profiles menu does include `missing value` entries (separators), confirmed against a live menu on my machine. Tested the exact pattern used here (`candidateName is in menuItems` where `menuItems` contains `missing value`, both a matching and a non-matching case) directly in `osascript`; it doesn't throw — `is in` list-membership comparisons tolerate `missing value` elements fine. The type-error risk applies to iterating `menuItems` directly and calling `contains` on each element (what the old code did, hence its `is not missing value` guard) — this code iterates `candidateNames` instead, which never contains `missing value`, and only uses `menuItems` as the static right-hand side of `is in`. No change needed here, but appreciate the scrutiny on this — AppleScript's `missing value` handling is exactly the kind of thing that's easy to get wrong silently.

extensions/google-chrome-profiles/src/open-profile.tsx
```diff
@@ -43,7 +44,11 @@ export default async function Command(props: LaunchProps) {
       break;
   }
 
-  const profile: Profile = { directory: profileDirectory, name: profileName };
+  const profile: Profile = {
+    directory: profileDirectory,
+    name: profileName,
+    ...(profileGivenName && { givenName: profileGivenName }),
```
frouo: Can be simplified:

```
givenName: profileGivenName,
```

extensions/google-chrome-profiles/src/util/util.ts
```diff
@@ -293,6 +294,45 @@ const launchInProfileCommand = (browser: BrowserConfig, profileDirectory: string
...
+const getGoogleAccountMenuLabel = async (
+  browser: BrowserConfig,
+  profile: { name: string; directory: string },
+): Promise<string | undefined> => {
+  try {
+    const path = join(homedir(), browser.dataPath, "Local State");
+    const localState: GoogleChromeLocalState = JSON.parse(await readFile(path, "utf-8"));
+    const givenName = localState.profile.info_cache[profile.directory]?.gaia_given_name;
+    return givenName ? `${givenName} (${profile.name})` : undefined;
+  } catch {
```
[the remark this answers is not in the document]
tos-jdkwon: Fixed — the catch now logs the caught error at `console.debug` verbosity before falling back, so an I/O error vs. malformed `Local State` vs. schema mismatch is distinguishable without changing the non-fatal fallback behavior.

tos-jdkwon:
@pernielsentikaer Merge conflicts are resolved — merged `origin/main` (which had picked up an unrelated profile-deletion feature touching the same files) and re-verified `npm run build`/`npm run lint` pass on the merged result.

frouo (changes requested):

tos-jdkwon:
Retriggering CI — the failed "Get Extensions" job was a runner-scheduling infra issue ("job was not acquired by Runner of type hosted"), not a code failure.

extensions/google-chrome-profiles/src/index.tsx
tos-jdkwon: This comment landed on `a682665db5`, which is exactly what that commit did — but I moved `givenName` out of `ga` and onto `Profile` directly in the very next commit (`3841928941`, pushed a couple minutes before this review), so this should be resolved on the current diff. Reasoning, since it's a deliberate change from what you're suggesting here rather than an oversight:

Nesting it in `ga` works fine for the main list, but breaks `open-profile.tsx` (the Quicklink/deeplink command) — its `Profile` only has `directory`/`name`/`givenName` from the deeplink, no `email`/`pictureURL`, so constructing a `ga` object there means either faking those fields or widening `ga`'s other fields to optional (which I did try first, then didn't love — it's a type-level workaround for a shape only one caller needs). Pulling `givenName` out to a top-level `Profile` field sidesteps that entirely: `open-profile.tsx` sets exactly the one field it has, `ga` keeps its original required shape, and `index.tsx` sets both `givenName` and `ga` independently since it has all of it. Happy to go back to nesting it if you'd still prefer that despite the open-profile.tsx wrinkle — let me know.

extensions/google-chrome-profiles/src/open-profile.tsx
tos-jdkwon: Same timing note as the index.tsx thread — this is exactly what it already reads as on the current diff (commit `3841928941` onward): `givenName` moved to a top-level `Profile` field, so this line is just `...(profileGivenName \&\& { givenName: profileGivenName })`, no `ga` wrapper.

extensions/google-chrome-profiles/src/index.tsx
frouo: Ok I see

extensions/google-chrome-profiles/src/open-profile.tsx
frouo:
```
  const profileGivenName = props.launchContext?.givenName;
```

so profileGivenName can be undefined. And Profile's `givenName` is also optional, so you don't need to check `profileGivenName` before insert (you can insert undefined, it's harmless and easier to read.

```
  const profile: Profile = {
    directory: profileDirectory,
    name: profileName,
    givenName: profileGivenName,
```

tos-jdkwon: Done in `a63c0511` — assigned directly. Applied the same simplification to the two matching sites in `index.tsx` (the `info_cache` extraction and the deeplink context; `JSON.stringify` drops an `undefined` field there anyway) so the pattern's consistent.

frouo (approved):
Thank you. Don't forget to add your @tos-jdkwon in the package.json `contributors`.

pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #30534, opened 2026-08-26
Fix Zed Recent Projects: focus existing window instead of opening a duplicate

dadia-willand:
## Description

Fixes "Search Recent Projects" always opening a new Zed window, even when the selected project is already open.

Recent Zed versions (observed on Zed 1.16.2) changed the CLI default: `zed <path>` now creates a new window on every invocation instead of reusing the window that already has that workspace. As a result, the action labeled "Focus Window" (shown when `entry.isOpen` is true) actually spawned a duplicate window.

The extension can't rely on the CLI for this (`-e/--existing` opens the project in the *frontmost* window, replacing its workspace, rather than focusing the correct one). So this PR raises the existing window directly:

- Added `focusZedWindow(windowTitle, bundleId)` to `lib/zed.ts` — a System Events AppleScript that finds the Zed window whose name contains the entry title and performs `AXRaise` on it. It mirrors the existing `closeZedWindow` helper, so it needs no new permissions beyond what "Close Project Window" already uses.
- In `search.tsx`, both CLI open paths (single-folder and multi-folder) now try `focusZedWindow` first when `entry.isOpen`, and fall back to the previous `openWithZedCli` behavior if the entry isn't open or no matching window is found. The CLI paths are macOS-only (`getZedCliPath` returns `null` elsewhere), so the AppleScript never runs on Windows.

Verified locally on macOS with Zed 1.16.2: selecting an already-open project raises its window (window count unchanged), selecting a closed project opens it as before, and cold start with Zed not running still works.

## Screencast

Behavior verified by counting Zed windows via `CGWindowListCopyWindowInfo` before/after: plain CLI open of an already-open project went 1 → 2 windows on Zed 1.16.2; with this change the count stays constant and the correct window is raised.

## Checklist


extensions/zed-recent-projects/src/search.tsx
```diff
@@ -245,7 +254,11 @@ function OpenInZedAction({ entry, revalidate }: { entry: Entry; revalidate: () =
   if (cliPath) {
     const openSingleFolder = async () => {
       try {
-        await openProject(() => openWithZedCli(cliPath!, [entry.paths[0]]), closeMainWindow);
+        if (await focusIfOpen()) {
+          await closeMainWindow();
+        } else {
+          await openProject(() => openWithZedCli(cliPath!, [entry.paths[0]]), closeMainWindow);
```
ewgenius, line 264: I don't think this is correct, especially under cli path branch. Also that might break it on windows.
I'm pretty sure zed cli or zed settings have some options on how to handle new vs existing window behavior, it is worth checking it

ewgenius (changes requested):

extensions/zed-recent-projects/src/search.tsx
dadia-willand: Thanks for taking a look! I checked both points before reaching for AppleScript — details below.

**Windows:** these branches can't execute on Windows. `cliPath` comes from `getZedCliPath()`, which returns `null` unless macOS, and both call sites are gated on `if (cliPath)`. I've still added an explicit `isMac` guard to `focusIfOpen` (0de2e76) so the platform assumption is visible at the call site instead of implied.

**CLI flags / settings:** none of them focus the window that already contains the project — that's the gap this PR works around:

- `-n/--new` always opens a duplicate window.
- `-a/--add` adds paths to the *currently focused* window, not the matching one.
- `-r/--reuse` / `-e/--existing` reuse the frontmost window, replacing its workspace. I tested this on Zed 1.16.2: selecting an already-open project switches whatever window you had focused to it, and you end up with the project open twice.
- `cli_default_open_behavior` only has two values, `existing_window` (open in the current window's sidebar) and `new_window` (always a new window, even if already open) — see https://zed.dev/docs/windows-and-projects.

"Focus the window whose workspace matches the path" (what `code .` does, and what plain `zed <path>` did before the multi-project overhaul in zed-industries/zed#59415) is an open feature request: zed-industries/zed#61575, also zed-industries/zed#54657 for Open Recent. Until Zed ships it, raising the window via System Events seemed like the least-bad option — it reuses the exact mechanism `closeZedWindow` already uses for the "Close Project Window" action, and falls back to the previous CLI behavior when no window matches. Happy to switch to a CLI flag if Zed adds the focus behavior, or to adjust the approach if you'd prefer something else.

dadia-willand: sorry im just using AI to address this. not really familiar with this extension or raycast extensions as a whole 🙇  just wanted to fix this cos it's breaking in my local mac 🙏

dadia-willand: if you have a better solution for this. please let me know 🙏

ewgenius: Got it, thanks for checking it.

> Until Zed ships it, raising the window via System Events seemed like the least-bad option

Agreed, but that shouldn’t break any existing extension behavior for anyone who relies on the Zed CLI.
I’d suggest making this an optional extension parameter in the settings.

Personally, I would like this to be consistent with the default or configured behavior of Zed "open project". If I have my Zed configured to open in new window, I would expect it to work the same way from Zed project picker, cli or Raycast.

And in general I would prefer this extension to be thin, and use native Zed cli and built in Raycast functionality as much as possible, and get away from custom hooks and platform dependent scripts.
Unfortunately, I don’t use it as much as I used to and don’t have time to revisit the UX and cleanup.

dadia-willand: Done in f280376 — the focus behavior is now behind a new "Focus Already-Open Projects" checkbox, off by default. With the preference off, nothing changes: every path goes through the Zed CLI exactly as before, so anyone relying on their Zed `cli_default_open_behavior` config keeps that behavior in Raycast too. Only users who explicitly opt in get the window-raise, and it still falls back to the CLI when no matching window is found.

Agreed on keeping the extension thin — if Zed ships the focus behavior from zed-industries/zed#61575, this preference can be dropped and the AppleScript replaced with a plain CLI flag. I've noted that in the code comment's intent; happy to do that follow-up when it lands.

pernielsentikaer (changes requested):
Hi @dadia-willand 👋

Thanks for working through the earlier feedback 💪

The window matching still relies on heuristics that don’t cover Zed’s supported titles. For example, a project `FOO` with an active extensionless file `BUILD` can have the title `FOO - BUILD`, but the matcher rejects it. Focus then opens another window, while the existing “Close Project Window” action now fails with “Window not found.”

The path fallback can also match a project path embedded inside a different absolute path, and it misses filesystem-root or trailing-slash project paths.

Could you replace the filename allowlist/title heuristics with a reliable path-based signal where available—such as Zed’s `AXDocument` value—and add coverage for arbitrary extensionless files, path containment, root paths, and trailing separators? Please ensure the existing close action doesn’t regress when the optional focus preference is disabled.

---

PR #30537, opened 2026-08-26
fix(standing-desk-tracker): count in-progress session in today's stats

esttorhe:
## Description

Fixes the `Standing Desk` menu bar command showing `0s` for **Today's Stats** while a session is in progress.

`loadDailyStats()` read the `currentState` React state variable, but it is called from `loadData()` on the line right after `setCurrentState(state)`:

```ts
async function loadData() {
  const { state } = await getCurrentState();
  setCurrentState(state);          // state setters don't mutate the existing closure
  ...
  await loadDailyStats();          // still sees the previous render's currentState
  await checkAndShowMotivation(state);
}
```

State setters don't mutate the closure that `loadDailyStats` captured, and the mount effect has an empty dependency list so it never re-runs. On every background refresh `currentState` is therefore still `null`, the `if (currentState)` guard is never taken, and the in-progress session is never added. Today's Standing / Sitting / Total only ever counted *completed* sessions — so a fresh install that has toggled once shows `0s` indefinitely.

The fix passes the resolved state in explicitly, the same way `checkAndShowMotivation(state)` already receives it, plus `newState` from `handleToggleState`. No behaviour change beyond the in-progress session now being counted.

Note `standing-desk-tracker.tsx` has the same shape, but it self-heals there because its effect re-runs on `[currentState]`, so I left it alone to keep this focused.

## Screencast

Straightforward logic fix with no UI change — the affected rows are the existing `Standing` / `Sitting` / `Total` items in the menu bar's "Today's Stats" section, which now include the running session instead of omitting it.

Reported from a fresh store install where "Today's Stats" read `Standing 0s / Sitting 0s / Total 0s` during an active sitting session.

To be precise about what I verified: `ray build` + `ray lint` pass, and I installed the distribution bundle and confirmed the command runs on its background interval. I have not captured a before/after screenshot of the menu bar itself, so the fix rests on the control-flow argument above rather than a visual diff. Happy to add a screencast if that's wanted for review.

## Checklist

### DIFF
--- a/extensions/standing-desk-tracker/CHANGELOG.md
+++ b/extensions/standing-desk-tracker/CHANGELOG.md
@@ -25,4 +25,3 @@
 - Quick toggle command for switching states
 - Session history and statistics calculation
 - Configurable preferences for notifications and goals
-
--- a/extensions/standing-desk-tracker/CHANGELOG.md
+++ b/extensions/standing-desk-tracker/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## [Fix today's stats ignoring the in-progress session] - {PR_MERGE_DATE}
+## [Fix today's stats ignoring the in-progress session] - 2026-08-26
 
 ### Fixed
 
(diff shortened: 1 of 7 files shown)


0xdhrv (approved):
Looks good to me, approved ✅

PR #30539, opened 2026-08-26
Update raycast-language-tool extension

pasqualesalza:
## Description

`Check Text Instant` builds its request by hand and leaves out six of the extension's preferences. `Check Text` passes all of them, so the same settings apply in one command and are silently ignored in the other:

```
motherTongue  preferredVariants  enabledRules
enabledCategories  disabledCategories  enabledOnly
```

They are extension-wide preferences, the manifest says the values are used whenever they are filled in, and `checkTextWithAPI` already accepts and serialises all six — nothing else was needed to make them work.

The effect is easy to miss because the command replaces the selection without showing anything. For example, setting *Disabled Categories* to `TYPOS` should stop LanguageTool from "correcting" English technical words inside Italian text — `commit` becomes `commi`, `deploy` becomes `dello`. That setting had no effect at all in this command:

```
$ curl -d "language=it-IT" --data-urlencode \
       "text=Ho fatto il commit e poi il deploy sul server." \
       https://api.languagetool.org/v2/check
3 matches

$ curl ... -d "disabledCategories=TYPOS" ...
0 matches
```

Verified in Raycast: with `TYPOS` in *Disabled Categories*, Check Text Instant now leaves those words alone, and mangles them again once the field is cleared. Nothing changes for anyone whose fields are empty, which is the default.

## Screencast

N/A

## Checklist


0xdhrv (approved):
Looks good to me, approved ✅

PR #30556, opened 2026-08-26
[Premier League] Split Matches and Fixtures into separate commands

kylemh:
## Description

**What:** Splits match browsing into two commands. Matches now covers what has been played, Fixtures covers what is coming, and each scrolls in its own direction. Follows on from #30514, which merged before the interaction model had settled.

**Why:** One command trying to be both past and future made every navigation choice ambiguous, and #30514 answered that by removing scrolling entirely and stacking two matchweeks on the landing view. Splitting the command instead gives each side a single obvious direction of travel, so scrolling can come back and mean something: down is further back in Matches, further ahead in Fixtures.

**Fix:** `MatchweekList` is a shared, direction-aware list that both commands render, paginating a matchweek at a time rather than by API cursor, so a page boundary always lands on a matchweek boundary and fixtures can't repeat. Matches starts on the current matchweek and sorts `kickoff:desc`; Fixtures starts on the next unplayed one and sorts `kickoff:asc`. The matchweek actions are now "Next: Matchweek n" and "Previous: Matchweek n", and each command offers only the one that keeps you on its side of today until you have moved, after which both are available. A "Jump to…" action on `J` takes a matchweek between 1 and 38. The guard from #30514 that stops any request going out before the season resolves is kept.

## Screencast

N/A

## Checklist


kylemh:
@anhthang

anhthang:
LGTM

kylemh:
@0xdhrv ready!

anhthang:
image

I haven't work with Form yet, but that's better if we can improve the UI (the red text). Also, use `Enter` as shortcut to jump is better imo

kylemh:
on it @anhthang -im sure it's obvious im vibe-coding here, but definitely hate that!

extensions/premier-league/src/components/jump_to_matchweek.tsx
```diff
@@ -2,21 +2,34 @@ import { Action, ActionPanel, Form, Icon, useNavigation } from "@raycast/api";
...
+  const digits = next.replace(/[^0-9]/g, "");
+
+  if (!digits) {
+    return "";
+  }
+
+  const value = Number(digits);
+
+  return value >= 1 && value <= TOTAL_MATCHWEEKS ? String(value) : previous;
+};
```
kylemh, line 15: avoided the error UI outright

extensions/premier-league/src/components/matchweek_list.tsx
```diff
@@ -211,8 +211,10 @@ export default function MatchweekList(props: {
       <Action.Push
         title="Jump to…"
         icon={Icon.MagnifyingGlass}
-        shortcut={{ modifiers: [], key: "j" }}
-        target={<JumpToMatchweek onJump={setSelectedMatchweek} />}
+        shortcut={{ modifiers: [], key: "return" }}
```
kylemh, line 221: J --> Return

pernielsentikaer (approved):
Looks good to me, approved 🔥

kylemh:
now that I'm using it @anhthang feels like "Matches" should be renamed to "Results"

gonna open a PR to revert "Enter" back to "J" since another command already has the same hotkey
Screenshot 2026-08-31 at 3 13 00 PM

PR #30561, opened 2026-08-26
Update tuya-smart extension

AndresMorelos:
## Description

Maintenance and feature update for the Tuya Smart extension.

- Tuya API failures are now reported. Tuya answers HTTP 200 on failure and reports the outcome in the payload, so errors were never raised: an expired IoT Core subscription showed an endless spinner or an empty device list with no explanation
- Device list regrouped into Controls, Sensors and Locks, with each device appearing once; sockets were previously listed twice, once as a device and again as a bare `switch_1` row
- Sensor readings are formatted: a temperature arriving as `294` now reads `29.4°C`, and a contact sensor reads Open or Closed instead of true or false
- Battery level and devices needing attention are surfaced in the list
- Offline devices are marked as such
- Encoded internal data points are hidden from the details view; a smart lock drops from 22 rows to the 11 that mean something
- Added brightness and colour temperature control for lights, using each product's own reported range
- Added a menu bar command for pinned devices
- Added a Temperature Unit preference; temperatures follow the device's own setting by default, matching the Tuya app, with Celsius or Fahrenheit available for products that never report one
- Added AI tools to list devices, toggle switches and set brightness
- Added a Control Device command taking arguments, so a device can be controlled from a deeplink through Apple Shortcuts and, through Shortcuts, by voice with Siri
- Curtains and light work modes are controllable again; the redesigned list had dropped every data point that takes one of a set of values
- Commands fall back to Tuya's local network protocol when the cloud is unavailable
- Refreshing takes one request per 20 devices instead of one request per device
- Fixed "Active Time" showing a meaningless duration; `active_time` is a Unix timestamp, not an elapsed time
- Fixed identically named devices hiding each other through duplicate React keys
- Fixed the On/Off filter applying only to the switches section
- Fixed turning a switch off reporting that it had been turned on
- Fixed the rename form opening empty instead of prefilled
- Updated to Raycast API 2.0, React 19 and ESLint 9, and cleared outstanding axios advisories

## Credit

The Control Device command comes from #29505 by @anwarulislam, who proposed deeplink control
and wrote the first version of it. This implementation is rewritten against the current
architecture, but the idea and the design of the argument surface are theirs. Supersedes
#29505, which can be closed once this merges.

## Behaviour changes for existing users

- Pinning is now per device rather than per switch; existing switch pins are not carried over
- The details panel starts closed and toggles with `Cmd+Shift+D`; it was previously always open, which left the list column too narrow for the state and battery labels
- Asking AI to turn a device off requests confirmation; turning one on does not

## Screencast

https://github.com/user-attachments/assets/a26bd56f-eb54-4642-aecc-dbd4cd867227

## Checklist


AndresMorelos:
Thanks — all three findings were real. Two are fixed as suggested; the third is fixed differently, and it seemed worth explaining why rather than closing it silently.

**1. Failed commands persist state — fixed, and it was worse than reported.**

The write lands in `useCachedState`, so the unapplied value was persisted to disk and survived a relaunch until the next successful refresh overwrote it, rather than being a transient UI glitch.

Both consumers now go through a shared `applyCommandResult`, which returns the device untouched when `result` is false. As a second line of defence, the failure path in `runCommand` now returns the value the device still holds, so a consumer that forgets the guard still cannot persist a phantom state. `menubar.tsx` already handled this correctly via `mutate(..., { rollbackOnError: true })`; the list now matches it.

**2. Unmatched switch name selects the first switch — fixed as suggested.**

The `?? switches[0]` fallback is gone. An explicit `switchName` that matches nothing now returns `undefined`, and the AI tool reports which switches the device actually has so the assistant can retry instead of operating a different gang. The no-query path still resolves to the single switch, which is the documented behaviour.

**3. Encoded statuses reach the assistant — symptom confirmed, fixed with a separate predicate.**

Confirmed against a real account: `synch_method = "AQE="` was reaching the model.

I did not reuse `isNoiseStatus` as suggested, because it also hides `countdown_N === 0`. Whether a timer is running is a reasonable thing to ask an assistant, and that answer would have silently disappeared. The two predicates answer different questions — what a person wants to read, versus what anything can parse — so there is now a separate `isOpaqueStatus` covering encoded blobs and internal bookkeeping, and the AI path filters on `sensitive || opaque`. Every human-meaningful data point still reaches the model, and the existing `omitted` line tells it something was withheld.

All three fixes are covered by tests that fail against the previously reviewed commit and pass now.

AndresMorelos:
Valid, and the same pattern was present in a place the review did not flag, so both are fixed.

**Ambiguous switch name selects first — fixed.**

The substring branch used `find`, so a partial name matching several gangs resolved to whichever came first.

**The device lookup had the identical defect, and it mattered more.**

`findDeviceByName` used the same `find`-per-tier shape. Measured against a real account with fifteen devices:

- `"tomacorriente"` matched 5 devices and silently resolved to one of them; another of the five is a fridge
- `"cuarto"` matched 3 devices spanning a socket, a door lock and a window sensor
- `"enchufe"` matched 3, `"ventana"` matched 3

So "turn off the outlet" could have reached the fridge, and a request naming a room could have reached a lock.

**Both now resolve only when exactly one target matches.** Two behaviours are deliberately preserved:

- an exact name still wins outright even when it is a prefix of others, so `"Ventanal"` resolves cleanly alongside `"Ventana Sama"` and `"Ventana cuarto principal"`
- a partial name matching exactly one target still resolves, so `"nevera"` works

Rather than just failing, the tools now return the candidates — `"tomacorriente" matches 5 devices: ... Ask the user which one they meant.` — and the extension instructions tell the assistant to ask instead of retrying, since acting on the wrong match switches real hardware.

Covered by tests that fail against the previously reviewed commit and pass now, including regression guards that exact matches and unique partial matches still resolve.

pernielsentikaer:
Thanks for the thorough pass, @AndresMorelos — this is a real maintenance of the extension, not a drive-by bump.

A couple of changelog items before this goes out:

1. Pinning is now per device, and existing switch pins are not carried over. That is in the PR body but not in `CHANGELOG.md`, which is what users will read after the update. Worth a sentence. The refresh bullet also overstates things: functions are one request per 20 devices, and the device list is still paginated.

2. The old nested commands hardcoded Open/Close/Stop for curtains and the white/colour/scene/music work modes for lights. Those were string enums, and the new actionable list only keeps booleans and numbers, so those controls disappear. If that is intentional, please say so in the changelog; if not, the string-enum actions should come back.

Happy to approve once those are in.

---

AndresMorelos:
Thanks, both fair. Fixed on the branch:

**1.** The changelog now says pinning moved from per switch to per device, and that pins set in an earlier version are not carried over. You are right that the refresh bullet overstated it: it now says instruction sets are fetched in batches of twenty devices, and that the device list itself is still paginated.

**2.** Not intentional, that one was a bug. The actionable filter documented itself as "booleans to flip, numbers to set, enums to choose" and then admitted only booleans and numbers, so every Enum data point fell out. `CommandActionPanel` already had a string branch waiting, just unreachable.

They are back as a submenu built from the range each product declares, read from the same `values` field the brightness range comes from, so nothing is hardcoded per category. Two gates were hiding them as well: "Show Switches" only appeared when a device had more than one boolean switch, so a single switch light could never reach its work modes, and a curtain has no boolean switch at all, so its panel offered nothing but Pin and Show Details. That push is now gated on how many controls a device actually has.

There is a changelog line for the restored controls too.

pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #30582, opened 2026-08-27
Update app-updates extension

gfazioli:
## Description

Maintenance update: moves the extension to `@raycast/api` 2.x and `@raycast/utils` 2.x.

`ray migrate` reports no codemods to apply and no source change was needed — the extension does not call any API introduced in 2.x, so it stays compatible with older Raycast versions. `tsconfig` now sets `module`/`moduleResolution` to Node16 for consistency with the rest of the toolchain, and the dependency bump clears the high-severity advisory the previous tree was reporting.

## Screencast

No user-visible changes: this is a dependency and toolchain update only. Both menu bar commands (update count badge and Brew Maintenance status) render exactly as before — they use built-in `Icon` values, which are unaffected.

## Checklist

### DIFF
--- a/extensions/app-updates/CHANGELOG.md
+++ b/extensions/app-updates/CHANGELOG.md
@@ -1,6 +1,6 @@
 # App Updates Changelog
 
-## [Raycast 2 Compatibility] - {PR_MERGE_DATE}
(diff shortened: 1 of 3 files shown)


pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #30584, opened 2026-08-27
Update dev-cache-cleaner extension

gfazioli:
## Description

Maintenance update: moves the extension to `@raycast/api` 2.x.

`ray migrate` reports no codemods to apply and no source change was needed — the extension does not call any API introduced in 2.x, so it stays compatible with older Raycast versions. The only source diff is formatting: the two menu bar deep links that were rewritten upstream to go through `process.env.RAYCAST_SCHEME` exceeded the print width and are now wrapped by Prettier.

`tsconfig` sets `module`/`moduleResolution` to Node16 for consistency, and the dependency bump clears the high-severity advisory the previous tree was reporting.

**Menu bar deep links.** The two menu bar items opened the extension's own commands by building a `raycast://extensions/Undolog/<extension>/<command>` URL by hand — which hardcodes the author handle, the extension name and the URL scheme, the last of which is exactly what had to be patched here when Raycast 2 changed it. They now go through `launchCommand`, the typed API for this, and failures surface as a HUD instead of an unhandled rejection. (Raised by Greptile on this PR.)

## Screencast

No user-visible changes: this is a dependency and toolchain update only. The menu bar command uses built-in `Icon` values, which are unaffected.

## Checklist

### DIFF
--- a/extensions/dev-cache-cleaner/CHANGELOG.md
+++ b/extensions/dev-cache-cleaner/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Dev Cache Cleaner Changelog
 
-## [Raycast 2 Compatibility] - {PR_MERGE_DATE}
+## [Raycast 2 Compatibility] - 2026-08-27
 
 - Updated to `@raycast/api` 2.x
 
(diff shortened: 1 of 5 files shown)


pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #30587, opened 2026-08-27
Update sync-folders extension

gfazioli:
## Description

Fixes the menu bar icons on Raycast 2, and moves the extension to `@raycast/api` 2.x.

**The bug.** The menu bar command used `extension-icon.svg` as its icon. That file is the wrapper produced by Raycast's icon maker: a nested `<svg>` embedding a base64 PNG, plus a `clipPath` referencing it through `<use xlink:href>`. On Raycast 2 it renders as a solid black square, so the extension showed up as a black block in the menu bar. The file itself is valid — `rsvg-convert` renders it correctly — so this is specific to the menu bar renderer.

**What I found while fixing it.** Custom SVG assets are not a usable replacement either: the menu bar renderer draws SVG *fills* but ignores SVG *strokes*, and a tint is not applied to SVG assets at all. So a stroke-only outline renders as nothing, whatever colour it declares. I verified this three ways — a stroke-based asset using `currentColor`, the same asset with an explicit colour, and a fill-based one tinted with `Color.PrimaryText` — all invisible.

**The fix.**

- the root icon is now the built-in `Icon.TwoArrowsClockwise`, which keeps the two-arrows sync motif of the original icon (and is available since 1.x, so this does not raise the compatibility floor)
- the **Massive Sync** submenu assets were all stroke-only outlines and are now rasterised PNGs. `bulk` gains an `@dark` variant since it had no colour of its own; the three coloured folder icons keep theirs.
- the SVG sources they replace are removed, along with six that were already orphaned before this change, so every file in `assets` is now actually referenced

**Also in this PR.** `@raycast/api` 2.1.0 and `@raycast/utils` 2.3.0 (`ray migrate` reports no codemods; no other source change needed), ESLint migrated to the flat config format, `module`/`moduleResolution` set to Node16, and the `platforms` field plus the real merge date in the changelog restored — both were applied by the reviewers here but never made it back into my local copy. The dependency bump also clears 8 high-severity advisories.

## Screencast

Before: the menu bar showed a solid black square. After: the built-in two-arrows sync icon, which follows the menu bar appearance in both light and dark themes. Verified in Raycast 2.1.0 on macOS.

The **Massive Sync** submenu icons only appear once at least one sync preset exists, and are rasterised versions of exactly the same artwork.

## Checklist


pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #30591, opened 2026-08-27
[Network Diagnostics] Fix Wi-Fi info on macOS Tahoe

### LINKED ISSUE #30583: [Network Diagnostics]
### Extension

https://www.raycast.com/hmarr/network-diagnostics

### Raycast Version

2.0.6.0

### OS Version

macOS Tahoe 26.5.2 arm64

### Description

When opening the app, I get an error that says:

```
Error fetching network info
Command failed with ENOENT:
/System/Library/PrivateFrameworks/Apple80211.framework/Versions/Current/Resources/airport -l
spawn
/System/Library/PrivateFrameworks/Apple80211.framework/Versions...
```

The Wi-Fi information is missing, and the error alert keeps popping up. I’ve tried reinstalling the app and setting the proper permissions, but nothing seems to work.

### Steps To Reproduce

1. Open Raycast with Cmd + Spacebar.
2. Search for “Network Diagnostics.”
3. Press Enter.
4. Raycast opens the extension, which displays the alert.

### Current Behaviour

_No response_

### Expected Behaviour

_No response_

szupzj18:
## Problem

Closes #30583

On macOS Tahoe (26), the `airport` utility was removed, so `wifiInfo()` fails with `ENOENT` and the extension shows a persistent "Error fetching network info" alert.

## Root cause

`src/network-info/wifi.ts` shelled out to `/System/Library/PrivateFrameworks/Apple80211.framework/.../airport -I`, which no longer exists on Tahoe.

## Fix

Read the SSID and RSSI through the CoreWLAN framework instead (via `osascript` JXA: `CWWiFiClient.sharedWiFiClient.interface`). CoreWLAN is available on all supported macOS versions, so no fallback is needed. When the Mac is not associated with a Wi-Fi network, the function still returns `{}` as before.

## Validation

- Verified on a Tahoe machine (macOS 26.4.1): the old `airport -I` invocation fails with `ENOENT` (RED); the new CoreWLAN invocation succeeds and returns `{}` on this Ethernet-only machine without crashing (GREEN).
- Parsing checks: associated-network output (`{"ssid":"HomeNet","rssi":"-62"}` — JXA returns rssi as a string) maps to `{ssid: "HomeNet", rssi: -62}`; empty/missing ssid or rssi maps to `{}`.
- `tsc --noEmit` clean; ESLint and Prettier clean on the changed file.

### DIFF
--- a/extensions/network-diagnostics/CHANGELOG.md
+++ b/extensions/network-diagnostics/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Network Diagnostics Changelog
 
+## [Update] - 2026-08-27
+
+- Fixed Wi-Fi info failing on macOS Tahoe, where the `airport` utility was removed; Wi-Fi details are now read via the CoreWLAN framework.
+
 ## [Update] - 2023-09-22
 
 - Added a new action inside local-network.tsx panel to open router IP address in browser (for accessing admin panel).
--- a/extensions/network-diagnostics/CHANGELOG.md
+++ b/extensions/network-diagnostics/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Network Diagnostics Changelog
 
-## [Update] - {PR_MERGE_DATE}
+## [Update] - 2026-08-27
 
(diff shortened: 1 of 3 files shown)


pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #30598, opened 2026-08-28
Update brew extension

chrismessina:
## Description

Adds Homebrew's install analytics to the extension. The detail sidebar shows a package's install counts and any deprecation warning, and Search gains an optional "Sort by Popularity" that orders results by installs over the last 30 days.

#### With sidebar hidden and stats accessory, sorted by popularity

image

Two data sources, deliberately. The per-package endpoint (~5KB) carries the 30/90/365-day counts plus the deprecation flags, and is fetched only for the selected row — Raycast constructs the detail element for every row in the list, so the fetch is gated on selection. The bulk 30-day rankings (~2.6MB, cached on disk) are only downloaded when the popularity sort is enabled, because ranking has to order the whole matching set *before* the search truncates it to 100 results; otherwise "most popular" would mean "most popular of the first hundred alphabetically".

#### The new toggle sidebar and show description actions

image

#### With sidebar and description visible

image

#### With sidebar visible and description hidden

image

This also moves the extension to `@raycast/api` 2.x. That needed no source changes — nothing here used the aliases v2 deprecates.

### Changes

- Search: "Sort by Popularity" (⇧⌘P) orders results by installs over the last 30 days. An empty query lists the most installed packages outright, and each row shows its install count while the sort is on.
- The detail sidebar shows install counts for 30, 90 and 365 days plus build errors, matching the analytics table on a package's formulae.brew.sh page.
- Deprecated and disabled packages carry a warning at the top of the sidebar, with the reason, the date, and the replacement package where Homebrew names one.
- "Hide Description" (⇧⌘Y) drops the description pane and shows the sidebar as a pure metadata list, with a Caveats indicator. The full caveats text stays in Show Details, which renders it as prose.
- "Show Details" is now offered even when the sidebar is open. It previously disappeared in that state, which left "Show in Finder" as the default action on an installed package. Install and Upgrade remain the default action where they apply.
- Display toggles are grouped in a "View" section. "Toggle Details" is now "Toggle Sidebar" (⇧⌘D) and works in Search as well as Show Installed; "Hide Dependencies" no longer appears in Search, where nothing filtered by it. Show Installed adds an install date and marks pinned formulae.
- Updated to `@raycast/api` 2.x, and to a `brace-expansion` release without GHSA-rgw5-rvv9-x895.

### Notes for review

- `src/utils/brew/analyticsParse.check.ts` is a self-check for the parsing that has no types to protect it — comma-formatted counts, per-invocation buckets, disabled-outranks-deprecated. It is not imported by any command and is not bundled. Run it with:
  `npx esbuild src/utils/brew/analyticsParse.check.ts --bundle --platform=node --outfile=/tmp/check.js && node /tmp/check.js`
- The bulk rankings refresh via a `Last-Modified` check on each launch, so the 2.6MB is paid once per Homebrew republish rather than once per launch.

## Checklist


nhojb (changes requested):
Very nice 😄 

Couple of minor comments:

- Because the install analytics load lazily, the detail metadata scrollview jumps after the analytics load. Perhaps we could add placeholder rows to avoid this?
- The install analytics are missing from the non-detail info view:

Image

chrismessina:
> Very nice 😄

Thanks! :) 

> Couple of minor comments:
> 
> * Because the install analytics load lazily, the detail metadata scrollview jumps after the analytics load. Perhaps we could add placeholder rows to avoid this?
> * The install analytics are missing from the non-detail info view...

Both addressed and pushed (`25c4b35`).

**1. The layout jump.** Root cause was mine: the statistics block returned `null` until the fetch landed, so the rows appeared underneath the cursor. `analyticsRows()` now always returns the three install rows, carrying an em dash until the counts are known — the panel reserves their height and only the *values* change. Build errors stay conditional, since they're absent for nearly every package and a permanent `0` row would be noise.

**2. Missing from the Details view.** Real gap — I'd only wired the sidebar. `Detail.Metadata.Label` and `List.Item.Detail.Metadata.Label` are different components so the JSX can't be shared, but everything deciding *what* to show now is (`analyticsRows`, `packageStatus`, `usePackageDetail`), so the two views can't drift in content. The deprecation warning comes along with it, at the top where it is in the sidebar.

image

chrismessina:
@nhojb since your first pass, my latest batch adds: outdated formulae detection so that available upgrades are called out in the metadata, changed from red to yellow the update icon, condensed the sidebar metadata renderer (which killed three duplicated formatters and `dependencies.tsx`), improved the skip toast and its "Skipped" marker for pinned formulae that shouldn't be upgraded, changed the Pin icon to use Tack, and added a Pinned Formulae section.

file-724e170482abb63cf50afa663f350ffb

file-b1d2ab57ee6e1f4dcde16193696a2cfa

file-60fe0bf8a5b16aa47bfa08018842c00b

file-219ce7723b8a1bac504982c0efab2063

chrismessina:

nhojb (approved):
LGTM 😄 

One minor thing: perhaps move the pinned section to the top in _Show Installed_?

Image

chrismessina:
> One minor thing: perhaps move the pinned section to the top in _Show Installed_?

@nhojb Funny enough — that was my preference, but the previous implementation showed the pinned formulae at the bottom because of the competing definition of "pinning" between Raycast and Homebrew. 

That is, "pinning" in Homebrew means "don't upgrade this formula". 

In Raycast, pinning is used to bookmark or to cause items to rise to the top of lists.

Thus pinning in the context of this extension means **exempting some formulae from being updated**, and therefore it kind of makes sense to anchor this list to the bottom of the lists, especially "Show Outdated", since those items won't be updated. 

Happy to revisit and adjust this behavior — I also suggested changing the action UI copy to use "Lock/Unlock" instead of "pin" given the collision, but Claude pushed back since it would conflict with what Homebrew user are familiar.

Given this, what do you suggest?

nhojb:
> Given this, what do you suggest?

Yeah, good points. Let's just leave it as-is for now. If we get any feedback we can easily move to the top later.

PR #30602, opened 2026-08-28
Add simplebanking extension

klotzbrocken:
## Description

A read-only Raycast front-end for [simplebanking](https://www.simplebanking.de), a German
macOS menu-bar banking app (Developer ID signed and notarized, distributed outside the App
Store). The extension shows what the app has already fetched — balances, transactions and a
monthly overview — so the user does not have to open the app for a quick look.

**The extension never talks to a bank and never sees a credential.** It shells out to the
CLI that ships inside the app bundle and parses its JSON output.

### Security and privacy

This is a banking extension, so the boundaries are worth spelling out:

- **No network access.** The extension contains no HTTP client, no `fetch`, no URLs. Its
  only I/O is `child_process.execFile` against the local binary and reading the resulting
  JSON from stdout.
- **No credentials, no secrets, no preferences.** `package.json` declares no preferences at
  all — there is nothing to store. Bank credentials live in the app, AES-GCM encrypted under
  a master password the extension never asks for and cannot use.
- **No shell.** Commands are invoked with `execFile` and a fixed argument array, never
  through a shell, and no user input is interpolated into arguments.
- **Local data only.** The CLI reads a SQLite cache and macOS preferences that the app wrote
  on its last refresh. Nothing is uploaded, no telemetry, no analytics.
- **It cannot move money.** Initiating a transfer is deliberately not exposed. Payments stay
  in the app, behind the master password and the bank's strong customer authentication.
- **One command does reach the bank, and says so.** `Konten Aktualisieren` asks the running
  app to refresh; depending on the bank that may require an approval in the user's banking
  app. It is a separate `no-view` command and its description states this.

### Requirements

The companion app must be installed. The extension looks for the binary in this order:
`~/.local/bin/sb` (the symlink the app offers to create), `/usr/local/bin/sb`, then directly
inside `/Applications/simplebanking.app`. If none is found it shows a plain error instead of
failing silently — no installation step is required beyond having the app.

### Screencast

The screenshots in `metadata/` were taken with the app's built-in **demo mode**, which
generates fictional banks, balances and transactions. No real account data is shown.

## Checklist


pernielsentikaer (changes requested):
Hi @klotzbrocken 👋

Thanks for contributing simplebanking 💪

I found a few things to address:

- Please translate the extension-owned UI into US English, including actions, placeholders, and errors. Keep the existing command `name` values unchanged.
- Please recapture the screenshots using [Window Capture → Save to Metadata](https://developers.raycast.com/basics/prepare-an-extension-for-store#screenshots). The current images retain the local-extension badge and show outdated command titles.
- The CLI can omit `balance` when no cached balance exists. This currently displays `NaN €` and breaks the combined total. Model it as optional and show an unavailable state:
  ```tsx
  text: k.balance == null ? "Not Available" : euro(k.balance, k.currency)
  ```
  Please also handle missing values in copied text and totals without treating them as zero.
- Monthly Overview currently combines different currencies and displays the result as EUR. Please calculate totals/categories per currency from transaction data, or explicitly restrict the summary to EUR accounts.
- Please address the [refresh-timeout feedback](https://github.com/raycast/extensions/pull/30602#discussion_r3878306114). The CLI already manages its own 180-second timeout, so let it finish rather than terminating it after 120 seconds:
  ```ts
  await run(pfad(), ["refresh"]);
  ```

---

klotzbrocken:
@pernielsentikaer All five points from your review are in (US-English UI, recaptured Retina screenshots, optional `balance`, per-currency Monthly Overview, no extension-side refresh timeout), plus a LaunchServices fallback so the CLI is found wherever the app is installed. Ready for another look whenever you have time, thanks!

PR #30611, opened 2026-08-28
zed-recent-projects: fix PATH missing entries from ~/.zshrc

PushUpek:
Fix for #25787

execWithCleanEnv spawned the user's shell as `-lc` (login, non-interactive). zsh only sources ~/.zshrc for interactive shells, so PATH additions from nvm/pyenv/oh-my-zsh/manual edits (the common case) were dropped, and Zed's formatters etc. only saw the bare system PATH when launched via this extension.

Switch to `-ilc` (interactive + login) so rc files get sourced too, matching what a normal terminal - and Zed itself when launched from Finder - already resolves.

Also extract the `env` argv building into buildCleanEnvArgs() so the flag choice is unit-testable without spawning a real shell, and add a regression test asserting -ilc is used.

## Checklist

### DIFF
--- a/extensions/zed-recent-projects/CHANGELOG.md
+++ b/extensions/zed-recent-projects/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Zed Recent Projects Changelog
 
-## [Fix Missing PATH Entries from ~/.zshrc] - 2026-08-28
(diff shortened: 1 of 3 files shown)


pernielsentikaer (changes requested):

pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #30619, opened 2026-08-28
Add emotecast extension

Kan-A-Pesh:
## Description

Emotecast searches Twitch emotes and pastes them into whatever you were typing in, animation included.

- `⏎` pastes at **32px**, roughly emoji size
- `⇧⏎` pastes at **128px**, roughly sticker size

Sources are picked from a dropdown: **7TV** (default), **BetterTTV** and **FrankerFaceZ**. Static and animated emotes are both listed, `⌘⇧A` narrows to animated ones, and NSFW emotes stay behind a placeholder until selected.

Emotes are pasted as image **files** rather than links, so the animation survives and the recipient sees the emote instead of a URL.

**Two details worth knowing during review:**

Variants are chosen by the **height measured in the downloaded bytes**, not by the scale name or the dimensions the API reports. Providers disagree on both: 7TV serves 32/64/96/128, BetterTTV 28/56/112, and FrankerFaceZ scales its own declared height. Sampling 12 emotes per provider, only 11/12 of BetterTTV and 10/12 of FrankerFaceZ report their real height — `pepeJAM` claims 112px but is 108px. Measuring is what keeps the output exact.

7TV needs no external tool, since it already serves GIF at exactly the two target heights. The other two require re-encoding: `ffmpeg` for GIF and PNG, and `ImageMagick` for FrankerFaceZ's animated WebP, which most ffmpeg builds cannot decode. When a tool is missing the command shows the matching `brew install` line instead of failing silently.

## Screencast

Searching 7TV, with the paste action showing in the footer:

7TV grid

Switching provider from the dropdown:

Source dropdown

Both paste sizes and the animated-only filter:

Action panel

## Checklist


pernielsentikaer (changes requested):
Hi @Kan-A-Pesh 👋

Thanks for contributing! We already have [7TV Emotes Search](https://www.raycast.com/Paranjayy/seventv-search) in the Store by @Paranjayy, which searches 7TV emotes and pastes them as files.

Your unified provider search, sizing, and animation handling add useful functionality, but we’d prefer bringing compatible improvements into existing extensions. Would you be open to contributing there instead? If combining these workflows would be impractical, could you explain why? 🙂

---

PR #30620, opened 2026-08-28
feat: add Fable support and per-model usage limits

bendrucker:
## Description

Adds Fable in the two places the extension hardcodes a list of Claude models.

Model Statistics matched on opus, sonnet, and haiku, so `claude-fable-5` landed in Unknown. Fable gets a new Frontier tier above Premium rather than joining it, since it costs roughly twice what Opus does per token. Tiers also render in a fixed order now instead of however the models arrived.

For usage limits there is no `seven_day_fable` field to add. The live endpoint returns model-scoped windows in a self-describing `limits` array, where each entry carries its own `group` and `scope.model.display_name`, so a model Anthropic adds later needs no extension update. The older `seven_day_sonnet` and `seven_day_opus` fields still supply any window the array does not carry.

Per-model windows used to render only in the menu bar. They now also appear in the Usage Limits detail view and the `get-usage-limits` AI tool, all three sharing one derivation. That fixes the menu bar title under-reporting whenever a model-scoped window was the binding limit.

Fable shows a whole-number percent below while the totals keep one decimal. `limits[].percent` arrives rounded, and `five_hour` and `seven_day` are floats.

The endpoint is undocumented, so the schema stays permissive. An entry naming no model drops its own row instead of failing the parse for the whole payload.

## Screencast

Usage Limits detail view, with the Fable weekly window beneath the 5-hour and 7-day totals:

Usage Limits detail view showing 5-Hour, 7-Day, and Fable Weekly Limit sections

Menu bar Rate Limits section, with the same three windows:

Menu bar Rate Limits dropdown showing 5-Hour at 69 percent, 7-Day at 53 percent, and Fable at 28 percent

Both are captures from a live account with a Fable weekly window.

## Checklist


pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #30621, opened 2026-08-28
Update kobbe extension: Live Visitors menu bar, Top Sources, Setup Health

michael-andreuzza:
## Description

Feature update for the Kobbe extension:

- Add **Live Visitors** menu bar command showing visitors currently online across your sites, with per-site breakdown and 5 minute background refresh.
- Add **Top Sources** command backed by the Kobbe sources endpoint, with referrer favicons.
- Add **Setup Health** command to check tracker installation and revenue webhook status per provider.
- Add a time range switcher (dropdown) to overview, top pages, and sources views.
- Redesign Site Overview with a Live section, tinted metric icons, and a Top Sources section.
- Show site favicons in site lists.
- Use `useCachedPromise` from `@raycast/utils` so commands open instantly with cached data.
- In Search Sites, pressing Enter now opens the in-Raycast overview instead of jumping to the browser.

## Screencast

Updated metadata screenshots included in the PR.

## Checklist


extensions/kobbe/src/api.ts
```diff
@@ -140,14 +85,59 @@ export async function getOverview(siteId: string, range: TimeRange): Promise<Ove
...
+      }
+    }),
+  );
+  return overviews;
+}
+
+function parseCompactNumber(display: string): number {
+  const match = /^([\d.,]+)\s*([KMB])?$/i.exec(display.trim());
+  if (!match) {
+    return 0;
```
pernielsentikaer: [should] Fake zero: the PR deliberately maps failed per-site requests to `online: null` so the menu bar never shows a false count — but an unparseable display string (`"—"`, `"N/A"`, `">5"`) falls through this `return 0` and renders as a real "0 online", also suppressing the `+` indicator. Since you already have the null convention, return `number | null` from `parseCompactNumber` and propagate it — then unknown-vs-zero is consistent end to end.

extensions/kobbe/src/api.ts
```diff
@@ -140,14 +85,59 @@ export async function getOverview(siteId: string, range: TimeRange): Promise<Ove
...
+  const [revenueResponse, overviewResponse] = await Promise.all([
+    getRevenue(siteId, range),
+    getOverview(siteId, range),
+  ]);
+  return { revenueResponse, overviewResponse };
+}
+
+export async function getLiveVisitors(): Promise<KobbeLiveSite[]> {
+  const sites = await listSites();
+  const overviews = await Promise.all(
```
pernielsentikaer: [should] Request fan-out: every 5-minute tick issues 1 + N requests (sites list, then one overview per site in parallel). Fine for a few sites; meaningful load for a 20-site account, on both the Kobbe API and the user's connection. If Kobbe has a batch/live endpoint, prefer it; otherwise cap the number of sites or add a short TTL so a manual refresh doesn't re-fetch everything.

extensions/kobbe/src/views.tsx
```diff
@@ -1,20 +1,35 @@
...
 
+export const RANGE_OPTIONS: Array<{ title: string; value: TimeRange }> = [
+  { title: "Today", value: "today" },
+  { title: "Last 24 Hours", value: "24h" },
+  { title: "Last 7 Days", value: "7d" },
+  { title: "Last 30 Days", value: "30d" },
+  { title: "All Time", value: "all" },
+];
+
+export function rangeTitle(range: TimeRange): string {
```
pernielsentikaer: [nit] `rangeTitle` is exported but never used anywhere in `src/`. Dead code — drop it unless a follow-up needs it.

extensions/kobbe/src/views.tsx
```diff
@@ -23,13 +38,49 @@ function siteCreatedDate(site: KobbeSite): Date | undefined {
...
 
+export function siteIcon(site: KobbeSite): Image.ImageLike {
+  if (site.domain) {
+    return getFavicon(`https://${site.domain}`, { fallback: "extension-icon.png" });
+  }
+  return "extension-icon.png";
+}
+
+function sourceIcon(source: KobbeSource): Image.ImageLike {
+  if (source.source.includes(".")) {
```
pernielsentikaer: [question] What can the `source` field contain? If it's always a bare domain, this heuristic is fine. If it can be a full URL or a dotted label that isn't a hostname, you'll build `https://https://…` and get the generic fallback instead of a favicon — `getFavicon` won't crash (it try/catches internally), the icon just silently won't be right. Worth documenting the assumption or normalizing with `new URL`.

extensions/kobbe/tsconfig.json
```diff
@@ -8,6 +8,7 @@
     "module": "ESNext",
     "moduleResolution": "Bundler",
     "noEmit": true,
+    "skipLibCheck": true,
```
pernielsentikaer, line 11: [nit] `skipLibCheck: true` — I verified the build passes without it. It loosens the type-safety net over dependencies; only keep it if something actually forces it.

pernielsentikaer (commented):
Reviewed at `bf9c141`. Verified locally: `npm run build` and `ray lint` both pass at the head commit; 20 files, +550/−142.

**Overall: approve** — nothing blocking. Two shoulds I'd like addressed, neither merge-gating. Details inline.

- **Shoulds** — `parseCompactNumber` can still produce a fake zero; `getLiveVisitors` fans out to N+1 requests every 5 minutes.
- **Nits** — unused `rangeTitle`; `skipLibCheck` loosening; `sourceIcon` heuristic.
- **Questions** — what can the `source` field contain?

**Praise** — the `online: null` + `+` indicator design in the menu bar distinguishes "couldn't count" from "zero", which most analytics menu bars get wrong; the `hideOverview` guard in `SiteActions` correctly avoids a self-push loop; standardizing on `Keyboard.Shortcut.Common.Refresh` and `useCachedPromise` is a clean consistency win.

extensions/kobbe/src/api.ts
michael-andreuzza: Fixed in 5fed3f2 — `parseCompactNumber` now returns `number | null` and the null flows straight into `KobbeLiveSite.online`, so unparseable display strings render as "Could not load" and trigger the `+` indicator instead of a fake zero.

extensions/kobbe/src/api.ts
michael-andreuzza: Went with the batch option: Kobbe now exposes `GET /api/agent/live` returning online counts for every site in the workspace in one request (deployed to production). The extension uses it first and only falls back to the per-site fan-out — now capped at 20 sites — when a self-hosted server predates the endpoint (404).

extensions/kobbe/src/views.tsx
michael-andreuzza: Removed in 5fed3f2.

extensions/kobbe/src/views.tsx
michael-andreuzza: Good catch — `source` is a sanitized referrer origin like `https://google.com` (protocol + host, set at ingestion) or a label like `Direct / unknown`, so the old heuristic built `https://https://…` and favicons never resolved. Now normalized with `new URL`: the hostname drives both the favicon and the display title, non-URL labels fall back to a globe icon, and Copy Source still copies the raw value.

extensions/kobbe/tsconfig.json
michael-andreuzza: It is actually forced: `tsc --noEmit` (which the Check workflow runs) fails inside `@raycast/utils/dist/types.d.ts` with TS2344 on `UnwrapReturn`'s `ReturnType<ReturnType<T>>` — reproduced on both TS 5.9.3 and 6.0.3. `npm run build` passes because ray/esbuild doesn't typecheck library declarations. The first CI run on this PR failed exactly there before I added `skipLibCheck`; happy to drop it as soon as @raycast/utils ships fixed declarations.

extensions/kobbe/src/api.ts
```diff
@@ -140,14 +86,71 @@ export async function getOverview(siteId: string, range: TimeRange): Promise<Ove
...
+    return response.sites;
+  } catch (error) {
+    // Self-hosted servers predating the batch endpoint return 404; fall back to per-site requests.
+    if (!(error instanceof KobbeApiError && error.status === 404)) {
+      throw error;
+    }
+  }
+  const sites = await listSites();
+  return Promise.all(
+    sites.slice(0, LIVE_FALLBACK_SITE_LIMIT).map(async (site) => {
```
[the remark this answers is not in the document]
michael-andreuzza: Fixed in 9449175e — the fallback now maps every site: the first 20 get real counts, the rest come back with `online: null`, so they stay visible as "Could not load" and the menu bar total shows the `+` truncation indicator instead of silently undercounting.

pernielsentikaer (approved):
Looks good to me, approved 🔥

extensions/kobbe/src/live-visitors.tsx
```diff
@@ -0,0 +1,48 @@
...
+export default function LiveVisitors() {
+  const live = useCachedPromise(getLiveVisitors, [], { keepPreviousData: true });
+  const loadedCounts = live.data?.filter((entry) => entry.online != null);
+  const total = loadedCounts?.length ? loadedCounts.reduce((sum, entry) => sum + (entry.online ?? 0), 0) : undefined;
+  const hasPartialFailures = live.data?.some((entry) => entry.online == null) ?? false;
+
+  return (
+    <MenuBarExtra
+      icon={Icon.Livestream}
+      title={total != null ? `${total}${hasPartialFailures ? "+" : ""}` : undefined}
```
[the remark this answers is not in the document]
michael-andreuzza: Valid catch — fixed in #30639: the menu bar count is now hidden while the command is in an error state, so a displayed number always reflects a successful fetch.

PR #30622, opened 2026-08-28
[Deepcast] Fix close window preference being ignored in default case

### LINKED ISSUE #30609: [Deepcast] Closes on translation even with stay open selected
### Extension

https://www.raycast.com/mooxl/deepcast

### Raycast Version

2.1.1.0

### OS Version

macOS Tahoe 26.5.1 arm64

### Description

When I translate something it copies to my clipboard and closes, with "View in raycast" selected, and "close raycast after translation" turned off

Image

### Steps To Reproduce

1. Translate text, with cmd + enter
2. Watch it disappear and copy to clipboard despite settings saying otherwise 

### Current Behaviour

_No response_

### Expected Behaviour

_No response_

szupzj18:
## Root cause

In commit f5276eb (Preserve hyperlinks in translated text), the `default` case of the `onTranslateAction` switch in `sendTranslateRequest` was changed from `toast.hide(); delayedCloseWindow(...)` to `copyTranslatedText(); showToast(...)`. The `delayedCloseWindow` call was removed, so the `closeRaycastAfterTranslation` preference is no longer respected in the default path.

The `onTranslateAction` dropdown has no `default` value in `package.json`, so when the preference is unset (or not yet saved), `prefs.onTranslateAction` is `undefined` and the switch falls through to `default`. This is the path used by all language-specific no-view commands (e.g. "Translate into Spanish") via the `translate()` helper, which does not pass `onTranslateAction`.

Users report that Raycast closes after translation even with "Close Raycast After Translation" turned off (#30609).

## Change

1. **`src/utils.ts`** — Add `delayedCloseWindow(closeRaycastAfterTranslation)` back to the `default` case, guarded by `onTranslateAction !== "none"` so the main `index.tsx` form (which explicitly passes `"none"`) is unaffected.
2. **`package.json`** — Add `"default": "clipboard"` to the `onTranslateAction` preference so it always has a valid value on fresh installs.

## Validation

- RED→GREEN TDD: wrote a test that calls `sendTranslateRequest` with no `onTranslateAction` and `closeRaycastAfterTranslation: true`, asserting `closeMainWindow` is called. RED (fails before fix), GREEN (passes after fix).
- Regression: `onTranslateAction: "none"` (main form) does not call `closeMainWindow` even with `closeRaycastAfterTranslation: true`.
- Regression: `closeRaycastAfterTranslation: false` does not call `closeMainWindow` in the default case.
- All 15 existing `hyperlinks.test.ts` tests pass.

Fixes #30609


lambegraham:
Thank you @szupzj18 for this, and @pernielsentikaer for reviewing. Using translation 50+ times a day, this got very annoying very quick.

12ian34:
Thank you all for fixing my error - so sorry for the inconvenience :(

pernielsentikaer (approved):
Looks good to me, approved 🔥

lambegraham:
@szupzj18 Unfortunately my issue was not solved by this. Using latest extension and raycast version, fully deleted settings and set up fresh. Same issue, still closes when I translate. #30609

PR #30624, opened 2026-08-28
[Parcel] Fix carrier dropdown failing to load

bendrucker:
## Description

Parcel changed `supported_carriers.json` from `{"ups": "UPS"}` to `{"ups": {"name": "UPS", ...}}`, and the Add Delivery carrier dropdown has failed with `s.name.localeCompare is not a function` ever since. The extension read that payload through an unchecked `as Record<string, string>` cast, so an upstream shape change surfaced as an internal `TypeError` in a user-facing toast. Three other response sites were cast the same way and would have broken the same way on whichever endpoint changed next.

Responses are now parsed with zod at the boundary. Both carrier shapes are accepted, so a revert upstream keeps working.

- Validation fails open per item and closed per envelope. One malformed delivery is dropped and logged rather than blanking the list. An unreadable envelope, or every item failing at once, raises "Parcel returned an unexpected response" with a Retry action.
- Schemas are `looseObject` throughout, because the AI tools spread deliveries and events straight into their output and a stripping schema would silently truncate what the assistant sees.

I have not run the distribution build in Raycast, so that checklist item is unticked.

## Checklist

### DIFF
--- a/extensions/parcel/CHANGELOG.md
+++ b/extensions/parcel/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Parcel Changelog
 
+## [Fix Carrier Loading] - {PR_MERGE_DATE}
+
+- Fixed the "Add Delivery" carrier dropdown failing to load after a change to Parcel's carrier list format
+- Carrier names display in "My Deliveries" again instead of falling back to carrier codes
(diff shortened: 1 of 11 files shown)


pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #30625, opened 2026-08-28
[Things] Fix list membership under Things 3.23

bendrucker:
## Description

Things 3.23 changed two things the extension's optional database path relied on. A to-do keeps its "scheduled" marker after its date arrives and Things rolls it into Today, so list membership now has to follow the date rather than that marker. And while a repeating to-do has an open occurrence, Things clears the template's next-occurrence date and computes it internally, so Upcoming has to show those templates without a date, the way Things itself does.

I checked this by diffing the database path's list membership against what Things reports over AppleScript, on a real database. Against 3.23 the old queries left Today short by 6 items and Upcoming short by 31; with these changes all five lists match Things exactly.

Two fixes came out of that comparison rather than from 3.23 itself. To-dos inside a trashed project were leaking into the lists, Quick Find, and search, because trashing a project leaves its to-dos untouched in the database. That applies to the Logbook too: Things omits those to-dos from its own Logbook, so the query filters them the same way. Separately, Things silently ignores an attempt to complete a repeating to-do's template, which the extension reported as success, so the write now reads the value back and surfaces a real error.

This all sits behind the Unofficial API preference, which is off by default. The only change on the default AppleScript path is that status read-back, which also gets error handling in the menu bar command where it was missing.

## Checklist

### DIFF
--- a/extensions/things/CHANGELOG.md
+++ b/extensions/things/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Things Changelog
 
-## [Things 3.23 Compatibility] - {PR_MERGE_DATE}
+## [Things 3.23 Compatibility] - 2026-08-29
 
 - Fixed Today and Anytime missing to-dos whose scheduled date had arrived when the Unofficial API preference is enabled. Things 3.23 leaves such a to-do marked as scheduled, so the extension has to decide list membership by date.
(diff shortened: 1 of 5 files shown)


pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #30629, opened 2026-08-29
fix: make Select Model apply to the running Handy app

### LINKED ISSUE #29966: [Handy] Select Model command not working
### Extension

https://www.raycast.com/mattiacolombomc/handy

### Raycast Version

1.104.24

### OS Version

macOS 26.5.1

### Description

The select model command of the handy extension is not working. It shows the available models correctly. However, when you switch the model and the extension says the model has been successfully selected in the handy app, still the previous model shows as selected. So the selection did not work.

Handy version: 0.9.4

### Steps To Reproduce

1. Open Select Model command
2. See correct list of available models
3. Select another model, and activate it by pressing Enter
4. Check Handy app - the model has not actually changed.

### Current Behaviour

The actual model selection is not changed in Handy app.

### Expected Behaviour

_No response_

Official-Krish:
Fixes #29966 

## Description

The handy extension's Select Model and Select Language commands only wrote settings_store.json, but the running Handy app never picked up the change until it was relaunched — so selecting a model appeared to "do nothing." 

## The problem (why)

- Handy is a single-instance Tauri app that loads settings_store.json into memory at launch and never re-reads it at runtime.
- There is no external channel to switch the model: no --set-model CLI flag, and no custom URL scheme (the Info.plist has no CFBundleURLTypes; setURLSchemeHandler: is WebKit internal IPC).
- The only way the active model changes in-process is Handy's own set_active_model Rust command, which the in-app tray menu calls. So writing the file alone was a no-op until a relaunch.

## The fix

- New src/lib/handy.ts — drives Handy via AppleScript GUI scripting:
- selectModelInHandy(name) / selectLanguageInHandy(label): activate Handy, then click the tray menu item whose title contains the target name. Handy exposes its model list under a submenu titled with the currently active model (not a fixed "Model" item), so the script searches every tray submenu for the target and clicks it — triggering Handy's in-app set_active_model.
- applySettingsAndReload(apply, binaryPath): the fallback when GUI scripting isn't possible (Accessibility not granted, hidden tray, or item not found). It persists the setting, quits Handy, waits for exit, persists again (to survive Handy's shutdown flush), then relaunches.
- select-model.tsx / select-language.tsx — handleSelect now: skips if already active; persists via writeSettings; attempts the live GUI switch; on success shows an HUD; on failure shows a Failure Toast with a "Restart Handy" primary action that calls applySettingsAndReload.
- settings.ts — hardening so the setting is actually honored/reloadable:
- readStore now tolerates a missing file (ENOENT → {}) instead of throwing.
- readSettings handles a missing settings key and defaults selected_model → "", selected_language → "auto".
- writeSettings always keeps the {"settings": {...}} envelope (even when the file/key was absent) and mkdirSyncs the directory first.

## Files changed

1. extensions/handy/src/lib/handy.ts	(new) - GUI-script + relaunch helpers
2. extensions/handy/src/select-model.tsx - live switch + restart fallback toast
3. extensions/handy/src/select-language.tsx - live switch + restart fallback toast
4. extensions/handy/src/lib/settings.ts - ENOENT/missing-key defaults, envelope safety
5. extensions/handy/tests/lib/settings.test.ts - tests for new hardening; replaced "throws if missing" with "reads defaults if missing"

## Checklist

### DIFF
--- a/extensions/handy/CHANGELOG.md
+++ b/extensions/handy/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Handy Changelog
 
-## [Fix live model & language selection] - {PR_MERGE_DATE}
+## [Fix live model & language selection] - 2026-09-01
 
 ### Fixed
 
--- a/extensions/handy/CHANGELOG.md
+++ b/extensions/handy/CHANGELOG.md
@@ -4,8 +4,8 @@
 
 ### Fixed
 
-- **Select Model** now actually switches the active model in the running Handy app, instead of only writing `settings_store.json` (which Handy only reads at launch, so the selection did nothing until a relaunch). The extension GUI-scripts Handy's tray model menu — the same path the in-app menu uses — so the model changes instantly with no restart. If Raycast lacks Accessibility permission, a "Restart Handy" toast offers the relaunch fallback.
-- **Select Language** now persists the choice and uses the same live-switch path, falling back to a relaunch when GUI scripting isn't available.
+- **Select Model** now actually switches the active model in the running Handy app, instead of only writing `settings_store.json` (which Handy only reads at launch, so the selection did nothing until a relaunch). The extension GUI-scripts Handy's tray model menu — the same path the in-app menu uses — so the model changes instantly with no restart. If Raycast lacks Accessibility permission, a "Restart Handy" toast offers the relaunch fallback. The store is updated only after Handy accepts the live switch (or after a successful restart), so a failed switch stays retryable.
+- **Select Language** now uses the same live-switch path and persist-after-success ordering, falling back to a relaunch when GUI scripting isn't available.
 
 ### Changed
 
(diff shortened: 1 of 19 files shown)


0xdhrv (changes requested):
The live-switch plus restart-fallback approach is reasonable, and the earlier process-matching, exact-menu-selection, and shutdown-handling issues look addressed.

Requesting a change for the persisted vs running-state mismatch in `handleSelect()`.

Both Select Model and Select Language write `settings_store.json` and update local "current" state *before* we know whether the running Handy app applied the change. Handy does not re-read that file while running, so a failed GUI-scripted switch that the user dismisses leaves Raycast believing the new selection is already active while Handy is still using the old one. That blocks retry and the restart fallback — the same class of discrepancy this PR is meant to fix.

Please persist (and mark active) only after a successful live switch, or otherwise keep a failed switch retryable until Handy has actually reloaded the setting.

0xdhrv (approved):
Looks good to me, approved ✅

PR #30631, opened 2026-08-29
Update resend extension

0xdhrv:
## Description

This update expands the Resend extension from its existing sent-email and legacy audience workflows into broader email, contact, and developer-resource management in Raycast and Raycast AI.

### Commands and workflows

- Add a **Received Emails** command for reading inbound messages and opening their attachments.
- Enhance **Emails** with expiring share links and scheduled-email cancellation.
- Move contact browsing and membership workflows from deprecated audiences to segments, while retaining the legacy audience tool for compatibility.
- Fix contact update/removal behavior so contacts can be identified by current Resend contact IDs or email addresses.

### AI tools

- Add tools for received email and attachments, temporary share links, email metrics, rescheduling, domains, templates, broadcasts, segments, topics, webhooks, and API logs.
- Move AI instructions and evals into `ai.yaml`, with confirmations for sending, sharing, rescheduling, cancellation, deletion, API-key creation, contact mutations, and domain verification.
- Expand email sending to support text and HTML bodies, CC/BCC/reply-to, custom headers, topics, scheduled delivery, idempotency keys, local files, and hosted attachments.

### Safety and reliability

- Show each local attachment path or hosted URL before an AI-initiated send, require absolute local paths, and accept hosted attachments only over HTTPS.
- Resolve a contact's authoritative email address before confirming deletion by contact ID.
- Reject conflicting pagination cursors, empty cursors, and limits outside Resend's supported `1–100` range.
- Update the Resend SDK, README, manifest metadata, and changelog for the expanded API coverage.

### Validation

- `npm run lint`
- `npx tsc --noEmit`
- `npm run build`
- `git diff --check`
- Focused pagination and attachment-safety assertions
- `npx ray evals`: latest full run passed 19/20; the remaining failure was a nondeterministic extra read-only list-tool call without corresponding mock data

## Screencast

Not attached yet. Manual Raycast verification and visual capture of the new **Received Emails** command and updated confirmation flows are still pending.

## Checklist

- [ ] I ran `npm run build` and [tested this distribution build in Raycast](https://developers.raycast.com/basics/prepare-an-extension-for-store#metadata-and-configuration) (`npm run build` passes; manual Raycast verification is pending)


pernielsentikaer (changes requested):
Awesome update, thanks for the push @0xdhrv 

A few things we could improve:
 
1. `createShareLink` / `cancelScheduledEmail` in `src/emails.tsx` handle `{ error }` from the SDK but don't catch a thrown request, so a network failure can leave the Animated toast up. A try/catch + `showFailureToast` (same pattern as `EmailSend`) would close that. Share/cancel also don't appear on the email detail screen.

2. Contacts cmd+D is now “Remove From Segment”, and there is no remaining command action to delete a contact (only the AI `remove-contact` tool). The confirmation copy is clear, so this may be intentional — if account-level delete should still live in the command, a separate Delete Contact action would avoid the old muscle-memory trap.

3. `send-email` accepts custom headers but the confirmation doesn't list them. Worth showing, since headers can change routing.

Noto sure if we can fix them all?

---

pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #30641, opened 2026-08-29
fix(virustotal): accept scheme-less URL input

DhanushRaj5297:
## Description

URL input without a scheme (e.g. `evil.com`, `evil.com/path`) was rejected with "Invalid input format" because `new URL()` requires a scheme, and only a scheme-qualified string (`https://evil.com`) would parse successfully.

The fix accepts scheme-less input by also checking it with an `https://` prefix, but **submits the original, unmodified input** to VirusTotal's `/urls` endpoint. It intentionally does not rewrite the input to add a scheme.

**Why:** VirusTotal's canonicalization only normalizes "minor aspects, like some escaped characters" — it does not insert a missing scheme. Since the URL identifier (`url_id`) is derived from the submitted string, `google.com` and `https://google.com` resolve to different IDs and are tracked as separate resources with independent detection stats and submission history. Silently rewriting user input to add `https://` would have shown a report for a different resource than what the user actually typed.

**Sources:**
- https://docs.virustotal.com/reference/url
- https://docs.virustotal.com/reference/domains-object

## Screencast
N/A - bug fix to existing validation logic in an existing command, no new extension/command added.

## Checklist

### DIFF
--- a/extensions/virustotal/CHANGELOG.md
+++ b/extensions/virustotal/CHANGELOG.md
@@ -1,5 +1,9 @@
 # VirusTotal Changelog
 
+## [Fix] - {PR_MERGE_DATE}
+
+- Fixed URL input without a scheme (e.g. `google.com`, `google.com/path`) being rejected as invalid; it's now accepted and submitted to VirusTotal unmodified
+
 ## [Add IP Address Support] - 2026-01-12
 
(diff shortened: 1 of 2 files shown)


pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #30646, opened 2026-08-30
Add Affipill extension

DerMatte:
## Description

Affipill is a local affirmation-track player for Raycast.

- Browse, add, edit, and play affirmation audio in the background
- Import multiple audio files and covers at once, with optional Raycast AI title cleanup and artwork matching (Raycast Pro)
- Dedicated **Add New Tracks Via AI** command
- **Stop Affipill Playback** command
- **Now Playing** menu bar item while a track is playing
- **Library Folder** preference to choose where audio and covers are stored (existing files move on the next open)

Tracks and covers are stored locally by the extension. No third-party API keys are required. AI import uses the built-in Raycast AI API.

## Screencast

https://github.com/DerMatte/raycast-extensions/releases/download/affipill-screencast/affipill-screencast.mp4

## Checklist


DerMatte:
Screencast for review:

https://github.com/DerMatte/raycast-extensions/releases/download/affipill-screencast/affipill-screencast.mp4

DerMatte:
Addressed the two remaining Greptile findings in 80b43536d:

- Playback ownership now includes the `afplay` process start time from `ps -o lstart=`, so a reused PID for another `afplay` of the same file is not killed.
- A failed `ps` lookup is treated as unknown: we keep the stored handle and skip both the kill and the state clear, so Stop/Now Playing can retry.

DerMatte:
Addressed the remaining Greptile finding in beec85cd:

- `stopPlayback` now throws when process ownership is unknown, so Stop reports failure instead of success while audio continues.
- `playTrack` will not start another `afplay` or overwrite the stored handle in that case. The handle stays so the next stop can retry.
- Also formatted `audio.ts` for the Prettier 3.9.4 Check failure.

DerMatte:
Addressed the latest Greptile finding in f20ac7a30:

- Ownership is re-checked immediately before SIGKILL, so a PID that is no longer the stored `afplay` process is not killed.
- Stop now kills only that PID, not the process group.

0xdhrv (commented):
Hey @DerMatte 👋

- **Now Playing** is a generic command name. Could you add `"subtitle": "Affipill"` so it is clear in root search? That follows the [command naming guidelines](https://developers.raycast.com/basics/prepare-an-extension-for-store#extensions-and-commands-naming).
- `FilesPickerForm` in `src/import-folder.tsx` sets `navigationTitle` to `"Add Tracks Via AI"` when this is the root **Add New Tracks Via AI** command. Root commands should keep the title from `package.json`. Could you drop `navigationTitle` on that root path? Nested **Import Tracks** from the library can still set it.
- `metadata/add-tracks-via-ai-1.png` is an empty file picker. Could you recapture it after matching files, so the Store listing shows the AI import flow?

---

PR #30648, opened 2026-08-30
Add Signal 500 extension

gableend:
## Description

Search the Signal 500 from Raycast: 500 hand-scored news and expert sources, ranked on editorial quality, ad restraint, signal-to-noise and independence, with the methodology published at https://feeds.bar/signal-500/method/. Fuzzy search with a pillar filter; actions open a source's Signal 500 page or its own site, or copy its feed URL for any RSS reader. Data comes from the same public API the feeds.bar site is built from, cached via useCachedPromise.

I'm the developer of FeedsBar, the app the catalogue was built for; the extension itself is free and works for anyone.

## Screencast

Tested live in Raycast 2026-08-30: list renders 499 sources with icons/scores, search narrows correctly, Open Signal 500 Page opens the right per-source URL.

## Checklist


gableend:
Both automated-review findings addressed in the latest commit: the browser actions are now grouped (Browse the Full Signal 500 sits with the other OpenInBrowser actions, copy actions follow as one block), and metadata/ now carries three 2000x1250 screenshots of the command running: the ranked list, a search, and the action panel.

0xdhrv (commented):
Hey @gableend 👋

- `package.json` does not declare `platforms`, so this installs as macOS-only. The command only fetches `https://feedsbar-edge-api.netlify.app/...` and opens or copies URLs. Could you add `"platforms": ["macOS", "Windows"]`?
- `src/search-signal-500.tsx` does not use the `error` from `useCachedPromise`, and there is no `List.EmptyView`. A failed fetch or empty catalogue would show a blank list. Could you surface the failure and an empty state?
- `assets/extension-icon.png` is a dark mark. In dark appearance it is hard to read at list size. Could you lighten it or add a theme variant so it stays clear in both appearances?

---

gableend:
Thanks @0xdhrv, all three addressed in 
ac72c813:

- `package.json` now declares `"platforms": ["macOS", "Windows"]`. The command only fetches the public feeds.bar endpoint and opens or copies URLs, so nothing is platform-specific.
- `useCachedPromise`'s `error` is surfaced: a failed fetch renders a `List.EmptyView` with the message and a Try Again action, and an empty catalogue or a pillar with no matches gets its own empty state instead of a blank list. Cached data still shows if a refresh fails.
- Added `assets/extension-icon@dark.png`, a light variant of the mark for dark appearance.

`ray lint` and `ray build` are green. Marked ready for review.

gableend:
Good catch on the shortcuts: all three now carry both forms (`macOS` cmd / `Windows` ctrl) in 990e850b. Lint and build green.

0xdhrv (approved):
Looks good to me, approved ✅

PR #30656, opened 2026-08-31
Update ssh-image-drop extension

free4416:
## Description

Adds a **Send Clipboard to Clipboard** command to the existing `ssh-image-drop` extension.

macOS Screen Sharing has its own clipboard sync, but it drops text often enough that pasting into a remote session is unreliable. This command bypasses that path: it pushes the clipboard over SSH straight onto the remote Mac's pasteboard, so `⌘V` in the Screen Sharing window just works.

The existing **Send Clipboard Image** uploads a file and hands back a remote *path*. This one leaves no file behind — it places the *content itself* on the remote clipboard.

Highlighted text in the frontmost app wins; otherwise the clipboard (text, then image). Every toast and HUD names the source and size — `Selection (69 B) → host` — so a mis-route is visible immediately.

The **remote server must be macOS**, and the SSH user must be the one logged in at that Mac's screen, since `pbcopy`/`osascript` are macOS-only. Both conditions are checked before any data is sent, each with its own message. This is documented in the README; the extension's other commands still support Linux servers.

One drive-by fix: `Clipboard.readText()` returns a placeholder like `Image (321x97)` for image-only clipboards, so macOS now reads the clipboard through `pbpaste` directly. This also fixes **Pull File from Server** misreading an image clipboard as a remote path.

## Checklist


free4416:
Closing this — withdrawing the submission.

The command worked around Screen Sharing's clipboard sync being unreliable,
but it turns out Screen Sharing's built-in clipboard sharing handles this
fine. The premise no longer holds, so there's no reason to add the command.

Sorry for the noise, and thanks for the review time.

PR #30668, opened 2026-08-31
Add gam-tools extension

jdietz02:
## Checklist


0xdhrv (commented):
Hey @jdietz02 👋

- `runGam` in `src/utils/gam.ts` runs `` `"${gamPath}" ${args}` `` through `child_process.exec`. Create/update user then interpolates form values into that string, including `password "${values.password}"`. The same pattern is in `runCommandInTerminal`. A name, OU, or password with a quote or shell metacharacter would execute as part of the shell command, not as a GAM argument. Could you spawn GAM with an argv array (`shell: false`) and pass each value as its own argument?

- That same `console.log` of the full command also writes the plaintext password into the Raycast logs. Could you keep secrets out of logs?

- `package.json` declares `["macOS", "Windows"]`, but install and long-running commands go through `osascript` + Terminal, `#!/usr/bin/env zsh`, and Homebrew/`~/bin/gam` paths. Could you set `"platforms": ["macOS"]` until a Windows path is implemented and tested?

- The three commands are view commands, but there is no `metadata/` folder. Could you add Store screenshots with Window Capture → Save to Metadata (2000×1250)?

- The manifest says MIT, but there is no `LICENSE` file. Could you add one?

- `package.json` is missing `categories`. Could you add at least one Store category?

- Command titles currently say `Manage Goolge Groups` and `Manage Googgle Drive`, and the group description has `Grous`. Could you fix those?

---

Issue #30674, opened 2026-08-31
[Set Audio Device] I'd like to be able to use something like the LEADER KEY app to set volume


admdrzc:
### Extension

https://www.raycast.com/benvp/audio-device

### OS Version

macOS Tahoe 26.6.0 arm64

### Description

I'd like to be able to use something like the LEADER KEY app to set volume. 

Basically I need a param on the deeplink that when executed will just set the volume silently. 

Here's an example from the Caffeine deeplink: raycast://extensions/mooxl/coffee/caffeinateUntil?arguments=%7B%22time%22%3A%225:11pm%22%7D

So something like raycast://extensions/benvp/audio-device/set-volume?arguments=20

### Who will benefit from this feature?

Fewer clicks! I guess alternatively there could be a Command that sets a specific volume and one could record a hotkey. 🤷

joshmedeski:
It's a smart idea, I don't think it will take much effort to implement.

Are you going to open a PR @admdrzc?

admdrzc:
Ah, I wasn’t planning on opening a PR, was mainly suggesting the idea hoping the owner/maintainer might be interested in implementing. 😃

PR #30678, opened 2026-09-01
Audio Device: Add optional volume argument for setting volume silently

joshmedeski:
## Description

Adds an optional `volume` argument to the **Set Output Volume** and **Set Input Volume** commands, so a specific level can be set silently — without opening the form — from a hotkey, a Quicklink, or a deeplink.

This implements the request in #30674: the reporter wants to drive volume from a launcher such as Leader Key, and [confirmed](https://github.com/raycast/extensions/issues/30674#issuecomment-5485701830) they weren't planning to open a PR themselves.

```
raycast://extensions/benvp/audio-device/set-volume?arguments=%7B%22volume%22%3A%2220%22%7D
```

Behavior:

- **Argument supplied** — the level is applied to the current default output/input device and confirmed with a HUD. Values are clamped to 0–100, and a non-zero level unmutes the device first, matching what the form already does on submit.
- **Argument empty** — the form opens exactly as before, so nothing changes for existing users.
- **Non-numeric input** — a failure toast explains that a number between 0 and 100 is expected.

The silent path reuses the existing `ioConfig` in `volume-form.tsx`, so output and input share one implementation.

## Checklist

- [x] I ran `npm run build` and [tested this distribution build in Raycast](https://developers.raycast.com/basics/prepare-an-extension-for-store#test-the-extension-in-raycast)

### DIFF
--- a/extensions/audio-device/CHANGELOG.md
+++ b/extensions/audio-device/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Audio Device Changelog
 
+## [Update] - {PR_MERGE_DATE}
+
+- Added an optional `volume` argument to the Set Output Volume and Set Input Volume commands, so a volume can be set silently from a hotkey or deeplink (e.g. `raycast://extensions/benvp/audio-device/set-volume?arguments=%7B%22volume%22%3A%2220%22%7D`)
+
 ## [Fix] - 2026-08-21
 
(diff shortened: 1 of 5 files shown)


0xdhrv (commented):
Hi @joshmedeski 👋

Thanks for the contribution 🔥

---

0xdhrv (approved):
Looks good to me, approved ✅

PR #30689, opened 2026-09-01
Add yatt extension

alexbartok:
### Yet Another Timezone Tool

New extension. Type a time or a time window; see it in every configured location.

Typical use: someone in New York asks "would 3 to 5 my time work?". Type `15-17 nyc`, read your own hours off the screen, press ⌘Enter and paste the result into the chat:

```
15:00–17:00 NYC (EDT) = 12:00–14:00 SFO (PDT) / 19:00–21:00 UTC / 21:00–23:00 MUC (CEST)
```

Expressions can mix a time, an end time or duration, a date and a place, in any order:

```
15-17 sf          15:00 to 17:00 San Francisco time
tomorrow 9 utc    tomorrow at 09:00 UTC
thu 17 et         next Thursday, 17:00 Eastern
9 sf 90m          a 90-minute window
lon               current time in London
```

**Commands**

- **Convert Time**: one row per location with time, business-hours colour, zone abbreviation and offset. Enter copies a row, ⌘Enter copies all rows as one line, ⌘I shows an hour strip, ⌘←/→ move the time.
- **Manage Locations**: 140,000 places (GeoNames, UN/LOCODE) with their airport and UN/LOCODE codes, all IANA zones, custom places, optional OpenStreetMap lookup. Aliases, export and import.
- **Menu Bar Clock**: selected locations in the menu bar.

**Compared with the existing timezone extensions**

- Time Zone Converter also accepts typed times and ranges; it runs as a form and pastes one line. This extension keeps the result on screen as a list, with colours, offsets, an hour strip and keys to nudge or re-anchor the time.
- In the Timezone and Time-Traveling World Clock select the time by scrubbing; Team Time and World Clock show the current time.
- Places are matched against the user's own list first (names, codes, aliases), then a bundled dataset, so it works offline.

**Notes**

- Place data (9 MB JSON) is generated by `scripts/generate-data.ts` and loaded only when a typed place is not in the list. Licences in `NOTICE.md`, shipped in `assets/`.
- `Intl.DateTimeFormat` only; DST edge cases are tested.
- OpenStreetMap lookup: identifying User-Agent, one request per second, only after a local miss.
- Dependencies: `@raycast/api`, `tz-lookup`.

## Screencast

https://github.com/user-attachments/assets/83da00ba-0071-4eb2-a0da-235b5224e97c

## Checklist


pernielsentikaer (changes requested):
Hi @alexbartok 👋

Thanks for the detailed contribution 💪

We already have [Time Zone Converter](https://www.raycast.com/alex_jones/time-converter) in the Store by @alexj, covering natural-language time and range conversion across saved locations with copyable output. There’s also substantial overlap with the Store’s world-clock and time-scrubbing extensions, as well as the queued [Timezone Travel PR](https://github.com/raycast/extensions/pull/30827).

Your location management, working-hours visualization, richer parser, and menu-bar workflow add useful capabilities. Could you look into contributing those improvements to an existing extension instead? If consolidation isn’t feasible, could you explain the technical or workflow differences that require a separate Store listing? 🙂

---

alexbartok:
Hi @pernielsentikaer 👋, thanks for the review, and for listing the neighbours.

I went through all of them before replying: the Time Zone Converter source, the Timezone Travel PR, In the Timezone, Time-Traveling World Clock, World Clock, Team Time and Timezone Converter. On paper they overlap with YATT a lot. In use they don't, and the difference is in what you type and what you get back.

**The job**

Someone in New York asks on a call: "would 3 to 5 my time work?" You type `15-17 nyc`, read your own hours off the list, and paste the whole line into the chat. The place is part of what you type, the anchor changes with every query, and you get a window, not a single instant.

**How the neighbours take input**

| | Time Zone Converter | Timezone Travel (PR) | In the Timezone, Time-Traveling World Clock | World Clock, Team Time | YATT |
|---|---|---|---|---|---|
| Input | A form: time field, locations field, submit | A time in one fixed anchor city | Arrow keys move the current time; nothing to type | Current time only | Time, window, date and place, in any order |
| Zone of the typed time | The Mac's local zone. `3pm ET` works through chrono-node (undocumented, fixed offsets); `3pm New York` is quietly read as local time | The anchor set in Manage Cities | n/a | n/a | Any location, code, abbreviation or zone in the expression; ⌘⇧Enter makes any row the anchor |
| Windows | Yes | No | No | No | Yes, with `+1d` markers and an hour strip |
| Places | About 350 hand-written aliases plus IANA names | A short hand-written city list | IANA zone IDs, about 600 | TimeAPI lookup | 140,000 places with IATA and UN/LOCODE codes, built from GeoNames, plus your own aliases |
| Result | Pastes a string and closes | Live list | Live list | List | Live list; copy one row or all, with a template |
| Menu bar | No | No | No | Yes | Yes, same list |

**Why this can't be a contribution**

Time Zone Converter is a form. You submit, it pastes, it closes. That's a good one-shot tool, and people who use it depend on that flow. YATT is a live list that recalculates on every keystroke. To bring YATT's behaviour there I'd have to replace the form, the parser (chrono in local time versus a parser that knows places), and the location data (a 350-line alias file versus a generated dataset with a build script and licence notices). Then I'd add location storage, a shared locations file and a menu-bar command. Every file would change. Your guidelines say to make a new extension when "the changes to an existing extension would be significant and might break other people's workflows", and that significant changes need the author's sign-off. That's this case.

Timezone Travel is the closest in spirit: an anchor, working hours, cities sorted west to east. But it's a queued PR from another author, not in the Store yet, and it works on one instant with a fixed anchor and a short city list. The same rewrite argument applies.

The scrubbing and world-clock extensions answer "what time is it there now, and an hour from now?". YATT answers "they said 3 to 5 their time, so what is that for me and for everyone else?".

That's the case for a separate listing.

pernielsentikaer (changes requested):
Hi @alexbartok 👋

Thanks for the detailed comparison. The query-specific anchor and live meeting-window workflow are useful differences.

Could these be added as a separate command in [Time Zone Converter](https://www.raycast.com/alex_jones/time-converter) by @alexj, preserving its existing form command? That wouldn’t require replacing the workflow its users depend on. Is there a technical reason that approach wouldn’t work? 🙂

alexbartok:
Hi @pernielsentikaer, thanks for narrowing it down to one question, that made it easier to check properly.

The honest answer is no, a separate command wouldn't work, and I'd like YATT reviewed as its own listing. The command can't be separated from the other two, the three of them share no code with Time Zone Converter, and moving them would change that extension's description and hand its author a codebase larger than their own. Details below.

**Why it is three commands, not one.** The live list reads from the location store (home, order, aliases, custom places, optional shared file). Manage Locations writes that store and Menu Bar Clock reads it, so all three would have to move together. They bring 13 preferences with them, 8.4 MB of generated data in three JSON files, the build script that makes those files from GeoNames, UN/LOCODE, OurAirports and tzdb, a NOTICE file that has to ship because the GeoNames and OpenStreetMap licences require attribution, and 600 lines of tests.

**Why nothing is shared.** Time Zone Converter is 1,150 lines in six files, with one command, two preferences and no data files. It parses with chrono-node, formats with date-fns-tz, resolves places through a 351-entry alias map, and sits on the 1.86 API. YATT has its own parser, formats with Intl, resolves through the dataset, and sits on 2.x. I went through the converter's source looking for anything the two could reuse and found nothing. Two extensions would be sitting in one folder under one icon.

**What the converter's users would see.** A Default Locations preference for the form and a Manage Locations command for the list, and the two wouldn't know about each other. `3pm ET` resolving through the alias map in one command and through the dataset in the other. A menu bar item and a file-picker preference in an extension whose description is "Convert times between different time zones and paste the results". The guidelines allow adding a command when it "compliments an existing extension without changing the extension title or description". This one would change the description.

**What @alexj would take on.** A data pipeline, licence notices, a menu bar command and a locked shared file store, none of it written by them. Since the guidelines want the author's sign-off on anything significant, every later YATT change would go through them as well.

The guidelines have a case that fits: "If your extension works standalone and brings something new to the Store, it's acceptable to create a new one instead of adding commands to an existing one." YATT runs on its own, and the table in my last reply covers what's new.

pernielsentikaer:
Thanks for explaining the shared storage, configuration, and maintenance implications. That answers the separate-command question. The remaining question is whether those differences justify a separate listing—not whether the existing form can simply be preserved.

The extension guidelines also state the following:

CleanShot 2026-09-15 at 08 54 58@2x

We should also strive to create as few duplicate extensions as possible, since end users are getting confused about which one is best. Currently, many of them share the same functionality, right? 

What do you think @alexj

alexj:
Hi, I'll be honest, there's a lot of AI-generated text to parse through in
the thread, so I don't see the significant change or advantage to
the approach. While I'm very happy to have others contribute improvements
to TimeConverter, I'm not going to sign up for all of the extra overhead
since I don't understand the improvements. I don't understand what 'YATT is
a live list that recalculates on every keystroke' means or the value it
brings. As another example, I chose to build via Raycast in order to not
have another menu bar item; I also don't get where the file picker
extension would come in..?

Note, I'm not saying the new extension isn't worthwhile, I just don't
understand the goal. But, I also shouldn't be the person making the call on
whether it's included or not. :) I'm just one person in the larger
community.

Cheers,

Alex

alexbartok:
@alexj
Agreed. Our extensions do different things and they should not be mashed together.

@pernielsentikaer 
I understand your mission is to maintain quality and usability of the Raycast store. I'm well aware that there are tons of extensions (and applications in general) that
deal with timezone conversion, hence the name YATT, but there's a valid reason: different workflows need different tools.
Please let me know if YATT has a place in the Raycast Store.

PR #30705, opened 2026-09-02
Add parallels-vm-switcher extension

Async23:
## Description

Adds **Parallels VM Switcher**, an independent extension focused narrowly on reliable VM switching rather than VM lifecycle management.

The Store already contains `parallels-virtual-machines`. This extension provides a distinct workflow and implementation:

- identifies every registered VM by UUID, revalidates before every action, and fails before side effects when normalized names collide;
- activates the UUID-specific regular Parallels Dock Helper, raises the exact Console window, and verifies matching AX identifier, CoreGraphics window ID, onscreen membership, frontmost focus, and menu-bar ownership;
- supports a direct name-or-UUID command in addition to the searchable list;
- starts or resumes a VM only as part of switching to it, and intentionally exposes no suspend, reset, force-stop, or Parallels shutdown controls.

The implementation invokes external programs with argument arrays, never interpolates VM-provided values into shell or JXA source, and includes tests for registry validation, ambiguity, timeout, concurrency, untrusted VM names, hidden windows, focus reclamation, and bounded activation retries.

Attribution for the MIT-licensed upstream work is preserved in `LICENSE`, `NOTICE.md`, and `pastContributors`. The extension uses its own icon and does not reuse the existing extension's screenshots or OS artwork.

Validation completed locally:

- `npm test` — 34 tests passed
- `npm run typecheck`
- `npm run lint`
- `npm run build`
- `xcrun swiftc -typecheck scripts/verify_space_switch.swift`
- deterministic cross-Space acceptance — target began hidden and offscreen; the command moved to the exact target Space/window and passed 10 consecutive stable focus samples

## Screencast

Not included. The UI is a standard Raycast searchable list; the behavior under review is the nonvisual cross-Space activation described and deterministically verified above. The second command accepts a VM name or UUID directly.

## Checklist


pernielsentikaer (changes requested):
Hi @Async23 👋

Thanks for the contribution 💪

We already have [Parallels Virtual Machines](https://www.raycast.com/danpalmer/parallels-virtual-machines) in the Store by @danpalmer, which already lists, opens, starts, and resumes Parallels VMs.

I see the meaningful improvements here around reliable cross-Space switching and the direct name-or-UUID command. Could you look into bringing those improvements into the existing extension instead? If combining them there would be impractical, could you explain why? 🙂

---

PR #30711, opened 2026-09-02
[Neovim] Fix nested-bracket framing in keymap JSON scanner

RG-IL:
## Description

The keymap JSON scanner in the Neovim extension can be hijacked by a bracketed string value inside the payload itself.

**Reproducer:** a keymap whose `rhs` or `desc` contains a standalone valid JSON array, e.g.:

`{"lhs":"<leader>x", "rhs":"echo [1, 2]"}`

The right-to-left scan finds `[1, 2]` (the rightmost `[`), bracket-matches it as a complete array, `JSON.parse` succeeds, and returns the inner slice instead of the enclosing keymap payload. The user's keymaps silently disappear and are replaced by malformed entries.

### Root cause

The scanner iterates every `[` in the output and tries to match/parse from each one, but starts the bracket-matcher with fresh `inString` state — so it doesn't know whether a given `[` sits inside a JSON string or at the top level.

### Fix

Anchor on a positional invariant: `vim.json.encode()` emits compact JSON with no embedded newlines, and `print()` appends exactly one `\n`. The real payload's `[` therefore always appears at position 0 or immediately after a newline. By requiring that precondition before accepting a candidate, brackets inside string values are skipped without re-introducing cross-output quote tracking.

18 test cases cover the exact adversarial scenario, startup noise with unmatched quotes, brackets inside values, and combinations of both.

## Screencast

No UI change — purely internal parser fix.

## Checklist

### DIFF
--- a/extensions/neovim/CHANGELOG.md
+++ b/extensions/neovim/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Neovim Changelog
 
+## [Fix keymap scanner nested-bracket framing] - 2026-09-02
+
+- Fix keymap scanner parsing a bracketed string value (e.g. `rhs "echo [1,2]"`) as the payload instead of the enclosing JSON array, which caused user keymaps to silently disappear.
+
 ## [Initial Version] - 2026-09-02
 
 - Search and open recent Neovim sessions (persistence.nvim sessions, recently opened directories)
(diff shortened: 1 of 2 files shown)


0xdhrv (approved):
Looks good to me, approved ✅

PR #30713, opened 2026-09-02
Add vitra extension

PTthe13:
## Description

Adds **Vitra**, a single-command extension showing where your day stands — readiness, sleep and recovery from an Oura ring — without leaving Raycast. Each value is shown against the person's own 30-day baseline rather than a population average.

**How it gets the data.** Vitra is a local-first desktop app: the Oura history lives in a SQLite database on the user's own machine and never reaches a server. The app publishes a small read-only JSON file in its application-support folder, and this extension reads that file. Nothing else.

Two things it deliberately does *not* do:

- **No network calls at all.** Verified in the compiled production bundle — no `fetch`, no `XMLHttpRequest`, no `http`/`https`, no HTTP client of any kind.
- **It does not read the app's database, nor its local HTTP API.** That API sits behind a per-launch token specifically so other processes on the machine cannot reach it, and the database is WAL-mode, which outside readers handle badly. A published, versioned file is the contract instead, so the schema underneath is free to move without breaking this extension.

It also honours two of the app's settings rather than second-guessing them:

- Vitra can hold scores back until the user has rated their day, so the rating is not coloured by the number. When that is on, the extension says so instead of showing them — including in **Copy Summary**.
- When someone has declared that a device or medication sets their heart rate, HRV and resting rate are shown **without comparison or verdict**, because that number is not a read on their recovery.

## Reviewing without the app

The extension requires the Vitra desktop app, so a reviewer who does not have it will see the empty state — *"No snapshot from Vitra yet"* — with guidance and a link. That is intended behaviour, not a failure.

The three screenshots show it running against real data.

## Notes

- The logic lives in `src/snapshot.ts`, free of any `@raycast/api` import, with unit tests (`npm test`, 9 passing) covering the formatting rules and the withheld-scores behaviour.
- Durations and their deltas are always expressed in the same units — sleep never shows `5h 31m` with a `+0.1h` delta under it — and signed numbers go through `toLocaleString` so a decimal separator cannot differ between two rows of one list.
- `assets/vitra.png` is the app's own icon.
- `platforms` declares macOS and Windows; the snapshot path resolution handles both.

## Checklist


PTthe13:
Good catch on the read-error classification — that one was real and is now fixed.

The distinction matters more than it looks: the empty state tells the user to open Vitra, which is exactly the wrong advice when Vitra *is* running and the file is there but unreadable. Only `ENOENT` now means "not this path" — which genuinely is the ordinary case, since the application-support folder is `Vitra` on some machines and `vitra` on others, so the first candidate missing is expected rather than a fault. Anything else is a fault worth naming.

`load()` also takes its candidate paths as an argument now, so the four results it distinguishes are testable — that was the second half of the review note, that the loader classification had no coverage. Six tests added covering: a good snapshot, a genuinely absent one, falling through `ENOENT` to a later candidate, an unreadable path, corrupt JSON, and a schema from a newer Vitra. 15 passing, `ray lint` clean, production build clean.

pernielsentikaer (changes requested):
Hi @PTthe13 👋

We already have the [Oura extension](https://www.raycast.com/Rob/oura) in the Store by @RobErskine, which covers the same core readiness, sleep, activity, HRV, and resting-heart-rate workflow.

I can see Vitra’s meaningful differences—local-only access, personal baselines, withheld-score handling, and Windows support. Could you look into contributing these capabilities to the existing extension instead? If integrating them there wouldn’t be feasible because of the Vitra app dependency, could you explain the constraints here? 🙂

---

PTthe13:
Hi @pernielsentikaer 👋 — thanks for the careful look, and for flagging the overlap.

I looked into merging this into @RobErskine's Oura extension, and I don't think it's feasible — the two extensions have no code in common, and the differences you listed aren't features I could port over.

**The Oura extension is an OAuth API client. Vitra is a file reader.**

The Oura extension runs a full OAuth 2.0 PKCE flow: the user registers their own Oura OAuth app, puts `client_id`/`client_secret` into Raycast preferences, authorizes in the browser at `cloud.ouraring.com`, and the extension then calls `api.ouraring.com` with the resulting token across nine commands.

This extension does none of that. No preferences, no OAuth app to register, no token stored in Raycast, no network request at all. One view command reads a JSON file the Vitra desktop app writes locally (`~/Library/Application Support/Vitra/snapshot.json`, `%APPDATA%\Vitra\` on Windows) and formats it — about 200 lines of parsing and number formatting.

**That's why the differentiators can't really be contributed upstream — they aren't extension features.**

- *Local-only access*: the point is that no Oura credentials exist in Raycast and no request leaves the machine. That property comes from not being an API client, so it can't be added to one — it would mean bolting a second, unrelated data path onto an extension whose entire architecture is the OAuth flow.
- *Personal baselines*: the 30-day baselines are computed in the Vitra app from its own stored history, and arrive in the snapshot pre-computed. Reproducing them upstream would mean fetching and retaining 30 days of history per metric inside Raycast — a substantially different extension than the one that exists today.
- *Withheld-score handling*: `scoresWithheld` is a Vitra app concept — scores suppressed when the ring's data doesn't support them, such as paced-heart sessions. There's no field in the Oura API that carries it.

(Windows support isn't really a differentiator on reflection — it falls out of reading a file Electron already writes to `%APPDATA%`, and the API-based extension is cross-platform anyway. I'd drop it from that list.)

**What upstreaming would actually look like**: a command inside the Oura extension that ignores its OAuth token, reads a third-party desktop app's file, and shows an empty state to every user who doesn't have that app installed. That seems worse for that extension's users than a separate extension clearly labeled as the companion to Vitra.

On that last point — I've pushed a commit sharpening both descriptions so the dependency is stated up front rather than implied:

- Extension: "Requires the Vitra desktop app. Reads Vitra's local snapshot to show today's readiness, sleep and recovery from your Oura ring — no Oura account, no network access."
- Command: "Requires the Vitra desktop app. Today's readiness, sleep and recovery, read from the snapshot Vitra writes on this machine."

The README already states the requirement and the no-network-calls behaviour, and the empty state points people to vitrahealth.app. Happy to go further — rename the command, or make that empty state more prominent — just say which.

If you'd still rather it live upstream, I'll take that direction and coordinate with @RobErskine — just let me know. 🙂

pernielsentikaer (changes requested):
Hi @PTthe13 👋

Thanks for the detailed explanation. The separate extension makes sense given that Vitra’s local snapshot is its own integration contract rather than another Oura API client.

I found a few things in the full review:

- `load()` currently accepts any JSON object with `schema: 1` as a complete `Snapshot`. For example, `{"schema":1}` returns `ok`, then the view crashes when it accesses `snapshot.scores.readiness`. Could you validate the required nested shape before returning `ok` and treat malformed snapshots as unreadable?
- `open("Vitra")` passes `"Vitra"` as the target, but Raycast’s `open` API expects a file, folder, or URL. Could you launch it through a registered deep link or resolve the installed application and open its absolute path, with a useful fallback if it isn’t installed?
- Please remove the dynamic `navigationTitle` from the root `List`; Raycast supplies the command title automatically for root commands.
- The Store description says “no Oura account,” but Vitra requires an Oura Ring and active Oura Membership. Something like “no Oura credentials or setup in Raycast” would preserve the intended distinction without implying that no Oura account is needed.
- The PR notes that the screenshots use real data. Since these expose personal health metrics, could you recapture them using synthetic or anonymized values in line with the [screenshot guidelines](https://developers.raycast.com/basics/prepare-an-extension-for-store#screenshots)?

---

PTthe13:
Thanks — all five are addressed.

**1. `load()` accepted any object with `schema: 1`.** Correct, and it crashed exactly as you described. There's now a `snapshotFault()` check that validates every nested group and field type before `load()` returns `ok`; anything malformed comes back as `unreadable` and lands in the empty state the other failures already use, with the specific fault named (`scores is missing or not an object`).

While fixing it I found a neighbouring case: a file with no `schema` key at all was reporting "Vitra is newer than this extension — snapshot version `undefined`", because `undefined !== 1` took the unsupported-version branch. A missing schema number is malformed, not from the future, so it's a fault now too. Seven tests cover the validator.

**2. `open("Vitra")`.** Right again — an app name isn't a file, folder or URL, so it did nothing. It now resolves the installed app through `getApplications()` (bundle id `com.vitra.app`) and opens its real path. If Vitra isn't installed, there's a failure toast with an "Open vitrahealth.app" action, since anyone reaching for that action is quite likely to be someone who doesn't have the app yet.

**3. `navigationTitle`.** Removed. The day it was showing moved into the Source row so the information isn't lost.

**4. "no Oura account".** You're right that this overclaimed — Vitra itself needs an Oura ring and an active membership. The distinction I was drawing is about Raycast, so it now reads: "Requires the Vitra desktop app. Reads Vitra's local snapshot to show today's readiness, sleep and recovery from your Oura ring — no Oura credentials or setup in Raycast, and no network calls."

**5. Screenshots.** Replaced. All three are recaptured against a made-up snapshot — readiness 82, HRV 48 ms, resting HR 58 bpm and so on are invented numbers, not anyone's morning. The old set also had a personal photo showing through the window's translucency, so the new ones use a plain backdrop.

They needed retaking regardless: the old footers still showed the `navigationTitle` from point 3.

`ray lint` and `ray build` are clean and the 22 tests pass. Marking it ready for review now — thanks for the thorough pass, the first two were real bugs.

pernielsentikaer (approved):
Looks good to me, approved 🔥

RobErskine:
Nice extension! I haven't heard of Vitra... I'll have to check it out. I agree that this is different enough from my Oura Extension to exist as its own thing.

PTthe13:
Thanks @RobErskine — and for taking the overlap question in good spirit. Your extension is the right call for anyone who wants their Oura data straight from the API; Vitra's just the local-file path for people already running the desktop app. Appreciate the note. 🙏

PR #30717, opened 2026-09-02
fix(runpool): keep overlapping resize confirmations apart

aicayzer:
Two follow-ups to the extension published on 1 September.

**A resize confirmation could act on the wrong count.** `Tool.Confirmation` returns only a message, so the count it read reaches the write through a module-scoped note, and there was a single slot. A second confirmation overwrote the first, so a change approved as growth could be written as a shrink that deregisters runners. Each note is now keyed to the pool and target count it asked about, and a shrink that cannot be tied to a confirmation is refused rather than falling back to the current count. This came from a Greptile review point raised on #30343 after the extension was published.

**The changelog had an `[Unreleased]` heading.** It matches neither `## [Title] - {PR_MERGE_DATE}` nor an explicit date, so the store listing shows it as a version by that name, although everything under it shipped in the first release. Folded into the dated entry and tightened, because twenty bullets for an initial release was not readable.

No new dependencies, no new commands, no metadata change. `ray lint` and `ray build -e dist` are both clean, and the note logic was exercised against the overlapping-confirmation cases directly, including the one that previously wrote a shrink.

### Checklist

### DIFF
--- a/extensions/runpool/CHANGELOG.md
+++ b/extensions/runpool/CHANGELOG.md
@@ -2,8 +2,8 @@
 
 ## [Fix a resize that could act on the wrong count] - {PR_MERGE_DATE}
 
-- Record each resize confirmation against the change it asked about, so two open at once cannot swap starting counts and turn approved growth into a shrink
-- Refuse a shrink that cannot be tied to a confirmation, rather than falling back to the current count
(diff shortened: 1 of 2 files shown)


0xdhrv (approved):
Looks good to me, approved ✅

PR #30731, opened 2026-09-02
Update stealth-ai-tool extension

AA1labs:
## Description

Adds **LM Studio** and **Ollama** as providers, so Stealth AI actions can run entirely against a local model with no cloud API key.

Local support was previously impossible: the request layer imported only `node:https`, so any `http://localhost` endpoint failed outright. The transport now selects `http` or `https` per URL scheme.

**Providers**

| | Chat | Model list |
| --- | --- | --- |
| LM Studio (default `http://localhost:1234`) | `/v1/chat/completions` | `/api/v0/models`, falling back to `/v1/models` |
| Ollama (default `http://localhost:11434`) | `/api/chat` (`stream: false`) | `/api/tags` |

Both need no API key and expose an editable **Server URL** that normalizes what users paste — `localhost:1234`, `http://127.0.0.1:1234/v1/` and LAN addresses all resolve correctly. The optional API key field is there only for servers placed behind auth.

LM Studio models are listed via the native `/api/v0/models` because the OpenAI-compatible `/v1/models` does not report model **type** — without it, embedding models appear in the picker as selectable chat models. The native endpoint also reports load state, so the loaded model is marked. Older builds without `/api/v0` fall back automatically. Ollama models show parameter size and quantization.

**Reliability and correctness fixes found while working in this code**

- Requests had no timeout and could hang indefinitely — now 60s for cloud providers, 180s for local ones, since a cold local model may need to load first
- Error handling keyed off `/model/i.test(message)`, so unrelated failures mentioning "model" were reported as misconfiguration while genuine ones were not — replaced with a typed `LLMConfigError`
- Anthropic responses were truncated at `max_tokens: 1024` — raised to 4096
- Gemini API key moved out of the request URL into the `x-goog-api-key` header
- Model dropdown could hold a value absent from the fetched list after switching providers
- Debounce was global, so one action blocked every other action for 3 seconds — now tracked per action
- Removed dead code: `aiProvider`/`apiKey` preference fallbacks reading preferences that are not declared in the manifest, an unused `fetchModels()`, and a full `AI.Model` dump logged on every run

**Docs**

The README documented a multiline prompt editor that no longer exists in the code (`forceEditor` is never set, and nothing writes the `action-configs` key that was read on every run). Corrected to describe the actual preference fields rather than re-adding the feature.

## Testing

Verified against live LM Studio and Ollama servers running locally:

- Model listing for both, including the LM Studio `/api/v0` → `/v1` fallback (exercised against a stub server) and the filtering of embedding models
- Server URL normalization across the URL shapes above
- End-to-end completions through `askAI` for both providers, returning correct output for the Fix Grammar prompt
- Error paths: unreachable server, missing API key and missing model each raise `LLMConfigError` with an actionable message

`ray lint` and `ray build` are clean.

Note: the checklist item about testing the distribution build inside Raycast is left unchecked — the provider layer was verified directly against live local servers rather than by exercising the packaged build in the Raycast client.

## Screencast

No new commands or UI surfaces were added. The existing **Configure AI Model** command gains two entries in the provider dropdown and a Server URL field when a local provider is selected.

## Checklist

### DIFF
--- a/extensions/stealth-ai-tool/.gitignore
+++ b/extensions/stealth-ai-tool/.gitignore
@@ -16,3 +16,4 @@ compiled_raycast_rust
 node_modules
 dist
 .DS_Store
+.serena
--- a/extensions/stealth-ai-tool/CHANGELOG.md
+++ b/extensions/stealth-ai-tool/CHANGELOG.md
@@ -1,5 +1,20 @@
 # Stealth AI Changelog
 
+## [Local Models: LM Studio & Ollama] - {PR_MERGE_DATE}
+
+- Add LM Studio and Ollama as providers, so prompts can run entirely on your own machine
+- Add an editable Server URL per local provider (defaults: `http://localhost:1234`, `http://localhost:11434`)
+- Accept pasted URLs in any common shape (`localhost:1234`, `http://127.0.0.1:1234/v1/`)
+- LM Studio model list uses the native `/api/v0/models` endpoint, hiding embedding models and marking the loaded one
+- Ollama model list shows parameter size and quantization
+- HTTP endpoints are now supported (previously every request was forced through HTTPS)
+- All requests now time out instead of hanging: 60s for cloud providers, 180s for local ones
+- Clearer failures: unreachable server, missing key and missing model each explain the fix and link to "Configure AI Model"
+- Anthropic responses are no longer truncated at 1024 tokens
+- Gemini API key moved out of the request URL into a header
+- Fix model dropdown breaking when a saved model is missing from the fetched list
+- Debounce is now per action, so one action no longer blocks a different one for 3 seconds
+
 ## [In-App AI Configuration] - 2026-02-09
 
 - Add in-app AI provider and model configuration (no more Raycast Settings)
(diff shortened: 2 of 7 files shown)


AA1labs:
Thanks — valid catch, fixed in the latest push.

\`handleSubmit\` wrote \`configured_provider\` as its first statement, so both rejection paths (missing API key, and no model selected) returned *after* the active provider had already been switched. The result was that a user moving from a working provider to, say, Anthropic without a key would end up with Anthropic active and unusable, and all nine actions would fail with a missing-key error until they went back and fixed it.

Validation now runs before any write, and \`configured_provider\` is persisted last — only once the key and model for that provider have been stored. Every rejection path returns without touching storage.

Worth noting the no-model case was pre-existing rather than new to this PR; the missing-key check I added widened the same flaw. Both are covered by the reordering.

valentindoering:
Hi @AA1labs — I tested the pending Stealth AI update and prepared a focused follow-up in https://github.com/AA1labs/raycast-extensions/pull/1. It fixes current GPT-5/OpenAI requests, restores multiline action editing, and adds two mutually exclusive macOS text workflows: Raycast native or Script Command-style Cmd+C/Cmd+V. I also moved provider, API key, model ID, and local server URL configuration into standard Raycast extension preferences, following the Store guidance against separate setup commands. Because manifest dropdown values are static, the model is entered as an exact ID instead of being fetched dynamically. Tests, lint, production build, and the Raycast development build pass. The PR remains a draft for UI testing, and I am happy to adjust or split the changes if useful.

PR #30737, opened 2026-09-02
Infisical: add folder support

aicayzer:
## Description

Secrets inside folders weren't reachable. `listSecrets` is called without `secretPath` and isn't recursive, so only the root of each environment ever showed up, with nothing to indicate the rest was there.

Folders are now listed for the current path and open into a nested view, one level at a time. The selected environment carries into the nested view.

While doing this I found `secretPath` was also missing from create, edit and delete, so those hit the root regardless of which folder you were in. Fixed here too, since folder browsing is what makes it reachable.

"Save All as .env" now covers the folder you're looking at rather than the root. At the root that's unchanged.

## Screencast

Tested against Infisical Cloud EU on a project with nested folders: folders list correctly, opening one shows its contents, the environment dropdown persists into the nested view, and a folder containing only subfolders shows those rather than the empty state.

## Checklist


pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #30741, opened 2026-09-03
Add ytranscript extension

leoisadev1:


pernielsentikaer (changes requested):
Hi @leoisadev1 👋

We already have [Fetch YouTube Transcript](https://www.raycast.com/apoorv_khandelwal/fetch-youtube-transcript) in the Store by @apoorv12, which supports the same core workflow of fetching a YouTube transcript and copying or saving it.

Your Windows support, setup-free approach, and additional export formats look like useful improvements. Could you explore adding them to the existing extension instead of creating a separate listing? If there’s a reason consolidation wouldn’t be feasible, please let us know 🙂

---

PR #30742, opened 2026-09-03
Update Digger extension

chrismessina:
## Description

Digger reported failed lookups as successful-but-empty ones. A Wayback check that timed out showed an invented snapshot count, a `robots.txt` request that 5xx'd read as "Not found", and a failed DNS or certificate lookup rendered a blank section with nothing saying why. This release makes each of those say what actually happened, and fixes a Refresh action that read straight through the cache it was meant to bypass.

Also migrates to Raycast API 2.x.

### Changes

- **Failed lookups report in place.** DNS, SSL certificate, Wayback and host metadata each read "Couldn't check" in their own section instead of a single truncated banner row, with a toast as the one-time heads-up. The statuses travel with the cached result, so a cached dig still explains itself.
- **`robots.txt` / `llms.txt` / `sitemap.xml` distinguish absent from unchecked.** A 5xx, timeout or refused connection reads "Couldn't check"; rows read "Checking…" until the request returns. Only a real answer from the server — a 404/410, or a page that is really an error page — says the file is absent.
- **Wayback no longer invents a snapshot count.** A failed precise-count fell back to a formula meant for very large archives, reporting a site with 8 snapshots as 5,000 under an "estimate" label; a non-2xx response fell through to "No snapshots available". Both now report the check as failed.
- **The Wayback lookup is bounded and cancellable.** Its four sequential requests each retried independently with no shared limit, so an unresponsive archive.org could hold the section for over a minute. They now share one budget and stop as soon as their dig is superseded.
- **"Refresh" actually refreshes.** It read through the cache, so on a stale result it re-rendered that same result. It now bypasses the cache; a failed refresh does not discard the cached copy.
- **Overview shows the site's own favicon**, falling back to a globe immediately rather than sitting on a chain icon while a third-party lookup runs, or rendering an empty icon slot for a host that never responds.
- **Host metadata (`/.well-known/host-meta`) is now displayed** under Data Feeds & API. It was fetched and cached on every dig but rendered nowhere.
- **Empty digs are no longer cached.** A site answering an unhappy status with no body parsed into a result with no title or metadata, and caching it pinned that empty view for the full 48-hour TTL.
- A non-2xx status is now marked in Overview; previously only 200 carried an indicator.
- Dropped six type fields nothing ever populated, including the two the Markdown export branched on — those rows could never appear.

## Checklist


chrismessina:
@greptile

**On "ENOTFOUND masks resolver failures"** (the live one, against `79f70df5`):

```
Not fixing this one — I think it would introduce a regression, and I'd rather
explain than silently decline.

All six queries returning ENOTFOUND is the normal result for any host that
isn't in public DNS, not a symptom of a broken resolver. `dns.resolve*`
bypasses the OS resolver by design, so:

  93.184.215.14   a:ENOTFOUND aaaa:ENOTFOUND cname:ENOTFOUND mx:ENOTFOUND ns:ENOTFOUND txt:ENOTFOUND
  localhost       a:ENOTFOUND ...
  my-nas.local    a:ENOTFOUND ...
  printer.lan     a:ENOTFOUND ...

Every one of those loads fine in a browser. Treating ENOTFOUND as a resolver
failure would put "Couldn't check" on all six rows for IP literals, localhost,
mDNS names, LAN hostnames, and anything in /etc/hosts — which is the exact
false positive the comment above the BENIGN set exists to prevent.

The filtering-resolver case you describe is real but not distinguishable:
NXDOMAIN is NXDOMAIN, whether the name is absent or the resolver is lying. It's
also already covered end-to-end — if the name genuinely doesn't resolve, the
main page fetch fails too and the user gets the error view, not a DNS section
claiming records are absent.

ENODATA and ENOTFOUND are the only two codes that mean "DNS answered, and the
answer is that this record isn't there." The other 22 Node DNS codes all reach
the failure path.
```

**On "Partial DNS failures look successful"** (anchored to `9e206182`, superseded):

```
This is fixed as of 79f70df5 — the comment is anchored to 9e206182, one commit
earlier. Current head is fa028db1.

DNSData now carries `unchecked`, the record types whose query failed for a
non-benign reason:

  dnsUtils.ts:135          sets dnsData.unchecked when a query fails non-benignly
  DNSCertificates.tsx:43   renders "Couldn't check" for those rows
  types/index.ts:206       `unchecked` lives inside DiggerResult, so it survives caching
```

You're right that the caller still marks lookups.dns as found — that's
deliberate. The per-record-type state travels in the data rather than in the
section status, so a partial failure shows per row without escalating to a
toast. Toasts are reserved for whole-subsystem failures (DNS entirely
unreachable, cert, Wayback, host metadata); a single record type that couldn't
be checked is reported in the section the reader is already reviewing row by row.

chrismessina:
> Thanks for the detailed clarification. I’m withdrawing both findings:
> 
> * `ENOTFOUND` is intentionally treated as benign because `dns.resolve*` does not cover OS, `/etc/hosts`, mDNS, or local-network resolution, and the main fetch already handles a genuinely unreachable hostname.
> * The partial DNS failure concern is resolved at the current head (`fa028db1`): non-benign failures are carried in `DNSData.unchecked` and rendered per record type, including through cached results.
> 
> No further change is needed for either comment.

@greptile ok great — so can you update your score?

0xdhrv (approved):
Looks good to me, approved ✅

PR #30750, opened 2026-09-03
Add lost extension

chasebank87:
## Checklist


pernielsentikaer (changes requested):
Hi @chasebank87 👋

Thanks for the contribution and for addressing the earlier focus and screenshot feedback 💪

We already have [Window Walker](https://www.raycast.com/nazzy_wazzy_lu/window-walker) in the Store by @luinbytes. It provides the same core workflow—searching open windows by app or title and bringing one to the front—but currently supports Windows only. Lost adds a macOS implementation, previews, and next/previous cycling, which are useful additions. Could you check whether these can enhance Window Walker instead of creating a separate listing, and coordinate with its author? If consolidation isn't practical, please explain the technical or workflow differences that require a separate extension.

I also found a few correctness issues in the current implementation:

- In `raiseAXWindow` (`swift/LostWindows.swift`), an exact Core Graphics window-ID mismatch falls through to fuzzy title matching. When multiple windows share a title, this can raise a different window from the one the user selected. Please only use title matching when ID lookup is unavailable or fails—not after it returns a confirmed different ID.
- `isPlaceholder` drops every window smaller than `200×120`, plus every `800×600` window at `(0,0)` and `500×500` window at `(0,617)`. Because this is used for both listing and front-window detection, legitimate windows can disappear or make cycling fail. Please identify placeholders using app/window identity or accessibility metadata rather than globally reserved dimensions.
- `assets/lost-windows` is committed as a prebuilt executable, while `npm run build` doesn't regenerate it because `build:swift` is separate. This means CI cannot verify that the shipped binary matches the reviewed Swift source. Please coordinate with a Raycast maintainer to have the final binary built/added through the repository's binary-dependency process, or add an equivalent reproducible verification step.

Please address these points, and we'll have another look 🙂

PR #30752, opened 2026-09-03
Capacities: pop to root search after a successful capture

mikolajadamowicz:
## Description

The Create Task, Save Weblink and Save to Daily Note commands ended their submit handler with an unawaited `showHUD(...)` followed by a bare `closeMainWindow()`. With no options `closeMainWindow()` uses `PopToRootType.Default` which follows the user's **Pop to Root Search** preference. The window hid, but the command's navigation stack stayed on the form. Reopening Raycast inside that window landed back on the Capacities form, still holding the values just submitted, which makes a duplicate save easy to trigger.

Each handler now ends with a single awaited call:

```ts
await showHUD("Notes saved to daily note", {
  clearRootSearch: true,
  popToRootType: PopToRootType.Immediate,
});
```

`showHUD` closes the main window itself and accepts the same options, so the old pair was redundant with no guaranteed ordering. This also removes a `Toast.Style.Success` toast that fired immediately before the close and was never visible.

No change to the API calls, validation, or space selection.

Tested on macOS. The extension also declares Windows support, which I had no way to verify.

## Checklist

### DIFF
--- a/extensions/capacities/CHANGELOG.md
+++ b/extensions/capacities/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Capacities Changelog
 
+## [Return to root search after capturing] - {PR_MERGE_DATE}
+
+- Fixed the Create Task, Save Weblink and Save to Daily Note commands leaving the form on the navigation stack after a successful save, so reopening Raycast returned to the filled-in form instead of the root search.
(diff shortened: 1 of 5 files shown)


0xdhrv (approved):
Looks good to me, approved ✅

PR #30756, opened 2026-09-03
Add hule extension

hule-do:
## Description

[Hule](https://hule-do.com) is a task manager. This extension adds four commands:

- **Quick Add Task** — one line, no view: the task lands in a configurable default list.
- **Create Task** — the full form: list, title, description, priority, due date, assignee.
- **Search Tasks** — search a workspace, with a workspace picker when you belong to several. An empty search bar shows the tasks that moved most recently.
- **My Tasks** — everything assigned to you, grouped into Overdue / Today / Tomorrow / Later.

From either list, `↵` opens the task's details inside Raycast. The action panel (`⌘K`) changes status (`⌥S`), priority (`⌥P`), due date (`⌥D`) and assignee (`⌥A`), opens the task in Hule, copies its link, or deletes it.

Authentication is a personal API token (`hule_pat_…`), created in Hule under **Settings → API tokens** and stored as a `password` preference. Two optional preferences: an API URL for self-hosted instances, and the default list Quick Add Task writes to. Authorization stays server-side — the extension is an ordinary authenticated REST client, and the API's own permission rules decide what a token may reach. No analytics, no bundled binaries, no Keychain access.

`platforms` is macOS only: the extension asks the system which application claims the `hule://` scheme, so "Open in Hule" can hand a task to the desktop app when it is installed and fall back to the browser otherwise. That path has not been tested on Windows.

## Screencast

Four screenshots in `metadata/` (2000×1250): the recent-tasks list, a task's details, the action panel, and the create form. They were captured against a demo workspace — no real data is shown.

**Reviewer access:** Hule needs an account to try the extension. I'm happy to provide a test account and an API token — tell me where to send them.

## Checklist


pernielsentikaer (changes requested):
Hi @hule-do 👋

Thanks for the Hule integration 💪

I found two issues while reviewing:

- `resolveList` silently falls back to the first available list when the configured **Default List** is missing, and duplicate names across workspaces also resolve to whichever appears first. This can create a task in the wrong place. Could you require a unique match whenever the preference is set and show an actionable error otherwise?
- After deleting a task from its detail view, the shared refresh handler calls `getTask` for the deleted ID and leaves the stale detail open. Could you return to the task list after a successful deletion instead?

Regression coverage would be helpful, but it isn’t required since the extension currently has no test setup.

---

pernielsentikaer (changes requested):
Hi @hule-do 👋

Thanks for fixing the default-list selection and returning to the list after deletion 💪

I found another issue in `src/helpers/images.ts`: image `fileId` values become local filenames directly. A value like `../outside.png` can resolve an existing image outside the cache, and a successful download can write outside it too. I reproduced both paths with isolated fixtures; I haven't tested whether the hosted API accepts such a description.

Could you use a filesystem-safe cache key instead? For example, add `import { createHash } from "node:crypto"` and replace the current `base` assignment with:

```ts
const base = createHash("sha256")
  .update(JSON.stringify([workspaceId, fileId]))
  .digest("hex");
```

This keeps remote IDs out of paths and preserves the existing MIME-based extension handling.


---

hule-do:
Thanks for catching this, @pernielsentikaer!

Fixed: image cache files are now keyed by `sha256([workspaceId, fileId])` as you suggested, so a remote id never becomes a path; the extension still comes from the MIME type.

While at it I audited the rest of the extension for the same class of issue and a few others:
- Task text is now escaped before it goes into Detail markdown, so a title or description can't inject remote images or live links; code blocks get a fence longer than any backtick run inside.
- Image downloads are capped (20 per description, 10 MB each, image MIME only, checked before reading the body) and written atomically.
- The token is validated before it reaches `fetch`, and the API URL must be `https` (plain `http` only for localhost).
- Failed loads now show an error view with Try Again instead of an empty result, and one failing workspace no longer blanks the others.
- Only lists you can edit are offered for new tasks, and read-only tasks don't offer edits the server would reject.

pernielsentikaer (changes requested):
Hi @hule-do 👋

Thanks for fixing the image cache paths 💪

The Markdown hardening still has a gap in `src/helpers/richText.ts`: code blocks nested inside blockquotes or lists lose their container prefix after the first line. A literal `remote` inside that code then becomes an active image instead of displayed code.

Could you prefix every blockquote line and indent every list continuation line? For blockquotes, replace the current mapping with:

```ts
.map((b) => b.split("\n").map((line) => `> ${line}`).join("\n"))
```

For lists, preserve block boundaries and indent continuation lines beneath the item marker rather than joining blocks with spaces. I reproduced both cases with a CommonMark parser, not a live Hule account.

---

hule-do:
Thanks, @pernielsentikaer — good catch.

Fixed at the root rather than per case: every container now owns every line it holds. Blockquotes prefix each line (blank lines included, as `>`), list and task items indent every continuation line under the marker and keep their blocks as separate blocks, and `\r`/`\r\n` are normalised to `\n` first so no line ending escapes the prefixes.

I verified it with two CommonMark parsers (commonmark.js and markdown-it) over 35k generated documents: quotes and all three list kinds nested up to three levels deep, with code blocks, paragraphs, headings, image alts, attachment names and mentions carrying 23 injection payloads. No link, HTML node or foreign image survives, while code stays code and the extension's own cached images still render. The same run against the previous version flags your case and 2,700+ others.

pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #30764, opened 2026-09-03
Link Commands: keep a typed sigil out of the package name

kud:
## Description

Two fixes to **Create Link Command**, plus a small tidy-up.

**A `@scope` or `#category` typed into the Package field was taken as part of the brand.** Package drives the generated filename, so `Linear · @work` produced a brand called literally that: the file landed as `linear-work.my-issues.sh` rather than `work.linear.my-issues.sh`, and the list read the scope back as part of the brand instead of as a scope. Typing a sigil there is someone reaching for a control that already exists a few rows away, so it is now moved to that control on blur.

Two deliberate details:

- **A control you have already set is never overridden.** If Environment is set and a conflicting `@scope` is typed into Package, the sigil is removed from the brand and reported as dropped rather than replacing your choice.
- **The move is confirmed in a line beneath the field.** The filename preview shows the *outcome* but never the *edit* — it would name the new filename without ever saying that two tokens were lifted out of the field you were typing in.

Read-side parsing is untouched: `facetsOf` still takes the scope from the title. The new `splitTypedPackage` describes what a person typed, never how a command on disk is interpreted.

**Icons were stored per command rather than per package, and never reused.** The asset directory was keyed on the script's filename, so every new command re-fetched a service's icon into a private folder, accumulating identical copies. It is now keyed on the package, and an icon already present in your script directory is used before the network is asked. That second half matters for private or intranet hosts, which are invisible to any public favicon service — a command for one fell back to a generic link glyph even when the correct mark was already sitting beside it. Commands with no package keep the previous filename-based key.

**Example values in source comments are now neutral placeholders.** Several doc comments illustrated the convention with names taken from one contributor's own command collection, including an internal hostname. They say the same thing with generic examples.

`npm run lint` and `npm run build` both pass. Draft until the distribution build has been click-tested in Raycast.

## Screencast

Not attached — no new UI surface. The only visible addition is one confirmation line under the Package field after a sigil moves, reading e.g. `Moved @work to Environment, moved #dev to Category.`

## Checklist

### DIFF
--- a/extensions/link-commands/CHANGELOG.md
+++ b/extensions/link-commands/CHANGELOG.md
@@ -1,5 +1,16 @@
 # Link Commands Changelog
 
+## [Sigils and Shared Icons] - {PR_MERGE_DATE}
+
+### Fixed
+
+- A `@scope` or `#category` typed into **Package** now moves to the control that owns it, instead of being taken as part of the brand. `Linear · @work` used to make a brand called literally that: it slugged into `linear-work.my-issues.sh` rather than `work.linear.my-issues.sh`, and the list read it back as a brand rather than a scope. A control you have already set is never overridden — a conflicting sigil is reported as dropped instead — and the move is confirmed in a line beneath the field, because the filename preview shows the outcome but never the edit.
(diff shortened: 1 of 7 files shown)


0xdhrv (approved):
Looks good to me, approved ✅

PR #30765, opened 2026-09-03
Link Commands: lead a list row with its name, not its host

kud:
## Description

**A list row now leads with the command's name and carries its host in the subtitle**, rather than the other way round.

Leading with the host is the right call in Raycast's own launcher: a row there appears among apps and extensions, and a leading domain says *this opens a browser* before anything else is read. But this list has no strangers in it — every row is already a link command, already grouped, already sorted by package. The signal fires on every row, so it distinguishes none of them, while spending the front of the only column that can differ.

Measured over a collection of 277 commands:

| | before | after |
| --- | --- | --- |
| titles longer than the visible column | 181 (65%) | 52 (18%) |
| rows made identical by a shared host prefix | 3 | 0 |
| titles left empty by the change | — | 0 |

```
figma.com · Recent Files   →   Recent Files   figma.com
linear.app · My Issues     →   My Issues      linear.app
notion.so · Roadmap        →   Roadmap        notion.so
```

The host is not discarded, it moves to the subtitle — where it answers *is this the one I want* rather than competing with the name. A title that is only a host keeps it and shows its package in the subtitle, as before; so does a command with no web target. The host is read from the command's own target rather than parsed out of the title, so a collection following no naming convention is unaffected.

**Nothing searchable changed.** The full title, the filename and the package name are all still in `keywords`, so every command is found by exactly what found it before — only what the row *displays* is different.

`npm run lint` and `npm run build` both pass. Draft until the distribution build has been click-tested in Raycast.

## Screencast

Not attached yet — the change is entirely in how each row is composed; the mock above is what it looks like.

## Checklist


0xdhrv (commented):

0xdhrv (approved):
Looks good to me, approved ✅

PR #30769, opened 2026-09-03
Wispr Flow: fix transcript selection position on open

artisticmedic:
## Description

Search Transcripts opens with the selection on the transcript that was first in the list on the previous run, not on the newest one. Any transcript recorded since that run is prepended above the selected row, so the list scrolls to the top while the highlight sits below the fold. After a dictation-heavy interval the selected row can be dozens of entries down, requiring a scroll or a mouse click to reach the newest transcript.

Cause: the command reads through `useCachedPromise`, which paints the previous run's rows before the fresh query resolves. Raycast anchors the selection to a row from that stale paint, and the anchor survives the data update because the row still exists.

The fix pins the selection to the first rendered row once the current query's own results arrive. The pin follows a complete loading cycle rather than a single `!isLoading` check: on a query change there is one render that carries the new query key alongside the previous query's data, and pinning there anchors to a row that may be absent from the new results.

Supporting changes:

- Pin to `groups[0].transcripts[0]` rather than `data[0]`. Date sections render in fixed Today through Older order regardless of the SQL `ORDER BY`, so `data[0]` is the first visible row only under Newest First. Under Oldest First it falls in the last section.
- Add an explicit `id` to `List.Item`. Without one Raycast generates its own identifiers and `selectedItemId` matches nothing.
- Record selection changes in refs rather than state. `onSelectionChange` fires on every arrow-key press, and a state write there re-renders the list, regrouping every loaded row and rebuilding each detail pane.
- Release `selectedItemId` once the pin has applied, so paging in another page of transcripts does not reassert the pinned row.
- Re-arm the pin on archive only when the archived row was selected and actually leaves the results. With Show Archived enabled the row remains, and re-arming would move the selection for no reason.
- Memoize the date grouping.

`package.json` reformatting is Prettier conformance. The file did not satisfy `ray lint` before this change.

## Screencast

Not included. The defect is a selection position that depends on transcript history accumulated between two launches of the command, which a short recording does not convey.

## Checklist

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -328,7 +328,7 @@
 /extensions/capd                                                    @jamiedavenport
 /extensions/capture                                                 @syhchen
 /extensions/capture-fullpage-of-website                             @Coun1er @j3lte @ridemountainpig
-/extensions/capture-quick-notes                                     
+/extensions/capture-quick-notes                                     @FurkanSimsir
 /extensions/capture-raycast-metadata                                @koinzhang
 /extensions/carbon-raycast                                          @adelowo
 /extensions/cardpointers                                            @emcro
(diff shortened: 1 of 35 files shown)


pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #30770, opened 2026-09-03
fix(runpool): exclude paused pools from menu bar fill fraction

aicayzer:
## Summary
- The menu bar icon and root-command subtitle summed `busy`/`count` across every pool, paused ones included, so a paused pool's idle slots diluted the fraction. A fully busy pool could read as half-full simply because a sibling pool was paused.
- Both aggregates now exclude paused pools, matching how each per-pool row already treats a paused pool's fraction as not meaningful.

## Test plan
- [x] `npm run lint`
- [x] `npm run build`
- [x] Verified against live `runpool status --json` output with one paused pool and one active pool


0xdhrv (approved):
Looks good to me, approved ✅

PR #30811, opened 2026-09-06
Add snapstate extension

SoulSniper-V2:
## Description

SnapState is a native macOS app that saves and restores complete workspaces. This extension adds a small Raycast companion for two focused actions:

- **Restore Workspace** — search saved SnapState workspaces and ask the app to restore one.
- **Save Current Workspace** — name the current workspace and ask SnapState to capture it.

The extension reads only the local summary index at `~/Library/Application Support/SnapState/raycast-workspaces.json`. It does not read window URLs, titles, or workspace snapshots, and it does not duplicate SnapState’s Accessibility or Apple Events implementation. The native app remains required and remains the paid product.

The app-side bridge is available in SnapState 2.5.2 and uses the `snapstate://restore` and `snapstate://capture` URL actions.

## Verification

- `npm run build` — passed.
- `npm run lint` — passed.
- `npx tsc -p tsconfig.json --noEmit` — passed.
- Manifest, icon, lockfile, README, metadata screenshots, and both command entry points are included.
- Native bridge and the notarized SnapState 2.5.2 DMG are live at https://getsnapstate.com.
- Metadata screenshots were captured from the local macOS Raycast development UI and pass the repository image checks.

The extension is intentionally thin: Raycast is a launcher and search surface; SnapState owns the actual capture and restore behavior.

## Checklist

- [x] Read the extension guidelines.
- [x] Read the publishing documentation.
- [x] Ran `npm run build` and `npm run lint`.
- [x] Checked that the icon is used by the extension.
- [x] Kept README assets outside the metadata folder.
- [x] Added and verified metadata screenshots.

The extension’s native URL actions and distribution build were verified independently; the full capture/restore behavior remains owned by the installed SnapState app.


pernielsentikaer (changes requested):
Hi @SoulSniper-V2 👋

Thanks for the clean companion bridge — the index reading is nicely defensive, the empty/error states are well separated, and the README explains the local-only privacy boundary clearly 💪

One thing I'd like to sort out before merging:

`createSnapStateURL()` in `src/lib.ts` builds the query with `URLSearchParams`, which form-encodes spaces as `+`. So a workspace named "Deep Work" is sent as `snapstate://capture?name=Deep+Work`. A custom URL scheme is normally parsed with strict URL rules, where `+` stays a literal plus rather than being decoded back to a space, so unless SnapState replaces `+` itself, multi-word names will be saved as `Deep+Work`. Our Store extension LayoutLock works around exactly this for its own scheme by emitting `%20`.

Could you confirm SnapState decodes `+`, or switch to `%20`? Something like:

```ts
export function createSnapStateURL(command: string, parameters: Record<string, string> = {}): string {
  const query = new URLSearchParams(parameters).toString().replace(/\+/g, "%20");
  return query ? `snapstate://${command}?${query}` : `snapstate://${command}`;
}
```

Two small things while you're in there:

- `src/restore-workspace.tsx` renders `${workspace.windowCount} windows`, so a single-window workspace shows `1 windows`. Could you use the same singular/plural handling as `workspaceAccessory()`?
- `package.json` doesn't include `@raycast/eslint-config`/`eslint`/`prettier` (or `@types/node`/`@types/react`), so `ray lint` prints "ESLint is not installed, skipping formatting check". Adding the usual devDependencies would let that check run — optional, but most new extensions include them.

On my side `npm run build`, `npm run lint`, `tsc --noEmit` and the metadata image checks all pass, so this is close 🙂

---

SoulSniper-V2 (commented):
Hi @pernielsentikaer, thanks for the thorough review!

**`+` vs `%20`:** You were right. I checked SnapState's handler — `handleExternalURL` reads params via `URLComponents.queryItems`, which percent-decodes but leaves `+` as a literal plus. So "Deep Work" would indeed have been saved as "Deep+Work". Switched `createSnapStateURL` to your suggested `%20` form.

**Pluralization:** Fixed. Added a shared `pluralize()` helper in `lib.ts` (also reused by `workspaceAccessory()`) — the restore list now renders "1 window".

**devDependencies:** Added `@raycast/eslint-config`, `eslint`, `prettier`, `@types/node`, and `@types/react`, and regenerated the lockfile. `tsc --noEmit` and eslint both pass locally, so `ray lint` should have its formatting check back.

All pushed to the branch — marking ready for review now. Thanks again!

Issue #29462, opened 2026-07-15
[Browser Bookmarks]


mikulabc:
### Extension

https://www.raycast.com/raycast/browser-bookmarks

### Raycast Version

0.68.0.0

### OS Version

Windows Windows 11 Professional 10.0.26200.0 amd64

### Description

I am getting this error below, could it be related to my 10k Bookmarks in Chrome, or that I have a lot of `javascript:xxx` bookmarklets too?

```
Error: Command terminated after reaching the extension memory limit (100 MB JS heap).
t.toRayError:index.mjs:927:11034
---
JavaScript Stack
t.toRayError:index.mjs:927:11034
Worker.<anonymous>:index.mjs:927:8278
Worker.emit:node:events:531:35
[kOnExit]:worker:316:12
```

### Steps To Reproduce

See Description

PR #30828, opened 2026-09-06
[Browser Bookmarks] Find Brave Nightly, Vivaldi, Helium and Dia on Windows

Just-Me-22:
On Windows this extension only ever showed me Edge, even though I had 1,403 bookmarks sitting in Brave Nightly.

When Windows support landed, Chrome, Brave and Edge were moved onto `getBrowserDataPath`. The other Chromium hooks still point at `~/Library/Application Support/...`, and `listAvailableBrowsers` drops any browser whose Windows path is missing, so they never get as far as reading a bookmark file.

This adds Windows paths for the four I could check against a real install, and puts their hooks on the same helper:

- Brave Nightly: `%LOCALAPPDATA%\BraveSoftware\Brave-Browser-Nightly\User Data`
- Vivaldi: `%LOCALAPPDATA%\Vivaldi\User Data`
- Helium: `%LOCALAPPDATA%\imput\Helium\User Data`
- Dia: `%LOCALAPPDATA%\Packages\TheBrowserCompany.Dia_ttt1ap7aakyb4\LocalCache\Local\Dia\User Data`

Dia is the odd one because it ships as an MSIX package on Windows, so its data sits inside the package container. The suffix on that folder is the publisher hash, which is the same for everyone installing the signed package, not something per machine.

`getBrowserDataPath` hands back the macOS path on anything that isn't Windows, so macOS behaviour doesn't move.

Two things I left alone on purpose. The remaining browsers (Comet, Sidekick, Whale, the channel variants and so on) aren't installed here and I'd only be guessing at their paths. And I didn't add executable matching: Helium ships its binary as `chrome.exe`, so matching on that would have it claim Chrome's entry. The path fallback lists these browsers without an application path, which means their bookmarks open in the default browser rather than in that specific browser. Happy to add proper matching for the ones where it's unambiguous if you'd rather.

Tested on Windows 11 with Raycast 2.2.0.0. All five browsers now appear in the dropdown and their bookmarks are searchable, including Dia's 1,616.

## Checklist

- [x] I ran `npm run build` and tested this distribution build in Raycast on Windows

### DIFF
--- a/extensions/browser-bookmarks/CHANGELOG.md
+++ b/extensions/browser-bookmarks/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## [Detect more Chromium browsers on Windows] - {PR_MERGE_DATE}
 
-- Find Brave Nightly, Vivaldi and Helium on Windows instead of only looking in the macOS application support folder
+- Find Brave Nightly, Vivaldi, Helium and Dia on Windows instead of only looking in the macOS application support folder
 
 ## [Chrome Account Bookmarks] - 2026-05-20
 
--- a/extensions/browser-bookmarks/CHANGELOG.md
+++ b/extensions/browser-bookmarks/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Browser Bookmarks Changelog
 
-## [Detect more Chromium browsers on Windows] - {PR_MERGE_DATE}
+## [Detect more Chromium browsers on Windows] - 2026-09-06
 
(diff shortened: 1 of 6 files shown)


Just-Me-22:
Related: #30830. I found this while digging into the memory report in #29462, so the two touch the same extension but are independent. This one is the detection bug, that one is the render cost. We'll also update those other browsers, like for example Opera. The dia browser is also available on Windows so that has to be added.

pernielsentikaer (approved):

PR #30832, opened 2026-09-06
Add keyden extension

RUBisco0211:
## Description

A [Raycast](https://www.raycast.com/) extension for browsing, searching, and copying TOTP codes managed by [Keyden](https://github.com/tasselx/Keyden), a native macOS menu bar authenticator.

### Features

- **Grouped Accounts**: Browse TOTP accounts grouped by issuer.
- **Pinned Accounts**: Access pinned accounts from a dedicated section while keeping them in their issuer groups.
- **Instant Search**: Search across account names and issuers using Raycast's built-in search.
- **Live Countdown**: See the remaining validity period update every second with a color-coded progress indicator.

### Command: List TOTP

keyden-1

## Checklist


pernielsentikaer (changes requested):
Hi @RUBisco0211 👋

Thanks for the Keyden extension — the pinned section and the live countdown are a nice touch 💪

One thing I found while reviewing:

**Accounts without an issuer break the whole list.** Keyden's CLI prints `issuer:account`, and for a token with no issuer it prints `:account` (`listAccounts()` uses `token.account.isEmpty ? token.issuer : "\(issuer):\(account)"`, and the app can store issuer-less accounts — `otpauth://totp/user@example.com?secret=…` imports fine and only the secret is required when saving). You handle the issuer-only case now, but an empty issuer still throws `Keyden returned an incomplete account on line N.`, and since `refresh` clears the snapshot on error the command ends up on "Unable to Load TOTPs" with *no* codes available — the entries that parsed fine disappear too. I reproduced it on the current head by compiling `src/keyden.ts` and calling `parseKeydenList(":user@example.com -> 123456 (23s)")`. Could you accept an empty issuer, e.g. group it under the account name or a fallback label?

The same thing happens to any line the code regex doesn't match: the CLI prints `------` instead of a code when it can't decode a token's secret, and one such line takes down the entire list. Skipping (or showing as unavailable) the lines that don't parse would make the command much more forgiving.

Two small ones while you're in there:

- `package.json` has `"tools": []`. The repo bot treats any truthy `tools` key as an AI extension (`aiFilesOrToolsExist = !!packageJsonObj.tools` in `scripts/bots/pr-bot.ts`), which is where the `AI Extension` label comes from, and our review instructions expect an `ai` section with evals when tools exist. Since this extension has no tools, could you drop that empty array?
- `metadata/keyden-1.png` shows a red badge-style icon in the bottom bar, while the extension icon is the blue key — it looks like that capture came from an earlier build. Could you retake it with Window Capture → Save to Metadata so it matches what ships? Using a demo account instead of your own would be nice too, since the screenshot is public.

Optionally: the countdown ring assumes a 30-second period (`remainingSeconds / 30`), so a token with a custom period (Keyden stores `period` per token, e.g. 60 from an `otpauth://` URI) sits at 100% for the first half. The seconds are still correct and the CLI doesn't expose the period, so feel free to leave it — just flagging it 🙂

Let me know if anything's unclear.

---

PR #30834, opened 2026-09-06
Add channel-mixer-helper extension

AyakaNya:
## Checklist


pernielsentikaer (changes requested):
Hi @AyakaNya 👋

Thanks for the Channel Mixer Helper extension 💪 The conversion math, history view and Color Picker integration all look good, and I checked the picker callback against `raycast-cross-extension` and the Color Picker command — that part works.

I found this while reviewing:

**The Store screenshots don't pass our metadata check.** The `metadata-images` job fails on all three images. `converter.png` reports padding left 34.5% / right 50.0% / top 1.2%, `history.png` top 0.9%, `result.png` bottom 5.4% / left 47.4%, the backgrounds don't match between the images (difference 18.9 and 15.5, max 12.0), and in all three the window is cut off at the right edge. I reproduced this locally with `python3 scripts/check_metadata_images.py extensions/channel-mixer-helper`.

Could you recapture all three with Raycast's built-in **Window Capture** and **Save to Metadata**, without cropping or resizing, so the tool applies the standard padding and background? The [screenshot guidelines](https://developers.raycast.com/basics/prepare-an-extension-for-store#screenshots) cover it. `history.png` also shows `#7A213E → #222222` with `Predicted RGB 0, 0, 0`, but the current code predicts `#222222` for that pair, so that image isn't from the current version.

**Greptile's open point on history reads.** It's right that `getHistory` returns `[]` both for "nothing stored" and for a failed read, so if `LocalStorage.getItem` fails once, the next `saveHistory` writes only the new entry and drops the previous 30. Could you make a read failure distinguishable from an empty history — for example let `saveHistory` skip the write (or rethrow) when the read failed — so a transient error can't erase saved conversions?

Optional, up to you: nothing in `src/` is macOS-only and the Color Picker dependency supports Windows, so you could add `"Windows"` to `platforms` if you're comfortable testing the picker there.

---

PR #30835, opened 2026-09-06
[Telegram] Fix blank photo previews and Unknown message senders

fatherkam:
## Description

Two independent fixes to the Telegram extension. Neither changes dependencies.

**Photos rendered as a blank detail pane.** `buildMarkdownWithMedia` read each cached photo off disk and inlined it into the markdown as a base64 data URI. A Telegram photo is around 400 KB, so the embed ran to roughly 550 KB of text, and Raycast's markdown renderer drops data URIs that large without raising an error — the pane kept its layout and showed no image.

It now references the cached file with a percent-encoded `file://` URL, matching how other extensions in this repo embed local images. That also avoids re-reading and re-encoding the file on every render.

**Senders in group chats showed as "Unknown User".** `processChatMessage` called `client.getEntity()` with a bare user id. That needs a populated entity cache, but `StringSession` persists only the auth key, so every Raycast command starts with an empty cache and the call throws `Could not find the input entity` for any sender not already fetched — the `catch` then wrote "Unknown User".

It now prefers `msg.sender`, which GramJS populates from the users map on the same response: no network call and no cache dependency, with `getEntity` kept as a fallback.

Also in this PR:

- Messages sent on behalf of a channel (channel posts, anonymous group admins) previously produced no sender at all; they are now attributed.
- The user-naming logic is hoisted into a `describeUser` helper shared by both paths, matching the surrounding module-level helpers.

## Screencast

Verified against a live Telegram account: photos that previously rendered a blank detail pane now display, and group senders resolve to real names.

## Checklist

### DIFF
--- a/extensions/telegram/CHANGELOG.md
+++ b/extensions/telegram/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Telegram Changelog
 
-## [Fix Photo Previews and Message Senders] - {PR_MERGE_DATE}
+## [Fix Photo Previews and Message Senders] - 2026-09-08
 
 - Show photos in the detail pane. They were embedded as base64 data URIs, which Raycast's markdown renderer drops once they grow large, leaving the pane blank
 - Resolve message senders from the entity attached to each message rather than `client.getEntity`, which throws in a fresh command process because the session does not persist an entity cache, showing group messages as "Unknown User"
--- a/extensions/telegram/CHANGELOG.md
+++ b/extensions/telegram/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Telegram Changelog
 
+## [Fix Photo Previews and Message Senders] - {PR_MERGE_DATE}
+
+- Show photos in the detail pane. They were embedded as base64 data URIs, which Raycast's markdown renderer drops once they grow large, leaving the pane blank
+- Resolve message senders from the entity attached to each message rather than `client.getEntity`, which throws in a fresh command process because the session does not persist an entity cache, showing group messages as "Unknown User"
+- Attribute messages sent on behalf of a channel, such as channel posts and anonymous group admins
+
 ## [Add 2-Step Verification Login Support] - 2026-03-26
 
 - Handle Telegram 2FA (`SESSION_PASSWORD_NEEDED`) with a password step
--- a/extensions/telegram/src/services/telegram-client.ts
+++ b/extensions/telegram/src/services/telegram-client.ts
@@ -326,6 +326,71 @@ async function downloadProfilePhoto(
   return undefined;
 }
 
+/**
+ * Works out who authored a message, and their id.
+ *
+ * Telegram fills this in three different ways, and only one of them sets `fromId`:
+ * group members set `fromId` to a user, channel posts and anonymous group admins set
+ * it to a channel, and both broadcast posts and private chats may omit it entirely --
+ * in which case the author is whatever the message is attached to.
+ *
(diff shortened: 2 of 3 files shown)


fatherkam:
Follow-up from the Greptile review on the stacked #30836: the `file://` URL helper here only split on `/`, which breaks on Windows (the extension declares `"platforms": ["macOS", "Windows"]`, and `MEDIA_CACHE_DIR` comes from `path.join`, so the path has backslashes and a drive letter there).

Now uses Node's `pathToFileURL`, which handles both the percent-encoding the support path needs and the Windows separators. Pushed as an amend to the existing commit.

pernielsentikaer (changes requested):
Hi @fatherkam 👋

Thanks for the fixes 💪

One thing before I can approve:

Channel posts can omit `fromId`, while GramJS still populates `msg.sender` with the channel. The new branch requires `msg.fromId instanceof Api.PeerChannel`, so those posts still end up without a sender name or ID.

Could you also handle `msg.sender instanceof Api.Channel` directly, using its `id` and `title`, and cover a channel post without `fromId` in a regression test?

Happy to approve once that’s in 🙂

---

fatherkam:
Thanks @pernielsentikaer — you're right, and it was worse than just channel posts.

The `!msg.fromId` branch only handled `chatEntity instanceof Api.User`, so **any** message without a `fromId` whose author is a channel fell through every branch and came out with no name and no id at all.

Rather than add a fourth condition to a chain that was already hard to reason about, I pulled the attribution rules out into a pure function. It has no network calls, so the rules are directly testable:

```ts
export function resolveMessageAuthor(
  msg: Api.Message,
  chatEntity?: Api.User | Api.Chat | Api.Channel,
): { senderId?: string; entity?: Api.User | Api.Channel } {
  if (msg.fromId instanceof Api.PeerUser) { /* ... */ }

  if (msg.fromId instanceof Api.PeerChannel) { /* ... */ }

  if (!msg.fromId) {
    const author = msg.sender ?? chatEntity;
    if (author instanceof Api.Channel || author instanceof Api.User) {
      return { senderId: author.id.toString(), entity: author };
    }
  }

  return {};
}
```

`msg.sender ?? chatEntity` covers the case you raised and one more: a channel post where `sender` isn't attached either, which now falls back to the chat entity the same way private chats already did. `processChatMessage` then branches on the returned entity — `title` and the channel photo for a `Channel`, the existing `describeUser` path for a `User` — and only falls back to `client.getEntity()` when the response didn't carry the user with it.

### On the regression test

There is one, covering a channel post with no `fromId`, plus five more around it. I checked it actually catches this: reverting to the old condition fails exactly those two channel cases and nothing else.

The catch is that it lives in **#30837**, the last PR in the stack, because that's where the test harness is added. I deliberately kept vitest out of this PR — adding it brings a devDependency and a ~3,000-line `package-lock.json` diff into what is otherwise a dependency-free fix, and I didn't want to couple these fixes to a decision about whether you want a test runner in extensions at all.

Happy to go either way:

- **Leave it** — the test is in #30837 and I've linked it here.
- **Move the harness into this PR** — say the word and I'll rebase it down, and this PR becomes self-contained at the cost of the lockfile diff.

Whichever you prefer. I'll leave this as a draft until you've had a chance to say which.

pernielsentikaer:
Thanks @fatherkam 💪 The channel-attribution fix looks good. Keeping the test harness in #30837 is fine—please mark this PR ready for review.

pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #30836, opened 2026-09-06
[Telegram] Fix bot messages showing as Unknown (migrate archived GramJS to teleproto)

fatherkam:
## Description

Closes #30817.

Messages from bots appeared as **"Unknown"** in the list, or vanished from it entirely. There are two causes, one on the wire and one in this extension.

### Telegram is not sending the content

Telegram added rich messages for bots in **API layer 228**. The `telegram` package (GramJS) is pinned to **layer 198**, and rather than downgrading newer content the server withholds it, substituting `messageMediaUnsupported` with the text stripped.

Captured from a live bot chat, every affected message arrives as:

```
out: false, hasText: false, textLength: 0,
mediaClass: "MessageMediaUnsupported"
```

`parseMessageMedia` maps that to type `"unknown"`, and the list item falls back to the media label — rendering the title as literally "Unknown". No parser change can recover this, because the bytes are never delivered. The client logs `Using LAYER 198 for initial connect` on every connection.

**GramJS has been archived by its authors.** Its npm deprecation notice reads:

> This package is archived and no longer maintained. Development continues in teleproto, a largely compatible, actively maintained fork.

This PR follows that instruction. teleproto is on layer 229. The swap is five import lines — `TelegramClient`, `StringSession`, `Api` and `computeCheck` are unchanged, and existing `StringSession` strings are read as-is, so **users stay logged in**. It typechecks with no other changes; `sendReadAcknowledge`, the one renamed method, is not used here.

For reviewer confidence: **this extension is the only consumer of GramJS in the repository** — nothing else is affected.

### The extension cannot read the content

With delivery restored, the text still needs reading. Rich message content lives in `message.richMessage` as a block tree, not in the flat `message.message` string, so both filters that decide whether a message is worth showing —

```ts
messages.filter((msg) => msg.message || msg.media)
```

— discarded every one of them. On the test chat that was **50 fetched, 30 kept**: the missing 20 were the bot's messages.

They are now read through teleproto's `Rich` helpers: whitespace-collapsed plain text for list titles, and Markdown for the detail pane so the bot's formatting survives.

## Screencast

Verified against a live bot chat: messages that rendered as "Unknown" — and then disappeared once delivery was restored — now show their real text, with formatting preserved in the detail pane.

## Checklist

- [x] I read the [extension guidelines](https://developers.raycast.com/basics/prepare-an-extension-for-store)
- [x] I read the [documentation about publishing](https://developers.raycast.com/basics/publish-an-extension)
[...]

PR #30837, opened 2026-09-06
[Telegram] Add a vitest suite

fatherkam:
## Description

Adds a `vitest` suite and an `npm test` script.

> **Stacked on #30836.** Please merge that first; the diff here will then reduce to the test files alone. (#30835 has merged.)

**No production behaviour changes.** The only edits outside test files are three functions in `telegram-client.ts` changed from module-private to exported, plus the new script and dev dependency.

23 tests across four files, chosen to guard behaviour that has actually broken rather than to chase coverage:

| Under test | Guards against |
|---|---|
| `renderMessageContent`, `hasRenderableContent` | Rich messages carry their text outside `msg.message`; filtering on `message \|\| media` alone silently dropped every one of them |
| `buildMarkdownWithMedia` | Photos being re-inlined as base64, which rendered a blank detail pane; asserts a percent-encoded `file://` URL |
| `parseMessageMedia` | The fallthrough that turns unrecognised media into `"unknown"` |
| `groupChatsByPinned`, `groupMessagesByDate`, `extractUrlFromText` | Existing pure helpers that had no cover |

The suite was checked to actually fail on regression: restoring the old `message || media` filter fails exactly one test and no others.

### One note on the setup

`@raycast/api` ships **types only** and has no runtime entry, so a test runner cannot resolve it — even `vi.mock` needs the specifier to resolve first. `vitest.config.ts` aliases it to `test/raycast-api.stub.ts`, which supplies the two members `telegram-client` touches at import time (`environment.supportPath` and `LocalStorage`). Happy to change the approach if there's a house convention here.

Also worth flagging: **no workflow in this repository runs tests**, so this is a guard for contributors running `npm test` locally rather than a CI gate. Let me know if you'd rather not carry it.

## Checklist


0xdhrv (commented):
Hey @fatherkam 👋

Could you resolve the merge conflict?

---

pernielsentikaer:
Hi @fatherkam 👋

Thanks for adding the test suite 💪 Lint, build, and all 30 tests pass locally, and I confirmed the regression claim (reverting the `message || media` filter fails exactly one test).

Two small notes, neither blocking:

- The description says 23 tests; the suite now has 30.
- The stub lives at `test/raycast-api.stub.ts`, but the repo convention is `src/__mocks__/@raycast/api.ts` aliased in the Vitest config (e.g. `extensions/clip`, `extensions/dagster`). Keeping it under `src/` also puts it inside `tsconfig.json`'s `include`. No strong rule here, just flagging since you asked 🙂

Nothing else from me.

pernielsentikaer (approved):
Looks good to me, approved 🔥

fatherkam:
@raycastbot I linked my Github profile to my Raycast account as requested.

PR #30838, opened 2026-09-06
Add bLine Clipboard extension

bhivedev:
## Description

**bLine Clipboard** searches the clipboard history that [bLine](https://bhive.software/bline) keeps in sync across a person's Macs, from Raycast: text, links, images and screenshots (searchable by the text inside them), from every Mac in their bLine.

One command, **Search Clipboard History**. Typing filters by a clip's text, its kind, the app it came from, the Mac it came from, or the words inside a screenshot. `⏎` copies the clip, `⌘⏎` pastes it in the frontmost app, `⌘Y` opens Quick Look, `⇧⌘P` pins it on every Mac, `⌃X` deletes it from every Mac, `⇧⌘D` toggles the preview pane.

**Requires bLine to be installed** (`/Applications/bLine.app`, build 63 or later). The extension reads history through bLine's own command-line helper, which ships inside the app, so nothing leaves the Mac and there is no account or token. Without bLine the command shows an error saying so. No binaries are downloaded by the extension.

**On the title:** `ray lint` asks for "BLine Clipboard" in title case. The product is written *bLine* (the b is lower case, as in the app, its site and its bundle); if that is a blocker I will change it.

## Screencast

Three screenshots are in `metadata/` (2000×1250): the list with the preview pane, a filtered search, and the actions panel. They were taken on a throwaway demo history.

## Checklist

- [x] I ran `npm run build` and [tested this distribution build in Raycast](https://developers.raycast.com/basics/prepare-an-extension-for-store#metadata-and-configuration) — `npm run build` succeeds; the extension was exercised in Raycast from the same source via `ray develop` on macOS 27, which is where the screenshots come from


bhivedev:
Addressed the five automated findings in the two follow-up commits:

- A failed history load now shows an empty view naming the reason (bLine not installed, or the helper's error) with a Try Again action, instead of an empty list.
- "Paste in Frontmost App" on an image now pastes the materialised file via `Clipboard.paste({ file })`, rather than only copying.
- The search haystack includes the clip's kind and the app it came from, as the README says.
- `getPreferenceValues` uses the generated `Preferences.SearchHistory` type; `tsconfig` includes `raycast-env.d.ts` so it resolves.
- README: Return copies, Command-Return pastes.

`npm run build` and `npm run lint` are clean apart from the title-case warning noted in the description.

bhivedev:
Withdrawing this submission — bLine is no longer being distributed, so there's nothing to review. Thank you for your time.

PR #30840, opened 2026-09-06
Add streakify extension

SerhiiNadoienko:
## Description

Track and manage daily habits directly from Raycast and your menu bar with custom emojis and strict streak rules.

## Screencast
Screencast preview 1

Screencast preview 1

Screencast preview 3


pernielsentikaer (changes requested):
Hi @SerhiiNadoienko 👋

Thanks for the submission and the screencast 💪 The extension builds and lints cleanly on my side, and the streak/undo/edit logic checks out.

I found these while reviewing:

**Screenshots are missing** — `create-streak` and `manage-streaks` are `view` commands, but there's no `metadata/` folder in `extensions/streakify`. Extensions with view commands need screenshots captured with Raycast's Window Capture tool (tick **Save to Metadata**). Could you add at least three? The [screenshot guidelines](https://developers.raycast.com/basics/prepare-an-extension-for-store#screenshots) cover the format — 2000×1250, same background across all shots.

**`launchCommand` is unhandled** — the three calls in `src/menu-bar.tsx` aren't wrapped, so a failed launch goes unnoticed. Could you wrap them, e.g.:

```tsx
try {
  await launchCommand({ name: "manage-streaks", type: LaunchType.UserInitiated });
} catch (error) {
  console.error("Failed to launch Manage Streaks", error);
}
```

One thing from the earlier Greptile thread, which I don't consider blocking: `writeQueue` in `src/storage.ts` serializes writes inside a single command process, but the menu-bar and view commands run as separate processes, so a simultaneous write from both could still drop a change. Did you consider that, or is the small window acceptable to you?

---

PR #30843, opened 2026-09-06
Add People extension

mathiswrong:
People is a contacts app for Mac, iPhone, and iPad. This extension searches the Mac app's authenticated, read-only local API and displays contacts and their fields in Raycast.

Users can launch People directly from Raycast before configuring an access key. Searching starts People when needed and retries while the app starts. Setup and connection errors include actions to open People, configure the key, or retry.

Commands:

- **Open People**: launch People or bring its main window forward, including before access-key setup.
- **Search People**: search by name, company, email address, or phone number; browse and copy contact fields; compose email, call, open links or addresses, or open the contact in People.
- **Report a Bug**: open a pre-addressed support email.

Requires People installed on the same Mac and subscription access for contact searches. Contact requests require a per-launcher access key. Setup is described in the README.

## Validation

- `npm test`: seven tests pass, covering cold startup, startup timeout, authentication errors, cancellation, and launch failure.
- `npm run build` (including an explicit `tsc --noEmit`) and `npm run lint`: pass.
- Manual testing of the new launch and recovery paths in Raycast remains pending.

## Open review item

The existing API sends a bearer key to a fixed loopback port. The server-authentication concern remains unresolved and requires a coordinated app and extension change.

## Screenshots

Existing contact-search screenshots are in `metadata/`.


pernielsentikaer (changes requested):
Hi @mathiswrong 👋

Thanks for the extension and the follow-up commits 💪 I checked out the branch and ran the tests, `ray build`, `tsc --noEmit`, `ray lint`, and the metadata validator locally — all pass, and the launch/retry handling reads well.

I found this while reviewing:

**The access key is sent to the loopback port before we know People owns it.** In `src/people-client.ts`, `requestPeople` puts `Authorization: Bearer ${accessKey}` on the very first request to `http://127.0.0.1:47631/v2...`, and the port is contacted without any check that the listener is actually People. If another local process happens to hold 47631 while People is closed, it receives the persistent key — and that key then lets it read the user's contacts through People's own API. Since the app already exposes this API, could you close this together with the app-side change, for example a handshake where the server proves it knows the key before the client sends it (or verifying the listening process before the first request)? Your PR description already flags it as unresolved, and I'd like it settled before we list the extension.

A few optional things, no rush on any of them:

- The three screenshots are all the search list with a different query. One showing the contact detail view (fields plus the email/phone/link/address actions) would show off the main value of the extension.
- A rejected key (401), a missing app, and an exhausted retry budget all end up as "People is unavailable" with the "copy your key" description. Distinguishing a bad key from "the app isn't running" would point people at the right fix.
- Each request uses a 1s timeout, and a timeout is handled the same as "app not running", so a slow first search on a large contact index can end in the unavailable state. If People can take more than a second, a longer first attempt would avoid that.
- `version` is `2.0.0` with a "Launch People from Raycast" entry above "Initial Version" — is this shipping on top of an earlier build for your app's users? Just want to be sure there's no earlier listing I should account for.

---

PR #30858, opened 2026-09-07
Add skills-manager extension

TDHSC:
## Description

Adds **Skills Manager**, a front end for the [Skills Manager](https://github.com/xingkongliang/skills-manager) desktop app and its `skills-manager-cli` binary, which maintains one shared library of AI agent skills and deploys them into Claude Code, Codex, Cursor and 50+ other agents.

The extension implements no logic of its own. Every action shells out to the CLI in `--json` mode and renders the result with native Raycast components; the CLI shares its SQLite database and repository lock with the desktop app, so state stays consistent between the two.

**Commands**

| Command | What it does |
| --- | --- |
| My Skills | Browse the library, filter by agent / preset / tag / source, read a skill's `SKILL.md`, deploy or undeploy per agent, manage tags, remove |
| Search Skills | Search the skills.sh marketplace and install, with the deploy step offered immediately |
| Presets | Deploy, undeploy and edit the membership of skill presets |
| Agents | Inspect detected agents and toggle which ones receive skills |
| Check for Skill Updates | Probe upstream for updates and apply the ones you pick |

**Requirements.** The Skills Manager app must be installed (`brew install --cask skills-manager`). Opening it once publishes a version-matched CLI to `~/.skills-manager/bin`, which the extension prefers; a standalone CLI on `PATH` also works. When neither is present the extension shows a guidance screen instead of failing, and there is a `CLI Path` preference for unusual installs.

**Notes for review**

- Destructive commands (`skills remove`, `presets delete`, `agents disable`) are always gated behind a `confirmAlert` before `--yes` is passed. Enabling an agent also confirms first, because the CLI syncs the legacy active preset into it immediately.
- A refused deployment (`TARGET_CONFLICT`) and a held-back update (`held_back_removals`) each get a dedicated screen rather than a toast: both need to say which paths are involved and that nothing was changed.
- Reading a skill is the default (Return) action in every list, so no bare Return mutates a deployment.
- Out of scope on purpose: the CLI's `git` command group, the legacy exclusive-sync workflow, the deprecated `skills enable/disable`, and `set-source --force`.

## Screencast

To follow — the extension drives a local CLI against the author's personal skill library, so a clean recording is being prepared separately. Happy to provide one on request if that blocks review.

## Checklist


pernielsentikaer (changes requested):
Hi @TDHSC 👋

Thanks for the very thorough write-up and the CLI-only design — the notes around refused deploys and held-back updates made this quick to follow 💪

**Overlap with an existing extension.** We already have [Skills](https://www.raycast.com/keito4/skills) in the Store by @keito4, which searches the skills.sh marketplace, installs for all supported agents, lists/removes installed skills per agent, and checks for updates. Your central library with a separate deploy step, presets, tags and the Skills Manager app state look like the real differences. Could you look into enhancing that extension instead of adding a separate one? If there's a difference I'm missing, could you explain it here? 🙂

---

PR #30861, opened 2026-09-07
Add kloak extension

naetikarvind:
## Checklist


pernielsentikaer (changes requested):
Hi @naetikarvind 👋

Thanks for the Kloak extension 💪 The command layout is clear, the build and metadata checks pass, and the concealed-copy/auto-clear flow is a nice touch.

I found a couple of things while reviewing:

**Stale import content.** In `src/import-vault.tsx`, if a newly selected file fails to load (`handleFilesChange`'s `catch`) or the picker is cleared, the old `content`/`format` stay in state. Since `handleImport` only reads `files[0]` when `content` is empty, pressing **Run Import** then sends the previously loaded export to `vault.import` instead of the file just selected. Could you clear `content`/`format` on read failure and on empty selection — or track the loaded file and refuse to import when it doesn't match the current `files[0]`?

**Favorites can't be saved.** `src/search-vault.tsx` calls `requestDaemon("vault.toggleFavorite", { id: item.id })`, but the Kloak daemon's RPC switch doesn't implement `vault.toggleFavorite`, so the optimistic update gets rolled back and users see "Failed to update favorite". The daemon supports `vault.updateItem`, so `requestDaemon("vault.updateItem", { id: item.id, updates: { favorite: updatedFav } })` should work.

Two smaller things: the `vault.generateTotp` fallback passes `{ secret, details }` while the daemon reads `params.options`, so non-default algorithm/digits/period are ignored on that path — passing `options` instead would fix it. And the README's Requirements section needs a link showing where users can get the Kloak app/daemon, since the extension can't work without it.

---

PR #30862, opened 2026-09-07
Add AWS frontier agents to Amazon AWS console list

matanbaruch:
## Description

Adds the AWS frontier agents to the `Open AWS Console` service list: AWS Security Agent, AWS DevOps Agent and AWS FinOps Agent. They all shipped after that list was last touched, so right now there is no way to reach any of them from Raycast.

Console targets, and how I checked them (an unknown prefix on `console.aws.amazon.com` returns 404, a real one returns 302):

1. `AWS Security Agent` goes to `/securityagent`. That is the exact URL behind the "Sign in" and "Get started today in the console" buttons on [aws.amazon.com/security-agent/getting-started](https://aws.amazon.com/security-agent/getting-started/).
2. `AWS DevOps Agent` goes to `/aidevops/home`, taken from the console link on [aws.amazon.com/devops-agent](https://aws.amazon.com/devops-agent/).
3. `AWS FinOps Agent` goes to `https://us-east-1.console.aws.amazon.com/finops-agent/home?region=us-east-1`. Full URL on purpose: the agent only runs in us-east-1 and `AWS_URL_BASE` carries no region, so a bare path would drop you into whatever region you used last. Same pattern as the existing Global Accelerator entry.

Kiro is the fourth frontier agent, but it has no `console.aws.amazon.com` prefix at all (I probed a dozen candidates, all 404), so I left it out rather than pointing this command at kiro.dev.

Two notes on the diff:

- The icons are the official service icons from the AWS console itself, nothing hand made.
- The entries are appended at the end of `aws-services.json`, which is how the newer services in that file (bedrock, grafana, kendra, ivs and friends) were added.

`frontier` is in the `match` string for all three so they come up as a group.

## Screencast

Data-only change to the console list, no UI change.

## Checklist

`npm run build` is clean. I did not load the dev build into the app, the only change is three rows in `aws-services.json` plus their icons, and the command that reads them is untouched.


matanbaruch:
@pernielsentikaer @0xdhrv quick review when you have a minute? Data-only change: three console entries plus their icons in `aws-services.json`, all checks green.

pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #30873, opened 2026-09-07
Add quick-groups extension

neilbartlett:
## Description

Quick Groups organizes Raycast around user-defined things and the actions associated with them. Users describe projects, machines, writing, clients, or other concepts in local YAML files, search for a thing once, and then open its files and URLs, launch it in a macOS application, connect over SSH, open an Obsidian note, copy protected values, or pass values to Raycast Script Commands.

The extension works immediately with an editable example, supports an optional custom groups directory, reloads files as they change, provides actionable YAML diagnostics, and supports sibling value substitution.

## Screenshots

Three Store screenshots are included in the extension metadata. They show the searchable group overview, actions on a writing project including opening it in Scrivener, and source-editing and reload actions.

## Checklist

- [x] I ran `npm run build` and tested the distribution build in Raycast


neilbartlett:
Addressed the automated review findings in the latest update:

- Propagate sensitivity when an explicit `value` is derived from a sensitive sibling; added a regression test ensuring the value is masked and excluded from searchable values.
- Version asynchronous directory loads so stale watcher/manual reloads and loads from a previous directory cannot overwrite newer state.
- Updated Prettier to the repository-required `printWidth: 120` and `singleQuote: false` settings, then reformatted the extension.
- Wrapped the ESLint flat configuration with `defineConfig`.

Verified locally with 44 tests, `npm run lint`, and `npm run build`.

pernielsentikaer (changes requested):
Hi @neilbartlett 👋

Thanks for this one — the YAML model, the diagnostics, and the first-run example all read really well, and the tests are a nice touch 💪

I found this while reviewing:

A scalar field named `pwd` isn't masked. `src/parser.ts` decides scalar sensitivity from `SENSITIVE_PATTERN`, which doesn't include `pwd`, while the `pwd` action registered in `src/actions/built-ins.tsx` is marked `sensitive: true`. So this:

```yaml
services:
  router:
    pwd: hunter2
```

renders `hunter2` in plaintext and includes it in the search index. The mapping form (`login: { value: admin, pwd: hunter2 }`) is handled correctly, so this is just the scalar case slipping past the same rule.

Could you make the scalar label check consult the registry too?

```ts
const sensitiveLabel =
  SENSITIVE_PATTERN.test(label) || getActionDefinition(label)?.sensitive === true;
```

That keeps scalar and mapping fields consistent and also covers custom actions registered with `sensitive: true`. While you're there, a `pwd` regression test in `tests/parser.test.ts` would be great.

Two small optional notes, no need to hold the PR for them: a `"subtitle": "Quick Groups"` on each command would help in Raycast root search, and the collection dropdown keeps a selected collection after a reload even if that collection no longer exists.

Thanks again 🙂

---

neilbartlett:
Hi @pernielsentikaer — Thx for the review—much appreciated. Agreed on scalar `pwd`: def an inconsistency. Scalar fields now consult the registry, and I added regression coverage checking that `pwd` values are masked and excluded from search.

Thx for the optional suggestions. I’ve added both: the commands now have a **Quick Groups** subtitle, and the collection filter resets if its selected collection disappears after a reload (that last one was prob more of a bug!).

All 45 tests pass, along with lint and the production build. Thanks again.

pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #30875, opened 2026-09-07
Add Macs Fan Control extension

dw2lam:
## Description

Adds **Macs Fan Control**, an extension for driving the [Macs Fan Control](https://crystalidea.com/macs-fan-control) app from Raycast.

Commands:

| Command | What it does |
| --- | --- |
| **Start Fan** | Live dashboard — app on/off, fan RPM, grouped temperature sensors, and every preset. |
| **Start Fan Control** | Launches Macs Fan Control (optionally toggles it off again). |
| **Start Fan Auto** | Hands every fan back to the system. |
| **Start Fan Full Blast** | Spins every fan to maximum. |
| **Start Fan Preset** | Applies any preset by name, including the user's own. |
| **Set Fan Speed** | Pins the fans to a specific RPM. |
| **Create Fan Preset** | Saves a new named preset into Macs Fan Control. |

### How it controls the fans

Writing to the SMC needs root, so **this extension never does it**. Every change is written into Macs
Fan Control's own preferences and the app is restarted, so it applies the preset through its own
privileged helper — the same path as clicking a preset in the app. The extension asks for no elevated
privileges.

Presets are read live from the app, so whatever the user has saved appears in Raycast. Sensor-based
fan rules are shown read-only and written back byte-identical, so presets built in the app are never
corrupted. Preset writes refuse to run on an unreliable read and keep rolling backups of the
preference domain.

### About the one bundled binary

`assets/smc-reader` reads live fan speeds and temperatures. The SMC has no shell or scripting
interface, so this needs native code. Please note:

- **Full source is in this PR** — `native/smc.c` (~200 lines) and `native/Makefile`.
- **Read-only by construction.** It issues only SMC *read* commands; it never writes an SMC key and
  therefore cannot change fan behaviour.
- **Reproducible.** `npm run native` rebuilds it (clang, universal arm64 + x86_64). Its SHA-256 is
  published in the README so it can be verified against a local rebuild.

Happy to change the approach if you'd rather it were fetched or handled differently.

### Hardware detection

Fan count comes from the SMC's own `FNum` key rather than a table of Mac models, so it is correct on
fanless Macs and on models newer than this extension. Fan locations use the `F<n>ID` key where the
hardware publishes it (Intel), falling back to positional names on Apple silicon, which has no such key.

## Screencast

Three screenshots are in `metadata/`: the dashboard (app status, live fan RPM, presets), the
temperature detail pane, and the speed form.

They are cropped just above Raycast's footer so the green *local extension* badge — which only
appears for a locally-installed extension and won't exist once this is published — is not in
frame. Verified against `scripts/check_metadata_images.py` before pushing.

## Checklist


dw2lam:
Thanks — most of these were real. Fixed in `060d42c`:

**1. Failed mutations left the app stopped** — valid, and the worst of the batch. `mutateAndRestart` now relaunches in a `finally`, so a failed write no longer leaves Macs Fan Control quit and every fan silently handed back to the system.

**2. Launch timeout reported success** — valid. `launchApp` now throws when the app never appears, so a preset can't be reported as applied while nothing is running.

**3. Stale index targeted the wrong preset** — valid. Rename and delete now resolve the target by name against the freshly-read list, only trusting the old slot if it still holds that name, and erroring otherwise.

**4. SMC errors looked like fanless hardware** — valid. Both the dashboard and the preset editor now separate `usePromise`'s error from an empty result; "no controllable fans" is only shown after a successful read.

**5. Prettier width** — fixed to 120.

**6. Duplicated preference types** — fixed; both call sites now use the generated `Preferences` type.

**7. Deeplink vs `launchCommand`** — partially. `launchCommand` only launches a command from inside a running one, so it can't back an `Action.CreateQuicklink`, which needs a URL the user triggers later from root search. I've removed the fragility instead: the author and extension name are now read from the manifest rather than written into the string, so they can't drift. Happy to change this if you'd prefer a different approach.

`npm run lint` and `npm run build` are clean, and the two suites in `scripts/` (a 21-check end-to-end run against a real Macs Fan Control install, and a regression test for preset loss) pass against the updated signatures.

dw2lam:
Follow-up on the `mutateAndRestart` error handling, since the last two review passes asked for opposite things:

- pass 2 flagged that the relaunch error *replaced* the mutation error;
- pass 3 flagged that the mutation error then *hid* the relaunch error.

Both are fair, and the stable answer is neither — it's to surface both. `25f8c01` now captures the two failures independently and, when both occur, raises a single message combining them: the mutation error says whether the preferences actually changed, and the relaunch note says Macs Fan Control is stopped so the fans are back under system control. When only one fails, that error is raised unchanged. The relaunch is still always attempted.

That's where I'll leave the automated pass unless a maintainer wants something different. Lint and build are clean and both suites in `scripts/` pass.

pernielsentikaer (changes requested):
Hi @dw2lam 👋

Thanks for the thorough work here — the guarded preference writes, rolling backups, and the read-only native helper all look solid. I verified the build, lint, and metadata checks pass, the bundled binary's hash matches the README, and the earlier review findings are addressed. I couldn't run this against real hardware, so these are optional notes rather than blockers.

A few small things:

- **`src/start-fan-preset.tsx`** — `getCustomPresets()` is awaited without a `try/catch`, so a failed preference read surfaces a generic Raycast error instead of the HUD the other no-view commands use. Easy to align with `applyPresetAsHud`.
- **Concurrent preset writes** — `resolveTarget` fixes the stale-index case, but two overlapping invocations can still lose an added preset: both read the list and the last write wins, while the removal guard only sees each caller's own snapshot. A small serialization or a re-read right before the write would close it. Edge case, so fine to leave if you'd rather not.
- **Screenshots** — the three images pass `check_metadata_images.py`, but the top/bottom padding is ~15% rather than the ~12% *Save to Metadata* produces. Could you confirm they came from Raycast's Window Capture and weren't cropped afterwards?
- `headlineTemperature` in `src/lib/smc.ts` and `APP_STORE_URL` in `src/lib/actions.ts` look unused; `APP_STORE_URL` points at a Softonic mirror rather than the vendor page, so worth dropping or switching to `https://crystalidea.com/macs-fan-control`.

Nothing here blocks a merge 🙂

---

Issue #28716, opened 2026-06-11
[Clockify] "Timer could not be started"


daviccarneiro:
### Extension

https://www.raycast.com/isma/clockify

### Raycast Version

0.64.3.0

### OS Version

macOS Tahoe 26.5.1 arm64

### Description

Unable to use the "Time tracking" command

https://cln.sh/pkFhstJ5

### Steps To Reproduce

1. Start "Time Tracking" command
2. Select "Start new timer"
3. Add description (mandatory for my workspace, all other fields are left blank, but adding them does not change anything)
4. Click "Start"
5. See error

### Current Behaviour

Error: "TImer could not be started"
Unfortunately does not allow me to generate a log

### Expected Behaviour

Tracking starts

SrAnthony:
Hitting the same issue :(

LeszekBlazewski:
Hitting same problem, did anyone manage to resolve it?

tonygauderman:
This may be the same root cause as #29613 — PR https://github.com/raycast/extensions/pull/30877 could fix it.

Both project dropdowns in `src/index.tsx` declared `onChange` *after* spreading `itemProps`, which replaced the `onChange` that `useForm` supplies. `values.projectId` therefore never got set, so `onSubmit` ran with no project, `POST /time-entries` was rejected, and `addNewTimeEntry` showed "Timer could not be started". The visible symptom is the project dropdown snapping back to the first entry every time you pick one.

If your dropdown was doing that too, this is very likely the same bug. If your project selection appeared to stick and it still failed, then it's something else and worth keeping open.

tonygauderman:
Follow-up: there is a second, separate bug that can produce this toast, and PR https://github.com/raycast/extensions/pull/30894 fixes it. I can't confirm from the reports here whether it's what you hit, so I'm not claiming it as a fix — but the symptom matches.

`useConfig` writes `userId`, `workspaceId` and `name` into `LocalStorage` without awaiting the writes, and every helper in `utils.ts` reads those ids back out of `LocalStorage` independently. A resolved `setItem` doesn't guarantee the key is readable, and `useConfig` can bootstrap from more than one mounted component at once — under that concurrency a write gets dropped:

```
wrote workspaceId="63c59a0a…c9"
getTimeEntries read workspaceId=undefined userId="63c59a0a…c8"
  | allItems keys=["name","userId"] allItems.workspaceId=undefined
POST /workspaces/undefined/time-entries -> 403
[...]

PR #30877, opened 2026-09-07
Clockify: fix project dropdown resetting selection and timer failing to start

### LINKED ISSUE #29613: [Clockify] Add Time Entry
### Extension

https://www.raycast.com/isma/clockify

### Raycast Version

0.69.0.0 (Raycast 2 Beta)

### OS Version

MacOS Tahoe 26.5.2

### Description

I can't select any projects anymore. Every time it resets itself to the 1e one in the list.
I know i'm using Raycast 2 Beta but all my other extensions are working fine.

### Steps To Reproduce

1. Open Raycast 2 Beta
2. Time Tracking
3. Add Time Entry
4. Select any other project

### Current Behaviour

Project resets itself to the 1e in the list.

### Expected Behaviour

Select any project I want.

tonygauderman:
## Description

Fixes the project dropdown in the Clockify extension resetting to the first project on every selection, which also made "Start Timer" fail with "Timer could not be started".

### Root cause

Both project dropdowns declared `onChange` *after* spreading `itemProps`:

```tsx
<Form.Dropdown {...itemProps.projectId} title="Project" onChange={(projectId) => fetchTasksForProject(projectId)}>
```

Because the later prop wins, this replaced the `onChange` that `useForm` supplies via `itemProps`. `values.projectId` was therefore never written. `itemProps` returns a *controlled* `value` (`useForm` deliberately returns `null` rather than `undefined` so the field isn't uncontrolled), so on the next render the dropdown was forced back to `null`, the selection snapped to the first item, and that re-emitted `onChange` — an unbounded loop of `/projects/{id}/tasks` requests, which in turn triggers Clockify rate limiting (HTTP 429).

The same missing write is why submitting the form failed: `onSubmit` received no `projectId`, so `POST /time-entries` was rejected and `addNewTimeEntry` showed "Timer could not be started".

### Fix

Drive the task fetch off the form value instead of intercepting the dropdown's `onChange`, leaving `useForm`'s handler intact:

```tsx
useEffect(() => {
  if (values.projectId) fetchTasksForProject(values.projectId);
}, [values.projectId]);
```

Applied to both `NewEntry` ("Start New Timer") and `AddTimeEntry` ("Add Time Entry"), which had identical code. No other `itemProps` spread in the extension overrides a handler.

Likely fixes #29613, and may be the same root cause behind #28716.

## Checklist

### DIFF
--- a/extensions/clockify/CHANGELOG.md
+++ b/extensions/clockify/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Clockify Changelog
 
+## [Fix project selection resetting and timer failing to start] - {PR_MERGE_DATE}
+
+- Fixed the project dropdown resetting to the first project on every selection in both "Start New Timer" and "Add Time Entry".
+- Fixed "Timer could not be started" caused by the selected project never reaching the form values.
+- Fixed the resulting loop of task requests, which triggered Clockify rate limiting (HTTP 429).
+
 ## [Fix cached workspace data] - 2026-05-15
 
 - Fixed cached projects, tags, and tasks falling back incorrectly when Clockify returns no data.
--- a/extensions/clockify/CHANGELOG.md
+++ b/extensions/clockify/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Clockify Changelog
 
-## [Fix project selection resetting and timer failing to start] - {PR_MERGE_DATE}
+## [Fix project selection resetting and timer failing to start] - 2026-09-08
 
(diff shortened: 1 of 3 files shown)


tonygauderman:
Screen recording of working after the fix.  This process failed prior because the project list was hitting the API too frequently and getting 429s back.

https://github.com/user-attachments/assets/62d97f57-8db5-471a-8578-d3b6a2fa3c32

pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #30887, opened 2026-09-08
Add Better JSON extension

XiaoDaiGua-Ray:
## Description

Better JSON is a Raycast developer tool for formatting, inspecting, searching, deserializing, and serializing JSON from the clipboard or manual input. It opens valid clipboard JSON directly, browses objects and arrays one level at a time, and shows complete decoded previews for nested JSON.

## Validation

- [x] Read the extension guidelines
- [x] Ran npm run build
- [x] Ran npm test (21 tests passed)
- [x] Ran npm run lint
- [x] Tested the distribution build in Raycast 2.2
- [x] Added a custom 512x512 PNG icon with transparent margins

## Notes

Nested JSON deserialization is iterative and unlimited in depth while ordinary strings such as \"123\" and \"true\" remain strings. Serialize JSON applies one JSON.stringify operation to the fully decoded value. The original input is retained for restoration and copying.


0xdhrv (commented):

We already have an extension in the Store that covers this JSON workflow:

* [Format JSON](https://www.raycast.com/destiner/json-format) by [@destiner](https://github.com/destiner)

The existing extension already includes **Format JSON**, formatting from the clipboard, formatting selected text, displaying formatted JSON in a viewer, converting JSON arrays to JSONLines, and converting text into a valid JSON value. It also already parses stringified JSON as part of its formatting flow.

- A hierarchical List that browses objects and arrays one level at a time
- Search across keys, values, and paths, plus copy of the current node
- Automatic nested JSON string deserialization, with the original input kept for restore

However, these appear to be natural enhancements to the existing Format JSON viewer rather than enough differentiation for a separate extension with the same primary workflow. Both take JSON from the clipboard or a form, inspect or re-serialize it, and copy the result.

Could we consider contributing the tree browser and nested deserialization to the existing extension instead?

If there is a technical or workflow-related reason why this needs to remain separate, please share more context about the distinction and the use cases that cannot be supported there.

> Consolidating this functionality would help avoid duplication, reduce confusion for users, and keep closely related JSON formatting and inspection workflows maintained in one place.
>
> See our guidance on [contributing to an existing extension versus creating a new one ↗](https://developers.raycast.com/basics/prepare-an-extension-for-store#contributing-to-existing-extensions-vs-creating-a-new-one).

0xdhrv:

XiaoDaiGua-Ray:
Thank you for the detailed review. I agree that **Format JSON** is valuable, and I do not want to duplicate its quick format-and-return workflow. It already provides capabilities that Better JSON intentionally does not try to replace:

- format a clipboard value or selected editor text and copy/paste it back;
- accept JSON/JS object input with configurable indentation;
- convert arrays to JSONLines and arbitrary text into a valid JSON value;
- show a formatted-code result, including parsing a top-level stringified JSON payload.

The distinction is the primary job:

| | Format JSON | Better JSON |
| --- | --- | --- |
| Primary workflow | Transform → copy/paste | Inspect → locate → extract |
| Unit of work | One whole clipboard value or selection | A retained JSON document and any selected subtree |
| Navigation | Formatted code result | Hierarchical, one-level browser with a live decoded preview |
| Finding data | Inspect the text | Search keys, primitive values, and full paths globally; filter by type |
| Output | Formatted whole result | Copy formatted or compact JSON, a raw string, a path, original input, or only the current node |

Better JSON is aimed at investigating long API responses and log payloads, where the useful value is usually buried inside objects, arrays, or several JSON.stringify layers. It opens valid clipboard JSON directly into that workspace, lets the user enter a matching result or return without losing selection, and enables extracting just the relevant subtree rather than reformatting and manually scrolling the full document.

Its nested deserialization is also document-wide: it recursively decodes nested JSON-like strings at every depth, while preserving ordinary strings such as `"123"` and `"true"`. The user can restore the original data types at any time, and the exact original input remains available for editing or copying. This is intentionally a reversible inspection aid rather than a different formatter setting.

There are corresponding workflow safeguards for large or imperfect payloads: children load in pages, search explicitly reports its 3,000-node coverage limit, previews can be shortened without truncating copied data, malformed clipboard JSON stays in the editor with a line/column error, and unfinished input is retained as a local draft.

So this is not just a request for a different viewer UI. Formatting is an export action inside a stateful JSON-inspection workspace. Adding this entire interaction model to Format JSON would turn its focused, one-shot formatting/selection/JSONLines workflow into a separate product with its own navigation, search, extraction, recovery, and nested-data semantics.

I believe the two extensions can remain clear and complementary:

- **Format JSON**: transform JSON or a selected JS/JSON value and send the result back quickly.
- **Better JSON**: inspect complex JSON, find a value by its content or path, understand embedded payloads, and extract exactly the needed piece.

For that workflow reason, I would appreciate reconsideration of this as a separate extension. I am happy to make the inspection-and-extraction focus even clearer in the Store copy if useful.

0xdhrv:
Thanks for clarifying your plans.

The distinction versus [Format JSON](https://www.raycast.com/destiner/json-format) is fair. That extension is mainly a one-shot format-and-return flow.

The closer Store overlap is [JSON Editor](https://www.raycast.com/HerbertLu/json-editor) by [@HerbertLu](https://github.com/HerbertLu). It already keeps a JSON document in a workspace, lets you locate a subtree with a filter (`.key.subkey`, `[0][1]`, `.map(...)`), and copies formatted, compact, or escaped output.

This PR adds a hierarchical List, search across keys, values, and paths, and nested JSON-string decoding. Those are useful, but they still serve the same inspect → locate → extract job.

Could you contribute these improvements to JSON Editor instead? That follows our [extension guidelines](https://manual.raycast.com/extensions-guidelines#reasons-to-be-rejected) and would bring them to existing users.

XiaoDaiGua-Ray:
Thanks. I understand why JSON Editor is the closest comparison, and I considered the integration path. However, I would like to keep Better JSON as a standalone extension because the product boundary is intentional.

JSON Editor is an expression-driven editor and converter: users paste JSON, provide a known JavaScript filter expression, edit the resulting value, and export it in several formats. That is useful when the user already knows the structure or wants to write a transformation.

Better JSON is designed for the opposite starting point: an unfamiliar API response or log payload. Its primary interface is a visual, zero-query JSON inspector:

- open valid clipboard JSON directly into a hierarchical browser;
- discover unknown data by searching keys, primitive values, and full paths;
- enter objects and arrays while keeping the current selection and context;
- inspect decoded nested payloads and extract exactly the selected node;
- restore original data types and input when needed.

These are not additional output actions for an editor. They establish a different interaction model: structural discovery and inspection rather than expression-based filtering and conversion. Adding that model to JSON Editor would require a separate navigation/browser experience alongside its current input → filter expression → result workflow.

Better JSON works independently and gives users a clearly different choice in the Store: a visual inspector for exploring complex or repeatedly stringified JSON without first knowing its path or writing an expression.

To make that distinction unmistakable, I am updating the Store copy to foreground “JSON Inspector” and “Browse, Search, and Extract Nested JSON.” I would appreciate reconsideration of Better JSON as a standalone extension on that basis.

0xdhrv (approved):
Looks good to me, approved ✅

XiaoDaiGua-Ray:
@raycastbot  hi, this is my raycast account "@xiaodaiguaray", thx.

PR #30889, opened 2026-09-08
Add path-tools extension

donghan364-hue:
## Checklist


pernielsentikaer (changes requested):
Hi @donghan364-hue 👋

Thanks for the Path Tools extension — the Finder/iTerm/VS Code round trip and the `openi` search view are a nicely scoped set of commands 💪

I found this while reviewing:

`filterExistingPaths` in `src/macos.ts` runs `Promise.all(paths.map(...))` over every candidate `mdfind` returns before the 100-result cap is applied. The search only starts at two characters, so a broad first query can return thousands of paths and each search fires a `stat()` for the whole set. On a large home folder that can leave the results list hanging for a while. Could you validate candidates with bounded concurrency and stop once you have enough valid results? For example, walk the candidates in small chunks (or a small worker pool) and return early after `maxSearchResults` existing paths are collected.

One smaller thing while you're in `src/openi.tsx`: the `error` branch returns a `<List>` without `searchText`/`onSearchTextChange`, so if a search fails the search bar stops responding and the command has to be reopened. Wiring the same handlers into that branch (or showing the error with `List.EmptyView` inside the main list) would keep it recoverable.

Optional cleanup: `getFinderTarget`/`finderTargetScript` in `src/macos.ts` are unused, and the `~` handling in `openInVisualStudioCode` is unreachable since the path always comes from Finder.

Thanks 🙂

---

PR #30892, opened 2026-09-08
windows-terminal: add New Tab Menu order preference

artistro08:
## Description

Adds a "Use New Tab Menu order" checkbox preference. When enabled, profiles are filtered and ordered to match Windows Terminal'"'"'s `newTabMenu` setting (`profile`, `matchProfiles`, `folder`, and `remainingProfiles` entries) instead of the raw `settings.json` list order. Off by default, so existing behavior is unchanged.

## Checklist

https://claude.ai/code/session_017qumXMNaDoczpLwhm54iPc

### DIFF
--- a/extensions/windows-terminal/CHANGELOG.md
+++ b/extensions/windows-terminal/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Windows Terminal Changelog
 
+## [New Tab Menu Preference] - 2026-09-08
+
+- Added the `Use New Tab Menu order` preference. When enabled, profiles are filtered and ordered to match Windows Terminal's `newTabMenu` setting (`profile`, `matchProfiles`, `folder`, and `remainingProfiles` entries), instead of the raw `settings.json` list order.
+
 ## [Always Open Administrator in a Normal Window] - 2026-07-22
 
 - "Open as Administrator" now always launches into a regular window, regardless of the `Open profiles in quake window` preference. Elevated quake windows don't respond to Windows Terminal's global quake shortcut (Win+`), so routing admin through quake left users unable to summon the drop-down back once it lost focus. The separate "Open as Administrator (Non-Quake)" action added in the previous release is no longer needed and has been removed.
--- a/extensions/windows-terminal/CHANGELOG.md
+++ b/extensions/windows-terminal/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Windows Terminal Changelog
 
-## [New Tab Menu Preference] - 2026-09-08
+## [New Tab Menu Preference] - {PR_MERGE_DATE}
 
 - Added the `Use New Tab Menu order` preference. When enabled, profiles are filtered and ordered to match Windows Terminal's `newTabMenu` setting (`profile`, `matchProfiles`, `folder`, and `remainingProfiles` entries), instead of the raw `settings.json` list order.
 
--- a/extensions/windows-terminal/package.json
+++ b/extensions/windows-terminal/package.json
@@ -23,6 +23,15 @@
       "type": "checkbox",
       "required": false,
       "default": false
+    },
+    {
+      "name": "useNewTabMenu",
+      "title": "New Tab Menu",
+      "label": "Use New Tab Menu order",
+      "description": "When enabled, shows profiles in the order defined by Windows Terminal's New Tab Menu setting (newTabMenu in settings.json) instead of the raw profile list order.",
+      "type": "checkbox",
+      "required": false,
+      "default": false
     }
   ],
   "dependencies": {
--- a/extensions/windows-terminal/package.json
+++ b/extensions/windows-terminal/package.json
@@ -51,6 +51,7 @@
     "lint": "ray lint",
     "fix-lint": "ray lint --fix",
     "build": "ray build",
+    "test": "node --test src/new-tab-menu.test.ts",
     "publish": "npx @raycast/api@latest publish"
   },
   "author": "lunaris",
--- a/extensions/windows-terminal/src/new-tab-menu.test.ts
+++ b/extensions/windows-terminal/src/new-tab-menu.test.ts
@@ -37,13 +37,6 @@ describe("buildProfileMatcher", () => {
     assert.equal(buildProfileMatcher({ type: "matchProfiles", name: "[" }), null);
   });
 
-  it("rejects patterns that can backtrack exponentially", () => {
-    assert.equal(buildProfileMatcher({ type: "matchProfiles", name: "(a+)+" }), null);
-    assert.equal(buildProfileMatcher({ type: "matchProfiles", name: "(a|aa)*" }), null);
-    assert.equal(buildProfileMatcher({ type: "matchProfiles", commandline: "((ab)+)+" }), null);
-    assert.equal(buildProfileMatcher({ type: "matchProfiles", source: "(a{1,2})+" }), null);
-  });
-
   it("keeps patterns that only look nested", () => {
     assert.deepEqual(matchedNames({ type: "matchProfiles", name: "(Power)+Shell" }), ["PowerShell"]);
     assert.deepEqual(matchedNames({ type: "matchProfiles", name: "(?:Power|Command)\\s.*" }), ["Command Prompt"]);
@@ -51,6 +44,28 @@ describe("buildProfileMatcher", () => {
       "PowerShell",
       "Command Prompt",
     ]);
+  });
+
+  it("keeps a repeated group whose alternatives are disjoint", () => {
+    assert.deepEqual(matchedNames({ type: "matchProfiles", commandline: "(pwsh|cmd)+\\.exe" }), [
+      "PowerShell",
+      "Command Prompt",
+    ]);
+    assert.deepEqual(matchedNames({ type: "matchProfiles", name: "(a|b)+" }), []);
+  });
+
+  it("bounds pathological patterns instead of rejecting or hanging on them", () => {
+    // (a+)+, (a?a?)+, ((ab)+)+, (a|aa)* all cause exponential backtracking in a native regex
+    // engine. The matcher still builds (the pattern is valid syntax) and returns fast with no
+    // match, instead of freezing or silently dropping a pattern that could otherwise be safe.
+    const adversarialName = "a".repeat(40) + "!";
+    for (const pattern of ["(a+)+", "(a?a?)+", "((ab)+)+", "(a|aa)*"]) {
+      const start = Date.now();
+      const matcher = buildProfileMatcher({ type: "matchProfiles", name: pattern });
+      assert.notEqual(matcher, null, pattern);
+      assert.equal(matcher!({ ...powershell, name: adversarialName }), false, pattern);
+      assert.ok(Date.now() - start < 1000, `${pattern} took too long`);
+    }
   });
 });
 
--- a/extensions/windows-terminal/src/new-tab-menu.test.ts
+++ b/extensions/windows-terminal/src/new-tab-menu.test.ts
@@ -105,6 +105,35 @@ describe("buildProfileMatcher", () => {
     const matcher = buildProfileMatcher({ type: "matchProfiles", name: "a?".repeat(5000) + "b" });
     assert.equal(matcher!({ ...powershell, name: "b" }), true);
   });
+
+  it("treats \\b and \\B as word boundaries, not literal letters", () => {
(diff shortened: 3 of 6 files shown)


pernielsentikaer (changes requested):
Hi @artistro08 👋

Thanks for adding this preference and addressing the previous feedback 💪

I found two remaining differences from Windows Terminal in `resolveNewTabMenuOrder` (`extensions/windows-terminal/src/open-profile.tsx`):

- `remainingProfiles` expands too early. For example:
  ```json
  [
    { "type": "remainingProfiles" },
    { "type": "profile", "profile": "PowerShell" }
  ]
  ```
  PowerShell should appear after the remaining profiles, but currently stays wherever it appeared in the raw list. Please collect explicit/matching references across the whole tree before expanding the remainder, as [Windows Terminal does](https://github.com/microsoft/terminal/blob/093e49e29a9f806ff83025c49be5d0c970673b00/src/cascadia/TerminalSettingsModel/CascadiaSettingsSerialization.cpp#L1753-L1898).

- `matchProfiles` should match when **any** configured field matches, using whole-string regex matching. Currently, multiple fields must all match, partial names match unexpectedly, and an entry without patterns selects everything. See [Windows Terminal's matching implementation](https://github.com/microsoft/terminal/blob/093e49e29a9f806ff83025c49be5d0c970673b00/src/cascadia/TerminalSettingsModel/MatchProfilesEntry.cpp#L78-L103).

Could you align these rules with Windows Terminal and add regression tests for these cases? 🙂

---

artistro08:
Thanks for the review! I'll make those changes now.

pernielsentikaer (changes requested):
Hi @artistro08 👋

Thanks for fixing the ordering and matching rules 💪

`buildProfileMatcher` still matches empty profile fields: `"source": ".*"` includes local profiles without a source, and `"commandline": ".*"` includes profiles without a command line. [Windows Terminal explicitly excludes empty field values](https://github.com/microsoft/terminal/blob/093e49e29a9f806ff83025c49be5d0c970673b00/src/cascadia/TerminalSettingsModel/MatchProfilesEntry.cpp#L78-L103).

Could you exclude empty values and add regression tests? Replace the matcher’s return in `extensions/windows-terminal/src/open-profile.tsx` with:

```typescript
return (profile: Profile) =>
  matchers.some(({ regex, get }) => {
    const value = get(profile);
    return value.length > 0 && regex.test(value);
  });
```

---

artistro08:
Added test and hopefully fixed everything. let me know if I need to add anything else

pernielsentikaer (changes requested):
Hi @artistro08 👋

Thanks for the fixes and regression tests 💪

Two things remain:

- CI fails on the `.ts` test import. Could you add `"allowImportingTsExtensions": true` to `compilerOptions` in `tsconfig.json` for the no-emit type check?
- Could you address the [regex feedback](https://github.com/raycast/extensions/pull/30892#discussion_r3963658880) without silently rejecting valid patterns? `(dev|prod)+` currently hides matching profiles, while `(a?a?)+` still bypasses the guard and stalls matching. Please use bounded, interruptible matching and add regression coverage for both cases.

---

artistro08:
Should be all good

pernielsentikaer (changes requested):
Hi @artistro08 👋

Thanks for addressing the earlier feedback 💪

The custom matcher in `extensions/windows-terminal/src/new-tab-menu.ts` still crashes on `commandline: ".*"` with a 2,000-character command line (`RangeError: Maximum call stack size exceeded`). It also rejects valid matches: `(a?)*b` against `b`, and `[\D]+` against `PowerShell`.

Could you make matching stack-safe and preserve these regex semantics? Please add regression coverage, including:

```typescript
assert.equal(
  buildProfileMatcher({ type: "matchProfiles", name: "[\\D]+" })!(powershell),
  true,
);
```

Please also prevent zero-length repetitions from consuming the entire budget instead of matching the remaining pattern.

---

artistro08:
Thank you for that. Regex isn't my strongsuit 😅

pernielsentikaer (changes requested):
Hi @artistro08 👋

Thanks for fixing the earlier issues 💪

Two things remain in `src/new-tab-menu.ts`:

- Valid patterns such as `\bPowerShell\b` and `(?i)powershell` still omit PowerShell, although they match with [Windows Terminal’s ICU matcher](https://github.com/microsoft/terminal/blob/d605121e14e281a7d06a8d93bf9d3141ead7024a/src/cascadia/TerminalSettingsModel/MatchProfilesEntry.cpp). Could you preserve these semantics rather than silently rejecting or interpreting them as literals? For example:
  ```typescript
  assert.equal(
    buildProfileMatcher({ type: "matchProfiles", name: "\\bPowerShell\\b" })!(powershell),
    true,
  );
  ```
- `MAX_QUANT_BOUND` limits individual quantifiers, but `(a{1000}){1000}` still expands into a huge program. Could you enforce a total instruction budget during compilation, including nested repetitions, and handle exhaustion safely?

---

artistro08:
I think I got it now

pernielsentikaer (changes requested):
Hi @artistro08 👋

Thanks for addressing the earlier issues 💪

The [remaining Greptile feedback](https://github.com/raycast/extensions/pull/30892#discussion_r3978333316) still applies. This fixture takes about 4.9 seconds synchronously in `resolveNewTabMenuOrder`:

```typescript
resolveNewTabMenuOrder(
  [{ guid: "test", name: "Test", commandline: "a".repeat(2000) }],
  [{ type: "matchProfiles", commandline: "a{0,12000}" }],
);
```

Could you bound total matching work or move matching into an interruptible worker, handling exhaustion explicitly rather than silently hiding profiles? The compilation cap alone doesn’t prevent the command from becoming unresponsive.

---

artistro08:
Sorry about that, totally missed it. I'll have that fixed shortly

artistro08:
Alright lets try this again, I wanted to clean up everything so I wasn't going in a loop

pernielsentikaer (changes requested):
Hi @artistro08 👋

Thanks for the fixes 💪

Two valid patterns still hide matching profiles in `src/new-tab-menu.ts`:

- `(?:x|xa)a{0,25}` fails against `"x" + "a".repeat(26)`. `capForDedup` merges counts from different entry positions; a fixed-width body doesn’t prevent that.
- The [nullable-repeat feedback](https://github.com/raycast/extensions/pull/30892#discussion_r3979750123) still applies: `(|a){2,4000}` now returns `null` rather than matching `"a"`.

Could you preserve these matches without silently rejecting valid patterns? This regression assertion currently fails:

```typescript
assert.equal(
  buildProfileMatcher({
    type: "matchProfiles",
    name: "(?:x|xa)a{0,25}",
  })!({ guid: "test", name: "x" + "a".repeat(26) }),
  true,
);
```

---

artistro08:
lets try now

pernielsentikaer (changes requested):
Hi @artistro08 👋

Thanks for fixing the previous counted-repeat cases 💪

The [budget feedback](https://github.com/raycast/extensions/pull/30892#discussion_r3979268583) still applies with a larger minimum. This assertion fails because matching exhausts its budget:

```typescript
assert.equal(
  buildProfileMatcher({
    type: "matchProfiles",
    commandline: ".*a{1000,4000}",
  })!({ guid: "test", name: "Test", commandline: "a".repeat(2000) }),
  true,
);
```

Nested repeats also remain unsupported: `(a{21}){21}` returns `null` instead of matching 441 `a` characters. Both patterns match with ICU, but their profiles disappear here.

Could you preserve these matches and distinguish resource exhaustion/unsupported patterns from genuine non-matches? Please handle those failures visibly rather than silently filtering profiles out; simply increasing the budget won’t resolve the underlying issue.

---

artistro08:
Hi @pernielsentikaer 👋

Pushed c240315279 addressing the three points:

- **`.*a{1000,4000}` against 2,000 characters now matches** (about 10ms). Root cause was that every count below `min` was its own thread, and `.*` enters the repeat at every position, so cost was O(length × min). Threads inside a counted repeat now carry a *range* of counts, and adjacent ranges at the same instruction merge, so that case is linear. `.*a{3000,4000}` against 3,900 characters passes too. Both are pinned by result in the tests, not by timing.
- **`(a{21}){21}` now matches 441 `a` characters** (and rejects 440/442). A large quantifier inside another large one is supported by unrolling the inner one within the counted repeat's body; `((a|b){200}){200}` matches 40,000 characters in ~90ms. The existing total program size cap still bounds the unrolled body.
- **Resource exhaustion and unsupported constructs are surfaced instead of silently dropped.** `buildProfileMatcher` now throws an exported `UnsupportedPatternError` (naming the pattern) for constructs the engine doesn't implement (`\p{L}`, lookaround, backreferences, ICU nested sets like `[[:alpha:]]` and `&&` intersection), patterns too large to compile, and matches that run past the step budget. Malformed patterns that Windows Terminal itself rejects still match nothing, as before. `resolveNewTabMenuOrder` lets the error propagate, and the Open Profile command catches it, shows a failure toast with the message, and falls back to the default profile sections rather than showing an order with profiles missing.

One known ceiling, documented in code and pinned by a test so it's visible rather than silent: a counted repeat whose alternatives differ in width by 2 or more (e.g. `(a|aaa){1000,4000}`) reaches counts with gaps between them that can't merge, so a long value still exhausts the budget and reports the error. With a `.*` prefix filling the gaps it matches normally.

Verified with a differential run of the new matcher against the previous exact-count engine (22,120 pattern/input checks, 0 mismatches), plus 37 tests, eslint, and prettier all clean.

artistro08:
I hope I got it this time

pernielsentikaer (changes requested):
Hi @artistro08 👋

Thanks for fixing the counted repeats and adding the visible fallback 💪

Some valid patterns still silently drop profiles in `src/new-tab-menu.ts`: `\w+` with `Développement`, `\bÉquipe\b` with `Équipe`, `\d+` with `١٢٣`, and `.` with `🚀` all match with ICU but fail here. Also, `a.b` incorrectly matches a newline, and the valid possessive pattern `a++` returns `null` for `aaa` instead of reaching the unsupported-pattern fallback.

Could you preserve ICU’s character semantics and route unsupported possessive quantifiers through `UnsupportedPatternError` rather than treating them as malformed? For example, this assertion currently fails:

```typescript
assert.equal(
  buildProfileMatcher({ type: "matchProfiles", name: "\\w+" })!(
    { guid: "test", name: "Développement" },
  ),
  true,
);
```

For the unsupported possessive suffix, `parseQuant` can explicitly reject it through the visible fallback:

```typescript
if (quantified && peek() === "+") {
  return unsupported("possessive quantifier not implemented");
}
```

---

artistro08:
Lets try this again

pernielsentikaer (changes requested):
Hi @artistro08 👋

Thanks for fixing the previous Unicode examples and possessive fallback 💪

Two character-semantics differences remain in `src/new-tab-menu.ts`. `(?i)straße` with `STRASSE` and `(?i)ss` with `ß` match in ICU but return `false` here, silently hiding those profiles. The per-character uppercase/lowercase checks don’t implement ICU case folding. Could you preserve those matches, or explicitly route unsupported case-folding cases through the visible fallback?

`\s` also misses vertical tab (U+000B) and next line (U+0085). You can cover ICU’s whitespace set with:

```typescript
const isSpace = (ch: string) => /\p{White_Space}/u.test(ch);
```

For the case-folding issue, this currently fails:

```typescript
assert.equal(
  buildProfileMatcher({ type: "matchProfiles", name: "(?i)straße" })!(
    { guid: "test", name: "STRASSE" },
  ),
  true,
);
```

Regression coverage for these cases would be helpful too.

---

artistro08:
Thanks for that! Should be good for review

pernielsentikaer (changes requested):
Hi @artistro08 👋

The literal fix helps, but case-insensitive character classes still differ from ICU: `(?i)[A-Z]+` matches `ı`, while `(?i)[ſ]` rejects `S`. Could you correct the class case-folding so profiles aren’t incorrectly included or hidden?

---

artistro08:
thanks again!

theycallhermax:
hello

pernielsentikaer:
Hi @theycallhermax 👋 

Do you have time to review this update?

theycallhermax:
I currently don't have access to a windows machine, but I will review as soon as I am able to

pernielsentikaer (approved):
I tested it on Windows. Looks good to me, approved 🔥

artistro08:
Thanks @pernielsentikaer for helping me and guiding me on what needed done. much appreciated.

PR #30899, opened 2026-09-08
Update Sunsama extension

artistro08:
## Description

Update to the Sunsama extension.

A pasted link can carry a channel automation on Sunsama's side — a Trello board mapped to a channel, for example. Add Task was always passing the default channel to `create_task`, which suppressed that automation, so a linked task never landed where the automation intended.

The channel is now left off when a link is being attached and the user hasn't picked one, and the default is applied afterwards only if the server picked nothing. An explicit pick is passed straight through and still wins.

Add Task also opens its connection to the MCP server when the form mounts. Nothing else in that form touches the server — the channel list is read from disk — so the handshake used to land on submit, where the user waits on it.

Verified against the live server: a plain task comes back with no `channel` field, while a Trello card on a mapped board comes back with the channel already filled in.

## Checklist

https://claude.ai/code/session_01KAYqb4ZRrrq7ukUszDQQXW

### DIFF
--- a/extensions/sunsama/CHANGELOG.md
+++ b/extensions/sunsama/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Sunsama Changelog
 
-## [Default Channel for Linked Tasks] - {PR_MERGE_DATE}
+## [Default Channel for Linked Tasks] - 2026-09-09
 
 - A pasted link now keeps whichever channel Sunsama's own automation assigns to
   it, and falls back to your default channel when no automation fires. Picking a
--- a/extensions/sunsama/CHANGELOG.md
+++ b/extensions/sunsama/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Sunsama Changelog
 
+## [Default Channel for Linked Tasks] - {PR_MERGE_DATE}
+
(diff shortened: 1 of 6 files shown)


0xdhrv (commented):
Hi @artistro08 👋

Thanks for the contribution 🔥

---

pernielsentikaer (changes requested):
Hi @artistro08 👋

Thanks for the fixes 💪

For a linked task, selecting another channel and then switching back to the initial channel still counts as “untouched” in `extensions/sunsama/src/add-task.tsx`, so Sunsama’s automation can override an explicit choice.

Could you track channel changes separately instead of comparing the final value? For example:

```tsx
const [channelChanged, setChannelChanged] = useState(false);
const deferToAutomation = !!url && !channelChanged;
```

```tsx
<ChannelDropdown
  {...itemProps.channel}
  channels={channels}
  onChange={(value) => {
    setChannelChanged(true);
    itemProps.channel.onChange?.(value);
  }}
/>
```

Please also prevent late default-channel initialization from overwriting a user selection, and add a regression test for switching away and back.

---

artistro08:
Thank you. should be ready again now

pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #30901, opened 2026-09-08
Agent Usage: Antigravity quota when the app is closed

JustMarkDev:
## Summary

Antigravity usage died the second you quit the app or `agy`. The probe only talked to a live localhost language server, so Raycast just said not running.

This keeps that local probe first. When it comes back `not_running` (or `missing_csrf` with no usable server), we read the Google OAuth token Antigravity already left in Keychain / `~/.gemini/antigravity-cli`, refresh it with Antigravity's public installed-app client, and call Cloud Code `retrieveUserQuotaSummary`. Same quota groups the UI already knows how to render. Email and plan come from Google userinfo and `loadCodeAssist`. We never write credentials back into Antigravity's store.

## How we tested

- `cd extensions/agent-usage && npm install && npm run typecheck && npm test && npm run lint` (276 tests passed, 1 skipped)
- Unit coverage for the Cloud Code quota fixture, fetcher OAuth fallback on `not_running`, and keeping the probe error when no credentials exist
- Live OAuth call on this machine with Antigravity closed: refreshed the Keychain token, hit `retrieveUserQuotaSummary`, got both Gemini and Claude/GPT weekly + 5h groups back
- Confirmed userinfo email and `loadCodeAssist` plan (`paidTier.name`) populate the detail view on the closed-app path


JustMarkDev:
Addressed the three Greptile notes:

- Access-token cache is keyed to the current credential and cleared when credentials disappear, so account switches/logouts do not keep serving the previous token.
- Stale access tokens no longer fall through when refresh is missing or fails; we return null and keep the probe error.
- Changelog heading now uses `{PR_MERGE_DATE}`.

JustMarkDev:
a screenshot to show the changes working

Screenshot 2026-09-08 at 20 45 35

pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #30902, opened 2026-09-08
Agent Usage: apply saved provider order in the menu bar

JustMarkDev:
## Summary
- The list command already saved Move Up/Down order to LocalStorage (`agent-order`), but the menu bar always sorted with the default provider order.
- Menu bar now loads that shared order on mount and again on Refresh All, and sorts items (including multi-account row ids) with it.
- Shared parse/sort helpers live in `order.ts` so list and menu bar stay in sync.

## Testing
- Ran `npm test` (including order helpers), `npm run build`, typecheck, and lint — all passed.
- Simulated the menu bar path locally: stored order JSON → `parseStoredAgentOrder` → `sortByAgentOrder` (with multi-account row ids); confirmed custom order differed from default and reset restored default.
- Started `npx ray develop`, reordered providers in Agent Usage with Move Up/Down, then used Refresh All in the menu bar — order matched the list.

### DIFF
--- a/extensions/agent-usage/CHANGELOG.md
+++ b/extensions/agent-usage/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Agent Usage Changelog
 
+## [Unreleased]
+
+### Bug Fixes
+
+- Apply the provider order from Move Up/Down in the list to the menu bar (use Refresh All after reordering)
+
 ## [Add OpenRouter credit balance] - 2026-09-08
 
 ### New Features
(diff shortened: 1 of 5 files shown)


EDM115:
yeah when I added that I couldn't test on the menu bar as i don't have a mac  
ggs if you made it work, maybe a screenshot for proof ? :)

JustMarkDev:
i'm sure you know that the current default order is different from this one, so it works, the only thing is that the user needs to refresh it or if there is an automatic refresh it refreshes the order as well. i also opened another PR #30901 because i don't always have agy open but i still want to know how much usage it has

Screenshot 2026-09-08 at 20 38 15

JustMarkDev:
there is a screenshot on the other pr as well, they are both safe to merge, fixed any greptile issues that popped up

pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #30904, opened 2026-09-08
Update slack extension

0xdhrv:
## Checklist

### DIFF
--- a/extensions/slack/.gitignore
+++ b/extensions/slack/.gitignore
@@ -12,9 +12,11 @@ compiled_raycast_rust
 
 # misc
 .DS_Store
-
 .raycast-swift-build
 .swiftpm
 compiled_raycast_swift
 /node_modules
 raycast-env.d.ts
+
+# Test output
+.test-dist/
--- a/extensions/slack/CHANGELOG.md
+++ b/extensions/slack/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Slack Changelog
 
-## [Fix recipient search, AI file uploads, and attachment downloads] - {PR_MERGE_DATE}
+## [Fix recipient search, AI file uploads, and attachment downloads] - 2026-09-10
 
 - Keep **Send Message** and **Open Channel** responsive in large workspaces by loading bounded, filtered directory pages instead of retaining and caching the entire workspace directory.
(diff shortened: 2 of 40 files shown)


pernielsentikaer (changes requested):
Hi @0xdhrv 👋

Thanks for improving large-workspace search 💪

Open Channel now loses frequently opened channels outside the initial result cap. I reproduced this with a top-ranked channel on page two: searching finds it, but reopening the command hides it again because `useFrecencySorting` only sorts the supplied rows.

Could you persist a bounded set of visited rows and merge them into the empty-query results before sorting? For example, retain the latest 100 visited items:

```ts
setRecentItems((items) =>
  [item, ...items.filter((entry) => entry.id !== item.id)].slice(0, 100),
);
```

Please add a regression test covering a visited channel outside the initial page.

---

pernielsentikaer (changes requested):
Hi @0xdhrv 👋

Thanks for addressing the recents issue 💪

I reproduced two issues in `src/shared/client/downloadFile.ts`:

- Simultaneous downloads with the same filename can overwrite each other while both report success. Could you reserve filenames atomically with exclusive creation (`await open(candidate, "wx")`), retry another filename on `EEXIST`, and only clean up files owned by that download?
- Valid HTML attachments are rejected as sign-in pages. Please distinguish authentication responses from actual HTML files rather than rejecting `text/html` outright.

Please also add `- files:read` to the README’s personal-token setup; it currently documents only `files:write`. The new download shortcuts need explicit macOS/Windows variants per the [Keyboard API](https://developers.raycast.com/api-reference/keyboard).

---

pernielsentikaer (changes requested):
Hi @0xdhrv 👋

Thanks for addressing the previous feedback 💪

The new HTML preview can freeze downloads when the response arrives in chunks. `readPrefix()` in `src/shared/client/downloadFile.ts` requests more bytes than are buffered, then immediately retries without letting network events run.

Could you consume the available bytes instead? This change passed my local HTTP reproduction:

```ts
const chunk = stream.read(
  Math.min(maxBytes - total, stream.readableLength || maxBytes - total),
) as Buffer | string | null;
```

Regression coverage for delayed, chunked HTML responses would be helpful too.

---

pernielsentikaer (approved):
Looks good to me, approved 🔥

Issue #30903, opened 2026-09-08
[Browser Tabs]


sotarko:
### Extension

https://www.raycast.com/koinzhang/browser-tabs

### Raycast Version

2.2.0.0

### OS Version

Windows Windows 11 Professional 10.0.26200.0 amd64

### Description

When Raycast brings a window belonging to another process to the foreground (tested with the Browser Tabs extension switching to a tab, and with window switching to other maximized apps), the window comes up in a small, non-maximized state instead of preserving its previous maximized size. Happens consistently, not just occasionally.
Raycast Version: 2.2.0.0
OS: Windows 11 Professional 10.0.26200.0 amd64s

### Steps To Reproduce

Have any app window maximized
Use Raycast to switch focus to a different already-open window (tested with Browser Tabs extension and with window switching)
The target window comes up small/non-maximized instead of keeping its previous size

artistro08:
Thanks for reporting! I added windows support and never checked this. I'll investigate and try and come up with a fix.

artistro08:
It should be fixed now. Tested on my machine

sotarko:
Tested on my side as well and it works perfectly now, thank you so much for the quick fix!

PR #30908, opened 2026-09-08
fix(browser-tabs): keep maximized windows maximized on Windows

artistro08:
## Description

On Windows, switching to a tab shrank a maximized browser window back to its restored size.

`bring_window_to_front` called `ShowWindow(hwnd, SW_SHOW)` on every window that was not minimized. A window that is already on screen does not need showing again, and that call drops a maximized window out of its maximized state. `SetForegroundWindow` on its own is all the focusing that is needed.

`ShowWindow` now runs only when the window is minimized, where `SW_RESTORE` already puts a window that was maximized before it was minimized back to maximized. Both `activate_tab` and `close_tab` go through this function, so both are fixed.

Ref https://github.com/raycast/extensions/issues/30903 — that report also covers Raycast's own window switching, which is outside this extension, so the issue is left open.

## Checklist

- [x] `cargo check` passes on the native helper
- [x] The change is confined to the Windows helper and does not touch the macOS code path
- [x] I updated the `CHANGELOG.md`

### DIFF
--- a/extensions/browser-tabs/CHANGELOG.md
+++ b/extensions/browser-tabs/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Browser Tabs Changelog
 
-## [Fix Maximized Windows on Windows] - {PR_MERGE_DATE}
+## [Fix Maximized Windows on Windows] - 2026-09-09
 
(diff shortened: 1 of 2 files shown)


0xdhrv (approved):
Looks good to me, approved ✅

PR #30911, opened 2026-09-09
Update whisper extension: zero-knowledge encryption + Multiple Values mode

quentinved:
## Description

Zero-knowledge encryption and a new Multiple Values mode for the Whisper extension.

- Secrets are now encrypted on the user's device (AES-256-GCM) before upload. The decryption key travels only in the link's `#` fragment and is never sent to any server.
- No plaintext ever leaves the device: a server without the zero-knowledge endpoint (Whisper older than 1.3, 404/405 on `/v1/ephemeral`) is rejected with a clear error asking to upgrade, same as the Whisper CLI. There is no fallback to server-side encryption.
- Keyboard shortcuts for Multiple Values are defined for both macOS and Windows; the server URL preference tolerates a trailing slash.
- New **Multiple Values** mode in Create Secret: structured key/value entries with optional sections, shared as one secret (e.g. a set of database credentials).
- README and CHANGELOG updated; added a screenshot of the new mode.

## Screencast

See `metadata/whisper-4.png` (Multiple Values mode).

## Checklist


0xdhrv (commented):
Hi @quentinved 👋

Thanks for the contribution 🔥

---

quentinved:
Hi @0xdhrv, thanks for the review! Both points are addressed in 144e46e:

- Removed the legacy fallback entirely: the extension now never sends plaintext. A server without the zero-knowledge endpoint gets a clear error asking to upgrade (Whisper 1.3+), same behaviour as the Whisper CLI. README and CHANGELOG updated accordingly.
- Multiple Values duplicate checks now use `Object.hasOwn` on null-prototype objects, so keys like `constructor` or `toString` are accepted (this also matches the web UI).

Marking as ready for review.

pernielsentikaer (changes requested):
Hi @quentinved 👋

Thanks for this one — moving encryption onto the device and adding Multiple Values both look good 💪 I verified the flow end to end against the hosted instance (a secret created with your code is accepted by `/v1/ephemeral` and decrypts with the `#k=` key, and the reveal page renders the nested sections), and both Greptile points are addressed in the current head.

Two things before we merge:

**1. Windows shortcuts.** `package.json` declares `platforms: ["macOS", "Windows"]`, but the new actions in `whisper-form.tsx` use `cmd` without platform variants:

```tsx
shortcut={{ modifiers: ["cmd"], key: "n" }}
```

Raycast needs both platforms when an ambiguous modifier like `cmd`/`ctrl` is involved ([docs](https://developers.raycast.com/api-reference/keyboard#keyboard.shortcut)), otherwise these shortcuts aren't available on Windows. Could you use `Keyboard.Shortcut.Common.New` for Add Entry where it fits and `{ macOS: { … }, Windows: { … } }` for the others? A Windows column in the README shortcut list would be great too.

**2. Server URL with a trailing slash.** With a preference like `https://whisper.example.com/`, the request goes to `//v1/ephemeral`, which 404s — and the new branch then reports "This server does not support zero-knowledge secrets (Whisper 1.3 or newer required)…", which sends self-hosted users after the wrong problem. You can reproduce it by pointing the preference at `https://whisper.quentinvedrenne.com/`. The concatenation predates this PR, but the new message makes the misdiagnosis confusing. Normalizing in `getWhisperUrl()` fixes both the request and the generated link:

```ts
return serverUrl?.trim().replace(/\/+$/, "") || DEFAULT_WHISPER_URL;
```

Optionally, while you're in `shared.ts`: a 201 response without an `id` currently produces a link with `shared_secret_id=undefined`, and a non-JSON 201 surfaces a raw `Unexpected token` message.

Also, the PR description still mentions the automatic legacy fallback that `144e46e` removed — could you update it so the security behaviour is described accurately?

---

I'll convert this PR into a draft after submitting this review. Please press Ready for review when it's ready and we'll have a look 😊

quentinved:
Hi @pernielsentikaer, thanks for the thorough review and for testing it end to end 🙏 All addressed in the latest commit:

1. **Windows shortcuts**: Add Entry uses `Keyboard.Shortcut.Common.New`; the other four actions now have explicit `macOS` / `Windows` variants (`ctrl`/`alt` on Windows). README has a shortcut table with a Windows column, and the in-form hint mentions `Ctrl+K`.
2. **Trailing slash**: `getWhisperUrl()` strips trailing slashes, so `https://whisper.quentinvedrenne.com/` now works and the generated link is clean.
3. **Response parsing**: a non-JSON 201 or a 201 without a string `id` now throws an explicit "Unexpected response from Whisper server" error instead of producing `shared_secret_id=undefined` or a raw parse error.
4. PR description updated to reflect that there is no legacy fallback anymore.

Marking as ready for review.

pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #30913, opened 2026-09-09
Add fastener-lookup extension

alejdiaz25:
## Description

Adds **Fastener Lookup** — tap drills, clearance holes, socket-head counterbores, and thread specs for CAD work, in one searchable command.

- Covers #0–1" UNC/UNF and M1.6–M30 coarse/fine pitch
- Natural size parsing: `#10`, `10-32`, `1/4-20`, `.375`, `number 10`, `m6 close`, `m10 x 1.25`, `tap 3/8`, `clearance for #10`, `cbore 1/4`
- Every row is a single value; **Enter copies the bare number** for pasting straight into a CAD dimension field. Also: paste into active app, copy in the other unit, copy the drill designation, copy a Markdown spec sheet, unit-override dropdown
- Data: Machinery's Handbook ~75% tap drills, ASME B18.2.8 and ISO 273 clearance fits, ASME B18.3 / ISO 4762 counterbores; nearest inch drill shown for every metric size
- Platform-specific shortcuts; built and tested on Windows, macOS supported
- Ships with parser and data-integrity tests (`npm test`) — every inch drill designation is cross-checked against the drill-size table

## Screencast

Screenshots to follow shortly.

## Checklist


pernielsentikaer (changes requested):
Hi @alejdiaz25 👋

Thanks for the Fastener Lookup extension — the dataset and the search-first workflow are really nice, and the parser/data tests are a great touch 💪

I found a few things while reviewing:

**Screenshots** — there's no `metadata/` folder yet, and this is a `view` command, so screenshots are required before merging. Could you add at least one native Raycast Window Capture saved with **Save to Metadata**? See the [screenshot guidelines](https://developers.raycast.com/basics/prepare-an-extension-for-store#screenshots).

**Unrecognized queries show the whole chart.** A query with no recognized size or keyword (`hello`, `xyz`, `1in`) falls through to `matchesSize()` returning `true` for everything, so the command renders all 508 rows and the "No fastener matched" empty view never appears. Partial matches are also silently dropped (`#1O` → `#1`). Could `search()` return `[]` when the query contains no recognized size, section, series, fit, or system, and treat leftover unrecognized tokens as a non-match?

**The 1" spec sheet is empty.** For the `1"` fastener `r.size` is `"1"`, and `specSheet("1")` re-searches `"1"`, which parses as `#1`, so the `r.size === "1"` filter leaves nothing — the detail panel is blank and "Copy 1 Spec Sheet" copies an empty string. Building the rows from the fastener directly (e.g. a `rowsForSize(size)` helper in `search.ts`) instead of re-parsing the size string would fix it.

**`1-8` and `1-12` don't resolve.** A bare `<n>-<tpi>` maps to `#<n>`, and `#1` has no 8/12 TPI thread, so the 1" UNC/UNF size is only reachable via `1"` / `1.0` / `1 in` even though the README advertises `#0–1"` coverage. When a TPI is present you could also match a whole-inch size:

```ts
case "number":
  return (
    f.system === "imperial" &&
    (f.size === `#${n}` || (q.tpi !== undefined && f.size === String(n)))
  );
```

**README example** — it says `6.6 mm ≈ H drill`, but 6.6 mm is 0.2598 in and `G` (0.261) is closer than `H` (0.266), which is what the extension shows. Could you change that example to `G`?

---

PR #30920, opened 2026-09-09
[UpNote] Fix special characters breaking note, notebook and tag actions

### LINKED ISSUE #30895: [UpNote] Apostrophe character prevents notes from being created
### Extension

https://www.raycast.com/ujjwal1998/upnote

### Raycast Version

2.2.1.0

### OS Version

macOS Tahoe 26.5.1 arm64

### Description

If I try to create a note that contains the `’` character, the note simply isn't created.

### Steps To Reproduce

1. Open the 'Create Note' Raycast UpNote command
2. Enter `'` in either the Title or Text fields
3. Submit the note

### Current Behaviour

Nothing happens

### Expected Behaviour

A note with the entered details is created

szupzj18:
## Description

Fixes #30895 — notes containing an apostrophe (e.g. `Bob's note`) were silently never created.

### Root cause

All three commands (`create-note`, `create-notebook`, `view-tag`) built the x-callback-url as a shell string and ran it through `child_process.exec`:

```ts
const cmd = `open 'upnote://x-callback-url/note/new?title=${values.title}&text=${values.textarea}&...'`;
exec(cmd);
```

User input was interpolated raw inside single quotes:

- An apostrophe (`'`) terminates the shell quote — bash fails with a syntax error before `open` ever runs, so nothing happens (exactly the reported symptom). Verified locally: running the old command with `Bob's note` exits `syntax error: unexpected EOF while looking for matching '`.
- Characters with URL meaning (`&`, `#`, `=`, spaces, newlines, non-ASCII) were not percent-encoded, so they truncated or garbled the parameters UpNote received.
- Passing user text through a shell string also left a command-injection surface.

### Changes

- Added a small shared `buildUpnoteUrl(action, params)` helper that percent-encodes every parameter value (`encodeURIComponent`) and omits undefined ones.
- All three commands now build the URL with it and open it via Raycast's `open()` (no shell involved — argv is passed directly).
- CHANGELOG entry added.

### Validation

- RED: reproduced the old failure — a stub `open` on PATH never runs because the shell rejects the command containing `Bob's note`.
- GREEN: asserts that titles/text/tags with apostrophes, spaces, `&`, `%`, newlines and unicode (`’`, `—`, `é`) encode correctly and round-trip through `URLSearchParams` exactly for all three commands.
- `tsc --noEmit` clean, `ray lint` clean (only a pre-existing title-casing warning), `ray build` succeeds.

### DIFF
--- a/extensions/upnote/CHANGELOG.md
+++ b/extensions/upnote/CHANGELOG.md
@@ -1,6 +1,6 @@
 # UpNote Changelog
 
-## [Fix Special Characters Breaking Note, Notebook and Tag Actions] - {PR_MERGE_DATE}
+## [Fix Special Characters Breaking Note, Notebook and Tag Actions] - 2026-09-09
 
 - Fixed Create Note, Create Notebook and View Tag failing silently when the entered title, text or tag contained an apostrophe or other special characters; x-callback-url parameters are now properly percent-encoded and the URL is opened without going through a shell.
 
--- a/extensions/upnote/CHANGELOG.md
+++ b/extensions/upnote/CHANGELOG.md
@@ -1,5 +1,9 @@
 # UpNote Changelog
 
+## [Fix Special Characters Breaking Note, Notebook and Tag Actions] - 2026-09-09
+
+- Fixed Create Note, Create Notebook and View Tag failing silently when the entered title, text or tag contained an apostrophe or other special characters; x-callback-url parameters are now properly percent-encoded and the URL is opened without going through a shell.
+
 ## [Security Maintenance] - 2026-05-21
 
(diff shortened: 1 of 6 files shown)


pernielsentikaer (commented):
@Ujjwal1998 do you want to check this?

Ujjwal1998:
@szupzj18 thanks for your work on this! changes look good to me

Ujjwal1998 (approved):

Ujjwal1998 (approved):

pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #30921, opened 2026-09-09
[Statamic Docs] Fix crash on launch in newer Raycast versions

duncanmcclean:
This pull request fixes an issue where the Statamic Docs extension crashes on launch in newer versions of Raycast, with an "Invalid parameters for method setLocalStorageItem" error.

This was happening because the version dropdown uses `storeValue`, so on mount it restores the previously selected version and fires `onChange` before the available versions have been fetched. The lookup returned `undefined`, which ended up being passed to `LocalStorage.setItem`. Older versions of Raycast silently accepted this, but newer versions validate the value and throw.

This PR fixes it by returning early from the change handler when the version can't be found. The selected version is still restored from `LocalStorage` once the available versions have loaded, so behaviour is otherwise unchanged.

### DIFF
--- a/extensions/statamic-docs/CHANGELOG.md
+++ b/extensions/statamic-docs/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Statamic Docs Changelog
 
-## [Fix] - {PR_MERGE_DATE}
+## [Fix] - 2026-09-09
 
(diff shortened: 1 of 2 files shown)


pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #30923, opened 2026-09-09
Font Search: fix font preview in dark mode

mindprint:
## Description

The font preview is invisible in dark mode. `src/fontPreview.ts` sets the SVG `fill` to `Color.PrimaryText`, but Raycast only resolves `raycast-*` color tokens for `Image.ImageLike` props. Inside a base64 data URI in a markdown `Detail` it is invalid CSS, so the fill falls back to black and the preview disappears against a dark background.

This tints the markdown image with `?raycast-tint-color=${Color.PrimaryText}` instead, which is the approach used by `nerd-font-picker`. Verified against the stock light and dark themes, and against custom themes with pink and blue text colors — the preview now matches the metadata labels beside it in every case.

Also escaped the preview text and font name before interpolating them into the SVG. An `&` in the Preview Text preference (for example `Fish & Chips`) produced invalid XML and the preview did not render at all.

## Screencast

Dark mode, after the fix:

Font Search in dark mode with a readable preview

## Checklist

### DIFF
--- a/extensions/font-search/CHANGELOG.md
+++ b/extensions/font-search/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Search Installed Fonts Changelog
 
-## [Fix Font Preview in Dark Mode] - {PR_MERGE_DATE}
+## [Fix Font Preview in Dark Mode] - 2026-09-09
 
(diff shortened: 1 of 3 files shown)


pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #30936, opened 2026-09-09
[Color Picker] Fix grey color swatches for some color formats

### LINKED ISSUE #30929: [Color Picker] "Organize Colors" shows a grey square instead of the color I picked
### Extension

https://www.raycast.com/thomas/color-picker

### Raycast Version

2.2.0.0

### OS Version

Windows Windows 11 Professional 10.0.26220.0 amd64

### Description

I picked a color with the Color Picker and it grabbed the hex correctly (5763ed, an indigo/blue). But when I open "Organize Colors" to see my history, the swatch just shows a plain dark grey square instead of the actual blue. The hex code and timestamp below it are right, it's only the color preview that's wrong.

Image

### Steps To Reproduce

Details:

- Picked color hex: 5763ed

- Swatch in Organize Colors: solid grey, doesn't match the hex

- Hex + timestamp display correctly

### Current Behaviour

_No response_

### Expected Behaviour

_No response_

Just-Me-22:
## Description

Closes #30929.

Pick a colour, open `Organize Colors`, and the swatch is a plain grey square. The hex text and the timestamp under it are correct, only the swatch is wrong.

The swatch comes from `getPreviewColor`, which formatted the colour using the `Color Format` preference. That preference is for copying and pasting, as its own description in the manifest says, so it can produce strings that are not CSS colours at all. On `HEX No Prefix` the swatch is handed `5763ed` with no `#`, Raycast cannot read that as a colour, and you get grey.

Three formats were already known to have this problem:

```ts
const unsupportedPreviewFormats = ["p3", "rgb", "rgb-percentage"];
```

`hex-no-prefix` and `hsva` were missing from the list. I would rather not add two more names to a list that has now been incomplete twice, so the swatch stops asking the preference and always uses oklch, which is what that list already fell back to. Every piece of text and every copy action still uses the preference exactly as before.

This reaches further than the report suggests. `getPreviewColor` feeds every swatch in the extension, and `getIcon` passes a string through untouched, so the same grey appears in the `Organize Colors` list view, `Favorite Colors`, `Generate Colors` and the menu bar icon.

## Screencast

The before is the screenshot in #30929.

On how I tested this: `ray develop` builds fine on my machine but never hands anything to Raycast, so I could not put a patched build on screen. Instead I tested the released build directly, since the preference lets you feed it any of these strings. On Windows 11 with Raycast 2.2.0, `#5763ED` and `oklch(38.16% 0.19 273)` both render correctly and `5763ed` renders grey, which is the whole bug and the reason oklch is a safe target. I also could not produce a `dist` build, because that step needs the Swift package and I am on Windows. `ray build -e dev` and `ray lint` both pass.

## Checklist

### DIFF
--- a/extensions/color-picker/CHANGELOG.md
+++ b/extensions/color-picker/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [Fix] - {PR_MERGE_DATE}
 
 - Fix color swatches rendering as a grey square when the color format preference is set to a format that is not a valid CSS color, such as `HEX No Prefix`
+- Fix the `OKLCH` format dropping the alpha channel and producing a `NaN` hue for black, white and gray
 
 ## [Favorite Colors & AI Tools] - 2026-08-29
 
--- a/extensions/color-picker/CHANGELOG.md
+++ b/extensions/color-picker/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Color Picker Changelog
 
-## [Fix] - {PR_MERGE_DATE}
+## [Fix] - 2026-09-10
 
 - Fix color swatches rendering as a grey square when the color format preference is set to a format that is not a valid CSS color, such as `HEX No Prefix`
 - Fix the `OKLCH` format dropping the alpha channel and producing a `NaN` hue for black, white and gray
--- a/extensions/color-picker/CHANGELOG.md
+++ b/extensions/color-picker/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Color Picker Changelog
 
+## [Fix] - {PR_MERGE_DATE}
+
+- Fix color swatches rendering as a grey square when the color format preference is set to a format that is not a valid CSS color, such as `HEX No Prefix`
+
 ## [Favorite Colors & AI Tools] - 2026-08-29
 
 - Add a `Favorite Colors` command to view and use saved favorites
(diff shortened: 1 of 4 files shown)


pernielsentikaer (changes requested):
Hi @Just-Me-22 👋

Thanks for the fix 💪

Could you address Greptile’s [alpha-loss finding](https://github.com/raycast/extensions/pull/30936#discussion_r3970521880)? I reproduced it, and also found that black, white, and gray produce an invalid `NaN` hue.

Using ColorJS’s serializer in the existing OKLCH case handles both:

```ts
case "oklch": {
  return color.to("oklch").toString();
}
```

Please add regression coverage for translucent and grayscale previews to the existing tests.

---

Just-Me-22:
Thanks for the quick review, and for catching the grayscale case on top of Greptile's finding. Both were real. Here is the fix 🙂

I switched the OKLCH case to ColorJS's serializer as you suggested, which handles both:

| Input | Before | After |
| --- | --- | --- |
| `#5763ed` | `oklch(56.94% 0.20527682343529072 274.1002878702785)` | `oklch(56.943% 0.20528 274.1)` |
| `#808080` | `oklch(59.99% 5.551115123125783e-16 NaN)` | `oklch(59.987% 0 none)` |
| `rgba(87, 99, 237, 0.5)` | `oklch(56.94% 0.20527682343529072 274.1002878702785)` | `oklch(56.943% 0.20528 274.1 / 0.5)` |


Both cases are covered now:

```
✔ color calculations
✔ oklch previews keep alpha
✔ oklch previews of grayscale colors have no NaN hue
```

`ray lint` and `ray build -e dev` pass. I cannot produce a `dist` build here since that step needs the Swift package and I am on Windows.

pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #30947, opened 2026-09-10
Add text-formatter extension

dakotayounger:
## Description

Adds **Magic Formatter** — a single `no-view` command that cleans up copied text and puts it back on the clipboard with **both** a plain-text (markdown) flavor and a rich-text (HTML) flavor, so one run pastes correctly into markdown apps and rich editors alike.

The problem it solves: copy prose out of an email client, a doc, or an AI chat response and paste it elsewhere, and you get one of two bad outcomes. Either paragraph breaks are lost and you paste a single smashed-together block, or you get raw markers — literal `**bold**` in Slack, which uses single-asterisk syntax and does not parse `**`. Raycast's built-in "paste without formatting" only strips formatting; it fixes neither case.

How it works:

- Reads the **HTML flavor** of the clipboard (via `osascript`) rather than only plain text, because rich sources keep bold, italics, links, and real paragraph breaks there. That HTML is converted to markdown with `turndown`.
- If the source was already rich HTML, the converted markdown passes through as-is — structure is already correct, so nothing needs guessing.
- If the source was plain text that lost its structure, an offline formatter normalizes list markers and spacing and restores paragraph breaks. Breaks are placed on linguistic cues — greetings, sign-offs, and transition phrases like "However" or "Two things" — and only when the text has no remaining structure to preserve, so already-formatted text and lists are never touched. This is the default and needs no setup or network access.
- Writes **both** clipboard flavors back — markdown as plain text and rendered HTML as rich text — so Slack and Notion get markdown while Gmail and Google Docs get real formatting.

Optional AI engine: if the user supplies their own Anthropic API key in preferences, the plain-text path can use Claude instead of the offline heuristics. Strictly opt-in — the `Engine` preference defaults to `auto`, which uses the offline path when no key is present, and the extension is fully functional with no key and no network calls. The system prompt forbids changing any wording. On `auto` a failed API call degrades to the offline path; on the explicit `ai` setting the failure surfaces in a HUD.

Third-party services: only the Anthropic API, only when the user has entered their own key. No telemetry, no other network calls.

Source repo: https://github.com/dakotayounger/raycast-extensions

## Screencast

This is a `no-view` command with no UI to capture — it operates on the clipboard and reports via HUD. Both behaviors below work with **no API key configured**.

**1. Paragraph restoration.** Clipboard before:

```
Hi Sarah, thanks for sending the deck over. I had a chance to review it last night and the framing in section two is much stronger now. Two things I'd change before Thursday. The pricing slide still leads with the annual number, which buries the per-seat comparison. And the case study needs a logo. Best, Dakota
```

After running Magic Formatter:

```
Hi Sarah,

thanks for sending the deck over. I had a chance to review it last night and the framing in section two is much stronger now.

Two things I'd change before Thursday. The pricing slide still leads with the annual number, which buries the per-seat comparison. And the case study needs a logo.

Best,
Dakota
```

No wording is altered — the only edits are inserted line breaks.

**2. Rich formatting survives the paste.** Copy text containing **bold** and *italics* from an email or a doc, run the command, then paste into Slack: you get real bold and italics rather than literal `**` and `*`. Paste the same clipboard into Notion or a plain-text field and you get clean markdown.

HUD confirmation on success: `✅ Formatted — ready to paste` (or `✅ Formatted (Claude) — ready to paste` when the optional AI engine ran).

## Checklist


0xdhrv (commented):

The problem statement is clear, and the dual-flavor clipboard idea (markdown plain text + rendered HTML) is a good fit for Slack vs Gmail/Docs. I do not see overlap with **Clean Text** or **Wrap Unwrap** — this is specifically about pasted prose and rich clipboard sources. `npm run build` and `ray lint` are clean on my side, and your offline path handles the smashed-email example from the PR body well.

A few things to fix before merge:

1. **Read HTML via `Clipboard.read()`** — The Raycast API exposes an `html` flavor on `Clipboard.read()` (see our clipboard docs and extensions like read-my-screen). `clipboard-html.ts` still shells out through `osascript` and returns `""` on failure, which silently drops bold, links, and structure from rich copies. Please switch the read path to the SDK (and only fall back to plain text when `html` is actually missing).

2. **Tighten offline sign-offs** — In `restore-paragraphs.ts`, the case-insensitive `SIGN_OFF` regex treats mid-body lines like “Thanks for the update” as closings because `/i` makes `[A-Z]` match lowercase words. I reproduced a spurious paragraph break on a normal sentence. Restrict sign-offs to real closings (e.g. line-start / short trailing name, or drop `/i` on the name capture).

3. **Preserve intentional line breaks** — When `isWallOfText` is true but the text already has one or two single newlines, `splitSentences` still collapses them to spaces before rebuilding. Consider honoring existing `\n` as hard breaks before heuristic paragraph insertion.

4. **AI truncation** — In `ai-format.ts`, treat `stop_reason` of `max_tokens` (or an incomplete response) as failure: fall back to offline on `auto`, or surface an error on `ai`, instead of writing a truncated clipboard.

5. **Preferences typing** — Drop the hand-written `Preferences` interface in `format-clipboard.tsx` and use the generated `Preferences` type from `raycast-env.d.ts`.

Optional follow-up: if `Clipboard.copy({ text, html })` reliably sets both flavors on current Raycast builds, you may be able to simplify `clipboard-write.ts`; if not, a short comment pointing to the reproduction helps future reviewers.

---

PR #30948, opened 2026-09-10
Add magic-eraser extension

dakotayounger:
## Description

Adds **Magic Eraser** — a single `no-view` command that removes list scaffolding from the clipboard while keeping every other piece of formatting intact.

The problem it solves: copying a Notion to-do list, a checklist, or a nested bullet list and pasting it into prose leaves literal `[ ]`, `•`, `-`, and `1.` markers behind. Raycast's "paste without formatting" does the opposite of what is needed here — it strips the bold, italics, and links you want to keep, while the list junk survives as plain text.

Magic Eraser inverts that: it removes only bullet, numbered, and checkbox markers at the start of each line, and preserves bold, italics, links, headings, and paragraph spacing.

How it works:

- Reads the **HTML flavor** of the clipboard (via `osascript`), since that is where rich sources keep bold, italics, and links, and converts it to markdown with `turndown`. Falls back to plain text when there is no HTML flavor.
- Strips leading list markers line by line: `-`/`*`/`+`, `1.`/`1)`, raw bullet glyphs (`•`, `●`, `▪`, `◦`, `‣`, `·`), and checkbox markers (`[ ]`, `[x]`).
- Writes **both** clipboard flavors back — cleaned markdown as plain text and rendered HTML as rich text — so Gmail and Google Docs get formatting while Slack and Notion get markdown.

No preferences, no configuration, no network calls, no third-party services.

Source repo: https://github.com/dakotayounger/raycast-extensions

## Screencast

This is a `no-view` command with no UI to capture — it operates on the clipboard and reports via HUD. Behavior:

**Before** (clipboard contents copied from a Notion to-do list):

```
[ ] Review the **Q3 pricing** deck before Thursday
[x] Send [the contract](https://example.com/contract) to legal
- Follow up with *Sarah* about the case study logo
1. Confirm the annual number on slide 4
```

**After** running Magic Eraser:

```
Review the **Q3 pricing** deck before Thursday
Send [the contract](https://example.com/contract) to legal
Follow up with *Sarah* about the case study logo
Confirm the annual number on slide 4
```

Note that the bold, italics, and the link all survive — only the markers are gone.

HUD confirmation on success: `🧹 List markers removed — formatting kept`

## Checklist


0xdhrv (commented):

This is distinct from **Clean Text** (its “remove numbering” option only strips plain-text bullets and numbers) and from **Magic Formatter** in #30947 (paragraph restoration). Neither removes Notion-style `[ ]` / `[x]` markers while keeping bold, italics, and links on both clipboard flavors.

- **Clipboard HTML:** `readClipboardHtml` in `src/lib/clipboard-html.ts` shells out to `osascript` and returns an empty string on failure, so rich copies silently fall back to plain text and lose the bold and links this command is meant to keep. `Clipboard.read()` already returns an `html` flavor. Could you read that, and only fall back to plain text when `html` is actually missing?
- **Markdown escaping:** `html-to-markdown.ts` replaces Turndown’s `escape` with an identity function, then `markdown-to-html.ts` runs the result through `marked`. Literal text such as `# Release notes` or `[label](value)` can turn into a heading or a link. Could you keep Turndown’s escaping, or only special-case the merge-tag syntax you care about?
- **Platforms:** The read and write paths use AppleScript and `NSPasteboard`, and `platforms` is unset. Could you set `"platforms": ["macOS"]` in `package.json`?
- **Unused assets:** The manifest uses `assets/icon-v2.png`. Could you remove `assets/icon.png` and `assets/icon.svg` if nothing references them?
- **Spacing (suggestion):** `htmlToMarkdown` collapses three or more newlines and trims, and the plain-text path also `trim()`s, so extra blank lines at the edges do not survive. If the Store copy promises spacing is left intact, could you preserve non-list whitespace?

---

PR #30949, opened 2026-09-10
fix(agent-usage): preserve 5h/weekly usage display when quota exhausted

skyblue:
When the 5h or weekly quota runs out, `getIntervalPercent` / `getWeeklyPercent` return null (total=0, status not active, no `remaining_percent` field returned). The detail panel used to hide the entire row in that case, but the user still wants to see "0% remaining" plus the reset countdown while reset is pending.

- `src/minimaxcn/renderer.tsx` — keep the 5h/weekly rows visible when `percent === null` but `remains_time > 0`; fall back to a "0% remaining" placeholder
- `src/minimaxcn/parser.test.ts` — add a case for the exhausted-but-reset-pending scenario
- `src/minimaxcn/formatMinimaxCNUsageText` — same rule applied to the copy/clipboard text format for consistency

Local `npm run typecheck`, `npm run build`, and `node --test src/minimaxcn/parser.test.ts` all pass (8/8 tests).

### DIFF
--- a/extensions/agent-usage/CHANGELOG.md
+++ b/extensions/agent-usage/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Agent Usage Changelog
 
+## [Fixed exhausted quota display] - {PR_MERGE_DATE}
+
+- Keep MiniMax CN 5h and weekly usage visible with 0% remaining and reset countdowns when quotas are exhausted.
+- Apply the same display to copied usage text.
+
 ## [Cached usage and background refresh] - 2026-09-09
 
 ### Improvements
--- a/extensions/agent-usage/CHANGELOG.md
+++ b/extensions/agent-usage/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Agent Usage Changelog
 
-## [Fixed exhausted quota display] - {PR_MERGE_DATE}
+## [Fixed exhausted quota display] - 2026-09-10
 
 - Keep MiniMax CN 5h and weekly usage visible with 0% remaining and reset countdowns when quotas are exhausted.
 - Apply the same display to copied usage text.
--- a/extensions/agent-usage/src/minimaxcn/auth.ts
+++ b/extensions/agent-usage/src/minimaxcn/auth.ts
@@ -29,6 +29,12 @@ function extractMarkedValue(output: string, startMarker: string, endMarker: stri
 }
 
 async function readShellEnvTokens(envKey: string): Promise<string | null> {
+  // The lookup below uses POSIX `printf` plus the `-ilc` flag, which Windows shells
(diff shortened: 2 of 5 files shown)


skyblue:
All 3 Greptile review findings addressed in this push:

1. **`renderer.tsx` (accessory function)** — Mirrored the detail-panel rule: when `getIntervalPercent` / `getWeeklyPercent` return null but `remains_time > 0` (reset still pending), treat as 0% instead of "no data". The accessory now uses `shownIntervalPercent = intervalPercent ?? (remains_time > 0 ? 0 : null)` so the list view and menu bar correctly display "0%" instead of "—" while reset is pending.

2. **`auth.ts` (readShellEnvTokens)** — Windows now skips the shell-lookup path entirely (POSIX `printf` + `-ilc` aren't implemented there) and relies on `process.env` inheritance. macOS path unchanged.

3. **`fetcher.ts` (parseMinimaxCNApiResponse)** — Added `Array.isArray(response.model_remains)` guard: a truthy non-array value now returns a `parse_error` instead of crashing later in `getCodingModelRemain(...).find(...)`.

Local `npm run typecheck`, `npm run build`, and `node --test src/minimaxcn/parser.test.ts` all pass (8/8). New commit `f2453a4c`.

skyblue:
@vladimirtomic @EDM115 (and @greptile-apps — please re-review against the current head)

This PR has been rebased onto the latest `origin/main` since the first review. The current head `a1daaf1f` addresses all 3 Greptile findings from the original review:

1. `renderer.tsx:152` accessory — `getMinimaxCNAccessory` now treats a null percent as 0% when `remains_time > 0`, so the list view and menu bar display "0%" instead of "—" while reset is pending. The exact suggestion Greptile offered.

2. `auth.ts:42` Windows shell lookup — `readShellEnvTokens` now early-returns null on `process.platform === "win32"` instead of failing on POSIX-only `printf -ilc`. The macOS path is unchanged.

3. `fetcher.ts:55` malformed response — `parseMinimaxCNApiResponse` now guards `response.model_remains` with `Array.isArray` and returns `parse_error` for truthy non-array values, instead of letting `.find` blow up later in `getCodingModelRemain`.

Local `npm run typecheck`, `npm run build`, and `node --test src/minimaxcn/parser.test.ts` all pass (8/8 tests). Diff is 4 files / +66 / -15.

EDM115 (approved):
lgtm  
idk if the same logic could be applied to the non-CN version of the MiniMax provider but if you don't have access to it let's not guess their API schema & do it without proper testing. surely someone else could pick this up if they want

EDM115:
oh didn't saw but you need to add an entry to the changelog !

pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #30951, opened 2026-09-10
Add copy-tab-link extension

jnbp:
## Checklist


pernielsentikaer (changes requested):
Hi @jnbp 👋

We already have [Copy Path](https://github.com/raycast/extensions/tree/main/extensions/copy-path) by @koinzhang in the Store, and it already copies the URL of the current browser tab (optionally with the tab title) from a hotkey — that's your "Copy Tab as Plain Text" workflow. Could you look into adding the formatted output (rich text, Markdown, Slack/Jira/AsciiDoc, etc.) and the URL/title cleanup to Copy Path instead of shipping a separate listing? If there's a difference I'm missing that justifies keeping this separate, could you explain it here? 🙂

---

PR #30954, opened 2026-09-10
Update threads extension

chrismessina:
## Description

**Download Threads Media had stopped working for every post.** Both third-party services it relied on are gone — `api.threadsphotodownloader.com` no longer resolves in DNS at all, and DolphinRadar answers `200` with `{"code":24001,"message":"Post not found"}` for public posts. The old code read the media list off that body without checking `code`, threw a `TypeError`, swallowed it, and surfaced "No images or videos found in this Threads post" — so every failure blamed the post rather than the dead dependency.

Media now comes from threads.com itself, with no intermediary. That also fixed the reported inability to use `threads.com/share/…` links, and made it possible to support carousels and voice posts, neither of which the old path could reach.

### Changes

- **Read media from Threads directly** instead of two dead third-party APIs.
- **Support `/share/…` short links**, resolved to the canonical post URL before downloading.
- **Support carousels** — every item is downloaded independently, so one failure no longer abandons the rest.
- **Support voice posts**, which previously reported "No media found". Saved as `.m4a` rather than the `.mp4` the server's content type claims, because the bytes are AAC audio with no video track.
- **Name files from the response, not the URL.** Threads serves a `.jpg` URL as WebP on some posts and JPEG on others; the old naming produced `.jpg` files that several macOS apps refuse to open.
- **Never overwrite an existing file.** Downloads stream to a temporary file and are renamed only on success; a taken name becomes `name (1).ext`.
- **New Image Format preference** — keep the original, or convert to JPEG or PNG (macOS only, via the built-in `sips`; the original is kept if conversion fails).
- **New Debug Logging preference** for diagnosing a post that will not download. Signed CDN URLs are redacted from logs and from the Copy Error payload.
- **Better failure messages** — a post Threads refuses to serve anonymously now says so, instead of claiming the link is invalid. Every failure toast carries a Copy Error action.
- **Robustness**: reject non-media and empty responses, time out a stalled download, refuse a concurrent run, and show progress in bytes when the server omits `Content-Length`.
- **Housekeeping**: migrate to `@raycast/api` 2.x, drop the now-unused `axios`, `cheerio` and `node-fetch`, add a vitest suite, add the MIT `LICENSE` file, and replace `WARP.md` with `AGENTS.md`.

## Checklist

- [x] I checked that this change does not break existing commands or remove
      functionality


pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #30956, opened 2026-09-10
Update attio extension

chrismessina:
## Description

Adds record creation and Raycast AI support to the Attio extension. New `Create Person`, `Create Company`, and `Create Deal` commands open a schema-driven form with draft support, and every record list gets a `New <Object>` action (`⌘N`) — including a create path straight from a "no matches" search. The edit form now also covers person names (split First/Last, matching Attio's editor), owners, and single record links such as a person's Company, via a live record-search picker.

### Changes

- Create Person / Create Company / Create Deal commands with Raycast draft support; `⌘N` from any record list (custom objects included, via an inline form)

image

image

image

- Create/edit forms render every writable attribute from the workspace schema: split name fields, workspace-member owner dropdown, tag-style multi-value entry (domains validated for well-formed TLDs), inline required/invalid validation
- Editing now supports names, owners, and single record links (async record search scoped to the attribute's allowed objects); one PUT per save, changed values only
- Record action panels reordered for consistency: open/pin/new, edit + copy, view controls, then export with delete always last
- Success toast after creating offers Open <Object> (in Raycast) and Open in Attio

- **AI Extension**: `@attio` tools for Search Records, Create Person, Create Company, and Create Deal — writes behind confirmations, deal stages resolved case-insensitively, owners matched by member name or email; `ai.yaml` ships instructions plus 11 evals (11/11 passing via `ray evals`)

## Checklist

### DIFF
--- a/extensions/attio/CHANGELOG.md
+++ b/extensions/attio/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Attio Changelog
 
-## [Create Records + AI Tools] - {PR_MERGE_DATE}
+## [Create Records + AI Tools] - 2026-09-12
 
 ### Create People, Companies, and Deals
 
--- a/extensions/attio/CHANGELOG.md
+++ b/extensions/attio/CHANGELOG.md
@@ -1,5 +1,25 @@
 # Attio Changelog
 
+## [Create Records + AI Tools] - {PR_MERGE_DATE}
+
+### Create People, Companies, and Deals
+
+- Dedicated Create Person / Create Company / Create Deal commands with Raycast draft support — dismiss mid-entry and pick the form back up later
+- New Record (`⌘N`) from every record list — the standard objects open their Create command; custom objects get the same schema-driven form inline
+- Schema-driven form: writable attributes from your workspace schema, inline required/invalid validation, tag-style entry for multi-value fields (domains validated for well-formed TLDs), person names parsed into Attio's name format, deal owners picked from workspace members
+- Success toast offers Open in Attio for the record you just created
+- Record action panels reordered: open/pin/new up top, edit and copy together, view controls, then export with delete always last
+
+### Editing
+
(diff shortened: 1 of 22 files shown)


pernielsentikaer (changes requested):
Hi @chrismessina 👋

Thanks for adding record creation and AI tools 💪

When reopening a Create Person draft containing a Company link, `RecordCreateForm.tsx` restores the encoded value but doesn’t seed `RecordRefPicker`’s items. The saved relationship is still submitted despite having no matching dropdown item.

Could you preserve the restored selection before search results load, resolving its title where possible? A fallback in `RecordRefPicker.tsx`, after assembling `items`, would be:

```tsx
if (props.value && !items.has(props.value)) {
  items.set(props.value, props.value);
}
```

---

chrismessina:
Thanks @pernielsentikaer 👋 Good catch — fixed in `RecordRefPicker.tsx`.

I went slightly past the suggested fallback: rather than showing the raw encoded ref as the title, the orphaned value is resolved through the existing `useRecordTitles` hook, so a reopened draft shows the actual company name. It falls back to the raw ref only if the lookup fails, so the saved relationship is never lost:

```tsx
const orphan =
  props.value && props.value !== props.initial?.value && props.value !== picked?.value && props.value.includes(":")
    ? props.value
    : "";
const { titleFor } = useRecordTitles(orphan ? [decodeRef(orphan)] : []);
// ...after assembling items:
if (orphan && !items.has(orphan)) items.set(orphan, titleFor(decodeRef(orphan).target_record_id)?.title ?? orphan);
```

It only fires when the value matches neither the initial value, the picked one, nor a search hit, so the normal path is unchanged. Verified by hand: create a Person, link a Company, pop back out to the draft, reopen — the selection is restored with its name.

Marking as ready for review 🙏

pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #30957, opened 2026-09-10
Update Digger extension

chrismessina:
## Description

Digger gains two sections and a set of export actions, and fixes three cases where it stated something it had not established.

image

**Well-Known** probes the paths under `/.well-known/` and reports which ones a site actually publishes. A path prefix cannot be listed, so each is requested individually — and judged by its opening bytes rather than its status code, because a single-page app answers `200` to every path it does not recognise.

image

image

**Theme** shows the colors a page declares: `theme-color` per media query, `color-scheme`, the `data-theme` family, and the CSS custom properties in its stylesheets. Modern color syntaxes and `var()` chains are resolved so each token draws a real swatch, viewable as a list or a color grid.

### Changes

**Added**

- **Well-Known section** — 82 GET-answerable paths from the IANA Well-Known URIs registry, plus a few widely deployed but unregistered ones (`apple-app-site-association`, `assetlinks.json`). Contents open in place, with copy and download actions. Paths that could not be checked are reported separately from paths the host answered for.
- **Theme section** — `oklch()`, `lab()`, `lch()`, `color(display-p3 …)` and `var()` references are resolved to hex so they render as swatches. A grid view shows the palette; the full set can be expanded to read every linked stylesheet.
- **Copy and Download actions** on `robots.txt`, `sitemap.xml` and each well-known file — as text, Markdown or CSV. A sitemap's `<url>` entries and `robots.txt`'s directives expand into rows.
- **Copy actions in Metadata** for page title, description, Open Graph description and Open Graph image URL.
- **Code Formatting preference** — pretty-prints minified JSON and XML in the file viewer. Copy and Download always yield the original bytes.

**Fixed**

- **Pages fetched in the wrong language.** Digger sent no `Accept-Language`, so sites that content-negotiate chose one; one returned `lang="ar-AR"` for an English page, and every metadata field read from it described the wrong translation.
- **`sitemap.xml` reported as malformed when a site simply has none.** A `200` serving an HTML error page now reads as absent, and a `406` as unavailable, rather than "could not parse sitemap structure".
- **Truncated overflow labels.** "…and 9 more" rendered as a clipped `......ore` in four sections; the text was in the value column, which is right-aligned and truncates from the left.

**Housekeeping**

- Removes `TODO.md`, which was included in an earlier release and is not useful to readers of this repository.
- `@raycast/api` 2.1.3 → 2.2.1.

## Checklist


0xdhrv (approved):
Looks good to me, approved ✅

PR #30959, opened 2026-09-10
Update google-maps-search extension

krsntn:
## Description

Setting **Preferred Starting Location** to `Home` (with a Home Address filled in) had no effect on the **Get Me Somewhere** command. The "Starting from" dropdown correctly displayed `Home`, but the generated directions always started from the user's current location.

## Root cause

In `src/travelTo.tsx`, the dropdown selection and the address used to build the URL are two separate pieces of state:

- `origin` was initialized from `preferences.preferredOrigin`, so the dropdown rendered `Home` as expected.
- `originAddress` was hardcoded to `""` on mount, and was only ever populated inside `handleOriginChange`, which fires exclusively on a *manual* dropdown change.

So on first open the two disagreed, and `makeDirectionsURL` was called with an empty origin. Google Maps interprets an empty `&origin=` as the current location, which is why the preference appeared to be ignored, selecting `Home` by hand in the dropdown worked, which masked the bug.

## Fix

Seed `originAddress` from the preference on mount so it matches the initial dropdown value:

```ts
const [originAddress, setOriginAddress] = useState<string>(
  preferences.preferredOrigin === OriginOption.Home ? preferences.homeAddress : ""
);
```

handleOriginChange is untouched; the Current Location and Custom defaults are unchanged (both correctly start empty).

## Checklist

### DIFF
--- a/extensions/google-maps-search/CHANGELOG.md
+++ b/extensions/google-maps-search/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Google Maps Search Changelog
 
-## [Prefill Home Origin Address] - {PR_MERGE_DATE}
+## [Prefill Home Origin Address] - 2026-09-10
 
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #30987, opened 2026-09-11
[Menubar Weather] Fix type error when showing feels-like temperature

### LINKED ISSUE #30896: [Menubar Weather] Menu bar is showing a type error when using feels like temperature
### Extension

https://www.raycast.com/koinzhang/menubar-weather

### Raycast Version

2.2.1.0

### OS Version

macOS [Unreleased] macOS 27 27.0.0 arm64

### Description

TypeError: Cannot read properties of undefined (reading 'apparent_temperature')

tl:common-utils.ts:199:28

---
196:     menuItems.push(
197:       Math.round(
198:         tempType == "apparent_temperature"
199:           ? weather?.hourly.apparent_temperature[timeHour()]
                                ^
200:           : weather?.current_weather?.temperature,
201:       ) + tempUnit,
202:     );
203:     if (menuUVI && weather.daily?.uv_index_max.length != 0) {

---

React Component Stack
Pc:menubar-weather.tsx:21:30
sA:index.js:177:7566
    at ray-navigation-stack (<anonymous>)
Nd:index.js:177:7067

JavaScript Stack
tl:common-utils.ts:199:28
menubar-weather.js:42:14673
Object.useMemo:react-reconciler.production.js:8:96884
exports.useMemo:react.production.js:514:33

### Steps To Reproduce

Use the menu bar command and enable feels like temperature

### Current Behaviour

_No response_

### Expected Behaviour

_No response_

szupzj18:
Fixes #30896

## Root cause

With the temperature preference set to "Feels like", `getMenuItem` read `weather?.hourly.apparent_temperature[...]`. The optional chaining only covered `weather` — when the latest weather data was missing the `hourly` section (a failed or partial API response, or a stale/invalid cached payload), reading `apparent_temperature` threw `TypeError: Cannot read properties of undefined (reading 'apparent_temperature')` and crashed the menu bar command.

## Changes

- `common-utils.ts`: guard every optional section in `getMenuItem`; when the hourly feels-like value is unavailable, fall back to the current temperature; skip the temperature item entirely if neither exists instead of rendering `NaN℃`. The wind item is similarly skipped when `current_weather` is missing, and the humidity unit falls back to an empty string.
- `axios-utils.ts`: responses that are not valid weather objects or carry an Open-Meteo `error` flag are rejected, so they are never written to the latest-weather cache (which `menubar-weather.tsx` falls back to while a refresh is in flight).

## Validation

- Reproduced the crash with a harness importing the real `getMenuItem`: partial payloads (`{}`, a non-object cached body, `current_weather` only, missing `daily`) all threw the reported `TypeError` before the fix and return safe item lists after; a complete payload still renders `25℃ | ☀ 5 | ↑ 10Km/h`.
- `ray lint` clean; `tsc --noEmit` reports only the pre-existing `swift:` module-resolution warning; `ray build` fails the same way on pristine `main` in this environment (Swift module export resolution) — the failure is unrelated to these changes.

### DIFF
--- a/extensions/menubar-weather/CHANGELOG.md
+++ b/extensions/menubar-weather/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Menubar Weather Changelog
 
-## [Fix Type Error When Showing Feels-Like Temperature] - {PR_MERGE_DATE}
+## [Fix Type Error When Showing Feels-Like Temperature] - 2026-09-11
 
 - Fix the menu bar command crashing with a "Cannot read properties of undefined (reading 'apparent_temperature')" type error when the temperature display is set to "Feels like" and the latest weather data is incomplete (e.g. a failed or partial API response); it now falls back to the current temperature and skips unavailable readings instead of failing.
 - Invalid or error responses from the weather API are no longer cached.
--- a/extensions/menubar-weather/CHANGELOG.md
+++ b/extensions/menubar-weather/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Menubar Weather Changelog
 
+## [Fix Type Error When Showing Feels-Like Temperature] - {PR_MERGE_DATE}
+
+- Fix the menu bar command crashing with a "Cannot read properties of undefined (reading 'apparent_temperature')" type error when the temperature display is set to "Feels like" and the latest weather data is incomplete (e.g. a failed or partial API response); it now falls back to the current temperature and skips unavailable readings instead of failing.
+- Invalid or error responses from the weather API are no longer cached.
+
 ## [Update icons] - 2024-09-02
 
 - Update icons for Street and Country
--- a/extensions/menubar-weather/package.json
+++ b/extensions/menubar-weather/package.json
@@ -230,5 +230,8 @@
     "fix-lint": "ray lint --fix",
     "lint": "ray lint",
     "publish": "npx @raycast/api@latest publish"
-  }
+  },
+  "platforms": [
+    "macOS"
+  ]
 }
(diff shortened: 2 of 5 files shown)


pernielsentikaer (changes requested):
Hi @szupzj18 👋

Thanks for the crash fixes 💪

Two partial-data paths still crash:

- `isValidWeatherResponse` accepts missing `daily.time` and `current_weather.time`, but the menu indexes/formats both. Could you validate the strings and array indexes used by the full menu? This is the remaining part of [Greptile’s validation feedback](https://github.com/raycast/extensions/pull/30987#discussion_r3985518062).
- `useLatestWeather` can return old `useCachedPromise` data, which bypasses the new validation. Please validate that before using it too:

```tsx
if (isValidWeatherResponse(weatherData)) {
  return weatherData;
}
// Otherwise, try the validated cache fallback.
```

I reproduced both with mocked rendering of the actual component.

---

szupzj18:
Thanks for the thorough mock reproduction — both are fixed in the latest push.

1. **Full-menu shape validation.** `isValidWeatherResponse` now validates every field the menu indexes or formats:
   - `current_weather.time` must be a string (used by the Last Updated item), alongside the numeric `temperature`/`windspeed`/`winddirection`/`weathercode` scalars.
   - The hourly arrays the menu reads at the current-hour index (`apparent_temperature`, `relativehumidity_2m`, `surface_pressure`, `visibility`) must each have at least 24 entries of `number | null`, so the index `timeHour()` can never be out of range.
   - Daily `time`/`sunrise`/`sunset` must be non-empty string arrays, and every daily array (strings + numerics) must have the same length, since the forecast submenus index `daily.time` at each mapped position and call `.substring()` on it.
   - `hourly_units` strings for the rendered units are checked; `daily_units.rain_sum` is checked when present.

2. **Stale `useCachedPromise` data.** `weatherData` is now validated before use, and only then do we fall back to the `Cache` entry — which is validated as well, with a corrupt/non-JSON cache entry ignored.

I verified with a harness that imports the actual source: 35 cases covering both of your reproductions (missing `daily.time`/`current_weather.time`, truncated hourly arrays, partial sections, stale cache shapes) — the validator rejects all of them and the full menu still renders correctly with a complete payload.

pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #31007, opened 2026-09-11
Update zlib-search extension

materialcritic:
## Description

This adds bulk-download support to the Z-Library extension, in two independent ways:

### 1. Multi-select in Search Books

- **Select** (⌘S) toggles a checkmark on any book in the results, across as many searches as you like in the same session
- **Download Selected (N)** (⌘⇧D) — appears once at least one book is selected — downloads them one at a time (sequential, not parallel, to stay kind to Z-Library's rate limits), with a single progress toast ("Downloading 2 of 5…")
- Books that fail stay selected so they're easy to retry; successful ones are cleared automatically

### 2. Download Queue (save for later)

- **Add to Queue** (⌘B) on any book in Search Books saves it for later without downloading immediately, persisted via Raycast's `LocalStorage` so it survives Raycast restarts
- New **Download Queue** command lists everything you've saved, split into "Queued" and "Downloaded" sections
- From there: download one book, download everything pending at once (same sequential download with progress toast), remove a single item, or clear everything already downloaded
- Downloaded items stay in the queue marked **Downloaded** (rather than disappearing) until removed or cleared, so you can see what you've already grabbed

Both features share the same download + progress-toast logic (now factored into `src/lib/zlib.ts`), which is the same logic already used by the existing single-book Download action.

## Screencast

Verified locally via `ray develop`: selected multiple books and downloaded them together, added a book to the queue, downloaded it from the new Download Queue command, and confirmed the queue (with its Downloaded marker) persists across a full Raycast restart.

## Checklist

### DIFF
--- a/extensions/zlib-search/CHANGELOG.md
+++ b/extensions/zlib-search/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 All notable changes to this project will be documented in this file.
 
-## [Bulk Download] - {PR_MERGE_DATE}
+## [Bulk Download] - 2026-09-11
 
 ### Added
 
--- a/extensions/zlib-search/CHANGELOG.md
+++ b/extensions/zlib-search/CHANGELOG.md
@@ -2,6 +2,15 @@
 
 All notable changes to this project will be documented in this file.
 
+## [Bulk Download] - {PR_MERGE_DATE}
+
+### Added
+
+- Select multiple books in Search Books and download them all at once with Download Selected
+- Add to Queue action to save books for later without downloading immediately
+- New Download Queue command to review, bulk-download, and manage saved books, even across Raycast restarts
+- Downloaded queue items are marked Downloaded and kept until manually cleared or removed
+
 ## [1.0.0] - 2026-09-10
 
 ### Added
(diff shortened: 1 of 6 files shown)


0xdhrv (commented):
Hi @materialcritic 👋

Thanks for the contribution 🔥

Also, please look into the failing check

---

materialcritic:
Thanks for the review, @greptile-apps! Both issues are fixed:

1. **Bulk status updates overwriting each other** — `handleDownloadAllPending` was calling `markDownloaded`, which closed over the `queue` value from render time. Since the bulk download loop is sequential and awaits between items, every `onEachResult` call within the same invocation saw that same stale `queue`, so each write started from the pre-download state and clobbered the previous item's `downloaded: true`. Fixed by tracking a running local copy that's updated and persisted after each result instead of recomputing from the stale closure.

2. **Cross-search selections silently dropped** — selections were a `Set<string>` of book ids resolved against the *current* search's `books` array. Once the user ran a new search, a previously selected book was no longer in `books`, so it was silently excluded from `Download Selected`, and the subsequent `setSelectedIds(failedIds)` reset discarded it entirely. Fixed by storing selections as `Map<string, Book>` (id → full book), so a selection keeps its own data and survives across searches, exactly as the README describes.

Marking as ready for review 🙏

0xdhrv (approved):
Looks good to me, approved ✅

PR #31008, opened 2026-09-11
Update Herdr: fix agent focus and agent names

mikluko:
## Description

Two bug fixes in the Herdr extension. No manifest, command, asset or dependency changes.

**Focusing an agent left its pane off screen.** `herdr agent focus` moves the server's focus but leaves the attached client drawing the tab it is already on, so the selected agent never came into view. Workspaces and tabs were unaffected, because `workspace focus` and `tab focus` are exactly what move the client. `focusPane` already solves this for panes, and `agent focus` answers with the agent record, so the pane id needed to reuse it comes free.

This works around herdr's own behaviour rather than fixing it there, so it may be worth a herdr-side issue too. Verified against herdr 0.9.0.

**Agents were named after their working directory.** `terminal_title_stripped` carries the session title an agent sets for itself, and the fallback chain skipped it, so two agents running in one repository were indistinguishable. It now precedes the directory basename. A title that is only the agent's own product name (`"Claude Code"`) identifies nothing once several agents run, so that case still yields to the directory.

Both changes are covered by unit tests. The focus test asserts the full command sequence, so it fails against the previous single-call behaviour.

## Screencast

No UI change to show: both fixes are behavioural. Verified instead by instrumenting the focus path in a dev build and driving it from the Dashboard, confirming herdr's snapshot landed on the requested workspace, tab and pane, and that the client followed.

Scope of what was exercised, for transparency: macOS with Ghostty as the terminal, herdr 0.9.0, Claude Code agents. Other terminals and agent kinds are unchanged by these diffs but were not exercised.

## Checklist

`npm run build`, `npm run typecheck`, `npm test` (69 tests) and `npm run lint` all pass. The two asset boxes are ticked as unchanged: this PR adds and touches no assets.

### DIFF
--- a/extensions/herdr/CHANGELOG.md
+++ b/extensions/herdr/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## [Fix Agent Focus and Agent Names] - {PR_MERGE_DATE}
+## [Fix Agent Focus and Agent Names] - 2026-09-11
 
 - Switch to the agent's tab when focusing an agent. `agent focus` moves the server's focus but leaves the attached client drawing the tab it is already on, so the agent's pane never came into view. Workspaces and tabs were unaffected, because `workspace focus` and `tab focus` are what move the client.
 - Name agents by the session title they set, rather than by their working directory. Two agents in one repository were indistinguishable. A title that is only the agent's own product name still yields to the directory.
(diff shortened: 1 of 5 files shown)


0xdhrv (approved):
Looks good to me, approved ✅

PR #31012, opened 2026-09-11
Add proton-vpn extension

jorgetoh:
## Description

Adds **Proton VPN**, an extension that controls the Proton VPN macOS app from Raycast: check the status, connect, disconnect, and switch countries.

The extension drives the app the user already has. It uses only the installed Proton VPN app plus built-in macOS tools (`scutil`, `defaults`, `sqlite3`, `open`, `osascript`). It needs no extra binaries, no credentials, and no WireGuard configuration files.

**Commands**

| --- | --- | --- |
| Manage Connection | `view` | Live status with disconnect and reconnect, a fastest-server action, recent countries, and all countries nearest first with distance, load, and server count. |
| Switch Country | `no-view` | Takes a country name or a two-letter code as an argument and connects to the fastest server there. |
| Toggle Connection | `no-view` | Connects if disconnected, disconnects if connected. Made for a global hotkey. |
| Menu Bar Status | `menu-bar` | Status icon with the country flag, connect and disconnect actions, and a nearest-countries submenu. |

**How it works** (also documented in the README)

- **Status and quick connect/disconnect** use the system VPN service that the app registers with macOS (`scutil --nc`).
- **The country list** comes from the local server database of the app (`database.sqlite` in the app container): country, city, coordinates, tier, and live load. Distances are computed from the current country as the app recorded it.
- **A country switch** writes a connection profile ("Raycast Switch") into the preferences of the app, marks it as the auto-connect profile, and relaunches the app in the background. The app then resolves the fastest server itself. The auto-connect flag is cleared immediately after, so a normal app launch does not auto-connect.
- **A full disconnect**: when the app made the connection, macOS enforces an on-demand rule that reconnects the tunnel after `scutil --nc stop`. Only the app can disable that rule, and it does so while it quits. The extension starts that path without the confirmation dialog, so the app disconnects cleanly and quits.

The README documents the known side effects openly: a "Raycast Switch" profile appears in the profile list of the app, a switch takes about 15 to 25 seconds because the app restarts, and macOS asks once for permission when Raycast sends the quit command.

**Notes**

- No network calls, no analytics, no Keychain access. Everything happens on the user's Mac.
- Requires the Proton VPN app installed and signed in, plus one connection made in the app, which registers the system VPN profile.
- Country selection needs a paid plan, because the free plan cannot select a country in the official app either.
- The extension is not affiliated with Proton AG. The disclaimer is in the README.

## Screencast

Four screenshots are in `metadata/`. All data in them is synthetic (an invented home location and server name), so the images contain no real IP address, location, or server.

## Checklist


jorgetoh:
All three review findings were valid and are fixed in `bcab65f`. Short note on each, since I resolved the threads:

**1. Saved profiles are replaced** — correct, and confirmed against a real preferences file: `profiles_<userId>` holds the whole saved-profile array, so writing a single-element array deleted every other profile. `connectToCountryImpl` now reads the existing array, filters out its own `raycastswitch0000001` entry, and appends the new one. The README claim that only the "Raycast Switch" profile is touched is now accurate.

**2. Reconnect stops before connecting** — correct. `scutil --nc start` returns while the service still reports `Disconnected`, so `waitForState(["Connected", "Disconnected"])` matched that initial state and reported a failure. The reconnect path now waits for `Connecting` or `Connected` first, and only then treats `Disconnected` as a failure. The unused `connectAndWait` helper in `vpn.ts` had the same flaw and is removed.

**3. Concurrent commands corrupt state** — correct. Every transition used one shared key with no owner. Each transition now carries a UUID, `clearTransition(id)` removes the record only when that id still owns it, and `withTransition` refuses to start a second operation while one is running. It reconciles against the real VPN state first, so a terminated command does not block the next one.

PR #31021, opened 2026-09-11
fix(skills): stop commands waiting indefinitely behind each other

DaleSeo:
## Description

Follow-up to #30985, which raised the timeout for mutating `skills` commands from 30 seconds to 5 minutes. Only one command may use the CLI at a time, and the wait for that turn was unbounded, so lengthening the run also lengthened how long everything else sat behind it. Triggering "Update All Skills" and then opening Manage Skills left the list empty and spinning, with no way to tell a slow update from a hang.

The read-only budget is an arbitrary number, and it only treats the symptom because the time is spent downloading, not listing skills. Separating resolution from execution would require spawning a second process for every command, which did not seem worthwhile here.

## Screencast

No UI Update

## Checklist


0xdhrv (approved):
Looks good to me, approved ✅

PR #31026, opened 2026-09-11
Toggl Track: fix duplicate billable checkbox form ID

speckofthecosmos:
## Description

`CreateTimeEntryForm` renders two `Form.Checkbox` elements that share the same `billable` form ID:

```tsx
// inside the project block
{selectedProject?.billable && <Form.Checkbox id="billable" label="" title="Billable" />}
// near the end of the form
{isWorkspacePremium && <Form.Checkbox id="billable" label="Billable" value={billable} onChange={setBillable} />}
```

When a **billable project** is selected in a **premium workspace**, both conditions are true and both render, so the form has a duplicate ID. The two also disagree: one is uncontrolled with an empty `label`, the other is controlled by the `billable` state.

This consolidates them into the single controlled checkbox, displayed whenever either condition applies — so the option is still offered in both situations, without the collision.

No change for workspaces/projects that only ever satisfied one of the two conditions.

## Screencast

Not included — the visible change is one `Billable` checkbox instead of two in the overlapping case.

## Checklist

### DIFF
--- a/extensions/toggl-track/CHANGELOG.md
+++ b/extensions/toggl-track/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Toggl Track Changelog
 
+## [Fix Duplicate Billable Checkbox] - {PR_MERGE_DATE}
+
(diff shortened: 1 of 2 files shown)


0xdhrv (approved):
Looks good to me, approved ✅

PR #28757, opened 2026-06-14
[Toggl Track] Add Quickstart New Timer command

speckofthecosmos:
## Description

Adds a **Quickstart New Timer** command that opens the new time entry form directly, skipping the recent-entries list. Assigned a global hotkey, it lets you start a timer from anywhere with a single keystroke; on submit, Raycast closes immediately and shows a HUD.

The change is intentionally small and additive:

- New `createTimeEntry` command (`src/createTimeEntry.tsx`) that renders the existing `CreateTimeEntryForm`.
- `CreateTimeEntryForm` gains an optional `closeWindowOnSubmit` prop (default `false`). Only the new command passes `true` (→ `closeMainWindow` + HUD on submit); every existing call site is untouched and behaves exactly as before (pops back to the list).
- README and CHANGELOG entries.

No new dependencies, and no behavior change for any existing command.

## Screencast

_Screencast to follow._

## Checklist

- [x] I read the [extension guidelines](https://developers.raycast.com/basics/prepare-an-extension-for-store)
- [x] I read the [documentation about publishing](https://developers.raycast.com/basics/publish-an-extension)
- [x] I ran `npm run build` and used the extension in development mode
- [x] I ran `npm run lint` and it passed (ESLint + Prettier)
- [x] I added a `CHANGELOG.md` entry using `{PR_MERGE_DATE}` as the date

PR #31027, opened 2026-09-11
Toggl Track: add Quickstop Timer command

speckofthecosmos:
## Description

Adds a **Quickstop Timer** command: stops the running time entry with no window at all, so it can be bound to a global hotkey and end a timer in one keystroke. It complements the existing **Quickstart New Timer** command (#28757) — start from a hotkey, stop from a hotkey.

A HUD reports what was stopped and how long it ran (`Stopped Writing docs · 01:12:44`), or says nothing was running.

### Notes on the implementation

- **`mode: "no-view"`** rather than `view`: the action takes no input, so Raycast can run the async default export headlessly and tear the process down when it resolves. No window is ever shown.
- **No cached read on the destructive path.** `getRunningTimeEntry()` serves from cache, for up to an hour in Low Data Mode and for the configured TTL otherwise. For a stop, a stale hit is worse than a miss: if the timer was stopped or replaced in another client, Quickstop stops the old entry and leaves the real one running. So the command always goes through a new `refetchRunningTimeEntry()`. That path also evicts `runningTimeEntry` when Toggl reports nothing current, matching the set/remove pairing `menuBar.tsx` and `me.ts` already use on that key.
- **A sibling function rather than `getRunningTimeEntry({ skipCache })`.** `useRunningTimeEntry` calls `useSafeCachedPromise(getRunningTimeEntry, [])`, whose generic infers `Parameters<T>` against the `[]` args tuple — adding even an optional parameter risks breaking inference at the existing call sites.
- **`refreshMenuBar()` moved to `helpers/common.ts`** (it was a private function in `useTimeEntryActions.ts`) and now **returns** the `launchCommand` promise. A one-shot `no-view` command has to `await` it: the process is torn down the moment the command resolves, which would otherwise cut off a fire-and-forget launch and leave the menu bar showing a stopped timer as still running. Existing callers are unaffected.
- Elapsed time is read *before* the stop and clamped with `Math.max(0, …)`, matching the existing `formatSeconds` guard against a mis-entered future start.

The `normalize` closure inside `getRunningTimeEntry` is hoisted to a module-level `normalizeTimeEntry` so the new API path shares it; behaviour is unchanged.

## Screencast

Not included — the command renders no UI by design; the only visible output is the confirmation HUD.

## Checklist

### DIFF
--- a/extensions/toggl-track/CHANGELOG.md
+++ b/extensions/toggl-track/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Toggl Track Changelog
 
+## [Quickstop Timer] - {PR_MERGE_DATE}
+
+- Added a "Quickstop Timer" command that stops the running time entry with no UI, so it can be bound to a global hotkey
+- In Low Data Mode the command confirms the running entry against the API before stopping, so it never acts on a stale cached timer
+
 ## [Fix Duplicate Billable Checkbox] - 2026-09-11
 
 - Fixed two `Billable` checkboxes rendering with the same `billable` form ID when a billable project is selected in a premium workspace — they are now a single checkbox, shown whenever either condition applies
--- a/extensions/toggl-track/README.md
+++ b/extensions/toggl-track/README.md
@@ -26,6 +26,13 @@ Raycast closes immediately and is ready for your next command.
 
 <img width="600" alt="Quickstart New Timer Command" src="./assets/new-entry.png">
 
+### Quickstop Timer
+
+Stops the running time entry immediately, without opening a window. Assign it a
+global hotkey to end a timer from anywhere with one keystroke — a HUD confirms
+what was stopped and how long it ran. If nothing is running, it says so and does
+nothing.
+
 ### Manage Tags
 
 This command list all the tags you have access to. You can create, rename, and delete them, assuming you have the appropriate permissions.
--- a/extensions/toggl-track/package.json
+++ b/extensions/toggl-track/package.json
@@ -40,6 +40,12 @@
       "mode": "view"
     },
     {
+      "name": "stopTimeEntry",
+      "title": "Quickstop Timer",
+      "description": "Stop the running time entry without opening a window.",
+      "mode": "no-view"
+    },
+    {
       "name": "updateTimeEntry",
       "title": "Update Time Entry",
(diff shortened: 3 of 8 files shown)


speckofthecosmos:
Good catch — confirmed and fixed in the latest push.

You're right that the truthy-stale case is the dangerous one, and it's worse than a failed stop: if the timer was replaced in another client, Quickstop stops an already-stopped entry *and* leaves the real one running.

I went slightly further than the suggestion and dropped the cache read entirely in Low Data Mode, rather than reading it and then refetching:

```ts
runningTimeEntry = liteMode ? await refetchRunningTimeEntry() : await getRunningTimeEntry();
```

Same request cost as the suggested version (one API call in LDM), but it can't act on cache state at all for the destructive path. Normal mode is unchanged — it already reaches the API on a miss.

`refetchRunningTimeEntry`'s doc comment was updated to say it exists so a one-shot action never stops the wrong timer.

pernielsentikaer (changes requested):
Hi @speckofthecosmos 👋

Thanks for adding the Quickstop workflow 💪

I found one remaining stale-cache path: in normal mode, `stopTimeEntry.ts` uses `getRunningTimeEntry()`, which can return a cached entry for up to the configured TTL. If the timer was stopped or replaced from another client during that window, Quickstop targets the old entry and leaves the actual timer running.

Could you always use `refetchRunningTimeEntry()` before this destructive action? The authoritative API path should also remove `runningTimeEntry` from the cache when the API returns no current entry, so the menu bar doesn’t retain a stale timer.

---

speckofthecosmos:
Both done in 6c45c23c.

`stopTimeEntry` no longer reads the cache at all. It's `await refetchRunningTimeEntry()` unconditionally, so there's no liteMode branch left on the destructive path. Normal mode now pays one extra request per stop in the case where the cache would have hit, which is the right trade for an action that can otherwise stop the wrong timer.

The authoritative read also removes `runningTimeEntry` when Toggl reports nothing current, instead of only writing on a hit. `menuBar.tsx` and `me.ts` already pair set/remove on that key, so the API path was the one writer that didn't.

`ray build`, `tsc --noEmit` and `ray lint` are clean.

pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #31028, opened 2026-09-11
Toggl Track: clamp running-entry elapsed time to zero

speckofthecosmos:
## Description

A running time entry whose `start` is in the future produces a **negative** elapsed time, which is then formatted directly. The result is garbled output like `-23:-35:-22` in the running-entry row and menu bar title, and a running entry that *subtracts* from today's total.

This is reachable through the extension's own UI: **Change Start Time** on the running entry is an `Action.PickDate` with no upper bound, so picking a future time is a couple of keystrokes away. Clock skew can do it too.

Three unguarded sites, all clamped with `Math.max(0, …)`:

| File | Expression |
|---|---|
| `components/RunningTimeEntry.tsx` | subtitle elapsed duration |
| `menuBar.tsx` | `currentDuration` (menu bar title) |
| `hooks/useTotalDurationToday.ts` | the running entry's contribution to today's total |

This matches the guard already present in `helpers/formatSeconds.ts`, which clamps for exactly this reason — these three call sites compute their durations with `dayjs` and never reach it.

Display-only; no API or cache behaviour changes.

## Screencast

Not included — a one-line before/after: `-23:-35:-22` becomes `00:00:00` until the start time is reached.

## Checklist

### DIFF
--- a/extensions/toggl-track/CHANGELOG.md
+++ b/extensions/toggl-track/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Toggl Track Changelog
 
+## [Fix Negative Running Durations] - {PR_MERGE_DATE}
+
+- Clamped elapsed-time calculations for the running entry to zero, so a start time in the future no longer renders a garbled negative duration in the list and menu bar, or subtracts from today's total
+
 ## [Fix Duplicate Billable Checkbox] - 2026-09-11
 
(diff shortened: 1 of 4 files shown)


0xdhrv (approved):
Looks good to me, approved ✅

PR #31029, opened 2026-09-11
Toggl Track: let a new timer be back-dated

speckofthecosmos:
## Description

Adds a **Start Time** option to the new time entry form, so a timer can be started as of a few minutes ago. The common case: you start working, and remember to start tracking ten minutes in — today that time is simply lost, or you start the timer and then go edit the entry afterwards.

Options are **Now** (the default, unchanged behaviour), **5 / 10 / 15 minutes ago**, and **Custom…**, which reveals a numeric field for any other offset.

### Notes on the implementation

- The offset is applied to the `Date` **before** the duration is derived, because a running entry encodes its start as `-1 * (Unix start time)` — `start` and `duration` have to agree or Toggl shows the wrong elapsed time.
- `startTimeOffset` is optional on `CreateTimeEntryParameters`, and `if (startTimeOffset)` skips the mutation for `0`/`undefined`, so every existing caller (including "Resume Time Entry") is byte-for-byte unchanged.
- The dropdown sits immediately before the Billable checkbox at the end of the form, after the tag picker, so it does not interrupt the description → project → task flow.
- The custom field is free text, so it is validated before submission: non-integers, negatives (which would start the timer in the *future*) and values beyond one day are rejected with an inline form error rather than silently clamped. A blank field means "now".
- The offset shifts the absolute timestamp rather than using `setMinutes()`, which is local calendar arithmetic and lands on the wrong instant across a DST transition.

## Screencast

Not included — the change is a "Start Time" dropdown added to the existing new-entry form.

## Checklist


speckofthecosmos:
Both findings were valid — fixed in the latest push, which is also rebased onto `main` now that #31026 has merged (this diff is the start-time work alone).

**DST.** Confirmed, and the impact is larger than described. Reproduced in `America/New_York` across the 2026-03-08 spring-forward gap, starting from an instant whose local time is 03:05 EDT and back-dating 10 minutes:

```
setMinutes  -> Sun Mar 08 2026 03:55:00 GMT-0400 | actual delta: -50 min
setTime     -> Sun Mar 08 2026 01:55:00 GMT-0500 | actual delta:  10 min
```

So the requested 10-minute back-date lands the start **50 minutes in the future**, which then produces a negative running duration. Switched to `now.setTime(now.getTime() - startTimeOffset * 60 * 1000)` as suggested, with a comment recording why.

**Invalid custom offsets.** Also confirmed — `-5` starts the timer five minutes in the future, and `999999999999` makes `toISOString()` throw a `RangeError` that surfaces only as a generic "Failed to start time entry".

Rather than clamping silently, the custom field is now validated and shows an inline form error, so submission is blocked and the user sees why:

```
"-5"            -> rejected     "20"   -> 20
"999999999999"  -> rejected     "1440" -> 1440
"abc"           -> rejected     "1441" -> rejected
"10.5"          -> rejected     ""     -> 0 (treated as "now")
```

The bound is one day (`MAX_START_OFFSET_MINUTES = 24 * 60`); back-dating further than that is better served by editing the entry. The error clears as soon as the field is edited.

0xdhrv (approved):
Looks good to me, approved ✅

PR #31031, opened 2026-09-11
Update accordance extension

### LINKED ISSUE #30988: [Accordance]...
### Extension

https://raycast.com/entupo/accordance

### Raycast Version

2.3.0

### OS Version

macOS 26.6.2

#### Error:
```
The variable verseText is not defined.
```

### Steps To Reproduce

1. when choosing either Accordance Read Bible or Get Verses, I get this error. 

### Current Behaviour

_No response_

### Expected Behaviour

_No response_

entupo:
## Description

Fixes raycast/extensions#30988 ([Accordance] — error The variable verseText is not defined.)
What was wrong
Get Verses and Read Bible retrieve text by sending Accordance a raw Apple event («event AccdTxRf») whose result is assigned inside the script:

```applescript
tell application "Accordance"
  try
    set verseText to «event AccdTxRf» {"ESVS", "John 3:16", true}
  on error errMsg
    return "Error: " & errMsg
  end try
end tell
```

On macOS 26, if the event fails (module not found, event not handled, Accordance unavailable), the AppleScript runtime no longer surfaces that failure as an on error message — it instead throws a new error, -2753 "The variable verseText is not defined.", referring to the assignment's target variable, outside the try block. The script exits with an uncaught error, runAppleScript throws, and the extension crashed the lookup with the raw message reported in the issue. The same fragile pattern existed in module listing («event AccdVerL») and the English-check («event AccdIsEg») call sites.

## Fix
All three AppleScript snippets now wrap the whole tell block in an outer try ... on error errorMessage number errorNumber, which macOS 26 does catch. Failures return "Error: [-2753] ..." instead of throwing, and the TypeScript layer already handles that Error:-prefixed output as a graceful failure (a proper failure toast / error verse item instead of a crash).

## Checklist

### DIFF
--- a/extensions/accordance/CHANGELOG.md
+++ b/extensions/accordance/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Accordance Changelog
 
-## [Fixes] - {PR_MERGE_DATE}
+## [Fixes] - 2026-09-12
 
 - Fix "The variable is not defined" error on macOS 26 by wrapping AppleScript calls in an outer try block (#30988)
 
--- a/extensions/accordance/CHANGELOG.md
+++ b/extensions/accordance/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Accordance Changelog
 
+## [Fixes] - {PR_MERGE_DATE}
+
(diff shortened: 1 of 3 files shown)


0xdhrv (approved):
Looks good to me, approved ✅

PR #31032, opened 2026-09-11
[Herdr] Fix agent focus not switching to the correct workspace

### LINKED ISSUE #31018: [herdr] Focus Agent Needing Attention broken after herdr v0.9.0 release
### Extension

https://www.raycast.com/vlades/herdr

### Raycast Version

2.3.0.0 (0)

### OS Version

macOS Tahoe 26.6.2

### Description

After herdr v0.9.0, the "Focus Agent Needing Attention" command doesn't work correctly. It opens the correct tab in my WezTerm. However, the herdr interface inside wezterm doesn't switch to the workspace where that agent lives. If the agent needing attention is in workspace `5 · docs`, the extension just brings me into the WezTerm tab, but I stay on my current workspace (we'll say `6 · notes`) instead of moving to workspace `5 · docs`.

### Steps To Reproduce

1. Set up herdr with at least two workspaces.
2. Start an agent in one workspace (e.g. `docs`).
3. Start / focus a different workspace (e.g. `notes`).
4. Run the "Focus Agent Needing Attention" command in Raycast.
5. Observe that herdr does not focus the `docs` workspace.

### Current Behaviour

The wezterm herdr tab opens (if not already open) and comes to the front. The sidebar shows my old workspace, not the workspace where the agent is. If I try to interact with the herdr interface, I am still on the wrong workspace. This started after the herdr v0.9.0 release.

### Expected Behaviour

The "Focus Agent Needing Attention" command opens wezterm and shows the herdr interface on the correct workspace where the agent is running. The sidebar updates to show the correct workspace name and number. I should be able to immediately work with the agent without manually switching herdr workspaces.

alastairsounds:
## Description

Fixes #31018

`herdr agent focus` (used by "Focus Agent Needing Attention" and other agent-focus actions) stopped switching the visible Herdr workspace after herdr v0.9.0. It still activates the correct terminal, but the attached Herdr TUI stayed on whatever workspace was already focused.

Root cause: herdr v0.9.0's server only pushes workspace changes to attached shell clients for `workspace.focus`, `tab.focus`, and `pane.focus`. So, `agent.focus` was left out of that list, so its workspace change never reached the client. This is confirmed against a vanilla v0.9.0 Homebrew install.

Since `agent.focus` isn't fixable from this extension's side of the wire, this PR routes agent-focus actions through `agent get` (to resolve the agent's pane) followed by the existing `focusPane` helper, which already uses `tab.focus`/`pane.focus` and works correctly. No changes are needed in herdr itself.

## Verify

To verify: have agents running in two different Herdr workspaces, then run "Focus Agent Needing Attention" (or any agent-focus action) on an agent in the non-active workspace. The Herdr TUI should switch to that agent's workspace, not just bring the terminal app forward.

## Checklist

### DIFF
--- a/extensions/herdr/CHANGELOG.md
+++ b/extensions/herdr/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## [Fix Agent Focus Not Switching Workspace] - {PR_MERGE_DATE}
+
+- Fix "Focus Agent Needing Attention" and other agent-focus actions not switching the visible Herdr workspace after herdr v0.9.0. `agent focus` now resolves the agent's pane and routes through pane focus, which correctly switches workspace.
+
 ## [Fix Stuck Menu Bar Toast] - 2026-08-29
 
 - Report menu bar failures with a HUD instead of a toast held open across the action. Clicking an item unloads the menu bar command, which left the toast on screen with nothing to resolve it.
--- a/extensions/herdr/package.json
+++ b/extensions/herdr/package.json
@@ -5,6 +5,9 @@
   "description": "Control Herdr workspaces and coding agents from Raycast",
   "icon": "extension-icon.png",
   "author": "vlades",
+  "contributors": [
+    "alastairsounds"
+  ],
   "type": "module",
   "license": "MIT",
   "platforms": [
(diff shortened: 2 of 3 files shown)


alastairsounds:
Closing this. #31008 was merged in this morning and fixes this bug with an equivalent approach.

PR #31038, opened 2026-09-11
Update mozilla-firefox extension

rockschtar:
## Description

Follow-up to #30663, where @0xdhrv suggested contributing the improvements from my Windows-only Firefox extension to this one instead of maintaining a second Store listing.

- **Search Bookmarks** now reads live from `moz_bookmarks` in `places.sqlite` instead of the daily `bookmarkbackups` lz4 file, so new bookmarks show up immediately
- **Search History** hides redirect artifacts (`hidden = 0`) and entries that were never visited
- Both commands match the query against the URL as well as the title
- History timestamps use `localtime`, so entries are grouped under the correct day when the UTC and local dates differ
- Search terms containing a single quote are escaped and no longer break the SQL query
- Removed the now-unused lz4 decoder, `getBookmarksDirectoryPath` and the "no bookmarks" error view (bookmarks come from the same database as the history, so `NotInstalledError` covers the missing-profile case)
- Added unit tests for the shared `searchWhereClause` helper

Tested on Windows 11 with Firefox release. The queries are platform-independent and work unchanged against a macOS `places.sqlite`.

## Screencast

n/a - no UI changes

## Checklist

### DIFF
--- a/extensions/mozilla-firefox/CHANGELOG.md
+++ b/extensions/mozilla-firefox/CHANGELOG.md
@@ -1,12 +1,13 @@
 # Mozilla Firefox Changelog
 
-## [Live Bookmarks & Search Fixes] - 2026-09-12
+## [Live Bookmarks & Search Fixes] - {PR_MERGE_DATE}
 
 - Search Bookmarks now reads live from `moz_bookmarks` in `places.sqlite` instead of the daily `bookmarkbackups` lz4 file, so new bookmarks show up immediately
 - Search History hides redirect artifacts (`hidden = 0`) and entries that were never visited
 - Search History and Search Bookmarks match the query against the URL as well as the title
 - Fixed history entries being grouped under the wrong day when the UTC and local dates differ
-- Search terms containing a single quote no longer break the SQL query
+- Tag assignments are excluded from Search Bookmarks, so tagged URLs no longer appear once per tag
+- Search terms containing a single quote no longer break the SQL query, and `%` / `_` are matched literally instead of acting as wildcards
 - Removed the built-in lz4 decoder and the "no bookmarks" error view, which are no longer needed
 
 ## [Windows Support] - 2026-09-09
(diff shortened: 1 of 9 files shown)


0xdhrv (commented):
Hey @rockschtar 👋

Thanks for contributing the live bookmark query and history filter to Mozilla Firefox.

- `getBookmarkQuery` in `src/hooks/useBookmarkSearch.tsx` selects every `moz_bookmarks.type = 1` row. Firefox also stores tag assignments as type-1 rows that point at the same `moz_places` record, so a tagged URL can show up once per tag and use extra slots in the `LIMIT 100`. Could you exclude children of the tags folder (`guid = 'tags________'`)?

- `searchWhereClause` in `src/util/index.ts` escapes `'` but still interpolates `%` and `_` into `LIKE`. Searching for `%20` or a title with `_` can match unrelated rows. Could you treat those as literals, and add a regression test for that case?

---

rockschtar:
Thanks for the review @0xdhrv! Both points are addressed in the latest commit:

- **Tag assignments**: `getBookmarkQuery` now excludes rows whose parent is a child of the tags folder (`guid = 'tags________'`), so a tagged URL shows up exactly once.
- **LIKE wildcards**: `searchWhereClause` escapes `%`, `_` and the escape character itself and adds `ESCAPE '\'` to every `LIKE`. Added a regression test covering `%20`, an underscore and a literal backslash.

Also switched the changelog heading to the `{PR_MERGE_DATE}` placeholder as flagged by Greptile. Ready for review again.

0xdhrv (approved):
Looks good to me, approved ✅

rockschtar:
> 🎉 🎉 🎉
> 
> Such a great contribution deserves a reward, but unfortunately we couldn't find your Raycast account based on your GitHub username (@rockschtar).
> 
> Please [link your GitHub account to your Raycast account](https://www.raycast.com/settings/account) to receive your credits and soon be able to exchange them for some swag.

GitHub Account is no connected with Raycast Account

PR #31048, opened 2026-09-12
[Dokploy] Add service lifecycle actions

merttopuz:
## Description

Adds `Deploy`, `Redeploy`/`Rebuild`, `Start`, `Stop` and `Reload` actions to the **Services** screen, so a service's lifecycle can be managed without leaving Raycast:

- `application`: Deploy, Redeploy, Start, Stop, Reload
- `compose`: Deploy, Redeploy, Start, Stop
- `mariadb` / `mongo` / `mysql` / `postgres` / `redis`: Deploy, Rebuild, Start, Stop, Reload

Each action posts to the same route Dokploy's own dashboard uses for that service kind (e.g. `application.deploy`, `postgres.rebuild`). Stopping a service asks for confirmation first, since it makes the service unreachable.

This follows up on the discussion in #29418, where @xmok suggested folding a broader Dokploy extension into this one instead of publishing it separately. Rather than one large PR, this brings those features over as a few small, focused PRs - this is the first.

## Screencast

Additive actions on the existing Services list, no new screens. Happy to add a screencast if it helps review.

## Checklist


merttopuz:
Thanks for the review, both fair points - pushed a fix for each:

1. Moved the new lifecycle actions section below the existing **Create** submenu, so Enter still opens Create as before instead of triggering an unconfirmed deploy.
2. Threaded the project list's `revalidate` down through `Environments`/`Services`, so a successful lifecycle action refreshes the displayed status instead of leaving it stale.

0xdhrv (commented):
Hi @merttopuz 👋

@xmok do you want to check this?

0xdhrv (approved):
Looks good to me, approved ✅

PR #31050, opened 2026-09-12
Qobuz: convert the Now Playing track, with clipboard fallback

kud:
## Description

`Convert Track Link` used to read only the clipboard, and rendered a blank view whenever the clipboard was empty.

Now the command converts, in order: a track link typed as the optional command argument; otherwise the track the Qobuz desktop app is currently on, read via `readNowPlayingTrackId()` from `@kud/qobuz`, the same state file the existing Now Playing menu-bar command already reads. When there is no argument and Qobuz has no current track, it falls back to a Spotify / YouTube Music / Qobuz track link on the clipboard, with the existing errors for an empty clipboard or an unsupported link.

When both sources are available, a switch action in the panel moves to the other one ("Use Clipboard Link Instead" or "Use Now Playing Instead"), and the metadata view shows a "Source" row (Now Playing in Qobuz / Clipboard). An empty clipboard with no current track shows a "Nothing to convert" view that lists the three sources in order, instead of a blank view. A typed link is deliberate, so an invalid one shows its error rather than falling through.

The two conversion directions were extracted into `convertFromQobuz` / `convertToQobuz` so the now-playing path reuses them instead of duplicating the lookup logic. `looksLikeTrackLink` in `resolve.ts` is the offline shape check of `resolveLink`, used only to decide whether the clipboard is worth offering as the alternative.

In the Qobuz → other services direction, the panel now also offers the Qobuz actions (Open in Qobuz, Play Track in Qobuz, Open in Browser, Copy Qobuz Link) under a "Qobuz" section, since the input is often the now-playing track rather than a link already in hand.

## Screencast

No UI layout changed (same detail view), so no screencast: this is a logic change to an existing command. Happy to add a screenshot of the new "Source" metadata row if useful for review.

## Checklist

`ray build` and `ray lint` both pass. This has **not** been click-tested inside Raycast yet, which is why this is opening as a draft; I'll convert it to ready for review once it has been.

### DIFF
--- a/extensions/qobuz/CHANGELOG.md
+++ b/extensions/qobuz/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Qobuz Changelog
 
-## [Now Playing fallback] - {PR_MERGE_DATE}
+## [Now Playing fallback] - 2026-09-12
 
 - Convert Track Link: convert the track Qobuz is currently on, falling back to a track link on the clipboard; when both are available, an action switches to the other one
 - Convert Track Link: accept an optional track link as a command argument, which takes precedence over both
--- a/extensions/qobuz/CHANGELOG.md
+++ b/extensions/qobuz/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## [Now Playing fallback] - {PR_MERGE_DATE}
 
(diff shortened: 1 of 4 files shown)


0xdhrv (approved):
Looks good to me, approved ✅

PR #31053, opened 2026-09-12
Orion: improve Command Bar Top Hit selection

szpony:
## Summary
- rank local candidates by match tier, then open tab, bookmark, reading list, and history
- collapse identical destinations and use history frecency only within history ties
- focus a newly resolved Top Hit without controlling later native list navigation
- recognize typed web addresses and add explicit default-browser actions

## Validation
- `npm ci`
- `npm run lint`
- `npm run build`
- `npm run dev`
- reviewed the staged diff for credentials, local Orion data, personal paths, and email addresses

## Privacy
This change reads the existing local Orion data sources only at runtime. It does not add telemetry, upload local data, or include any user data, database files, credentials, or screenshots.

### DIFF
--- a/extensions/orion/CHANGELOG.md
+++ b/extensions/orion/CHANGELOG.md
@@ -5,6 +5,7 @@
 - Improve Top Hit ranking with deterministic match tiers and source precedence: open tabs, bookmarks, reading list, then history. Deduplicate matching destinations and use history frecency only to resolve ties within history.
 - Keep Top Hit selected as local sources resolve, while preserving an explicit Ctrl+N/Ctrl+P selection when later results arrive.
 - Reset the selection session whenever the query or profile changes, so deleting or appending text cannot leave Web Search focused after a prior candidate disappears.
+- After the first Ctrl+N/Ctrl+P navigation, release selection to Raycast's native list so repeated navigation remains smooth while later results do not steal focus.
 - Recognize typed web addresses and offer opening them in the system default browser. Add the same default-browser action to local Orion results.
 
 ## [Command Bar] - 2026-06-22
--- a/extensions/orion/CHANGELOG.md
+++ b/extensions/orion/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## [Command Bar] - {PR_MERGE_DATE}
 
 - Improve Top Hit ranking with deterministic match tiers and source precedence: open tabs, bookmarks, reading list, then history. Deduplicate matching destinations and use history frecency only to resolve ties within history.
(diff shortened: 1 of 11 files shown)


pernielsentikaer (changes requested):
Hi @szpony 👋

Thanks for improving the Command Bar ranking and selection behavior 💪

I found three things while reviewing:

- Please apply each section’s limit before adding URLs to the shared deduplication set. Currently, results hidden by the limit can suppress matching results from later visible sections.
- Please preserve scheme and path distinctions when deduplicating URLs. `http://` and `https://`, or `/path` and `/path/`, can lead to different destinations.
- In `BookmarkListItem` and `HistoryListItem`, please preserve the existing default-browser action as the primary action and append the new Orion action afterward. Otherwise, pressing Return changes behavior for the standalone commands.

---

pernielsentikaer (changes requested):
Hi @szpony 👋

Thanks for the follow-up — the limit ordering, the scheme/slash-preserving dedup key, and keeping "Open in Default Browser" primary in the Bookmarks/History commands all look right now 💪

One thing I found while checking the ranking path. `LAST_VISIT_TIME` comes back from Orion's `history_items` as a numeric Core Data/WebKit epoch (seconds since 2001), but it's typed as a string and passed straight to `new Date(...)`. That has two effects in `command-bar.tsx`:

- `historyFrecency` evaluates `new Date(790000000.5)`, which is 1970, so `recencyBoost` is always `0` and history recency never breaks a tie.
- `compareRankedHits` then calls `bLastVisit.localeCompare(aLastVisit)`. The value is a number at runtime, so this throws `TypeError: bLastVisit.localeCompare is not a function`. It's reached whenever two history results share a tier and visit count (two once-visited pages is enough), which crashes the Command Bar.

Could you convert the column when selecting, like the other history extensions do?

```sql
datetime(LAST_VISIT_TIME + 978307200, 'unixepoch', 'localtime') as lastVisitTime,
DATE(LAST_VISIT_TIME + 978307200, 'unixepoch', 'localtime') as lastVisitDate
```

That keeps `lastVisitTime` a string, so both the `Date` parse and the tie-break compare work (it also fixes the standalone History command grouping everything under 1970).

Two smaller ones:

- `canonicalUrl` only trims now, so `https://Example.com/path` and `https://example.com/path` can both show. Normalizing just the host case would close that.
- `isWebAddress` accepts any `x.yy` as a TLD, so `index.html`, `main.js`, or `file.txt` get offered as web addresses. A Public Suffix List check or a short TLD allowlist would avoid opening those in the browser.

---

pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #31055, opened 2026-09-12
Orion: add configurable suggestions and tab fallback

szpony:
## Summary
- add a Command Bar preference for 0, 3, 6, 9, or 12 live suggestions
- disable suggestion network requests when the limit is zero
- add Fuse and pinyin-based open-tab fallback only when exact local tab matching has no result
- label fallback matches and keep them out of Top Hit

## Validation
- `npm ci`
- `npm run lint`
- `npm run build`
- `npm run dev`
- local Fuse/pinyin dependency check using `网易首页` and `wangyi`
- reviewed the fully staged diff, including the new source file, for credentials, local Orion data, personal paths, and email addresses

## Privacy
No user data, database files, credentials, or screenshots are included. When enabled, live suggestions continue to send only the text typed into Command Bar to the user-selected search engine; selecting a limit of zero prevents those requests.

### DIFF
--- a/extensions/orion/CHANGELOG.md
+++ b/extensions/orion/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Orion Changelog
 
+## [Command Bar] - {PR_MERGE_DATE}
+
+- Add a configurable limit for live search suggestions; setting it to zero hides suggestions and prevents suggestion requests.
+- Add fuzzy and pinyin matching as a labeled fallback for open tabs when there is no exact local tab match. Fallback matches are not eligible for Top Hit.
+
 ## [Command Bar] - 2026-06-22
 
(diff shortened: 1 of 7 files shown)


0xdhrv (commented):
Hey @szpony 👋

Thanks for the Command Bar suggestion limit and tab fallback.

Search Tabs used to fold diacritics in `search()` (`normalize("NFD")` and strip combining marks). `tabSearch.ts` only trims and lowercases, so a query like `creme brulee` can miss a tab titled `Crème brûlée`, and Fuse at `0.3` is not a guaranteed fallback.

Could you reuse that folding for the exact match and as a Fuse field, so Search Tabs keeps the old matching and Command Bar fallback can use it too?

---

pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #31068, opened 2026-09-13
[Dokploy] Add View Logs action to Services

merttopuz:
## Description

Adds a `View Logs` action to the **Services** screen, showing the last 200 lines of a service's logs (via its `readLogs` route) in a Detail view, with `Refresh` and `Copy Logs` actions.

Uses the same per-kind id field lookup already used for the lifecycle actions added in #31048. Compose stacks can scope logs to one container on Dokploy's side, but this first version calls the same route uniformly across all kinds and leaves per-container scoping for a follow-up if it turns out to be needed.

This is another small, focused piece following up on #29418.

## Screencast

Adds one new action and screen off the existing Services list. Happy to add a screencast if it helps review.

## Checklist

### DIFF
--- a/extensions/dokploy/CHANGELOG.md
+++ b/extensions/dokploy/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## [View Service Logs] - {PR_MERGE_DATE}
 
-- Add a `View Logs` action to the **Services** screen, showing the last 200 lines of a service's logs with `Refresh` and `Copy Logs` actions.
+- Add a `View Logs` action to the **Services** screen, showing the last 200 lines of a service's logs with `Refresh` and `Copy Logs` actions. Not yet available for Compose stacks, which need a container to be picked first.
 
 ## [Service Lifecycle Actions] - 2026-09-13
 
--- a/extensions/dokploy/CHANGELOG.md
+++ b/extensions/dokploy/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Dokploy Changelog
 
+## [View Service Logs] - {PR_MERGE_DATE}
+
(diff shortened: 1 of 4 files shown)


merttopuz:
Thanks, all three were valid - pushed a fix for each:

1. Hid `View Logs` for compose services, since `compose.readLogs` requires a `containerId` this screen doesn't collect yet. Left it as a follow-up rather than adding a container picker to this PR.
2. `parseResponse` now rejects non-ok responses instead of returning the error body as if it were log text.
3. Escaped triple backticks in the log content so they can't close the markdown fence early.

pernielsentikaer (changes requested):
Hi @merttopuz 👋

Thanks for adding service logs 💪

I found this while reviewing:

`readLogs` isn't exposed through Dokploy's OpenAPI handler. It's a [native tRPC query](https://github.com/Dokploy/dokploy/blob/main/apps/dokploy/server/api/routers/application.ts#L1227-L1260), so `/api/application.readLogs?...` won't work and the logs view won't load. Could you call `/api/trpc/<type>.readLogs` using its superjson input format and unwrap the tRPC response?

---

merttopuz:
Good catch, thanks for pointing to the router source - confirmed `readLogs` isn't wired into the OpenAPI bridge on any of the service routers, so the REST-style call I had never reaches it.

Pushed a fix: it now calls the raw tRPC endpoint directly (`/api/trpc/<type>.readLogs`) with a superjson-encoded `input` and unwraps the response from `{ result: { data: { json } } }`. Auth is unchanged - the same `x-api-key` header resolves a session on this endpoint the same way it does on the REST-bridged ones.

Testing this against a live instance now, will mark ready for review once confirmed.

merttopuz:
Confirmed working against a live instance - logs load correctly now. Marking ready for review.

pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #31072, opened 2026-09-13
Update series-rating-graphs extension

muhammadrizo-y:
## Description

- Add new detail view using markdown table
- Add List item detail view

## Screencast

N/A

## Checklist


muhammadrizo-y:
@greptileai Downgraded to @raycast/api: ^2.4.0, which will work because Raycast v2.4.0.0 is released.

pernielsentikaer (changes requested):
Hi @muhammadrizo-y 👋

Thanks for the detail-view and list-detail work, and for the dependency refresh 💪

I found a couple of things while reviewing:

**Failed poster actions now report success.** In `src/utils/image.ts`, `downloadImage` and `copyImage` catch their errors and call `showFailureToast`, but they no longer rethrow or return a failure result. `DownloadPosterAction` and `CopyPosterAction` then continue into their success branch, so a failed download/copy still shows “Poster downloaded…” / “Poster copied to clipboard” (and the download toast offers “Show in Finder” for an empty path). Could you keep the error propagating to the caller, or return an explicit failure result the callers check, so a failure isn’t reported as success?

**`@raycast/api` still resolves to 2.4.1.** `package.json` says `^2.4.0`, but `package-lock.json` still pins `@raycast/api` to `2.4.1`, which Greptile flagged as newer than the released Raycast app (it made another extension uninstallable). The caret range also allows 2.4.1 on a fresh install. Could you pin to the exact released version and regenerate the lockfile, and confirm which Raycast app version this targets?

**Layout preference is overridden by the cached state.** `useCachedState("layout", defaultLayout)` in `src/search-shows.tsx` only uses the preference as the initial value, so after a toggle the “Layout” preference no longer changes anything. Could you sync the cached value when the preference changes, or key the cache to the preference?

Smaller note: the Grid item-size dropdown (Large/Medium/Small) was removed — was that intentional?

---

muhammadrizo-y:
Hi @pernielsentikaer

I've addressed the issues. About the Grid dropdown, I initially removed it because there seems to be a bug? Or maybe I'm doing something wrong, but I've seen this in other extensions that have the Toggle Layout action as well: the dropdown resets to the Large option every time you switch from List view to Grid view using the toggle. I've re-added it now, but could you check if this happens for you too?

Also, the Metadata Image Enforcer check shows two invalid errors.

pernielsentikaer (changes requested):
Hi @muhammadrizo-y 👋

Thanks for the detail-view and list-detail work, and for reworking the earlier feedback 💪

I found two things while reviewing:

**The `metadata-images` check is failing.** `series-rating-graphs-1.png` and `series-rating-graphs-4.png` show the local extension icon in the footer, so they look like they were captured from a local dev build. Could you recapture both with Raycast's **Window Capture** tool and **Save to Metadata**, from the installed extension? Details: https://developers.raycast.com/basics/prepare-an-extension-for-store#screenshots

**The API downgrade isn't reflected in the lockfile.** `package.json` says `@raycast/api: ^2.3.1`, but `package-lock.json` still resolves `@raycast/api` to `2.4.1`, so a fresh install still pulls the version that was flagged as newer than the released app supports. The caret range also allows 2.4.x. Could you pin the exact released version, regenerate the lockfile, and confirm which Raycast app version this targets?

Smaller note: in `src/search-shows.tsx` the list detail uses `show.rating?.aggregateRating.toFixed(1)`, while the grid accessory guards with `?.toFixed(1)`. If a result ever comes back with a `rating` object missing `aggregateRating`, the list detail would throw — making them consistent would be a nice bit of hardening.

---

muhammadrizo-y:
@pernielsentikaer 

@raycast/api is now pinned to 2.3.1 and the list detail now uses `show.rating?.aggregateRating?.toFixed(1)`.
But the `metadata-images` check failure is a false positive, there aren't any development icons in screenshot 1 and 4.

Also, could you help me with this?
> About the Grid dropdown, I initially removed it because there seems to be a bug? Or maybe I'm doing something wrong, but I've seen this in other extensions that have the Toggle Layout action as well: the dropdown resets to the Large option every time you switch from List view to Grid view using the toggle. I've re-added it now, but could you check if this happens for you too?

pernielsentikaer (changes requested):
Hi @muhammadrizo-y 👋

Thanks for the detail-view and list-detail work, and for reworking the earlier feedback 💪

I found two things while reviewing:

**The `metadata-images` check is failing.** `series-rating-graphs-1.png` and `series-rating-graphs-4.png` show the local extension icon in the footer, so they look like they were captured from a local dev build. Could you recapture both with Raycast's **Window Capture** tool and **Save to Metadata**, from the installed extension? Details: https://developers.raycast.com/basics/prepare-an-extension-for-store#screenshots

**The API downgrade isn't reflected in the lockfile.** `package.json` says `@raycast/api: ^2.3.1`, but `package-lock.json` still resolves `@raycast/api` to `2.4.1`, so a fresh install still pulls the version that was flagged as newer than the released app supports. The caret range also allows 2.4.x. Could you pin the exact released version, regenerate the lockfile, and confirm which Raycast app version this targets?

Smaller note: in `src/search-shows.tsx` the list detail uses `show.rating?.aggregateRating.toFixed(1)`, while the grid accessory guards with `?.toFixed(1)`. If a result ever comes back with a `rating` object missing `aggregateRating`, the list detail would throw — making them consistent would be a nice bit of hardening.

**Grid item-size dropdown resets to Large.** You asked whether we see the dropdown reset too. I couldn't reproduce it in the app from here, but the code explains the mismatch: the `Grid.Dropdown` in `src/search-shows.tsx` is uncontrolled (no `value`/`defaultValue`), and `columns` starts at `5` while the first item is `Large` (`3`), so on first open — and after a List→Grid toggle where the dropdown remounts — the label can disagree with the actual column count. Making it controlled and persisting the value fixes it, e.g. `const [columns, setColumns] = useCachedState("grid-columns", 5)` and `value={String(columns)}` on the dropdown (that's what `csfd` and `chiikawa-character` do). That also sidesteps the `storeValue` remount behavior.

---

muhammadrizo-y:
@pernielsentikaer 

`metadata-images` check, the API downgrade and `?.toFixed(1)` issues are repeated here, they are already fixed.

About the Grid dropdown, I tried what you suggested but it's still the same thing. Can you please check your Slack DM?

pernielsentikaer (changes requested):
Hi @muhammadrizo-y 👋

Thanks for working through the earlier rounds — the exact `@raycast/api: 2.3.1` pin with a matching lockfile, the layout-preference sync, the `?.toFixed(1)` guard, and the controlled grid dropdown all look right now 💪

I found this while reviewing:

The `metadata-images` check is still failing on `series-rating-graphs-1.png` and `series-rating-graphs-4.png`. The checker flags a green square in the bottom band of the window. Digging into the pixels, screenshot 4's flagged area is two 28×28 green squares that look like the 🟩 rating emoji from the episode table, and the flagged spot in screenshot 1 is wider in-app green content rather than a dev-build badge — so it may well be a false positive on our side. That said, it's a required check and it's still red at the current head, so it needs to be sorted before we can merge. Could you recapture those two with **Window Capture** → **Save to Metadata** so the rating table sits clear of the bottom band, or let us know if you're confident the detector is wrong so we can take a look at the checker instead? I couldn't visually confirm a real local extension badge here.

Smaller note: in the list detail, `${show.rating?.voteCount.toLocaleString() || "N/A"}` still assumes `voteCount` exists whenever `rating` does — worth guarding for consistency with the `aggregateRating` fix.

---

muhammadrizo-y:
@pernielsentikaer

Fixed: `${show.rating?.voteCount.toLocaleString() || "N/A"}` -> `${show.rating?.voteCount?.toLocaleString() || "N/A"}`

I'm confident the `metadata-images` check is wrong; there is no development icon anywhere in screenshots 1 and 4. This requires manual review (not an automated AI response).

And the `Grid` dropdown issue is still there; it's probably a bug.

pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #31075, opened 2026-09-13
Google Chrome: sort bookmark results by frecency

### LINKED ISSUE #31074: [Google Chrome] Sort Search Bookmarks results by frecency
### Extension

### OS Version

macOS 26.6.2

### Description

`Search Bookmarks` never reorders its results. `useBookmarkSearch` only filters the list, so the order is whatever `getBookmarks` returns from Chrome's bookmarks file. A bookmark I open through Raycast every day ranks the same as one I opened once, and when several bookmarks match a query the order looks arbitrary to the user.

Proposal: sort the matches with `useFrecencySorting` from `@raycast/utils`, call `visitItem` when a bookmark is opened, and expose a `Reset Ranking` action so a mis-hit is undoable.

Two details that matter:

- The list item component is shared with `Search History` (`ChromeListItems.TabHistory`, distinguished by its `type` prop), so ranking has to be wired at the bookmark call sites only. `Search History` keeps its `ORDER BY last_visited DESC` and must not be reordered.
- `useFrecencySorting` preserves the input order for items that were never visited, so the list looks exactly as it does today until you actually open something.

Related: #8665 (Arc, closed as stale rather than declined), #18214 (Safari), #26013 (Google Chrome, ordering inside `Search All`).

Anyone who uses `Search Bookmarks` as their main way to open sites. The more bookmarks you keep, the more the current order costs: matches for a short query are effectively unranked, so you type more characters than the bookmark needs.

### Anything else?

_No response_

kirill-markin:
## Description

`Search Bookmarks` never reordered its results. `useBookmarkSearch` only filtered the list, so the order was whatever `getBookmarks` returns from Chrome's bookmarks file: a depth-first walk of the folder tree. A bookmark you open through Raycast every day ranked the same as one you opened once.

The filtered list is now sorted with `useFrecencySorting`, and opening a bookmark bumps its ranking. This covers `Search Bookmarks` and the Bookmarks section of `Search All`.

Closes #31074

## Details

- The frecency key is `bookmark.url`, so ranking follows the page rather than Chrome's per-profile bookmark id. Two bookmarks pointing at the same URL share one entry, which is the intent.
- Bookmarks you have never opened through Raycast keep exactly the order they have today. `useFrecencySorting` sorts its input in place, so the hook is handed a copy and `sortUnvisited` falls back to an index map captured from the pristine `getBookmarks` order. That map is keyed by `bookmark.id` rather than by URL, because a `Map` would otherwise collapse two same-URL bookmarks onto one index and reshuffle entries nobody ever opened.
- Every action that opens the bookmark records the visit: `Open`, `Open in Current Profile`, `Open in Original Profile` and `Open in Guest Window`. Copying a URL does not. The visit is recorded when the action is triggered, as in the hook's own example and in the other extensions here that use it — and it has to be, because `openNewTab` awaits `popToRoot`, which unloads the command before anything after it can run.
- A `Reset Ranking` action (`ctrl` `shift` `R`) clears the stored ranking for a bookmark and restores it to its original position immediately, without waiting for a refetch.

## Scope

- `Search History` is untouched. `HistoryItem` and `HistoryItemActions` are shared between history and bookmarks, so both new props are optional and every history call site passes neither: no reordering, no extra action, no shortcut change. `useHistorySearch` keeps its `ORDER BY last_visited DESC`.
- The `search-bookmarks` AI tool is deliberately left unranked. It answers one-shot queries rather than presenting a list to pick from, so frecency has nothing to disambiguate there.
- No new preference. The behaviour is invisible until you open a bookmark, and `Reset Ranking` covers a mis-hit. Happy to add a toggle if you would rather it were opt-in.

## Screencast

The same `gmail` query in `Search Bookmarks`, before and after opening the three inbox bookmarks a few times. The three that were opened move to the top; the three that were never opened keep their order relative to each other.

**Before**

Screenshot 2026-09-13 at 21 29 14

**After**

Screenshot 2026-09-13 at 21 29 32

## Checklist

### DIFF
--- a/extensions/google-chrome/CHANGELOG.md
+++ b/extensions/google-chrome/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Google Chrome Changelog
 
+## [Sort Bookmarks by Frecency] - {PR_MERGE_DATE}
+
+- Rank bookmarks opened from Raycast above rarely used ones in Search Bookmarks and Search All.
+- Add a Reset Ranking action to restore a bookmark to its original position.
+
 ## [Add Search Windows Command] - 2026-08-16
 
 - Add a new Search Windows command to search and select open Google Chrome windows.
--- a/extensions/google-chrome/package.json
+++ b/extensions/google-chrome/package.json
@@ -25,7 +25,8 @@
     "xmok",
     "hayk_safaryan",
     "alexi.build",
-    "oussama_moussaoui"
+    "oussama_moussaoui",
+    "kirill-markin"
   ],
   "pastContributors": [
     "jonasengelmann"
--- a/extensions/google-chrome/src/components/ChromeActions.tsx
+++ b/extensions/google-chrome/src/components/ChromeActions.tsx
@@ -85,52 +85,58 @@ function HistoryItemActions({
   title,
   url,
   profile: profileOriginal,
+  onVisit,
+  onResetRanking,
 }: {
   title: string;
   url: string;
   profile: string;
+  onVisit?: () => void | Promise<void>;
+  onResetRanking?: () => void | Promise<void>;
 }): ReactElement {
   const { openTabInProfile } = getPreferenceValues<Preferences>();
   const [profileCurrent] = useCachedState(CHROME_PROFILE_KEY, DEFAULT_CHROME_PROFILE_ID);
 
+  async function openAndRecordVisit(openBehaviour: SettingsProfileOpenBehaviour): Promise<void> {
+    await onVisit?.();
+    await openNewTab({ url, profileOriginal, profileCurrent, openTabInProfile: openBehaviour });
+  }
+
   return (
     <ActionPanel title={title}>
-      <Action onAction={() => openNewTab({ url, profileOriginal, profileCurrent, openTabInProfile })} title={"Open"} />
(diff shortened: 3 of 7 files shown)


kirill-markin:
Added the before/after screenshots to the description right after opening this PR, in case anything here only read the body as it was created.

They show the same `gmail` query in `Search Bookmarks`: the three bookmarks that were opened through Raycast move to the top, and the three that were never opened keep their order relative to each other.

pernielsentikaer (changes requested):
Hi @kirill-markin 👋

Thanks for adding bookmark frecency sorting 💪

The latest change starts opening the bookmark without awaiting `onVisit`, but `void onVisit?.()` still leaves a rejected storage promise unhandled. That rejection can terminate the command before the asynchronous open finishes.

Could you explicitly catch and ignore ranking failures while continuing the open operation?

---

kirill-markin:
Thanks @pernielsentikaer — fixed in 688aba26. Both bookmark-opening paths now attach `.catch` to the visit and log the failure with `console.error`, like `tools/search-history.ts` already does for non-fatal errors, so a ranking-storage failure can neither block the open nor become an unhandled rejection.

pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #31087, opened 2026-09-14
[Git] Fix PAGER/EDITOR "not permitted" errors and rebase actions

### LINKED ISSUE #31066: [Git] Adding an existing repository throws an error on submission
### Extension

https://www.raycast.com/ernest0n/git

### Raycast Version

2.3.1.0

### OS Version

macOS Tahoe 26.6.2 arm64

### Description

Error: Use of "PAGER" is not permitted without enabling allowUnsafePager

Object.action:block-unsafe-operations-plugin.ts:15:21

---
12:       action(args, { env }) {
13:          for (const vulnerability of vulnerabilityCheck(args, env)) {
14:             if (options[vulnerability.category] !== true) {
15:                throw new GitPluginError(undefined, 'unsafe', vulnerability.message);
                        ^
16:             }
17:          }
18: 
19:          return args;

---

JavaScript Stack
Object.action:block-unsafe-operations-plugin.ts:15:21
db.exec:manage-repositories.js:28:10800
Ca.attemptRemoteTask:manage-repositories.js:27:6368
Ca.attemptTask:manage-repositories.js:27:6026

### Steps To Reproduce

1. Run "Manage Git Repositories"
2. Hit enter to add repository
3. Click or arrow key down to "Add Existing Directory"
4. Click "Add Folder"
5. Choose a folder and click "Open"
6. Command + Enter to add repository

### Current Behaviour

_No response_

### Expected Behaviour

_No response_

anandghegde:
## Description

Fixes #31066

The lockfile update in #29840 moved `simple-git` from 3.33.0 to 3.36.0. Since 3.36 (`@simple-git/argv-parser` 1.1.0), simple-git throws before running any command if `PAGER`, `GIT_PAGER`, `EDITOR`, `GIT_EDITOR` or `GIT_SEQUENCE_EDITOR` appear in the environment given to it, unless `allowUnsafePager` / `allowUnsafeEditor` is set. `GitManager` passes the user's whole login-shell environment (`shellEnvironmentVariables`) through `.env(...)`, so anyone who exports a pager or editor gets this error on every Git command, starting with adding a repository.

The check looks only at whether the variable is present, not its value, so setting `PAGER=cat` doesn't help. The variables have to be removed. Git never runs in a terminal inside the extension, so they aren't needed. This PR drops them in `environment-utils.ts` (in both the zsh path and the `process.env` fallback) rather than turning on the unsafe options. The other variables simple-git blocks (`GIT_SSH_COMMAND`, `GIT_ASKPASS`, `GIT_CONFIG_GLOBAL`, ...) are left alone, because removing them would change how Git behaves for people who rely on them.

The same checks also broke both rebase actions:

- "Continue Rebase": `continueRebase()` called `this.git.env("GIT_EDITOR", "true")`, and simple-git rejects that without `allowUnsafeEditor`. `.env(name, value)` also writes into the instance's env object, which is the shared `shellEnvironmentVariables`. So after one failed attempt, `GIT_EDITOR` stayed in it and every later Git command failed until the extension reloaded.
- Interactive rebase (reword/squash/fixup/drop/edit): `interactiveRebase()` passes `-c sequence.editor=<script>`, which is rejected with `Configuring sequence.editor is not permitted without enabling allowUnsafeEditor`.

Both now run on a separate simple-git instance, created by `createGitWithoutEditor()`. It has `allowUnsafeEditor` and its own copy of the environment with `GIT_EDITOR=true` added. The unsafe option applies only to these two calls, and the shared env object is never changed. Squash needs `GIT_EDITOR=true` too: without it, and with no terminal, git stops with "Please supply the message using either -m or -F option". The shared simple-git setup (binary, error toast and the `[GIT]` debug logging) moved into a small `createSimpleGit` helper, so commands on these instances show up in the log like all others.

### How I tested

- **Pager/editor:** Bundled `src/utils/environment-utils.ts` with esbuild (with `@raycast/utils` stubbed) and ran `simpleGit(repo).env(shellEnvironmentVariables)`, the same setup `GitManager` uses, followed by `status()` and `log()` on a local repo. I did this once with `PAGER=less`, once with `GIT_PAGER=less` and once with `EDITOR=vim` exported.
  - Before: every call failed with `Use of "PAGER" is not permitted without enabling allowUnsafePager` (and the matching `GIT_PAGER` / `EDITOR` errors).
  - After: every call succeeded.
- **Continue Rebase:** Bundled `GitManager` the same way (with `@raycast/api` stubbed). In a temp repo I started a real rebase, let it stop on a conflict, then resolved and staged the file. After that I called `new GitManager(repo).continueRebase()` and ran another command.
  - Before: `continueRebase` failed with `Use of "GIT_EDITOR" is not permitted without enabling allowUnsafeEditor`, `GIT_EDITOR` was left in the shared env, and the next `status()` failed with the same error.
  - After: the rebase completed (no `.git/rebase-merge` left), the shared env was unchanged, and the next `status()` succeeded.
- **Interactive rebase:** Same harness. In a temp repo with commits one..four, I called `interactiveRebase(two, [pick two, reword three -> "three renamed", squash four])`.
  - Before: it failed with `Configuring sequence.editor is not permitted without enabling allowUnsafeEditor`.
  - After: history became `three renamed, two, one`, with all four files in HEAD. The shared env was unchanged, the next `status()` succeeded, and the `[GIT]` log line for the rebase command was printed.
- `npm run build`, `npm run lint` and `npx tsc --noEmit` pass.

## Screencast

Before
After

## Checklist


pernielsentikaer (approved):
Looks good to me, approved 🔥

pernielsentikaer:
Hi @anandghegde 👋

Thanks for fixing the pager/editor and rebase failures 💪

While testing the restored interactive-rebase flow, I found a related existing issue: reword messages are interpolated directly into an `exec git commit --amend -m "..."` line in the rebase todo. Shell expressions such as `$(...)` and backticks are therefore executed instead of being preserved literally. I reproduced this with a harmless `$(touch …)` message, which created the file and changed the resulting commit subject.

Could you store each message in a temporary file and use `git commit --amend -F <shell-quoted-path>` instead, ensuring no user-controlled message content is inserted into the rebase todo syntax?

anandghegde:
Thanks for catching that @pernielsentikaer! Fixed in #31091: each reword message now goes to a file and is applied with `git commit --amend -F '<quoted path>'`, so no message text is written into the rebase todo. The file is kept in `.git/rebase-merge`, so it's still there if the rebase stops at an `edit` step or a conflict before the reword runs, and Git deletes it when the rebase ends. I checked it with a `$(touch …)` message and a multi-line message: no file was created and the commit message came through unchanged.

PR #31096, opened 2026-09-14
Update seo-lighthouse extension

guelug:
## Description

Adds a Speedtest-style SVG scorecard to SEO Lighthouse: circular category gauges and Core Web Vitals bars rendered as a markdown data-URI image in the report `Detail` view, plus copy/save PNG actions on macOS.

Also drops unused `lighthouse` / `chrome-launcher` dependencies (the extension already uses the Lighthouse CLI), switches invocation to `execFile`, removes custom i18n (Store is English-only), and exposes an `audit-page` AI tool.

## Screencast

The report view now shows live SVG gauges instead of a static markdown table. Existing metadata screenshots are included in `extensions/seo-lighthouse/metadata/`.

## Checklist


pernielsentikaer (changes requested):
Hi @guelug 👋

Thanks for expanding the Lighthouse report and adding the AI audit tool 💪

I found a few things that need addressing:

- Please remove Chrome's `--no-sandbox` flag. Both the form and AI tool can audit arbitrary URLs, so Chrome's renderer sandbox needs to remain enabled.
- Rework Re-analyze so it starts only one forced audit and resets its progress. It currently changes the `usePromise` arguments and calls `revalidate()`, which can start two audits, while retained data makes the loading view show 100% before the new audit finishes.
- Require at least one selected category or restore the all-category fallback. Clearing all four checkboxes currently opens a terminal audit error.
- Keep the existing report actions first and restore "Open With"; append the new actions afterward so the established workflow and default action are preserved.
- Regenerate `package-lock.json` from the committed manifest. `npm ci` currently fails because `@types/node` and the root dependency metadata are out of sync.
- Correct the Chrome-missing guidance: the Lighthouse Path preference expects the Lighthouse executable, not a Chromium browser binary.
- Use the standard Copy shortcut if the README should advertise ⌘⇧C. `Keyboard.Shortcut.Common.CopyName` maps to ⌘⇧. on macOS.
- Please recapture the report metadata screenshots from the current English UI using [Raycast Window Capture with Save to Metadata](https://developers.raycast.com/basics/prepare-an-extension-for-store#screenshots). The current report captures still show the removed Spanish UI rather than this PR's scorecard dashboard.

Please also address the still-valid Greptile findings for the [Chrome sandbox](https://github.com/raycast/extensions/pull/31096#discussion_r4005049527), [duplicate re-analysis](https://github.com/raycast/extensions/pull/31096#discussion_r4005049534), [empty categories](https://github.com/raycast/extensions/pull/31096#discussion_r4005049546), [action order](https://github.com/raycast/extensions/pull/31096#discussion_r4005049553), and [lockfile](https://github.com/raycast/extensions/pull/31096#discussion_r4005049563).

---

I'll convert this PR into a draft after submitting this review. Please press Ready for review when it's ready and we'll have a look 😊

guelug:
Thanks for the detailed review 🙏 All points are addressed:

- Removed Chrome's `--no-sandbox`; the renderer sandbox stays enabled for both the form and the AI tool.
- Re-analyze now drives a single forced audit through a generation counter passed to `usePromise` (no `revalidate()` call) and resets the progress ring first.
- Submitting with no category selected is blocked with a toast instead of opening a terminal audit error.
- Restored the original report actions first, including **Open With**; the new actions are appended afterwards.
- Regenerated `package-lock.json` from the committed manifest; `npm ci` now succeeds.
- Corrected the Chrome-missing guidance: Lighthouse Path expects the Lighthouse CLI, not a browser binary.
- Switched the scorecard action to `Keyboard.Shortcut.Common.Copy` so it matches the ⌘⇧C documented in the README.
- Recaptured the metadata screenshots from the current English UI (form, audit progress, scorecard dashboard) and updated the `screenshots` list.

Also moved the DNS preflight after the cache lookup, so a valid unexpired cached report still opens when DNS is unavailable.

Verified with `npm ci`, `npm run lint`, `npm run build` and `tsc --noEmit`.

pernielsentikaer (changes requested):
Hi @guelug 👋

Thanks for working through the whole list 💪 The sandbox flag, single re-analysis, category guard, action order, lockfile, Chrome guidance, copy shortcut and the recaptured screenshots all check out on my side. I also went through Greptile's findings again — the only one left doesn't reproduce for me, so no need to chase it (details below).

Two things I found on this pass:

**Saved reports now expire with the cache.** The file returned as `reportPath` in `utils/lighthouse.ts` is the cache entry itself — `lighthouse-cache-<hash>.json` inside the user's Output Path — and `cleanExpiredCache()` deletes anything matching `lighthouse-cache-*.json` after 24 hours. The form says "JSON reports are saved to this folder" and the README says the same, so someone keeping yesterday's report will find it gone, and the name is a hash rather than something recognisable. Could you keep a readable report in the Output Path and move the cache to `environment.supportPath`, or at least make the 24-hour lifetime and the naming explicit in the form description and README?

**The "Choose Output Directory" action disappeared from the form.** Before this PR the form had that action (the `osascript` folder picker) next to the path field; now the field is the only way to set it. Was dropping it intentional? If not, restoring it — or using `Form.FilePicker` with `canChooseDirectories` — would keep the old workflow.

On Greptile's last point about the screenshots: I couldn't reproduce it. The bottom-bar icon in all three images is your own `assets/extension-icon-dark.png`, and the repository's metadata-image check passes both locally and in CI on this commit, so I'm treating that finding as resolved.

Small optional thing: the `audit-page` tool description still lists Spanish triggers (`analizar`, `auditar`, `revisar`, `ver`) now that the custom i18n is gone. `ver` is generic enough that it may pull unrelated Spanish requests into the tool, so it may be worth trimming.

Happy to take another look once those two are sorted 🙂

---

guelug:
Thanks for the thorough second pass 🙏 Both points addressed:

- **Reports no longer expire with the cache.** The 24h cache now lives under `environment.supportPath`, separate from the user's Output Path. Each audit writes a readable, non-expiring report to the Output Path as `lighthouse-<host>-<timestamp>.json`; a cache hit reuses the cached JSON but still writes/keeps that readable file if it's missing. Updated the form description, README, and preference copy to state this explicitly.
- **Restored Choose Output Directory** on the form (AppleScript folder picker via `useForm`'s `setValue`), next to the path field as before.

Also trimmed the `audit-page` tool description to drop the leftover Spanish trigger words now that the custom i18n is gone.

Verified with `npm ci`, `npm run lint`, `npm run build`, `tsc --noEmit`, and a real example.com run confirming: readable report lands in Output Path, cache is isolated in the support folder, and a previous report survives a cache-hit re-run.

Ready for another look 🙂

pernielsentikaer (approved):
Looks good to me, approved 🔥

Issue #15990, opened 2024-12-21
[Connect to VPN] Create defunct processes every 30 seconds


arthur-fontaine:
### Extension

https://www.raycast.com/rasmusbe/connect-to-vpn

### Raycast Version

1.88.4

### macOS Version

15.1.1

### Description

For a month, the Connect to VPN make my Macbook Pro M1 Pro crashes every few days. In the Terminal it shows `fork failed: Resource temporarily unavailable` message, in the browser it can't open any new tab, etc.

After investigating (https://github.com/raycast/extensions/issues/9410#issuecomment-2558125222), I figured out that the zombie processed were created by the Raycast Helper (Extensions).

<img width="488" alt="Capture d’écran 2024-12-21 à 14 29 46" src="https://github.com/user-attachments/assets/215ff0ce-a89e-4800-bbaa-edd7b0600387" />

And after running `ps -eo pid,ppid,etime,comm | grep defunct`, I found that a defunct process was created every 30 seconds.

By running `log stream --predicate "subsystem == 'com.raycast.macos'" --level debug --style compact`, I found that this was correlated to the Connect to VPN extension.

Checking the changes, I found that there was a change a month ago that trigger something every 30 seconds.

https://github.com/raycast/extensions/commit/989f3ec7e74a3e5400b5f1d6f899aa2df231287f#diff-291e4da920717a0f856d10b8a6d0d2b79a1caf7376c8bc8c34b736af7992002bR28

Looking at this comment https://github.com/raycast/extensions/issues/9410#issuecomment-2387605074, and the code and this code https://github.com/raycast/extensions/commit/989f3ec7e74a3e5400b5f1d6f899aa2df231287f#diff-87c853ea201b0ac6d5e80278f9511e5e411a81ecc6069eb4411fa81ab5fcb449R19-R24, I think that the issue is that there is no cleanup function returned by the `useEffect`.

### Steps To Reproduce

I think the bug is on every Mac, so just install the Connect to VPN extension.

NvL-WEAP:
I had the same issues on all my workstations. After some digging and raycast pointing out to the extensions-diagnostics i figured this VPN extension is causing the defunct zombie processes until the system has no resources anymore.

Only temporary fix is restarting the Mac, but depending on the specs it wil crash in a couple of days.

For now i removed the Connect to VPN extension, which solved the crashes for now.

**Example zombie processes;**
```
[...]

PR #31097, opened 2026-09-14
Update connect-to-vpn extension

rasmusbe:
## Description

This updates the extension to `@raycast/api` 2 and fixes several problems in how service status is read and how the menu bar refreshes.

### Status reads no longer block the extension host

Every `networksetup` call went through `execSync`, so `Promise.all` gave no parallelism and the event loop stood still for the duration. The menu bar command refreshes every 30 seconds and Raycast times a command out after five, which leaves little room. Measured on one machine with nine network services: 1835 ms of blocked event loop per refresh, now 2 ms.

`execSync` came in with #16004, to fix the zombie processes reported in #15990. That is covered here by `execFile`, which runs the tool directly instead of through a shell and so spawns one process per call rather than two, together with a timeout and `SIGKILL` so a call cannot hang around.

### Status comes from scutil

`scutil --nc list` reports every VPN service and its status in one call, where `networksetup -showpppoestatus` needs one call per service. Both read the same `SCNetworkConnection` state, so the status words match apart from the capital letter. On the same machine: 38 ms and two processes against 343 ms and ten, with identical statuses for all nine services. A service `scutil` does not list is either a physical interface, which has no connection status, or one to ask `networksetup` about directly.

### The menu bar no longer signals itself into a refresh loop

It polls a stored timestamp and refreshes every service when it changes, but each of those refreshes wrote the same timestamp again, so it kept refreshing itself roughly every 620 ms for as long as it ran. The guard meant to prevent that tested `launchType`, which says how a command started rather than which command is running, so opening the menu bar got through it and also produced `Command cannot launch itself`.

### Connecting no longer holds the action callback open

It polled for a final status for up to a minute, and running a second action meanwhile made Raycast unload the session while the first was still polling, which failed with `Worker unloaded`. The callback now sends the command and returns, and an effect watches any service left in transition.

### The menu bar icon follows the menu bar again

The `@dark` asset variant is chosen from Raycast's own theme, not from the appearance of the macOS menu bar. With a light Raycast theme and a dark menu bar, Raycast served the black file and the icon disappeared into the background. One monochrome asset tinted with `Color.PrimaryText` resolves against the menu bar itself, so the two `@dark` copies have no job left. Reported in #31304.

Also fixed:

- `networksetup` answers an unknown service name with an empty string and exit code 0, which was cast straight to a status without being checked
- the last used service was stored under its shell-escaped name, so a service with a quote in its name could not be toggled
- the error toast was raised from the render body, so it came back on every re-render
- a service now shows `Connecting…` or `Disconnecting…` while it changes, in both the menu bar and the list

Dependencies go from 295 packages to 177, and from 22 `npm audit` advisories to none.

## Checklist


pernielsentikaer (changes requested):
Hi @rasmusbe 👋

Thanks for improving the VPN status handling 💪

I found a race in `setServiceStatus`: `updateVpnStatus` launches the menu-bar refresh before `networksetup` runs. The launched command can therefore load the old status, treat the transition timestamp as its initial baseline, and receive no further update after the VPN command completes. This means the menu-bar icon, status, and action can remain stale until the next scheduled refresh.

Could you ensure the menu bar consumes and polls the transition correctly, or trigger its refresh after `networksetup` has started or completed?

---

rasmusbe:
Hi @pernielsentikaer, thanks for catching that.

The signal now goes out after `networksetup` has run rather than before it, so the menu bar launches against the requested change instead of the status from before the request. `networksetup` returns before macOS has finished the transition, so a service left connecting or disconnecting is watched from an effect in `useNetworkServices`, and `updateVpnStatus` is sent again the moment it settles. The menu bar gets two signals, one for the transition and one for the end state, and no longer sits on either until the next scheduled refresh.

Three more things from the same round:

- the settle grace period used one shared counter, so a second connection starting while a first was still settling got no grace and could accept the status from before its own request. It is counted per service now.
- the settling loop read the `scutil` map alone, which left services handled through the `networksetup -showpppoestatus` fallback stuck in transition. It goes through `currentStatus`, so those settle too.
- `toggle` passed `disconnected` where the list and the menu bar pass `disconnecting`, so it published a terminal status the instant the command was sent.

The branch also carries a fix for #31304: the menu bar icon shipped a black copy and a white one and picked between them on Raycast's own theme, which is not the appearance of the menu bar. One asset tinted with `Color.PrimaryText` resolves against the menu bar itself.

pernielsentikaer (approved):
Looks good to me, approved 🔥

PR #31102, opened 2026-09-14
[OmniFocus] Add Planned Date support

anant1811:
## Description

Adds **Planned Date** support to the OmniFocus extension. OmniFocus 4.7 introduced planned dates, and OmniFocus exposes them in its scripting dictionary as `planned date` (`task.plannedDate` in JXA), which this extension already uses for defer and due dates.

Changes:

- **Add a Task** form: new *Planned Date* picker between Defer Date and Due Date. It is set on the task in the same JXA script that sets the defer and due dates.
- **List Tasks**: planned dates are read back from OmniFocus and shown as a blue tag accessory, in the detail metadata panel, and in the task detail view.
- Types updated (`plannedDate` on `OmniFocusTask` and `CreateOmniFocusTaskOptions`), CHANGELOG entry added, contributor added.

The `plannedDate` reads are guarded (`task.plannedDate && task.plannedDate()`) so listing still works on OmniFocus versions older than 4.7 where the property does not exist. On those versions the Planned Date field is simply ignored by OmniFocus.

Tested on macOS 26 with OmniFocus 4.9: created a task with a planned date from the form and confirmed the date appears in OmniFocus, and confirmed the value round-trips through the list command.

## Screencast

Straightforward form field addition (one `Form.DatePicker` alongside the existing Defer/Due pickers); no new commands or preferences.

## Checklist

https://claude.ai/code/session_01Fku6VoYPD1cjfnSRdyjmsy


pernielsentikaer (changes requested):
Hi @anant1811 👋

Thanks for adding Planned Date support 💪

OmniFocus’s automation documentation notes that getting or setting `plannedDate` requires the database to have been migrated, and that migration is optional even on OmniFocus 4.7+.

In `list-perspectives-tasks.ts`, `task.plannedDate()` is called for every task. On an unmigrated database it throws, and the surrounding catch returns `null` for the entire task, so the task is removed from the list.

Could you catch failures around the planned-date accessor itself and fall back to `null`, preserving the rest of the task? Please also provide an actionable message when someone selects a Planned Date without a compatible, migrated database.

---

anant1811:
Thanks for the review @pernielsentikaer, good catch on the migration requirement 🙏

Addressed in 09584a1:

- **Listing**: `list-perspectives-tasks.ts` and `list-tasks.ts` now read the planned date through a `safePlannedDate()` helper that wraps the accessor in its own try/catch and returns `null`. On an unmigrated database (or OmniFocus < 4.7) the task is still listed, just without a planned date, instead of being dropped by the outer catch.
- **Adding**: setting `task.plannedDate` is wrapped in try/catch inside the JXA script and the result reports `plannedDateUnsupported`. The task is still created (so nothing the user typed is lost), and the form shows a toast: *"Task added without a Planned Date — Planned dates need OmniFocus 4.7 or later with a migrated database. Migrate it in OmniFocus Settings > Database, then set the date in OmniFocus."*

Lint and build are clean; re-tested the add and list paths against OmniFocus 4.9 with a migrated database. Marking as ready for review.

pernielsentikaer (approved):
Looks good to me, approved 🔥

